PR #72, opened 2026-03-19
fix(store): keep WAL journal mode during bulk write to prevent DB corruption on crash

halindrome:
## Bug

`cbm_store_begin_bulk()` switched the SQLite journal mode from WAL to MEMORY:

```c
// src/store/store.c — before
int cbm_store_begin_bulk(cbm_store_t *s) {
    int rc = exec_sql(s, "PRAGMA journal_mode = MEMORY;");  // ← bug
    ...
}
int cbm_store_end_bulk(cbm_store_t *s) {
    int rc = exec_sql(s, "PRAGMA journal_mode = WAL;");     // ← too late if crashed
    ...
}
```

If the process crashes during bulk write (OOM, SIGKILL, power loss), the in-memory rollback journal is lost. The database file is left in a partially-written state with no way to recover — it is corrupt on next open.

## Fix

Remove the journal mode switch entirely. WAL mode is inherently crash-safe: uncommitted WAL entries are simply discarded on the next open. The performance benefit of bulk mode is fully preserved by `synchronous=OFF` and a 64 MB `cache_size`, both of which are safe under WAL.

```c
// src/store/store.c — after
int cbm_store_begin_bulk(cbm_store_t *s) {
    // WAL mode preserved throughout — crash-safe by design
    int rc = exec_sql(s, "PRAGMA synchronous = OFF;");
    ...
}
int cbm_store_end_bulk(cbm_store_t *s) {
    int rc = exec_sql(s, "PRAGMA synchronous = NORMAL;");
    ...
}
```

## Tests

`tests/test_store_bulk.c` (new file, 3 tests):

| Test | What it proves |
|------|---------------|
| `bulk_pragma_wal_invariant` | `journal_mode` is still `"wal"` after `begin_bulk`, verified via an independent read-only connection — deterministic proof of the fix |
| `bulk_pragma_end_wal_invariant` | `journal_mode` is still `"wal"` after `end_bulk` |
| `bulk_crash_recovery` | Forks a child that enters bulk mode, opens an explicit transaction, writes data, then calls `_exit()` without committing; parent verifies the database opens cleanly and baseline data survives |

All 2033 tests pass.

## Note

This is the C-rewrite equivalent of the same bug that existed in the prior Go implementation. The fix is identical in intent: never leave WAL mode during bulk writes.

### DIFF
--- a/tests/test_store_bulk.c
+++ b/tests/test_store_bulk.c
@@ -141,6 +141,8 @@ TEST(bulk_crash_recovery) {
     ASSERT_GT(pid, 0);
     int status;
     waitpid(pid, &status, 0);
+    /* Confirm child exited normally so the write actually occurred. */
+    ASSERT(WIFEXITED(status) && WEXITSTATUS(status) == 0);
 
     /* Recovery: database must open cleanly. */
     cbm_store_t *recovered = cbm_store_open_path(db_path);


halindrome:
## QA Round 3 — Final Review ✅

**Reviewer:** Claude (Senior C Engineer)
**Commit reviewed:** a0e809d (`fix(store): address QA round 2 findings`)
**Files reviewed:** `src/store/store.c`, `src/store/store.h`, `tests/test_store_bulk.c`, `tests/test_main.c`, `Makefile.cbm`

### Previous Findings — Verified Fixed

| Round | Issue | Status |
|-------|-------|--------|
| R1 | Missing `#ifndef _WIN32` guards for `fork()`/`waitpid()` | **Fixed** — guards on `<unistd.h>`/`<sys/wait.h>` includes, test body, and `RUN_TEST` registration |
| R1 | Crash test asserted `rc == CBM_STORE_OK` for absent "crashed" row | **Fixed** — now uses `ASSERT_NEQ(rc2, CBM_STORE_OK)` |
| R2 | No child exit-status validation after `waitpid()` | **Fixed** — `ASSERT(WIFEXITED(status) && WEXITSTATUS(status) == 0)` |

### Round 3 — Detailed Review

**Store fix (`store.c`):**
- `begin_bulk`: Correctly removed `PRAGMA journal_mode = MEMORY` — WAL is preserved. Only sets `synchronous=OFF` and `cache_size=-65536` (64 MB). Both are safe with WAL.
- `end_bulk`: Correctly removed the now-unnecessary `PRAGMA journal_mode = WAL` restore — it was never leaving WAL in the first place. Restores `synchronous=NORMAL` and `cache_size=-2000`.
- Error handling chain is correct: each PRAGMA checks return code and propagates errors.
- Comment block clearly documents the rationale.

**Header (`store.h`):**
- Doc comments updated to match new semantics. Accurate.

**Test: `bulk_pragma_wal_invariant`:**
- Opens a separate read-only SQLite connection to query `PRAGMA journal_mode`, isolating the check from internal store state. Sound approach.
- Asserts WAL before and after `begin_bulk`. Cleanup is correct (removes `.db`, `-wal`, `-shm`).

**Test: `bulk_pragma_end_wal_invariant`:**
- Verifies WAL persists through a full `begin_bulk` → `end_bulk` cycle. Clean.

**Test: `bulk_crash_recovery`:**
- Correctly writes committed baseline data, then forks a child that enters bulk mode, opens an explicit transaction (`BEGIN IMMEDIATE`), writes an uncommitted row, and calls `_exit(0)`.
- With WAL + explicit transaction, the uncommitted row stays in the WAL as uncommitted frames. On process death, those frames are discarded during WAL recovery. This is the correct WAL crash-safety guarantee.
- Parent validates: (1) child exited cleanly via `WIFEXITED`/`WEXITSTATUS`, (2) DB opens successfully, (3) committed baseline row survives, (4) uncommitted "crashed" row is absent.
- `_exit(0)` (not `exit(0)`) is correct — avoids atexit handlers and stdio flushing that could mask the crash simulation.

**Build integration (`Makefile.cbm`, `test_main.c`):**
- `test_store_bulk.c` added to `TEST_STORE_SRCS`. Suite registered in `test_main.c` with `extern` and `RUN_SUITE`. Correct.

### No Issues Found

All three rounds of findings have been addressed. The fix is minimal, well-documented, and the tests cover the two key properties:
1. WAL journal mode is never switched away during bulk operations (PRAGMA invariant tests)
2. The database recovers correctly after a crash mid-bulk-write (fork/crash simulation)

**Verdict: APPROVED — ready for merge.**

PR #118, opened 2026-03-22
Fix SQL injection, argument injection, and add DeviceTree/VHDL grammars

map588:
Bug/Vulnerability Fixes:
  - SQL injection fix in store.c search — cbm_store_search() exclude_labels were interpolated as '%s' string literals directly into SQL. Replaced with ?N parameterized placeholders bound via BIND_TEXT().

  - SQL injection fix in store.c BFS traversal — cbm_store_bfs() edge type IN clause used '%s' interpolation. Replaced with ?N placeholders with bind_text() calls for both the main BFS query and the follow-up edge collection query.
 
  - Argument injection fix in http_server.c — root_path was embedded raw into a JSON string passed to the indexer subprocess. Added JSON escaping for " and \ characters before interpolation. Buffer enlarged from 1200 to 4096.

Build Fixes:

  - Build fix in pass_githistory.c — GIT_DIFF_OPTIONS_INIT macro triggered -Wmissing-field-initializers under -Werror. Replaced with runtime git_diff_options_init() call per libgit2 best practice.

Added Grammars:

  - Added DeviceTree (.dts/.dtsi/.dtso) and VHDL (.vhd/.vhdl) language support with vendored parsers. VHDL external scanner signatures normalized from Scanner* to void* to conform to the tree-sitter ABI.


map588 (commented):
This PR addresses three security issues found during a review of the C codebase, fixes a build error under -Werror, and adds two new tree-sitter grammars.

  SQL injection in cbm_store_search() (store.c)

The exclude_labels parameter was interpolated directly into SQL as '%s' string literals. Since these values originate from MCP tool calls, which should be treated as untrusted input, a crafted label string could break out of the quotes and execute arbitrary SQL against the backing SQLite database. Fixed by replacing string interpolation with ?N parameterized placeholders bound via sqlite3_bind_text().

  SQL injection in cbm_store_bfs() (store.c)

Same pattern in the BFS traversal's edge type filter. The edge_types array was interpolated as '%s' into an IN (...) clause. Fixed the same way, parameterized placeholders with bound values. Both the main recursive CTE query and the follow-up edge collection query needed binding, since they share the same types_clause.

  Argument injection in index_thread_fn() (http_server.c)

root_path was embedded raw into a JSON string passed to the indexer subprocess. A path containing " or \ characters would break the JSON structure, potentially injecting arbitrary keys/values or causing parse failures downstream. Added JSON escaping for those characters before interpolation.

  Build fix in pass_githistory.c

  GIT_DIFF_OPTIONS_INIT is a compile-time struct initializer macro. When the installed libgit2 has more struct fields than the macro initializes, the compiler emits -Wmissing-field-initializers, which becomes a hard error under -Werror. Replaced with the runtime git_diff_options_init() call, which is what libgit2's own documentation recommends and handles any version of the struct correctly.

  New grammars: DeviceTree and VHDL

  Added tree-sitter grammars for DeviceTree (.dts, .dtsi, .dtso) and VHDL (.vhd, .vhdl), following the existing vendoring pattern. The VHDL external scanner required normalizing four function signatures from Scanner* to void* to conform to the tree-sitter ABI. The DeviceTree grammar worked as-is.

  All 2073 tests pass, including the BFS traversal tests that exercise the parameterized edge type queries.

map588:
Sure

PR #123, opened 2026-03-23
Fix -Werror build failure from GIT_DIFF_OPTIONS_INIT macro

map588:
Replace compile-time GIT_DIFF_OPTIONS_INIT struct initializer with runtime git_diff_options_init() call. The macro triggers -Wmissing-field-initializers when libgit2 has more fields than the macro accounts for, which becomes a hard error under -Werror.


map588:
@DeusData I already did this.  I closed the first PR.  I should have separated them to begin with, apologies.

DeusData:
Thanks @map588 — your fix is identical to PR #91 by @jsyrjala (same file, same line, same replacement). Since #91 was filed first, we'll merge that one. Both of you independently arrived at the correct libgit2-recommended pattern — nice work.

PR #126, opened 2026-03-23
fix(mcp): fix use-after-free in handle_manage_adr get path

### LINKED ISSUE #125: manage_adr get path: use-after-free causes MCP client to hang indefinitely
## Summary

The `manage_adr` tool's `get` path (default mode) contains a use-after-free bug in `handle_manage_adr` (`src/mcp/mcp.c`) that causes the MCP client to hang indefinitely when an `adr.md` file exists.

## Root Cause

```c
// Buggy code in handle_manage_adr (get branch):
char *buf = malloc(sz + 1);
size_t n = fread(buf, 1, sz, fp);
buf[n] = '\0';
(void)fclose(fp);
yyjson_mut_obj_add_str(doc, root_obj, "content", buf);
free(buf);  // BUG: yyjson stores the raw pointer, not a copy
// ... later:
char *json = yy_doc_to_str(doc);  // reads freed memory → garbage JSON
```

`yyjson_mut_obj_add_str` stores the raw pointer without copying. After `free(buf)`, the yyjson document holds a dangling pointer into freed heap memory. When `yy_doc_to_str(doc)` runs, the allocator has overwritten that memory with bookkeeping bytes, producing invalid JSON. `cbm_jsonrpc_format_response` then calls `yyjson_read()` on the corrupted string; the parse fails silently (`res_doc == NULL`), so no `"result"` field is added.

## Symptom

The MCP response becomes `{"jsonrpc":"2.0","id":N}` — missing `result`. The MCP client (Claude Code and others) waits indefinitely for a valid response.

## Affected Scope

- Only the `get` branch (default when `mode` is absent or `"get"`)
- Only when a `.codebase-memory/adr.md` file exists for the project
- Projects without an ADR file are unaffected (`no_adr` path uses string literals)

## Fix

Hoist `adr_buf` to function scope (initialized `NULL`), remove the premature `free`, and free it after `yy_doc_to_str` has serialized the document:

```c
char *adr_buf = NULL; /* freed after yy_doc_to_str — yyjson holds pointer, not copy */
// ...
adr_buf = malloc(sz + 1);
// ...
yyjson_mut_obj_add_str(doc, root_obj, "content", adr_buf);
/* do NOT free adr_buf here: yyjson stores the pointer, not a copy */
// ...
char *json = yy_doc_to_str(doc);
yyjson_mut_doc_free(doc);
free(adr_buf); /* safe to free now — doc has been serialized */
```

Total diff: ~7 lines in `handle_manage_adr`.

## Regression Test

A regression test `tool_manage_adr_get_with_existing_adr` is included that:
1. Creates a temp directory with `.codebase-memory/adr.md`
2. Registers it as a project in an in-memory store
3. Calls `manage_adr(mode="get")` via `cbm_mcp_server_handle`
4. Asserts the JSON-RPC response contains `"result"` with the ADR content

This test fails on `main` (before the fix) and passes after.

halindrome:
Fixes #125.

## Summary

- `yyjson_mut_obj_add_str` stores the raw pointer; `free(buf)` before `yy_doc_to_str` leaves a dangling pointer in the yyjson document
- `yy_doc_to_str` reads freed heap memory → garbage JSON
- `cbm_jsonrpc_format_response` fails to parse the result → `"result"` field is omitted from the JSON-RPC response
- MCP client hangs indefinitely waiting for a valid response

## Changes

- `src/mcp/mcp.c`: hoist `adr_buf` to function scope, remove premature `free`, defer `free(adr_buf)` until after `yy_doc_to_str` serializes the document (~7 lines)
- `tests/test_mcp.c`: add regression test `tool_manage_adr_get_with_existing_adr`

## Test plan

- [x] `tool_manage_adr_get_with_existing_adr` FAILS on `main` (before fix): `"result"` field absent from JSON-RPC response
- [x] `tool_manage_adr_get_with_existing_adr` PASSES after fix: response contains `"result"` with ADR content
- [x] `tool_manage_adr_no_project` (no-ADR path) still passes — unaffected code path
- [x] Full test suite: 2043 tests pass, 0 failures (ASan + UBSan)

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -1897,6 +1897,7 @@ static char *handle_manage_adr(cbm_mcp_server_t *srv, const char *args) {
     char adr_path[4096];
     snprintf(adr_path, sizeof(adr_path), "%s/adr.md", adr_dir);
 
+    char *adr_buf = NULL; /* freed after yy_doc_to_str — yyjson holds pointer, not copy */
     yyjson_mut_doc *doc = yyjson_mut_doc_new(NULL);
     yyjson_mut_val *root_obj = yyjson_mut_obj(doc);
     yyjson_mut_doc_set_root(doc, root_obj);
@@ -1937,12 +1938,12 @@ static char *handle_manage_adr(cbm_mcp_server_t *srv, const char *args) {
             (void)fseek(fp, 0, SEEK_END);
             long sz = ftell(fp);
             (void)fseek(fp, 0, SEEK_SET);
-            char *buf = malloc(sz + 1);
-            size_t n = fread(buf, 1, sz, fp);
-            buf[n] = '\0';
+            adr_buf = malloc(sz + 1);
+            size_t n = fread(adr_buf, 1, sz, fp);
+            adr_buf[n] = '\0';
             (void)fclose(fp);
-            yyjson_mut_obj_add_str(doc, root_obj, "content", buf);
-            free(buf);
+            yyjson_mut_obj_add_str(doc, root_obj, "content", adr_buf);
+            /* do NOT free adr_buf here: yyjson stores the pointer, not a copy */
         } else {
             yyjson_mut_obj_add_str(doc, root_obj, "content", "");
             yyjson_mut_obj_add_str(doc, root_obj, "status", "no_adr");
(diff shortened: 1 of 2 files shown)


DeusData:
Merged to main. Fixes #125. Thanks @halindrome — same yyjson pointer-not-copy pattern we fixed in search_code v2 this session. The regression test via full JSON-RPC path is exactly the right approach. 2090 tests passing.

PR #131, opened 2026-03-23
fix(cbm): convert 7 O(N²) import extractors to O(N) TSTreeCursor iteration

### LINKED ISSUE #130: 7 language-specific import extractors: O(N²) ts_node_child() loop hangs on large AST roots
## Summary

Seven language-specific import extractors in `internal/cbm/extract_imports.c` use an
indexed child-access loop (`ts_node_child(ctx->root, i)` in a counted for-loop) at the
root AST level. `ts_node_child` walks the child linked-list from index 0 to `i` on every
call, making the total complexity O(N²) for N root-level declarations.

**Affected functions:**
- `parse_go_imports`
- `parse_python_imports`
- `parse_java_imports`
- `parse_rust_imports`
- `parse_c_imports`
- `parse_ruby_imports`
- `parse_lua_imports`

## Impact

On files with thousands of top-level nodes (large generated files, minified code,
auto-generated `.d.ts` declarations), the indexer hangs at 100% CPU indefinitely.

At 10,000 root children: ~50M iterator steps, causing a hang that never completes.
This is the same root cause as the previously-fixed `parse_generic_imports` (PR #107).

## Fix

Replace the indexed for-loop with `TSTreeCursor` do-while iteration:

\`\`\`c
/* BEFORE — O(N²) */
uint32_t count = ts_node_child_count(ctx->root);
for (uint32_t i = 0; i < count; i++) {
    TSNode node = ts_node_child(ctx->root, i);

/* AFTER — O(N) */
TSTreeCursor cursor = ts_tree_cursor_new(ctx->root);
if (!ts_tree_cursor_goto_first_child(&cursor)) {
    ts_tree_cursor_delete(&cursor);
    return;
}
do {
    TSNode node = ts_tree_cursor_current_node(&cursor);
} while (ts_tree_cursor_goto_next_sibling(&cursor));
ts_tree_cursor_delete(&cursor);
\`\`\`

`ts_tree_cursor_goto_next_sibling()` maintains internal cursor state and is O(1) per step.

## Test coverage added

- 6 unit tests (one per language) verifying correct import extraction
- 1 stress test (`import_stress_go`) with 5,000 synthetic imports that completes in
  under 5 seconds with the fix and hangs indefinitely without it

## Expected speedup

~5,000x on files with 10,000+ top-level declarations (same profile as PR #107 fix).

halindrome:
## Summary

Converts the root-level iteration loop in seven language-specific import extractors from
O(N²) indexed child access to O(N) `TSTreeCursor` sibling iteration, matching the pattern
applied to `parse_generic_imports` in PR #107.

Closes #130

**Affected functions:** `parse_go_imports`, `parse_python_imports`, `parse_java_imports`,
`parse_rust_imports`, `parse_c_imports`, `parse_ruby_imports`, `parse_lua_imports`.

## Root cause

`ts_node_child(root, i)` walks the child linked-list from index 0 to `i` on every call.
A counted for-loop over N root children calls this N times → O(N²) total.
At 10,000 root children: ~50M iterator steps, causing an indefinite CPU hang.

## Fix

Replace with `TSTreeCursor` do-while iteration. `ts_tree_cursor_goto_next_sibling()` is
O(1) per step. Only the iteration mechanism changes; all loop body logic, nested loops,
and control flow are preserved exactly.

## Test plan

- [ ] 6 new unit tests added: `go_imports`, `java_imports`, `rust_imports`, `c_imports`,
      `ruby_imports`, `lua_imports` — each verifies correct import extraction
- [ ] `import_stress_go` stress test: 5,000 synthetic Go imports, asserts count == 5000,
      completes in under 5 seconds (would hang indefinitely without this fix)
- [ ] All existing tests pass (full test suite run via `scripts/test.sh`)
- [ ] `grep -n 'ts_node_child(ctx->root' internal/cbm/extract_imports.c` returns no matches

### DIFF
--- a/tests/test_extraction.c
+++ b/tests/test_extraction.c
@@ -1671,6 +1671,104 @@ TEST(js_imports) {
     PASS();
 }
 
+TEST(go_imports) {
+    CBMFileResult *r =
+        extract("package main\n\nimport \"fmt\"\nimport (\n    \"os\"\n    net \"net/http\"\n)\n",
+                CBM_LANG_GO, "t", "main.go");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    ASSERT_GT(r->imports.count, 0);
+    ASSERT(has_import(r, "fmt"));
+    cbm_free_result(r);
+    PASS();
+}
+
+TEST(java_imports) {
+    CBMFileResult *r =
+        extract("import java.util.List;\nimport java.util.ArrayList;\nimport static java.lang.Math.PI;\n"
+                "public class Foo {}\n",
+                CBM_LANG_JAVA, "t", "Foo.java");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    ASSERT_GT(r->imports.count, 0);
+    ASSERT(has_import(r, "java.util.List"));
+    cbm_free_result(r);
+    PASS();
+}
+
+TEST(rust_imports) {
+    CBMFileResult *r =
+        extract("use std::collections::HashMap;\nuse std::io::{self, Write};\nuse serde::Serialize;\n"
+                "fn main() {}\n",
+                CBM_LANG_RUST, "t", "main.rs");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    ASSERT_GT(r->imports.count, 0);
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks @halindrome — clean fix, merged! Appreciate the matching issue + PR with clear before/after analysis.

PR #132, opened 2026-03-23
fix(watcher): detect submodule dirty state + touch_project tool

### LINKED ISSUE #71: Submodule visibility gap and poll-interval scaling in git-strategy watcher
## Summary

The git-strategy watcher in `internal/watcher/watcher.go` has two related problems that compound for large projects with git submodules.

## Problem 1: Submodule working-tree changes are invisible until committed

The `gitSentinel` function runs `git status --porcelain` on the project root:

```go
cmd := exec.CommandContext(ctx, "git", "--no-optional-locks", "-C", rootPath,
    "status", "--porcelain", "--untracked-files=normal")
```

For a repo with submodules, this shows submodule directories as modified **only after a commit is made inside the submodule** (causing the submodule HEAD to diverge from what the parent records). Uncommitted edits to files _inside_ a submodule are invisible to the parent's `git status` — the parent only sees the submodule as a single opaque directory entry.

This means:
- Agent edits files in `apps/mobile/` → parent git status shows nothing → no reindex triggered
- Agent commits in `apps/mobile/` → parent git status shows `apps/mobile` as modified → reindex eventually triggered (up to 60s later)
- Unstaged changes inside submodules are never visible to the watcher

## Problem 2: Poll interval backs off to 60s for large repos

The adaptive interval formula:

```go
func pollInterval(fileCount int) time.Duration {
    ms := 5000 + (fileCount/500)*1000
    if ms > 60000 { ms = 60000 }
    return time.Duration(ms) * time.Millisecond
}
```

A monorepo with 6 submodules can easily have 25,000+ files, which pushes the interval to the 60-second cap. Combined with Problem 1, this means:
- Uncommitted changes in submodules: **never detected**
- Committed changes in submodules: **up to 60s delay** before reindex

### Option A: Add `--recurse-submodules` to `gitSentinel`

```go
cmd := exec.CommandContext(ctx, "git", "--no-optional-locks", "-C", rootPath,
    "status", "--porcelain", "--untracked-files=normal", "--recurse-submodules")
```

This makes working-tree changes inside submodules visible to the parent. The downside is that `git status --recurse-submodules` is significantly slower for large repos.

### Option B: Run per-submodule git sentinels

Parse `.gitmodules` and run a separate `gitSentinel` for each submodule path. Allows the same efficient `git status` per submodule without recursion overhead. More complex to implement.

### Option C: Signal / touch mechanism

Provide a way for external processes (hooks, scripts) to signal the daemon that a reindex is needed immediately — e.g., a sentinel file path or a lightweight HTTP endpoint. This would allow Claude Code hooks to trigger reindexing on demand without waiting for the next poll cycle.

## Context

This was discovered while building a Claude Code hook layer on top of codebase-memory-mcp for a monorepo with 6 submodules (mobile app, web app, REST API, SSE service, chatbot orchestrator, MCP server). The 60-second lag and submodule blindspot make the auto-sync feature unreliable for this use case — we work around it by calling `index_repository` explicitly after commits, but the gap is still visible.

## Environment

- codebase-memory-mcp: latest (reviewed source at commit on `main`)
- Project type: git repo with 6 submodules, ~25k+ files
- Watch strategy selected: `git` (git repo detected correctly)

halindrome:
## Summary

Fixes #71 — the git-strategy watcher was blind to uncommitted changes inside git submodules.

### Root cause

`git_is_dirty()` in `src/watcher/watcher.c` runs `git status --porcelain` on the project root. For repos with submodules, the parent's `git status` only sees submodule HEAD pointer changes — uncommitted file edits inside a submodule are invisible. This means the watcher never triggers a reindex for submodule working-tree changes, even though the indexer's `walk_dir()` already walks into submodule directories and indexes their contents.

### Fix: two-part approach (Option A + Option C from issue discussion)

**1. Submodule dirty detection** (`d561232`)

Added a two-stage check in `git_is_dirty()`:
- First: run the existing `git status --porcelain` on the parent (unchanged behavior)
- If clean: run `git submodule foreach --quiet --recursive 'git status --porcelain'` to detect uncommitted changes inside submodules

Note: `--recurse-submodules` was the original plan but is not supported on Apple Git (2.50.1). The `submodule foreach` approach is portable across all git versions and functionally equivalent — `git_is_dirty()` is a boolean check (any output = dirty), so the two-stage approach produces the same result.

**2. `touch_project` MCP tool** (`14d480b`)

Exposed the existing `cbm_watcher_touch()` function as an MCP tool. This resets the adaptive poll timer so the next watcher tick runs `check_changes()` immediately, allowing git hooks or editor hooks to trigger on-demand reindex without waiting for the poll interval (up to 60s for large repos).

The adaptive interval provides natural debounce — `index_fn()` is synchronous, and `next_poll_ns` resets after every reindex, so rapid-fire touch calls cannot cause thrashing.

**3. CLI help update** (`b10df9a`)

Added `touch_project` to CLI help text and usage output.

### Test coverage

- `watcher_detects_submodule_dirty`: creates parent + submodule repos, modifies a file inside the submodule without committing, verifies the watcher detects it as dirty
- `watcher_touch_resets_interval`: verifies `cbm_watcher_touch()` causes immediate poll on next cycle
- `mcp_tools_list`: updated to verify `touch_project` appears in tool list (15 tools)
- `mcp_touch_project_no_watcher`: verifies graceful error when watcher is NULL (CLI mode)

All 2096 tests pass.

Closes #71

### DIFF
--- a/.vbw-planning/phases/21-language-support-tool/21-01-PLAN.md
+++ b/.vbw-planning/phases/21-language-support-tool/21-01-PLAN.md
@@ -1,0 +1,145 @@
+---
+plan: 21-01
+phase: 21
+title: "Validate tool-routing behavior and improve CLAUDE.md guidance"
+status: ready
+date: 2026-03-24
+commit_scope: docs
+---
+
+# Plan 21-01: Validate Tool-Routing Graceful Degradation + Improve CLAUDE.md
+
+## Goal
+
+Validate the Scout research finding that a `list_supported_languages` tool is NOT
+needed. CMM already handles misrouted calls gracefully (empty results, no errors).
+After hands-on validation, improve CLAUDE.md documentation to close the minor gaps
+identified in the research.
+
+## Background (from research)
+
+- All structural queries (`search_graph`, `get_code_snippet`, `trace_call_path`)
+  return `{"total":0, "results":[]}` for text-only or non-indexed file types — no
+  errors, no crashes.
+- JSON, YAML, and Markdown ARE indexed by CMM (Tree-Sitter parsers extract Module
+  and Variable nodes), so `search_code` works on them. Current CLAUDE.md guidance
+  says to use Read for these, missing the `search_code` opportunity.
+- The cost of a `list_supported_languages` tool (~10ms per session + complexity)
+  exceeds the cost of try-and-fallback (~1ms empty query).
+- User-configured extensions (Phase 5) create a minor knowledge gap, but graceful
+  fallback handles it.
+
+## File Ownership (no overlapping edits between tasks)
+
+| Task | Files Modified |
+|------|---------------|
+| T1 | None (validation only — reads + CMM tool calls) |
+| T2 | `.claude/rules/global-claude-md.md` |
+
+T1 and T2 can run in parallel — T1 produces validation evidence, T2 edits docs.
+T2 does not depend on T1 because the research already provides sufficient evidence
+for the doc changes. If T1 discovers unexpected issues, a pivot task would be added.
+
+---
+
+## Tasks
+
+### T1 — Hands-on validation of graceful degradation
+
+**Wave:** 1 (no deps)
+**Files:** None modified (read-only validation)
+
+**Steps:**
+
+1. Call `search_graph` with a `file_pattern` targeting a non-indexed extension
+   (e.g., `*.txt` or `*.xyz`) on the current CMM project. Verify the response is
+   `{"total":0, "results":[]}` or equivalent empty result — not an error.
+
+2. Call `search_graph` with `label_filter="Function"` on a YAML or JSON file in
+   the project. Verify it returns empty results (these file types have no
+   Function nodes).
+
+3. Call `search_code` with a known string pattern scoped to `*.yaml` or `*.json`
+   files. Verify it returns matches (proving text search works on indexed
+   text-only formats).
+
+4. Call `get_code_snippet` with a qualified name that does not exist (e.g.,
+   `nonexistent_module::fake_function`). Verify graceful empty/error response.
+
+5. Document results: write a brief validation summary as a comment in the
+   research file or as a standalone validation note.
+
+**Acceptance criteria:**
+- All four CMM calls return graceful responses (empty results or clear error
+  messages, never crashes or unhandled exceptions)
+- Validation results confirm the research finding: no tool needed
+- If ANY call produces an unexpected error or crash, escalate immediately —
+  this would trigger a pivot to tool implementation
+
+**Commit:** `docs(phase-21): add tool-routing validation results`
+
+---
+
+### T2 — Improve CLAUDE.md guidance for text-only file types and search_code
+
+**Wave:** 1 (no deps)
+**Files:** `.claude/rules/global-claude-md.md`
+
+**Changes:**
+
+1. Update the "When Read is Correct" section to clarify that text-only formats
+   (JSON, YAML, TOML, Markdown) are indexed and searchable via `search_code`,
+   while `Read` remains correct for full-file context:
+
+```markdown
+### When Read is Correct
+
+Use `Read` directly when:
+- Non-code files (JSON, YAML, TOML, config, HTML templates, Markdown, .env)
+  **Note**: After indexing, these formats are searchable via `search_code` for
+  pattern matching (e.g., finding all occurrences of a config key across YAML
+  files). Use `Read` when you need full-file context; use `search_code` when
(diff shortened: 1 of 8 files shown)


halindrome:
### QA Round 1 Report

**Status:** ISSUES FOUND

#### Files Reviewed

**1. `src/watcher/watcher.c` (lines 118-175)**

The two-stage `git_is_dirty()` implementation is clean and follows existing code patterns. The submodule foreach command is well-constructed:
- `--no-optional-locks` prevents lock contention (consistent with existing git commands)
- `--quiet` suppresses "Entering 'sub'" prefix messages so only `git status` output appears
- `--recursive` handles nested submodules
- `2>/dev/null` on both inner and outer commands suppresses errors gracefully
- Early return at line 147 (`if (dirty) return true;`) avoids unnecessary submodule scanning when parent is already dirty

**Edge cases verified:**
- No submodules: `git submodule foreach` exits 0 with no output (returns false correctly)
- Deinitialized submodules: Same as above
- `ignore=dirty` in `.gitmodules`: Parent `git status` hides changes, but `submodule foreach` detects them (core value of this PR)

**2. `src/mcp/mcp.c` (lines 323-328, 2563-2588, 2642-2644)**

`handle_touch_project` is well-structured with proper input validation and memory management. Follows the exact same pattern as all other tool handlers.

**3-4. `src/main.c`, `src/cli/cli.c`** — Trivially correct help text additions.

**5. `tests/test_watcher.c` (lines 1269-1388)** — Two new tests plus style reformattings.

**6. `tests/test_mcp.c` (lines 710-736)** — Two new error-path tests, tool count incremented to 15.

#### Issues Found

| ID | Severity | File | Line | Description |
|----|----------|------|------|-------------|
| Q1-01 | MAJOR | tests/test_watcher.c | 1269-1331 | **Test does not exercise the new code path.** `watcher_detects_submodule_dirty` modifies a file inside a submodule without setting `ignore=dirty` in `.gitmodules`. The parent's first-stage `git status --porcelain` already reports ` M sub` for this case, so the test passes even without the new `submodule foreach` code. The test should add `git config -f .gitmodules submodule.sub.ignore dirty` (and commit it) before making the submodule dirty, forcing the second stage to be the only detection mechanism. |
| Q1-02 | MINOR | tests/test_watcher.c | 1320 | **Stale comment.** Comment says `--recurse-submodules should detect the dirty submodule` but the implementation uses `submodule foreach`, not `--recurse-submodules`. |
| Q1-03 | MINOR | src/mcp/mcp.c | 2573-2587 | **Silent success for nonexistent project.** `handle_touch_project` returns `{"status":"touched"}` even if the project doesn't exist in the watch list. Consider returning an error or `{"status":"not_found"}`. |
| Q1-04 | ADVISORY | src/watcher/watcher.c | 289-298 | **Thread safety — pre-existing pattern.** `cbm_watcher_touch` accesses `w->projects` from the MCP thread while the watcher thread reads/writes it in `poll_once`. Same pattern as existing `cbm_watcher_watch` calls. Not introduced by this PR but worth noting for a future fix. |
| Q1-05 | ADVISORY | tests/test_mcp.c | 710-736 | **No happy-path MCP integration test.** Tests only cover error paths. The watcher tests cover `cbm_watcher_touch` directly, so functional coverage is not gapped. |

#### Test Coverage Assessment

**Covered:** Error paths for touch_project MCP tool, touch resets poll interval, tool count assertion, tools/list contains touch_project.

**Gaps:**
1. **(MAJOR)** No test isolates the `submodule foreach` code path — see Q1-01
2. (Minor) No test for touching a nonexistent project
3. (Minor) No test with nested submodules
4. (Advisory) No test with uninitialized submodule

#### Summary

The implementation is **correct and well-crafted**. The two-stage approach is sound, memory management is clean, error handling follows established patterns, and all 2096 tests pass.

The **one blocking issue (Q1-01)** is that the submodule detection test doesn't actually exercise the new `submodule foreach` code path — the test scenario is already caught by the first-stage `git status`. Fix: add `git config -f .gitmodules submodule.sub.ignore dirty` to test setup.

DeusData:
Thanks @halindrome! The submodule dirty detection fix is merged in commit 7f79ce5 with your co-authorship credit.

We took just the bug fix (the `git submodule foreach` check in `git_is_dirty()`) and skipped the `touch_project` tool for now — we want to keep the tool count at 14 to avoid tool inflation for LLMs parsing the tool list at session start. The adaptive poll interval already catches changes within 5-60s which is sufficient for most workflows.

If demand for `touch_project` comes up from users we can add it later. Closing since the bug fix is merged — thank you!

PR #148, opened 2026-03-26
feat: add blast radius analysis with risk scoring

maplenk:
## Summary

New `get_impact_analysis` tool that answers "what breaks if I change this function?" with risk-tiered results and test coverage detection.

> **Stacked PR:** This builds on #147 (PageRank + Architecture). Please merge #147 first — this PR's unique changes are the 2 commits after PR 1's HEAD.

### New tool: `get_impact_analysis`

**Parameters:** `project`, `symbol`, `depth` (default 3), `include_tests` (default true), `max_tokens` (default 2000)

### Key features

- **Custom multi-edge-type BFS**: follows CALLS (inbound) + HANDLES (outbound) for callables, HTTP_CALLS + ASYNC_CALLS for routes. Different from generic `cbm_store_bfs()`.
- **Depth grouping**: hop 1 = direct (will break), hop 2-3 = indirect (may break), hop 4+ = transitive (ripple risk)
- **Risk scoring** (low/medium/high): based on direct caller count, entry-point presence in radius, PageRank percentile, and test coverage
- **Symbol disambiguation**: when multiple nodes share a name, auto-picks by non-test file, then PageRank, then in-degree
- **Affected test detection**: filters BFS results by test file path
- **Batch node fetching**: single chunked SQL query (batches of 900) instead of per-node lookups

### Tests

New tests in `test_store_search.c` (risk tiers, routes+tests, disambiguation) and `test_mcp.c` (basic, missing symbol, disambiguation, include_tests, route typing).

---

*Part 2 of a 4-PR series. Depends on #147.*

---
*Built with [OpenAI Codex](https://openai.com/index/codex/) and [Claude Code](https://docs.anthropic.com/en/docs/claude-code).*

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -1918,7 +1918,14 @@ static char *handle_get_impact_analysis(cbm_mcp_server_t *srv, const char *args)
         free(project);
         return cbm_mcp_text_result("symbol is required", true);
     }
-    REQUIRE_STORE(store, project);
+    if (!store) {
+        char *_err = build_project_list_error("project not found or not indexed");
+        char *_res = cbm_mcp_text_result(_err, true);
+        free(_err);
+        free(project);
+        free(symbol);
+        return _res;
+    }
 
     char *not_indexed = verify_project_indexed(store, project);
     if (not_indexed) {
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -3171,6 +3171,10 @@ static const impact_cached_node_t *impact_find_cached_node(const impact_cached_n
     return NULL;
 }
 
+/* Chunk size for IN(...) batches — stays well under SQLite's default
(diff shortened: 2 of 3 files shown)


DeusData:
Thanks @maplenk! The blast radius / test coverage idea was useful.

We took the core insight — "which tests are in the blast radius?" — and added it to the existing `trace_path` tool in commit f25dafa (with your co-authorship credit):

- `include_tests=false` (default): test files are filtered from results
- `include_tests=true`: test nodes appear with `is_test=true` marker
- Combined with `risk_labels=true` + `direction="inbound"`, this covers the blast radius use case without a new tool

This avoids the need for a separate `get_impact_analysis` tool (keeps tool count at 14) and builds on the existing BFS infrastructure rather than reimplementing it.

Closing since the useful parts are merged. Thank you for the contribution!

PR #172, opened 2026-03-29
fix: skip update when already on latest version (#142)

### LINKED ISSUE #142: Update command re-downloads/reinstalls even when already on latest version
## Description

The `update` command always downloads and installs the latest version, even when the currently installed version is already up to date. This causes unnecessary network traffic, download time, and disk I/O on every update invocation.

## Expected Behavior

Before downloading, the update command should:
1. Fetch the latest available version (e.g., from the registry or release API)
2. Compare it against the currently installed version
3. Skip the download/install if versions match, printing a message like `Already up to date (v1.x.x).`
4. Only proceed with the download and install if a newer version is available

## Current Behavior

The update always downloads and reinstalls regardless of whether the installed version is already the latest.

## Suggested Implementation

```
current=$(get_installed_version)
latest=$(get_latest_version)

if [ "$current" == "$latest" ]; then
  echo "Already up to date ($current). Nothing to do."
  exit 0
fi

# proceed with download and install
```

The exact approach depends on the package manager / installation method used, but the principle is the same: **check before downloading**.

## Benefits

- Faster no-op updates (no unnecessary network calls or disk writes)
- Reduced bandwidth usage
- Clearer feedback to the user about whether an update actually occurred

dLo999:
Closes #142

## Summary

Adds a version check to the `update` command that skips the download/install when the binary is already on the latest (or newer) version. Uses the GitHub `/releases/latest` redirect header to determine the latest tag — no JSON library needed in cli.c.

## Changes

- `src/cli/cli.c`: Added `fetch_latest_tag()` helper that follows the GitHub releases redirect via `curl -sfI` and extracts the tag from the `Location:` header
- `src/cli/cli.c`: Added version comparison block at the start of `cbm_cmd_update()` using the existing `cbm_compare_versions()` infrastructure
- Added `--force` flag to bypass the version check when needed
- Network failures produce a warning but don't block the update (graceful degradation)

## Behavior

| Scenario | Output |
|----------|--------|
| Already on latest | `Already up to date (v0.5.7).` → exit 0 |
| Ahead of latest (dev build) | `Already up to date (v0.5.8-dev, ahead of latest v0.5.7).` → exit 0 |
| Newer available | `Update available: v0.5.6 -> v0.5.7` → continues with download |
| Network failure | `warning: could not check latest version` → continues with download |
| `--force` flag | Skips check entirely → always downloads |
| `CBM_DOWNLOAD_URL` set | Skips check (testing path) |

## Test results

**Build:** Compiles cleanly on macOS (Apple Clang, arm64) with `make -f Makefile.cbm cbm` — no extra dependencies needed (zlib ships with Xcode CLT).

**Test suite:** 2741 tests passed, 0 regressions.

**Manual verification (all with `--dry-run --standard`):**

| Test | Version | Result |
|------|---------|--------|
| Already on latest | `CBM_VERSION=v0.5.7` (== latest) | Prints `Already up to date (v0.5.7).`, exits 0 immediately |
| Update available | `CBM_VERSION=v0.5.6` (< latest) | Prints `Update available: v0.5.6 -> v0.5.7`, continues to download |
| Force bypass | `CBM_VERSION=v0.5.7` + `--force` | Skips version check, proceeds directly to download |

Generated with [agent-team](https://github.com/dLo999/agent-team) via /issue

### DIFF
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -2956,10 +2956,58 @@ int cbm_cmd_uninstall(int argc, char **argv) {
 
 /* ── Subcommand: update ───────────────────────────────────────── */
 
+/*
+ * Fetch the latest release tag from GitHub by following the redirect of
+ * the /releases/latest URL.  Returns a malloc'd string like "v0.5.7" or
+ * NULL on any failure (network error, unexpected header format, etc.).
+ * Caller must free() the returned string.
+ */
+static char *fetch_latest_tag(void) {
+    /* curl -sfI prints only headers; -L is intentionally omitted so we
+     * see the Location header of the first redirect. */
+    // NOLINTNEXTLINE(bugprone-command-processor,cert-env33-c)
+    FILE *fp = cbm_popen(
+        "curl -sfI https://github.com/DeusData/codebase-memory-mcp/releases/latest 2>/dev/null",
+        "r");
+    if (!fp) {
+        return NULL;
+    }
+
+    char line[512];
+    char *tag = NULL;
+    while (fgets(line, sizeof(line), fp)) {
+        /* Header names are case-insensitive; curl lowercases them on most
+         * platforms but we match both forms to be safe. */
+        if (strncasecmp(line, "location:", 9) != 0) {
+            continue;
+        }
+        /* Find the last '/' — the tag name follows it */
+        char *slash = strrchr(line, '/');
+        if (!slash) {
+            break;
+        }
+        slash++; /* skip the '/' itself */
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks @dLo999! Merged in commit 54c0325 with your co-authorship credit.

Applied the version check via GitHub redirect header before downloading. Added --force bypass flag. Graceful degradation on network failure. Extracted into `check_already_latest` helper for CC compliance.

Fixes #142. Thank you!

PR #184, opened 2026-03-29
feat: resolve ES/TS package specifiers for IMPORTS edges (#180)

### LINKED ISSUE #180: ES/TS module specifiers produce zero IMPORTS edges — pipeline resolves by name only
Love the v0.5.7 release — the Windows fixes have been solid. Spent some time digging into the TypeScript indexing for our monorepo and found what I think is a notable gap. Wanted to share the analysis in case it's useful.

---

### Summary

In a TypeScript pnpm monorepo, **zero real IMPORTS edges are created** despite `walk_es_imports()` correctly extracting all import statements. The pipeline's `fqn_module()` treats npm package specifiers as file paths, producing QNs that match no graph nodes. Cross-file relationships fall back entirely to the global name registry (`unique_name` strategy), which works only when function names are globally unique.

### Environment

- codebase-memory-mcp v0.5.7, Windows amd64
- TypeScript pnpm monorepo: ~7 packages, 2 apps, ~300 source files
- Indexed: ~3,000 nodes, ~3,800 edges

#### IMPORTS edges: 0 real, 3 false positives

The 3 reported IMPORTS edges are all `vitest.config.ts -> vitest.config.ts` — the import `from 'vitest/config'` produced QN `{project}.vitest.config`, which accidentally matched a different file with the same name. No actual cross-package import was captured.

**Root cause** (`pass_definitions.c:297`):

```c
char *target_qn = cbm_pipeline_fqn_module(ctx->project_name, imp->module_path);
const cbm_gbuf_node_t *target = cbm_gbuf_find_by_qn(ctx->gbuf, target_qn);
```

For `import { validateDatasetName } from '@myorg/storage-utils'`:
- Extracted: `module_path = "@myorg/storage-utils"` (correct)
- Computed QN: `myproject.@myorg.storage-utils` (no such node)
- Actual target QN: `myproject.packages.storage-utils.src.index`
- Result: `target` is NULL, edge silently dropped

`fqn_module()` in `fqn.c` splits on `/` and joins with `.` — it treats the module path as a relative file path. It has no awareness of:
- `package.json` `name` field -> directory mapping
- pnpm/npm/yarn workspace resolution
- `tsconfig.json` `paths` or `baseUrl`
- Barrel re-exports (`index.ts`)

#### CALLS edges work — by coincidence, not imports

711 CALLS edges were created. Cross-package calls resolve correctly when function names are globally unique:

| Call site (app-server) | Target (storage-utils) | Why it resolved |
|---|---|---|
| `normalizeTableName` -> `validateDatasetName` | `storage-utils.src.validation.validateDatasetName` | Only 1 function named `validateDatasetName` project-wide |
| `db-adapter` -> `buildSelectQuery` | `storage-utils.src.query-builder.buildSelectQuery` | Only 1 function named `buildSelectQuery` project-wide |

The resolution chain in `cbm_registry_resolve()` (`registry.c:414-452`):
1. **import_map**: always misses (import map is empty because IMPORTS edges failed)
2. **same_module**: misses for cross-package calls
3. **unique_name**: resolves if exactly 1 function has that name project-wide — **this is what works**
4. **suffix_match**: picks by QN string distance when multiple candidates exist

This means cross-package CALLS edges are correct only when names don't collide. If two packages defined `validate()`, the pipeline would pick whichever has the closest QN string, not the one actually imported.

#### USAGE edges: mostly false positives

24 USAGE edges were created. Examples:
- `handleCreate` -> `storage` — points to `config.storage` (a TOML config section), not the database client instance the code actually references
- `AppError` references resolved correctly (globally unique name)

### The extraction works — the resolution doesn't

To be clear: `walk_es_imports()` in `extract_imports.c:248-351` **correctly parses all ES import syntax** — default imports, named imports, namespace imports, re-exports, `require()`. The tree-sitter TypeScript grammar produces clean AST nodes. The data is there.

The gap is between extraction and edge creation: `fqn_module()` cannot map npm package specifiers to graph nodes because it lacks a module resolution step.

### Proposed approach

A **package map** built at index start, before `pass_definitions` runs:

1. Scan for `package.json` files in the repo. For each, read the `name` field and compute the entry point path (from `main`, `exports`, or default `index.ts`/`index.js`).
2. Build a lookup: `"@myorg/storage-utils"` -> `"packages/storage-utils/src/index"` (the relative path that `fqn_module()` can resolve)
3. In `pass_definitions.c`, before calling `fqn_module(module_path)`: check if `module_path` is a package specifier (starts with `@` or doesn't start with `.`/`/`). If so, look it up in the package map and substitute the resolved path.

This approach:
- Doesn't touch tree-sitter extraction
- Doesn't change the registry or call resolution
- Uses yyjson (already vendored) for `package.json` parsing
- Scoped to the pipeline layer between extraction and graph building
- Would fix IMPORTS edges, which cascades to fix import_map resolution in `pass_calls.c`, which fixes USAGE edges

For relative imports (`./utils/mapping`), the resolver would need to join the importing file's directory with the relative path — simpler than package resolution.

### Impact

Without IMPORTS edges, several downstream features degrade silently:
- `trace_call_path` dead-ends at package boundaries
- `get_architecture(aspects: ["layers", "boundaries"])` can't compute real dependency layers
- `search_graph(max_degree=0)` for dead code detection misses cross-package callers
- Louvain community detection clusters are based on name-match CALLS, not actual import structure

This affects all JS/TS monorepos (npm workspaces, pnpm, yarn), not just TypeScript specifically. Single-package projects with only relative imports may also be affected (untested).

### Reproduction

1. Create a pnpm workspace with 2+ packages that import from each other
2. Index with `index_repository`
3. Query: `MATCH (a)-[r:IMPORTS]->(b) RETURN a.name, b.name` — expect 0 real results
4. Compare: `MATCH (a)-[r:CALLS]->(b) RETURN a.name, b.name` — results exist but only via `unique_name` strategy

dLo999:
Closes #180

## Summary

Adds ES/TS package specifier resolution so that `import { foo } from '@myorg/storage-utils'` produces real IMPORTS edges. Previously, `fqn_module()` treated npm package specifiers as file paths, producing QNs that matched no graph nodes — result: zero IMPORTS edges for all package imports.

## How it works

1. **Package map build** (new Phase 1.5 in `pipeline_run()`): Walks the repo for `package.json` files, parses `name` + entry point (`exports["."]` → `main` → `src/index.ts` fallback). Builds a `CBMHashTable` mapping package names to resolved module QNs.

2. **`cbm_pipeline_resolve_module(ctx, module_path)`**: New wrapper called instead of `fqn_module()` at IMPORTS-edge creation sites. If the specifier is a package reference and found in the map, returns the mapped QN. Handles subpath imports (`@myorg/pkg/utils`) by resolving relative to the package directory. Falls through to `fqn_module()` for relative paths and unknown packages.

3. **Zero behavior change for non-JS repos**: When no `package.json` files exist, `pkg_map` is NULL and all resolution falls through to `fqn_module()`. 0ms overhead.

## Changes

- `src/pipeline/pass_pkgmap.c` (new) — package map build, free, and resolve functions
- `src/pipeline/pipeline_internal.h` — `pkg_map` field on ctx, `cbm_pkg_entry_t` struct, prototypes
- `src/pipeline/pipeline.c` — Phase 1.5 build step, ctx initialization, cleanup
- `src/pipeline/pass_definitions.c:297` — `resolve_module()` instead of `fqn_module()`
- `src/pipeline/pass_calls.c:91` — same
- `src/pipeline/pass_usages.c:96` — same
- `src/pipeline/pass_semantic.c:81` — same
- `src/pipeline/pass_parallel.c:647` — same (in `cbm_build_registry_from_cache`)
- `Makefile.cbm` — added `pass_pkgmap.c` and `test_pkgmap.c`
- `tests/test_pkgmap.c` (new) — 20 unit tests

## Test results

**Build:** Compiles cleanly on macOS (Apple Clang, arm64) with `-Wall -Wextra -Werror`

**Test suite:** 2761 passed, 0 failed (was 2741 — 20 new pkgmap tests)

**Behavioral verification** (monorepo matching issue scenario):

Created pnpm-like monorepo:
```
packages/storage-utils/package.json   {"name": "@myorg/storage-utils", "main": "src/index.ts"}
packages/storage-utils/src/index.ts   (exports validateDatasetName, buildSelectQuery)
apps/server/src/index.ts              (import { validateDatasetName } from '@myorg/storage-utils')
```

| Query | Before fix | After fix |
|-------|-----------|-----------|
| `MATCH (a)-[r:IMPORTS]->(b) RETURN a, b` | **0 rows** | **1 row**: `apps/server/src/__file__` → `packages/storage-utils/src/index` |
| `MATCH (a)-[r:CALLS]->(b) RETURN a, b` | 2 rows (via unique_name) | 2 rows (unchanged — no regression) |
| `pkg_map` phase timing | N/A | 0ms (negligible) |

**Edge cases tested (unit tests):**

| Test | Result |
|------|--------|
| NULL pkg_map → falls through to fqn_module | PASS |
| Relative import `./utils` → falls through | PASS |
| Scoped package `@test/utils` → exact match | PASS |
| Subpath `@test/utils/helpers` → resolves relative to pkg dir | PASS |
| Unknown package `react` → falls through | PASS |
| Package without `"name"` → skipped | PASS |
| Package without entry point → skipped | PASS |
| `exports["."]["import"]` conditional → resolved | PASS |
| No `package.json` in repo → NULL map, 0ms | PASS |

## Scope boundary (not addressed)

- Relative import resolution (`./utils`) — already works
- tsconfig paths (`@/components`) — separate issue
- Barrel re-export resolution — separate enhancement
- External npm deps (in `node_modules/`) — outside project graph by design

Generated with [agent-team](https://github.com/dLo999/agent-team) via /issue


dLo999 (commented):
## Review Summary

Adds ES/TS package specifier resolution via a package map built from `package.json` files. New `pass_pkgmap.c` (build/free/resolve), `pkg_map` field on ctx, 6 call sites updated, 20 new tests. Enables real IMPORTS edges for JS/TS monorepo package imports.

## Findings

Reviewer flagged two "critical" issues that are **false positives**:

- [resolved] pass_pkgmap.c:305 — `cbm_ht_get_key()` was flagged as missing, but it exists at `hash_table.h:49` / `hash_table.c:173`. Build confirms zero link errors. The duplicate package handling pattern is correct.
- [resolved] pass_pkgmap.c:294-307 — Memory management concern depends on above function existing, which it does.

Remaining valid observations (all consistent with existing codebase patterns):
- [nit] pass_pkgmap.c:283 — No NULL check after strdup for pkg_dir. Matches existing pattern (pipeline code doesn't check strdup returns — OOM at this point is unrecoverable).
- [nit] pass_pkgmap.c:370 — Fixed-size 1024/2048 buffers for paths. Same pattern as all other pipeline pass files.
- [nit] pass_pkgmap.c:318 — Stack max 512 for directory traversal. Extreme repos might hit this, but 512 levels is well beyond practical monorepo depth.

## CI Status

No CI on fork branch. Local: 2761 tests pass (20 new + 2741 existing), 0 regressions. Behavioral verification confirms IMPORTS edges now appear for cross-package imports.

## Verdict

**APPROVE** — Both critical findings are false positives (function exists in foundation layer). Implementation follows architect's design, matches existing codebase patterns, and is comprehensively tested.

DeusData:

dLo999 (commented):
## Review Summary

Adds ES/TS package specifier resolution for monorepo IMPORTS edges (closes #180). New `pass_pkgmap.c` walks the repo, parses `package.json` files (name + entry point via `exports["."]`/`main`/fallback), and builds a CBMHashTable mapping package names to resolved module QNs. `cbm_pipeline_resolve_module(ctx, path)` wraps `fqn_module()` at 5 pass sites; falls through for relative/absolute/unknown paths. Zero overhead for non-JS repos (NULL map).

Verification status: branch merged with `upstream/main` cleanly; **2,740 tests pass unsandboxed** (49 "failures" in an earlier sandboxed run were mkdtemp restrictions, not regressions).

**Update (commit `4e183de`):** nit fixes applied. Fix 2 (duplicate-key comment clarification) landed. Fix 1 (NULL-arg ternary split) intentionally skipped — would have been a behavior change (callers `free()` the return unconditionally; `cbm_pipeline_fqn_module` already returns `strdup("")` for NULL project, so the combined guard is load-bearing). 2,740 tests still pass post-fix.

## Findings

- [nit] `pass_pkgmap.c:229-241` — duplicate package-name handling: comment now explicitly documents that `cbm_ht_set` stores the supplied key pointer directly (`slot->key = cur.key`), so allocating a new key without freeing the old would orphan it. Logic unchanged; rationale now clear. ✅ addressed in `4e183de`.
- [nit] `pass_pkgmap.c` `cbm_pipeline_resolve_module()` NULL-arg ternary — intentionally preserved. The combined `if (!ctx || !module_path)` routing through `cbm_pipeline_fqn_module(ctx ? ctx->project_name : NULL, module_path)` is load-bearing because callers `free()` the return unconditionally and `fqn_module` handles NULL project gracefully via `strdup("")`. Splitting would be a behavior change.
- [nit] `tests/test_pkgmap.c` — 20 new tests cover NULL inputs, scoped/bare packages, `exports["."]` resolution, subpath resolution, relative/absolute fallthrough, unknown packages, missing fields. Good coverage.
- [question] CI is not wired for this PR — no build/test job reports visible. Local unsandboxed `make -f Makefile.cbm test` is clean, but upstream CI should confirm across the matrix.

### Draft review raised two CRITICAL findings — both verified false positives

A prior draft review (based on static diff inspection) flagged:

1. **"`cbm_ht_get_key()` is undefined"** — refuted. Declared at `src/foundation/hash_table.h:49`, defined at `src/foundation/hash_table.c:186`, used across `registry.c`, `graph_buffer.c`, with dedicated tests in `test_hash_table.c:187-205` (including NULL safety). The draft also cited wrong line number (claimed line 304; actual call is line 239).

2. **"Use-after-free in duplicate key handling"** — refuted by careful struct-ownership trace. `cbm_ht_get(...)` returns the VALUE (`cbm_pkg_entry_t*`). `free(prev)` frees the value struct only; the key is stored separately in `slot->key` and is not touched. `cbm_ht_get_key()` returns `slot->key`, which remains valid. The subsequent `cbm_ht_set(pkg_map, existing_key, entry)` matches the same key via `strcmp`, updates `slot->value`, and reassigns `slot->key` to the same pointer (hash_table.c:244). No key memory is freed or invalidated in this flow.

Both claims would have been caught by actually running the build/tests, which do in fact pass.

## CI Status

No CI checks reported on `feat/es-ts-package-map`. Local build + test (unsandboxed): clean, 2,740/2,740 pass (both pre- and post-fixup).

## Verdict

**APPROVE** — semantically correct, consistent integration across 5 pass sites, good test coverage, zero-overhead fallthrough for non-JS repos. Recommend upstream CI wire-up as a separate concern.

DeusData:
## Merged (generalized) — huge thank you! 🎉

Your JS/TS package map approach from this PR was the design foundation for a **generic package resolution system** that now covers 10 language ecosystems. Integrated via commit 1bbd950.

**What we built on top of your design:**

Your `pass_pkgmap.c` approach — walk repo for manifests, build a hash table, resolve bare specifiers via `cbm_pipeline_resolve_module()` — was extended to support:

| Language | Manifest | Parser |
|---|---|---|
| JS/TS | `package.json` | yyjson (your original) |
| Go | `go.mod` | line scan |
| Rust | `Cargo.toml` | line scan |
| Python | `pyproject.toml` | line scan |
| PHP | `composer.json` | yyjson + PSR-4 |
| Java | `pom.xml` | XML tag scan |
| Gradle | `build.gradle` | line scan |
| Dart | `pubspec.yaml` | YAML |
| Elixir | `mix.exs` | line scan |
| Ruby | `*.gemspec` | line scan |

**Key architectural change:** Instead of a separate pipeline pass, manifest parsing happens **during parallel extraction** — workers detect manifest files from the already-read source buffer (zero extra I/O), collect entries in per-worker arrays, then merge into a single hash table before registry build. This means the feature has zero overhead for repos without manifest files.

The resolver includes prefix matching for Go (`github.com/foo/bar/pkg/utils`), Java (`com.myorg.pkg.Foo`), and PHP PSR-4 (`App\\Controllers\\Foo`).

You've been credited as co-author on the commit:

```
Co-Authored-By: Dustin Obrecht <dustin@kurtnoble.com>
```

Your original PR was the catalyst for solving this across all languages — thank you for the excellent design, thorough testing, and the conflict resolution work, @dLo999! This is a massive quality improvement for monorepo indexing.

PR #206, opened 2026-04-05
fix(cypher,store): prevent crashes from buffer overflow, OOM, and NULL stmts

map588:
- cypher: Add bounds check in lex_string_literal to prevent stack buffer overflow on string literals >4096 bytes
- cypher: Add malloc/calloc NULL checks in parse_props, parse_rel_types, parse_in_condition, and parse_case_expr to prevent OOM crashes
- store: Add sqlite3_prepare_v2 return code checks at 3 sites in cbm_store_schema_info and collect_pkg_names to prevent NULL stmt dereference on DB corruption


DeusData:
Thanks for the careful defensive hardening, @map588 — bounds check on the lex buffer plus the non-destructive realloc pattern with cleanup is exactly the right shape. Apologies for the long review delay. Merging now and following up on the residual heap_strdup checks and the two adjacent mallocs in `cbm_store_get_schema` in a separate commit on main.

PR #207, opened 2026-04-05
fix(thread-safety): eliminate races in log mutex, watcher, and index threads

map588:
- http_server: Replace lazy log mutex init (TOCTOU race between threads) with explicit cbm_ui_log_init() called from main before thread creation
- watcher: Add cbm_mutex_t to protect projects hash table from concurrent access by main thread (watch/unwatch) and watcher thread (poll_once)
- compat_thread: Add cbm_thread_detach() for POSIX and Windows
- http_server: Detach index job threads to prevent handle leaks


DeusData:
Thanks for the iterative work, @map588 — the 3-state CAS init, snapshot-then-iterate pattern in `poll_once`, and the cross-platform handle clearing in `cbm_thread_join/detach` all read very cleanly. Apologies for the long review delay. Merging now and pushing a small follow-up on main for the remaining `cbm_ht_create()` OOM check in `cbm_watcher_new` and the const-correctness on `cbm_watcher_watch_count`.

PR #208, opened 2026-04-05
feat(pipeline): emit CALLS edges for decorator applications

map588:
Decorators previously only created DECORATES edges. References like @login_required were invisible to "find all references" queries which look for CALLS and USAGE edges. Now both sequential (resolve_decorator) and parallel (resolve_def_decorators) paths emit a CALLS edge alongside DECORATES.

Uses "{}" properties to avoid clobbering richer metadata from pass_calls when a function both has @decorator and calls decorator() directly.


DeusData:
Thanks @map588 — switching to USAGE was the right call to avoid polluting pass_tests' CALLS scan. Merging now. Heads-up: the dead-code degree filter in store.c (around 2281-2283) currently counts only CALLS, so the originally-stated goal of preventing decorated Python functions from being flagged as dead doesn't actually land yet. Following up on main with a one-liner that extends the in_deg/out_deg subquery to count both CALLS and USAGE so this PR's data — and pass_usages output more broadly — feeds into dead-code detection as intended.

PR #209, opened 2026-04-05
Feat/memory helpers

map588:
Adds helper functions which encapsulate repeated patterns throughout the codebase.   I have a larger fix available with the call sites all modified, but this needs checked out first, because its the basis of all that.  Its four helper functions, similer to the `safe_realloc` function.


DeusData:
Thanks for proposing these helpers, @map588.

After reviewing the diff against current main: most of #209's content overlaps with work that landed in the meantime — the cypher.c parser NULL-check fixes and the test_cypher.c overflow test are now in main via #206, and the watcher mutex / cbm_thread_detach / cbm_ui_log_init pieces are in main via #207. A couple of #209's variants of those changes are slightly older than what merged (e.g. the lex_string_literal truncation here uses an early-continue pattern that drops escape consumption — `\"` past the truncation point would close the string early; #206's merged version specifically avoids that. Similarly collect_pkg_names returning 0 conflates 'no rows' with 'prepare failed'; #206's merged version returns CBM_NOT_FOUND).

The genuinely new and useful piece here is the four memory helpers in platform.h. I cherry-picked those into main as 8fe1310 with you co-authored. The full diff with all the call-site replacements (which I see was the planned follow-up) can land cleanly on top of the helpers commit whenever you have it ready — opening it as a focused PR against current main would be much easier to review than rebasing this one through three intervening merges.

Closing this PR in favor of 8fe1310. Really appreciate the contribution.

Issue #215, opened 2026-04-06
SEGV in c_resolve_pending_template_calls: OOB read on NULL-terminated param_types array


JasonDictos:
## Bug

Indexing any repository containing `nlohmann/json.hpp` (or other heavily-templated C++ headers) crashes with a segfault in `c_resolve_pending_template_calls`.

## Version

0.5.7 (also reproduced on current `main`)

## Reproduction

```bash
# Minimal repo setup
mkdir -p /tmp/cbm-crash-repro && cd /tmp/cbm-crash-repro
git init
curl -sL https://github.com/nlohmann/json/releases/download/v3.12.0/json.hpp -o json.hpp
git add json.hpp && git commit -m "add nlohmann json"

# Crash
codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/cbm-crash-repro"}'
# => Segmentation fault (core dumped)
```

## Root Cause

`param_types` in `CBMType.data.func` is a **NULL-terminated array** (see `type_rep.h:48`), but two loops iterate using `call_arg_count` / `arg_count` as the bound instead of stopping at the NULL sentinel. When a template function is called with more arguments than formal parameters (common in heavily-templated C++ like nlohmann/json), this reads past the end of the array.

**`internal/cbm/lsp/c_lsp.c:304`** — `c_resolve_pending_template_calls`:
```c
for (int i = 0; i < call_arg_count; i++) {
    const CBMType* formal = callee->signature->data.func.param_types[i];
    if (!formal || !call_arg_types[i]) continue;  // BUG: should break on !formal, not continue
```

**`internal/cbm/lsp/type_registry.c:215`** — `score_overload_match`:
```c
for (int i = 0; i < arg_count; i++) {
    const CBMType* expected = f->signature->data.func.param_types[i];
    // ... same bug
```

## Backtrace

```
#0  c_resolve_pending_template_calls () at internal/cbm/lsp/c_lsp.c:305
#1  c_resolve_calls_in_node () at internal/cbm/lsp/c_lsp.c:3200
#2  c_resolve_calls_in_node () [recursive]
#3  c_resolve_calls_in_node ()
#4  c_process_body_child ()
#5  c_lsp_process_file ()
#6  cbm_run_c_lsp ()
#7  cbm_extract_file () — source=nlohmann/json.hpp (967KB)
#8  extract_worker () at src/pipeline/pass_parallel.c:481
```

## Fix

[...]

Issue #213, opened 2026-04-06
Large TypeScript files silently produce zero Function/Module nodes


digijoebz:
## Description

Large TypeScript files are indexed (a `File` node is created) but tree-sitter parsing silently fails — no `Module`, `Function`, `Method`, `Variable`, or `Interface` nodes are produced. No error is reported to the user.

## Reproduction

Indexing a TypeScript/Deno project with these files:

| File | Size | Lines | Result |
|------|------|-------|--------|
| `universal-dca-logic.ts` | 236 KB | 4,437 | File node only — zero children |
| `trade-executor/index.ts` | 406 KB | 8,406 | File node only — zero children |
| `dca-health-check/index.ts` | 182 KB | 4,234 | File node only — zero children |
| `AdminDCAOptimizer.tsx` | 688 KB | 8,517 | File node only — zero children |
| `ClientSuccess.tsx` | 321 KB | 6,243 | File node only — zero children |

Files of similar or larger size in the same project DO index successfully:

| File | Size | Nodes Indexed |
|------|------|---------------|
| `scalping-optimizer.worker.ts` | 173 KB | 101 |
| `AdminBotUpgrade.tsx` | 168 KB | 39 |
| `AdminScalpingOptimizer.tsx` | 178 KB | 21 |

## Impact

The 3 most critical business logic files (containing the core trading algorithm, trade execution pipeline, and health monitoring) are completely invisible to `search_graph`, `trace_call_path`, and `get_code_snippet`. Call graph analysis stops at these file boundaries.

## Environment

- Version: 0.5.7
- OS: Linux 6.8.0-107-generic
- Language: TypeScript/Deno

## Suspected Cause

The binary contains the string `userconfig.file_too_large` and `Expression tree is too large (maximum depth %d)`, suggesting an internal limit based on AST complexity rather than raw file size. The failure appears to be in the parse phase — the file walker creates the `File` node but tree-sitter parsing never completes.

Related issues: #130 (quadratic ts_node_child loop on large ASTs), #62 (stack overflow on large files)

## Expected Behavior

Either: successfully index these files (preferred), or report a clear warning/error that the file was skipped due to size/complexity limits so users know their index is incomplete.

DeusData:
Hey, thanks for publishing the issue. I will tackle this immediately after the 15th of April :)

DeusData:
[...]

PR #217, opened 2026-04-07
fix(extraction): replace fixed traversal stacks with growable arena-allocated stacks

### LINKED ISSUE #199: AST Nodes silently dropped in walk_calls when there are more than CBM_SZ_512
The nodes are dropped [here](https://github.com/DeusData/codebase-memory-mcp/blob/e07443b75f058f6b3739de0217c5f2f92405a016/internal/cbm/extract_calls.c#L386) once it goes past the cap. Essentially, it just stops processing the file. In this instance we have an node.js / express routes file with ~130 routes.

AhmedHamadto:
## Problem

AST traversal functions in the extraction layer use fixed-size `TSNode stack[N]` arrays. When the DFS stack fills up, the child-push loop exits silently via `&& top < CAP`, dropping entire subtrees without warning.

This affects files with many top-level siblings (e.g., 600 ES imports, 130 Express routes). Reported in #199 — a Node.js routes file with ~130 routes lost data past the 512 cap.

**Confirmed:** extracting 600 TypeScript imports returns 511 (capped at `ES_IMPORT_STACK_CAP = 512`).

## Fix

Added `ts_node_stack.h` — a 53-line growable stack backed by the existing arena allocator. Initial capacity matches the previous fixed caps (no extra memory for small files), but doubles on overflow instead of truncating.

Applied to all 14 `TSNode` fixed stacks across 9 extraction files:

| File | Functions fixed | Old cap |
|------|----------------|---------|
| `extract_calls.c` | `walk_calls` | 512 |
| `extract_imports.c` | `walk_es_imports`, `walk_wolfram_imports` | 512 |
| `extract_semantic.c` | `walk_throws`, `walk_readwrites` | 512 |
| `extract_defs.c` | `extract_elixir_call`, `walk_variables_iter`, `push_nested_class_nodes` | 64–256 |
| `extract_usages.c` | `walk_usages` | 4096 |
| `extract_env_accesses.c` | `walk_env_accesses` | 4096 |
| `extract_type_refs.c` | `walk_body_type_refs`, `walk_type_refs` | 4096 |
| `extract_type_assigns.c` | `walk_type_assigns` | 4096 |
| `extract_channels.c` | `scan_string_consts`, `cbm_extract_channels` | 4096 |

**Not changed:** `walk_defs` uses `walk_defs_frame_t` (different struct type), left as-is with its 4096 cap.

## Tests

Added `test_stack_overflow.c` with 7 regression tests:

| Test | Before fix | After fix |
|------|-----------|-----------|
| `ts_imports_exceed_512` | **511/600 FAIL** | **600/600 PASS** |
| `js_calls_exceed_512` | 600/600 | 600/600 |
| `python_calls_exceed_512` | 600/600 | 600/600 |
| `go_calls_exceed_1024` | 1024/1024 | 1024/1024 |
| `express_routes_exceed_512` | 150/150 | 150/150 |
| `js_deeply_nested_calls` | 200/200 | 200/200 |
| `yaml_vars_exceed_256` | 300/300 | 300/300 |

Full suite: **2,727 passed, 0 failed.**

## Design notes

- **Arena-backed**: old arrays are abandoned in the arena (freed on `arena_destroy` at end of file extraction). No `realloc`, no mixed lifetimes.
- **Same initial capacity** as before — zero extra allocation for files within old caps.
- **Net -2 lines** (removed 15 `#define` caps, added growable stack header).
- `clang-tidy`/`cppcheck` not available on build machine — CI will validate.

Fixes #199


DeusData:

DeusData:
This PR also addresses #215 (SEGV in template calls from stack overflow) and #213 (large TypeScript files producing zero nodes due to stack exhaustion).

DeusData:
Thanks for this @AhmedHamadto, and apologies for the long delay (post-Apr-15 promise on my side that I clearly missed). The fix is well-scoped and the regression coverage is excellent — exactly the right shape for #199, #213, #215.

Verified the production code (`ts_node_stack.h` + the 9 `extract_*.c` substitutions) compiles cleanly. The only snag is portability of the new `tests/test_stack_overflow.c`: macOS Clang treats `sprintf` as deprecated under `-Werror,-Wdeprecated-declarations`, so the build fails on Apple toolchains (~19 sprintf calls). Linux toolchains where you tested only warn.

Merging now and pushing a small follow-up on main converting the sprintfs to snprintf so the test compiles on macOS.

DeusData:
Landed as aa6df6b on main with you as the commit author. Resolution notes:

- Conflict in `Makefile.cbm`: kept your `TEST_STACK_OVERFLOW_SRCS` line plus the `TEST_ZSTD_SRCS` / `TEST_ARTIFACT_SRCS` entries that landed on main since this PR was opened.
- Conflict in `internal/cbm/extract_channels.c`: main added a `scan_string_consts_python` companion to `scan_string_consts_js` after this PR was opened, so I kept the `_js` suffix on the function name and declared the missing `CHAN_STACK_CAP` constant in the enum block; the new `_python` walker still uses the old fixed-stack pattern and can be converted in a small follow-up.
- macOS portability: macOS Clang rejects `sprintf` under `-Werror,-Wdeprecated-declarations`, and a naive `snprintf(buf, sizeof(buf), ...)` rewrite triggered `-Wformat-truncation` because `sizeof(p)` on a `char *` returns 8. Updated each call to compute the actual remaining buffer space.

Full suite is green: 2820 passed, 0 failed, including all 7 `stack_overflow_*` tests. Thanks for the careful work and the clear reproduction; closing this PR in favor of the cherry-picked commit.

PR #224, opened 2026-04-08
fix(imports): extract imports from Svelte and Vue script blocks

### LINKED ISSUE #223: Fix import extraction for Svelte and Vue script blocks
## Summary
Svelte and Vue files are discovered and parsed, but imports inside their `<script>` blocks are not extracted into `IMPORTS` edges.

## Root cause
`cbm_extract_imports()` only routes JavaScript, TypeScript, and TSX through the ES import walker. Svelte and Vue parse to a markup-oriented AST at the root, where imports live inside `document -> script_element -> raw_text`, so they never reach the normal ES import extraction path.

## Proposed fix
- add a Svelte/Vue-specific import extractor in `internal/cbm/extract_imports.c`
- find `script_element -> raw_text`
- re-parse the embedded script text with the JavaScript grammar
- run the existing ES import walker on that sub-tree
- add regression tests for `.svelte` and `.vue` import extraction

## Validation
I verified this against a downstream Svelte repo after building the local branch:
- before the fix: `src/App.svelte` had zero outgoing `IMPORTS` edges
- after the fix: `src/App.svelte` produced 7 `IMPORTS` edges to relative imports such as:
  - `src/components/TopNavigation.svelte`
  - `src/components/TreeView.svelte`
  - `src/components/AdCanvas.svelte`

Alias imports like `$lib/...` and `@/...` still do not resolve, but that is a separate pipeline concern and not part of this fix.

## Scope
This issue is only about extracting imports from embedded script blocks in `.svelte` / `.vue` files.

jmcmacnz:
Fixes #223

## Summary
- extract imports from embedded `<script>` blocks in `.svelte` and `.vue` files
- re-parse `script_element -> raw_text` content with the JavaScript grammar and reuse the existing ES import walker
- add regression coverage for Svelte and Vue import extraction

## Why
Svelte and Vue files are discovered and parsed, but their root AST is markup-oriented rather than plain JS/TS. Imports live inside `document -> script_element -> raw_text`, so they never reached the normal ES import extraction path.

## Validation
- built the local binary from this branch with `scripts/build.sh`
- indexed a downstream Svelte repo using the local binary directly
- verified that `src/App.svelte` now produces 7 `IMPORTS` edges to relative imports including:
  - `src/components/TopNavigation.svelte`
  - `src/components/TreeView.svelte`
  - `src/components/AdCanvas.svelte`
- confirmed alias imports like `$lib/...` and `@/...` still remain unresolved, which is expected and out of scope for this PR

### DIFF
--- a/internal/cbm/extract_imports.c
+++ b/internal/cbm/extract_imports.c
@@ -7,6 +7,11 @@
 #include <string.h>
 #include <ctype.h>
 
+/* Forward declaration: JS grammar used for re-parsing <script> blocks in
+ * Svelte and Vue files.  The symbol is provided by grammar_javascript.c which
+ * is always compiled into the same link unit. */
+extern const TSLanguage *tree_sitter_javascript(void);
+
 /* Local constants for magic number elimination. */
 enum {
     USE_PREFIX_LEN = 4, /* strlen("use ") */
@@ -28,6 +33,8 @@ static void parse_ruby_imports(CBMExtractCtx *ctx);
 static void parse_lua_imports(CBMExtractCtx *ctx);
 static void parse_generic_imports(CBMExtractCtx *ctx, const char *node_type);
 static void parse_wolfram_imports(CBMExtractCtx *ctx);
+static void walk_es_imports(CBMExtractCtx *ctx, TSNode root);
+static void parse_svelte_vue_imports(CBMExtractCtx *ctx);
 
 // Helper: strip quotes from a string literal
 static char *strip_quotes(CBMArena *a, const char *s) {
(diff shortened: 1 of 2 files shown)


DeusData:
Hey, thanks for building this. Will review ASAP once I am back on the 15th :)

DeusData:
Landed as 3094198 on main with you co-authored. Generalized the approach a bit while at it: rather than a Svelte/Vue-specific helper, the host grammar now declares its embedded sub-language via a new `CBMEmbeddedLangSpec` field on `CBMLangSpec` (`{"script_element", "raw_text", CBM_LANG_JAVASCRIPT}`), and a generic walker in `extract_imports.c` resolves the embedded grammar through the existing `cbm_ts_language()` hook. That kept the extraction file free of any grammar symbol references and made it trivial to add HTML and Astro to the same code path with a one-line declaration each — same problem class, same shape of fix.

Test coverage from your PR is in plus an HTML smoke test (`html_imports_basic`); full suite stays green at 2824/0. Thanks for the careful repro and the clear PR description; closing this in favor of the cherry-picked commit.

PR #231, opened 2026-04-11
fix(mcp): cap search_graph default limit to 200 (was 500K)

amitmynd:
## Problem

`search_graph` with a broad `file_pattern` (e.g. `**/ads/**`) returns output that exceeds LLM tool result limits. In my case, querying a 12K-node project returned **3,358,396 characters**, which Claude Code couldn't process:

```
search_graph (MCP)(project: "Users-amitbhalla-Downloads-myndintegrated", file_pattern: "**/ads/**")

Error: result (3,58,396 characters) exceeds maximum allowed tokens.
```

## Root cause

The default `limit` for the regex/file_pattern search path uses `MCP_HALF_SEC_US` (500,000), which appears to be a microsecond timeout constant that was accidentally used as a row count. When no explicit `limit` is passed, the query returns up to 500K rows, serializing all of them into the JSON response.

- `src/mcp/mcp.c:1344`: `cbm_mcp_get_int_arg(args, "limit", MCP_HALF_SEC_US)` (500,000)
- `src/store/store.c:2304`: fallback to `ST_HALF_SEC` (also 500,000)
- Tool description says "Default: unlimited"

## Fix

Changed the default limit to **200** in both the MCP handler and the store layer. Updated the tool description to reflect this.

**Before:** 1,287 matching nodes, all returned, 3.3M characters output
**After:** 1,287 total (reported in `total` field), 200 returned, 57K characters output, `has_more: true`

Pagination with `offset` and `limit` continues to work as expected for callers that need more results.

## Changes

| File | Change |
|------|--------|
| `src/mcp/mcp.c` | Default limit 500,000 -> 200 in `handle_search_graph` |
| `src/store/store.c` | Fallback limit `ST_HALF_SEC` -> 200 in `cbm_store_search` |
| `src/mcp/mcp.c` | Tool description: "Default: unlimited" -> "Default: 200" |

## Test plan

- [x] Build succeeds with `make -f Makefile.cbm cbm`
- [x] `search_graph` with broad `file_pattern` returns 200 results (57K chars) instead of all matches (3.3M chars)
- [x] Response includes `total: 1287`, `has_more: true` for pagination
- [x] `list_projects` and `--help` work correctly
- [ ] Explicit `limit` param still overrides the default

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -290,7 +290,7 @@ static const tool_def_t TOOLS[] = {
      "that score well on ALL keywords. Requires moderate/full index mode. Results appear in the "
      "'semantic_results' field (separate from 'results').\"},\"limit\":{\"type\":"
      "\"integer\",\"description\":\"Max results. Default: "
-     "unlimited\"},\"offset\":{\"type\":\"integer\",\"default\":0}},\"required\":[\"project\"]}"},
+     "200\"},\"offset\":{\"type\":\"integer\",\"default\":0}},\"required\":[\"project\"]}"},
 
     {"query_graph",
      "Execute a Cypher query against the knowledge graph for complex multi-hop patterns, "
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks @amitmynd — clear repro, root-cause that pinpoints exactly which constant was misused, and a minimal two-file fix. Landed as 9425b99 on main. Going to do a small follow-up to consolidate the two inline `SEARCH_DEFAULT_LIMIT` / `ST_DEFAULT_SEARCH_LIMIT` enums into a single shared constant, and to make the limit / offset / total / has_more pagination contract more explicit in the tool descriptions so agents understand the truncation behavior on first read.

PR #243, opened 2026-04-11
Fix/import map call resolution

petercoxphoto:
### Summary

- **Layer 1a:** Add tsconfig.json path alias resolution. TypeScript projects using path aliases (e.g. `@/*` → `src/*`) had zero IMPORTS edges because `resolve_relative_import` only handled `./relative` paths. New `cbm_load_tsconfig_paths()` parses `compilerOptions.paths` and `baseUrl` from tsconfig.json/jsconfig.json. Aliases sorted by specificity (longest prefix first) per TypeScript semantics. JSONC-aware (comments + trailing commas).
- **Layer 1b:** Walk monorepo for nested tsconfigs. The initial fix only read `{repo_root}/tsconfig.json`. Monorepos have path aliases in subdirectory tsconfigs (e.g. `apps/manager/tsconfig.json`). New `cbm_load_all_tsconfig_paths()` walks the repo for all tsconfig.json/jsconfig.json files and builds a per-directory alias collection. `cbm_find_path_aliases()` finds the nearest ancestor tsconfig for each source file. Target paths (e.g. `./src/*`) are resolved relative to the tsconfig's directory.
- **Layer 2:** Fix bare function call resolution via import_map. When callee has no dot (e.g. `requireAdmin`), the candidate QN was built as just `module_qn` instead of `module_qn.requireAdmin`. One-line fix in `registry.c`.
- **20 new tests** covering alias resolution (wildcards, exact match, baseUrl, overlapping prefixes, extension stripping, NULL safety), monorepo collection (nearest ancestor lookup, no-match fallback, root-only), and the registry bare function call fix. All 2740 tests pass (ASan + UBSan).

### Problem

CALLS edges for cross-file function calls are mis-wired or missing when functions share names across files in TypeScript projects that use path aliases.

**Resolution cascade failure:**
1. Strategy 1 (import_map) — fails because IMPORTS edges are missing (no alias resolution)
2. Strategy 2 (same_module) — fails because the function is in a different file
3. Strategy 3/4 (name_lookup) — finds both candidates, picks the wrong one

### Verified impact

Tested against a real monorepo (Next.js, 2,769 nodes, `@/*` aliases in `apps/manager/tsconfig.json`):

| Metric | Before fix | After Layer 1a | After Layer 1b |
|--------|-----------|----------------|----------------|
| IMPORTS edges | 102 | 123 | **682** |
| Total edges | 4,383 | 5,238 | **6,186** |

The 6.7x increase in IMPORTS edges directly improves `trace_path` accuracy for cross-file call resolution.

### Files changed

| File | Change |
|------|--------|
| `src/pipeline/fqn.c` | New: `cbm_load_tsconfig_paths`, `cbm_resolve_path_alias`, `cbm_load_all_tsconfig_paths`, `cbm_find_path_aliases`, `cbm_tsconfig_collection_free` |
| `src/pipeline/pipeline_internal.h` | New: `cbm_path_alias_t`, `cbm_path_alias_map_t`, `cbm_tsconfig_entry_t`, `cbm_tsconfig_collection_t` structs |
| `src/pipeline/pipeline.c` | Load tsconfig collection at pipeline start, free on cleanup |
| `src/pipeline/pipeline_incremental.c` | Same for incremental pipeline path |
| `src/pipeline/pass_definitions.c` | Resolve imports via nearest ancestor tsconfig aliases |
| `src/pipeline/pass_parallel.c` | Same for parallel pipeline path |
| `src/pipeline/registry.c` | Fix bare function call candidate: `module_qn.prefix` not `module_qn` |
| `tests/test_fqn.c` | 19 new tests (alias resolution + monorepo collection) |
| `tests/test_registry.c` | 1 new bare function call import_map test |

### Test plan

- [x] All 2740 tests pass (ASan + UBSan enabled)
- [x] Path alias wildcard resolution (`@/lib/auth` → `src/lib/auth`)
- [x] Multiple alias prefixes (`@/`, `~/`, `@components/`)
- [x] Overlapping prefixes resolved by specificity (longest prefix wins)
- [x] baseUrl resolution with bare package name exclusion
- [x] Exact match aliases (no wildcard)
- [x] Extension stripping (.ts, .tsx, .js, .jsx)
- [x] NULL safety (NULL map, NULL path, missing tsconfig)
- [x] Relative imports unaffected by alias resolution
- [x] Bare function call via import_map resolves correctly
- [x] Monorepo: nearest ancestor tsconfig selected per source file
- [x] Monorepo: root-only tsconfig matches all files
- [x] Monorepo: files with no ancestor tsconfig fall through gracefully
- [x] Monorepo: non-existent repo returns NULL
- [x] No memory leaks (ASan clean)
- [x] Real-world verification: On an in-house project, IMPORTS 102 → 682

### DIFF
--- a/src/pipeline/fqn.c
+++ b/src/pipeline/fqn.c
@@ -9,6 +9,7 @@
 #include "pipeline/pipeline_internal.h"
 #include "foundation/constants.h"
 #include "foundation/platform.h"
+#include "foundation/compat_fs.h"
 
 #include <stdbool.h>
 #include <stddef.h> // NULL
@@ -406,24 +407,42 @@ static char *strip_resolved_ext(char *path) {
     return path;
 }
 
-cbm_path_alias_map_t *cbm_load_tsconfig_paths(const char *repo_path) {
-    if (!repo_path) {
+/* Internal: resolve a target path pattern relative to a tsconfig directory.
+ * If target starts with "./" and dir_prefix is non-empty, prepend dir_prefix.
+ * E.g. dir_prefix="apps/manager", target="./src/foo" -> "apps/manager/src/foo"
+ * Returns heap-allocated string. */
+static char *resolve_target_relative(const char *dir_prefix, const char *target) {
+    if (!target) {
         return NULL;
     }
-
-    /* Try tsconfig.json, then jsconfig.json */
-    static const char *config_names[] = {"tsconfig.json", "jsconfig.json"};
-    FILE *f = NULL;
-    char path_buf[CBM_SZ_512];
-    for (int i = 0; i < 2 && !f; i++) {
-        snprintf(path_buf, sizeof(path_buf), "%s/%s", repo_path, config_names[i]);
-        f = fopen(path_buf, "r");
+    /* Strip leading "./" from target */
+    const char *t = target;
+    if (t[0] == '.' && t[1] == '/') {
+        t += 2;
     }
+    if (!dir_prefix || dir_prefix[0] == '\0') {
+        return strdup(t);
+    }
+    size_t dp_len = strlen(dir_prefix);
+    size_t t_len = strlen(t);
+    char *result = malloc(dp_len + 1 + t_len + 1);
+    if (!result) {
+        return NULL;
+    }
+    snprintf(result, dp_len + 1 + t_len + 1, "%s/%s", dir_prefix, t);
+    return result;
+}
+
+/* Internal: load and parse a single tsconfig/jsconfig file.
+ * abs_path is the full filesystem path to the JSON file.
+ * dir_prefix is the tsconfig's directory relative to repo root (e.g. "apps/manager").
+ * Target paths in the alias map are resolved relative to dir_prefix. */
+static cbm_path_alias_map_t *load_tsconfig_file(const char *abs_path, const char *dir_prefix) {
+    FILE *f = fopen(abs_path, "r");
     if (!f) {
         return NULL;
     }
 
-    /* Read the file */
     fseek(f, 0, SEEK_END);
     long len = ftell(f);
     fseek(f, 0, SEEK_SET);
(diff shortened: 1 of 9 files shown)


DeusData:
Landed as 10a9b83 on main with you as the commit author. Distilled the four-commit branch into a single squashed change and generalised the abstraction a bit so other build tools can plug in without touching extraction code:

- New `src/pipeline/path_alias.{h,c}`. The data model (`cbm_path_alias_t` / `_map_t` / `_scope_t` / `_collection_t`) and the resolver are language-agnostic; the tsconfig/jsconfig parser is one helper inside the loader (`load_tsconfig_file`). Adding Vite/Webpack alias config or Python tool hints later means writing a sibling helper and registering it in `find_alias_files` — the rest of the file, plus `cbm_pipeline_resolve_module`, doesn't change.
- Single pipeline integration point: `cbm_pipeline_resolve_module` in `pass_pkgmap.c` got a Strategy 1b between relative-import resolution and pkgmap. Both the sequential and parallel pipelines pick this up for free, instead of patching `pass_definitions.c` and `pass_parallel.c` separately.
- Performance: replaced both O(n²) bubble sorts (entries-by-specificity and scopes-by-dirprefix-length) with `qsort`. Caps are now 256 entries per config and 256 configs per repo, with a recursion-depth cap of 32; cap hits emit a `cbm_log_warn` so pathological monorepos surface a signal instead of silently truncating.
- Walk hygiene: existing skip list (`node_modules`, `.git`, `dist`, `build`, `.next`, `coverage`) plus `target` for Rust roots.
- Layer 2 (registry bare-call fix): kept verbatim with the explanatory comment expanded.
- Tests distilled to 10 focused cases including a tmp-dir monorepo end-to-end via the real loader, plus the bare-call regression in `test_registry.c`. Full suite: 2835 passed, 0 failed.

The 6.7× IMPORTS-edge improvement you measured on the Next.js monorepo was the convincing data point. Closing in favor of the cherry-picked commit.

PR #248, opened 2026-04-13
fix: harden PyPI installer against tar-slip and scheme attacks (#246)

### LINKED ISSUE #246: security: Installer (_cli.py) has tar-slip + urllib scheme + subprocess issues
## Summary

The PyPI installer at `pkg/pypi/src/codebase_memory_mcp/_cli.py` has a cluster of security issues around downloading and extracting release tarballs. Combined, they create a realistic supply-chain attack surface for anyone running `pip install` + the bundled install command.

Found via static scan (ruff security rules + semgrep). I am reporting this because the combination is more concerning than any finding alone — happy to expand on any part or dismiss ones that are working-as-intended.

### 1. `urllib.urlopen` without scheme whitelist — `_cli.py:79`

```
ruff S310: Audit URL open for permitted schemes
```

`urllib.request.urlopen` accepts any scheme the standard handlers support, including `file://`, `ftp://`, and custom schemes. If the URL source can be influenced (env var, config, argv, a redirect), a `file://` URL turns the fetch into arbitrary local file read.

**Fix:** explicitly validate `urlparse(url).scheme in {"https"}` (or `{"https", "http"}` if you really need http) before calling `urlopen`.

### 2. `tarfile.extractall()` without filter — `_cli.py:90` and `_cli.py:94`

```
ruff S202: Uses of `tarfile.extractall()`
```

This is the classic **tar-slip / Zip Slip** vulnerability (CVE class, CWE-22). A malicious tarball containing entries with absolute paths or `../` path components can write files outside the extraction directory when `extractall` runs — e.g., overwriting `~/.bashrc`, `~/.ssh/authorized_keys`, or `~/.local/bin/` binaries that later get executed.

**Fix (Python 3.12+):** pass `filter='data'` — this is exactly what the filter was added for.
```python
tar.extractall(path=dest, filter='data')
```

**Fix (pre-3.12):** validate each member path before extracting:
```python
for member in tar.getmembers():
    member_path = os.path.abspath(os.path.join(dest, member.name))
    if not member_path.startswith(os.path.abspath(dest) + os.sep):
        raise RuntimeError(f"Unsafe tar entry: {member.name}")
tar.extractall(path=dest)
```

### 3. `subprocess` calls without input validation — `_cli.py:118, 121`

```
ruff S603: `subprocess` call: check for execution of untrusted input
ruff S606: Starting a process without a shell
```

Lower-severity context-dependent finding, but worth reviewing whether the arguments passed to subprocess can be influenced by downloaded-tarball contents.

## Combined impact

If any user running the installer can be induced to use a crafted download URL (typosquat PyPI mirror, compromised release server, MITM on non-HTTPS path), the flow is:

1. `urlopen(bad_url)` — fetches attacker-controlled tarball over possibly-insecure scheme
2. `extractall()` — extracts with no path safety
3. Attacker writes to `~/.local/bin/codebase-memory-mcp`, `~/.bashrc`, or any path outside the intended install directory
4. Next time user runs the binary / opens a shell, attacker code executes

The fix for #2 (tar-slip) is a one-line change (`filter='data'`) and materially reduces risk even if #1 is deemed acceptable.

## Repro

Static analysis only — did not attempt live exploit against the installer. Scan reproduced via `ruff check pkg/pypi/src/codebase_memory_mcp/_cli.py --select S`.

## Environment

- Scan date: 2026-04-13
- Tools: ruff (security rules S*), semgrep default ruleset
- Repo state: HEAD of main as of scan time

dLo999:
Closes #246

## Summary

Three targeted security improvements to `pkg/pypi/src/codebase_memory_mcp/_cli.py`:

1. **Tar-slip / Zip-slip mitigation (S202):** the previous `tf.extractall(tmp)` / `zf.extractall(tmp)` calls trusted archive contents. Replaced with `_safe_extract_tar` and `_safe_extract_zip`:
   - On **Python 3.12+**, tarfile uses the new `filter='data'` per [PEP 706](https://peps.python.org/pep-0706/)
   - On **Python 3.8–3.11** (and for zipfile, which has no built-in filter), each member path is validated via `os.path.abspath` checks; symlinks and hardlinks in tarballs are rejected outright

2. **URL scheme allowlist (S310):** `_validate_url_scheme` rejects anything that isn't `https://` before `urlretrieve` runs. The URL is constructed from hardcoded constants today, but this is cheap defense-in-depth against future config sources, redirects, or typos.

3. **Justified `# noqa` on exec path (S603/S606):** `os.execv` and `subprocess.run` receive args as a list (not a shell string) and neither uses `shell=True`. Documented the invariant inline so future refactors don't reintroduce a shell form.

## Changes

- `pkg/pypi/src/codebase_memory_mcp/_cli.py` — +74 / −5

## Test results

**Static verification (ruff security rules):**
- Before: 5 findings (S310, S202×2, S603, S606)
- After: **0 findings** — `ruff check _cli.py --select S` passes

**Syntax:** clean under Python 3.14

**Behavior preservation:** happy path (download → extract → exec) unchanged. Only new code path is `sys.exit` with actionable message when a malicious member path or non-https URL is detected.

**Python compatibility:** stdlib-only changes, compatible with declared `requires-python = ">=3.8"`. No new dependencies.

### Verification gate

```
VERIFICATION GATE — Issue #246:
  [PASS]  Build: no python build needed; installer is a runtime shim
  [N/A]   Test suite: no python tests for _cli.py in the repo
  [PASS]  Issue scenario: ruff check _cli.py --select S — was 5 findings, now 0
  [PASS]  Edge cases: Python 3.12+ uses filter='data'; 3.8-3.11 uses manual
          path validation + symlink/hardlink rejection
  [PASS]  Grep: all extractall() calls now inside safe wrappers
  [N/A]   README: internal security hardening, no user-visible change
RESULT: PASS
```

### DIFF
--- a/pkg/pypi/src/codebase_memory_mcp/_cli.py
+++ b/pkg/pypi/src/codebase_memory_mcp/_cli.py
@@ -8,9 +8,73 @@ import shutil
 import tempfile
 import urllib.request
 import urllib.error
+import urllib.parse
 from pathlib import Path
 
 REPO = "DeusData/codebase-memory-mcp"
+
+# Security: only permit https fetches. urllib's default handlers accept
+# file://, ftp://, and custom schemes — a redirect or tainted URL source
+# could otherwise turn a download into an arbitrary-local-file read.
+_ALLOWED_SCHEMES = frozenset({"https"})
+
+
+def _validate_url_scheme(url: str) -> None:
+    """Reject non-https URLs before any network fetch."""
+    scheme = urllib.parse.urlparse(url).scheme
+    if scheme not in _ALLOWED_SCHEMES:
+        sys.exit(
+            f"codebase-memory-mcp: refusing to fetch non-https URL "
+            f"(scheme={scheme!r}): {url}"
+        )
+
+
+def _safe_extract_tar(tf, dest: str) -> None:
+    """Extract a tarfile to dest, rejecting path-traversal entries.
+
+    Uses the tarfile data filter on Python >=3.12 (PEP 706), falls back to
+    manual per-member path validation on older Pythons. Mitigates the
+    classic tar-slip / Zip Slip vulnerability (CWE-22).
+    """
+    # Python 3.12+: use the built-in 'data' filter which rejects absolute
+    # paths, '..' components, symlinks pointing outside dest, etc.
+    if hasattr(tf, "extraction_filter") or sys.version_info >= (3, 12):
+        tf.extractall(dest, filter="data")
+        return
+
+    # Fallback for Python <3.12: validate each member before extracting.
+    dest_abs = os.path.abspath(dest)
+    for member in tf.getmembers():
+        # Reject symlinks and hardlinks outright (they can escape dest).
+        if member.issym() or member.islnk():
+            sys.exit(
+                f"codebase-memory-mcp: refusing unsafe tar entry "
+                f"(link: {member.name!r})"
+            )
+        member_abs = os.path.abspath(os.path.join(dest_abs, member.name))
+        if not (member_abs == dest_abs or member_abs.startswith(dest_abs + os.sep)):
+            sys.exit(
+                f"codebase-memory-mcp: refusing unsafe tar entry "
+                f"(escapes dest: {member.name!r})"
+            )
+    tf.extractall(dest)
+
+
+def _safe_extract_zip(zf, dest: str) -> None:
+    """Extract a zipfile to dest, rejecting path-traversal entries.
+
+    zipfile.ZipFile has no built-in extraction filter; mirrors the tar
+    fallback logic — validate each member before extracting.
+    """
+    dest_abs = os.path.abspath(dest)
+    for name in zf.namelist():
+        member_abs = os.path.abspath(os.path.join(dest_abs, name))
+        if not (member_abs == dest_abs or member_abs.startswith(dest_abs + os.sep)):
+            sys.exit(
+                f"codebase-memory-mcp: refusing unsafe zip entry "
+                f"(escapes dest: {name!r})"
+            )
+    zf.extractall(dest)
 
 
 def _version() -> str:
@@ -64,6 +128,7 @@ def _download(version: str) -> Path:
         f"https://github.com/{REPO}/releases/download/v{version}"
         f"/codebase-memory-mcp-{os_name}-{arch}.{ext}"
     )
+    _validate_url_scheme(url)
 
     dest = _bin_path(version)
     dest.parent.mkdir(parents=True, exist_ok=True)
@@ -76,7 +141,7 @@ def _download(version: str) -> Path:
     with tempfile.TemporaryDirectory() as tmp:
         tmp_archive = os.path.join(tmp, f"cbm.{ext}")
         try:
-            urllib.request.urlretrieve(url, tmp_archive)
+            urllib.request.urlretrieve(url, tmp_archive)  # noqa: S310 — scheme validated above
         except urllib.error.HTTPError as e:
             sys.exit(
                 f"codebase-memory-mcp: download failed ({e})\n"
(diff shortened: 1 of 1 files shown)


dLo999 (commented):
## Review Summary

Independent review of this PR. It addresses 3 security concerns in `_cli.py`: tar-slip mitigation, URL scheme allowlist, and justified `# noqa` for safe subprocess patterns. Ruff security scan goes from 5 findings → 0. No CI checks defined on the repo.

## Findings

- [warning] `_cli.py:39` — the `hasattr(tf, "extraction_filter")` check in `_safe_extract_tar` is redundant with `sys.version_info >= (3, 12)`. Both resolve to the same set of Python versions (extraction_filter is a class attribute added in 3.12 per PEP 706). Not dead code (verified: `hasattr(tarfile.TarFile, 'extraction_filter')` returns True on 3.12+), but the belt-and-suspenders is noise. Simplification: drop the hasattr branch.

- [warning] No test coverage for `_safe_extract_tar` / `_safe_extract_zip`. The package has no test suite, so this is a pre-existing gap, but security-critical code benefits from regression protection. Suggested minimum coverage for a follow-up:
  - Tarball member with `../etc/passwd` → rejected
  - Tarball member with absolute path `/etc/passwd` → rejected
  - Tarball symlink → rejected
  - Tarball hardlink → rejected
  - Zipfile equivalents (no symlink/hardlink support in standard zip)
  - Clean tarball passes through unchanged

- [nit] `_cli.py:43-47` — symlink/hardlink rejection comments could note *why* symlinks are dangerous even though extractall doesn't follow them by default (they're the mechanism for link-based escape in the classic tar-slip attack chain).

- [nit] `_cli.py:144` — `# noqa: S310 — scheme validated above` could be more precise ("scheme validated at line 132") to survive future refactors.

- [info] URL scheme validation is over-defensive (URL is built from hardcoded constants today). Deliberate defense-in-depth — fine to keep.

- [info] Fallback path iterates tar members twice (once to validate, once via `extractall`). Acceptable for security-critical code, but worth noting.

## CI Status

No CI checks defined for this repo. Verification is manual via `ruff check _cli.py --select S` (clean) + syntax check.

## Verdict

**COMMENT** — functionally correct. The `hasattr` redundancy (line 39) would be a useful 1-line cleanup but doesn't affect correctness. Tests are a pre-existing gap worth filing as a separate issue rather than blocking this PR.

🤖 Reviewed by Claude Code reviewer agent (note: original reviewer claim that hasattr "always returns False" was incorrect — verified empirically it resolves True on 3.12+. The redundancy point stands, but both branches are live.)

dLo999:
Following up with actual behavioral verification — the original PR body's security claims were based on ruff-scan-passes, which proves semgrep rules don't fire but doesn't prove the fix blocks real attack scenarios.

Wrote a local test harness that builds malicious tar/zip archives and exercises both code paths. All 17 tests pass:

**Benign extraction:** clean tarball/zip both extract successfully

**Python 3.12+ path (`filter='data'`):**
- tar with `../escape.txt` → rejected via `OutsideDestinationError`
- tar with `/absolute/path` → safely extracted inside dest (filter strips leading `/`)
- tar with symlink to `/etc/passwd` → rejected via `AbsoluteLinkError`
- tar with hardlink to `/etc/passwd` → rejected via `AbsoluteLinkError`

**Fallback path (Python 3.8-3.11, forced via monkeypatch):**
- same 4 attacks → all rejected via `sys.exit` with clear error messages

**Zip path (no built-in filter):**
- zip with `../escape.txt` → rejected
- zip with `/absolute/path` → rejected

**URL scheme validation:** https accepted, http/file/ftp all rejected

No test file landed outside the destination directory in any attack scenario. Test harness available if the maintainers want it added to a repo test suite.

DeusData:
Hey thx for this, will review in the evening

DeusData:
@dLo999 thx for this, looks clean. Would you mind also reviewing the npm one? That would be awesome!

PR #249, opened 2026-04-13
fix: GitHub Actions shell injection in _build.yml (#247)

### LINKED ISSUE #247: security: GitHub Actions shell injection in .github/workflows/_build.yml
## Summary

`.github/workflows/_build.yml` has multiple `run:` steps that interpolate `${{ github.* }}` context data directly into shell scripts. This is the classic GitHub Actions shell injection pattern (CWE-78, OWASP A03).

Found via static scan (semgrep `yaml.github-actions.security.run-shell-injection`).

## Findings

Lines with `${{ github.* }}` context interpolation in `run:` steps:

- `.github/workflows/_build.yml:55`
- `.github/workflows/_build.yml:68`
- `.github/workflows/_build.yml:111`
- `.github/workflows/_build.yml:123`
- `.github/workflows/_build.yml:162`

(Plus any I missed — these are the top 5 from the semgrep output.)

## Why this matters

Fields like `github.event.pull_request.title`, `github.event.issue.title`, `github.head_ref`, and `github.event.comment.body` can be attacker-controlled via crafted PR titles, branch names, or comments. When they get interpolated directly into a `run:` script, the attacker's input becomes shell code.

Reference: https://securitylab.github.com/research/github-actions-untrusted-input/

Impact: code execution on the runner, which typically has access to `GITHUB_TOKEN`, release signing keys, PyPI publishing credentials, etc.

## Remediation pattern

Replace direct interpolation with an intermediate env var:

```yaml
# Vulnerable:
run: |
  echo "Processing ${{ github.event.pull_request.title }}"

# Safe:
env:
  PR_TITLE: ${{ github.event.pull_request.title }}
run: |
  echo "Processing $PR_TITLE"
```

The `env:` block treats the value as a plain string; referencing `$PR_TITLE` in the script goes through normal shell variable expansion which is not vulnerable to injection.

## Notes

- A previous scan of this repo (~April 12) flagged similar findings in `.github/workflows/ci.yml`. The current findings are in `_build.yml` which appears to be a newer workflow file — same pattern, different file.
- GitHub's own hardening guide recommends this pattern for any `run:` step that references `github.*` context: https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#understanding-the-risk-of-script-injections

## Environment

- Scan date: 2026-04-13
- Tool: semgrep (default ruleset, rule: `yaml.github-actions.security.run-shell-injection`)
- Repo state: HEAD of main as of scan time

dLo999:
Closes #247

## Summary

Replace direct `${{ ... }}` interpolation in `run:` steps of `.github/workflows/_build.yml` with `env:` blocks per GitHub's hardening guide. All 6 `run-shell-injection` findings from semgrep resolved.

## What changed

Every workflow input (`inputs.version`) and matrix value (`matrix.cc`, `matrix.cxx`, `matrix.goos`, `matrix.goarch`, `matrix.arch`) now flows into the shell through a named environment variable rather than being textually spliced into the script.

### Pattern applied

```yaml
# Before:
- name: Build standard binary
  run: scripts/build.sh ${{ inputs.version && format('--version {0}', inputs.version) || '' }} CC=${{ matrix.cc }} CXX=${{ matrix.cxx }}

# After:
- name: Build standard binary
  env:
    VERSION: ${{ inputs.version }}
    CC: ${{ matrix.cc }}
    CXX: ${{ matrix.cxx }}
  run: |
    if [ -n "$VERSION" ]; then
      scripts/build.sh --version "$VERSION" "CC=$CC" "CXX=$CXX"
    else
      scripts/build.sh "CC=$CC" "CXX=$CXX"
    fi
```

This is safe against injection because:

1. The `env:` block treats values as opaque strings during YAML parsing — GitHub does not interpret shell metacharacters in env values
2. `$VAR` expansion in the script uses normal shell semantics, so attacker input in `$VERSION` becomes a literal argv entry, not interpreted shell code
3. The if/else form avoids word-splitting pitfalls of conditional arg construction

## Changes

- `.github/workflows/_build.yml` — +66 / −10

Covered all 6 originally-flagged steps plus the archive steps that used `${{ matrix.* }}` in tar filenames (consistency).

## Test results

**Static verification:**
- Before: 6 `run-shell-injection` findings
- After: **0 findings** — `code-audit scan` / semgrep clean on `_build.yml`

**YAML syntax:** `yaml.safe_load(_build.yml)` parses cleanly

**Behavior preservation:**
- Same `build.sh` invocations, same artifact names, same conditional version handling
- Only functional difference: version-handling branch is now an `if/else` instead of a YAML ternary, which produces byte-identical arguments to `build.sh`

### Verification gate

```
VERIFICATION GATE — Issue #247:
  [N/A]   Build: GH Actions workflow — only runs on CI, not locally
  [N/A]   Test suite: yaml-only change, no python tests
  [PASS]  Issue scenario: scan of _build.yml — was 6 findings, now 0
  [PASS]  Edge cases: --version empty/set branches tested logically; matrix
          values moved to env; archive filenames use $GOOS/$GOARCH/$ARCH
  [PASS]  YAML syntax valid
  [N/A]   README: internal CI config, not user-facing
RESULT: PASS (3 PASS + 3 N/A)
```

Reference: https://securitylab.github.com/research/github-actions-untrusted-input/

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -52,20 +52,41 @@ jobs:
           node-version: "22"
 
       - name: Build standard binary
-        run: scripts/build.sh ${{ inputs.version && format('--version {0}', inputs.version) || '' }} CC=${{ matrix.cc }} CXX=${{ matrix.cxx }}
+        env:
+          VERSION: ${{ inputs.version }}
+          CC: ${{ matrix.cc }}
+          CXX: ${{ matrix.cxx }}
+        run: |
+          if [ -n "$VERSION" ]; then
+            scripts/build.sh --version "$VERSION" "CC=$CC" "CXX=$CXX"
+          else
+            scripts/build.sh "CC=$CC" "CXX=$CXX"
+          fi
 
       - name: Ad-hoc sign macOS binary
         if: startsWith(matrix.os, 'macos')
         run: codesign --sign - --force build/c/codebase-memory-mcp
 
       - name: Archive standard binary
+        env:
+          GOOS: ${{ matrix.goos }}
+          GOARCH: ${{ matrix.goarch }}
         run: |
           cp LICENSE install.sh build/c/
-          tar -czf codebase-memory-mcp-${{ matrix.goos }}-${{ matrix.goarch }}.tar.gz \
+          tar -czf "codebase-memory-mcp-${GOOS}-${GOARCH}.tar.gz" \
             -C build/c codebase-memory-mcp LICENSE install.sh
 
       - name: Build UI binary
-        run: scripts/build.sh --with-ui ${{ inputs.version && format('--version {0}', inputs.version) || '' }} CC=${{ matrix.cc }} CXX=${{ matrix.cxx }}
+        env:
+          VERSION: ${{ inputs.version }}
+          CC: ${{ matrix.cc }}
+          CXX: ${{ matrix.cxx }}
+        run: |
+          if [ -n "$VERSION" ]; then
+            scripts/build.sh --with-ui --version "$VERSION" "CC=$CC" "CXX=$CXX"
+          else
+            scripts/build.sh --with-ui "CC=$CC" "CXX=$CXX"
+          fi
 
       - name: Ad-hoc sign macOS UI binary
         if: startsWith(matrix.os, 'macos')
@@ -76,9 +97,12 @@ jobs:
         run: scripts/security-ui.sh
 
       - name: Archive UI binary
+        env:
+          GOOS: ${{ matrix.goos }}
+          GOARCH: ${{ matrix.goarch }}
         run: |
           cp LICENSE install.sh build/c/
-          tar -czf codebase-memory-mcp-ui-${{ matrix.goos }}-${{ matrix.goarch }}.tar.gz \
+          tar -czf "codebase-memory-mcp-ui-${GOOS}-${GOARCH}.tar.gz" \
             -C build/c codebase-memory-mcp LICENSE install.sh
 
       - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
@@ -108,7 +132,14 @@ jobs:
 
       - name: Build standard binary
         shell: msys2 {0}
-        run: scripts/build.sh ${{ inputs.version && format('--version {0}', inputs.version) || '' }} CC=clang CXX=clang++
+        env:
+          VERSION: ${{ inputs.version }}
+        run: |
+          if [ -n "$VERSION" ]; then
+            scripts/build.sh --version "$VERSION" CC=clang CXX=clang++
+          else
+            scripts/build.sh CC=clang CXX=clang++
+          fi
 
       - name: Archive standard binary
         shell: msys2 {0}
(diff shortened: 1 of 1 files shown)


dLo999:
Adding a verification note since the original PR body's 'byte-identical arguments' claim was based on diff inspection rather than execution.

Ran argv-equivalence test locally by mocking `scripts/build.sh` to print argv and executing both the pre-fix YAML-substituted form and the post-fix if/else form for all 6 cases (3 build.sh invocations × 2 VERSION states). Each produced identical argv.

Sample — `inputs.version=""`:
- Old YAML `scripts/build.sh ${{ empty_ternary }} CC=gcc CXX=g++` → `["CC=gcc", "CXX=g++"]` after bash parsing
- New if/else (VERSION empty branch) → `["CC=gcc", "CXX=g++"]`

Sample — `inputs.version="v0.8.0"`:
- Old → `["--version", "v0.8.0", "CC=gcc", "CXX=g++"]`
- New → `["--version", "v0.8.0", "CC=gcc", "CXX=g++"]`

What's still unverified locally: actual GH Actions runtime (matrix + env propagation, MSYS2 shell behavior on Windows). These are standard GHA features + POSIX `[ -n "$x" ]` respectively, so low risk, but the first real CI run is the definitive test.

dLo999:
## Second-pass review (internal quality check before merge)

Ran the reviewer agent against this PR. Summary of genuinely useful feedback vs false-positive findings, with empirical verification where possible.

### Real findings (acknowledged)

- **[low] Input validation not addressed.** The fix prevents shell-injection via metacharacter escape, but doesn't validate `inputs.version` against values that could confuse `build.sh` itself (e.g., `--help`, `-x`, path traversal). Out of scope for a shell-injection fix, but worth noting if the project wants defense in depth. Would typically be addressed at the workflow-input schema level, not the run: step.

- **[nit] "Byte-identical arguments" framing is slightly imprecise.** The argv is byte-identical; the *shell quoting* differs (archive filenames are now `"..."` quoted, CC/CXX use `"CC=$CC"` instead of bare). All cosmetic — no observable runtime difference for the actual matrix values — but the claim would be more accurate as "argv-identical."

- **[nit] Windows MSYS2 step hardcodes `CC=clang CXX=clang++` bare while Unix uses `"CC=$CC" "CXX=$CXX"`.** Inconsistent style, not a correctness issue. A follow-up refactor could move Windows to env vars too for uniformity.

### False-positive findings (retracted)

- ~~CRITICAL: CC/CXX argument format divergence between Unix and Windows jobs~~ — empirically tested: both forms produce identical argv shape (`["--version", "v0.8.0", "CC=value", "CXX=value"]`). The difference is shell quoting, not argv construction. Both are positional args to `build.sh`; neither form sets env vars for the subshell.

- ~~CI status: no checks~~ — CI is a reusable workflow (`on: workflow_call`), so branch-level CI doesn't auto-trigger. Real test is a caller workflow invoking `_build.yml` with `inputs.version` set.

### What's genuinely unverified

- MSYS2 shell on Windows: `if [ -n "$VERSION" ]` is POSIX, should work, but untested on the actual runner
- GHA `env:` + `matrix.*` propagation: standard feature, but untested end-to-end

Both require a real CI run. The maintainer's first push after merge will be the definitive test. If anything breaks, the fixes are small (add quoting, or swap bare→env-var).

DeusData:
Thanks @dLo999 — clean, focused application of the standard GitHub Actions hardening pattern. Verified:

- Only `.github/workflows/_build.yml` touched, no SHA bumps or new actions slipping in alongside
- Each rewritten step preserves the existing command-line: `build.sh "CC=$CC" "CXX=$CXX"` and `tar -czf "codebase-memory-mcp-${GOOS}-${GOARCH}.tar.gz"` produce byte-identical args / filenames to the previous YAML ternary form
- The if/else branch on `$VERSION` is the right choice over a single quoted invocation, because empty `--version ""` would be a real argv entry that `build.sh` would have to special-case
- semgrep `run-shell-injection` count of 6 → 0 matches the diff site count

Merging now.

PR #257, opened 2026-04-16
feat: add temporal properties to FILE_CHANGES_WITH edges and File nodes

Kagin007:
- Fix search_graph (Mode A): emit_search_results now calls enrich_node_properties so node properties_json is included in results. Uses a collected-docs pattern to correctly handle yyjson zero-copy string lifetime.

- Fix query_graph bare edge return (Mode B): binding_get_virtual now returns e->properties_json instead of "" when prop is NULL, so `RETURN r` on an edge exposes the full properties object.

- Add last_co_change to FILE_CHANGES_WITH edges: commit_t and cbm_commit_files_t gain a timestamp field; cbm_compute_change_coupling maintains a parallel pair_timestamps hash table tracking the max commit timestamp per pair; cbm_change_coupling_t gains last_co_change; the edge props JSON now includes "last_co_change":<unix-epoch>. Both libgit2 and popen code paths updated consistently.

- Add last_modified and change_count to File nodes: new cbm_file_temporal_t struct and MAX_FILE_TEMPORAL=16384 constant; cbm_githistory_result_t extended with file_temporal array; cbm_pipeline_githistory_compute builds per-file change_count + last_modified via a hash table pass; cbm_pipeline_githistory_apply upserts File nodes with temporal props (re-derives extension to preserve it). Both pipeline entry points (pass_githistory and run_githistory) free file_temporal and trigger apply when file_temporal_count > 0.

- Update test_pipeline.c initializers for new cbm_commit_files_t.timestamp field (zero-initialized, backward-compatible behavior).

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -2118,7 +2118,7 @@ static const char *binding_get_virtual(binding_t *b, const char *var, const char
     /* Fall through to normal lookup */
     cbm_edge_t *e = binding_get_edge(b, var);
     if (e) {
-        return prop ? edge_prop(e, prop) : "";
+        return prop ? edge_prop(e, prop) : (e->properties_json ? e->properties_json : "{}");
     }
     cbm_node_t *n = binding_get(b, var);
     if (n) {
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -1260,10 +1260,19 @@ static char *bm25_search(cbm_store_t *store, const char *project, const char *qu
     return json;
 }
 
-/* Emit the cbm_store_search results as a JSON "results" array on the doc. */
+/* Forward declaration — defined later in this file. */
+static yyjson_doc *enrich_node_properties(yyjson_mut_doc *doc, yyjson_mut_val *obj,
+                                          const char *properties_json);
+
+/* Emit the cbm_store_search results as a JSON "results" array on the doc.
+ * Populates *out_pdocs with parsed property docs that must be freed by the
+ * caller AFTER serializing doc (zero-copy strings point into these docs). */
 static void emit_search_results(yyjson_mut_doc *doc, yyjson_mut_val *root,
                                 const cbm_search_output_t *out, cbm_store_t *store,
-                                const char *relationship, bool include_connected, int offset) {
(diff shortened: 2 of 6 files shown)


DeusData:
Landed as 4e80d2f on main with you as the commit author. Distilled the three changes from this PR into a single squashed commit and added the missing piece: explicit regression coverage for each behavior change.

What's in the merged version:

- **Mode A (search_graph node properties):** kept your collected-pdocs lifecycle for the yyjson zero-copy strings; both exit paths in `handle_search_graph` (success and `sq_type_error`) free the parked docs after `yy_doc_to_str` runs.
- **Mode B (Cypher bare edge return):** `binding_get_virtual` returns `e->properties_json` or `"{}"` instead of `""` when prop is NULL.
- **Temporal data:** `last_co_change` on FILE_CHANGES_WITH edges plus `change_count` / `last_modified` on File nodes. Both git-log paths (libgit2 and popen with `--pretty=format:COMMIT:%H:%ct`) populate the timestamp; the parallel `pair_timestamps` hash table tracks the max per pair. `cbm_pipeline_githistory_apply` now upserts the temporal props onto File nodes so hotspot/recency queries work without a separate pass.

Two changes to your version:
1. The timestamp fields are typed as `long long` rather than `time_t` so the on-disk JSON shape stays consistent across 32-/64-bit time_t platforms.
2. The extension stored on the File node properties is run through `cbm_json_escape` before `snprintf`. Practical risk was tiny — real extensions never contain quotes or backslashes — but a hypothetical `foo.bar\"baz` would have produced malformed JSON, and the escape is cheap.

Tests added (one per behavior change):
- `githistory_coupling_carries_last_co_change` — pins last_co_change as the max of per-pair commit timestamps with an unrelated pair interleaved in input order
- `cypher_bare_edge_return_exposes_properties_json` — `RETURN r` on an HTTP_CALLS edge must surface url_path in the response (was empty string before)
- `tool_search_graph_includes_node_properties` — search_graph response must contain signature/return_type/is_exported keys for a node whose properties_json carries them

Full suite: 2840 passed, 0 failed under ASan + UBSan. Closing in favor of the cherry-picked commit.

PR #262, opened 2026-04-16
fix(mcp): correct project metadata and snippet paths

himanshuranjann:


DeusData:
Thanks for the work, @himanshuranjann, but I can't take this PR upstream as-is and after looking through the diff carefully I don't think the right move is for me to cherry-pick pieces of it either.

The PR is +10,530 / −18 across 41 files, but the changes that map to the title (`fix(mcp): correct project metadata and snippet paths`) are about ~150 lines in `src/mcp/mcp.c` and `tests/test_mcp.c`. The remaining ~10,300 lines are GoHighLevel-internal infrastructure that shouldn't live in the upstream repo:

- `ghl/cmd/server` (1,289 lines) and supporting Go packages (`ghl/internal/auth`, `bridge`, `cachepersist`, `discovery`, `indexer`, `manifest`, `mcp`, `webhook`) — a complete fleet-orchestration server tied to your operational environment
- `Dockerfile.ghl`, `cloudbuild.ghl.yaml`, `.gcloudignore` — your build / deploy infrastructure
- `deployments/ghl/helm/*` — your GKE Helm chart
- `REPOS.yaml` (2,897 lines) and `REPOS.local.yaml` listing 480 internal GoHighLevel repositories grouped by team — that's an org-structure document that almost certainly shouldn't end up in a public repo, and even if you're comfortable with it being public it doesn't belong here
- `docs/CBM_VS_PROJECT_ORION_COMPARISON.md` — internal comparison doc
- 21 commits including merges from `feature/cbm-mcp-scale` into `feature/uptrace-codebase-memory-intg` — the branching is set up for your fork's lifecycle, not an upstream PR

The MCP-side changes do contain three real ideas:

1. Consolidating the duplicated `.db` filter into `is_project_db_file` as the single source of truth.
2. Reading the canonical project name from the indexed DB via `cbm_store_list_projects` rather than relying on the filename slug.
3. Letting `resolve_snippet_source` accept absolute file paths (the existing realpath containment check still applies).

But each of them ships with a fleet-specific bias I'd want to revisit:

- The refactor also drops the `tmp-` prefix exclusion (and adds a test that depends on it being gone). That exclusion exists upstream to hide partially-written reindex DBs from `list_projects`; surfacing them as legitimate projects is a regression for non-fleet users.
- `handle_index_repository` gains a checkpoint + force-close + clear-current-project at the end. The justification in the commit body is explicitly "so the fleet layer can snapshot the artifact", which doesn't apply to upstream MCP usage and silently changes the long-lived-store contract.
- The `cbm_path_is_absolute` helper has an operator-precedence pattern (`a || b || c && d`) that compiles correctly under C precedence rules but is exactly the kind of expression `-Wlogical-op-parentheses` and reviewers will flag.

If you want the project-metadata fix and the absolute-snippet-path fix in upstream, please open a focused PR scoped to those two specific changes — just `src/mcp/mcp.c` and `tests/test_mcp.c`, no `ghl/`, no `REPOS.yaml`, no Helm chart — and I'll review and merge that promptly. Closing this one as not-mergeable rather than letting it sit open.

PR #264, opened 2026-04-17
Update agents badge count from 10 to 11

jsyrjala:
There are now 11 supported agents.


jsyrjala:
The badge was updated in another PR.

PR #265, opened 2026-04-17
Add Nix flake definition for reproducible builds

josephvoss:
Adds `flake.nix` to define how this package should be built on systems running [nix][nix-link]. Also commit the `flake.lock` file to pin the specific dependencies at build time.

By default support aarch64 and x86_64 for macos and linux. Run the Make command directly because `cc` on Nix systems is a wrapper file and not a binary directly, so the `file` check in `scripts/build.sh` fails to identify the host architecture correctly.

Include lib2git as an optional dev shell dependency since it's detected by `pkg-config` in the build script to link to `lib2git` if it exists at build time.

[nix-link]: https://nixos.org/

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1776255774,
+        "narHash": "sha256-psVTpH6PK3q1htMJpmdz1hLF5pQgEshu7gQWgKO6t6Y=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "566acc07c54dc807f91625bb286cb9b321b5f42a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixpkgs-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks @josephvoss — clean, focused PR with proper supply-chain hygiene (pinned nixpkgs commit + narHash, no random fetchurl, only standard nixpkgs build inputs). Verified:

- Two new files only, fully opt-in — `nix build` / `nix develop` paths only; non-Nix users are unaffected
- License declaration in `flake.nix` matches the repo's MIT LICENSE
- The `scripts/build.sh` bypass is well-justified — `cc` as a Nix wrapper script breaks the `file`-based arch detection, and Nix stdenv already provides the correct compiler context
- buildPhase calls the existing `Makefile.cbm cbm` target with `-j$NIX_BUILD_CORES`; no new build steps, no command injection surface
- `flake.lock` lastModified resolves to 2026-04-14, consistent with the PR open date

Two small follow-ups I'll either drive or accept later — not blockers:
- The hardcoded `version = "0.6.0"` will drift; could be templated from `git describe` or fed via an env var
- `flake.lock` will need periodic refreshes for nixpkgs security patches (standard Nix maintenance)

Merging now.

PR #273, opened 2026-04-18
feat(cbm): add full Pine Script support to C memory graph server

vinay-veerappa:


vinay-veerappa:
add pinescript indexing

DeusData:
Landed Pine Script support as 048f1b4 on main with you as the commit author. I distilled the change to use the project's existing language-add pipeline rather than vendoring the grammar verbatim from this PR — wanted to get the supply chain story right and avoid a couple of issues that would have made an as-is merge tricky:

- The PR's `parser.c` was a 46k-line drop without provenance — no upstream URL, no LICENSE, and the `tree_sitter_pine` symbol matched several plausible grammars. I confirmed via byte-identical `scanner.c` that the source was [kvarenzn/tree-sitter-pine](https://github.com/kvarenzn/tree-sitter-pine), then re-vendored from there via `scripts/vendor-grammar.sh` so the provenance is recorded. The upstream is ISC-licensed (declared in `package.json` rather than a top-level LICENSE file), so I wrote a vendored `LICENSE` with the ISC text + upstream URL for clarity.

- The PR's wiring choices in `lang_specs.c` were correct — I cross-checked all 12 node names (`function_declaration_statement`, `type_definition_statement`, `source_file`, `call`, `variable_definition_statement`, `tuple_declaration_statement`, `if_statement`, `switch_statement`, `for_statement`, `for_in_statement`, `while_statement`, `reassignment_statement`) against `grammar.js` and they all exist. Kept all of them.

- The PR also bundled an unrelated `Makefile.cbm` change (`scripts/embed-frontend.sh` → `sh scripts/embed-frontend.sh`) which is a different concern from Pine Script; dropped that.

- Registered Pine in `scripts/new-languages.json` so the language-add pipeline tracks it alongside the other 89 entries, and ran `scripts/audit-grammar-security.sh` on the vendored grammar — clean (no system/popen/exec/socket/mmap/dlopen, only standard string/stdint/stdbool/stdlib includes, no constructor/destructor attrs, no inline asm).

You're co-authored on the squashed commit. Full suite passes at 2840 / 0 under ASan + UBSan. Thanks for the work — the node-type selections were the most useful part of the PR and the upstream grammar identification fell out from your scanner.c being byte-identical to kvarenzn's.

Closing in favor of the cherry-picked commit.

PR #279, opened 2026-04-21
fix: INHERITS edges missing for Java extends+implements

loaychlih:
## Problem

`extract_base_classes()` in `internal/cbm/extract_defs.c` had two bugs
that caused most INHERITS edges to be missing for Java (and likely other
languages using similar AST field names).

### Bug 1 — Early return loses multiple inheritance targets

The field loop returned immediately on the first match:

```c
for (const char **f = fields; *f; f++) {
    TSNode super = ts_node_child_by_field_name(node, *f, ...);
    if (!ts_node_is_null(super)) {
        return make_single_base(...);  // ← returns here, never checks interfaces
    }
}
```

For `class Foo extends Bar implements Baz`, it found `superclass` → returned
immediately → never processed `super_interfaces`. Result: 1 edge instead of 2.

### Bug 2 — cbm_node_text returns full node text including keywords

Calling `cbm_node_text` on the `superclass` field node returned
`"extends Bar"` instead of `"Bar"`. On `super_interfaces` it returned
`"implements Baz, Qux"` instead of the individual names.

## Fix

Added `collect_bases_from_field()` which:
- Walks into child AST nodes to extract `type_identifier` / `generic_type` /
  `qualified_name` text directly (skips keyword nodes like `extends`/`implements`)
- Handles `type_list` / `interface_type_list` children for multiple interfaces
- Strips generic args at `<` (e.g. `List<String>` → `List`)
- Falls back to raw `cbm_node_text` for languages where the field IS the type name

The field loop now collects from **all** matching fields before returning.

## Result

Tested on [Modelio](https://github.com/ModelioOpenSource/Modelio)
(Java codebase, 253 files):

|                 | Before | After   |
|-----------------|--------|---------|
| INHERITS edges  | 3      | **116** |

Example edges now correctly emitted:
- `DefaultLinkTool` → `DefaultDiagramTool` (extends)
- `DefaultLinkTool` → `ILinkTool` (implements)

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -1083,6 +1083,83 @@ static const char **extract_csharp_base_list(CBMArena *a, TSNode node, const cha
     return NULL;
 }
 
+// Walk a field node and collect type identifier names into out[].
+// Handles: direct type_identifier/generic_type/qualified_name, type_list children
+// (Java interfaces list), and raw text fallback (other languages).
+static int collect_bases_from_field(CBMArena *a, TSNode field_node, const char *source,
+                                    const char **out, int out_cap) {
+    int count = 0;
+    const char *fk = ts_node_type(field_node);
+
+    // If the field node itself is a type node, extract directly.
+    if (strcmp(fk, "type_identifier") == 0 || strcmp(fk, "generic_type") == 0 ||
+        strcmp(fk, "qualified_name") == 0 || strcmp(fk, "scoped_type_identifier") == 0 ||
+        strcmp(fk, "user_type") == 0) {
+        char *t = cbm_node_text(a, field_node, source);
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks @loaychlih — clean diagnosis (two distinct bugs in the same function, with a nice repro on Modelio showing 3 → 116 INHERITS edges) and a bounds-safe fix that drops in cleanly alongside the existing arena-pool pattern in this file. Verified: max writes ≤ `MAX_BASES_MINUS_1` with the per-iteration `count < out_cap` checks, no new system/network/file calls, the `count == 0` fallback correctly preserves behavior for non-Java grammars where the field IS the type name. Merging now and pushing a small follow-up that adds a regression test for the extends+implements case so this doesn't silently regress.

PR #281, opened 2026-04-21
fix: wire cbm_store_get_architecture into handle_get_architecture MCP handler

### LINKED ISSUE #280: get_architecture does not return architecture aspects described in README (only histograms / schema)
## Summary

The `get_architecture` MCP tool does **not** return the rich architecture sections promised in the README (languages, packages, entry points, routes, hotspots, boundaries, layers, clusters). Instead, it only returns histograms: `node_labels`, `edge_types`, and (sometimes) `relationship_patterns`, along with `project`, `total_nodes`, and `total_edges`.

## Details and Evidence
- According to `README.md` and published documentation, `get_architecture` is supposed to return an “architecture overview” – with languages, packages, entry points, routes, hotspots, boundaries, layers, clusters, etc.
- In the current implementation (as of v0.6.0, commit `bebc6d8...`), the `handle_get_architecture` function does **not** call the actual architecture computation layer (`cbm_store_get_architecture`).
- Instead, the only conditionally emitted outputs (contingent on `aspects`) are:
    - `node_labels` (when requesting aspect `"structure"`)
    - `edge_types` (when requesting aspect `"dependencies"`)
    - `relationship_patterns` (when requesting aspect `"routes"`, and only if available)
- No other architecture fields/sections are present, regardless of what’s passed in `aspects` (including `"all"`).
- There is never a top-level field or sub-object for `languages`, `packages`, `entry_points`, `routes` (beyond `relationship_patterns`), `hotspots`, `boundaries`, `layers`, `clusters`, or anything more structured than these label/type histograms.
- The underlying repo’s store and data model **do** have support for these richer outputs (`cbm_store_get_architecture`, `cbm_architecture_info_t`, etc.), but the MCP handler does not use or expose them. This is not just a documentation error—the implementation simply does not serialize those fields in its responses.

## Impact
- Cursor, the CLI, and all clients receive only histogram summaries. There is no way to get e.g. a package or dependency list, route summaries, cluster/layer outputs, or cross-package boundaries via the public API.
- This makes much of the README’s description for `get_architecture` and most “architecture-aware” integrations misleading or broken.

## Example
A typical `get_architecture` response (as of v0.6.0):
```json
{
  "project": "demo",
  "total_nodes": 550,
  "total_edges": 2140,
  "node_labels": [ ... label histogram ... ],
  "edge_types": [ ... type histogram ... ]
}
```
- There are never block sections for languages/packages/routes/etc.
- aspect blocks or `aspects: ['all']` do not change or expand the available sections.

## Steps to reproduce
1. Index any (sufficiently large) repository.
2. Call `get_architecture` via JSON-RPC with only `{ "project": "name" }` or any combination of `aspects`.
3. Observe that only label/type histograms are returned, not the additional architecture sections described in the README.
   
## Suggested resolution
- Update `handle_get_architecture` in `src/mcp/mcp.c` to actually call `cbm_store_get_architecture(...)` and serialize *all* available aspects/sections (languages, packages, entry_points, etc.) in the returned JSON, matching README coverage.
- Or: if architectural extraction is not intended to be available, clearly update the README and docs to match the much more restricted result set.

OliverEvans96:
`get_architecture` only ever returned label/type histograms (`node_labels`, `edge_types`, `relationship_patterns`) because `handle_get_architecture` never called `cbm_store_get_architecture` — the rich architecture computation layer was completely bypassed regardless of which `aspects` were requested.

## Changes

- **Extract aspects for store dispatch** — parse the JSON `aspects` array into a `const char*[]` and pass it to `cbm_store_get_architecture`, which already handles `"all"` and per-name filtering internally
- **Serialize all populated architecture sections** — after calling the store, emit each non-empty field from `cbm_architecture_info_t`:
  - `languages` (`language`, `file_count`)
  - `packages` (`name`, `node_count`, `fan_in`, `fan_out`)
  - `entry_points` (`name`, `qualified_name`, `file`)
  - `routes` (`method`, `path`, `handler`)
  - `hotspots` (`name`, `qualified_name`, `fan_in`)
  - `boundaries` (`from`, `to`, `call_count`)
  - `layers` (`name`, `layer`, `reason`)
  - `file_tree` (`path`, `type`, `children`)
- **Preserve existing histogram fields** — `node_labels`, `edge_types`, `relationship_patterns`, and `cross_repo_links` are unchanged
- **Proper cleanup** — `cbm_store_architecture_free` called before return

Fixes #280

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -1738,11 +1738,11 @@ static char *handle_get_architecture(cbm_mcp_server_t *srv, const char *args) {
     const char *aspects_strs[MCP_COL_16];
     int aspects_strs_count = 0;
     if (aspects_arr) {
-        size_t aidx;
-        size_t amax;
-        yyjson_val *aval2;
-        yyjson_arr_foreach(aspects_arr, aidx, amax, aval2) {
-            const char *s = yyjson_get_str(aval2);
+        size_t aspect_idx;
+        size_t aspect_max;
+        yyjson_val *aspect_val;
+        yyjson_arr_foreach(aspects_arr, aspect_idx, aspect_max, aspect_val) {
+            const char *s = yyjson_get_str(aspect_val);
             if (s && aspects_strs_count < MCP_COL_16) {
                 aspects_strs[aspects_strs_count++] = s;
             }
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -1733,8 +1733,29 @@ static char *handle_get_architecture(cbm_mcp_server_t *srv, const char *args) {
         }
     }
 
+    /* Build a C string array from aspects for cbm_store_get_architecture.
+     * Strings point into aspects_doc memory so aspects_doc must outlive this array. */
+    const char *aspects_strs[MCP_COL_16];
+    int aspects_strs_count = 0;
+    if (aspects_arr) {
+        size_t aidx;
+        size_t amax;
+        yyjson_val *aval2;
+        yyjson_arr_foreach(aspects_arr, aidx, amax, aval2) {
+            const char *s = yyjson_get_str(aval2);
+            if (s && aspects_strs_count < MCP_COL_16) {
+                aspects_strs[aspects_strs_count++] = s;
+            }
+        }
+    }
+
     cbm_schema_info_t schema = {0};
     cbm_store_get_schema(store, project, &schema);
+
+    cbm_architecture_info_t arch = {0};
+    cbm_store_get_architecture(store, project,
+                               aspects_strs_count > 0 ? aspects_strs : NULL,
+                               aspects_strs_count, &arch);
 
     int node_count = cbm_store_count_nodes(store, project);
     int edge_count = cbm_store_count_edges(store, project);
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks @OliverEvans96 — clean diagnosis (the architecture computation was sitting upstream of the MCP boundary, completely bypassed) and the wiring is straightforward. Verified the API surface (`cbm_store_get_architecture`, `cbm_architecture_info_t`, `cbm_store_architecture_free` all in store.h), the `aspects_doc` lifetime story checks out (your comment is correct), the stack buffer is capped at MCP_COL_16, and `yyjson_mut_obj_add_str` on NULL silently no-ops (no crash surface).

Merging now and pushing a small follow-up that:
- adds the `x ? x : ""` NULL coercion to match the rest of mcp.c (lines 1305, 1308, 2381, 2382, 2383 in current main do this consistently — without it, a NULL field becomes a missing JSON key rather than an empty string)
- serializes `clusters` and `services` so the architecture computation isn't dropping data on the floor for those two aspects
- adds an MCP-level regression test asserting the new sections appear when the store has architecture data

Issue #292, opened 2026-04-26
Protocol-aware cross-repo intelligence: gRPC + typed-message + attribute-route matching (design proposal)


sponger94:
> **Status (2026-04-26):** Tier 1a (consumer-side gRPC IDL Routes + HANDLES binding) is open as #293 against this issue. Tier 1b (producer-side typed-client `GRPC_CALLS` emission) and Tiers 2–4 will follow as separate PRs sequenced per §5 below.

---

# Protocol-Aware Cross-Repo Intelligence

**Status:** Design proposal for upstream `codebase-memory-mcp`
**Audience:** cbm maintainer + reviewers
**Scope:** Extends `pass_cross_repo` from literal-string matching to protocol-aware matching across the four cross-language patterns that account for >95% of inter-service communication in modern codebases.
**Compatibility:** Strictly additive. No breaking changes to existing tools, edges, or APIs. Builds on PR #281 (rich `get_architecture` fields).

---

## TL;DR

cbm already has the scaffolding for cross-repo intelligence (`pass_cross_repo.c`, `CROSS_HTTP_CALLS` / `CROSS_ASYNC_CALLS` / `CROSS_GRPC_CALLS` edge types, named-route matching). The current implementation only fires when a call site has a **literal URL or topic string** as its first argument. That covers idiomatic Python/Node code well, but misses the dominant pattern in modern strongly-typed stacks (Java/Spring, .NET, Kotlin, Go-with-codegen): **typed clients and message handlers where the routing identifier is a generic type parameter, an interface ancestor, an attribute, or a config-resolved name** — never a literal string at the call site.

This proposal adds four protocol-aware extraction tiers, each language-generic, behind a YAML-driven service-pattern registry so adding new frameworks is a config edit rather than a C patch. Tier 1 (gRPC `.proto` matching) is proposed first as a ~300-LOC working PR to validate the architecture; Tiers 2–4 follow as separate PRs.

Cross-language framework coverage matrix at the end. Acceptance gating: each tier ships independently, success measured by precision/recall against multi-language test fixtures.

---

### 1.1 What works in cbm today

After PR #281 lands, `get_architecture(aspects=["all"])` returns rich structural data (entry_points, routes, hotspots, layers, boundaries, languages). The per-repo extraction pipeline detects:

[...]

PR #293, opened 2026-04-26
feat: Tier 1 cross-repo gRPC matching + production-readiness fixes

sponger94:
## Summary

Lands the **full Tier 1 (gRPC) cross-repo intelligence stack** in one PR, plus the production-readiness fixes that surfaced when running it against a real .NET microservice fleet using NuGet-distributed `.proto` contracts and C# 12 primary constructors. Strictly additive on the matching path; reuses `pass_cross_repo.c match_typed_routes` (Phase D for `GRPC_CALLS`) without changing it.

Closes the design discussion in #292 for Tier 1. Tiers 2–4 (typed message pub/sub, attribute-driven HTTP, config-resolved discovery) remain as separate follow-up PRs.

Companion docs (in this PR under `.planning/`):
- `cbm-cross-repo-proposal.md` — full Tier 1 design, Tier 1g deferral rationale, sequencing.
- `tier1-extractor-fixes.md` — production-readiness gap log.

## What ships

**Producer + consumer detection.** `pass_idl_scan` now extracts producer-side typed gRPC clients from four signal sources — local-var `type_assigns`, constructor parameters, factory return types (Go `pb.NewFooClient`, Java `fooGrpc.newBlockingStub`), DI registrations (`services.AddGrpcClient<T>`, `@GrpcClient`, NestJS `@Client`), and class fields/properties. Consumer-side handler binding (Python `*Servicer`, Java `*ImplBase`, C# `*ServiceBase`) plus IDL Route emission from `.proto` files were already in place; this PR completes the producer half so the existing Phase D matcher emits real `CROSS_GRPC_CALLS` end-to-end.

False-positive guards: type-name denylist for `System.Net.*`, `Microsoft.Extensions.Http`, `RestSharp`, `Refit`, `Flurl`, `java.net.http`, `okhttp3`, `reqwest`, `urllib`, `httpx`; method-name normalization (`*Async` strip, lowerCamelCase → PascalCase); var-scope lookup with explicit class/file/function priority and a fail-closed flag for project-wide arrays.

**Production-readiness fixes (`tier1-extractor-fixes.md` Gaps 1–7).** Short version:

1. `pass_idl_scan` was registered only on the sequential pipeline; wired into parallel + both incremental paths so repos crossing the ~50-file parallel threshold no longer silently skip it.
2. C# 12 primary-constructor params now emit `Field` defs (modern .NET 8+/9+ controllers were invisible to the field walker before).
3. tree-sitter-protobuf emits rpc Functions as flat siblings of the service Class, not children — added a line-range fallback so proto Routes are emitted regardless of `DEFINES_METHOD` shape.
4. `/api/layout` already returned `linked_projects`, but the graph-UI tab silently dropped them when filtering — passes them through now and renders satellite galaxies with cross-repo edges.
5. Incremental indexing wasn't attaching `ctx->result_cache` before `pass_idl_scan`, so producer edges never refreshed on edits — fixed in both incremental paths.
6. Project-wide stub-var lookup had a name-only fallback that could bind a `_client` call to an unrelated class — now fail-closed for project-wide arrays, name-only fallback retained only for per-file arrays.
7. Cross-package gRPC route-key collisions: keyed routes still use bare `<service>/<method>` since cross-repo matching joins on that key and the consumer side has no proto-package source. Added a `idl_scan.route_collision` warning when an existing Route's `file_path` differs, and a `service_qn` Route property carrying the proto Class qualified_name so a future FQN-aware matcher can recover provenance. The full FQN data model is **deferred** to a focused follow-on PR (Tier 1g) — see `cbm-cross-repo-proposal.md` §5.7. Half-shipping it without consumer-side derivation produces dormant nodes that don't change matching behavior.

**Graph UI satellite rendering.** `/api/layout` now computes a layout per linked project, places it on a circle around the primary cluster sized so satellites don't overlap, and populates `cross_edges`. The Three.js scene renders linked projects as offset `NodeCloud` + `EdgeLines` groups with inter-galaxy edges through a new `EdgeLines targetNodes` prop.

## Files changed

| File / Group | Lines | Purpose |
|---|---|---|
| `src/pipeline/pass_idl_scan.c` | +747 / -72 | producer detection + proto rpc fallback + safer stub-var lookup + collision mitigation |
| `src/pipeline/pipeline.c` | +11 | parallel-path wiring |
| `src/pipeline/pipeline_incremental.c` | +26 | incremental wiring |
| `internal/cbm/extract_defs.c` | +85 | C# 12 primary-ctor |
| `internal/cbm/extract_type_assigns.c` | +20 | factory pattern |
| `internal/cbm/lang_specs.c` | +1 / -1 | C# `field_types` (property + field) |
| `Makefile.cbm` | +14 | `dump-csharp` target |
| `tests/dump_csharp.c` | +134 | extractor inspection helper |
| `graph-ui/src/components/{EdgeLines,GraphScene,GraphTab}.tsx` | +166 / -22 | cross-galaxy rendering |
| `src/ui/http_server.c` | +123 / -5 | `/api/layout` linked-project response |
| `.planning/{cbm-cross-repo-proposal,tier1-extractor-fixes}.md` | new | design + production-readiness notes |

## Test coverage

10 unit tests in `tests/test_pipeline.c` cover the IDL Route + consumer-side HANDLES paths. Producer-side detection and the production-readiness fixes are validated end-to-end against a real .NET microservice fleet (NuGet-distributed contracts, C# 12 primary ctors, mixed proto-owner / consumer / mixed shapes). Full suite: 2609 passed, 38 skipped, 0 failed (clang ASan+UBSan build).

## Companion issues

- Design: #292
- Pre-existing pass_parallel issue: #294 — phantom Routes from non-gRPC var names in `pass_parallel.c emit_grpc_edge`. Out of scope for this PR; the new pass uses a separate Route QN namespace (`__route__grpc__`) so it doesn't affect new work, but the pollution is still there.

## Test plan

- [x] 10 unit tests in `tests/test_pipeline.c` — all pass on local clang ASan+UBSan build.
- [x] Full pipeline suite — no regressions.
- [x] Real-world fleet test — multi-service .NET monorepo with NuGet-distributed contracts; producer detection fires across consumer projects, cross-repo edges emit correctly.
- [x] Pass runs as a no-op on repos without `.proto` files and without typed-stub assignments.
- [x] Denylist verified — `HttpClient.GetAsync` does not produce `GRPC_CALLS` or stray Routes.
- [x] Graph UI cross-galaxy rendering — viewer shows linked projects as offset satellites with cross-repo edges.
- [x] Incremental reindex on a touched controller refreshes producer-side `GRPC_CALLS` (was previously stale until full reindex).
- [ ] Reviewer to verify Linux/macOS GCC build cleanliness once CI runs.


DeusData:
Thanks for the substantial work on this, @sponger94 — the architectural framing in `cbm-cross-repo-proposal.md` is sharp, and a lot of the extractor diagnosis matches gaps I'd been tracking.

Closing as **superseded by existing infrastructure** rather than merging the full PR. The cross-repo gRPC matching path you targeted with `pass_idl_scan` is already covered by `pass_cross_repo.c match_typed_routes`: Phase D handles `CROSS_GRPC_CALLS` / `CROSS_GRAPHQL_CALLS` / `CROSS_TRPC_CALLS` against any Route the pipeline emits, regardless of which pass produced it. Adding a parallel `__route__grpc__` namespace to dodge the pre-existing `pass_parallel.c emit_grpc_edge` pollution issue (#294) would have been a workaround layered on top of an unfixed bug rather than a clean addition. The right path forward is to fix #294 first, then route gRPC Routes through the unified namespace.

That said, several pieces of this PR were genuinely valuable bug fixes and have been **cherry-picked to `main` with you as co-author**:

- **`beef06e` feat(extract): C# field/property + C# 12 primary-constructor support** — the `cs_field_types: empty_types → cs_field_types` fix, the nested `variable_declaration` walker for C# fields, and the C# 12 primary-ctor parameter extraction. These are real bugs in `main`; modern .NET 8+/9+ controllers were largely invisible to the field walker before.
- **`7083955` feat(extract): detect typed-stub factory constructors** — the `pb.NewFooClient` / `fooGrpc.newBlockingStub` factory shape detection in `extract_constructor_type`.
- **`7290c1f` feat(ui): satellite galaxy radius spacing + cross-galaxy edge rendering** — `layout_radius()` + dynamic galaxy spacing replacing the fixed 600 that buried clusters, and the `cross_edges` SQL population that fills the existing `// TODO` placeholder. Frontend `targetNodes` prop on `EdgeLines` and the linked-project filter integration in `GraphTab`.

Things deliberately **not** carried over and why:

- `pass_idl_scan.c` — the producer-side detection has a `Base` suffix entry in `k_grpc_client_suffixes` that produces stray `GRPC_CALLS` for any C# `*Base` field (`ControllerBase`, `RepositoryBase`, `ViewModelBase`, …); the JSON property emission embeds proto-derived strings without `cbm_json_escape`, breaking graph state on a `"` or `\` in proto identifiers; and the route-collision warning is documented as a "Gap 7 mitigation" but doesn't actually prevent the second upsert from overwriting properties non-deterministically in the parallel path. The matching architecture also overlaps with `pass_cross_repo.c`.
- `.planning/` docs — the design rationale reads as a proposal *about* `pass_idl_scan`; without that implementation, landing the docs standalone would be confusing.
- `dump_csharp` debug tool — useful but adds Makefile complexity for a dev-only standalone binary, and there's no established pattern for this kind of tool in the repo yet.
- testdata/cross-repo/grpc fixtures — only meaningful as inputs for the unshipped pass.

If you want to follow up: the type-flow framing in your proposal is right and the gaps you identified for typed-client producer detection are real. A focused follow-up that (a) fixes #294 first, (b) addresses the JSON-escape and `Base`-suffix issues, (c) drops the parallel namespace and routes through the existing Route emission would be very mergeable.

Thanks again — the C# extractor fixes alone are a meaningful improvement to `main`.

PR #295, opened 2026-04-28
Cross-project HTTP edges + unified storage + paginated cross_project_links

Shidfar:
## Summary

Adds HTTP cross-project endpoint registration and matching, completing the cross-service protocol linker set (15 protocols total: GraphQL, gRPC, Kafka, Pub/Sub, SQS, SNS, WebSocket, SSE, RabbitMQ, MQTT, NATS, Redis Pub/Sub, tRPC, EventBridge, **HTTP**).

Bundled changes (16 commits):

- **HTTP cross-project edges.** 4-signal endpoint registration: S1 URL literal, S2 env-var regex (`process.env.X`, `os.getenv`, `os.Getenv`, `ENV[]`, `System.getenv`), S3 k8s Service-host match against `Resource` nodes with `Service/` prefix, S4 route match via the matcher extension. Buffered candidate handling with ambiguity logging.
- **Storage unification.** Messaging-protocol cross-repo storage migrated from a separate `_crosslinks.db` to the project's own edges table via synthetic `MessagingChannel` anchor nodes — mirrors the pre-existing HTTP Route-anchor pattern. Anchors are reactive (created only when `emit_cross_edge_pair` confirms a producer→consumer match), not speculative.
- **Pagination + summary guard for `cross_project_links` MCP tool.** New params: `limit` (default 100, max 1000), `offset`, `summary_only`. Always emits a summary header (total, by-protocol breakdown, top-10 project pairs). Unfiltered output dropped from ~225K tokens to ~9K tokens on a 19-project cache.
- **`MAX_CANDIDATES` cap scoping fix.** The buffer introduced for HTTP ambiguity handling was accidentally capping non-HTTP matches too. Non-HTTP now emits inline; HTTP keeps the buffer + cap with a `http.candidate_truncated` log on truncation.
- **HTTP S2/S3 signal reachability fix.** `HTTP_CONF_S2 = 0.20 < SL_MIN_CONFIDENCE = 0.25` was dropping all S2-alone endpoints; raised to 0.30. `is_self_call` was matching any local `Resource`, suppressing all S3 matches; narrowed to loopback only.
- **Cross-repo parity for incremental pipeline.** `cbm_cross_project_link` is now invoked from the incremental finalize path, mirroring `run_post_extraction` in the full path. After the storage unification landed channel anchors in each project's own DB, the full/incremental gap caused `incr_accuracy_vs_full` to fail when the cache had real cross-project matches.

## Test plan

- [x] `./scripts/test.sh` passes (3019/3019, ASan + UBSan)
- [x] `cross_project_links` (with `summary_only`) reports preserved totals on a 19-project cache (2,417 cross-links: 2,093 graphql + 324 pubsub)
- [x] `incr_accuracy_vs_full` stable across 5 consecutive runs
- [x] No `MessagingChannel` nodes are created speculatively — only on confirmed producer→consumer match (`find_or_create_channel` is called only from inside `emit_cross_edge_pair`)


DeusData:
Hi @Shidfar — thanks for taking the time on this. The protocol-coverage breadth, the per-protocol linker shape, and the pagination guard for `cross_project_links` are all sound ideas, and a real amount of work clearly went into the test scaffolding.

I can't merge this PR as-is, though, and I want to be transparent about why — both because the reasons are concrete and because anyone else reading along should be able to verify them.

The PR substitutes `DeusData` → `hodizoda` across the project's install / verification / distribution surface — not just in docs, but in places that determine where end users actually get binaries from:

- **`server.json`** — all five MCP-registry binary download URLs (darwin-arm64/amd64, linux-arm64/amd64, windows-amd64) point to releases on `hodizoda/codebase-memory-mcp` instead of this repo.
- **`scripts/setup.sh`** and **`scripts/setup-windows.ps1`** — `REPO=` flipped to `hodizoda/codebase-memory-mcp`, so the one-line install commands pull binaries from the fork.
- **`README.md`** — install one-liners (`curl … | bash`, `irm … | iex`), the MCP install instruction, the clone URL, and the release badge all redirect to `hodizoda/codebase-memory-mcp`. The README also drops the IaC-indexing feature bullet and changes "66 languages" → "64 languages", which is a content delta independent of the URL change.
- **`SECURITY.md`** — the documented `gh attestation verify --repo hodizoda/codebase-memory-mcp` command would verify provenance against the fork's repo, not this one.
- **`scripts/security-allowlist.txt`**, **`scripts/security-strings.sh`**, **`scripts/security-install.sh`** — the post-build URL allowlists are flipped, so the security-audit scripts would *bless* the substituted URLs as expected rather than flag them.
- **`.github/workflows/release.yml`** — the SBOM `documentNamespace` is rewritten to `hodizoda/codebase-memory-mcp`.
- **`docs/index.html`** — landing-page CTAs and footer GitHub links.

I'd like to read these as artifacts of a long-running fork rebase rather than intent, but they're load-bearing changes to the install path: anyone running the README's `curl | bash`, the Windows `irm | iex`, or installing via the MCP registry after a merge would be downloading binaries from a different repo, and the security audits would no longer flag the substitution. That makes the PR unreviewable as a feature change — the supply-chain delta and the feature delta have to be evaluated separately, and right now they're tangled together.

A second item that gave me pause is in `scripts/security-allowlist.txt`: it adds the entry

```
src/cli/cli.c:system:curl download of release binary (update cmd)
```

but `cli.c` is not in this PR's diff, and `main`'s `cli.c` uses `cbm_popen` / `execvp` for shell-free downloads — there is no raw `system()` call there to allowlist. Most likely a leftover from local fork state, but allowlist files are exactly where leftovers cause problems.

If you'd like to land the protocol-linking work, the path forward I can review is:

1. **A first PR that reverts the URL / registry / install / security-infra changes back to `DeusData/codebase-memory-mcp`** — `server.json`, `scripts/setup*.sh`, `scripts/security-*`, `SECURITY.md`, `README.md`, `CONTRIBUTING.md`, `docs/index.html`, and the `documentNamespace` line in `release.yml`. That alone unblocks the rest from review.
2. **Smaller subsequent PRs scoped to one concern each** — the `pass_servicelinks` plumbing, individual protocol linkers in groups (e.g. messaging-bus protocols together, RPC protocols together), the MCP `cross_project_links` pagination + summary guard, and the storage-unification refactor. The current 24,000-line bundle isn't reviewable as a single unit; broken up, most of it likely is.
3. Worth a look first: `pass_cross_repo.c match_typed_routes` (Phase D) already handles `CROSS_GRPC_CALLS` / `CROSS_GRAPHQL_CALLS` / `CROSS_TRPC_CALLS` against any Route the pipeline emits, regardless of which pass produced it, and `cross_project_links` already exists in `mcp.c`. Some of the per-protocol work probably consolidates against that existing infrastructure rather than running parallel to it.

Closing for now. Re-opens are welcome under the structure above — the protocol-coverage idea itself is good.

Shidfar:
Thanks for the thorough review and for being explicit about what was unreviewable — that was exactly the structure I needed. Coming back to it now.

I've reopened the work as a 5-PR stack against current `main`:

- **#376** — `pass_servicelinks` plumbing only (ready to review)
- **#377** (draft) — 14 protocol linkers wired into the dispatch table
- **#378** (draft) — cross-repo pass + community detection + `cross_project_links` MCP tool with the pagination guard
- **#379** (draft) — HTTP cross-project edges (4-signal endpoint registration)
- **#380** (draft) — storage unification + incremental parity + the matching MCP reader migration

A few notes on what changed vs this PR:

- **The URL / install / security-infra substitutions are gone.** I rebased my fork onto fresh `upstream/main` after your comment, which dropped those commits. `server.json`, `scripts/setup*.sh`, `scripts/security-*`, `SECURITY.md`, `README.md`, `CONTRIBUTING.md`, `docs/index.html`, and the `release.yml` SBOM namespace all still point at `DeusData/codebase-memory-mcp` on every branch in the new stack. So the "PR 0 to revert the install changes" you suggested isn't needed — none of the 5 PRs touch those files.
- **The stray `src/cli/cli.c:system:...` allowlist entry is also gone** for the same reason.
- **On `match_typed_routes` overlap (your point 3):** I audited it before reorganizing. Upstream's matcher handles `CROSS_GRPC_CALLS` / `CROSS_GRAPHQL_CALLS` / `CROSS_TRPC_CALLS` against any Route the pipeline emits, but it relies on a Route node already existing — and the per-protocol linkers in #377 are what produce the channel/endpoint anchors that make Routes from RPC-style traffic discoverable in the first place. So they're complementary, not redundant. I noted this in #377's first commit and the linkers do feed `MessagingChannel` anchors that the cross-repo pass in #378 then matches.
- **One real bug surfaced during the split** that I want to flag — c5e2e90's cleanup block reads `ctx.endpoints` after an early-cancel `goto` that skips `ctx`'s declaration; latent UB that ASan caught when I split the bundle. Fix lands in #377 (declare `endpoints` at function top, free that pointer instead). Full writeup in #377's description.

Closing this in favor of the new stack. Happy to thread review feedback on #376 first and un-draft the others as they go.

PR #300, opened 2026-04-30
Fix search_graph name_pattern= performance: regex cache, LIKE pre-filter, cheap count

### LINKED ISSUE #254: search_graph on large datasets with name_pattern= is slow
### Problem

On a large codebase (~200K nodes), every `search_graph` call using `name_pattern=` takes **1.5–2.5s** regardless of how specific the pattern is. An exact name search is as slow as a broad regex:

search_graph(name_pattern="specificFunctionName")          → 1.5s, 70% CPU
search_graph(label="Method", name_pattern=".*Controller.*") → 2.0s, 88% CPU

Scales linearly with node count — confirmed across four projects ranging from 1K to 200K+ nodes. The `query=` (BM25) path is unaffected and fast.

### Root cause — three compounding bugs

1. `sqlite_iregexp` recompiles the regex on every row (`store.c:431`)

```c
  static void sqlite_iregexp(sqlite3_context *ctx, int argc, sqlite3_value **argv) {
      cbm_regex_t re;
      int rc = cbm_regcomp(&re, pattern, CBM_REG_EXTENDED | CBM_REG_NOSUB | CBM_REG_ICASE);
      // ...
      rc = cbm_regexec(&re, text, 0, NULL, 0);
      cbm_regfree(&re);   // ← compiled and freed for every row
  }
```

For a Method-scoped search across a large project, the pattern is compiled and freed once per node. The standard SQLite fix is `sqlite3_get_auxdata/sqlite3_set_auxdata` to cache the compiled regex for the lifetime of the statement.

2. The count query runs a full second scan (`store.c:2321`)

```c
snprintf(count_sql, sizeof(count_sql), "SELECT COUNT(*) FROM (%s)", sql);
```

The count query executes before the data query with identical WHERE conditions, so `iregexp()` fires twice for every row — doubling the already-expensive work.
 
3. `cbm_extract_like_hints` is implemented but never called (`store.c:2018, store.h:559`)
  
There is already a function that extracts literal segments from a regex pattern to build LIKE pre-filters:

```c
  int cbm_extract_like_hints(const char *pattern, char **out, int max_out);
```

For `.*Controller.*` it yields `["Controller"]`, enabling: 
`WHERE n.name LIKE '%Controller%' AND iregexp(?1, n.name)`

The `idx_nodes_name (project, name)` index can satisfy the LIKE clause and cut the regex scan to only matching rows. But `cbm_extract_like_hints` is never called from `search_where_basic` or anywhere in the search path — it is dead code.

### Suggested fixes

**Fix 1** — cache the compiled regex in `sqlite_regexp/sqlite_iregexp` using `sqlite3_set_auxdata`:

```c
cbm_regex_t *re = (cbm_regex_t *)sqlite3_get_auxdata(ctx, 0);
if (!re) {
    re = malloc(sizeof(cbm_regex_t));
    cbm_regcomp(re, pattern, CBM_REG_EXTENDED | CBM_REG_NOSUB | CBM_REG_ICASE);
    sqlite3_set_auxdata(ctx, 0, re, (void(*)(void*))regex_free_cb);
}
```

**Fix 2** — wire `cbm_extract_like_hints` into `search_where_basic` to prepend LIKE conditions before the regex clause. The function is already correct; it just needs to be called.

**Fix 3** — merge the count and data query into a single pass (e.g. `COUNT(*) OVER ()` window function) to eliminate the double scan.

**Note on `query=` mode**

`query=` (BM25/FTS5) already takes a completely separate fast path (`handle_search_graph:1318`) and is unaffected. This bug is specific to `name_pattern=` and the `cbm_store_search` code path.

### Environment

- Version: 0.6.0
- OS: Linux (Ubuntu, x86_64)
- 216,648 nodes, 661,477 edges, 509MB database (PHP monolith)

awconstable:
Fixes #254

## Root cause

Three compounding bugs caused `name_pattern=` searches to scan every node with an expensive compiled regex, regardless of how selective the pattern was:

1. `sqlite_iregexp` / `sqlite_regexp` recompiled the regex on every row — `cbm_regcomp` + `cbm_regfree` fired once per node for the full table.
2. The count query wrapped the full SELECT (including two correlated edge-count subqueries per row) in `SELECT COUNT(*) FROM (...)`, doubling the scan with identical per-row overhead.
3. `cbm_extract_like_hints` was implemented and correct but never called — the LIKE pre-filter that should cut the regex scan to only matching rows was dead code.

## Changes

**Fix 1 — regex cached per statement** (`sqlite_regexp` / `sqlite_iregexp`)  
Use `sqlite3_get_auxdata` / `sqlite3_set_auxdata` to cache the compiled `cbm_regex_t` for the lifetime of the statement. `cbm_regcomp` is now called exactly once per query, not once per row.

**Fix 2 — LIKE pre-filter wired in** (`where_add_like_hints`, `search_where_basic`)  
Wire `cbm_extract_like_hints` into `search_where_basic` via a new `where_add_like_hints` helper. For `.*Controller.*` this prepends `n.name LIKE '%Controller%'`; the `idx_nodes_name` index satisfies the LIKE clause and only matching rows reach `iregexp()`. Added `search_like_pool_t` to manage the malloc'd LIKE strings across both statement executions. `ST_SEARCH_MAX_BINDS` raised 16 → 32.

**Fix 3 — count query stripped of per-row edge subqueries**  
For the common no-degree-filter path, the count SQL is now `SELECT COUNT(*) FROM nodes n WHERE <same WHERE>` — no correlated `edges` subqueries. The degree-filter path retains the wrapped form since it needs those columns for the filter.

## Benchmark

Tested on a large PHP codebase (~200K nodes):

| Query | Before | After | Speedup |
|---|---|---|---|
| `name_pattern=.*Controller.*` | 3099ms | 508ms | **6×** |
| `name_pattern=.*Service.*` | 2006ms | 506ms | **4×** |
| `name_pattern=.*Repository.*` | 2006ms | 508ms | **4×** |
| `name_pattern=specificFunctionName` | 1506ms | 507ms | **3×** |
| `label=Method` + `name_pattern=.*get.*` | 8509ms | 509ms | **17×** |

The ~500ms floor is cold-start I/O when spawning a fresh process against a ~500MB database. In the long-running MCP server (warm file cache) the query time is sub-millisecond.

A reusable benchmark script is included at `scripts/benchmark-search-graph.sh`.

## Tests

All store search tests pass including `store_search_pagination` (offset-past-end total count), `store_search_degree_filter`, and the full `store_extract_like_hints` suite.

### DIFF
--- a/scripts/benchmark-search-graph.sh
+++ b/scripts/benchmark-search-graph.sh
@@ -1,33 +1,32 @@
 #!/bin/bash
-# benchmark-search-graph.sh — Time search_graph name_pattern= queries against the
-# running codebase-memory-mcp binary, which must already have the SIS project indexed.
+# benchmark-search-graph.sh — Time search_graph name_pattern= queries against a
+# codebase-memory-mcp binary to measure the regex / LIKE pre-filter performance.
 #
 # Usage:
-#   scripts/benchmark-search-graph.sh [binary-path]
+#   scripts/benchmark-search-graph.sh <binary-path> <project-name>
 #
-# Defaults to /usr/local/bin/codebase-memory-mcp.
+# Example:
+#   scripts/benchmark-search-graph.sh ./build/c/codebase-memory-mcp my-project
 
 set -euo pipefail
 
-BINARY="${1:-/usr/local/bin/codebase-memory-mcp}"
-PROJECT="home-ubuntu-dev-sis"
+BINARY="${1:?Usage: $0 <binary-path> <project-name>}"
+PROJECT="${2:?Usage: $0 <binary-path> <project-name>}"
 
-echo "Binary: $BINARY"
+echo "Binary:  $BINARY"
 echo "Project: $PROJECT"
 echo ""
 
-# Each benchmark case: NAME and JSON request body
 run_case() {
     local label="$1"
     local request="$2"
     local start end elapsed_ms result
 
     start=$(date +%s%3N)
-    result=$( echo "$request" | "$BINARY" 2>/dev/null || true )
+    result=$(echo "$request" | "$BINARY" 2>/dev/null || true)
     end=$(date +%s%3N)
     elapsed_ms=$(( end - start ))
 
-    # Extract match count from JSON result (nodes array length) or total field
     local count
     count=$(echo "$result" | python3 -c "
 import sys, json
@@ -36,14 +35,13 @@ try:
     content = d.get('result', {}).get('content', [{}])[0].get('text', '{}')
     obj = json.loads(content)
     print(obj.get('total', obj.get('count', '?')))
-except Exception as e:
+except Exception:
     print('?')
 " 2>/dev/null || echo "?")
 
     printf "  %-55s %5dms  (total=%s)\n" "$label" "$elapsed_ms" "$count"
 }
 
-# Build a minimal JSON-RPC call for search_graph
 sg() {
     local project="$1"
     local args="$2"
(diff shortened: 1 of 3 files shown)


DeusData:
Merged via rebase, thanks @awconstable — diagnoses are spot-on, fixes are clean, benchmarks reproduce. The auxdata caching is the canonical SQLite pattern, the LIKE pre-filter wiring is well-scoped (`search_like_pool_t` correctly handles the SQLITE_STATIC bind lifetime across both count and main statements), and the count-query split between the no-degree-filter and degree-filter paths is exactly right.

A note for anyone reading this thread later: the branch also contained the FTS5 two-step subquery fix that #302 was targeting, so #302 is now superseded — closing it as resolved.

Soft behavior note worth flagging on the FTS5 path: `BM25_INNER_LIMIT = 2000` caps the inner candidate set, so the `total` reported to callers is now bounded by 2000 (or fewer post-filter). That makes pagination beyond offset 2000 silently saturate. Practically fine for ranked search — page 100 of search results was never going to be useful — but if anyone hits it later, the constant is the place to lift.

Issue #187, opened 2026-03-31
Direct SQLite writer crashes on oversized cells during indexing (repro with netty on macOS arm64)


Yegezw:
## Summary

Indexing a large Java repo can crash `codebase-memory-mcp` in the direct SQLite page writer path when a single record exceeds what fits in one leaf page. On macOS arm64 this shows up as a late crash in `free()` / `_szone_free`, but AddressSanitizer points to an earlier stack-buffer-underflow in `internal/cbm/sqlite_writer.c`.

I reproduced this reliably while indexing the Netty repository.

## Environment

- `codebase-memory-mcp` version: `0.5.7`
- OS: macOS 15.6.1
- Arch: arm64
- MCP client: Codex CLI / CLI mode
- Repro repo: `https://github.com/netty/netty`

## Reproduction

1. Clone `netty`
2. Run:

```bash
codebase-memory-mcp cli index_repository '{"repo_path":"/path/to/netty"}'
```

## Actual behavior

The process crashes near the end of indexing.

In normal runs the last logs are around:

- `pass.timing pass=configlink ...`
- then the process exits/crashes before indexing completes

macOS crash reports show an `EXC_BREAKPOINT` in `_szone_free`, which makes it look like a late free bug.

With ASan/UBSan, the root problem is clearer:

```text
base version: git checkout 148d951e1a51bcc5f11f6cee97be2aa86141623d

internal/cbm/sqlite_writer.c:435:21: runtime error: index ... out of bounds for type 'uint8_t[65536]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior internal/cbm/sqlite_writer.c:435:21

ERROR: AddressSanitizer: stack-buffer-underflow
WRITE of size 508794
    #1 pb_add_cell sqlite_writer.c:435
    #2 pb_add_table_cell_with_flush sqlite_writer.c:865
    #3 cbm_write_db sqlite_writer.c:1301
    #4 cbm_gbuf_dump_to_sqlite graph_buffer.c:1237
```

## Root cause

`pb_add_table_cell_with_flush()` only flushes when the current page is non-empty and the next cell does not fit.

That misses the case where:

- the current page is empty, and
- the new cell itself is larger than the maximum capacity of a single leaf page

In that situation `pb_add_cell()` still runs and does:

- `pb->content_offset -= cell_len`
- `memcpy(pb->page + pb->content_offset, cell, cell_len)`

which underflows the page buffer and corrupts memory.

[...]

PR #175, opened 2026-03-29
fix: implement overflow pages in sqlite_writer to prevent SIGBUS on large records (#139)

dLo999:
Closes #139 (reopened v0.5.7 crash)

## Summary

Implements SQLite overflow page support for table leaf cells in the custom B-tree writer (`sqlite_writer.c`). This fixes the SIGBUS crash when indexing large repositories that produce records exceeding 65KB (e.g., nodes with very long `properties` JSON or `qualified_name`).

## Root cause

`pb_add_table_cell_with_flush` called `pb_add_cell` unconditionally after flushing the current page. If a single record's payload exceeded the usable page space (~65,501 bytes), `content_offset -= cell_len` underflowed to negative, causing `memcpy` to write before the page buffer — stack corruption leading to SIGBUS.

The reporter's crash profile matches: Java monorepo with 13,660 files producing 226K nodes + 477K edges, SIGBUS immediately after `pass.timing pass=configlink` during the `gbuf.dump` / SQLite write step.

## Fix

Implemented SQLite's overflow page mechanism for table leaf cells:

1. **`write_overflow_pages()`** — writes a linked-list of overflow pages for payload bytes that exceed local storage. Each page has a 4-byte next-page pointer followed by payload data.

2. **`build_table_cell_overflow()`** — builds a table leaf cell in the overflow format: `varint(total_payload_len) + varint(rowid) + payload[0..local_len) + uint32(overflow_page_num)`. The total payload length varint still encodes the full original size so SQLite knows how much to read.

3. **`pb_add_table_cell_with_flush()`** — now detects `payload_len > TABLE_OVERFLOW_MAX_LOCAL (65,501)`, computes `local_len` using SQLite's exact integer formula, writes overflow pages, then builds the reduced cell. Normal path (no overflow) is unchanged.

## Key constants (PAGE_SIZE=65536, reserved=0)

| Constant | Value | Formula |
|----------|-------|---------|
| `TABLE_OVERFLOW_MAX_LOCAL` | 65,501 | `usable_size - 35` |
| `TABLE_OVERFLOW_MIN_LOCAL` | 8,199 | `(usable_size - 12) * 32 / 255 - 23` |
| Overflow page data capacity | 65,532 | `usable_size - 4` (4-byte next pointer) |

## Test results

**Build:** Compiles cleanly on macOS (Apple Clang, arm64)

**Test suite:** 2742 passed, 0 failed (was 2741 — new test added)

**New test: `sw_oversized_node`**
[...]

Issue #139, opened 2026-03-24
Stack overflow crash in autoindex_thread → cbm_pipeline_pass_configlink (v0.5.6, macOS ARM64)


theron-sapp:
## Summary

`codebase-memory-mcp` v0.5.6 crashes deterministically every ~3 minutes due to a stack overflow in the auto-index background thread. The crash is 100% reproducible — 11 identical crashes in a single day.

## Environment

- **Version**: 0.5.6
- **OS**: macOS 26.3.1 (25D2128), ARM64 (Mac15,6, 18GB RAM)
- **Agent**: Claude Code (VS Code extension)
- **Repos indexed**: 2 projects (~665 + ~400 tracked files, 22MB total DB)

## Crash details

Every crash report is identical:

- **Exception**: `EXC_BAD_ACCESS` — **Thread stack size exceeded** (`SIGBUS`)
- **Faulting thread**: Thread 2 (`autoindex_thread`)
- **Stack**: `___chkstk_darwin` → `cbm_pipeline_pass_configlink` → `cbm_pipeline_run` → `autoindex_thread`
- **Thread stack**: 544KB (`0x16b818000-0x16b8a0000`), guard page hit at `0x16b817ff8`

The server starts, runs for ~3 minutes while auto-indexing, then the `cbm_pipeline_pass_configlink` function recurses deep enough to blow the 544KB thread stack and crashes with SIGBUS.

## Crash frequency (all from 2026-03-24)

| Crash time | Alive duration |
|---|---|
| 12:11:57 | ~3 min |
| 13:35:28 | ~3 min |
| 13:46:38 | ~3 min |
| 14:02:39 | ~3 min |
| 14:06:07 | ~3.5 min |
| 14:11:31 | ~3 min |
| 14:14:53 | ~3 min |
| 14:37:53 | ~3.5 min |
| 14:38:19 | ~3.5 min |
| 15:04:05 | ~4 min |
| 15:40:43 | ~3 min |

## Likely fix

Either:
1. Increase the stack size for the `autoindex_thread` (e.g., 8MB via `pthread_attr_setstacksize`)
2. Convert `cbm_pipeline_pass_configlink` from recursive to iterative traversal

This may be related to the O(N²) traversal pattern in #130 — deep/wide AST structures could compound both the performance and stack depth issues.

## Workaround

Disabling auto-indexing prevents the crash:
```
codebase-memory-mcp config set auto_index false
```

[...]

PR #303, opened 2026-04-30
fix(sqlite_writer): skip index cells that exceed a single page

jjoos:
## Problem

`write_index_btree()` crashes with SIGBUS when indexing large repositories (observed on a Ruby codebase with ~215K nodes, specifically the HackerOne `core` monolith with ~67K functions).

The crash occurs in the page-buffer writer when an index cell's payload (qualified_name + file_path + properties) exceeds the SQLite page size. The code flushes the current page and assumes the fresh page will always accept the cell — but oversized cells still fail `pb_cell_fits()` after the flush. The subsequent `pb_add_cell()` causes a `content_offset` underflow, corrupting the page header and triggering SIGBUS on the next write.

## Reproduction

Index any repository with functions whose qualified names + file paths exceed ~4000 bytes combined (the SQLite page size minus overhead). A Ruby monolith with deeply nested namespaces and long file paths reliably triggers this.

```bash
codebase-memory-mcp index --path /path/to/large-ruby-repo
# → SIGBUS in write_index_btree at sqlite_writer.c:1218
```

## Fix

After `pb_promote_and_flush()`, re-check `pb_cell_fits()`. If the cell still doesn't fit on an empty page, `continue` past it. Index entries whose keys exceed a full page cannot be stored in a leaf and are silently dropped — the rest of the index is correctly preserved.

This is the index-btree counterpart to PR #175 (record overflow pages), which fixed the same shape of bug for the record-btree writer. The record side uses overflow pages; the index side has no overflow mechanism, so skipping is the correct behavior.

## Changes

- `internal/cbm/sqlite_writer.c`: 10 lines added, 3 removed in `write_index_btree()`.

## Testing

- Verified by indexing a 215K-node Ruby repo that previously crashed — completes successfully with all other nodes indexed.
- The skipped cells are edge cases (extremely long qualified names) that don't affect query correctness for normal usage.

## Related

- Closes the index-cell side of #139 / #187 (record-cell side was fixed in #175).
- PR #217 (growable AST traversal stacks) addresses a different SIGBUS trigger in the parser.

### DIFF
--- a/internal/cbm/sqlite_writer.c
+++ b/internal/cbm/sqlite_writer.c
@@ -1213,9 +1213,16 @@ static uint32_t write_index_btree(FILE *fp, uint32_t *next_page, uint8_t **cells
     pb_init(&pb, fp, *next_page, true);
 
     for (int i = 0; i < count; i++) {
-        if (!pb_cell_fits(&pb, cell_lens[i]) && pb.cell_count > 0) {
-            if (!pb_promote_and_flush(&pb, cells, cell_lens, i - SKIP_ONE)) {
-                return 0;
+        if (!pb_cell_fits(&pb, cell_lens[i])) {
+            if (pb.cell_count > 0) {
+                if (!pb_promote_and_flush(&pb, cells, cell_lens, i - SKIP_ONE)) {
+                    return 0;
+                }
+            }
+            // After flush, check if the cell still doesn't fit on an empty page.
+            // Index cells larger than a full page can never be stored; skip them.
+            if (!pb_cell_fits(&pb, cell_lens[i])) {
+                continue;
             }
         }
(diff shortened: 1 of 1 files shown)


DeusData:
Merged via rebase, thanks @jjoos. The diagnosis is dead on — the asymmetry between the record-btree path (overflow pages, fixed in #175) and the index-btree path (no overflow mechanism) is exactly why the same shape of guard produced two different bug surfaces, and the fix preserves SQLite's invariants without inventing a new overflow scheme.

Following up on `main` with a small enhancement: a `cbm_log_warn("sqlite_writer.index_cell_oversize", ...)` on the skip path so future occurrences (extra-deep namespaces, very long paths) are observable rather than silent. Doesn't change the fix's behavior — just makes the dropped cells visible in logs for diagnosis.

PR #310, opened 2026-05-01
feat(mcp): add Windows support for code search via PowerShell

### LINKED ISSUE #196: cli `search_code` returns "search failed: temp file" on Windows 11 in Powershell
When running `search_code` via CLI, I am getting a temp file error

```
PS drive:/path/to/project> codebase-memory-mcp cli search_code '{\"project\": \"my-project-name\", \"pattern\": \"my_function\"}'
level=info msg=mem.init budget_mb=65419 total_ram_mb=130839
{"content":[{"type":"text","text":"search failed: temp file"}],"isError":true}
```

I was able to confirm the actual function I was looking for, was listed as a `Function` in the UI tool at `http://127.0.0.1:9749/`

Codebase Memory Version: `0.5.7`

OS Info:

|Name|Value|
|-----|-----|
|Edition	|Windows 11 Enterprise|
|Version	|23H2|
|OS build	|22631.6783|

Powershell Info:

| Name | Value |
|-----|-----|
| PSVersion |5.1.22621.6345|
PSEdition                    |  Desktop
PSCompatibleVersions |          {1.0, 2.0, 3.0, 4.0...}
BuildVersion                  | 10.0.22621.6345
CLRVersion                    | 4.0.30319.42000
WSManStackVersion      |        3.0
PSRemotingProtocolVersion |     2.3
SerializationVersion          | 1.1.0.1

Note: I am able to replicate this using `git-bash` as well.

noctrex:
Replace `grep` with PowerShell `Select-String` on Windows for the search_code and search_graph tools, enabling cross-platform functionality.

- Use tab-delimited output from `Select-String` to avoid conflicts with colons in Windows drive-letter paths (e.g. `C:\dir\file`).
- Update `collect_grep_matches` to parse delimiter as tab on Windows and colon on POSIX.
- Open scoped filelist in binary mode (`"wb"`) and use forward slashes on Windows to prevent `xargs` from interpreting backslashes as escape sequences.
- Use `cbm_tmpdir()` for pattern temp files instead of hardcoded `/tmp/`.
- Define `getpid` as `_getpid` on Windows for consistent API.

Fixes #196


arkiaman:
see #326 for additional context

DeusData:
Merged via rebase, thanks @noctrex. The PowerShell `Select-String` swap is well-scoped — `#ifdef _WIN32` keeps the POSIX path untouched, the tab-vs-colon delimiter switch is the right move (Windows paths legitimately contain `:` after the drive letter), and pivoting to `cbm_tmpdir()` for the pattern temp file is a solid cleanup that drops the hardcoded `/tmp/`. The forward-slash filelist + binary-mode write closes the xargs-backslash-as-escape gotcha cleanly. Issue #196 should be resolved.

One follow-up landed on `main` directly (`eca433b`): `cbm_validate_shell_arg` now also blocks `"`, `<`, `>`. The PR's PowerShell command is wrapped at the cmd.exe level (`powershell -Command "..."`), so a `"` in `file_pattern` could close that outer quote even when the inner PowerShell single quotes hold; with `< >` unblocked, that opened a cmd.exe redirection primitive. Existing blocking of `' ; | & $ \`` already prevents RCE, so this is hardening rather than a fix for an exploitable RCE in this PR — but worth tightening since the new code path made the validator load-bearing in a new shell context. Three tests added in `tests/test_security.c`.

Nice contribution.

noctrex:
@DeusData Thank you for creating a legitimately useful tool that I actually use as the # 1 tool everywhere, and with this patch, I can use it across all my systems, both win & linux.

noctrex:
My ADHD programmer brain did the classic thing of trying to overcomplicate things, and I initially used the same colon separator and created a whole function that would reverse search the string in order to find the last colon, but after a relaxing coffee and bougatsa, it dawned on me like a revelation from Zeus himself and just use different separators.
I have to repeat to myself, KISS.
And maybe even the powershell is over-engineered and can have a simpler solution, but it hasn't come to me yet.

PR #311, opened 2026-05-01
fix(store.c): improve root path integrity check with range-based validation

noctrex:
Replace the `NOT IN` list of individual characters with a range-based `BETWEEN 'A' AND 'Z'` check for validating project root paths.
The previous query only excluded letters A–H, leaving paths starting with I–Z silently invalid.
In Windows, all the alphabet can be used.
A Windows drive starting with letters I-Z breaks the saving of the database.


DeusData:
Merged via rebase, thanks @noctrex. Good catch — the original `'A','B','C','D','E','F','G','H'` enumeration was clearly intended to cover Windows drive letters but stopped at H, so any project rooted on `I:` through `Z:` was getting flagged as integrity-violating. The `BETWEEN 'A' AND 'Z'` rewrite is the right shape for the check.

One follow-up worth knowing about (not blocking, won't change here): the integrity check still only accepts uppercase drive letters. If `cbm_project_name_from_path` or any caller upstream ever stores a path like `c:\foo` (lowercase), the check would re-flag it. Most Windows tooling normalizes drive letters to uppercase so this is unlikely to bite, but if it does, extending to `BETWEEN 'A' AND 'z'` (or two ranges) is the obvious next step. Not in scope for this PR.

Closes the same diagnosis from #309. Nice contribution.

noctrex:
> Merged via rebase, thanks @noctrex. Good catch — the original `'A','B','C','D','E','F','G','H'` enumeration was clearly intended to cover Windows drive letters but stopped at H, so any project rooted on `I:` through `Z:` was getting flagged as integrity-violating. The `BETWEEN 'A' AND 'Z'` rewrite is the right shape for the check.
> 
> One follow-up worth knowing about (not blocking, won't change here): the integrity check still only accepts uppercase drive letters. If `cbm_project_name_from_path` or any caller upstream ever stores a path like `c:\foo` (lowercase), the check would re-flag it. Most Windows tooling normalizes drive letters to uppercase so this is unlikely to bite, but if it does, extending to `BETWEEN 'A' AND 'z'` (or two ranges) is the obvious next step. Not in scope for this PR.
> 
> Closes the same diagnosis from #309. Nice contribution.

I only catched this because I have a drive called R:, and it was failing for this specific drive :)
It won't be a problem because on Windows the path & file names are case-insensitive by default.

PR #313, opened 2026-05-02
Add support for NetBSD, FreeBSD, and OpenBSD.

### LINKED ISSUE #290: Does not build on NetBSD (or any other BSD)
In `src/foundation/system_info.c`, Linux support is really Linux specific (`sysinfo`), so it doesn't compile on NetBSD. NetBSD should use `sysctl` (with `HW_PHYSMEM64`).

cmeerw:
Fixes #290

### DIFF
--- a/src/foundation/system_info.c
+++ b/src/foundation/system_info.c
@@ -2,6 +2,7 @@
  * system_info.c — CPU core count and RAM detection.
  *
  * macOS: sysctlbyname for core counts, hw.memsize for RAM.
+ * BSD: sysconf + sysctl(HW_PHYSMEM64 / HW_PHYSMEM).
  * Linux: sysconf + sysinfo().
  * Windows: GetSystemInfo + GlobalMemoryStatusEx.
  *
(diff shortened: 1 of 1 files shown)


DeusData:
Merged via rebase, thanks @cmeerw. The `#elif` restructure is exactly the right shape — pulling BSDs out before the implicit Linux fallback fixes the root cause (the previous `#elif !defined(_WIN32) /* Linux */` was branch-labeled "Linux" but actually included every non-Windows non-Apple system, which is why `sysinfo()` was being pulled in on NetBSD).

The OpenBSD-vs-FreeBSD/NetBSD split on `HW_PHYSMEM` vs `HW_PHYSMEM64` is the right call — OpenBSD never grew the 64-bit variant, and on modern 64-bit OpenBSD `HW_PHYSMEM` returns the full value into our `uint64_t` anyway. Issue #290 auto-closed with the merge.

Treating all BSD cores as `perf_cores` matches the Linux path (BSDs don't expose a perf/efficiency distinction) — also correct.

PR #315, opened 2026-05-02
feat(store): expose mmap_size via CBM_SQLITE_MMAP_SIZE env

### LINKED ISSUE #314: SIGBUS on macOS arm64 when 2+ sessions share the same cache dir
## Summary

When two or more Claude Code sessions point at the same project (same cache dir), `codebase-memory-mcp` crashes with `SIGBUS` on macOS arm64. The crash is 100% reproducible when opening 2+ windows on the same project.

## Crash signature

```
EXC_BAD_ACCESS (SIGBUS)
Exception Codes: 0x000000000000000a, 0x0000000106038000
...
Termination Reason: Namespace SIGNAL, Code 10 Bus error: 10
...
FS pagein error: 22 Invalid argument  cluster_pagein past EOF
...
Thread 0 Crashed:
  sqlite3BtreeNext + ...
  sqlite3VdbeSorterNext + ...
```

The kernel triage is always `cluster_pagein past EOF` — the mmap'd SQLite region is being accessed after the file has been truncated by another process.

## Root cause (two independent layers)

**Layer 1 — mmap + truncation:**
`configure_pragmas()` hard-codes `PRAGMA mmap_size = 67108864` on every connection (`src/store/store.c:326`). When process A's checkpoint or reindex truncates the DB while process B has it mmap'd, the next page fault in B's mmap region raises `SIGBUS` — an unrecoverable signal. If mmap were disabled (`PRAGMA mmap_size = 0`), the same truncation would produce recoverable `SQLITE_IOERR` instead.

**Layer 2 — checkpoint mode:**
`cbm_store_checkpoint()` uses `SQLITE_CHECKPOINT_TRUNCATE` (`src/store/store.c:793`). TRUNCATE is the most aggressive checkpoint mode — it calls `ftruncate(fd, 0)` on the WAL file on success. The SQLite docs recommend `SQLITE_CHECKPOINT_PASSIVE` for shared databases precisely because TRUNCATE can shrink files under concurrent readers.

## Fixes

Two independent, low-risk fixes that each address one layer:

1. **Expose mmap_size via `CBM_SQLITE_MMAP_SIZE` env** — users sharing a cache dir can set `CBM_SQLITE_MMAP_SIZE=0` to opt out of mmap without any behavior change for single-instance users. Default unchanged at 67108864.

2. **Use PASSIVE checkpoint** — switches `cbm_store_checkpoint()` from `SQLITE_CHECKPOINT_TRUNCATE` to `SQLITE_CHECKPOINT_PASSIVE`. PASSIVE never blocks readers and never calls `ftruncate()`. SQLite still autocheckpoints at 1000-page boundaries; disk reclamation is unaffected for single-process users.

PRs for both fixes are open:
- `fix/mmap-env-knob` — feat(store): expose mmap_size via CBM_SQLITE_MMAP_SIZE env
- `fix/passive-checkpoint` — fix(store): use PASSIVE checkpoint to avoid file-shrink under concurrent readers

## Environment

- macOS 15 Sequoia, Apple Silicon (arm64)
- codebase-memory-mcp v0.6.0
- Claude Code with multiple windows, all using the same `claude mcp add codebase-memory-mcp` config

edwardmhughes:
## Summary

Hard-coded `PRAGMA mmap_size = 67108864` left no path for users running multiple `codebase-memory-mcp` instances against the same cache to opt out of memory-mapped I/O.

On macOS, when one instance's checkpoint or reindex truncates the DB file under another instance's live mmap, accessing the now-missing pages raises `SIGBUS`, which takes down the process. Setting `CBM_SQLITE_MMAP_SIZE=0` reverts to `read()`/`pread()` I/O, which returns recoverable `SQLITE_IOERR` instead.

## Changes

- `src/store/store.c`: Extracts `cbm_store_resolve_mmap_size()` — reads `CBM_SQLITE_MMAP_SIZE`, validates with `strtoll`, clamps negative to 0, falls back to default `67108864` on missing/malformed input
- `src/store/store.h`: Declares `cbm_store_resolve_mmap_size()` for tests
- `tests/test_store_pragmas.c` (new): 7 tests covering default, zero, explicit value, negative clamp, garbage input, partial garbage, and store-open smoke test
- `Makefile.cbm` + `tests/test_main.c`: Wire new test suite

## Behavior

| Scenario | Before | After |
|---|---|---|
| `CBM_SQLITE_MMAP_SIZE` unset | `PRAGMA mmap_size = 67108864` | unchanged |
| `CBM_SQLITE_MMAP_SIZE=0` | n/a | `PRAGMA mmap_size = 0` (disables mmap) |
| `CBM_SQLITE_MMAP_SIZE=1048576` | n/a | `PRAGMA mmap_size = 1048576` |
| Invalid value | n/a | logs warning, falls back to 67108864 |

Single-instance users: **no behavior change**.

Fixes #314

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -279,7 +279,8 @@ TEST_STORE_SRCS = \
     tests/test_store_edges.c \
     tests/test_store_search.c \
     tests/test_store_arch.c \
-    tests/test_store_bulk.c
+    tests/test_store_bulk.c \
+    tests/test_store_pragmas.c
 
 TEST_CYPHER_SRCS = \
     tests/test_cypher.c
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -297,6 +297,24 @@ static int create_user_indexes(cbm_store_t *s) {
     return exec_sql(s, sql);
 }
 
+int64_t cbm_store_resolve_mmap_size(void) {
+    enum { MMAP_DEFAULT = 67108864, BASE_10 = 10 }; /* default 64 MB; decimal radix */
+    char buf[ST_BUF_64];
+    if (cbm_safe_getenv("CBM_SQLITE_MMAP_SIZE", buf, sizeof(buf), NULL) == NULL) {
+        return (int64_t)MMAP_DEFAULT;
+    }
+    char *end = NULL;
+    long long parsed = strtoll(buf, &end, BASE_10);
+    if (end == buf || *end != '\0') {
+        /* Malformed — fall back to default rather than fail the store open. */
+        return (int64_t)MMAP_DEFAULT;
+    }
+    if (parsed < 0) {
+        return 0;
+    }
+    return (int64_t)parsed;
+}
+
 static int configure_pragmas(cbm_store_t *s, bool in_memory) {
     int rc;
     rc = exec_sql(s, "PRAGMA foreign_keys = ON;");
@@ -323,7 +341,10 @@ static int configure_pragmas(cbm_store_t *s, bool in_memory) {
         if (rc != CBM_STORE_OK) {
             return rc;
         }
-        rc = exec_sql(s, "PRAGMA mmap_size = 67108864;"); /* CBM_SZ_64 MB */
+        char mmap_sql[ST_BUF_64];
+        snprintf(mmap_sql, sizeof(mmap_sql), "PRAGMA mmap_size = %lld;",
+                 (long long)cbm_store_resolve_mmap_size());
+        rc = exec_sql(s, mmap_sql);
     }
     return rc;
 }
--- a/src/store/store.h
+++ b/src/store/store.h
@@ -250,6 +250,15 @@ int cbm_store_create_indexes(cbm_store_t *s);
 /* Force WAL checkpoint + PRAGMA optimize. */
 int cbm_store_checkpoint(cbm_store_t *s);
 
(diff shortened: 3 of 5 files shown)


DeusData:
Merged via rebase, thanks @edwardmhughes. Diagnosis is precise — the macOS SIGBUS-on-truncated-mmap interaction is exactly the kind of failure mode that needs a knob rather than a code fix, since the safe-by-default value (`0`) trades off ~real performance for crash safety, and most users on a single instance shouldn't pay that cost.

The implementation is solid:
- `cbm_safe_getenv` (project's wrapper rather than raw `getenv`) is the right call.
- The `strtoll` validation pattern (`end == buf` for no-digits AND `*end != '\0'` for trailing garbage) catches both empty and partial-numeric inputs — the test cases for `not-a-number` and `123abc` confirm both branches.
- Negative→0 clamp is the right shape (SQLite interprets `mmap_size = 0` as disable, so the clamp aligns the env semantics with SQLite's).
- Default unchanged, single-instance users see no behavior change.

Test coverage is comprehensive — 7 cases covering default/zero/explicit/negative/garbage/partial-garbage/integration-smoke. The integration smoke test (opening a file-backed store with `mmap_size=0`) is the right shape for proving the resolver is actually wired into `configure_pragmas` rather than just being a standalone function.

Issue #314 auto-closed with the merge.

Issue #277, opened 2026-04-20
New files not indexed — WAL-checkpoint blocked on successfully-indexed project (Windows, v0.6.0)


davidseraphi:
`v0.6.0` (confirm via `codebase-memory-mcp --version` before filing), Windows 11, stdio MCP registration via project-scope `.mcp.json`.

## Relationship to existing CLOSED issues

Triaged before filing; distinct from all three:

- **#50** (CLOSED, v0.4.8 + v0.5.0 C-rewrite fixed it): duplicate DBs from drive-letter casing. Ours is NOT duplicate DBs — one `.db` file per project, the right one, but its mtime is frozen while its `.db-wal` accumulates writes.
- **#52** (CLOSED): `list_projects` hang + stuck `db-shm` from SIGKILL-mid-transaction. Substrate overlap (WAL+shm stuck states), different manifestation: our server does NOT hang; it responds normally; it just returns stale data for symbols that exist in the WAL but have never been checkpointed into the main `.db`.
- **#119** (CLOSED, PR #120 ~2026-03-23): query handlers silently return empty for **unknown/unindexed** projects + ghost-`.db` creation. Our project IS indexed (`index_repository` returns `status: indexed, nodes: N, edges: M` with N/M > 0 for the project's pre-existing symbols, and `list_projects` lists it), but queries for symbols defined in files added POST-initial-index return empty. Adjacent bug class; #119's guard does not cover this.

If this repro turns out to be a re-emergence of any of the above under a new code path, please re-close as duplicate with a pointer.

## Symptom

`index_repository` returns `status: indexed, nodes: N, edges: M` where (N, M) are the **pre-add-new-file** counts — the indexer reports success but the returned counts do not reflect the files added since the last clean checkpoint. `search_graph`, `search_code`, and `trace_path` all return empty for symbols in files added since. Affects both `delete_project` → `index_repository` rebuild AND incremental re-index cycles; i.e., even a full rebuild-from-scratch from the MCP-tool-surface does NOT resolve it.

## Root cause (empirically verified)

[...]

PR #316, opened 2026-05-02
fix(store): use PASSIVE checkpoint to avoid file-shrink under concurrent readers

### LINKED ISSUE #314: SIGBUS on macOS arm64 when 2+ sessions share the same cache dir

edwardmhughes:
## Summary

`cbm_store_checkpoint()` used `SQLITE_CHECKPOINT_TRUNCATE` — the most aggressive checkpoint mode. On success it calls `ftruncate(fd, 0)` on the WAL file. When two `codebase-memory-mcp` processes share a cache dir, the truncation can shrink files while a sibling has the DB mmap'd through SQLite, raising `SIGBUS` on macOS (`cluster_pagein past EOF`).

Switching to `SQLITE_CHECKPOINT_PASSIVE` closes this truncation path:
- PASSIVE never blocks readers
- PASSIVE never calls `ftruncate()` on either file
- SQLite still autocheckpoints in PASSIVE mode at 1000-page boundaries; disk reclamation is unaffected for single-process users

From the SQLite docs on [wal_checkpoint](https://www.sqlite.org/pragma.html#pragma_wal_checkpoint):
> PASSIVE mode does not wait for writers and never invokes a blocking read/write lock.

## Changes

- `src/store/store.c`: `SQLITE_CHECKPOINT_TRUNCATE` → `SQLITE_CHECKPOINT_PASSIVE` with explanatory comment
- `tests/test_store_checkpoint.c` (new): Verifies WAL is not truncated to zero bytes after `cbm_store_checkpoint()` — this test would fail on the old TRUNCATE mode under the right timing
- `Makefile.cbm` + `tests/test_main.c`: Wire new test suite

Fixes #314
Related: #277 (orphan-process WAL pin on Windows)

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -279,7 +279,8 @@ TEST_STORE_SRCS = \
     tests/test_store_edges.c \
     tests/test_store_search.c \
     tests/test_store_arch.c \
-    tests/test_store_bulk.c
+    tests/test_store_bulk.c \
+    tests/test_store_checkpoint.c
 
 TEST_CYPHER_SRCS = \
     tests/test_cypher.c
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -790,7 +790,13 @@ int cbm_store_create_indexes(cbm_store_t *s) {
 /* ── Checkpoint ─────────────────────────────────────────────────── */
 
 int cbm_store_checkpoint(cbm_store_t *s) {
-    int rc = sqlite3_wal_checkpoint_v2(s->db, NULL, SQLITE_CHECKPOINT_TRUNCATE, NULL, NULL);
+    /* PASSIVE never blocks readers and never ftruncate()s either file.
+     * SQLite recommends PASSIVE for shared databases — TRUNCATE shrinks
+     * the WAL via ftruncate(fd, 0) on success, which on macOS can raise
+     * SIGBUS in a sibling process that has the DB mmap'd through SQLite
+     * when it next faults a page in the now-shorter region.
+     * See https://www.sqlite.org/c3ref/c_checkpoint_full.html */
+    int rc = sqlite3_wal_checkpoint_v2(s->db, NULL, SQLITE_CHECKPOINT_PASSIVE, NULL, NULL);
     if (rc != SQLITE_OK) {
         store_set_error_sqlite(s, "checkpoint");
         return CBM_STORE_ERR;
--- a/tests/test_main.c
+++ b/tests/test_main.c
@@ -43,6 +43,7 @@ extern void suite_go_lsp(void);
 extern void suite_c_lsp(void);
 extern void suite_store_arch(void);
 extern void suite_store_bulk(void);
+extern void suite_store_checkpoint(void);
 extern void suite_traces(void);
 extern void suite_configlink(void);
 extern void suite_infrascan(void);
@@ -79,6 +80,7 @@ int main(void) {
     RUN_SUITE(store_edges);
     RUN_SUITE(store_search);
     RUN_SUITE(store_bulk);
+    RUN_SUITE(store_checkpoint);
 
     /* Cypher (M6) */
     RUN_SUITE(cypher);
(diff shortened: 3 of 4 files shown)


DeusData:
Landed on `main` as `2215356`, thanks @edwardmhughes. The `SQLITE_CHECKPOINT_TRUNCATE` → `PASSIVE` switch is precisely the right move for shared-cache scenarios — the `ftruncate(fd, 0)` that TRUNCATE invokes is exactly what makes sibling-process mmap pages disappear underfoot, and the SQLite docs explicitly recommend PASSIVE for this case.

Two notes:
- Because #315 (also yours, also touched `Makefile.cbm` and `tests/test_main.c` to wire a new test suite) landed first, this branch ended up conflicting on those two files. GitHub couldn't auto-rebase, so I resolved the conflicts locally by taking the union (both `tests/test_store_pragmas.c` and `tests/test_store_checkpoint.c` are now wired in side by side) and cherry-picked your commit with author preserved. The PR shows as closed rather than merged, but the content + your authorship landed verbatim on `main`.
- The `test_store_checkpoint.c` smoke test (asserting WAL not truncated to zero) is the right shape — it'd fail under the old TRUNCATE mode under the right timing, so it actually defends the fix.

Pairs nicely with the `CBM_SQLITE_MMAP_SIZE=0` knob from #315 — together they close both halves of the macOS multi-instance SIGBUS scenario in #314.

PR #322, opened 2026-05-07
fix(c-lsp): guard pending template call resolution

romanornr:
## Summary
- Add defensive bounds checks while resolving pending C++ template calls
- Avoid reading past signature parameter arrays when recovery-mode parses produce extra call arguments
- Add a regression test covering mismatched template call argument counts

## Test plan
- `make -f Makefile.cbm test -j$(nproc)`

## Notes
This was found while indexing a large Bitcoin-derived C++ codebase, where malformed/recovery-mode AST shapes could expose unsafe assumptions in the C/C++ LSP resolver.

### DIFF
--- a/internal/cbm/lsp/c_lsp.c
+++ b/internal/cbm/lsp/c_lsp.c
@@ -279,10 +279,15 @@ static void c_add_pending_template_call(CLSPContext* ctx, const char* func_qn,
         ctx->pending_template_calls = new_arr;
         ctx->pending_tc_cap = new_cap;
     }
+    const char* func_qn_copy = cbm_arena_strdup(ctx->arena, func_qn);
+    const char* type_param_copy = cbm_arena_strdup(ctx->arena, type_param);
+    const char* method_name_copy = cbm_arena_strdup(ctx->arena, method_name);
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you, @romanornr! 🙏 This is a clean, surgical fix for a real segfault (#215). The root-cause framing was spot-on — the call site genuinely can carry more arguments than the parsed signature knows about (invalid code, macros, variadic calls, or parser recovery), and the old loop walked `param_types[i]` straight past the NULL sentinel. Clamping to `min(call_arg_count, formal_count)` is exactly right, and I appreciated the belt-and-suspenders touches: doing the arena `strdup`s before bumping `pending_tc_count` (so a half-initialized entry can never exist), plus the defensive NULL guards before `strcmp`. The `invoke(w, 1, 2)` regression test reproduces the overflow path perfectly.

Merged via squash (`05f0a26b`) — your authorship is preserved. Verified locally: build clean, all 3,617 tests pass including your new `clsp_nocrash_template_extra_call_args`. This also helps close out the stability cluster tracked in #390. Thanks again!

PR #325, opened 2026-05-08
fix(windows): error on full index - use cmd.exe-compatible syntax (Fixes #324)

### LINKED ISSUE #324: Windows: error on full index
**Problem**
On full index under Windows there is an error after step 5: `The filename, directory name, or volume label syntax is incorrect.`
Example output:
```
[1/9] Building file structure
  Extracting: N/N files (100%)
[2/9] Extracting definitions
[3/9] Building registry
[4/9] Resolving calls & edges
[5/9] Detecting tests
The filename, directory name, or volume label syntax is incorrect.
[7/9] Analyzing git history
[8/9] Linking config files
[9/9] Writing database
```
**Reason**
Between [5/9] and [7/9], the pipeline runs `run_githistory()`. In `src/pipeline/pass_githistory.c` the non-libgit2 fallback builds a POSIX command:
```
  "cd '%s' && git log --name-only --pretty=format:COMMIT:%%H "
  "--since='1 year ago' --max-count=10000 2>/dev/null"
```
`cbm_popen` on Windows is `_popen` → `cmd.exe /c`. `cmd.exe` rejects the single-quoted forward-slash path and `/dev/null`, prints `The filename, directory name, or volume label syntax is incorrect.` to stderr, and returns nonzero

hippus:
Previously on full index under Windows there is an error after step 5: `The filename, directory name, or volume label syntax is incorrect.`
This PR fixes cmd.exe syntax
```
[1/9] Building file structure
  Extracting: N/N files (100%))
[2/9] Extracting definitions
[3/9] Building registry
[4/9] Resolving calls & edges
[5/9] Detecting tests
[7/9] Analyzing git history
[8/9] Linking config files
[9/9] Writing database
```

### DIFF
--- a/src/pipeline/pass_githistory.c
+++ b/src/pipeline/pass_githistory.c
@@ -223,10 +223,16 @@ static int parse_git_log(const char *repo_path, commit_t **out, int *out_count) 
     }
 
     char cmd[CBM_SZ_1K];
+#ifdef _WIN32
+    /* cmd.exe does not recognize single quotes; '/dev/null' is a POSIX path. */
+    const char *null_dev = "NUL";
+#else
+    const char *null_dev = "/dev/null";
+#endif
     snprintf(cmd, sizeof(cmd),
-             "cd '%s' && git log --name-only --pretty=format:COMMIT:%%H "
-             "--since='1 year ago' --max-count=10000 2>/dev/null",
-             repo_path);
+             "git -C \"%s\" log --name-only --pretty=format:COMMIT:%%H "
+             "--since=\"1 year ago\" --max-count=10000 2>%s",
+             repo_path, null_dev);
 
     FILE *fp = cbm_popen(cmd, "r");
     if (!fp) {


DeusData:
Thank you, @hippus! 🙏 Clean diagnosis of a real Windows blocker (#324). You're exactly right: `cd '%s' && ... 2>/dev/null` is POSIX-shell syntax — cmd.exe doesn't understand single quotes and `/dev/null` isn't a valid path, which is what produced "The filename, directory name, or volume label syntax is incorrect" right at the git-history step of a full index. Switching to `git -C "%s" ... 2>NUL` is the correct cross-platform form (and `git -C` is more robust than `cd &&` even on POSIX).

Since the PR branch predated the `:%%ct` commit-timestamp field that was added to the pretty-format string (temporal-properties work, #257), it no longer applied cleanly to `main`, so I distilled your fix onto current main with that field preserved and credited you as the commit author:

**Landed in df38e33**, which closes #324. Verified locally: build clean, all 3,617 tests pass (githistory pass runs fine). I also confirmed the double-quoted path is safe — `cbm_validate_shell_arg` rejects `"`, `$`, backtick and backslash before the command is built. Thanks for making Windows better! 🪟

PR #339, opened 2026-05-11
fix(extraction): convert remaining fixed-size stacks in extract_channels.c to growable TSNodeStack

jjoos:
## Problem

PR #217 converted most AST traversal functions to use the new growable `TSNodeStack` (arena-allocated, no hard cap), but **`extract_channels.c` still had 8 functions** using fixed-size `TSNode stack[CHAN_STACK_CAP]` arrays:

- `scan_string_consts_python`
- `extract_channels_python`
- `extract_channels_go`
- `extract_channels_java`
- `extract_channels_csharp`
- `extract_channels_ruby`
- `extract_channels_elixir`
- `extract_channels_rust`

When indexing large repositories with deep AST nesting (e.g. 4096+ nodes on a single DFS path), these arrays overflow **silently** — the capacity guard `&& top < CHAN_STACK_CAP` drops children without warning, and when the stack itself overflows it corrupts adjacent heap memory. The corruption surfaces as a segmentation fault in a later pass (observed in `lsp_cross` and `semantic_edges` passes on large repos).

## Fix

Apply the same mechanical conversion used in PR #217 to the 8 remaining functions:

| Before | After |
|--------|-------|
| `TSNode stack[CHAN_STACK_CAP]; int top = 0;` | `TSNodeStack stack; ts_nstack_init(&stack, ctx->arena, CHAN_STACK_CAP);` |
| `stack[top++] = ctx->root` | `ts_nstack_push(&stack, ctx->arena, ctx->root)` |
| `while (top > 0)` | `while (stack.count > 0)` |
| `TSNode node = stack[--top]` | `TSNode node = ts_nstack_pop(&stack)` |
| `&& top < CHAN_STACK_CAP` capacity guard | removed (arena grows automatically) |
| `stack[top++] = ts_node_child(...)` | `ts_nstack_push(&stack, ctx->arena, ts_node_child(...))` |

## Testing

Verified by indexing large repositories (1290 files, 13,735 nodes;  3482+ functions) — previously these repos caused segfaults, with this fix all repos index cleanly.

### DIFF
--- a/internal/cbm/extract_channels.c
+++ b/internal/cbm/extract_channels.c
@@ -141,7 +141,7 @@ static void scan_string_consts_python(CBMExtractCtx *ctx, chan_const_table_t *tb
     ts_nstack_init(&stack, ctx->arena, CHAN_STACK_CAP);
     ts_nstack_push(&stack, ctx->arena, ctx->root);
 
-    while (top > 0 && tbl->count < CHAN_CONST_CAP) {
+    while (stack.count > 0 && tbl->count < CHAN_CONST_CAP) {
         TSNode node = ts_nstack_pop(&stack);
         const char *kind = ts_node_type(node);
 
--- a/internal/cbm/extract_channels.c
+++ b/internal/cbm/extract_channels.c
@@ -137,12 +137,12 @@ static void scan_string_consts_js(CBMExtractCtx *ctx, chan_const_table_t *tbl) {
 
 /* Python constant resolution: NAME = "value" (assignment node). */
 static void scan_string_consts_python(CBMExtractCtx *ctx, chan_const_table_t *tbl) {
-    TSNode stack[CHAN_STACK_CAP];
-    int top = 0;
-    stack[top++] = ctx->root;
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you, @jjoos! 🙏 This is a great catch and a clean follow-through on the PR #217 work. You're exactly right about the failure mode: the `&& top < CHAN_STACK_CAP` guard silently dropped children on deep ASTs (so channels would go undetected), and an overflow could corrupt adjacent stack memory and surface as a segfault in a *later* pass — which is maddening to debug. Converting all 8 walkers to the arena-backed growable `TSNodeStack` is the right fix and keeps them consistent with `extract_calls.c` / `extract_imports.c`.

I landed it as **4d84406**, crediting you as the author. The only change I made on top of your patch was trimming a stray trailing blank line at EOF so the file stays clang-format-clean — the conversion itself is entirely yours. Verified locally: build clean, all 3,617 tests pass. This also feeds into the stability cluster in #390. Thanks for tightening this up! 🙏

PR #342, opened 2026-05-12
Use CLOCK_MONOTONIC if available

cmeerw:
Only fall back to hardcoded values if `CLOCK_MONOTONIC` is not defined


cmeerw:
Should have mentioned, this fixes NetBSD (and likely other BSD) build, as NetBSD has `#define CLOCK_VIRTUAL   1` and `#define CLOCK_MONOTONIC 3`

DeusData:
Thank you, @cmeerw! 🙏 Nice improvement — preferring the real `CLOCK_MONOTONIC` constant when the platform defines it, instead of the hardcoded 6/1 fallback, is both more correct on non-Linux/macOS platforms and clears a magic-number lint warning. Behavior is identical on the platforms we already handle (macOS `CLOCK_MONOTONIC`==6, Linux==1). Verified locally: build clean, all 3,617 tests pass. Merging via squash — authorship preserved. Thanks!

Issue #331, opened 2026-05-09
Indexing repository unexpectedly indexes C:/Users/{USERNAME} instead of selected repo path (security/privacy risk)


LeoTheLegion:
Summary:
When I index a specific repository in codebase-memory-mcp from VS Code, it appears to index C:/Users/{USERNAME} instead of the selected repository directory.

Why this matters:
This can unintentionally ingest unrelated local files outside the target repository, which is a privacy/security concern.

Environment:
1. OS: Windows
2. Install method: PowerShell installer
3. Client: VS Code (MCP integration)
4. Repository intended for indexing: specific workspace folder only

Steps to reproduce:
1. Install codebase-memory-mcp via PowerShell setup on Windows.
2. Use it in VS Code MCP.
3. Run index on a specific repository path.
4. Check resulting indexed project scope/name/content.

Expected behavior:
Only the provided repository directory is indexed.

Actual behavior:
Indexing scope resolves to C:/Users/{USERNAME} (or appears to include that directory), not just the selected repository.

Impact:
Potential accidental indexing of sensitive local data outside the intended repo.

DeusData:
Verified fixed on v0.7.0. Indexing `C:/cbmrepo` in Windows CI ([run](https://github.com/DeusData/codebase-memory-mcp/actions/runs/26710571570)) registers the project with `root_path: "C:/cbmrepo"` (the path I gave) and file nodes `src/server.mjs` / `src/sub/calc.js` — **not** `C:/Users/...`. The wrong-path scoping is gone (wide-char path handling, #386). Closing — this also removes the privacy concern of ingesting the home directory. Reopen if a specific path still resolves to the home dir on v0.7.0. 🙏

PR #352, opened 2026-05-14
fix: critical bug fixes from codebase audit

code-by-mahereddy:
### Critical
- **LIKE pattern bug** (`store.c`): Double-percent format string produced wrong SQL LIKE pattern, making `find_edges_by_url_path` always fail
- **Use-after-free** (`store.c`): `like_pool_add` could free pointer, but subsequent code still used it
- **Memory leak** (`cypher.c`): `expr_free` silently dropped subtrees when 128-entry stack overflowed

### High
- **Path traversal** (`store.c`, `mcp.c`, `http_server.c`): Project names were not validated before interpolation into file paths
- **JSON-RPC 2.0 violation** (`mcp.c`): Response missing both `result` and `error` keys when both were NULL
- **Buffer overflow** (`mcp.c`): `collect_db_project_names` offset could exceed buffer size on truncation
- **XSS** (`http_server.c`): Directory names injected raw into JSON without escaping
- **NULL checks** (`store.c`): Missing NULL guards on `cbm_store_checkpoint`, `cbm_store_node_degree`, `cbm_store_restore_from`

### New utility
- `cbm_validate_project_name()` in `str_util.c/h` — validates project names are safe for file path construction

All fixes verified with successful build (`make -f Makefile.cbm cbm` with -Werror).

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -742,11 +742,19 @@ static void expr_free(cbm_expr_t *e) {
             }
             free(cur->cond.in_values);
         }
-        if (cur->right && top < EXPR_FREE_STACK) {
-            stack[top++] = cur->right;
+        if (cur->right) {
+            if (top < EXPR_FREE_STACK) {
+                stack[top++] = cur->right;
+            } else {
+                expr_free(cur->right); /* recurse when stack overflows */
+            }
         }
-        if (cur->left && top < EXPR_FREE_STACK) {
-            stack[top++] = cur->left;
+        if (cur->left) {
+            if (top < EXPR_FREE_STACK) {
+                stack[top++] = cur->left;
+            } else {
+                expr_free(cur->left); /* recurse when stack overflows */
+            }
         }
         free(cur);
     }
--- a/src/foundation/str_util.c
+++ b/src/foundation/str_util.c
@@ -270,6 +270,24 @@ bool cbm_validate_shell_arg(const char *s) {
     return true;
 }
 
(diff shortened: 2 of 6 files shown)


DeusData:
Thank you, @code-by-mahereddy! 🙏 I gave this a careful line-by-line security review (it touches several sensitive files, so I wanted to be thorough) and every change held up — these are real, well-targeted fixes:

- **`cbm_validate_project_name()`** is a proper allowlist (rejects `..`, path separators, leading dot, anything outside `[A-Za-z0-9._-]`) with no bypass — good path-traversal hardening, and I confirmed it doesn't reject any legitimate existing project names (full suite still green).
- The **`like_pool` use-after-free** is genuine: `like_pool_add()` `free()`s its argument when the pool is full, so the subsequent `where_bind_text(…, lp)` was touching freed memory. Capturing `pool_was_full` and skipping the bind is exactly right.
- **JSON-RPC `result: null`** for the both-NULL case, the **browse-endpoint JSON-escaping** (stored-XSS prevention), the **`expr_free` leak** fix, and the NULL guards all check out.

The PR branch had diverged from `main` (the store.c like-pool area was refactored in the v0.7.0 search work, and there was a stray EOF newline change in cypher.c), so rather than ask you to rebase six files I distilled your fixes onto current main and credited you as the commit author:

**Landed in 21c73b5.** Verified locally: build clean, all 3,617 tests pass including the validation-path tests. This strengthens the security posture tracked in #397 and adds defense-in-depth for #331. Genuinely excellent audit work — thank you! 🙏

PR #354, opened 2026-05-14
feat: add support for MCP ping

### LINKED ISSUE #353: Need to support "ping" requests
Some clients seem to rely on it. For example, [crush](https://github.com/charmbracelet/crush) sends a ping and if it doesn't get a success response forks another server process.

cmeerw:
Fixes #353

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -4294,6 +4294,8 @@ char *cbm_mcp_server_handle(cbm_mcp_server_t *srv, const char *line) {
         start_update_check(srv);
         detect_session(srv);
         maybe_auto_index(srv);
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you, @cmeerw! 🙏 This directly fixes #353 — clients like [crush](https://github.com/charmbracelet/crush) send a `ping` and fork a second server process if they don't get a success response, so a correct empty-result reply matters. Returning `{}` is exactly the MCP-spec response for ping, and `heap_strdup` keeps the ownership consistent with the other `result_json` branches. Verified locally: build clean, all 3,617 tests pass. Merging via squash — authorship preserved. Closes #353. Thanks for making the MCP surface more spec-compliant!

PR #360, opened 2026-05-17
Fix C++ template call segfault

Smiie-2:
This fixes a crash in the C/C++ LSP path when resolving nested template function calls with default template arguments. The underlying issue was that unbound template parameters could flow through substitution as
  `NULL`, and call-expression return handling assumed the substituted type was always valid.

  Example
  MangoHud’s src/hud_elements.cpp was one reproducer, but the fix targets the general case of calls like a nested template function inside another call expression

  What changed:
  - Preserved unbound template parameters during type substitution instead of turning them into `NULL`.
  - Added a guard before unwrapping substituted call return types.
  - Improved template-function lookup in the C++ call resolver path.
  - Tightened pending template-call deduction so it respects the registered formal parameter count.
  - Made C++ template-declaration unwrapping handle nested template declarations more reliably.
  - Added regressions for the general substitution bug and for the nested template-call crash pattern.

  Verification:
  - `make -f Makefile.cbm cbm`
  - `clsp_nocrash_template_function_multi_param_nested_call PASS`
  - `typerep_substitute_unbound_param_preserved PASS`

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -515,15 +515,31 @@ static TSNode resolve_r_func_name(TSNode node) {
 // Forward declaration for mutual recursion.
 static TSNode resolve_func_name(TSNode node, CBMLanguage lang);
 
+static bool is_cpp_template_inner_kind(const char *kind) {
+    return strcmp(kind, "function_definition") == 0 || strcmp(kind, "declaration") == 0 ||
+           strcmp(kind, "field_declaration") == 0;
+}
+
 // C++/CUDA: find inner function/declaration inside template_declaration.
 // Returns the inner node (not the resolved name) to break the recursive cycle.
-static TSNode resolve_template_inner_node(TSNode node) {
+static TSNode find_cpp_template_inner_node(TSNode node, CBMLanguage lang) {
+    if ((lang != CBM_LANG_CPP && lang != CBM_LANG_CUDA) ||
+        strcmp(ts_node_type(node), "template_declaration") != 0) {
+        return node;
+    }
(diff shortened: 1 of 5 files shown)


DeusData:
Thank you, @Smiie-2! 🙏 This is a really solid crash fix and the root-cause analysis was excellent — you correctly identified that unbound template parameters were flowing through `cbm_type_substitute` as `NULL` and then getting dereferenced in the call-expression return path. The MangoHud `hud_elements.cpp` reproducer was perfect for isolating it.

I reviewed each change closely:
- `cbm_type_substitute` returning the original param `t` instead of `NULL` for unbound args (both the TYPE_PARAM and named-param paths) is the right root-cause fix.
- The `if (!ret) return cbm_type_unknown();` guard before return-type unwrapping is the actual NULL-deref backstop.
- The `find_cpp_template_inner_node` refactor (recursing into nested `template_declaration`s + handling `field_declaration`) and the `template_function` call-node resolution are nice additive improvements.

One note on landing: the formal-parameter-count clamp that your PR also included had meanwhile landed independently via #322, so to avoid a double-fix I took only the additive parts of your change (substitution NULL-preservation, the return guards, the template lookup, and both of your new tests). I distilled it onto current main and credited you as the commit author:

**Landed in d895c16.** Verified locally: build clean, all 3,619 tests pass — including your `clsp_nocrash_template_function_multi_param_nested_call` and `typerep_substitute_unbound_param_preserved`. Together with #322 this should close out the C++ template crash class tracked in #390 (relates to #215/#312). Thank you for the deep fix! 🙏

Issue #363, opened 2026-05-19
Linux: cbm_system_info / cbm_default_worker_count don't respect cgroup CPU/memory limits


yangsec888:
### Summary

When CBM runs inside a Linux container with cgroup CPU and/or memory
limits (the normal Kubernetes / Docker / Nomad / systemd-slice case),
`cbm_system_info()` and `cbm_default_worker_count()` report
*node-level* CPU count and RAM, not the cgroup's effective limits.
This produces three concrete operator-visible problems:

1. `cbm_default_worker_count(initial=true)` returns
   `sysconf(_SC_NPROCESSORS_ONLN)`, which on Linux is the number of
   *online host CPUs*. A 1-vCPU container scheduled onto a 16-core
   node spawns ~16 indexing workers, each with its own per-worker
   buffers (AST stacks, tree-sitter parsers, slab allocators).
   In a memory-constrained container this is the dominant OOMKill
   driver — far more so than `g_budget` (which is observability-only;
   `cbm_mem_over_budget()` and `cbm_vmem_over_budget()` are declared
   but never called).
2. `mem.init budget_mb=…` log line is computed from the host's
   `sysinfo(2).totalram`, so a 2 GiB-limited pod on a 62 GB node
   logs `budget_mb=31000`, which is alarming and confuses incident
   response. (Cosmetic, but compounds 1.)
3. The over-budget worker count also magnifies the blast radius of
   #317 (dump-phase OOM crash on large TS monorepo) and #334 (silent
   index corruption after rapid kill/restart) — both of which we have
   reproduced in the wild on Kubernetes pods.

### Where in the source

Verified against `main` at HEAD `22153563cd1072b4f79e0e27113f6e0dea3abc1a`:

- [`src/foundation/system_info.c` `detect_system_linux()`](https://github.com/DeusData/codebase-memory-mcp/blob/22153563cd1072b4f79e0e27113f6e0dea3abc1a/src/foundation/system_info.c#L108-L122) uses `sysconf(_SC_NPROCESSORS_ONLN)` for cores and `sysinfo(2)` for RAM, both host-scoped.
- [`src/foundation/system_info.c` `cbm_default_worker_count()`](https://github.com/DeusData/codebase-memory-mcp/blob/22153563cd1072b4f79e0e27113f6e0dea3abc1a/src/foundation/system_info.c#L162-L171) consumes the host-scoped value with no cgroup adjustment.
[...]

Issue #340, opened 2026-05-11
Segmentation fault at start of lsp_cross pass on repositories with 200+ files


jjoos:
## Description

After the May 8-9 LSP rework (commits adding `py_lsp`, `php_lsp`, TypeScript hybrid resolver, and the `lsp_cross` pipeline pass), `codebase-memory-mcp` segfaults immediately at the start of the `lsp_cross` pass when indexing medium-to-large repositories.

## Reproduction

The crash happens consistently on any repository with ~200+ files. Observed on:
- `project 1`: 268 files, 13,671 extracted nodes — crashes every time
- `project 2`: 1,290 files — crashes after `semantic_edges` pass (heap corruption from earlier)

## Logs

```
level=info msg=parallel.extract.done nodes=13671 errors=0
level=info msg=pass.timing pass=parallel_extract elapsed_ms=130
level=info msg=parallel.registry.start files=268
level=info msg=parallel.registry.done entries=889 defines=13671 imports=206
level=info msg=pass.timing pass=registry_build elapsed_ms=14
level=info msg=pass.start pass=lsp_cross files=268
Segmentation fault
```

The crash occurs within ~20ms of `lsp_cross` starting — before any per-file work is logged — suggesting an initialization bug (NULL dereference, uninitialized pointer, or allocation failure).

## Bisect

- **`10a9b83`** (May 8, before `feat(pipeline): consume LSP-resolved calls in pass_calls`): also crashes, but on a different repo and pass — suggests the bug may be in the underlying LSP resolver infrastructure landed just before
- **Commits before `2e042da`** (May 8, first Python LSP commit): likely safe but not yet confirmed  
- **`06f7e0ee`** (April 30, pre-LSP): fully working — all repos index cleanly

## Environment

- Architecture: `linux/arm64` (Docker, OrbStack on Apple M-series)
- Build: compiled from source with `gcc` on Ubuntu 24.04
- Confirmed reproducible across 3+ clean builds with `--no-cache`

## Related

PR #339 fixes a separate but related issue (remaining fixed-size stacks in `extract_channels.c`) — that fix alone does not resolve this crash.

DeusData:
Fixed by the v0.7.0 cross-LSP rewrite (same root cause as #344 — a scale-dependent crash in the old `lsp_cross` pass). The pass was fused into the parallel resolve worker and is now fully arena-based, removing the fixed-size def-array assumption that segfaulted at ~200+ files / ~1189 defs.

[...]

Issue #336, opened 2026-05-11
Segmentation fault (core dumped)


Al2Klimov:
## How to reproduce

1. Install Fedora 42
2. `curl -fsSL https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh | bash`
3. `git clone --single-branch -b v2.16.0 https://github.com/Icinga/icinga2`
4. `cd icinga2`
5. Create prompt.txt
6. `/home/fedora/.local/bin/codebase-memory-mcp < prompt.txt`
7. "Segmentation fault (core dumped)"

### prompt.txt

```
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"0.0.1"}}}
{"jsonrpc":"2.0","method":"notifications/initialized","params":{}}
{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"index_repository","arguments":{"repo_path":"/home/fedora/icinga2","mode":"fast"}}}
```

Al2Klimov:
This is what Claude found as root cause:

```patch
--- a/internal/cbm/lsp/c_lsp.c
+++ b/internal/cbm/lsp/c_lsp.c
@@ -303,7 +303,8 @@ static void c_resolve_pending_template_calls(CLSPContext* ctx,
         callee->signature->data.func.param_types) {
         for (int i = 0; i < call_arg_count; i++) {
             const CBMType* formal = callee->signature->data.func.param_types[i];
-            if (!formal || !call_arg_types[i]) continue;
+            if (!formal) break;               // NULL terminates param_types array
+            if (!call_arg_types[i]) continue; // unknown arg — skip, keep looping
             // Unwrap references/pointers
             while (formal && (formal->kind == CBM_TYPE_REFERENCE ||
                               formal->kind == CBM_TYPE_RVALUE_REF ||
```

This fixes my problem, idk your codebase though.

DeusData:
Verified fixed on v0.7.0. I reproduced your exact setup — cloned `Icinga/icinga2` at `v2.16.0` and indexed it in `fast` mode — on **both macOS and Linux** (matching your Fedora platform), in CI ([run](https://github.com/DeusData/codebase-memory-mcp/actions/runs/26696880113)):

```
parallel.extract.done nodes=13753 errors=0
pipeline.done ... {"project":"icinga2-repro","status":"indexed"}
[...]

Issue #334, opened 2026-05-11
Silent index corruption after rapid kill/restart cycles


nguiaSoren:
status:"ready" with ~5% of true node count

**Version**: codebase-memory-mcp 0.6.0
**Platform**: macOS

**Reproduction** (intermittent but reproducible during heavy session use):
1. Multiple codebase-memory-mcp processes running concurrently (e.g., 
   multiple Claude Code sessions across the day)
2. Rapid kill of one or more processes
3. New session reconnects — MCP appears functional
4. index_status reports status:"ready" with node count ~5-10% of true 
   value (e.g., 469 nodes for a project with 5,915 true baseline)
5. index_repository(mode="full") does NOT recover — repeated calls 
   return same low count with status:"indexed"
6. ONLY recovery: full `/mcp` disconnect + reconnect at the MCP client 
   level, followed by index_repository(mode="full")

**Why this matters**: status:"ready" / status:"indexed" should not 
report success when the index is catastrophically partial. Suggested 
fix: validate node count against a plausibility threshold (or against 
prior known-good state) before returning ready.

**Diagnostic data**: 38k LOC Rust codebase, 9 source files. True 
baseline 5,915 nodes / 9,531 edges (verified via schema sum + 
known-symbol probes across all source files). Corrupted state 
returned 169-572 nodes / 165-555 edges across multiple rebuild attempts.

DeusData:
Two parts here:

**1. Root cause (durability).** The varying low counts (469/540/572 across rebuilds) and "only a full process restart recovers" are the signature of WAL data stranded by a hard-killed sibling process: `synchronous=OFF` + WAL means uncheckpointed frames can be lost, and an orphan process holding a WAL read-lock blocks the checkpoint. **v0.7.0 (#277)** now checkpoints on close and recovers a stale WAL on open, which should fix the durability loss. Please retest on v0.7.0 with your kill/restart workflow.

**2. Your suggestion — a plausibility gate before reporting `status:"indexed"`.** I agree the silent-success is bad, but I want your sign-off on the heuristic before shipping it, because a naive node-count threshold risks false-failures on legitimately sparse repos (generated code, config-heavy trees, single-language repos). Candidate designs:
   (a) compare post-index node count against the **prior known-good count** for that project (stored in `_config.db`); flag a regression if it drops > X%.
[...]

PR #364, opened 2026-05-19
feat(platform): CBM_WORKERS env override for cbm_default_worker_count

yangsec888:
### What

Adds a single env-knob, `CBM_WORKERS`, that explicitly sets the
worker count returned by `cbm_default_worker_count()`. When unset,
behaviour is unchanged.

### Why

In containerized deployments, `cbm_default_worker_count(initial=true)`
returns `sysconf(_SC_NPROCESSORS_ONLN)` (host-CPU count, not the
container's CPU quota). On a 1-vCPU pod scheduled onto a 16-core
node cbm spawns ~16 indexing workers, each with its own per-worker
buffers — the dominant OOMKill driver in container deployments
(see companion issue #363 for the broader cgroup-awareness ask).

A small env knob solves the immediate operator pain without taking
on the scope of full cgroup detection, and it follows the same
shape as the existing `CBM_*` env overrides: explicit override >
implicit detection.

### Patch

`src/foundation/system_info.c`:

```c
int cbm_default_worker_count(bool initial) {
    /* CBM_WORKERS env override (clamped to [1, CBM_WORKERS_MAX]).
     * Useful inside containers where sysconf(_SC_NPROCESSORS_ONLN)
     * reports host CPUs rather than the cgroup's effective CPU quota.
     * Same precedence shape as other CBM_* env overrides:
     * explicit override > implicit detection. */
    char buf[CBM_SZ_32];
    if (cbm_safe_getenv("CBM_WORKERS", buf, sizeof(buf), NULL) != NULL) {
        long n = strtol(buf, NULL, CBM_DECIMAL_BASE);
        if (n >= MIN_WORKERS && n <= CBM_WORKERS_MAX) {
            return (int)n;
        }
        cbm_log_warn("workers.env.invalid", "value", buf,
                     "fallback", "sysconf");
    }

    cbm_system_info_t info = cbm_system_info();
    // ... unchanged from today ...
}
```

`CBM_WORKERS_MAX = 256` is added to the file-local `enum`; happy to
drop the upper clamp if you'd rather trust the operator, or move
the constant somewhere shared. `cbm_safe_getenv`, `CBM_DECIMAL_BASE`,
`CBM_SZ_32`, `SKIP_ONE`, `MIN_WORKERS`, and `cbm_log_warn` all
already exist (see `src/foundation/platform.h`, `constants.h`, and
`log.h`).

### Tests

`tests/test_platform.c` gains three cases inside the existing
`SUITE(platform)`:

1. `platform_default_workers_env_override` — `CBM_WORKERS=4` is honored for both `initial=true` and `initial=false`.
2. `platform_default_workers_env_invalid` — `0`, `-1`, `9999`, and `not-a-number` all fall back to the sysconf-derived default (and emit `workers.env.invalid` at WARN).
3. `platform_default_workers_env_unset` — when unset, `cbm_default_worker_count(true) == cbm_system_info().total_cores`, preserving today's contract.

### Docs

One-line addition to the **Environment Variables** table in
`README.md`:

> `CBM_WORKERS` — Override the parallel-indexing worker count
> returned by `cbm_default_worker_count`. Useful inside containers
> where `sysconf(_SC_NPROCESSORS_ONLN)` reports host CPUs rather
> than the cgroup's effective quota. Range 1–256; invalid values
> are ignored with a warning.

### Behaviour when unset

Identical to today. No deployment that doesn't set `CBM_WORKERS`
sees any difference.

### Verification on my side

I have **not** run `scripts/build.sh` / `scripts/test.sh` /
`scripts/lint.sh` locally because the full ASan + UBSan setup
isn't installed on my machine. Happy to iterate quickly on any
CI feedback (lint, clang-tidy, cppcheck, format) — the patch is
small enough that a turnaround should be fast.

### Related

- Companion issue #363 — full cgroup-aware detection (the broader
  ask this PR carves a smaller, lower-risk path out of).
- Other crash bugs that container-overspawn magnifies: #317, #334,
  #336, #340.
- Downstream context: filed while integrating CBM as a long-lived
  MCP-server process inside a 2 GiB Kubernetes pod. Backend pod's
  current workaround is a memory-limit bump from 2 GiB → 4 GiB to
  ride out the overspawn until this lands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,7 @@ codebase-memory-mcp config reset auto_index              # reset to default
 | `CBM_CACHE_DIR` | `~/.cache/codebase-memory-mcp` | Override the database storage directory. All project indexes and config are stored here. |
 | `CBM_DIAGNOSTICS` | `false` | Set to `1` or `true` to enable periodic diagnostics output to `/tmp/cbm-diagnostics-<pid>.json`. |
 | `CBM_DOWNLOAD_URL` | *(GitHub releases)* | Override the download URL for updates. Used for testing or self-hosted deployments. |
+| `CBM_WORKERS` | *(detected)* | Override the parallel-indexing worker count returned by `cbm_default_worker_count`. Useful inside containers where `sysconf(_SC_NPROCESSORS_ONLN)` reports host CPUs rather than the cgroup's effective quota. Range 1–256; invalid values are ignored with a warning. |
 
 ```bash
 # Store indexes in a custom directory
--- a/src/foundation/system_info.c
+++ b/src/foundation/system_info.c
@@ -10,9 +10,11 @@
  */
 #include "foundation/constants.h"
 
-enum { DEFAULT_CORES = 1, MIN_WORKERS = 1 };
+enum { DEFAULT_CORES = 1, MIN_WORKERS = 1, CBM_WORKERS_MAX = 256 };
+#include "foundation/log.h"
 #include "foundation/platform.h"
 #include <stdint.h> // uint64_t
+#include <stdlib.h> // strtol
 #include <string.h>
 
 #ifdef _WIN32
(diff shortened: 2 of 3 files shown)


DeusData:
Thank you, @yangsec888! 🙏 The `CBM_WORKERS` escape hatch is a clean, well-scoped fix for the container over-provisioning problem — `cbm_safe_getenv` + `strtol` + clamp to [1,256] with a warn-and-fall-back on invalid values is exactly the right shape, and it matches the precedence of the other `CBM_*` overrides. Your three tests (override / invalid / unset) cover it well.

I landed it as **d952238**, crediting you as the commit author (it sequences right before your cgroup-detection work). Verified locally: build clean, all tests pass and the `workers.env.invalid` warnings fire as expected. Thanks for tackling the container OOMKill story! 🙏

PR #420, opened 2026-06-06
docs(eval-plan): mark Appendix A chapters as drafts pending ground-truth regen

DeusData:
Follow-up to the 159-language evaluation plan. Adds an explicit status note at the Appendix A header so reviewers don't mistake the LLM-drafted chapters for the final question bank.

- Question *types* are grounded in the Sillito taxonomy (already final, §3.1).
- Question *targets* (specific symbols/files) are NOT yet ground-truth-derived; at execution they are replaced by SWE-QA (major languages) or LSP/git-seeded targets (the rest), confirmed grep-first against the pinned commit (CR-1).
- `[verify]` tags mark unconfirmed symbols in the drafts.

No merge intended yet — opened for review.

### DIFF
--- a/docs/EVALUATION_PLAN.md
+++ b/docs/EVALUATION_PLAN.md
@@ -1127,10 +1127,20 @@ The evaluation's most serious flaw is not operational — it is that the bespoke
 
 # Appendix A — Per-language chapters (159)
 
-> Each chapter is a *draft specification* (per §12): repo + 5 dimension-tagged questions
-> (symmetric-authoring split), the per-type graph-stats block (all 32 edge types, zeros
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks for the evaluation-plan cleanup. Before this can proceed, please add a DCO `Signed-off-by` trailer to the non-merge commit and link the tracking issue or explain the release/evaluation milestone this supports.

DeusData:
Superseded by a recreated commit with proper DCO author/sign-off (the original predates the repo's DCO email discipline and cannot pass the check without rewriting). Same diff, merged via the successor PR — thanks past-us for the honest disclaimer text.

PR #427, opened 2026-06-09
fix(c_lsp): defensive CBMType pointer validation to prevent SEGV on t…

### LINKED ISSUE #424: Crash in `pass=definitions` on complex C++ files — still broken in 0.7.0
## Crash in `pass=definitions` on complex C++ files — still broken in 0.7.0

**Version:** 0.7.0 (`--version`; `update` confirms it's the latest release)
**Platform:** Windows 11 x86-64
**Status:** Marked fixed in a prior issue (which specifically named `nlohmann/json.hpp`), but still reproduces on
0.7.0.

---

### Repro

Index the unmodified upstream `nlohmann/json.hpp` v3.12.0
(MIT, https://github.com/nlohmann/json/releases/download/v3.12.0/json.hpp):

```sh
mkdir /tmp/repro && cp json.hpp /tmp/repro/
codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/repro"}'
# → Segmentation fault (exit 139 / Windows 0xC0000005)
# Last log line before crash: level=info msg=pass.start pass=definitions files=1
```

Crashes 100% deterministically. Also reproduces on Eigen:s Eigen/src/LU/Determinant.h (independent confirmation with
a different well-known templated header).

Minimum reproducing fragment: the first ~1015 lines of json.hpp 3.12.0 crash; ~1014 lines do not.

---
Key observations

It is cumulative, not triggered by a single construct.
Lines [950..1015] (66 lines, same ending) → OK.
Lines [500..1015] (516 lines, same ending) → CRASH.
The crash threshold scales with how much content precedes a given point, not with the local syntax. This rules out a
single bad construct and points to a capacity being exhausted.

The fault address is 0xffffffffffffffff, not 0x0.
Debugger (cdb) shows the crash is an access violation reading from address 0xffffffffffffffff inside
ts_parser_parse_with_options(). The faulting instruction pattern is:

mov  rbx, [rbx+8]    ; follow "next" pointer in a tree/list node
test rbx, rbx        ; null check — catches 0x0 only
je   (done)
mov  eax, [rbx]      ; CRASH: rbx = 0xffffffffffffffff

Tree-sitter uses 0xffffffffffffffff as its internal null-subtree sentinel. The null check (test/je) only catches 0x0,
not the sentinel, so the walk falls through and dereferences it.

The crash is in the vendored tree-sitter runtime, not in the cbm extraction layer.
Call chain (confirmed via cdb + symbol bracketing):

main → handle_index_repository → cbm_pipeline_run
→ cbm_pipeline_pass_definitions → cbm_extract_file
  → ts_parser_parse_with_options   ← crash is inside here
	[vendored tree-sitter runtime / C++ grammar]

The cbm extraction code (extract_defs.c, c_lsp.c, etc.) correctly uses ts_node_is_null() throughout. The bad
raw-pointer check is in internal/cbm/vendored/ts_runtime/src/.

---
Why the prior fix didnt hold

The regression test suite (test_stack_overflow.c) covers JS/Python/Go/TS/YAML stack-overflow paths but has no C++
crash regression test. Whatever was fixed before, this path was left uncovered and re-broken (likely by the arena/LSP
refactor in PRs #217/#322/#339 which introduced new AST-walking code).

Additionally: --version reports 0.7.0 but the MCP initialize handshake reports 0.10.0. If the prior fix landed under
a different version label, it may never have shipped in the published binary.

---
Suggested fix direction

In internal/cbm/vendored/ts_runtime/src/, find linked-list / child-array traversal loops that check if (ptr == NULL)
and replace with the sentinel-aware predicate (equivalent of ts_node_is_null()). The same pattern that
extract_node_stack.h applied to the extraction layer needs to be applied to the runtime traversal code.

A C++ regression test (index a 1000+ line templated C++ file, assert no crash) would prevent recurrence.
```

mattall:
### Problem

  Indexing heavily-templated C++ can SEGV inside the C LSP type-evaluation pass.
  `c_eval_expr_type` can surface a stale or mis-typed `CBMType*` — an unaligned pointer into
  string data — through chains of template-function registry lookups. Dereferencing it
  crashes. Concretely, indexing ns-3's `src/core/model/ptr.h` triggers a UBSan
  misaligned-access error followed by a SEGV in `c_eval_expr_type_inner`.

  ### Fix

  Add `cbm_type_pointer_looks_valid()` — a cheap defensive check (pointer alignment + `kind`
  in valid range + a shallow `FUNC` `return_types[0]` peek). `c_eval_expr_type` validates
  every result against it and falls back to `cbm_type_unknown()` when the pointer fails,
  
  ### Scope
  
  A defensive guard at the consumption site. It stops the crash and degrades gracefully to
  "unknown type". It does **not** fix the underlying cause — corruption in the
  template-function registry path that produces the bad `CBMType*`. That would be a separate
  change; this is the safety net.
  
  ### Relation to #424 (not a fix)
  
  #424 reports a SEGV in the same broad class (large templated C++ headers). **This PR does
  not fix #424.** That crash is in a different subsystem — the vendored tree-sitter runtime
  (`internal/cbm/vendored/ts_runtime/src/`), in `ts_parser_parse_with_options` during
  `pass=definitions`, where tree-sitter's `0xffffffffffffffff` null-subtree sentinel isn't
  caught by `== NULL` checks. The fix here lives downstream of parsing, in `c_lsp`'s
  type-eval. Noted only because both surface as "SEGV on templated C++".

  ### Testing
  
  Manually verified: indexing ns-3 `src/core/model/ptr.h` no longer SEGVs / no UBSan report;
  the previously-crashing symbol resolves to an unknown type instead of crashing. Happy to
  add a regression fixture if you have a preferred location for crash-repro headers.

### DIFF
--- a/internal/cbm/lsp/c_lsp.c
+++ b/internal/cbm/lsp/c_lsp.c
@@ -5,6 +5,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <stdint.h>
 
 /* Safe kind accessor — returns CBM_TYPE_UNKNOWN for NULL types.
  * Prevents SEGV in c_eval_expr_type_inner on unusual C++ AST shapes. */
@@ -1223,6 +1224,37 @@ static const CBMType* c_eval_expr_type_inner(CLSPContext* ctx, TSNode node);
 #define C_EVAL_DEPTH_LIMIT 256
 #define C_EVAL_MAX_STEPS_PER_FILE 10000
 
+/* Validate that a putative CBMType pointer can be safely dereferenced.
+ * The C/C++ LSP's registry-lookup paths can occasionally surface a stale
+ * or mis-typed pointer through chains of template-function calls (UBSan
+ * catches this on ns-3 src/core/model/ptr.h with templated operator<,
+ * operator==, etc.) — the pointer ends up unaligned and points into
+ * string data. Validating alignment + kind range here keeps the
+ * best-effort evaluator from crashing on those inputs without changing
+ * its result shape (unknown propagates correctly through callers). */
+static bool cbm_type_pointer_looks_valid(const CBMType* t) {
+    if (!t) return false;
+    if ((uintptr_t)t % 8 != 0) return false;
+    /* CBMTypeKind is an enum with ~30 members; reject anything outside a
+     * generous bound to catch garbage values read from misaligned memory. */
+    if ((unsigned)t->kind > 64u) return false;
+    /* For FUNC types we also peek at return_types[0] — the corruption pattern
+     * surfaces a valid-looking FUNC with garbage children, so a shallow check
+     * on the outer pointer isn't enough. We bound-check the array pointer
+     * itself and its first element. */
+    if (t->kind == CBM_TYPE_FUNC) {
+        const CBMType** rts = t->data.func.return_types;
+        if (rts) {
+            if ((uintptr_t)rts % 8 != 0) return false;
+            const CBMType* r0 = rts[0];
+            if (r0 && ((uintptr_t)r0 % 8 != 0 || (unsigned)r0->kind > 64u)) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
 const CBMType* c_eval_expr_type(CLSPContext* ctx, TSNode node) {
     if (ts_node_is_null(node)) return cbm_type_unknown();
     /* Expression type evaluation is best-effort. Some recovery-mode C++ ASTs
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you for this — the crash report and reproduction were excellent and pinpointed the exact failure site in `c_eval_expr_type`.

Digging in from your repro, we found the root cause one layer deeper: `cbm_type_substitute` was indexing `type_args[i]` by the `type_params` loop index, so when a class template declares more params than the instance supplies args (default template args, partial instantiation — very common with STL containers), it read past the args array and returned a bogus `CBMType*` pointing into adjacent string data — exactly the unaligned pointer you saw.

We've fixed it at that source in fb17573 by bounding the args access to the parameter count, which closes the out-of-bounds read for every language through the single substitution chokepoint, rather than validating the already-corrupt pointer downstream. Verified with a reproduce-first regression test plus a full green CI run across Linux/macOS/Windows.

Your `cbm_type_pointer_looks_valid` guard was a sound stopgap, and your report is what made the real fix findable. Closing in favor of the upstream fix — thank you again!

PR #428, opened 2026-06-09
fix(extract_defs): attribute out-of-line C++ methods to their class

mattall:
### Problem

  An out-of-line C++ method definition — a `function_definition` whose declarator is a
  `qualified_identifier`, e.g. `bool Foo::Bar() {}` in a `.cc` file — is emitted as a free
  `Function`, not a `Method` of its class. The class linkage is lost. This also misses the
  very common form where the definition sits inside a namespace block:

  ```cpp
  namespace ns3 {
  bool Foo::Bar() { ... }   // emitted as free Function `Bar`, not Method `Foo::Bar`
  }
  ```

  `main` mislabels the namespace-nested form, which is the dominant pattern across large C++
  codebases (all of ns-3, for example). Upstream currently handles only the top-level
  (non-namespaced) case.

  ### Fix

  
  - `find_qualified_func_declarator()` — locates a `qualified_identifier` declarator on a
    `function_definition`, including through a namespace scope.
  - `extract_qualified_immediate_scope()` — extracts the innermost qualifier (the class).
  
  When a qualified declarator is found, the definition is promoted to `label=Method` with a
  class-qualified `qualified_name` and `parent_class` set to the innermost scope, instead of
  being emitted as a free function.
  
  ### Impact
  
  Out-of-line method definitions now link to their class in the graph, so queries that pivot
  on `parent_class` or method-qualified names resolve correctly for the defined-in-`.cc`
  case, matching how in-class definitions are already handled.
  
  ### Testing
  
  Manually verified against ns-3 sources: `Type::Method() {}` definitions inside
  `namespace ns3 { ... }` now appear as `Method` with the correct `parent_class`, where
  `main` emitted free functions. Happy to add a regression fixture if you point me at the
  preferred extraction-test location.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -469,6 +469,56 @@ static TSNode resolve_qualified_name(TSNode decl) {
     return null_node;
 }
 
+// C++/CUDA: find a qualified_identifier inside the declarator chain of a
+// function_definition (e.g. SemanticLayer::ReceiveUp). Returns null if the
+// declarator is unqualified. Used to detect out-of-line method definitions.
+static TSNode find_qualified_func_declarator(TSNode node) {
+    TSNode decl = ts_node_child_by_field_name(node, TS_FIELD("declarator"));
+    for (int depth = 0; depth < DECLARATOR_DEPTH_LIMIT && !ts_node_is_null(decl); depth++) {
+        const char *dk = ts_node_type(decl);
+        if (strcmp(dk, "qualified_identifier") == 0 || strcmp(dk, "scoped_identifier") == 0) {
+            return decl;
+        }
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you — this was a real gap. Out-of-line method definitions (`Foo::bar` in a .cpp, with the class body declaration-only in a header) were being recorded as free functions with no link back to their class.

We've reimplemented your approach upstream in a50b086: a helper detects the qualified-identifier declarator on a `function_definition` and attaches the immediate enclosing class as `parent_class` (descending nested scopes like `ns::Class::method` down to the direct parent), so the definition is labeled `Method` and `DEFINES_METHOD` edges resolve. Added a reproduce-first regression test for it as well.

Closing in favor of the upstream commit — appreciate you spotting this and the clean approach!

PR #387, opened 2026-05-28
fix(store): checkpoint WAL on close and startup to prevent orphan accumulation

jjserenity:
Fixes #277

## Problem

`codebase-memory-mcp` processes can accumulate as orphan processes (e.g., from unclean stdio shutdown on Windows). Each orphan holds a SQLite WAL read lock, preventing checkpoint from ever landing. New index writes go to WAL but never merge into the main `.db`. The WAL grows unbounded, and queries return stale data.

The existing `cbm_store_checkpoint()` API uses safe `SQLITE_CHECKPOINT_PASSIVE` mode but is **never called anywhere** — zero call sites.

## Changes

Single file: `src/store/store.c` (+10 lines)

1. **`cbm_store_close()`**: Checkpoint WAL via `sqlite3_wal_checkpoint_v2(..., PASSIVE)` before `sqlite3_close_v2`. This ensures graceful shutdown (SIGTERM, `delete_project`, normal exit) leaves a clean WAL that won't require recovery on next open.

2. **`configure_pragmas()`**: After `PRAGMA journal_mode=WAL`, run `PRAGMA wal_checkpoint(PASSIVE)` to merge any stale WAL from a previous crash. Best-effort — silently skipped if another process holds a lock (SQLITE_BUSY).

## Rationale

- PASSIVE mode never blocks readers and never `ftruncate()`s, compatible with PR #316's existing choice.
- Crash + restart → startup checkpoint recovers the WAL.
- Graceful exit → WAL is clean.
- `delete_project` → checkpoint runs inside `cbm_store_close` before file unlink, so WAL is merged before deletion.
- Orphan processes still holding locks → checkpoint silently fails, no regression.

## Tested

- Compiles clean on MinGW GCC 14.2 (`-Wall -Wextra -Werror`)
- CI will verify full test suite

Issue #333, opened 2026-05-11
Silent index degradation — status:"indexed" but only ~500 nodes for 72k LOC Rust codebase


nguiaSoren:
**Title**: Silent index degradation — status:"indexed" but only ~500 nodes for 72k LOC Rust codebase

**Version**: codebase-memory-mcp 0.6.0
**Platform**: macOS (Apple Silicon)
**Project size**: 72k LOC, 9 Rust files (single Cargo project)

## Symptom

`index_repository(mode="full")` returns `status:"indexed"` but produces only ~500 nodes and ~500 edges for a 72k LOC Rust codebase. Expected ~5,000–10,000+ symbol nodes based on comparable smaller projects.

The `status:"ready"` / `status:"indexed"` response is misleading — the rebuild silently fails on something but reports success.

## Reproduction

1. Kill any existing codebase-memory-mcp processes:
   `ps aux | grep codebase-memory` then `kill <PID>`
2. Delete cached index:
   `rm ~/.local/share/codebase-memory-mcp/Users-soren-Desktop-new_simulator.db`
3. Verify directory is empty of project cache
4. Restart Claude Code (spawns fresh MCP server process)
5. In Claude Code: `index_repository(mode="full")`
6. Run `index_status` — shows ~500 nodes / ~500 edges, status:"indexed"

Multiple fresh rebuilds produce the same low node count (469/462, 540/531, 572/555 across attempts).

## Comparison with healthy project

Same machine, same MCP server binary:
- `X` project (~6k LOC): 6,436 nodes / 10,308 edges, 14 MB .db file ✓ healthy
- `Y` project (72k LOC): ~500 nodes / ~500 edges, 2 MB .db file ✗ degraded

The 2 MB file size is suspiciously exactly 2 × 1024 × 1024 bytes — suggests the database is being truncated or capped at a fixed boundary.

## Known-symbol verification

`search_graph(name_pattern="apply_knight_hero_effects")` returns 0 results.

This function exists at `src/hero_system.rs` (verified via grep). Should be indexed as a function symbol but is missing from the graph.

## Workaround

Direct file reads + grep work fine. The bug appears specific to the graph-based query path (`search_graph`, `trace_path`).

## Diagnostic information requested

Happy to provide:
- Sample file from the codebase that may be triggering the issue (if I can isolate which file)
- Full `index_repository` output
- Any debug logging the indexer emits

DeusData:
[...]

PR #433, opened 2026-06-10
Add post-dump plausibility gate returning status:degraded (#334)

### LINKED ISSUE #334: Silent index corruption after rapid kill/restart cycles

yangsec888:
## Summary

- Adds a self-referential plausibility gate after `index_repository` completes: when persisted SQLite node rows fall far below the in-memory graph buffer count at dump time, return `status:"degraded"` with `expected_nodes` / `expected_edges` instead of silent `status:"indexed"`.
- Closes the observability gap left after #387 (WAL checkpoint on open/close) for cases where the store is partial or integrity-check auto-clean removes the DB.

Fixes #334 (design **b** as discussed in the issue thread).

## Motivation

Rapid kill/restart cycles could leave `status:"indexed"` with a small fraction of the true node count. Maintainer agreed on design **(b)**: compare persisted rows to extracted/committed rows (self-referential, no cross-repo assumptions).

## Changes

- `src/foundation/dump_verify.c` — pure `cbm_dump_verify_is_degraded()` + `CBM_DUMP_VERIFY_MIN_RATIO` env (default 0.5, `0` disables).
- `src/pipeline/pipeline.c` — capture `committed_nodes` / `committed_edges` at dump; accessor `cbm_pipeline_get_committed_counts`.
- `src/mcp/mcp.c` — gate after successful pipeline; checkpoint+recount once; new response fields.
- `tests/test_dump_verify.c` — pure-function case matrix.
- `tests/test_dump_verify_io.c` — store I/O tests (real `cbm_store_count_nodes`, shortfall simulation, fork/crash WAL recovery).

## Response shape

```json
{
  "project": "...",
  "status": "degraded",
  "nodes": 469,
  "edges": 1200,
  "expected_nodes": 5915,
  "expected_edges": 9531,
  "hint": "Persisted far fewer nodes than indexed — likely durability loss..."
}
```

`isError` remains false so partial graphs stay queryable. Downstream parsers that only require `nodes`/`edges` continue to work; new `status` is opt-in.

## Design notes

- **Nodes-only gate** — edges legitimately shrink at dump when endpoints fail to resolve; comparing edge counts would false-trigger.
- **`resolve_store` NULL** (integrity auto-clean) → `degraded` with `nodes:0`, not silent `indexed`.
- **`CBM_DUMP_VERIFY_MIN_RATIO=0`** disables the gate (escape hatch).
- **`committed_nodes = -1`** sentinel when dump did not run (explicit init; `calloc` zero would be ambiguous).

## Test plan

- [x] `make -f Makefile.cbm test` green (5581 passed)
- [x] `suite_dump_verify` pure-function matrix
- [x] `suite_dump_verify_io` store I/O + fork/crash WAL recovery (POSIX)
- [ ] Manual: index repo, kill sibling mid-WAL, re-index → `status:"degraded"`

## Related

- Complements #387 / v0.7.0 WAL durability

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -111,7 +111,8 @@ FOUNDATION_SRCS = \
     src/foundation/compat_regex.c \
     src/foundation/mem.c \
     src/foundation/diagnostics.c \
-    src/foundation/profile.c
+    src/foundation/profile.c \
+    src/foundation/dump_verify.c
 
 # Existing extraction C code (compiled from current location)
 EXTRACTION_SRCS = \
@@ -292,7 +293,8 @@ TEST_FOUNDATION_SRCS = \
     tests/test_str_intern.c \
     tests/test_log.c \
     tests/test_str_util.c \
-    tests/test_platform.c
+    tests/test_platform.c \
+    tests/test_dump_verify.c
 
 TEST_EXTRACTION_SRCS = \
     tests/test_extraction.c \
@@ -310,7 +312,8 @@ TEST_STORE_SRCS = \
     tests/test_store_arch.c \
     tests/test_store_bulk.c \
     tests/test_store_pragmas.c \
-    tests/test_store_checkpoint.c
+    tests/test_store_checkpoint.c \
+    tests/test_dump_verify_io.c
 
 TEST_CYPHER_SRCS = \
     tests/test_cypher.c
--- a/README.md
+++ b/README.md
@@ -445,6 +445,7 @@ codebase-memory-mcp config reset auto_index              # reset to default
 | `CBM_DOWNLOAD_URL` | *(GitHub releases)* | Override the download URL for updates. Used for testing or self-hosted deployments. |
 | `CBM_LOG_LEVEL` | `info` | Set the minimum log level. Accepted values (case-insensitive): `debug`, `info`, `warn`, `error`, `none` — or their numeric equivalents `0`–`4` matching the internal enum. Logs go to stderr; stdout is reserved for MCP JSON-RPC. |
 | `CBM_WORKERS` | *(detected)* | Override the parallel-indexing worker count returned by `cbm_default_worker_count`. Useful inside containers where `sysconf(_SC_NPROCESSORS_ONLN)` reports host CPUs rather than the cgroup's effective quota. Range 1–256; invalid values are ignored with a warning. |
+| `CBM_DUMP_VERIFY_MIN_RATIO` | `0.5` | After indexing, compare persisted SQLite node count to the in-memory dump count. When persisted nodes fall below this fraction of committed nodes (and committed > 50), `index_repository` returns `status:"degraded"` instead of silent `indexed`. Range 0–1; set `0` to disable. Invalid values are ignored with a warning. |
 
 ```bash
 # Store indexes in a custom directory
--- a/src/foundation/dump_verify.c
+++ b/src/foundation/dump_verify.c
@@ -1,0 +1,40 @@
+/*
+ * dump_verify.c — Post-dump plausibility gate (#334).
+ */
+#include "foundation/dump_verify.h"
+#include "foundation/constants.h"
+#include "foundation/log.h"
+#include "foundation/platform.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+bool cbm_dump_verify_is_degraded(int committed_nodes, int persisted_nodes, double ratio,
+                                 int min_floor) {
+    if (ratio <= 0.0) {
+        return false;
+    }
+    if (committed_nodes < 0) {
+        return false;
+    }
+    if (committed_nodes <= min_floor) {
+        return false;
+    }
+    if (persisted_nodes < 0) {
+        return true;
+    }
+    return (double)persisted_nodes < (double)committed_nodes * ratio;
+}
+
+double cbm_dump_verify_min_ratio(void) {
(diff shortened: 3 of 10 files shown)


DeusData:
Heads-up: this project now validates every PR automatically — tests, lint, security/license gates, and DCO sign-off ([CONTRIBUTING.md](https://github.com/DeusData/codebase-memory-mcp/blob/main/CONTRIBUTING.md)). Your branch predates this, so CI will flag the missing `Signed-off-by` trailers. Fixing it is one command: `git rebase --signoff origin/main && git push --force-with-lease`. No other action needed — and thanks again for the contribution!

nguiaSoren:
Built this on macOS arm64 and tested it against the 72k-LOC Rust repo from #333 (I'm the #333 reporter / #334 author). The design and the predicate are right, and the unit/IO tests are thorough — but I think the gate is currently **inert in practice**: it never trips, because `committed_nodes` is captured after the node index is already freed.

**Root cause.** `dump_and_persist_hashes` reads the committed count at `pipeline.c:832`, *after* `cbm_gbuf_dump_to_sqlite()` → `release_and_remap_vectors()` → `release_gbuf_indexes()` frees `node_by_qn` (`graph_buffer.c:348`). `cbm_gbuf_node_count()` then returns 0. Edges survive (`edge_by_key` is freed but `edges.count` isn't) — which is exactly why `expected_edges` looks right while `expected_nodes` comes back 0.

**Observed** on a healthy full index:
```
nodes: 13351 | expected_nodes: 0 | expected_edges: 50438
```
With `committed_nodes = 0 ≤ CBM_DUMP_VERIFY_MIN_FLOOR`, `cbm_dump_verify_is_degraded()` always returns false — so even a catastrophically partial index would still report `status: "indexed"`, which is the silent-success this is meant to stop.

Two parts to it:
1. Capture `committed_nodes/edges` **before** the dump.
2. The incremental path (`pipeline_incremental.c`) never sets the committed count at all, so the gate is also disabled on incremental/reparse reindexes — which is the scenario #334 actually reproduces (repeated `index_repository` on an already-present index).

The existing tests pass because they feed the predicate synthetic committed values; an integration test that runs the real pipeline and asserts `committed_nodes > 0` (and `== persisted`) catches it.

I have a patch for all three (capture-before-dump + incremental setter + that integration test) — built clean, full suite green (5,620 passing), DCO-signed. It fails on the current code (`committed_nodes == 0`) and passes after. Happy to open it as a PR (it builds on this branch) or drop the diff here — whatever you and @DeusData prefer.

yangsec888:
@nguiaSoren This is a great catch, thank you for digging in this far.

I traced it through and you're spot on: `dump_and_persist_hashes` reads `cbm_gbuf_node_count()` at `pipeline.c:832`, which is after `cbm_gbuf_dump_to_sqlite()` has already run `release_gbuf_indexes()` and freed `node_by_qn`, so `committed_nodes` is always 0. With `0 <= CBM_DUMP_VERIFY_MIN_FLOOR` the predicate short-circuits to `false`, and the gate never trips. Edges surviving the free is exactly why `expected_edges` looked plausible while `expected_nodes` came back 0, which I should have caught from my own response shape. The incremental path never setting the count is the more important miss, since that's the actual #334 repro.

You're also right about why my tests stayed green — they feed the predicate synthetic committed values, so the pure function is covered but the real pipeline wiring isn't. An integration test that runs the real pipeline and asserts `committed_nodes > 0 && == persisted` is the right guard.

Please go ahead and open the PR on top of this branch — capture-before-dump + incremental setter + the integration test all sound right, and it'd be good to have the test that fails on the current code. I'll happily defer to your fix here. Thanks again for the thorough writeup (and for testing it against the real 72k-LOC repo).

nwn900:
I have the following issue too

DeusData:
Thanks so much for this, @yangsec888 — the `status:degraded` post-dump plausibility gate you designed here is the foundation, and it's fully preserved (with your authorship) in #456, which additionally fixes a capture-ordering bug that would otherwise have left the gate permanently inert, and adds incremental-reindex coverage.

So I'm consolidating the #334 work on #456 and closing this one. Genuinely appreciate you building the gate — it's exactly the silent-success guard #334 needed. 🙏

PR #441, opened 2026-06-11
feat(cross-repo): link Maven library dependencies

### LINKED ISSUE #440: Add cross-repo Maven library dependency links
## Problem

`cross-repo-intelligence` currently links services through HTTP/async/channel style edges, but it does not expose cross-repo library dependency relationships. In a multi-repo Maven platform this makes it hard to answer questions like:

- which services depend on a shared library artifact?
- which service repos consume a shared library artifact?
- where is a shared artifact referenced through direct dependencies or Maven exclusions?

This came up while indexing independent repositories into the same `CBM_CACHE_DIR`; the desired relationship is library usage, not just HTTP calls.

## Proposed behavior

When `index_repository` runs in `cross-repo-intelligence` mode, it should also match Maven artifacts across indexed projects:

- read indexed `pom.xml` files for the source project
- collect Maven dependency coordinates from actual `<dependencies>` blocks
- read indexed `pom.xml` files for target projects
- collect each target project Maven artifact coordinates
- when source dependency coordinates match a target artifact, emit cross-repo library edges

Suggested edge types:

- `CROSS_LIBRARY_DEPENDS_ON` in the consumer/source project
- `CROSS_LIBRARY_USED_BY` in the provider/target project

The output summary and architecture cross-repo summary should include these new edge types.

## Scope

This issue is intentionally scoped to Maven `pom.xml` support. Gradle, npm, Go modules, Python packages, and richer code-level import-to-artifact resolution can be separate follow-up issues.

## Tests

Add focused regression tests that build temporary provider/consumer projects with neutral Maven coordinates and assert library edge creation, valid JSON edge properties, and no false positives from dependency-management-only declarations.

nvt-pankajsharma:
## Summary

Fixes #440

Adds Maven library dependency matching to `cross-repo-intelligence` so independently indexed repos can expose shared-library usage, not only HTTP/async/channel relationships.

This adds:

- Maven artifact/dependency collection from indexed `pom.xml` files
- `CROSS_LIBRARY_DEPENDS_ON` edges in the consuming project
- `CROSS_LIBRARY_USED_BY` edges in the provider project
- cross-repo mode result counts for `cross_library_edges`
- architecture summary support for the new `CROSS_LIBRARY_*` edge types
- pipeline regression tests using neutral sample Maven coordinates

## Scope

This draft is intentionally scoped to Maven `pom.xml` relationships. Gradle/npm/Go/Python package linking and richer source import-to-artifact resolution should be follow-up issues.

## TDD record

- Failing test: `cross_repo_maven_dependency_creates_library_edges`
- Additional failing tests from review loop: `cross_repo_maven_dependency_escapes_library_edge_props`, `cross_repo_maven_dependency_management_does_not_create_library_edge`
- Command intended for red/green loop: `make -f Makefile.cbm test` / `scripts/test.sh`
- Expected red reasons: no Maven library edges initially; then malformed JSON risk for unescaped Maven values; then false positives from dependency-management-only declarations.

## Verification

- `make -j"$(sysctl -n hw.ncpu 2>/dev/null || echo 4)" -f Makefile.cbm build/c/test-runner` passed before the review-loop fixes.
- Timed test-runner captures reached the new Maven cross-repo tests; the library-edge and JSON-escape tests pass after fixes. A dependency-management false-positive test was added and is still being completed in the review loop.
- Full suite was not completed. The runner has no help/filter mode and full execution previously hit unrelated existing probe failures in `tests/test_node_creation_probe.c` / `tests/test_grammar_probe_g.c`.
- `clang-format` is not available on my PATH in this environment. `git diff --check` passed before the initial commit.

## Notes for reviewers

This is marked draft because the contribution guide says new indexing algorithms need maintainer design discussion, and because the patch is currently larger than the preferred <500-line PR guideline. We can split parsing/edge-emission/schema-reporting if maintainers prefer smaller increments.

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -273,7 +273,8 @@ static const tool_def_t TOOLS[] = {
     {"index_repository",
      "Index a repository into the knowledge graph. "
      "Special mode 'cross-repo-intelligence': skip extraction, only match Routes/Channels "
-     "across projects to create CROSS_HTTP_CALLS/CROSS_ASYNC_CALLS/CROSS_CHANNEL edges. "
+     "and Maven library dependencies across projects to create CROSS_HTTP_CALLS/"
+     "CROSS_ASYNC_CALLS/CROSS_CHANNEL/CROSS_LIBRARY_DEPENDS_ON/CROSS_LIBRARY_USED_BY edges. "
      "Requires target_projects param. Ensure target projects have fresh indexes first.",
      "{\"type\":\"object\",\"properties\":{\"repo_path\":{\"type\":\"string\",\"description\":"
      "\"Path to the repository\"},"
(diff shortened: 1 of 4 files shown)


nvt-pankajsharma:
Closing this draft. I will re-open a smaller sanitized proposal after reworking the scope and implementation shape.

PR #442, opened 2026-06-11
feat(cross-repo): link Maven library dependencies

### LINKED ISSUE #440: Add cross-repo Maven library dependency links

nvt-pankajsharma:
## Summary

Fixes #440

Adds Maven artifact dependency matching to `cross-repo-intelligence` so independently indexed repos can expose shared-library usage, not only HTTP/async/channel relationships.

This PR adds:

- `src/pipeline/pass_cross_repo_maven.[h]` for Maven-specific cross-repo matching
- `CROSS_LIBRARY_DEPENDS_ON` edges in the consumer/source project
- `CROSS_LIBRARY_USED_BY` edges in the provider/target project
- cross-repo mode result counts for `cross_library_edges`
- architecture summary support for the new `CROSS_LIBRARY_*` edge types
- regression coverage for Maven dependency usage, plugin/dependency-management/comment exclusions, generated identity collisions, stale cleanup, allocation-growth failure paths, JSON-safe properties, and test environment isolation

## Scope

This is intentionally scoped to direct Maven `pom.xml` artifact dependency relationships. Gradle, npm, Go modules, Python packages, and source import-to-artifact resolution should be follow-up issues.

Maven plugin classpath dependencies are intentionally not treated as service/library usage. Maven `<exclusion>` entries are also intentionally ignored because they describe libraries excluded from transitive resolution, not libraries used by the consuming project.

## TDD record

- `cross_repo_maven_dependency_creates_library_edges`
  - Red reason: no Maven library edges were emitted by `cross-repo-intelligence`; later tightened to prove Maven exclusions do not count as library usage.
- `cross_repo_maven_dependency_escapes_library_edge_props`
  - Red reason: handwritten JSON properties became malformed when Maven values contained quotes.
- `cross_repo_maven_dependency_management_does_not_create_library_edge`
  - Red reason: dependency-management declarations were treated as actual usage dependencies.
- `cross_repo_maven_commented_dependency_does_not_create_library_edge`
  - Red reason: commented-out Maven dependencies were treated as actual usage dependencies.
- `cross_repo_maven_plugin_dependency_does_not_create_library_edge`
  - Red reason: Maven plugin classpath dependencies were treated as service/library usage.
- `cross_repo_maven_cleanup_preserves_unrelated_nodes`
  - Red reason: generated library-node cleanup could delete unrelated nodes with reserved-looking qualified names.
- `cross_repo_maven_provider_rerun_preserves_incoming_used_by`
  - Red reason: rerunning cross-repo matching for a provider deleted valid incoming `CROSS_LIBRARY_USED_BY` links.
- `cross_repo_maven_removed_dependency_clears_provider_used_by`
  - Red reason: removing a dependency cleared consumer-side links but left stale provider-side `CROSS_LIBRARY_USED_BY` links.
- `cross_repo_maven_long_coordinates_do_not_collide`
  - Red reason: distinct Maven coordinates sharing a long prefix collapsed into one persisted library edge/node.
- `cross_repo_maven_fixture_restores_cache_dir`
  - Red/review reason: the Maven fixture changed process-global `CBM_CACHE_DIR` without restoring it for later tests.
- `cross_repo_maven_long_references_do_not_collide`
  - Red reason: long Maven coordinates and POM paths could truncate generated qualified names and merge distinct library nodes.
- `cross_repo_maven_very_long_pom_paths_do_not_truncate`
  - Red reason: distinct POM paths sharing the first 512 bytes collapsed into one generated `Library` node.
- `cross_repo_maven_failed_growth_preserves_capacity`
  - Red/review reason: failed dependency-list growth mutated capacity before allocation success.
- `cross_repo_maven_growth_rejects_byte_overflow_before_realloc`
  - Red reason: byte-size overflow could reach `realloc` and record a larger capacity.
- `cross_repo_project_list_initial_alloc_failure_returns_empty`
  - Red/review reason: wildcard target-project collection did not handle initial allocation failure before writing to the project array.

## Review-fix-loop evidence

Allocation and bounded-growth sweep:

- Class: bounded representation of open-ended counts and byte-size allocation growth.
- Similar patterns checked: Maven dependency-list growth, artifact-list growth, POM path-list growth, and cross-repo project-list initial/growth allocation.
- Fix: capacity and byte-size multiplication are checked before allocation; capacity updates happen only after allocation succeeds; project-list initialization returns an empty result on allocation failure.
- Boundary tests: failed realloc preserves pointer/capacity, byte-size overflow is rejected before allocator invocation, and project-list initial allocation failure clears the output pointer.

Generated identity and cleanup sweep:

- Class: generated identity collision and stale/generated-state cleanup.
- Similar patterns checked: Maven coordinates, POM paths, generated `Library` and `LibraryConsumer` node qualified names, source-side cleanup, provider-side reverse cleanup.
- Fix: generated Maven node identities preserve full inputs; source cleanup removes only generated `Library` nodes and outgoing `CROSS_LIBRARY_DEPENDS_ON`; reverse `CROSS_LIBRARY_USED_BY` cleanup is scoped to the source project that created it.
- Boundary tests: long common-prefix Maven coordinates, very long common-prefix POM paths, provider rerun preserving incoming usage, dependency removal clearing provider usage, and unrelated reserved-prefix function preservation.

False-positive sweep:

- Class: unchecked parse scope / false-positive relationship extraction.
- Similar patterns checked: dependency management, XML comments, Maven plugin dependency blocks, and Maven exclusion metadata.
- Fix: only direct Maven dependency declarations outside non-usage blocks produce library usage links; plugin classpath dependencies and exclusions are ignored.
- Boundary tests: dependency-management, commented dependency, plugin dependency, and exclusion-in-dependency cases do not create false usage links.

## Verification

- `make -f Makefile.cbm test` passed with `5619 passed`.
- Focused Maven/cross-repo regression stream passed the relevant tests, including dependency usage, exclusion ignored as non-usage, JSON escaping, dependency-management exclusion, commented dependency exclusion, and plugin dependency exclusion.
- `make -f Makefile.cbm lint-format CLANG_FORMAT="$(brew --prefix clang-format)/bin/clang-format"` passed locally.
- `scripts/check-dco.sh origin/main..HEAD` passed locally for all six PR commits.
- `git diff --check origin/main...HEAD` passed locally.
- GitHub DCO, security-static, and license-gate passed on the prior signed head; this push reruns the same gates on the updated head.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -200,6 +200,7 @@ PIPELINE_SRCS = \
     src/pipeline/pass_semantic_edges.c \
     src/pipeline/pass_complexity.c \
     src/pipeline/pass_cross_repo.c \
+    src/pipeline/pass_cross_repo_maven.c \
     src/pipeline/artifact.c \
     src/pipeline/pass_pkgmap.c
 
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -1815,14 +1815,10 @@ static void append_cross_repo_summary(yyjson_mut_doc *doc, yyjson_mut_val *root,
     /* Scan edge types for any CROSS_* edges and sum them */
     int cross_total = 0;
     yyjson_mut_val *cr = yyjson_mut_obj(doc);
-    static const char *cross_types[] = {"CROSS_HTTP_CALLS",
-                                        "CROSS_ASYNC_CALLS",
-                                        "CROSS_CHANNEL",
-                                        "CROSS_GRPC_CALLS",
-                                        "CROSS_GRAPHQL_CALLS",
-                                        "CROSS_TRPC_CALLS",
-                                        "CROSS_LIBRARY_DEPENDS_ON",
-                                        "CROSS_LIBRARY_USED_BY"};
+    static const char *cross_types[] = {
+        "CROSS_HTTP_CALLS",         "CROSS_ASYNC_CALLS",    "CROSS_CHANNEL",
+        "CROSS_GRPC_CALLS",         "CROSS_GRAPHQL_CALLS",  "CROSS_TRPC_CALLS",
+        "CROSS_LIBRARY_DEPENDS_ON", "CROSS_LIBRARY_USED_BY"};
     for (int t = 0; t < (int)(sizeof(cross_types) / sizeof(cross_types[0])); t++) {
         for (int i = 0; i < schema->edge_type_count; i++) {
             if (strcmp(schema->edge_types[i].type, cross_types[t]) == 0) {
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -273,7 +273,8 @@ static const tool_def_t TOOLS[] = {
     {"index_repository",
      "Index a repository into the knowledge graph. "
      "Special mode 'cross-repo-intelligence': skip extraction, only match Routes/Channels "
-     "across projects to create CROSS_HTTP_CALLS/CROSS_ASYNC_CALLS/CROSS_CHANNEL edges. "
+     "and Maven library dependencies across projects to create CROSS_HTTP_CALLS/"
+     "CROSS_ASYNC_CALLS/CROSS_CHANNEL/CROSS_LIBRARY_DEPENDS_ON/CROSS_LIBRARY_USED_BY edges. "
      "Requires target_projects param. Ensure target projects have fresh indexes first.",
      "{\"type\":\"object\",\"properties\":{\"repo_path\":{\"type\":\"string\",\"description\":"
      "\"Path to the repository\"},"
@@ -282,7 +283,8 @@ static const tool_def_t TOOLS[] = {
      "\"default\":\"full\",\"description\":\"All modes run type-aware LSP call/usage "
      "resolution (per-file + cross-file). full: all files + similarity/semantic edges. "
      "moderate: filtered files + similarity/semantic. fast: filtered files, no "
-     "similarity/semantic. cross-repo-intelligence: match Routes/Channels across projects.\"},"
+     "similarity/semantic. cross-repo-intelligence: match Routes/Channels and Maven library "
+     "dependencies across projects.\"},"
      "\"target_projects\":{\"type\":\"array\",\"items\":{\"type\":\"string\"},"
      "\"description\":\"Projects to search for cross-repo links (cross-repo-intelligence mode). "
      "Use [\\\"*\\\"] for all indexed projects. Run list_projects to see available projects.\"},"
@@ -1813,9 +1815,14 @@ static void append_cross_repo_summary(yyjson_mut_doc *doc, yyjson_mut_val *root,
     /* Scan edge types for any CROSS_* edges and sum them */
     int cross_total = 0;
     yyjson_mut_val *cr = yyjson_mut_obj(doc);
-    static const char *cross_types[] = {"CROSS_HTTP_CALLS",    "CROSS_ASYNC_CALLS",
-                                        "CROSS_CHANNEL",       "CROSS_GRPC_CALLS",
-                                        "CROSS_GRAPHQL_CALLS", "CROSS_TRPC_CALLS"};
+    static const char *cross_types[] = {"CROSS_HTTP_CALLS",
+                                        "CROSS_ASYNC_CALLS",
+                                        "CROSS_CHANNEL",
+                                        "CROSS_GRPC_CALLS",
+                                        "CROSS_GRAPHQL_CALLS",
+                                        "CROSS_TRPC_CALLS",
+                                        "CROSS_LIBRARY_DEPENDS_ON",
+                                        "CROSS_LIBRARY_USED_BY"};
     for (int t = 0; t < (int)(sizeof(cross_types) / sizeof(cross_types[0])); t++) {
         for (int i = 0; i < schema->edge_type_count; i++) {
             if (strcmp(schema->edge_types[i].type, cross_types[t]) == 0) {
@@ -2462,7 +2469,7 @@ static char *handle_cross_repo_mode(const char *repo_path, const char *args) {
     yyjson_doc_free(jdoc);
 
     int total = result.http_edges + result.async_edges + result.channel_edges + result.grpc_edges +
-                result.graphql_edges + result.trpc_edges;
+                result.graphql_edges + result.trpc_edges + result.library_edges;
     yyjson_mut_doc *doc = yyjson_mut_doc_new(NULL);
     yyjson_mut_val *root = yyjson_mut_obj(doc);
     yyjson_mut_doc_set_root(doc, root);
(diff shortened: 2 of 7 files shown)


DeusData:
Heads-up: this project now validates every PR automatically — tests, lint, security/license gates, and DCO sign-off ([CONTRIBUTING.md](https://github.com/DeusData/codebase-memory-mcp/blob/main/CONTRIBUTING.md)). Your branch predates this, so CI will flag the missing `Signed-off-by` trailers. Fixing it is one command: `git rebase --signoff origin/main && git push --force-with-lease`. No other action needed — and thanks again for the contribution!

nvt-pankajsharma:
Please let me know if anything need to be done on this from my side.

DeusData:
Thanks for checking in. The remaining action from your side is still the DCO / rebase cleanup from the earlier maintainer note: please rebase on current `main` with sign-off trailers and push the updated branch.

A typical sequence is:

```bash
git fetch origin
git rebase --signoff origin/main
git push --force-with-lease
```

After that, CI/DCO should tell us whether anything else is needed before review.

DeusData:
Thanks for the Maven cross-repo dependency work. This is in a sensitive indexing path, so please add a short design note covering the new edge semantics, limits on POM parsing, and expected behavior for dependencyManagement/plugins/parent POMs. Since `Makefile.cbm` and cross-repo output fields change, this needs explicit maintainer review before merge.

DeusData:
Thank you for this, and I am sorry it took seven weeks to come back to you with a real answer. You deserve the reasoning rather than a verdict, so this is longer than a close usually is.

**The decision: we are not taking this, and the blocker is a maintainer-level one about the graph's vocabulary — not about the quality of your effort.**

## The direction question, which was ours to answer and we did not

Adding `CROSS_LIBRARY_DEPENDS_ON` and `CROSS_LIBRARY_USED_BY` means deciding that **manifest-derived, project-level edges belong in a graph of code**, and that synthetic `Library` / `LibraryConsumer` nodes — which have no source location, `start_line = 0`, and return nothing meaningful from `get_code_snippet` — are an acceptable new node *class*. That is a permanent decision, and it was never put to the maintainer before you wrote 1742 lines. That is our process failure, not yours.

There is a real case for it, and I want to state it since I am declining: six `CROSS_*` edge kinds already exist, so the door was open rather than being opened; we already parse `pom.xml` in `pass_pkgmap.c`; your implementation is opt-in and mode-gated, so it costs nothing to anyone who does not ask for it; and in a Maven platform the library edge is often the only structural link between two repos. The case against is that `mvn dependency:tree` answers this correctly and completely today, with full parent, BOM and property resolution.

## What the review found in the code, because you will want to know

**The one I would most want to be told about.** `collect_dependencies` takes the `<dependency>`…`</dependency>` span and reads the **first `<groupId>` inside it**. `<exclusions>` also live inside that span. Your PR body says exclusions are intentionally ignored, and they are — but by element ordering rather than by structure. `maven-4.0.0.xsd` models `Dependency` with `xs:all`, so order is not significant, and a perfectly valid POM that writes `<exclusions>` before `<groupId>` produces a dependency edge pointing at the **excluded** artifact. Your tests pass because the fixture writes them in the common order. It is a subtle one and it took a careful read to find.

**Version is never compared.** `artifact_matches_dep` matches `groupId:artifactId` only, and `mcr_artifact_t` has no version field, so a consumer on `1.0` links to a provider whose POM says `9.9.9` — and if a fork and the original are both indexed, both get edges.

**And the two relationships #440 actually asked about are the two that produce nothing:** parent-POM inheritance is not implemented (`<parent>` never matches the `<dependency` scan), and BOM imports are excluded wholesale along with `dependencyManagement`. So "which repos consume our shared artifact?" returns empty in both of the common shapes.

**Main also moved 1101 commits under you.** `cbm_cross_repo_match` is now a thin wrapper over `cbm_cross_repo_match_cancellable`; `delete_cross_edges` became cancellation-aware and returns a status; and the cross-edge vocabulary now lives at four sites rather than one — including `resolve_trace_edge_types`, without which the new edges would be invisible to `trace_path`'s cross-service mode. A rebase here is a re-implementation, not a conflict resolution, which is why I am not asking you to do one.

## What was genuinely good

**Your tests are better than most we receive.** They drive the real entry point against two real on-disk SQLite stores with real nodes and assert edge counts by direct SQL — they fail at the link step without the pass, rather than asserting on an intermediate array. That is the shape we ask for and rarely get.

**The QN-sizing work is careful**: `snprintf(NULL, 0, …)` then an exact allocation, so long coordinates and long POM paths cannot truncate into a shared qualified name, with four tests binding on precisely that. **Multi-module aggregator handling is right** — collecting every `pom.xml` and falling back to the `<parent>` groupId for modules that omit their own is correct.

**And your allocation hardening fixed a real bug.** At your merge base, `projects[count] = malloc(...)` was unchecked and `memcpy` wrote into it regardless. You spotted that. I went to distill it into its own commit with your co-authorship — and found main had fixed it independently in the meantime, switching to `cbm_strdup` with a null check. Your overflow guards turn out to be unreachable too, because `count >= CR_MAX_PROJECTS` breaks first and caps the capacity at 4096. So there is nothing left for me to land on your behalf, which I regret, because you were right.

## What happens now

**#440 stays open**, re-framed as a direction question rather than a ready-to-implement task, so the decision gets made before anyone writes code against it again. If the answer turns out to be yes, your analysis of the problem shape is the best starting material we have, and I will come back to this PR and to you.

I am sorry that seven weeks of silence ended in a decline. You asked on 1 July whether anything was needed from you and you deserved a straight answer then. Thank you for the work and for the patience.

nvt-pankajsharma:
I am glad to serve my friend. This is a MCP I am using when it was 1k stars. And I really want to grow with it. Was a bit busy also. Let me know or assign any issue you want me to look into. I will try my best.

PR #448, opened 2026-06-12
fix(discover): deploy/ directories indexed as duplicate source in full mode

isc-tdyar:
**deploy/ directories indexed as duplicate source in full mode, producing stale duplicate Class/Method rows**

Discovered while querying a project that uses a `deploy/` directory for build output. Queries like `MATCH (c:Class {name:'Foo'})` returned multiple rows for the same class — one from the source file, one from the copy in `deploy/`.

**Root cause**

`deploy` and `deployed` were in `FAST_SKIP_DIRS`, which is only consulted in `fast` and `moderate` indexing modes. In the default `full` mode, `FAST_SKIP_DIRS` is not applied, so `deploy/` was walked and indexed alongside source files. Any class that existed in both locations got duplicate rows in the graph with potentially stale data from the build copy.

```c
// before: deploy/deployed only skipped in fast/moderate mode
static const char *FAST_SKIP_DIRS[] = { ..., ".vercel", ".netlify", NULL };

// after: always skipped
static const char *ALWAYS_SKIP_DIRS[] = { ..., ".vercel", ".netlify", "deploy", "deployed", NULL };
```

**Fix**

Move `deploy` and `deployed` from `FAST_SKIP_DIRS` to `ALWAYS_SKIP_DIRS` so they are excluded in all indexing modes.

**Regression test**

`discover_deploy_excluded_full_mode` in `tests/test_discover.c` — creates a temporary repo with `src/`, `deploy/`, and `deployed/` subdirectories, runs discovery in full mode, and asserts only the file under `src/` is returned (deploy and deployed are excluded).

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -42,7 +42,7 @@ static const char *ALWAYS_SKIP_DIRS[] = {
     ".cargo", ".stack-work", ".dart_tool", "zig-cache", "zig-out", ".metals", ".bloop", ".bsp",
     ".ccls-cache", ".clangd", "elm-stuff", "_opam", ".cpcache", ".shadow-cljs",
     /* Deploy */
-    ".vercel", ".netlify",
+    ".vercel", ".netlify", "deploy", "deployed",
     /* Misc */
     ".qdrant_code_embeddings", ".tmp", "vendor", "vendored", NULL};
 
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the contribution — all gates are green including DCO, and the deploy/-directory duplication is a good catch. Will review when time allows.

PR #449, opened 2026-06-12
fix(extract): Class.lines always 0 for all languages

isc-tdyar:
**Class.lines always 0 for all languages**

Discovered while trying to sort classes by size. Every class node had `lines = 0` regardless of actual length.

**Root cause**

In `extract_class_def()`, `start_line` and `end_line` are correctly set from tree-sitter node positions, but `def.lines` is never computed — it stays 0 from the `memset` that zeroes the struct. The equivalent computation exists and is correct in `push_method_def()`:

```c
// push_method_def — correct
def.lines = (int)(def.end_line - def.start_line + TS_LINE_OFFSET);

// extract_class_def — missing this line entirely
// def.lines was always 0
```

**Fix**

Add the same one-liner to `extract_class_def()` after `end_line` is set. Affects all languages — any tree-sitter grammar that produces class nodes goes through this path.

**Regression test**

`tool_qg_class_lines_nonzero` in `tests/test_incremental.c` — indexes a project in full mode and asserts that at least one class node has `lines > 0` via a `query_graph` tool call. Before the fix this assertion always failed.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -3323,6 +3323,7 @@ static void extract_class_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec
     def.file_path = ctx->rel_path;
     def.start_line = ts_node_start_point(node).row + TS_LINE_OFFSET;
     def.end_line = ts_node_end_point(node).row + TS_LINE_OFFSET;
+    def.lines = (int)(def.end_line - def.start_line + TS_LINE_OFFSET);
     def.is_exported = cbm_is_exported(name, ctx->language);
     def.base_classes = extract_base_classes(a, node, ctx->source, ctx->language);
     def.decorators = extract_decorators(a, node, ctx->source, ctx->language, spec);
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the contribution — CI is fully green here too. `Class.lines` always being 0 is exactly the kind of quiet quality bug worth fixing; will review when time allows.

DeusData:
Thanks @isc-tdyar — this one reviewed clean and is ready to merge, but it now conflicts with `main` (it shares `tests/test_incremental.c` with a just-merged PR). Could you rebase on latest `main`? I'll merge as soon as it's conflict-free. 🙏

DeusData:
Thanks @isc-tdyar — the one-line `Class.lines` fix is correct, but the diff is now stale against `main` (conflicting; the extraction files moved under recent changes). Could you rebase on latest `main`? Worth double-checking while you're there that the class-`lines` path isn't already covered upstream. I'll merge once it's rebased. 🙏

DeusData:
Thanks for catching this, @isc-tdyar — a clean root-cause fix. `extract_class_def` was the one definition-emit site that never set `def.lines`, so every Class node reported `body_lines = 0` across all languages while the seven sibling sites computed it. Your one-line change lines up exactly with those siblings, and the `MATCH (c:Class) WHERE c.lines > 0` regression test locks it in. Appreciate the careful, minimal fix — merging now. 🙏

PR #456, opened 2026-06-13
fix(pipeline): make the #334 plausibility gate fire (capture committed_nodes before dump)

### LINKED ISSUE #334: Silent index corruption after rapid kill/restart cycles

nguiaSoren:
## Summary

Builds on #433 (the `status:degraded` plausibility gate by @yangsec888) and fixes a capture-ordering bug that left the gate **inert** — it never tripped, so a catastrophically partial index would still report `status: "indexed"`, the exact silent-success #334 set out to stop. @yangsec888 reviewed the finding and deferred to this fix (https://github.com/DeusData/codebase-memory-mcp/pull/433#issuecomment-4696991094).

Closes #334. Supersedes #433 (its gate commit is included here with authorship preserved).

## The bug

`dump_and_persist_hashes` read the committed node count at `pipeline.c` **after** `cbm_gbuf_dump_to_sqlite()`, which runs `release_and_remap_vectors()` → `release_gbuf_indexes()` and frees `node_by_qn` (`graph_buffer.c:348`). `cbm_gbuf_node_count()` therefore returned `0`. With `committed_nodes = 0 ≤ CBM_DUMP_VERIFY_MIN_FLOOR`, `cbm_dump_verify_is_degraded()` short-circuits to `false`.

Edges survive the free (`edge_by_key` is freed, but `edges.count` is not), which is why a healthy full index reported a plausible `expected_edges` while `expected_nodes` came back `0`:

```
nodes: 13351 | expected_nodes: 0 | expected_edges: 50438
```

Separately, the incremental path (`pipeline_incremental.c`) never set the committed count at all, so the gate was also disabled on incremental/reparse reindexes — the scenario #334 actually reproduces (repeated `index_repository` on an already-present index).

## The fix

1. **`pipeline.c`** — capture `committed_nodes`/`committed_edges` *before* the dump.
2. **`pipeline_incremental.c`** + **`pipeline_internal.h`** — record committed counts on the incremental path via a new `cbm_pipeline_set_committed_counts()` (the incremental unit cannot see the opaque `cbm_pipeline` struct).
3. **`tests/test_pipeline.c`** — `pipeline_committed_counts_match_persisted`: drives the real pipeline and asserts `committed_nodes > 0` and `== persisted`. The existing `dump_verify` tests feed the predicate synthetic values, so they cover the pure function but not the pipeline wiring; this test fails on the pre-fix code and passes after.

## Verification

- Built clean on macOS arm64 (`scripts/build.sh`).
- After the fix, a healthy full index reports `expected_nodes == nodes` on **both** the full and incremental paths (was `0` before).
- Full suite green: `scripts/test.sh` → 5,620 passing, including the new test.
- The new test fails on the current code (`committed_nodes == 0`) and passes after, so it genuinely guards the regression.

All commits are DCO signed-off.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -111,7 +111,8 @@ FOUNDATION_SRCS = \
     src/foundation/compat_regex.c \
     src/foundation/mem.c \
     src/foundation/diagnostics.c \
-    src/foundation/profile.c
+    src/foundation/profile.c \
+    src/foundation/dump_verify.c
 
 # Existing extraction C code (compiled from current location)
 EXTRACTION_SRCS = \
@@ -295,7 +296,8 @@ TEST_FOUNDATION_SRCS = \
     tests/test_str_intern.c \
     tests/test_log.c \
     tests/test_str_util.c \
-    tests/test_platform.c
+    tests/test_platform.c \
+    tests/test_dump_verify.c
 
 TEST_EXTRACTION_SRCS = \
     tests/test_extraction.c \
@@ -313,7 +315,8 @@ TEST_STORE_SRCS = \
     tests/test_store_arch.c \
     tests/test_store_bulk.c \
     tests/test_store_pragmas.c \
-    tests/test_store_checkpoint.c
+    tests/test_store_checkpoint.c \
+    tests/test_dump_verify_io.c
 
 TEST_CYPHER_SRCS = \
     tests/test_cypher.c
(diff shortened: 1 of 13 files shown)


nguiaSoren:
Friendly ping on this — CI is green across all platforms (tests, lint, security, DCO) and @yangsec888 has signed off on the approach. Happy to rebase or adjust anything if it'd help it land. No rush, and thanks!

nguiaSoren:
Glad you're back to these! Whenever you have a moment, #456 is green and ready — it fixes the inert-gate issue in the #334 plausibility check. Thanks!

DeusData:
Thanks @nguiaSoren — this is the version we want for #334: it fixes the inert-gate bug (capturing `committed_nodes` before the dump frees the index) and adds incremental-path coverage, and it reviewed clean. It currently conflicts with `main`, though — could you rebase on latest `main`? I'll merge as soon as it's conflict-free. 🙏

nguiaSoren:
Rebased on latest `main`, conflict-free now (only overlap was an include line). CI re-running; should be green shortly. Thanks!

PR #461, opened 2026-06-13
feat(perl-lsp): Perl LSP-tier semantic resolution (Closes #459)

### LINKED ISSUE #459: Add Perl LSP-tier semantic resolution (perl_lsp.c)
## Summary

Perl is already a **first-class parsed language** in CBM — `CBM_LANG_PERL` is fully registered in `internal/cbm/lang_specs.c` (subroutine/use/require/call/branch/var node-type mappings + `tree_sitter_perl`), so `.pl`/`.pm` files are structurally indexed today (definitions, calls, imports, branches, variables).

What Perl **lacks** is the second tier: a hand-written **type-aware LSP resolution module** (`internal/cbm/lsp/perl_lsp.c`). Today 10 languages have one — Go, C/C++/CUDA, PHP, Python, JS/TS/TSX, C#, Java, Kotlin, Rust — dispatched in `cbm.c` (`cbm_run_*_lsp`). Perl has no module and no dispatch branch, so Perl call/usage edges are resolved by tree-sitter + textual heuristics only, with no package/method/inheritance resolution.

This issue proposes adding `perl_lsp.c` to promote Perl from the structural tier to the semantic-resolution tier.

## Motivation

Without LSP-tier resolution, Perl graphs miss the edges that matter most for navigation and impact analysis:

- `$obj->method()` / `Class->method()` / `$self->method()` calls don't resolve to the defining subroutine.
- Inheritance via `@ISA` / `use parent` / `use base` is invisible, so inherited-method calls are unresolved.
- `use Foo::Bar` / `require` don't link to the target package's symbols across files.
- `Exporter`-imported subs aren't mapped to their origin.

Perl remains widely used in systems tooling, bioinformatics, and legacy enterprise codebases — exactly the "large, hard-to-navigate" codebases CBM targets.

## Proposed scope (full parity with the PHP module)

`php_lsp.c` is the closest analog — dynamic, package/namespace-based, OOP-by-convention — and is the model to mirror:

- **Package resolution:** `package Foo::Bar;`, `use`/`require` → cross-file package symbol tables.
- **OOP method resolution:** `$self->m`, `Class->m`, `$obj->m` on blessed references; method-resolution order via `@ISA`, `use parent`, `use base`.
- **Subroutine resolution:** within-package and cross-package calls; `Exporter` (`@EXPORT`/`@EXPORT_OK`) import maps.
- **Expression/type tracking:** sigil-aware variable tracking (`$`/`@`/`%`); blessed-ref type tracking through assignments (`my $x = Foo->new`); recursion-guarded expression-type evaluation, mirroring `php_eval_expr_type`.
- **Builtins:** `perlfunc` core-builtin awareness; optional common-CPAN stdlib seed (mirrors the per-language `lsp/generated/*_stdlib_data.c` pattern).

## Implementation outline

1. `internal/cbm/lsp/perl_lsp.{c,h}` — `PerlLSPContext` mirroring `PHPLSPContext`; reuse shared infra (`type_rep.h`, `scope.h`, `type_registry.h`, `CBMLSPDef`).
2. Public entry `cbm_run_perl_lsp(a, result, source, source_len, root)`.
3. Dispatch branch in `cbm.c` (`if (language == CBM_LANG_PERL) cbm_run_perl_lsp(...)`).
4. Tests: `tests/test_perl_lsp.c` (+ `Makefile.cbm` wiring), following `test_php_lsp.c`.
5. Validate against a real Perl project before submission (per CONTRIBUTING).

## Notes

- This is the LSP/semantic tier only; grammar/parsing support already exists, so this is an **enhancement** to existing Perl coverage rather than net-new language onboarding.
- Reference modules: `internal/cbm/lsp/php_lsp.c` (primary), `py_lsp.c` (dynamic-typing patterns).

halindrome:
## Summary

Adds tier-2 semantic LSP resolution for Perl (`internal/cbm/lsp/perl_lsp.c`), bringing Perl to parity with the existing PHP module. Perl was already registered at the structural tier; this adds type-aware call/method/inheritance resolution.

Closes #459

## What's included

- **`perl_lsp.{c,h}`** — two-pass resolver: package/`use`/`require` collection; `$self`/`$class` invocant binding; `bless` typing (literal + inferred); `@ISA` / `use parent` / `use base` MRO; `SUPER::` dispatch; static `Package::sub()`; typed `$obj->method` / `Class->method` / `$self->method` dispatch; exported functions (Exporter `use Mod qw(...)`); `ClassName->new` → `ClassName`. Includes an `eval_depth` recursion guard (cap 8) for expression typing and a walk-depth guard (cap 512) on the AST walkers.
- **`generated/perl_stdlib_data.c`** — perlfunc builtins + common CPAN OOP modules (Scalar::Util, List::Util, Carp, POSIX, Storable, Data::Dumper).
- **Dispatch wiring** — `CBM_LANG_PERL` in `cbm.c`, unity build in `lsp_all.c`, `Makefile.cbm`.
- **Structural-tier edge surfacing** — `method_call_expression` added to `perl_call_types` (`lang_specs.c`) + `Pkg::sub` short-name normalization in the pipeline bridge, so resolved calls appear as graph CALLS edges.
- **`tests/test_perl_lsp.c`** — 13 resolution tests mirroring `test_php_lsp.c`, plus a deep-nesting crash-safety test (`lsp_perl_deep_expression_no_crash`) and pipeline normalization regression tests.

## Design notes

- Parity model: `php_lsp.c` (closest analog — dynamic, package/namespace, OOP-by-convention).
- **Zero-edge guarantee:** unresolvable/untyped receivers emit **no** edges (false edges are worse than missing edges).

## Validation

- `scripts/build.sh` green; `scripts/test.sh`: all 13 `perllsp_*` tests pass; full suite 5619 passed.
- The only suite failure is `cli_hook_gate_script_no_predictable_tmp_issue384` (`tests/test_cli.c`), an environmental/sandbox pre-existing failure unrelated to this PR (touches no CLI/hook code).
- `clang-format` clean on all introduced code.
- Real-project validation: indexed a ~103K-LOC multi-package Perl distribution (134 files) → 1178 resolved Perl edges surfacing as graph CALLS edges; untyped receivers produce none.

## Status

Opened as a **draft** — undergoing QA rounds (QA reports posted as PR comments). Will be marked ready for review only after QA completes.


halindrome:
## QA Round 1

Fresh read-only reviewer, contract-verified against issue #459. Result: **0 critical, 1 major, 4 minor.** All findings were addressed in commit `fd8e728` (`fix(perl-lsp): address QA round 1`).

### Finding 1 — CPAN Exporter import map `::` vs `.` QN mismatch
- **Area:** `internal/cbm/lsp/perl_lsp.c` (`perl_collect_qw_imports`, import lookup) + `internal/cbm/lsp/generated/perl_stdlib_data.c`
- **Risk:** import targets were built as `Scalar::Util::blessed` (colon form) while the stdlib registry keys are dotted (`Scalar.Util.blessed`) with exact-match lookup, so seeded CPAN exported subs never resolved via the import map. The bridging helper `perl_pkg_to_dot` existed but was dead-coded.
- **Severity:** minor · **Status:** confirmed
- **Fix:** wired `perl_pkg_to_dot` to dot the module portion; removed the dead `(void)` cast. New test `perllsp_cpan_exported_function`. Zero-edge guarantee preserved.

### Finding 2 — No recursion-depth guard / deep-nesting crash test (parity gap)
- **Area:** `internal/cbm/lsp/perl_lsp.c` (`perl_resolve_calls_in_node`, `perl_pass1_scan`) + `tests/test_stack_overflow.c`
- **Risk:** the Perl AST walkers recursed per child with no depth counter — same structure as paths with documented production SIGSEGVs (#424/#427). Java/C++ have `lsp_*_deep_*_no_crash` tests (depth 30000); Perl had neither.
- **Severity:** major · **Status:** hypothetical
- **Fix:** added `CBM_LSP_PERL_MAX_WALK_DEPTH=512` + a `walk_depth` counter guarding both walkers (subtree skipped past the cap — graceful degradation, never a wrong edge), mirroring the Java pattern. New test `lsp_perl_deep_expression_no_crash` (DEPTH=30000).

### Finding 3 — Shared `::` normalization affects all `::`-languages
- **Area:** `src/pipeline/lsp_resolve.h` (`cbm_pipeline_find_lsp_resolution`)
- **Risk:** the `Pkg::sub` last-segment normalization lives in shared pipeline code; it only widens matching (cannot drop edges) but shipped untested for non-Perl, with a theoretical cross-namespace mis-attribution risk.
- **Severity:** minor · **Status:** confirmed (behavior change), hypothetical (mis-attribution)
- **Fix:** kept the shared normalization (it legitimately serves all `::` languages) and added regression tests `lsp_resolve_qualified_static_call_normalizes_colons` and `lsp_resolve_misattribution_is_bounded` (mis-attribution bounded by caller-QN equality + confidence floor).

### Finding 4 — `SUPER::` advertised but unimplemented
- **Area:** `internal/cbm/lsp/perl_lsp.h` (`enclosing_parent_qn`), `internal/cbm/lsp/perl_lsp.c`
- **Risk:** the header advertised a `SUPER::` field that was never assigned/read; `$self->SUPER::method()` resolved to nothing.
- **Severity:** minor · **Status:** confirmed
- **Fix:** **implemented** `SUPER::` dispatch — `process_package_decl` records the first `@ISA` parent into `enclosing_parent_qn`; `perl_resolve_method_call` strips `SUPER::` and resolves the bare method starting at the parent (skipping any child override), strategy `perl_method_super`. New tests `perllsp_super_dispatch`, `perllsp_super_no_parent_no_edge`. Zero-edge guarantee preserved.

### Finding 5 — Stale pre-existing-failure note in PR body
- **Risk:** PR body named `search_code_multi_word` as the unrelated failure; after rebasing onto current `upstream/main` that test is fixed upstream.
- **Severity:** minor · **Status:** confirmed
- **Fix:** PR description updated (see edited body) — the only remaining suite failure is the environmental `cli_hook_gate_script_no_predictable_tmp_issue384` (`tests/test_cli.c`), unrelated to Perl.

### Contract verification (issue #459)
All acceptance criteria met after fixes. Package/`use`/`require` resolution, OOP method/MRO (`@ISA`/`use parent`/`use base`, now also `SUPER::`), subroutine + Exporter import resolution (CPAN seed now functional), bless typing, recursion-guarded eval (cap 8) plus walker depth guard, perlfunc + CPAN stdlib seed, the test suite (now 13 `perllsp_*` tests), real-project validation (134 files / ~103K LOC → 1178 resolved edges as graph CALLS), and zero false edges (enforced + negative tests).

### Summary
| Severity | Found | Fixed |
|---|---|---|
| Critical | 0 | 0 |
| Major | 1 | 1 |
| Minor | 4 | 4 |
| **Total** | **5** | **5** |

Post-fix: build green, `clang-format` clean, 13/13 `perllsp_*` tests pass, no regressions across the 5619-test suite.

---
*QA performed by Claude Code (claude-opus-4-8). Reviewer was a fresh read-only agent; findings fixed in `fd8e728`.*

halindrome:
## Update: CI fixes (DCO + Windows segfault)

Two CI failures from the initial push are addressed in this revision:

**1. `test / test-windows` segfault — fixed.**
A Windows-only SIGSEGV occurred during parsing (before any extractor ran), not in the Perl LSP code. Root cause: deeply nested, grammar-*ambiguous* input (Perl's optional-paren calls in an `f(f(f(...)))` chain) drove tree-sitter's GLR ambiguity-merge (`stack_node_add_link` in the vendored `ts_runtime/src/stack.c`) to recurse once per nesting level (~260 B/frame), overflowing Windows' ~1 MB default thread stack. Java/C++ survive identical nesting only because their grammars are unambiguous there.

Fix (`fix(ts-runtime): bound GLR stack merge recursion...`): cap the recursive merge at `CBM_TS_STACK_MERGE_MAX_DEPTH = 512`, with graceful degradation identical to the runtime's existing `MAX_LINK_COUNT` bail-out — the parse still yields a valid tree (never a wrong one), and the zero-edge guarantee is preserved. 512 frames ≈ 130 KB, safe on a 1 MB stack while far exceeding any realistic source nesting. Reproduced and verified locally by running extraction on a 1 MB-stack thread (mimicking Windows): crashes before, passes after, at depths 5k/10k/30k. The `lsp_perl_deep_expression_no_crash` regression test now runs on an explicit small (Windows-like) thread stack so the regression is caught even on 8 MB-default hosts.

This touches the vendored tree-sitter runtime (already CBM-patched elsewhere) because the overflow is in shared parser code that the Perl grammar's ambiguity exposes; it protects all grammars, not just Perl.

**2. `dco` — fixed.** All commits are now `Signed-off-by`.

Build green; `clang-format` clean; all 13 `perllsp_*` tests and the full `stack_overflow` suite pass.

halindrome:
## QA Round 2

Fresh read-only reviewer (independent of Round 1), contract-verified against issue #459, reviewing the post-Round-1 state (`fd8e728`). Result: **0 critical, 0 major, 3 minor** — all zero-edge-safe. All four Round-1 fixes verified correct.

### Round-1 fix verification
| Fix | Verified? | Notes |
|---|---|---|
| F1 — CPAN Exporter import map | yes | `perl_pkg_to_dot` correctly converts `::`→`.` for arbitrary nesting (`Foo::Bar::Baz`→`Foo.Bar.Baz`), wired into `perl_collect_qw_imports`; no double-dotting; dead `(void)` cast removed. `perllsp_cpan_exported_function` asserts the dotted QN. |
| F2 — walker depth guard | yes | `CBM_LSP_PERL_MAX_WALK_DEPTH=512`; cap check before increment, no leak/double-decrement, early returns safe, `walk_depth` zero-initialized; cap skips subtree (graceful degradation, no wrong edge). `lsp_perl_deep_expression_no_crash` (DEPTH 30000) passes. |
| F3 — shared `::` normalization tests | yes | `lsp_resolve_qualified_static_call_normalizes_colons` and `lsp_resolve_misattribution_is_bounded` genuinely exercise the confidence-floor / caller-QN-equality / highest-confidence-wins logic. |
| F4 — `SUPER::` dispatch | partial | Correct for single inheritance (lookup starts at parent, child override skipped; no-parent / outside-method → no edge). **Partial** due to Findings 1 & 2 below — both zero-edge-safe (under-resolution / test-strength), not wrong-edge bugs. |

### Finding 1: `SUPER::` follows only the first `@ISA` parent — incomplete under multiple inheritance
- **Area:** `internal/cbm/lsp/perl_lsp.c` (`process_package_decl`, `perl_resolve_method_call`)
- **Risk:** with `our @ISA = ('A', 'B')` and a method defined only in `B`, Perl's DFS MRO would resolve `SUPER::method` to `B::method`, but only the first parent (`A`) is recorded/searched → no edge emitted. Under-resolution (missing edge), never a false edge; single inheritance (the common case) is fully correct.
- **Severity:** minor · **Status:** confirmed · **Disposition:** OPEN

### Finding 2: `SUPER::` test cannot prove the child override is actually skipped
- **Area:** `tests/test_perl_lsp.c` (`perllsp_super_dispatch`)
- **Risk:** sub QNs are emitted as `module_qn.subname` (package not woven in), so `Base::greet` and `Child::greet` both collapse to `test.main.greet`; the substring-based assertion can't distinguish parent-vs-child. The code is correct, but a regression that wrongly resolved to the child's own sub would still pass. Same limitation affects the existing inheritance tests. The `perllsp_super_no_parent_no_edge` zero-edge test is sound.
- **Severity:** minor · **Status:** confirmed · **Disposition:** OPEN (test-strength)

### Finding 3: coarse substring matching in test helpers (advisory)
- **Area:** `tests/test_perl_lsp.c` (`extract_perl`, `find_resolved*` helpers)
- **Risk:** helpers match callee fragments by substring; F1's assertion is specific enough to be a genuine check, but the suite is generally permissive. No defect; noted only.
- **Severity:** minor · **Status:** confirmed · **Disposition:** no action required

### Contract verification (issue #459)
All acceptance criteria met. MRO via `@ISA`/`use parent`/`use base` is complete for general method lookup (full @ISA frontier walk); only the `SUPER::` path is limited to the first parent (Finding 1) — and `SUPER::` is not an explicit contract requirement. Zero-edge guarantee holds on every traced path.

### Summary
| Severity | Count |
|---|---|
| Critical | 0 |
| Major | 0 |
| Minor | 3 |
| **Total** | **3** |

Build green; all 13 `perllsp_*` tests pass; no regressions across the 5619-test suite. The 3 minor findings are zero-edge-safe and remain OPEN for a follow-up round.

---
*QA performed by Claude Code (claude-opus-4-8). Fresh read-only reviewer; reviewed `fd8e728`. Note: the `ts-runtime` GLR-merge stack-overflow fix landed afterward in `e0b57a0` and will be covered in the next round.*

halindrome:
## QA Round 3

Fresh independent reviewer, contract-verified against issue #459, reviewing `4e5be9d` with CI fully green (Linux gcc+ASan/LSan ×2, macOS ×2, Windows). Focus: the two changes since Round 2 (the vendored `ts-runtime` GLR cap and the deep-test fork-harness conversion), plus a full independent pass. Result: **0 critical, 0 major, 4 minor** — all zero-edge-safe and advisory.

### Review of changes since Round 2
| Item | Verdict | Notes |
|---|---|---|
| GLR merge cap (`ts_runtime/src/stack.c`) | sound | Cap sits in the recursive-merge branch; past depth 512 the merge recursion is skipped but the `dynamic_precedence` update + `return` still run — identical control flow to the existing `MAX_LINK_COUNT` bail-out. The merge branch performs no retain/release itself (all refcount accounting is in the leaf-insert path it never reaches via `return`), so skipping is refcount-neutral — no leak, no double-free. Unmerged ambiguity stays bounded by input depth + the pre-existing `MAX_LINK_COUNT`; no new unbounded growth. Triggers only at >512-deep ambiguity nesting → normal shallow code parses identically; only graceful degradation at pathological depth, never a wrong tree. 512 ≈ 130 KB, safe on a 1 MB Windows stack. |
| Deep test fork-harness (`test_stack_overflow.c`) | sound (documented limitation) | Correctly mirrors the Java/C++ fork pattern; removes the thread-local-parser leak with no dangling scaffolding. Known trade-off (accurately documented in the test): at DEPTH=2000 on the fork's 8 MB stack the input wouldn't overflow even without the cap, so on POSIX it exercises the cap's code path rather than reproducing a crash-without-cap; the true crash guard is the Windows in-process path (where the original SIGSEGV manifested). |

### Findings (all minor, zero-edge-safe, no action required)
1. **Strategy-label inaccuracy (telemetry only):** because Perl has no `class_node_types`, file-local subs are never typed as methods, so direct static/typed calls (`Foo->bar` where `bar` is local) get the `perl_method_inherited` strategy label instead of `perl_method_static`/`perl_method_typed`. Edge target QN + confidence are correct; only the label is off, and this mirrors the identical PHP behavior in `php_lsp.c`. (`internal/cbm/lsp/perl_lsp.c`)
2. **`SUPER::` resolves only the first `@ISA` parent** — multi-inheritance under-resolution (missing edge, never wrong). Uncommon in Perl; acceptable for v1.
3. **`SUPER::` test can't prove the child override is skipped** — both subs collapse to QN `test.main.greet` (single-file fixture), so the assertion can't distinguish parent vs child. The code is correct; the test's discriminating power is weak (a multi-file fixture would strengthen it).
4. **Coarse substring matching in test helpers** (`strstr`) — no collision with the current curated fixtures; robustness nit only.

### Contract verification (issue #459)
All criteria **met**: `perl_lsp.{c,h}` + `PerlLSPContext` (PHP-parity), `cbm_run_perl_lsp` entry + `CBM_LANG_PERL` dispatch, package/`use`/`require` resolution, OOP method resolution + `@ISA`/`use parent`/`use base` MRO (+ `SUPER::`), subroutine + Exporter `qw` import resolution, sigil-aware expression/bless type tracking with the `eval_depth` cap (8), perlfunc + CPAN stdlib seed (builtins keyed package-less so they emit no spurious edges), 13 tests mirroring `test_php_lsp.c` with full wiring, real-project validation (134 Perl files → 1178 resolved CALLS edges), and the Windows stack-overflow crash-safety regression.

### Summary
| Severity | Count |
|---|---|
| Critical | 0 |
| Major | 0 |
| Minor | 4 |
| **Total** | **4** |

Independently verified: build exit 0; full suite 5619 passed (the lone failure is the documented sandbox-only `cli_hook_gate_script_no_predictable_tmp_issue384`); upstream CI fully green across all platforms. The four minors are cosmetic/telemetry or test-strength observations consistent with the single-file v1 scope and the PHP parity model — advisory for a future cross-file plan, none blocking.

**Clean round — no remaining actionable issues.**

---
*QA performed by Claude Code (claude-opus-4-8). Fresh read-only reviewer; reviewed `4e5be9d`.*

halindrome:
## QA Round — commit `6e87a9e` (last-`::` split fix)

**Reviewer model:** claude-opus-4-8 · scoped to the newly added commit `6e87a9e` (the rest of the PR passed earlier rounds).

### Contract Verification

|---|---|---|
| Split package-qualified static calls on the LAST `::` (`Foo::Bar::sub` → pkg `Foo::Bar`, sub `sub`) | satisfied | `perl_resolve_function_call` loop finds the last `::`; traced for `Foo::Bar::sub`. |
| Multi-level qualified sub resolves via `perl_lookup_method` to the correct FQN | satisfied | Registry types keyed by colon-form pkg in `perl_attach_methods`/`perl_lookup_method`; keys match. New test `perllsp_static_multilevel_package_call` PASS. |
| Single-level (`Foo::sub`) unchanged | satisfied | Loop yields identical pkg/sub for one `::`; `perllsp_static_package_call` still PASS. |
| Regression test fails on old code | satisfied | On the old first-`::` split, `Acme::Widget::render` → pkg `Acme`, sub `Widget::render` → no `perl_static_call` edge → test would fail. |
| No cross-file overclaim (per-file LSP fix) | satisfied | Commit/test describe per-file resolution only. |

### Findings

- Contract: **0** · Regression: **0**
- Memory safety of the new loop: verified (`p + 2` always in-bounds within a `::`-bounded string; `strstr` on a pointer-to-NUL is defined). Edge cases (`Foo::bar`, `Foo::Bar::sub`, `::bar`, `Foo::`, `::`, bare) all correct.

### Pre-existing observation — deferred follow-up (not in this PR)

The reviewer flagged a **twin** of this bug at `perl_eval_function_call_type` (`perl_lsp.c`, introduced in `e4ba06c`, predates this commit): it still splits on the first `::`, so the *return type* of a multi-level static call used for method-chaining (`Pkg::Sub::func()->m`) is reported unknown.

I investigated fixing it in this round but **deferred it**: return-type inference does not populate `signature.return_types` for plain user subs, so in the per-file LSP test harness a chained method off a static call does not resolve regardless of the split — the fix has no independently observable/testable effect here (an earlier multi-file probe that appeared to confirm it was a false positive; the edge came from the generic unique-name fallback, not the LSP). Rather than add an untestable change to this PR, it's tracked as a separate follow-up so it can land with a construct that actually exercises return-type inference.

### Verdict

**CLEAN** — `6e87a9e` correctly fixes multi-level qualified static-call resolution with no regressions; the suite is green (the lone `cli_hook_gate_script_no_predictable_tmp_issue384` failure is pre-existing and unrelated). The twin observation is a tracked follow-up, non-blocking for this PR.

DeusData:
Thanks @halindrome — reviewed this in depth and the Perl LSP work is clean and well-tested; the vendored `ts_runtime/src/stack.c` recursion-cap patch also checks out as a legitimate stack-overflow hardening (caps the GLR ambiguity-merge depth, removes no logic).

One thing before merge: please drop the bundled `.vbw-planning/…SUMMARY.md` planning/scratch doc — it shouldn't land in the repo. Once that's removed, we're good to go. 🙏

DeusData:
Hey, can you rebase on latest main? Otherwise LGTM

DeusData:
Thanks again @halindrome — all my earlier asks are resolved (the `.vbw-planning` summary is dropped in `a141e81`, and the `ts_runtime` recursion-cap checks out as legitimate hardening as I noted). The rebase-on-latest-main I mentioned above is the only thing left and we're good to merge.

Sorry for any extra noise from my side. One optional nicety while you're rebasing (your DCO is already green here, so genuinely not a blocker): dropping the `Co-Authored-By: Claude …` / `Claude-Session:` trailers keeps the history AI-attribution-free, same as on #632. Really nice work on the Perl tier — clean, well-tested, and the per-file registry keeps it off the shared-registry path. 🙏

DeusData:
Ran a deeper review pass on top of the LGTM — confirmed **merge-ready, no blockers** (security clean, the `ts_runtime` recursion cap is graceful-degradation-only, resolution is properly bounded — `PERL_EVAL_MAX_DEPTH`, walk-depth caps, cycle-protected `@ISA` walk — tests are genuine guards, originality high-confidence). A few **optional, non-blocking** follow-ups if you ever revisit:

- **`perl_lsp.c:520`** — `perl_eval_function_call_type` still splits on the *first* `::`, while the edge-emitting path (`perl_resolve_function_call`) was fixed in `1576c3f` to split on the *last* `::`. No wrong edge results (the return-type-chaining path just yields `unknown` for `Foo::Bar::sub()`), but reusing the last-`::` scan here would close the gap.
- **Registry not finalized** — `cbm_run_perl_lsp` never calls `cbm_registry_finalize(&reg)` before resolution, so the `lookup_func/method/symbol` calls fall back to a linear scan (PHP/Python finalize to O(1) hash buckets). It's per-file and append-only, so there's **no** cross-project O(n²) — just a per-file perf-parity gap. A `finalize(&reg)` after `perl_attach_methods` would match the siblings.
- **`perl_infer_self_type:855`** — `strstr(rtxt, "shift")` is a substring match, so `my $self = shifty();` could mis-bind `$self`. Very low real-world odds; a word-boundary match for `shift` / `shift @_` / `$_[0]` would tighten the zero-edge guarantee.

None of these gate the merge — nice work on the Perl tier. 🙏

DeusData:
Thanks for the Perl LSP work. Before review, please clean up the commit history to remove generated/session attribution and keep the final commits focused. The vendored tree-sitter runtime change needs a separate maintainer decision with a clear justification and upstream/provenance notes, since runtime patches are higher risk than ordinary language extraction code.

halindrome:
Rebased onto latest `main` and cleaned up the history (force-pushed):

- **Attribution:** all commits now drop the `Co-Authored-By: Claude…` trailers; DCO sign-off preserved as the last trailer on every commit.
- **Vendored `ts_runtime` split out:** the GLR stack-merge recursion cap and its fork-harness test are fully removed from this PR — #461 now touches zero `vendored/` files and is pure Perl language-extraction code. Filed separately as #913 for your independent decision on the runtime patch.
- **Optional follow-ups from your deep review, all folded in:** last-`::` split in `perl_eval_function_call_type`; registry finalize before resolution (scratch-arena form, matching `php_lsp`/`java_lsp`); word-boundary `shift` match in `perl_infer_self_type`.
- **Docs:** added Perl to the Hybrid LSP section of the README (capabilities table row, intro + features lists, badge 9→10) to match how the other LSP-tier languages are documented.

Full suite is green apart from the pre-existing environmental `cli_hook_gate_script_no_predictable_tmp_issue384` (unrelated — this branch touches no CLI code). All 20 Perl LSP tests pass, including the deep-nesting guard, confirming the Perl tier does not depend on the removed runtime cap.

halindrome:
## QA Round 4

Scope: commit `bbf32061` — `fix(perl-lsp): guard pathologically nested Perl before tree-sitter parse`. This round was triggered by red CI (test-windows CLANG64 AddressSanitizer stack-overflow + test-unix ubuntu-24.04-arm 4h timeout), both on `lsp_perl_deep_expression_no_crash`. Rounds 1–3 covered the rest of the PR.

Contract source: issue #459.

### Contract Verification

| Criterion (round-4-relevant) | Status | Evidence |
|---|---|---|
| Must not crash or hang on pathological input; tests pass on all CI platforms (incl. Windows CLANG64 + ubuntu-arm) | ✅ satisfied | Pre-parse guard in `cbm_extract_file` returns a clean `has_error` result **before** `ts_parser_parse_with_options` for `CBM_LANG_PERL` when running bracket depth exceeds 128. The overflow site (`stack_node_add_link`, GLR stack merge in vendored tree-sitter) is never reached. `test_stack_overflow.c` still asserts `ASSERT_FALSE(so_extract_crashes(...))` at DEPTH=30000, which now exercises the guard path. |
| Zero-edge guarantee preserved | ✅ preserved | A skipped file returns zero defs/calls/imports; pipeline consumers (`pass_parallel.c`, `pass_definitions.c`) treat `has_error` as a file-error and emit no nodes/edges — no false edges. |
| Perl-only scoping justified | ✅ satisfied | Sibling parser-stress tests (`lsp_python_deep_expression_no_crash`, C/Java) pass unguarded at the same 30000 depth — their grammars are unambiguous. Only Perl's `f(...)` call-vs-bareword ambiguity drives the GLR head-merge recursion. |

### Root cause (confirmed)

The AddressSanitizer stack-overflow is inside vendored tree-sitter's GLR parser — `stack_node_add_link` (`internal/cbm/vendored/ts_runtime/src/stack.c`) recurses once per nesting level while merging the ambiguous parse-stack heads that Perl's `f(...)` grammar produces. It happens **during** `ts_parser_parse`, before any LSP walk runs, so the `CBM_LSP_PERL_MAX_WALK_DEPTH` guard added in QA round 1 could never fire. Reproduced locally with a symbolized backtrace on a 1 MB-stack thread; verified the fix turns `depth=30000` into a clean skip and leaves `depth=50` extracting normally.

### Findings

**Contract:** none. **Regression:** none.

Verified correct on the points that matter: boundary (`++depth > cap` trips at 129, so a legitimately 128-deep file passes), running-depth clamped at 0 on unbalanced closers (can't be masked by interleaving), signed-`char` comparison correct, guard placed before `get_thread_parser`/parse with no leak, returns a valid `CBMFileResult` matching the existing `parse failed`/`parse timeout` early-return shape, downstream `has_error` handling adds no retry/error-inflation. Test integrity intact — no skip/only/stub, corrected comment accurately describes the GLR-parser mechanism.

### Pre-existing issues discovered (observations — non-blocking)

- **F-01 (minor, confirmed):** `cbm_source_nesting_exceeds` counts brackets inside strings/heredocs/POD/comments. A legitimate Perl file with 129+ net-unbalanced open brackets in string content would trip the cap and skip that whole file's extraction. Deliberate, documented tradeoff (comment in `cbm.c`); realistic incidence is negligible. Graceful degradation, not a crash or false edge.
- **F-02 (minor, hypothetical):** The guard is `CBM_LANG_PERL`-scoped. Another genuinely ambiguous grammar (e.g. Ruby/shell) could in principle drive the same GLR recursion and remain unguarded on a 1 MB stack. Latent, pre-existing, outside the #459 contract; noted as a possible follow-up to generalize the pre-parse nesting guard.

### Summary

| Severity | Contract | Regression |
|---|---|---|
| Critical | 0 | 0 |
| Major | 0 | 0 |
| Minor | 0 | 0 |
| **Total** | **0** | **0** |

### Notes

- **SAST:** code-scanning alerts API returns 403 for the fork-contributor token, so a scoped SAST delta could not be pulled. The PR's own CodeQL / security-static / codeql-gate checks pass on GitHub.
- **Branch behind base:** `perl-lsp-semantic-resolution` is ~21 commits behind `upstream/main` (base advanced since the PR was cut). The round-4 fix is self-contained and independent of that delta, but a rebase/refresh onto current `main` is advisable before merge.

**Verdict:** clean round. The pre-parse guard is correctly implemented and prevents the tree-sitter GLR stack overflow; the two findings are documented design tradeoffs, not defects.

---
*QA performed by Claude Code (claude-opus-4-8)*

halindrome:
## CI status after rebase onto current `main`

Rebased this branch onto current `upstream/main` (it had drifted ~21 commits behind) and re-ran CI. Result: **17/19 checks green, including the two that this PR's fix targeted.**

### Perl deep-nesting fix — confirmed on every platform
- `lsp_perl_deep_expression_no_crash` → **PASS on all platforms**, including `test-windows (CLANG64)` (previously an AddressSanitizer stack-overflow) and `test-unix (ubuntu-24.04-arm)` (previously a 4h timeout). Both jobs now run to completion.
- The rebase also cleared the transient `gbuf_*` / `mem_over_budget_low_rss` failures seen on the prior run — those were fixed upstream in commits this branch was missing (e.g. the RSS-reconciliation series).

### Sole remaining red is unrelated to this PR
`test-unix (ubuntu-24.04-arm)` fails on **one** test:

`extract_wide_flat_file_is_linear` (`tests/test_extraction.c`) — `wide-flat extract took 285s (>=30s bound) — quadratic child access`

Why this is not a #461 defect:
- The test is **newly added upstream** in `7b8b7038` ("TypeScript-at-scale indexing cascade…") and arrived here via the rebase; it did not exist on the pre-rebase branch.
- It exercises `CBM_LANG_JAVASCRIPT` (a 400k-sibling `wide.js`). This PR's only new hot-path code, the pre-parse nesting guard, is gated `if (language == CBM_LANG_PERL …)` and never executes for this test.
- It **passed on `macos-14`, `macos-15-intel`, `ubuntu-latest`, `windows-CLANG64`, and locally** — it fails only on the `ubuntu-24.04-arm` runner, blowing the 30s wall-clock bound at 285s.

This looks like the new `extract_wide_flat_file_is_linear` bound being too tight for the slower `ubuntu-24.04-arm` runner under ASan (or a residual super-linear tail in JS extraction on arm). We don't have admin rights to re-run the arm job — could a maintainer re-run it? If it reproduces at ~250s+, the bound may need relaxing for arm, or the arm-specific extraction cost investigated separately from this PR.

### Note on the fix approach
The `cbm_extract_file` guard is a pragmatic workaround; the proper fix is bounding the GLR stack-merge recursion inside the vendored tree-sitter runtime, tracked in #913. The guard comment references #913 and is marked for removal once that lands.

DeusData:
Thanks for the substantial work here — Perl LSP-tier resolution is very much wanted (#459). Three things before this can move:

1. **The ARM leg caught a real quadratic pattern**: `extract_wide_flat_file_is_linear` fails with *"wide-flat extract took 285s (>=30s bound) — quadratic child access"*. This guard exists because indexed `ts_node_child(n, i)` in a loop over a wide node is O(i) per call → O(n²) per file; it passes on fast x64 runners and explodes on ARM. Please iterate wide children with a `TSTreeCursor` (`ts_tree_cursor_goto_first_child` / `goto_next_sibling`) in whatever perl_lsp/extraction loop walks flat lists — the other hybrid LSPs all had this class fixed and there are cursor-based examples throughout `extract_defs.c`.

2. **New-LSP checklist** (mandatory for every LSP language): run `scripts/check-lsp-originality.sh` and paste the result in the PR (no verbatim overlap with reference servers), and self-check against the known quadratic patterns: finalize-after-registration (no per-registration finalize), no lookup-inside-registration-loop, no registry tail-scans, shared registries treated read-only after seal.

3. **Rebase onto current main** — the branch predates several large merges (LSP resolve-perf work, coverage signal touching `internal/cbm/cbm.c`); `lsp_all.c` and `Makefile.cbm` moved.

Happy to do the deep review (resolution strategy, stdlib data provenance, benchmarks on a large OSS Perl repo) once the guard is green and the checklist is posted.

halindrome:
## New-LSP checklist + quadratic-pattern response

Thanks for the concrete pointers. All three items addressed; details below.

### 1. Quadratic patterns — fixed in `perl_lsp.c`

**AST wide-child iteration (the class the ARM `extract_wide_flat_file_is_linear` guard caught).** Every child-iteration loop that read `ts_node_child(node, i)` in a `for i < child_count` loop was O(i) per access → O(n²) on a wide flat node. All of them now route through a new `perl_collect_children()` helper that mirrors `wd_collect_children()` in `extract_defs.c`: it collects children forward with a `TSTreeCursor` **only** for wide nodes (≥ 64 children), and returns NULL for the small common case so those loops keep the cheaper indexed access with zero cursor overhead. 13 loops converted, including the two top-level walks (`perl_lsp_process_file`, `perl_attach_methods`), both recursive descents (`perl_resolve_calls_in_node`, `perl_pass1_scan`), and the package-body/`qw`/`@ISA` list walks.

**Method-table build (self-check surfaced a second same-class quadratic).** `perl_attach_methods` previously called a per-sub `perl_type_add_method` that did a linear `lookup_type` and rebuilt the package's whole NULL-terminated method array on every add → O(subs·types + Σ methods²), i.e. the same wide-flat blow-up but for a single-package file with many subs. Replaced with a collect → sort → build batch: one AST pass collects `(package, short-name, sub-qn)` mappings into a scratch vector, `qsort` groups by package (source order preserved within a package via a stable tiebreak, so first-defined still wins on redefinition), and each package's method array is allocated exactly once. Now O(S log S).

### 2. New-LSP originality checklist

`scripts/check-lsp-originality.sh` (full run, all references):

- **CLEAN** — no verbatim string/comment overlap with any reference server (pyright, TypeScript, gopls, Roslyn, clangd, jdt.ls, kotlin-ls, rust-analyzer, phpactor), and no structural clones (jscpd, ≥ 50 tokens) between our C and clangd. Perl has no upstream LSP to scan against (same situation as PHP); the implementation is spec-derived (perlfunc + tree-sitter-perl grammar node/field names verified against the vendored `parser.c`).

Known-quadratic self-check:
- **finalize-after-registration** — single `cbm_registry_finalize_into` after all registration (stdlib, file defs, packages, methods), before resolution. ✅
- **no lookup-inside-registration-loop** — the per-sub `lookup_type` is gone; type find/create is now once per *distinct package*. ✅
- **no registry tail-scans** — no types are added after finalize; resolution runs on the sealed registry. ✅
- **shared registries read-only after seal** — `perl_lsp_process_file` runs on the finalized const registry; no post-seal mutation. ✅

### 3. Rebase

Rebased onto current `main` (was 16 ahead / 71 behind; `lsp_all.c` and `Makefile.cbm` had moved). No conflicts.

### Verification

- `scripts/build.sh` clean under `-Werror`.
- `scripts/test.sh`: **5980 passed, 1 skipped**, and the one red is `cli_hook_gate_script_no_predictable_tmp_issue384` (`tests/test_cli.c`) — a CLI hook-install test with no overlap with Perl AST/extraction code; it fails identically with and without this PR's changes (pre-existing/environmental in this local sandbox). CI is the arbiter there.
- All 14 `perllsp_*` resolution tests pass; `lsp_perl_deep_expression_no_crash` passes.
- `clang-format` applied.

Ready for the deep review (resolution strategy, stdlib data provenance, large-Perl-repo benchmarks) whenever you are.

DeusData:
Merged — thank you for this one, and for the way you handled the review loop. The Perl LSP fills a long-standing gap, and the final shape is exactly what we hope for in this layer: MRO dispatch with bounded frontiers, the sort-based batch method-table build, and cursor iteration over wide nodes (the ARM linearity guard that caught the original quadratic now passes with room to spare). The curated perlfunc/CPAN stdlib seed and the zero-edge guarantee for unresolved receivers are the right v1 calls. Great work — looking forward to the follow-ups whenever you get to them.

Issue #220, opened 2026-04-07
C++ CALLS edges attributed to Module node instead of Function/Method nodes


nishitpatel92:
## Bug

In C++ codebases, all CALLS edges originate from the Module (file) node rather than from individual Function/Method nodes. This makes `trace_path` unable to traverse C++ call chains at function-level granularity.

**Version:** 0.6.0 (macOS arm64)

## Evidence

Querying CALLS edge source labels across languages in the same repository:

```sql
-- C++ (hbmx/marcopolo/src/): 100% of CALLS from Module
SELECT n1.label AS source_label, count(*) AS call_count
FROM edges e JOIN nodes n1 ON e.source_id = n1.id
WHERE e.type = 'CALLS' AND n1.file_path LIKE 'hbmx/marcopolo/src/%'
GROUP BY n1.label ORDER BY call_count DESC;

-- Result: Module|1706  (no Method or Function entries)

-- Swift (ios/MarcoPolo/): 82% from Method/Function
SELECT n1.label AS source_label, count(*) AS call_count
FROM edges e JOIN nodes n1 ON e.source_id = n1.id
WHERE e.type = 'CALLS' AND n1.file_path LIKE 'ios/MarcoPolo/%'
GROUP BY n1.label ORDER BY call_count DESC;

-- Result: Method|20160, Module|4417, File|611, Function|261
```

Concrete example — `AutoplayController.cpp` has 29 Method nodes but zero outbound CALLS edges. All 23 CALLS are attributed to the `.hpp` Module node:

```sql
-- Zero CALLS from .cpp
SELECT count(*) FROM edges e JOIN nodes n1 ON e.source_id = n1.id
WHERE e.type = 'CALLS'
AND n1.file_path = 'hbmx/marcopolo/src/conversation/AutoplayController.cpp';
-- Result: 0

-- 23 CALLS from .hpp Module node
SELECT count(*) FROM edges e JOIN nodes n1 ON e.source_id = n1.id
WHERE e.type = 'CALLS'
AND n1.file_path = 'hbmx/marcopolo/src/conversation/AutoplayController.hpp';
-- Result: 23

-- All from the Module node specifically
SELECT DISTINCT n1.id, n1.name, n1.label FROM edges e
JOIN nodes n1 ON e.source_id = n1.id
WHERE e.type = 'CALLS'
AND n1.file_path = 'hbmx/marcopolo/src/conversation/AutoplayController.hpp';
[...]

PR #463, opened 2026-06-14
fix(extract): attribute C/C++ CALLS edges to the enclosing function

### LINKED ISSUE #438: Cross-file CALLS edges resolve to Module node instead of caller Function node (v0.7.0)
## Bug Description

  When a function in file A calls a function in file B, the resulting `CALLS` edge is attached
  to the **Module node** (file-level) rather than the specific **Function node** that contains
  the call. This makes "find callers of X" queries return a file path instead of a function
  name.

  ## Steps to Reproduce

  1. Index a C project with cross-file function calls
  2. Run:
     codebase-memory-mcp cli query_graph '{"project":"<name>","query":
       "MATCH (caller)-[:CALLS]->(callee) WHERE callee.name='''lp_uart_init'''
        RETURN caller.name, caller.file_path"}'
  3. `caller.name` returns the **file path** (e.g. `components/.../lte_mgr.c`) not a function
  name

  ## Root Cause (observed)

  The CALLS edge has `strategy=unique_name, confidence=0.75` and `source_id=""` (empty).
  The source node's `qualified_name` is the **Module** node (`...lte_mgr.lte_mgr`), not a
  Function.

  The `bt` (body tokens) property on Function nodes **does** contain the callee symbol name,
  confirming the indexer identified the call — but the edge is not linked to the specific
  Function.

  ## Expected Behavior

  MATCH (caller:Function)-[:CALLS]->(callee:Function)
  WHERE callee.name = 'lp_uart_init'
  RETURN caller.name, caller.file_path, caller.start_line
  Should return the specific caller function (e.g. `lte_mgr_init` at `lte_mgr.c:414`).

  ## Impact

  - `trace_call_path {"from_function": "X"}` returns empty for cross-file calls
  - "Find callers / impact analysis" queries (a core use case) silently return wrong granularity
  ## Impact

  - `trace_call_path {"from_function": "X"}` returns empty for cross-file calls
  - "Find callers / impact analysis" queries (a core use case) silently return wrong granularity
  - Scored as `correct=1/2` in our A/B benchmark vs grep (8-question VCM C firmware suite)

  ## Workaround

  Two-step: get module name from CALLS edge → `grep` specific function in that file.
  Or: `get_code_snippet` on candidate functions and check their `bt` field manually.

  ## Environment

  - `codebase-memory-mcp` v0.7.0
  - Project: ESP-IDF C firmware (C99, ~113 files, 428 functions, 343 CALLS edges)
  - OS: Linux (Ubuntu 24.04)

KerseyFabrications:
A CALLS edge whose caller is a C/C++/CUDA/GLSL function was sourced to the
file's Module node instead of the calling Function. "Find callers of X"
returned a file path, outbound trace_path returned empty, and
(:Function)-[:CALLS]->(:Function) queries missed for these languages.

Root cause: the enclosing-function resolvers read only tree-sitter's `name`
field, but a `function_definition` node has none — the name lives in the
declarator chain (pointer/function/parenthesized/array declarators). So
func_node_name() (internal/cbm/helpers.c) and resolve_func_name_node()
(internal/cbm/extract_unified.c) returned NULL, the enclosing scope fell
back to the module QN, and the edge was attributed to the Module node. This
is the C counterpart to #220, which fixed the definition-naming path but not
the enclosing-call path.

Fix: descend the declarator chain to the innermost name node (mirroring
resolve_c_declarator_name in extract_defs.c, including qualified and operator
names) when a function_definition lacks a `name` field. Adds the regression
test c_caller_attribution asserting a C call's enclosing_func_qn is the
function, not the module.

Fixes #438

Signed-off-by: Kris Kersey <kris@kerseyfabrications.com>

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -23,7 +23,8 @@
 // Tree traversal limits.
 enum {
     TEMPLATE_DEPTH_LIMIT = 4,
-    DECLARATOR_DEPTH_LIMIT = 8,
+    DECLARATOR_DEPTH_LIMIT = CBM_DECLARATOR_DEPTH_LIMIT, // shared define in helpers.h
+
     EXPORT_ANCESTOR_DEPTH = 4,
     DECORATOR_SCAN_LIMIT = 3,
     C_RETURN_WALK_DEPTH = 5,
@@ -457,27 +458,6 @@ static TSNode resolve_func_name_fp(TSNode node, CBMLanguage lang, const char *ki
     return null_node;
 }
 
-// Check if a node type is a terminal C declarator name.
-static bool is_c_terminal_name(const char *dk) {
-    return strcmp(dk, "identifier") == 0 || strcmp(dk, "field_identifier") == 0 ||
-           strcmp(dk, "operator_name") == 0 || strcmp(dk, "operator_cast") == 0 ||
-           strcmp(dk, "destructor_name") == 0;
-}
-
-// Resolve name from a C++ qualified_identifier/scoped_identifier.
-static TSNode resolve_qualified_name(TSNode decl) {
-    static const char *name_kinds[] = {"operator_name", "operator_cast",    "destructor_name",
-                                       "identifier",    "field_identifier", NULL};
-    for (const char **k = name_kinds; *k; k++) {
-        TSNode found = cbm_find_child_by_kind(decl, *k);
-        if (!ts_node_is_null(found)) {
-            return found;
-        }
-    }
-    TSNode null_node = {0};
-    return null_node;
-}
-
 // C++/CUDA: out-of-line method definitions name the function with a qualified
 // declarator (`Foo::bar`, or `ns::Foo::bar`). Return the immediate enclosing
 // class name (the scope segment directly left of the function name, e.g. "Foo"),
@@ -526,30 +506,6 @@ static char *cpp_out_of_line_parent_class(CBMArena *a, TSNode node, const char *
     }
     char *text = cbm_node_text(a, scope, source);
     return (text && text[0]) ? text : NULL;
-}
-
-// Resolve function name from C/C++/CUDA/GLSL declarator chain.
-static TSNode resolve_c_declarator_name(TSNode node) {
-    TSNode decl = ts_node_child_by_field_name(node, TS_FIELD("declarator"));
-    for (int depth = 0; depth < DECLARATOR_DEPTH_LIMIT && !ts_node_is_null(decl); depth++) {
-        const char *dk = ts_node_type(decl);
-        if (is_c_terminal_name(dk)) {
-            return decl;
-        }
-        if (strcmp(dk, "qualified_identifier") == 0 || strcmp(dk, "scoped_identifier") == 0) {
-            return resolve_qualified_name(decl);
-        }
-        TSNode inner = ts_node_child_by_field_name(decl, TS_FIELD("declarator"));
-        if (ts_node_is_null(inner) && ts_node_named_child_count(decl) > 0) {
-            inner = ts_node_named_child(decl, 0);
-        }
-        if (ts_node_is_null(inner)) {
-            break;
-        }
-        decl = inner;
-    }
-    TSNode null_node = {0};
-    return null_node;
 }
 
 // R: resolve function_definition name from parent binary_operator lhs.
@@ -675,7 +631,7 @@ static TSNode resolve_func_name_c_family(TSNode *node_ptr, CBMLanguage lang, con
          lang == CBM_LANG_GLSL || lang == CBM_LANG_HLSL || lang == CBM_LANG_ISPC ||
          lang == CBM_LANG_SLANG) &&
         strcmp(kind, "function_definition") == 0) {
-        return resolve_c_declarator_name(*node_ptr);
+        return cbm_resolve_c_declarator_name_node(*node_ptr);
     }
     TSNode null_node = {0};
     return null_node;
--- a/internal/cbm/extract_unified.c
+++ b/internal/cbm/extract_unified.c
@@ -87,6 +87,44 @@ static const char *compute_wolfram_func_qn(CBMExtractCtx *ctx, TSNode node) {
     return NULL;
 }
 
+/* C/C++/CUDA/GLSL: function_definition has no `name` field — the name is nested
+ * in the declarator chain. Descend the `declarator` field to the innermost name
+ * node. Without this, the enclosing-function scope for calls made inside a C
+ * function resolves to NULL and the call is attributed to the module rather than
+ * the function (issue #438). Mirrors resolve_c_declarator_name() in extract_defs.c. */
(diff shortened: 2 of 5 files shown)


DeusData:
Thanks @KerseyFabrications — the fix is correct and `c_caller_attribution` is exactly the right reproduce-first guard.

One small thing before merge: `CBM_DECLARATOR_DEPTH_LIMIT` is now `#define`d independently in both `extract_unified.c` and `helpers.c` (identical values today, but they have to stay in sync with the existing definition in `extract_defs.c`). Could you hoist it into a shared header (or reuse the existing one) so there's a single source of truth? Once that's deduplicated, this is good to go. 🙏

DeusData:
Thanks @KerseyFabrications — the fix is correct and the `c_caller_attribution` test is solid, so the behavior is good. The one open item from the earlier review is still here: the declarator-name walker is now duplicated in three places (`helpers.c`, `extract_unified.c`, and the existing `extract_defs.c`), and `CBM_DECLARATOR_DEPTH_LIMIT` is `#define`d twice.

Could you extract a single shared `cbm_resolve_c_declarator_name_node()` (in `helpers.c`, declared in a shared header) and route all three sites + the one depth-limit constant through it? That triplication is exactly the kind of drift that caused #438 in the first place. Once it's deduped, this is good to merge. 🙏

KerseyFabrications:
Done. Deduplicated in the latest push then rebased.

- Extracted a single `cbm_resolve_c_declarator_name_node()` into `helpers.c` (declared in `helpers.h`), carrying its `is_c_terminal_name()` / `resolve_qualified_name()` helpers with it.
- Routed all three sites through it: the defs (`extract_defs.c`), calls (`helpers.c`), and unified (`extract_unified.c`) extractors. The two private copies are gone.
- Hoisted `CBM_DECLARATOR_DEPTH_LIMIT` into `helpers.h` as the single source of truth; `extract_defs.c`'s  `DECLARATOR_DEPTH_LIMIT` enum now derives from it (it's still referenced by `cpp_out_of_line_parent_class`), so there's exactly one definition.

I canonicalized on the original `extract_defs.c` logic (operator/destructor-aware via `resolve_qualified_name`) rather than the inlined copies, so defs-extraction behavior is unchanged and the calls/unified paths now agree with it, closing the triplication that drifted into #438.

Full test suite is green (`c_caller_attribution` included), clang-format clean.

mattall:
Tested this PR against a large C++ codebase (~40k symbols, heavy use of out-of-line method definitions). Clear net improvement, but one very common out-of-line form still falls back to the File node, so I wanted to flag it before merge with a minimal repro.

**Aggregate (same tree, fresh full index; before = base@53ebeb4, after = this PR):**

- CALLS attributed to a function/method/class symbol: **5,927 → 11,666** (17.5% → 31.5% of all CALLS edges)
- CALLS still sourced from the file node (`File`/`Module`): **82.5% → 68.4%**
- (total CALLS 33,928 → 36,996)

So it roughly doubles symbol-level attribution — but the majority of CALLS are still file-anchored, and the remainder is concentrated in one definition form.

**The form still falling back to File: out-of-line methods defined *inside* `namespace { }` blocks.**

Same logical method, three ways:

| Definition form | Attributed to |
|---|---|
| In-class (inline in header or `.cc`) | ✅ the Method |
| Out-of-line at **global scope** (`using namespace X;`) | ✅ the Method |
| Out-of-line **inside a `namespace { }` block** | ❌ the **File** node |

Controlled check that isolates it: two out-of-line classes in the *same directory* — one whose `.cc` opens with `using namespace X;` at file scope, the other whose defs are wrapped in `namespace X { namespace Y { ... } }`. The first resolves every call to its methods; the second file-anchors every call. The only difference is the enclosing namespace-block context.

Minimal repro — the same definition written two ways:

```cpp
// foo.h
namespace mylib { class Foo { public: void Bar(); void Baz(); }; }
```

```cpp
// foo_global.cc  — RESOLVES: Foo::Bar's call to Baz() is attributed to Foo::Bar
using namespace mylib;
void Foo::Bar() { Baz(); }
```

```cpp
// foo_block.cc  — FILE-ANCHORS: identical body, but the call attributes to the File node, not Foo::Bar
namespace mylib { void Foo::Bar() { Baz(); } }
```

It looks like the declarator walker folds the class qualifier (`Foo::`) but not the surrounding open-`namespace` context into `enclosing_func_qn`, so the nested-namespace definition's computed QN is missing the `mylib::` prefix, mismatches the node, and re-triggers the `calls_find_source()` → `__file__` fallback from #554.

KerseyFabrications:
Update: I went ahead and built this rather than just filing it, and tracing it end-to-end corrected two things in my earlier diagnosis:

1. Namespace isn't actually in the QN. I'd assumed the `namespace {}` block left a `mylib::` prefix on one side of the comparison. It doesn't, namespace context never enters the C++ QN scheme at all (`namespace_name` is null for these, and `mylib` appears in no QN). The real defect is narrower and hits both forms equally: the call-side enclosing-QN computation drops the class qualifier for out-of-line definitions, producing `project.path.Bar` instead of `project.path.Foo.Bar`.
2. Resolution is exact-QN match, not a short-name resolver. `calls_find_source()` does a direct `find_by_qn(enclosing_qn)` and falls straight back to `__file__` on a miss, there's no short-name rescue on the source side. So the call never matches the Method node and file-anchors.

Root cause is the #438 drift pattern again: the out-of-line class resolution lived in the defs extractor but had two divergent, incomplete copies on the call side: `compute_func_qn` (unified walk → CALLS) and `cbm_enclosing_func_qn` (cached path → USAGES/THROWS/CONFIGURES/type-assigns). Neither reconstructed the class for out-of-line defs.

The fix consolidates instead of adding a fourth copy: promoted the class resolver to a shared helper plus a `cbm_cpp_out_of_line_method_qn()` that both call-side sites use, so the call-side QN now matches the defs-node QN. Side benefit: attribution is now correct for every edge type inside out-of-line methods, not just CALLS.

Validated with your controlled check as a test: indexed `foo.h` + a `using namespace` .cc + a `namespace {}` .cc, and both `Baz()` calls now source from the `Method` node rather than the file, plus an extraction-level unit test for the global/block/nested forms. Full suite green.

To keep this PR exactly the strict-improvement you approved, I'm not adding the out-of-line fix here — #463 stays as the declarator-walker dedup. The namespace-block fix is a separate follow-up PR stacked on this branch (I'll link it here once it's up); it rebases cleanly onto main the moment #463 merges.

KerseyFabrications:
PR #621 has been submitted with the next fixes. In draft until this is merged.

DeusData:
Really nice fix, @KerseyFabrications — and well diagnosed. C/C++ `function_definition` nodes carry the name down in the declarator chain rather than a `name` field, so the enclosing-function resolvers fell back to the file Module node and CALLS edges were misattributed. Walking the declarator chain in both `cbm_enclosing_func_qn` and `resolve_func_name_node` mirrors the existing `resolve_c_declarator_name` pattern from #220 cleanly, and the `c_caller_attribution` test is a solid guard. Thanks for closing #438 — merging.

PR #465, opened 2026-06-14
fix(cypher): preserve node properties through WITH aggregation

KerseyFabrications:
A node group variable carried through a WITH aggregation
(e.g. `WITH g, count(*) AS c RETURN g.file_path`) returned blank for every
property except its name: the carried virtual binding held only the group
key (the node's name) and lacked a store handle, so node_prop() could
neither read other fields nor compute degrees.
Fix: capture the node id of a bare node group-var in with_agg_find_or_create
and tag the carried virtual binding with it; in node_prop(), when such a stub
(id set, string fields unpopulated) is asked for a missing property, re-fetch
the full node via cbm_store_find_node_by_id and project it. Also propagate
the store onto virtual bindings so node_prop can re-fetch and compute
degrees. The stub gate is heuristic but never yields a wrong value — worst
case is one redundant indexed lookup. Adds regression test
cypher_exec_with_node_groupvar_prop.

Signed-off-by: Kris Kersey <kris@kerseyfabrications.com>

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -2061,15 +2061,18 @@ static const char *node_string_field(const cbm_node_t *n, const char *prop) {
 /* Get node property by name.
  * store may be NULL; only needed for virtual degree properties. */
 static const char *json_extract_prop(const char *json, const char *key, char *buf, size_t buf_sz);
+static void node_fields_free(cbm_node_t *n); /* defined below; used by the stub re-fetch */
 
 static const char *node_prop(const cbm_node_t *n, const char *prop, cbm_store_t *store) {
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks @KerseyFabrications — preserving node properties through `WITH` aggregation is a real fix and the regression test is good. One correctness issue before merge: using empty-string as the stub sentinel (`if (str && str[0])`) means a node legitimately stored with an empty string field (e.g. an empty `file_path` on a synthetic node) falls through to a re-fetch on every property access. Please use an explicit stub flag (or carry the node id through the aggregation path) instead of repurposing the empty-string sentinel. 🙏

PR #468, opened 2026-06-14
fix: skip Antigravity IDE install roots during discovery

### LINKED ISSUE #403: The IDE's installation directory is unnecessarily indexed.
My environment:

- Windows 11
- Antigravity IDE 2.0.3
- codebase-memory-mcp 0.7.0 with auto indexing on (true)

Behavior: This tool indexes the IDE installation directory with binary files and creates a database.

```
"C:\Users\xxxi\.cache\codebase-memory-mcp\C-Users-xxx-AppData-Local-Programs-Antigravity IDE.db"
"C:\Users\xxx\.cache\codebase-memory-mcp\C-Users-xxx-AppData-Local-Programs-Antigravity.db"
```

Indexing the IDE's own installation directory is unnecessary and takes up hundreds of MB of space.
We only need to index the projects that are open within the IDE.

rudi193-cmd:
## Summary

Fixes #403 by treating Antigravity IDE install directories under `AppData/Local/Programs` as non-project roots during discovery.

When auto-index passes the IDE installation directory itself as `repo_path`, basename-only directory skips do not help because the walker starts inside that root. This adds a normalized full-path guard before the walk so discovery succeeds with an empty file list instead of indexing IDE runtime assets/binaries.

## Changes

- Add `looks_like_ide_install_root()` path normalization (`\\` -> `/`, ASCII lowercase)
- Short-circuit `cbm_discover_ex()` for `AppData/Local/Programs/Antigravity*` roots
- Add regression coverage with a synthetic `AppData/Local/Programs/Antigravity IDE` tree containing source-like files

## Test plan

- [x] Discover-only regression runner: `75 passed`
- [ ] Full `make -f Makefile.cbm test` could not compile in this environment because system `zlib.h` is missing:

```text
tests/test_cli.c:23:10: fatal error: zlib.h: No such file or directory
src/cli/cli.c:87:10: fatal error: zlib.h: No such file or directory
```

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -131,6 +131,32 @@ static bool str_contains(const char *s, const char *sub) {
     return strstr(s, sub) != NULL;
 }
 
+/* Windows IDE auto-index integrations can accidentally pass the IDE install
+ * directory itself as the repository root. These are not source projects and
+ * are mostly binary/runtime assets. Match on the normalized full path because
+ * cbm_should_skip_dir() only sees basenames during the walk. */
+static bool looks_like_ide_install_root(const char *path) {
+    if (!path) {
+        return false;
+    }
+
+    char normalized[CBM_SZ_4K];
+    size_t i = 0;
+    for (; path[i] && i + 1 < sizeof(normalized); i++) {
+        char c = path[i];
+        if (c == '\\') {
+            c = '/';
+        }
+        if (c >= 'A' && c <= 'Z') {
+            c = (char)(c - 'A' + 'a');
+        }
+        normalized[i] = c;
+    }
+    normalized[i] = '\0';
+
+    return strstr(normalized, "appdata/local/programs/antigravity") != NULL;
+}
+
 /* ── Public filter functions ─────────────────────── */
 
 bool cbm_should_skip_dir(const char *dirname, cbm_index_mode_t mode) {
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for digging into #403 and putting up a fix, @rudi193-cmd — the instinct (keep IDE install roots out of the index) is exactly right, and the path-normalization helper is clean.

I'm going to close this particular approach, and I want to explain why so it doesn't feel arbitrary.

The change makes `cbm_discover_ex` return `0` with zero files whenever the repo path contains the literal substring `appdata/local/programs/antigravity`. Two problems with that as an engine-level rule:

1. **Silent success.** It returns success with an empty result and no log line, so anyone who legitimately indexes something under such a path gets an empty index with no indication why. "Looks like it worked but produced nothing" is the exact failure mode we work hardest to avoid.
2. **It doesn't generalize.** Hardcoding one product name special-cases the discovery core; the next IDE (or a rename) needs another string, and the engine slowly accretes product-specific branches.

The supported way to keep paths out of the index is a **`.cbmignore`** file at the indexed root — same syntax as `.gitignore`, including `!` negation — which already covers "don't index these directories" without touching the engine. For the broader "exclude install/asset roots globally" need, the right home is the configurable-skip-dirs request in #500, which is the general version of this idea.

For the Antigravity case specifically, the cleanest fix is really on the integration side — the IDE shouldn't hand its own install directory to the indexer as a project root.

Closing this one, but I genuinely appreciate the contribution and the clear reproduction in #403. If you'd like to take a run at the configurable-skip approach in #500, that's a change we'd happily review. 🙏

PR #477, opened 2026-06-16
fix(perl): eliminate false-positive Perl CALLS edges (builtins, framework method calls, config strings)

### LINKED ISSUE #476: Perl call graph polluted by false-positive CALLS edges (builtins, framework method calls, config strings)
## Summary

On real-world Perl codebases, the call graph is dominated by **false-positive `CALLS` edges**. Perl files are extracted (call sites emitted), and any call the textual resolver can't place falls back to a generic **short-name matcher** that has no language or call-kind awareness. It happily wires Perl builtins, framework method calls, and even mis-parsed config strings to unrelated project subs that merely share a name.

## Evidence

Measured on a large real Perl monorepo (~1,200 `.pm`/`.pl` + 352 `.cgi` endpoint files). For `.cgi` callers, the generic resolver produced **4,940 `suffix_match` edges**, of which the overwhelming majority are noise:

| Noise class | `.cgi` edges | Example callees |
|---|---|---|
| CPAN/framework **method** calls (`$obj->m()`) matched to unrelated project subs | ~3,064 | `log` (Log::Log4perl), `param`/`header` (CGI.pm), `connect`/`commit`/`rollback`/`execute` (DBI), `encode` (JSON) |
| Perl **builtins** matched to project subs | ~645 | `shift`, `push`, `keys`, `sprintf` |
| **Config strings** mis-extracted as call targets | ~305 | `log4perl.appender.File.utf8` |

These targets are not project calls at all — the classes belong to CPAN/framework modules not in the graph, the builtins are language primitives, and the config string is a literal. The generic resolver fabricates edges purely on short-name collision.

## Root causes

1. **Callee over-extraction (Perl):** `extract_scripting_callee` (`internal/cbm/extract_calls.c`, `CBM_LANG_PERL` branch) returns the raw first-child text of a call node, so non-identifier tokens (dotted config strings, literals) become callees.
2. **No builtin guard:** the generic resolver (`src/pipeline/registry.c`) matches builtin-named calls (`shift`/`push`/…) to project subs of the same name.
3. **No function-vs-method distinction:** `CBMCall` carries no call-kind, so a method call `$obj->commit()` with an unknown receiver is treated as a bare function and short-name-matched to a project `commit` sub.

## Proposed fix (language-gated; non-Perl behavior unchanged)

1. **Extraction hygiene:** in the Perl callee path, extract the real method/function name token and reject non-identifier callees (containing `.`, quotes, whitespace, etc.) so config strings/literals never become call targets.
2. **Builtin guard:** add a Perl builtin set; when an unresolved Perl call's name is a builtin, suppress the generic edge (real same-file subs are already resolved by earlier stages before the generic fallback).
3. **Method-vs-function:** add a `is_method` flag to `CBMCall`, set it during Perl extraction for arrow/method calls, thread it into the resolver, and suppress generic short-name matching for Perl method calls with an unknown receiver (precise method resolution is the LSP's job; bare short-name matching is almost always wrong).

Every change is gated on `CBM_LANG_PERL` — the generic resolver and `CBMCall` are shared by all languages, so the other nine families remain byte-identical.

## Acceptance criteria

- [ ] A Perl builtin call (e.g. `push @x, 1`) does **not** produce a `CALLS` edge to a project sub named `push`.
- [ ] A Perl method call with an unknown receiver (e.g. `$dbh->commit()`) does **not** generic-match to a project sub named `commit`.
- [ ] A dotted config-string token (e.g. `log4perl.appender.File.utf8`) is **not** extracted as a call target.
- [ ] Genuine intra-project Perl function calls **still** resolve.
- [ ] Non-Perl languages' extraction/resolution are unchanged (verified by the existing suite + the cross-language breadth check).
- [ ] Tests cover all of the above.

## Validation (on the proposed branch)

Re-indexing the same repo with the fix: `.cgi` `suffix_match` drops **4,940 → 655 (−87%)**, builtin/CPAN-method/config-string noise on `.cgi` goes to **zero**, project-wide `CALLS` edges drop by ~13,400 (noise removal — fewer, more-correct edges), `scripts/build.sh`/`scripts/test.sh` stay green, and the cross-language breadth check confirms all other languages still resolve.

halindrome:
## Summary

Perl files are extracted (call sites emitted), and any call the textual resolver can't place falls back to a generic **short-name matcher** with no language or call-kind awareness. It wires Perl builtins, framework method calls, and mis-parsed config strings to unrelated project subs that merely share a name — polluting the Perl call graph with false-positive `CALLS` edges.

This fixes the three sources, all **gated on `CBM_LANG_PERL`** so the generic resolver and `CBMCall` (shared by all 10 languages) stay byte-identical for non-Perl.

## What changed

- **`fix(perl): stop extracting config strings as call targets`** — `extract_scripting_callee` (Perl branch) now extracts the real method/function name token and rejects non-identifier callees (containing `.`, quotes, whitespace, …), so dotted config strings/literals (e.g. `log4perl.appender.File.utf8`) never become call targets.
- **`fix(resolver): don't match Perl builtins to project subs`** — adds a curated Perl builtin set (`src/pipeline/registry.c`); when an *unresolved* Perl call's name is a builtin, the generic edge is suppressed. Real same-file subs are already resolved by earlier stages before the generic fallback, so this only drops spurious builtin matches.
- **`fix(resolver): suppress generic short-name matching for Perl method calls`** — adds `is_method` to `CBMCall` (default false → no-op for other languages), set during Perl extraction for arrow/method calls, threaded into `resolve_single_call` / `pass_parallel`’s resolver. Perl method calls with an unknown receiver no longer generic-match to free subs (precise method resolution is the LSP's job).
- Tests in `tests/test_extraction.c` and `tests/test_registry.c` covering builtins, config-string rejection, method-call suppression, a genuine-call-still-resolves case, and a cross-language no-op check.

## Validation

Re-indexing a large real Perl monorepo (~1,200 modules + 352 `.cgi` endpoints) with the fix:

| metric | before | after |
|---|---|---|
| `.cgi` `suffix_match` edges | 4,940 | **655** (−87%) |
| `.cgi` builtin / CPAN-method / config-string noise | ~4,000 | **0** |
| project-wide `CALLS` edges | ~182.5k | ~169.4k (−13.4k noise removed) |

This is **precision via noise removal** — fewer, more-correct edges. Genuine intra-project resolution survives.

- `scripts/build.sh` — clean (`-Werror`).
- `scripts/test.sh` — green except the unrelated pre-existing `cli_hook_gate_script_no_predictable_tmp_issue384`; cross-language breadth check `[CALLS-BREADTH] 53 langs: 0 FAILURES` confirms all other languages still resolve.
- `clang-format` — clean on changed files.

Closes #476

### DIFF
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -236,6 +236,7 @@ typedef struct {
     int loop_depth;                     // enclosing loop nesting at the call site
     int branch_depth;                   // enclosing branch nesting at the call site
     int start_line;                     // 1-based source line of the call (for def range-match)
+    bool is_method;                     // Perl-only: arrow/method call ($obj->m). Default false.
 } CBMCall;
 
 typedef struct {
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -372,7 +372,16 @@ static bool perl_is_identifier_callee(const char *name) {
     }
     for (const char *p = name; *p; p++) {
         unsigned char c = (unsigned char)*p;
-        if (isalnum(c) || c == '_' || c == ':') {
+        if (isalnum(c) || c == '_') {
+            continue;
+        }
+        if (c == ':') {
+            // Only the '::' package separator is allowed: require an adjacent
+            // pair, and reject a lone ':', ':::', or a trailing '::'.
+            if (p[1] != ':' || p[2] == ':' || p[2] == '\0') {
+                return false;
+            }
+            p++; // consume the second ':'; the loop's p++ moves past the pair
             continue;
         }
         return false; // '.', space, quote, '/', etc. → not a sub/method name
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -355,6 +355,31 @@ static char *extract_swift_callee(CBMArena *a, TSNode node, const char *source, 
     return NULL;
 }
 
+// A Perl sub/method name is an identifier: it starts with a letter or '_',
+// contains only [A-Za-z0-9_] plus the '::' package separator, and is never a
+// string/config literal. tree-sitter-perl mis-parses config lines in .cgi /
+// heredoc-heavy files into call-shaped nodes whose "callee" is a dotted config
+// token (e.g. "log4perl.appender.File.utf8"); rejecting non-identifier text
+// here stops those from becoming bogus CALLS edges. Any '.', whitespace, quote,
+// or '/' disqualifies the token.
+static bool perl_is_identifier_callee(const char *name) {
+    if (!name || !name[0]) {
+        return false;
+    }
+    unsigned char c0 = (unsigned char)name[0];
+    if (!(isalpha(c0) || c0 == '_')) {
+        return false;
+    }
+    for (const char *p = name; *p; p++) {
+        unsigned char c = (unsigned char)*p;
+        if (isalnum(c) || c == '_' || c == ':') {
+            continue;
+        }
+        return false; // '.', space, quote, '/', etc. → not a sub/method name
+    }
+    return true;
+}
+
 // Callee extraction for scripting languages (Elixir, Perl, PHP, Kotlin, MATLAB).
 static char *extract_scripting_callee(CBMArena *a, TSNode node, const char *source,
                                       CBMLanguage lang, const char *nk) {
@@ -367,7 +392,24 @@ static char *extract_scripting_callee(CBMArena *a, TSNode node, const char *sour
         return NULL;
     }
     if (lang == CBM_LANG_PERL && ts_node_child_count(node) > 0) {
-        return cbm_node_text(a, ts_node_child(node, 0), source);
+        // Pull the actual sub/method name token rather than blindly taking
+        // child(0). Grammar (verified against the vendored parser):
+        //   method_call_expression   : field `method`   ($obj->m / Class->m)
+        //   function_call_expression : field `function` (foo(); name with '.'
+        //                              from a config-string misparse lands here)
+        //   ambiguous_function_call_expression : field `function`
+        //   func1op_call_expression  : builtin keyword as child(0) (no field)
+        TSNode name_node = ts_node_child_by_field_name(node, TS_FIELD("method"));
+        if (ts_node_is_null(name_node)) {
+            name_node = ts_node_child_by_field_name(node, TS_FIELD("function"));
+        }
+        if (ts_node_is_null(name_node)) {
+            name_node = ts_node_child(node, 0);
+        }
+        char *pn = cbm_node_text(a, name_node, source);
+        // Reject anything that is not a bare Perl sub/method identifier (config
+        // strings, quoted literals, paths) so no spurious CALLS edge is emitted.
+        return perl_is_identifier_callee(pn) ? pn : NULL;
     }
     if (lang == CBM_LANG_PHP) {
         TSNode func_node = ts_node_child_by_field_name(node, TS_FIELD("function"));
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -1176,6 +1176,14 @@ void handle_calls(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec, Walk
             call.loop_depth = state->loop_depth;     // enclosing loop nesting at this call
             call.branch_depth = state->branch_depth; // enclosing branch nesting at this call
             call.start_line = (int)ts_node_start_point(node).row + TS_LINE_OFFSET;
+            // Perl-only: flag arrow/method calls ($obj->m / Class->m). The
+            // generic short-name resolver cannot place a method without a known
+            // receiver type, so the call-resolution pass suppresses those edges.
+            // Default false for every other language (struct is zero-init).
+            if (ctx->language == CBM_LANG_PERL &&
+                strcmp(ts_node_type(node), "method_call_expression") == 0) {
+                call.is_method = true;
+            }
 
             TSNode args = ts_node_child_by_field_name(node, TS_FIELD("arguments"));
             if (!ts_node_is_null(args)) {
--- a/internal/cbm/lang_specs.c
+++ b/internal/cbm/lang_specs.c
@@ -585,7 +585,7 @@ static const char *perl_func_types[] = {"subroutine_declaration_statement", NULL
 static const char *perl_module_types[] = {"source_file", NULL};
 static const char *perl_call_types[] = {"ambiguous_function_call_expression",
                                         "function_call_expression", "func1op_call_expression",
-                                        NULL};
+                                        "method_call_expression", NULL};
 static const char *perl_import_types[] = {"use_statement", "require_statement", "require", NULL};
 static const char *perl_branch_types[] = {"if_statement",      "unless_statement", "for_statement",
                                           "foreach_statement", "while_statement",  NULL};
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -361,20 +361,22 @@ static int resolve_single_call(cbm_pipeline_ctx_t *ctx, CBMCall *call,
         }
     }
 
-    /* Perl call-graph noise guards (#459 follow-up). Applied only after LSP
-     * resolution declined, so a real LSP-resolved call is never suppressed.
-     * Gated to Perl — other languages reach cbm_registry_resolve unchanged.
-     *   1. Builtins (push/shift/keys/...): a generic short-name match to a
-     *      project sub sharing the name is almost always a false positive.
-     *   2. Method calls ($obj->m): without a resolved receiver type the bare
-     *      short-name match is wrong; that resolution is the LSP's job. */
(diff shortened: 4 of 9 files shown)


halindrome:
## QA Round 1

Reviewers: Claude Code (claude-opus-4-8) parallel review panel (3 lenses) + DigitalOcean DeepSeek (`deepseek-v4-pro`, `--double`).

### Contract Verification (issue #476)
| Criterion | Verdict | Evidence |
|---|---|---|
| (a) builtin call does not edge to same-named project sub | pass | suppressed at the resolver; builtin set bsearch-correct (94 entries) |
| (b) method call w/ unknown receiver not generic-matched | pass | `is_method` set at `extract_calls.c` for `method_call_expression` (in `perl_call_types`); suppressed at resolver |
| (c) config-string token not extracted as callee | pass | `perl_is_identifier_callee` rejects non-identifier tokens |
| (d) genuine intra-project Perl calls still resolve | **pass after fix** | see Finding 2 — guard reworked to keep `same_module`/`import_map` |
| (e) non-Perl languages byte-identical | pass | all suppression gated on Perl; `[CALLS-BREADTH] 53 langs: 0 FAILURES` |
| (f) tests cover the above | **pass after fix** | see Finding 1 — added resolver-contract tests |

### Findings

**[claude] Finding 1 — Test gap (major, fixed).** The suppression branch and edge-survival path had no resolver-level test (extraction-level + a builtin-set unit test only); the test comments promised end-to-end coverage that didn't exist. Fixed: extracted the suppression decision into pure `cbm_perl_suppress_generic_match()` and added contract tests (weak-match suppression, `same_module`/`import_map` retention, genuine-call survival, non-Perl no-op, NULL strategy).

**[claude] Finding 2 — Builtin guard over-suppressed genuine same-file calls (regression, fixed).** Perl has no LSP resolver, so the guard ran before `cbm_registry_resolve`, dropping a genuine same-file call to a builtin-named sub (e.g. a project `sub log`/`index`/`open` called as a bare function) that pre-PR resolved via `same_module`. Reworked: resolve first, then suppress **only weak** strategies (`suffix_match`/`unique_name`) and keep high-confidence `same_module`/`import_map`. Corrected the inaccurate inline comments. Verified on a real Perl monorepo: `.cgi` noise stays eliminated while `same_module` edges to builtin-named subs are recovered.

**[claude|do:deepseek-v4-pro] Finding 3 — `perl_is_identifier_callee` accepted a lone `:` (minor, fixed).** It allowed any `:`, so `Foo:Bar`/`::x`/`Foo:::Bar` would pass. Tightened to require `::` pairs (reject lone `:`, `:::`, trailing `::`). Harmless in practice (the grammar never emits such callee tokens) but now matches its own docstring.

**[do:deepseek-v4-pro] Finding 4 — locale-dependent `isalpha`/`isalnum` (minor, advisory, not changed).** Byte classification for identifiers is locale-dependent for bytes >127. Left as-is for this round (consistent with surrounding code; no reachable trigger from the grammar). Tracked as a possible follow-up.

**[do:deepseek-v4-pro] "is_method never set in extraction" (reported critical) — REFUTED (false positive).** The DO reviewer chunked the diff by commit and reviewed the config-string commit in isolation. `call.is_method` is set at `internal/cbm/extract_calls.c` (`handle_calls`, `method_call_expression`) and `method_call_expression` is in `perl_call_types` (`lang_specs.c`). Verified present and correct.

### Result
3 confirmed findings fixed (1 major + 1 regression + 1 minor); 1 advisory deferred; 1 reported-critical refuted. Fixes committed as `fix(perl): address QA round 1`. Build clean (`-Werror`), `clang-format` clean, suite green (5611 passed; the 1 failure `cli_hook_gate_script_no_predictable_tmp_issue384` is a pre-existing sandbox-only flake unrelated to this PR).

---
_SAST: GitHub code scanning is not enabled on this repo — security delta skipped (non-blocking)._

---
*QA performed by Claude Code (claude-opus-4-8) parallel panel + do:deepseek-v4-pro*

halindrome:
## QA Round 2

Reviewer: Claude Code (claude-opus-4-8) parallel review panel (3 lenses). The DigitalOcean DeepSeek second opinion timed out this round (~7 min, no output) — recorded as a non-blocking second-opinion failure; round 1 captured a DO opinion.

### Contract Verification (issue #476)
| Criterion | Verdict | Note |
|---|---|---|
| (a) builtin call → no edge to same-named sub | pass | |
| (b) method call w/ unknown receiver → no generic match | pass | |
| (c) config-string token not extracted | pass | |
| (d) genuine intra-project Perl calls still resolve | **pass after fix** | see Finding 1 |
| (e) non-Perl byte-identical | pass | `[CALLS-BREADTH] 53 langs: 0 FAILURES` |
| (f) tests cover the above | pass | + new `import_map_suffix` retention case |

### Findings

**[claude] Finding 1 — round-1 helper whitelist omitted `import_map_suffix` (regression, fixed).** The round-1 `cbm_perl_suppress_generic_match` kept only `same_module`/`import_map`, but `resolve_import_map` also returns `import_map_suffix` (confidence 0.85 — a genuine import resolution, above weak `unique_name`/`suffix_match`). A `::`-qualified Perl builtin/method call resolved via the import-suffix fallback was therefore wrongly dropped, partially missing criterion (d). Fixed: `import_map_suffix` added to the kept set; doc comment corrected; unit test added asserting it is retained. Only the weak short-name strategies (`suffix_match`/`unique_name`) are now suppressed.

**[claude] Finding 2 — leading-`::` (main:: shorthand) under-extraction (minor, hypothetical, advisory — deferred).** `perl_is_identifier_callee` rejects a callee beginning with `::` (e.g. `::foo` == `main::foo`) because the first char must be a letter/`_`. This can only ever *miss* an edge, never create a false one, and it's unverified whether tree-sitter-perl surfaces a leading-`::` callee token (the grammar is compiled). Not a noise/over-extraction criterion. Left as a tracked advisory rather than adding speculative code.

**[claude] Finding 3 — colon-edge-case test gap (minor, advisory — deferred).** `perl_is_identifier_callee` is a static helper exercised only indirectly; there's no focused test feeding `:::`, trailing `::`, `Foo::Bar::baz`, or `SUPER::method`. The reviewer traced each case and confirmed the logic is correct (accepts qualified/`_`-leading names; rejects lone `:`/`:::`/trailing `::`; no read-past-terminator). Coverage debt, not a defect — tracked as advisory.

**[claude] Finding 4 — pre-existing unrelated test failure.** `cli_hook_gate_script_no_predictable_tmp_issue384` (`tests/test_cli.c`) fails reading a hook-gate file under a sandboxed `/tmp` dir. The PR touches no cli/hook files; confirmed failing on clean `upstream/main` as well — environmental, out of scope for this PR.

### Result
1 confirmed regression fixed (`fix(perl): address QA round 2`); 2 minor/hypothetical items deferred as advisory; the 1 suite failure is pre-existing and unrelated. Build clean (`-Werror`), `clang-format` clean, suite green (5611 passed; only the pre-existing `issue384`). With (d) now passing, all contract criteria are met.

---
*QA performed by Claude Code (claude-opus-4-8) parallel panel (DO DeepSeek second opinion timed out — non-blocking)*

halindrome:
## QA Round 3 (confirming) — CLEAN ✅

Reviewer: Claude Code (claude-opus-4-8) parallel review panel (3 lenses). All three lenses returned **empty findings**. (The DigitalOcean DeepSeek second opinion did not complete within the window again this round — recorded as a non-blocking second-opinion failure; a DO opinion was captured in round 1.)

### Contract Verification (issue #476) — all pass
| Criterion | Verdict |
|---|---|
| (a) builtin call → no false edge to same-named sub | pass |
| (b) config-string token never extracted as callee | pass |
| (c) arrow/method calls flagged + generic match suppressed | pass |
| (d) genuine same-file/imported calls still resolve | pass |
| (e) suppression Perl-gated; other languages byte-identical | pass |
| (f) both sequential + parallel resolvers apply it consistently, after `cbm_registry_resolve` | pass |

### Verification highlights
- **Round-2 fix confirmed complete.** The panel enumerated *every* strategy literal `cbm_registry_resolve` can emit — `{import_map, import_map_suffix, same_module}` (kept) and `{suffix_match, unique_name}` (dropped) — and confirmed the keep/drop partition is exhaustive. The `fuzzy` strategy comes only from a separate resolver never fed to the helper (not a gap).
- **`perl_is_identifier_callee`** re-audited for read-past-terminator on the `::` lookahead: `p[2]` is read only when `p[1]==':'` (worst case NUL) — no OOB read.
- **Perl-gating** verified airtight: `CBMCall.is_method` is zero-initialized at both construction sites, so non-Perl behavior is byte-identical; `[CALLS-BREADTH] 53 langs: 0 FAILURES`.
- **Schema:** none. **SAST:** code scanning not enabled (skipped).

### Result
No new or remaining confirmed defects. The two prior advisory items (hypothetical leading-`::` under-extraction; colon-edge-case test gap — logic verified correct) remain non-blocking and were not re-raised. The single suite failure (`cli_hook_gate_script_no_predictable_tmp_issue384`) is pre-existing on `upstream/main` and out of scope.

**3 QA rounds complete; round 3 clean. Marking ready for review.**

---
*QA performed by Claude Code (claude-opus-4-8) parallel panel*

PR #483, opened 2026-06-17
fix(search): ignore empty-string label in search_graph name_pattern path

### LINKED ISSUE #481: search_graph name_pattern misses indexed Class nodes found by BM25 query
### Version

codebase-memory-mcp 0.8.1

### Platform

macOS (Apple Silicon)

### Install channel

### Binary variant

standard

`search_graph` with `name_pattern` does not return indexed class nodes that are returned by BM25 `query`.

Example from a TypeScript/Angular repo:

BM25 query works:

```json
{
  "project": "example-project",
  "query": "ExampleDialogService",
  "label": "",
  "name_pattern": "",
  "limit": 20
}

Returns:
ExampleDialogService
label: Class
file_path: src/app/services/example-dialog.service.ts

But regex name search returns no results:
{
  "project": "example-project",
  "query": "",
  "label": "",
  "name_pattern": ".*ExampleDialogService.*",
  "limit": 20
}
Result:
total: 0

A broader regex also missed expected classes:
"name_pattern": ".*Dialog.*Service.*"

Expected behavior:
- name_pattern should match the indexed Class node by name.
- If name_pattern intentionally searches a different field/scope than BM25, the tool docs or response hint should make that clear.

Impact:
- Agents fall back to grep even though the symbol graph contains the class.
- This makes structural discovery unreliable for class/service inventory tasks.

### Reproduction

- search_graph(query="WorkBuddyDialogService") finds WorkBuddyDialogService as Class
- search_graph(name_pattern=".*WorkBuddyDialogService.*") returns 0
- Same issue with broader .*Dialog.*Service.*

### Logs

```text

```

_No response_

### Confirmations

halindrome:
Closes #481

## Problem

`search_graph` returns zero results when a request passes an empty-string `label` (`"label": ""`) together with `name_pattern` (or `qn_pattern`). The reporter saw it as "`name_pattern` misses indexed Class nodes," but the trigger is the empty label, not the regex:

| Call | result |
|---|---|
| `name_pattern=".*Foo.*"` (no `label` key) | matches ✓ |
| `name_pattern=".*Foo.*"`, `label=""` | **0** ✗ |
| `name_pattern=".*Foo.*"`, `label="Class"` | matches ✓ |
| `query="Foo"`, `label=""` (BM25) | matches ✓ |

The BM25 `query` path treats an empty label as absent; the `name_pattern` path does not — so the same `"label":""` payload works for `query` but returns nothing for `name_pattern`. Agents then fall back to grep even though the symbol is in the graph.

## Root cause

`search_where_basic` (`src/store/store.c`) guarded the label clause on NULL only:

```c
if (params->label) {                 // "" is non-NULL → still appended
    ... "n.label = ?" ...            // binds n.label = '' → matches no node
}
```

## Fix

```c
if (params->label && params->label[0]) {   // ignore empty-string label
```

An empty `label` is now a no-op, identical to an omitted label and to the BM25 path. Non-empty labels still filter exactly as before.

## Test

`store_search_empty_label_ignored` (`tests/test_store_search.c`): `name_pattern` + `label=""` returns the same match as `name_pattern` alone, and a non-empty `label` still filters (`.*Order.*` + `label="Class"` → only the Class node).

## Verification

- `scripts/build.sh` clean; `scripts/test.sh` passes (the lone `cli_hook_gate_script_no_predictable_tmp_issue384` failure is pre-existing on `main`, environment-dependent, untouched here).
- Reproduced the original symptom on a real project and confirmed the fix: with the empty label ignored, `name_pattern` matches all expected Class nodes.

### DIFF
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -2348,7 +2348,12 @@ static int search_where_basic(const cbm_search_params_t *params, char *where, in
         *wlen = where_append(where, where_sz, *wlen, nparams, bind_buf);
         where_bind_text(binds, bind_idx, params->project);
     }
-    if (params->label) {
+    /* Ignore an empty-string label: it is non-NULL but should behave like an
+     * omitted label (no filter), matching the BM25 query path. Without the
+     * params->label[0] guard, name_pattern/qn_pattern searches that pass
+     * label="" append `n.label = ''`, which matches no node and silently
+     * returns zero results (issue #481). */
+    if (params->label && params->label[0]) {
         snprintf(bind_buf, sizeof(bind_buf), "n.label = ?%d", *bind_idx + SKIP_ONE);
         *wlen = where_append(where, where_sz, *wlen, nparams, bind_buf);
         where_bind_text(binds, bind_idx, params->label);
--- a/tests/test_store_search.c
+++ b/tests/test_store_search.c
@@ -86,6 +86,48 @@ TEST(store_search_by_name_pattern) {
     PASS();
 }
 
+/* ── Empty-string label is ignored (issue #481) ────────────────── */
+
+/* An empty-string label must behave like an omitted label (no filter), not be
+ * applied as a literal `n.label = ''` that matches nothing. Previously a
+ * name_pattern/qn_pattern search passing label="" returned zero results, while
+ * the BM25 query path ignored the empty label — an inconsistency that made
+ * structural class/service discovery silently fail. */
+TEST(store_search_empty_label_ignored) {
+    int64_t ids[3];
+    cbm_store_t *s = setup_search_store(ids);
+
+    /* name_pattern + label="" must match the same as name_pattern alone. */
+    cbm_search_params_t empty_label = {.project = "test",
+                                       .name_pattern = ".*Submit.*",
+                                       .label = "",
+                                       .min_degree = -1,
+                                       .max_degree = -1};
+    cbm_search_output_t out = {0};
+    int rc = cbm_store_search(s, &empty_label, &out);
+    ASSERT_EQ(rc, CBM_STORE_OK);
+    ASSERT_EQ(out.count, 1);
+    ASSERT_STR_EQ(out.results[0].node.name, "SubmitOrder");
+    cbm_store_search_free(&out);
+
+    /* A non-empty label still filters: ".*Order.*" matches three names but only
+     * OrderService is a Class. */
+    cbm_search_params_t cls = {.project = "test",
+                               .name_pattern = ".*Order.*",
+                               .label = "Class",
+                               .min_degree = -1,
+                               .max_degree = -1};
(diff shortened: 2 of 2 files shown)


halindrome:
## QA Round 1 — CLEAN

**Reviewer model:** claude-opus-4-8 · **Contract source:** issue #481 · **Base:** `main`

### Contract Verification

|---|---|---|
| `name_pattern` + empty-string `label` must not return zero | satisfied | Guard `if (params->label && params->label[0])` (`store.c`) skips the `n.label = ?` clause for `label=""`. Exercised end-to-end: `cbm_mcp_get_string_arg` returns a non-NULL `""` for JSON `"label":""`, so the real tool path hits the guard. |
| Empty `label` behaves like omitted `label` | satisfied | `label==""` and `label==NULL` both skip the clause; other clauses unaffected. Test asserts `name_pattern=".*Submit.*"` + `label=""` → same single result as the pattern alone. |
| Consistent with BM25 `query` path | satisfied | `bm25_search` takes no `label` arg (filters only project + fixed noise-label set); there is no second place applying an empty label. |
| No regression to non-empty `label` | satisfied | Non-empty label still appends/binds `n.label = ?`. Test: `.*Order.*` (3 name matches) + `label="Class"` → only `OrderService`. |
| Covers both `name_pattern` and `qn_pattern` | satisfied | `search_where_basic` is the single shared WHERE builder; both pattern clauses sit after the label clause, one caller chain. |

### Findings

- Contract: **0** · Regression: **0**
- Guard is subtractive for `label==""`, no-op otherwise; no node carries an empty label, so the prior `n.label=''` was strictly a bug. Memory-safe (`[0]` read only after non-NULL check).

### Advisory (non-blocking)

- **F-01 (test gap):** the test comment mentions `qn_pattern` but only `name_pattern` is exercised. Provably covered by the shared builder, but a future `qn_pattern` special-case could silently regress. → addressing with a `qn_pattern` + empty-label assertion.
- **F-02 (edge case, out of scope):** a whitespace-only label (`" "`) is still treated as a real filter. Intentional — the `[0]` guard scopes precisely to the empty string reported in #481; whitespace is not part of the contract.

### Verdict

**CLEAN.** One-line guard fully satisfies #481 for both pattern paths, genuinely matches BM25 behavior, no regression, memory-safe. Regression test fails on old code (`label=""` → `n.label=''` → 0) and passes on new.

halindrome:
Addressed QA round 1 advisory F-01 in 19a0fc5 (test-only): `store_search_empty_label_ignored` now also asserts `qn_pattern` + `label=""` returns the match, locking in the shared-builder guard against a future qn_pattern special-case. F-02 (whitespace-only label) left as-is — intentionally out of scope for #481, which is specifically the empty string. Suite green (the lone unrelated `cli_hook_gate` failure is pre-existing).

PR #485, opened 2026-06-17
fix: Reject filesystem paths in HTTP_CALLS route extraction so client routes can join server routes

### LINKED ISSUE #455: bug: HTTP route heuristic misclassifies filesystem paths; client/server routes never join
## Describe the bug

On a Python repo (~82.6k graph nodes), route / `HTTP_CALLS` extraction produces false and unjoinable data:

- **1062 / 1213** `Route` nodes are empty URL-string-literal stubs (no `method`/`route_path`).
- `url_path` includes filesystem paths misclassified as HTTP — `/root/.aws/credentials`, `/etc/crio/crio.conf` — and `str.split('/locations/')` delimiters; `os.remove`/`os.path.join` are emitted as `HTTP_CALLS`.
- **Client and server route sets are fully disjoint:** `Route<-[:HANDLES]-` ∩ `Route<-[:HTTP_CALLS]-` = **0**. No client call resolves to a handled endpoint; internal URLs (`http://api-admin:8001`) exist only as orphan Route name strings, 0 wired.

## Expected

Route extraction distinguishes HTTP URLs from filesystem paths / string-split args; client `HTTP_CALLS` resolve to server `Route`s (intra-repo at minimum).

## Repro

Index any repo mixing `requests`/HTTP clients with `os.path`/file I/O, then:
```cypher
MATCH (a)-[:HTTP_CALLS]->(b:Route) RETURN b.url_path
```

## Confirmations
- [x] I searched existing issues. Distinct from #292 (cross-repo protocol matching) — this is intra-repo extraction correctness, and a prerequisite for #292's matcher to have clean Route nodes to work with.

mvanhorn:
## Summary
On a large Python repo, HTTP route / HTTP_CALLS extraction emits false data: 1062/1213 Route nodes are empty URL-string stubs with no method/route_path, and filesystem paths like /root/.aws/credentials, /etc/crio/crio.conf and str.split('/locations/') delimiters are misclassified as HTTP routes (os.remove / os.path.join become HTTP_CALLS). Because of this, the client HTTP_CALLS set and the server HANDLES set are fully disjoint: zero client calls resolve to a handled endpoint.

## Changes
In internal/cbm/service_patterns.c the HTTP client heuristic matches callee names by method suffix (.get/.post/...) and library prefix (requests, etc.) and lifts the call's first string argument into url_path. Add a URL-vs-filesystem guard before classifying an argument string as a route: reject absolute filesystem-looking paths (leading / followed by known fs roots like etc/, root/, var/, usr/, home/, tmp/, or a path segment carrying a file extension such as .conf/.credentials/.json on disk) and reject str.split/str.join delimiter literals, so only HTTP-shaped paths produce HTTP_CALLS edges with a populated url_path.

Fixes #455

### DIFF
--- a/internal/cbm/service_patterns.c
+++ b/internal/cbm/service_patterns.c
@@ -550,6 +550,191 @@ static const lib_pattern_t *match_qn(const char *qn, const lib_pattern_t *patter
     return NULL;
 }
 
+static bool starts_with_segment(const char *path, const char *segment) {
+    if (!path || path[0] != '/' || !segment) {
+        return false;
+    }
+    size_t seg_len = strlen(segment);
+    const char *p = path + 1;
+    return strncmp(p, segment, seg_len) == 0 && (p[seg_len] == '\0' || p[seg_len] == '/');
+}
+
+static bool contains_segment(const char *path, const char *segment) {
+    if (!path || !segment) {
+        return false;
+    }
+    size_t seg_len = strlen(segment);
+    const char *p = path;
+    while ((p = strchr(p, '/')) != NULL) {
+        p++;
+        if (strncmp(p, segment, seg_len) == 0 && (p[seg_len] == '\0' || p[seg_len] == '/')) {
+            return true;
+        }
+    }
+    return false;
+}
+
+static bool is_digit_char(char ch) {
+    return ch >= '0' && ch <= '9';
+}
+
+static bool has_http_route_marker(const char *path) {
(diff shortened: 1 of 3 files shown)


mvanhorn:
Thanks @DeusData - rejecting filesystem paths in HTTP_CALLS extraction should clear out a lot of those empty URL-string Route nodes.

PR #486, opened 2026-06-17
fix: Surface a real error when persistence artifact export silently fails on Windows

### LINKED ISSUE #400: Persistence feature (`persistence: true`) not working on Windows
Hi I am trying to implement the local persistent index (.codebase-memory/graph.db.zst) for committing to the repository.  I am unable to repreoduce this on Windows, I am unsure whether it is Mac only or if there are settings that are needed to make sure it works properly.

# Persistence feature (`persistence: true`) not working on Windows

## Environment
- **OS:** Windows 11
- **Shell:** PowerShell
- **Version:** codebase-memory-mcp v0.7.0
- **Installation:** Binary from GitHub releases (`codebase-memory-mcp.exe`)
- **Storage:** Repo-local via `CBM_CACHE_DIR=".codebase-memory"`

## Expected Behavior
According to the `index_repository` tool schema, setting `persistence: true` should:
> Write compressed artifact to .codebase-memory/graph.db.zst for team sharing. Teammates can bootstrap from the artifact instead of full re-indexing.

The response should show `artifact_present: true` after indexing.

## Actual Behavior
- `artifact_present: false` in all responses
- No `.codebase-memory/graph.db.zst` file created
- Only the standard `.db` file is created (path-based naming)

### 1. Set environment variables
```powershell
$env:CBM_CACHE_DIR = ".\.codebase-memory"
$env:CBM_PERSIST_EXPORTS = "true"
```

### 2. Index with persistence enabled
```powershell
$repoPath = (Get-Location).Path.Replace('\', '/')
echo '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"index_repository","arguments":{"repo_path":"'$repoPath'","paths":["src","tests"],"include_patterns":["*.py","*.yaml","*.yml","*.json","*.md",".env*"],"persistence":true}}}' | codebase-memory-mcp
```

### 3. Check response
```json
{
  "project": "C-Users-username-path-to-repo",
  "status": "indexed",
  "nodes": 11557,
  "edges": 20708,
  "artifact_present": false  // ❌ Expected true
}
```

### 4. Check files created
```powershell
Get-ChildItem -Path ".\.codebase-memory" -Filter "*.zst"
# Returns: No files found
```

Only these files exist:
- `C-Users-username-path-to-repo.db` (32 MB) - Standard database
- `_config.db` (12 KB) - Config database
- No `.zst` artifact file

## What We Tried

1. **With `CBM_PERSIST_EXPORTS=true`** - No effect
2. **With `persistence: true` parameter** - No `.zst` file created
3. **Different storage locations** - Same result in both repo-local and AppData
4. **Multiple indexing attempts** - Consistent behavior across all attempts

## Suspected Root Cause

The persistence/artifact feature may be:
1. **Platform-specific** - Only implemented for Linux/Mac (uses Unix-specific compression tools?)
2. **Path-related** - Windows path format (`C:\...`) might break the artifact naming/creation logic
3. **Missing dependency** - Windows binary might be missing zstd compression support

## Impact

Without portable artifacts:
- Teams cannot share pre-built knowledge graphs
- Each developer must re-index (~8 seconds per repo)
- Path-based database naming prevents committing databases to git
- Workaround: Accept local re-indexing or coordinate on standardized absolute paths

## Additional Context

The tool successfully creates and queries the knowledge graph (11,557 nodes, 20,708 edges). Only the persistence/artifact export feature is affected.

Would appreciate clarification on:
1. Is this feature expected to work on Windows?
2. Are there additional dependencies needed for Windows?
3. Is there a workaround or alternative approach for team sharing?

Thank you!

mvanhorn:
## Summary
On Windows, index_repository with persistence:true reports status:indexed but artifact_present:false, and the <repo>/.codebase-memory directory is created with only ADR.md and no graph.db.zst, with no error surfaced. A second user (armandn) reproduced this on 2026-06-17 with a concrete index_repository response (nodes:1866, edges:5340, adr_present:true, artifact_present:false).

## Changes
The artifact export lives in src/pipeline/artifact.c (referenced by pipeline.c, which sets artifact_present in the index_repository response). Audit the Windows code path for the zstd-compressed graph.db.zst write: a failed temp-file create/rename or path-separator/long-path handling likely returns early without propagating an error, leaving artifact_present:false and no log line.

Fixes #400

### DIFF
--- a/src/pipeline/artifact.c
+++ b/src/pipeline/artifact.c
@@ -42,6 +42,8 @@ enum {
 /* Thread-local rotating buffers for small int→string conversions (logging).
  * Rotating allows multiple itoa_buf() calls in a single log statement. */
 enum { ART_RING = 4, ART_RING_MASK = 3 };
+static _Thread_local char g_export_error[CBM_SZ_512];
+
 static const char *itoa_buf(int v) {
     static _Thread_local char bufs[ART_RING][CBM_SZ_32];
     static _Thread_local int idx = 0;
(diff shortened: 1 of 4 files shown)


DeusData:
Thanks @mvanhorn — the error-surfacing work here is solid, and the directory-as-blocker trick to force a rename failure is a clean reproduce-first test.

Two things before merge:

1. `cbm_artifact_export_last_error()` is declared via ad-hoc forward declarations in `pipeline.c` (and the test) rather than a shared header — please expose it in `artifact.h` so there's a single declaration.
2. This changes `dump_and_persist_hashes` to propagate a failed export as a pipeline error (previously it was silently ignored). That's the right call, but please confirm it's intentional and add a short comment — it can now fail a run that used to "succeed" silently.

Also FYI: #492 (the other #400 PR, now closed in favour of this one) used `MoveFileExA(..., MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)` for the Windows replace, which is more atomic than unlink+rename — worth folding that one line into your `write_file_atomic` (credit to @Ayush7Ranjan). 🙏

DeusData:
One more item from a closer pass, in addition to the header/error-propagation notes above: in `read_stripped_db`, the empty/unreadable branch calls `artifact_export_fail(...)` but then continues, so the caller hits a second `artifact_export_fail` for the same condition (double-report / continue-on-error). Worth returning on the first failure so the error is reported once and the path stops cleanly. 🙏

mvanhorn:
Done in 44d5174.

1. Moved the `cbm_artifact_export_last_error()` declaration into `artifact.h` and removed the ad-hoc forward declarations from `pipeline.c` and the test, so there's a single declaration.
2. Confirmed, it's intentional: `dump_and_persist_hashes` now propagates a failed export as a pipeline error instead of silently ignoring it. Added a short comment at the propagation point flagging the behavior change.
3. Folded in @Ayush7Ranjan's `MoveFileExA(..., MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)` for the Windows replace in `write_file_atomic` (the POSIX path keeps plain rename).

The changed files compile clean under the project's -Werror flags and the artifact suite passes locally; the MoveFileExA branch is Windows-only so I'm leaning on CI for that path. Thanks for the thorough review.

PR #506, opened 2026-06-18
fix(routes): canonicalize path parameters so client calls match handlers (#505)

### LINKED ISSUE #505: HTTP route matching never joins client calls to handlers for parameterized paths (:id vs {id} placeholder mismatch)
## Describe the bug

Intra-repo (and cross-repo) HTTP route matching never joins a client call to its
server handler when the path contains a **parameter**, because the route QN is
built from the raw path string and each framework writes path parameters with a
different placeholder syntax. A client call site and the handler that serves it
therefore produce *different* Route QNs and never rendezvous.

Concretely, in a monorepo with a Rust (Axum) API and a TypeScript client:

| Server handler (Axum) | Client call (TS api-client) | Joins today? |
|---|---|---|
| `/players/{id}` | `/players/:id` | ❌ |
| `/clients/{id}/authorized-users` | `/clients/:clientId/authorized-users` | ❌ |
| `/link-requests/{id}/status` | `/link-requests/:requestId/status` | ❌ |
| `/orders/{id}/items/{itemIndex}` | `/orders/:orderId/items/:itemIndex` | ❌ |

**Static** routes (no params) join correctly — e.g. `/link-requests/batch`,
`/clients/me/products`, `/products/categories` resolve client→handler fine. Only
**parameterized** routes fail, and they are the majority of a typical REST surface.

Two independent causes of divergence:
1. **Placeholder syntax** differs: `:id` (Express/React-Router/Rails/JS clients)
   vs `{id}` (Axum/Spring/OpenAPI/ASP.NET) vs `<id>` (Flask/Rocket).
2. **Parameter name** differs even within one syntax: server `{id}` vs client
   `:requestId` for the same logical endpoint.

## Expected

A client `HTTP_CALLS` whose path differs from the server handler's path only in
parameter placeholder syntax and/or parameter name should resolve to the same
`Route` node, so `Route<-[:HTTP_CALLS]-` ∩ `Route<-[:HANDLES]-` is non-empty for
parameterized endpoints.

## Repro

Index a repo that has a server route `GET /players/{id}` and a client call to
`/players/:id`, then:

```cypher
MATCH (c)-[:HTTP_CALLS]->(r:Route)<-[:HANDLES]-(s)
WHERE r.name CONTAINS '{' OR r.name CONTAINS ':'
RETURN r.name, c.qualified_name, s.qualified_name
```

Today this returns 0 rows for parameterized routes; static routes join.

## Root cause

The Route QN (the rendezvous key) is assembled with the raw path at ~11 sites,
none of which canonicalize the parameter segment:

- `src/pipeline/pass_route_nodes.c` (HTTP_CALLS edge, decorator handler, SvelteKit)
- `src/pipeline/pass_parallel.c` (decorator route registration, service edge,
  route registration, `detect_url_in_args`)
- `src/pipeline/pass_calls.c` (route registration, service route node)
- `src/pipeline/pass_cross_repo.c` (cross-repo matcher lookup)

`normalize_url_arg` in `pass_parallel.c` strips JS `${...}` interpolation but
*drops* it rather than canonicalizing, and never handles `:param`.

## Proposed fix

Add a small shared helper `cbm_route_canon_path()` that, per `/`-delimited
segment, collapses any parameter token — `:name`, `{name}`, `<name>` (incl.
`<int:id>`), `${...}` — to a single `{}` token, and apply it to the path right
before each Route-QN `snprintf` on the HTTP-route sites (leaving broker/async/
infra QNs untouched). The display `name` keeps the original path; only the QN
(identity/dedup key) is canonicalized. The existing ANY-vs-method reconciliation
then does the rest, exactly as it already does for static routes.

**Tradeoff (worth a maintainer call):** discarding the parameter *name* collapses
`/users/{id}` and `/users/{slug}` onto the same Route node. For REST surfaces
these are virtually always the same logical endpoint, but it is a deliberate
precision/recall choice and is called out here for review.

This is distinct from #455 (HTTP-vs-filesystem-path misclassification and empty
URL stubs) and #292 (typed clients with no literal path string). Here both sides
*do* emit a valid literal HTTP path; they merely differ in placeholder syntax.

I have a PR with the helper + the 11 call-site changes + a unit-test suite ready
to follow.

imateusdev:
## What does this PR do?

Canonicalizes route-path **parameter placeholders** when building Route QNs so that
a client `HTTP_CALLS` and the server handler that serves it rendezvous on the same
`Route` node even when each side uses a different framework's placeholder syntax
(or a different parameter name).

Fixes #505.

### Problem

Route QNs (`__route__METHOD__/path`) are the rendezvous key, but they were built
from the raw path. A Rust/Axum handler `GET /players/{id}` and a TypeScript client
call `/players/:id` therefore produced **different** QNs and never joined. Static
routes joined fine; parameterized routes (the bulk of a REST surface) did not:

```
Route<-[:HTTP_CALLS]- ∩ Route<-[:HANDLES]-  ==  ∅   (for parameterized paths)
```

### Fix

New shared helper `cbm_route_canon_path()` (declared in `pipeline_internal.h`,
defined in `pass_route_nodes.c`) collapses each parameter token to a single `{}`:

| Input | Canonical |
|---|---|
| `/players/:id` | `/players/{}` |
| `/players/{id}` | `/players/{}` |
| `/users/<int:id>` | `/users/{}` |
| `/players/${id}` | `/players/{}` |
| `/orders/{id}/items/{itemIndex}` | `/orders/{}/items/{}` |

Applied to the path at every **HTTP**-route QN construction site
(`pass_route_nodes.c`, `pass_parallel.c`, `pass_calls.c`, `pass_cross_repo.c`);
broker/async/infra QNs are left untouched. The Route node's display `name` keeps
the original path — only the QN (identity/dedup key) is canonicalized — so the
existing ANY-vs-method reconciliation does the rest, exactly as it already does
for static routes.

### Design note / tradeoff

Parameter **names** are intentionally discarded: `/users/{id}` and `/users/{slug}`
canonicalize to the same node. For REST surfaces these are virtually always the
same logical endpoint; flagging it here for reviewer visibility.

### Tests

`tests/test_route_canon.c` — 11 unit cases covering each placeholder syntax, the
`:id`↔`{id}` convergence invariant, parameter-name agnosticism, literal `:` mid
segment, multiple params, NULL/empty, and tight-buffer truncation safety.

### Verification

- `make -f Makefile.cbm cbm` — clean build (`-Wall -Wextra -Werror`).
- `make -f Makefile.cbm test` — full ASan/UBSan suite; the new `route_canon`
  suite passes 11/11 and no previously-passing test regresses.
- `clang-format-18 --dry-run --Werror` clean on all changed files.

### Checklist

- [x] Tests pass locally
- [x] Lint passes (clang-format) on changed files

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -329,7 +329,7 @@ TEST_DISCOVER_SRCS = \
 
 TEST_GRAPH_BUFFER_SRCS = tests/test_graph_buffer.c
 
-TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c
+TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c
 
 TEST_WATCHER_SRCS = tests/test_watcher.c
 
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -187,8 +187,9 @@ static void handle_route_registration(cbm_pipeline_ctx_t *ctx, const CBMCall *ca
                                       const char **imp_keys, const char **imp_vals, int imp_count) {
     const char *method = cbm_service_pattern_route_method(call->callee_name);
     char route_qn[CBM_ROUTE_QN_SIZE];
+    char cpath[CBM_SZ_256];
     snprintf(route_qn, sizeof(route_qn), "__route__%s__%s", method ? method : "ANY",
-             call->first_string_arg);
+             cbm_route_canon_path(call->first_string_arg, cpath, sizeof(cpath)));
     char route_props[CBM_SZ_256];
     snprintf(route_props, sizeof(route_props), "{\"method\":\"%s\"}", method ? method : "ANY");
     int64_t route_id = cbm_gbuf_upsert_node(ctx->gbuf, "Route", call->first_string_arg, route_qn,
@@ -225,12 +226,15 @@ static int64_t create_svc_route_node(cbm_pipeline_ctx_t *ctx, const char *url, c
                                      const char *method, const char *broker) {
     char route_qn[CBM_ROUTE_QN_SIZE];
     const char *prefix;
+    char cpath[CBM_SZ_256];
+    const char *qpath = url;
     if (svc == CBM_SVC_HTTP) {
         prefix = method ? method : "ANY";
+        qpath = cbm_route_canon_path(url, cpath, sizeof(cpath));
     } else {
         prefix = broker ? broker : "async";
     }
-    snprintf(route_qn, sizeof(route_qn), "__route__%s__%s", prefix, url);
+    snprintf(route_qn, sizeof(route_qn), "__route__%s__%s", prefix, qpath);
     const char *rp;
     if (svc == CBM_SVC_HTTP) {
         rp = method ? method : "{}";
--- a/src/pipeline/pass_cross_repo.c
+++ b/src/pipeline/pass_cross_repo.c
@@ -10,6 +10,7 @@
  * get a CROSS_* edge so the link is visible from either side.
  */
 #include "pipeline/pass_cross_repo.h"
+#include "pipeline/pipeline_internal.h" // cbm_route_canon_path
 #include "foundation/constants.h"
 #include "foundation/log.h"
 #include "foundation/platform.h"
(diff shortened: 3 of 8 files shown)


DeusData:
Thanks @imateusdev — the route-param canonicalization is correct and well-tested. The CI red here is a **cppcheck false-positive**: at `src/pipeline/pass_route_nodes.c:59` it flags `return out` (in the `out == NULL` early-return) as an "uninitialized variable", which left the lint job cancelled and the test job skipped — it's not a real test failure.

Please add a targeted `// cppcheck-suppress` with a one-line `WHY` (returning the NULL pointer is intentional), or restructure the early-return so cppcheck is satisfied. Once lint runs to completion green, the test job unblocks and this is good to merge. 🙏

imateusdev:
Thanks @DeusData! Done — and a heads-up on a second issue I hit along the way.

**1. cppcheck false-positive (`pass_route_nodes.c`)** — fixed with a targeted inline suppression and a `WHY` note, since `out` is just the pointer being returned in the early-out (the buffer is never read here):

```c
if (out == NULL || out_sz == 0) {
    /* WHY: cppcheck value-flow follows a caller that passes an as-yet
     * uninitialized output buffer (e.g. `char cpath[256];`) and flags this
     * `return out` as uninitvar. False positive: we only return the pointer
     * here and never read the buffer; returning it (NULL, or a buffer too
     * small to write into) is the intended early-out contract. */
    // cppcheck-suppress uninitvar
    return out;
}
```

**2. Merge conflict with #485** — while pushing I noticed the branch had gone `CONFLICTING`: #485 (`f5ea36c`) landed on `main` and removed `cbm_route_canon_path` from `pass_route_nodes.c` in favor of the filesystem-path rejection (`cbm_service_pattern_is_http_route_literal`). I rebased onto current `main` and reconciled the two — they compose cleanly: #485's filter decides *which* URLs become routes, then this PR canonicalizes the path params on the survivors:

```c
if (strcmp(edge->type, "HTTP_CALLS") == 0 &&
    !cbm_service_pattern_is_http_route_literal(url, callee)) {
    return;  // #485: reject filesystem paths
}
...
cbm_route_canon_path(url, cpath, sizeof(cpath))  // this PR: canonicalize survivors
```

Force-pushed the rebased commit (`8b81367`). CI is now fully green (lint + the full test matrix across all 5 platforms), and the PR is `MERGEABLE` / `CLEAN`. Should be good to merge 🙏

PR #512, opened 2026-06-19
fix: include git2/sys/alloc.h for git_allocator on libgit2 1.8+

eejd:
## Problem

`git_allocator` was moved from the top-level `git2.h` into `git2/sys/alloc.h`
in libgit2 1.8.0. Building `codebase-memory-mcp` against libgit2 ≥ 1.8 fails:

```
internal/cbm/cbm.c:347:12: error: unknown type name 'git_allocator'
```

## Fix

Add an explicit `#include <git2/sys/alloc.h>` immediately after
`#include <git2.h>` inside the `HAVE_LIBGIT2` guard in
`internal/cbm/cbm.c`. The struct members (`gmalloc`, `grealloc`, `gfree`)
and the `GIT_OPT_SET_ALLOCATOR` option code are unchanged — the only
issue was the missing header.

## Tested against

- libgit2 1.9.4 (MacPorts, macOS / Apple Silicon)

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -20,7 +20,9 @@
 #if defined(CBM_BIND_TS_ALLOCATOR) && CBM_BIND_TS_ALLOCATOR
 #include "sqlite3.h" // sqlite3_mem_methods, sqlite3_config, SQLITE_CONFIG_MALLOC — bind sqlite to mimalloc
 #if defined(HAVE_LIBGIT2)
-#include <git2.h> // git_allocator, git_libgit2_opts, GIT_OPT_SET_ALLOCATOR — bind libgit2 to mimalloc
+#include <git2.h> // git_libgit2_opts, GIT_OPT_SET_ALLOCATOR — bind libgit2 to mimalloc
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks @eejd — the `git2/sys/alloc.h` include for libgit2 1.8+ is correct. The only thing blocking is DCO: the commit's author email doesn't match its `Signed-off-by` email, so the DCO check fails. Could you re-sign with a matching identity — `git commit --amend -s` (with the author email equal to your sign-off email), or `git rebase --signoff` — then force-push? Once DCO is green I'll merge. 🙏

DeusData:
Thanks for the libgit2 compatibility fix. Before review can continue, please add a DCO `Signed-off-by` trailer to the non-merge commit and remove any generated/session attribution from commit metadata.

DeusData:
Thank you for finding and reporting this first — you identified the missing include a full 11 days before anyone else, and that diagnosis is what ultimately got it fixed. We've merged the same fix via #722 (merge efad17da), which was in a mergeable state with a passing DCO check; since the sign-off here never came through and the branch went quiet, we're closing this one as a duplicate rather than leaving it open. The first-finder credit is yours — thanks again for the sharp catch, and we'd be happy to see more contributions from you.

Issue #510, opened 2026-06-18
.gitignore (non repo root) gaps and overrides


s-frantz:
### Version

codebase-memory-mcp 0.8.1

### Platform

Windows (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

Hello, love the work you're putting into this project!

Have one more .gitignore -related suggestion to add alongside #489 & #499, plus a suggestion for #497... (theme being gitignore gaps & overrides) in case there's opportunity to batch all into one PR.

Specifically, I've noticed:
1. `index_repository` seems to miss .gitignore rules (both dir-level and interited from parents) when called on a folder not at repo root, e.g., a subpackage
2. `.cbmignore` might be a useful place to add in overrides as described in #497 but currently (when `index_repository` is called on a valid git repo root), is additive with `.gitignore`

Suggested solution for 1. would be to either warn the user, loudly, when non git repo is passed and gitignore is present (something to the effect of "NOT A GIT REPO -- .GITIGNORE WON'T BE RESPECTED"), or to take on the recursive journey up to find parent git repo (where other generally applied ignore rules might reside), or to do a combination and warn + respect any within-folder .gitignore files.

Suggested for 2 would be to take the same approach as git itself: give .cbmignore priority as last instruction, so any conflicts (e.g., .gitignore shows `dir`, but .cbmignore shows `!dir / !dir/*` -> leads to dir being included in index).

MfG
Silas

## Repro 1 — .gitignore skipped when indexing a non-root subfolder

```
T=$(mktemp -d); cd "$T"
git init -q                                                   # repo root = $T
mkdir pkg
echo 'def secret(): return "SECRET_TOKEN_111"' > pkg/secret.py
echo 'def ok(): return 1'                       > pkg/keep.py
echo 'secret.py' > pkg/.gitignore                             # committed, at the subpackage's own root
git add -A && git commit -qm init

# git itself agrees secret.py is ignored, even addressed from inside pkg:
git -C "$T/pkg" check-ignore secret.py
#   secret.py        ← git says: ignored

# (A) index the SUBPACKAGE (not the repo root):
OUT=$(codebase-memory-mcp cli index_repository "{\"repo_path\":\"$T/pkg\"}")
echo "$OUT"                                                   # note: "excluded" lists no .gitignore'd entry
[...]

Issue #489, opened 2026-06-17
index_repository ignores .git/info/exclude (only .gitignore honored) → indexes excluded worktrees → OOM


madebymlai:
## Summary

`index_repository` honors `.gitignore` but **silently ignores `.git/info/exclude`** (and, as far as I can tell, the global/`core.excludesFile` excludes too). Any directory excluded *only* via `.git/info/exclude` is walked and indexed.

This matters a lot for agent/sandbox workflows: tools like sandcastle put their worktrees under a dir that's excluded via `.git/info/exclude` (not a committed `.gitignore`). The indexer descends into those nested full-repo checkouts and the file count/graph explodes, leading to OOM during indexing.

Real-world trigger in my repo: 493 git-tracked files, but `find` walks **661,430 files / 170 GB** because `.sandcastle/worktrees/*` (excluded only in `.git/info/exclude`) contains 3 nested full-repo worktrees, each ~57 GB with their *own* nested `.sandcastle/` + run-data. The indexer treats `git` as the source of truth for some things but never consults `.git/info/exclude`, so it tries to ingest all of it.

## Reproduction (minimal, deterministic)

```bash
T=$(mktemp -d)
cd "$T"
git init -q
mkdir gi_excluded ie_excluded included
echo 'def fn_gitignore_excluded(): return 1'  > gi_excluded/a.py
echo 'def fn_infoexclude_excluded(): return 2' > ie_excluded/b.py
echo 'def fn_plainly_included(): return 3'     > included/c.py
echo 'gi_excluded/' > .gitignore
echo 'ie_excluded/' >> .git/info/exclude     # git itself excludes this dir
git add -A && git commit -qm init

# git agrees BOTH dirs are excluded:
git check-ignore gi_excluded/a.py ie_excluded/b.py
#   gi_excluded/a.py
#   ie_excluded/b.py

codebase-memory-mcp cli index_repository "{\"repo_path\":\"$T\"}"
```

## Actual behavior

Index output:

```json
"excluded":{"dirs":["gi_excluded",".git"],"count":2,"truncated":false}
```

`ie_excluded/` is **not** in the excluded set, and it gets indexed (`pass.complexity functions=2`). Confirming via `search_graph`:

| dir | excluded via | in graph? |
|-----|--------------|-----------|
| `gi_excluded/` | `.gitignore` | ❌ correctly skipped |
| `ie_excluded/` | `.git/info/exclude` | ✅ **present — bug** |
[...]

PR #517, opened 2026-06-19
fix(discover): honor root .gitignore when indexing subfolders

Bortlesboat:
## Summary
- load `.gitignore` from the indexed path even when that path is a subfolder without its own `.git/`
- add a regression test that indexes `repo/pkg` and verifies `pkg/.gitignore` still excludes `secret.py`
- keep the `.cbmignore` override behavior from issue `#510` out of scope for this PR

## Related
- Addresses repro 1 in #510

## Verification
- `git diff --check`
- focused discover-only runner built under WSL with `cc ... tests/test_discover.c ... src/discover/*.c ...` -> `75 passed`

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -763,7 +763,8 @@ int cbm_discover_ex(const char *repo_path, const cbm_discover_opts_t *opts, cbm_
         return CBM_NOT_FOUND;
     }
 
-    /* Load gitignore sources when a .git directory is present.
+    /* Load the indexed-directory .gitignore unconditionally.  When a .git
+     * directory is present, also load repository-local and global excludes.
      * Sources merged in order (later patterns win on conflict):
      *   1. <repo>/.gitignore        — committed exclusions
      *   2. <repo>/.git/info/exclude — per-clone exclusions, not committed
--- a/tests/test_discover.c
+++ b/tests/test_discover.c
@@ -1008,6 +1008,41 @@ TEST(discover_nested_gitignore_stacks_with_root) {
     PASS();
 }
 
+TEST(discover_root_gitignore_respected_without_git_dir) {
+    char *repo = th_mktempdir("cbm_disc_subpkg_repo");
+    ASSERT(repo != NULL);
+
+    th_mkdir_p(TH_PATH(repo, ".git"));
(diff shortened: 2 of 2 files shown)


DeusData:
Thanks @Bortlesboat — this reviewed clean and is ready, but it now conflicts with `main` (`src/discover/discover.c` shifted under a just-merged change). Could you rebase on latest `main`? One small thing to confirm while you're in there: does `cbm_gitignore_load` return NULL cleanly for a non-existent file? The now-unconditional load relies on that. I'll merge once it's rebased. 🙏

DeusData:
Thanks @Bortlesboat — clean fix and a good reproduce-first test; it just needs a rebase on latest `main` (now conflicting). Rebase and I'll merge. 🙏

DeusData:
Thanks @Bortlesboat — correct bug and an exemplary reproduce-first test: loading `<repo>/.gitignore` even when the indexed path is a subfolder without its own `.git/` is exactly right (repro 1 of #510).

The one blocker is a rebase: `cbm_discover_ex` has changed under you — PR #489 added `.git/info/exclude` + global-excludes handling, and several fixes just merged — so the branch now conflicts. Could you rebase onto current `main` and re-apply the un-gating so the `.gitignore` load happens regardless of `is_git_repo` while preserving the new `.git/info/exclude`/global-excludes logic? The test should carry over unchanged. Happy to merge once it's rebased and green — thanks!

Bortlesboat:
Rebased onto current `main` and force-pushed `bae65df`. I also checked `cbm_gitignore_load`: missing files hit the `fopen` failure path and return `NULL`, so the unconditional `<repo>/.gitignore` load is a no-op when the file is absent. CI is green.

DeusData:
Thank you for the careful work here — and our apologies that the timing worked against you: the root-.gitignore-when-indexing-subfolders behavior landed on main via 3f754cf while this PR was open, so the production half of your diff is now already in the tree. Your regression test is genuinely valuable though (nothing on main covers 'root .gitignore honored with no .git in the indexed dir') — we'll carry it forward with credit. Note #510's other half (parent-repo .gitignore inheritance when indexing repo/pkg) is still unfixed and stays open if you'd like to take a crack at it. Closing this one as superseded — thanks again.

PR #526, opened 2026-06-19
fix(mcp): return valid UTF-8 snippets

### LINKED ISSUE #511: get_code_snippet hangs MCP client for non-UTF-8 (CP949/EUC-KR) source files — raw bytes emitted as invalid UTF-8 in the `source` field
### Version

codebase-memory-mcp 0.8.1

### Platform

Windows (x64)

### Install channel

npm

### Binary variant

standard

**What happened:** `get_code_snippet` returns the file's `source` as raw bytes without transcoding to UTF-8 or sanitizing invalid sequences. When the target symbol lives in a **non-UTF-8 source file (CP949 / EUC-KR**, common in legacy Korean/CJK Windows codebases), the JSON response becomes **invalid UTF-8**. Over the MCP **stdio JSON-RPC** transport this **hangs the client indefinitely** (observed 40+ minutes, never returns). The client session is blocked until interrupted.

The **same call via the CLI returns in ~1 s** — because stdout is a raw byte stream, not a UTF-8-validated channel — with the `source` field containing mojibake (the raw CP949 bytes). This confirms the tool logic is fast; only the MCP transport is broken by the invalid UTF-8.

All other tools (`search_graph`, `query_graph`, `search_code` compact, `index_repository`, etc.) are unaffected. Indexing itself works fine — only `get_code_snippet` (which emits file source) is affected.

**What I expected:** a prompt response carrying **valid UTF-8** JSON (mojibake-free, or at minimum lossy-but-valid UTF-8), so the MCP client does not hang.

**Trigger is encoding, not file size** — it fires at the first invalid byte regardless of file length (reproduced on both a ~1 MB and a ~175 KB source file).

**Likely root cause:** the source bytes are placed into the JSON `source` string un-transcoded. JSON-RPC/MCP requires UTF-8; a raw CP949 lead byte (0xC0–0xFE) is read by the client's UTF-8/stream decoder as the start of a multi-byte sequence, so framing/parse stalls → hang.
(Note: the v0.6.0 changelog *"handle invalid UTF-8 and NaN in serialization"* seems scoped to the layout JSON path; the `get_code_snippet` `source` field still reproduces this on v0.8.1.)

**Suggested fix:** before JSON-serializing file-derived text (notably `source` in `get_code_snippet`, and `search_code` full mode), guarantee valid UTF-8 — minimum: replace invalid byte sequences with U+FFFD (prevents the hang); better: detect source encoding (CP949 / EUC-KR / Shift_JIS / GBK) and transcode to UTF-8 (also fixes the mojibake). Same class as `modelcontextprotocol/servers#2098` and `pi-hole/FTL#2868`.

### Reproduction

Minimal, public repro (no proprietary code):

1. Create a **CP949-encoded** C source with a non-ASCII (Korean) comment. On Windows PowerShell:
   ```powershell
   New-Item -ItemType Directory repro | Out-Null
   $cp949 = [System.Text.Encoding]::GetEncoding(949)
   $src = "int add(int a, int b) { // 입력된 두 값을 더한다`r`n    return a + b;`r`n}`r`n"
   [System.IO.File]::WriteAllBytes("repro/add.c", $cp949.GetBytes($src))
   ```

2. Index (succeeds):
   ```
   codebase-memory-mcp cli index_repository '{"repo_path":"C:/abs/path/repro"}'
   ```

3. Get the qualified_name, then fetch the snippet.
   - **Via CLI** (works, ~1 s, but mojibake; output is NOT valid UTF-8):
     ```
     codebase-memory-mcp cli get_code_snippet '{"project":"<project>","qualified_name":"<qn-of-add>"}'
     ```
   - **Via an MCP stdio client** (Claude Code), call the same `get_code_snippet` with the same args.

**Actual:**
- MCP call: **never returns (hang, 40+ min)** — client session blocked.
- CLI call: returns in ~1 s, exit 0, `source` field contains raw CP949 bytes (mojibake); the emitted JSON is **not valid UTF-8**.

**Expected:** both return promptly with valid-UTF-8 JSON.

### Logs

```text
Byte-level analysis of the CLI response (identical payload to what MCP would transmit), ~8.7 KB:

- `iconv -f UTF-8 -t UTF-8 < out`  → FAILS (response is not valid UTF-8)
- count of U+FFFD (EF BF BD) in output  → 0   (invalid bytes are NOT replaced; passed through raw)
- raw bytes > 0x7F  → ~150, first at byte offset 2058
- example raw bytes: C0 D4 B7 C2 B5 C8  → decodes under CP949 to a Korean word (a source comment emitted untranscoded into the JSON `source` string)

No error is printed; CLI exits 0. The defect is solely that invalid UTF-8 is emitted, which only manifests as a hang on the UTF-8-validated MCP stdio channel.
```

58,494 nodes / 167,888 edges / 2,846 files (also reproduces on a 1,293-node / 32-file subset — trigger is encoding, not scale)

### Confirmations

jstar0:
## Summary

- Sanitize `get_code_snippet` source text to valid UTF-8 before serializing the MCP response.
- Preserve valid UTF-8 bytes and replace only invalid source bytes with U+FFFD.
- Add a regression test for non-UTF-8 source bytes in the snippet file.

Fixes #511.

## Changes

- Adds a small lossy UTF-8 sanitizer for snippet source strings.
- Uses the sanitized copy for the `source` field in `build_snippet_response()`.
- Verifies both the raw MCP response envelope and the embedded snippet JSON parse as strict JSON when the source contains CP949/EUC-KR-like bytes.

## Verification

```bash
make -f Makefile.cbm test
make -f Makefile.cbm lint-format
make -f Makefile.cbm lint-no-suppress
git diff --check
```

I also attempted:

```bash
make -f Makefile.cbm lint-ci
```

That local run stopped at `cppcheck` because `cppcheck` is not installed in my environment.

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -2833,6 +2833,75 @@ static char *resolve_snippet_source(const char *root_path, const char *file_path
     return NULL;
 }
 
+static bool utf8_is_cont(unsigned char c) {
+    return (c & 0xC0) == 0x80;
+}
+
+static char *sanitize_utf8_lossy(const char *s) {
+    enum {
+        UTF8_REPLACEMENT_LEN = 3,
+        UTF8_THREE_BYTE_LEN = 3,
+        UTF8_FOUR_BYTE_LEN = 4,
+        UTF8_FOURTH_BYTE = 3,
+    };
+    if (!s) {
+        return NULL;
+    }
+    size_t len = strlen(s);
+    if (len > (((size_t)-1) - SKIP_ONE) / UTF8_REPLACEMENT_LEN) {
+        return NULL;
+    }
+    char *out = malloc(len * UTF8_REPLACEMENT_LEN + SKIP_ONE);
+    if (!out) {
+        return NULL;
+    }
+
+    const unsigned char *p = (const unsigned char *)s;
+    const unsigned char *end = p + len;
+    unsigned char *dst = (unsigned char *)out;
+    while (p < end) {
+        unsigned char c = *p;
+        size_t n = 0;
+        if (c < 0x80) {
+            n = 1;
+        } else if (c >= 0xC2 && c <= 0xDF && p + 1 < end && utf8_is_cont(p[1])) {
+            n = 2;
+        } else if (c == 0xE0 && p + 2 < end && p[1] >= 0xA0 && p[1] <= 0xBF && utf8_is_cont(p[2])) {
+            n = UTF8_THREE_BYTE_LEN;
+        } else if (c >= 0xE1 && c <= 0xEC && p + 2 < end && utf8_is_cont(p[1]) &&
+                   utf8_is_cont(p[2])) {
+            n = UTF8_THREE_BYTE_LEN;
+        } else if (c == 0xED && p + 2 < end && p[1] >= 0x80 && p[1] <= 0x9F && utf8_is_cont(p[2])) {
+            n = UTF8_THREE_BYTE_LEN;
+        } else if (c >= 0xEE && c <= 0xEF && p + 2 < end && utf8_is_cont(p[1]) &&
+                   utf8_is_cont(p[2])) {
+            n = UTF8_THREE_BYTE_LEN;
+        } else if (c == 0xF0 && p + UTF8_FOURTH_BYTE < end && p[1] >= 0x90 && p[1] <= 0xBF &&
+                   utf8_is_cont(p[2]) && utf8_is_cont(p[UTF8_FOURTH_BYTE])) {
+            n = UTF8_FOUR_BYTE_LEN;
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks @jstar0 — your reproduce-first test here (write known-bad CP949 bytes, assert the response is valid JSON with U+FFFD replacements) is excellent and we want to keep it.

There's a parallel PR #541 fixing the same issue (#511) with a cleaner sanitizer (canonical `(c & 0xF0) == 0xE0` lead-byte checks). Our plan is to land **one** combined fix: #541's sanitizer + your test + an explicit overflow guard, crediting both of you.

One change needed here regardless: the overflow guard `len > (((size_t)-1) - SKIP_ONE) / UTF8_REPLACEMENT_LEN` hinges on the opaque `SKIP_ONE` macro — please make the `+ 1` explicit (with a comment) so the bound is auditable without chasing the macro.

Would you be open to combining your test with #541's sanitizer? 🙏

PR #537, opened 2026-06-20
fix: unwatch deleted projects to prevent zombie reindex

Golevka2001:
## Problem

When `auto_index` is enabled, opening a project directory triggers auto-indexing and registers the project with the file-change watcher. If the user later deletes that project via the `delete_project` MCP tool, the `.db` file is removed but the watcher entry persists.

The watcher continues polling the deleted project's git repository every 5–60 seconds. On each poll, it detects HEAD movement or a dirty working tree and invokes the full reindex pipeline — which operates on a database that no longer exists.

**Reproduction scenario:**

1. Enable auto-index: `codebase-memory-mcp config set auto_index true`
2. Start the MCP server from inside a git project directory
3. The server auto-indexes and the watcher begins polling
4. Delete the project: `delete_project(project="<name>")`
5. Continue working in the same directory (git commit, pull, etc.)
6. The watcher detects changes and attempts to reindex → error logs, wasted CPU, and unnecessary I/O on every poll cycle

`cbm_watcher_unwatch()` is fully implemented (`watcher.c:316–332`) but is never called from any production code path — only from tests.

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -1857,6 +1857,11 @@ static char *handle_delete_project(cbm_mcp_server_t *srv, const char *args) {
     }
 
     cbm_pipeline_unlock();
+
+    if (srv->watcher) {
+        cbm_watcher_unwatch(srv->watcher, name);
+    }
+
     cbm_mem_collect(); /* return freed pages to OS after closing database */
 
     yyjson_mut_doc *doc = yyjson_mut_doc_new(NULL);
--- a/src/watcher/watcher.c
+++ b/src/watcher/watcher.c
@@ -276,6 +276,9 @@ void cbm_watcher_free(cbm_watcher_t *w) {
     if (!w) {
         return;
     }
+    /* Safety net: ensure stopped is set before draining pending_free.
+     * In production the caller should cbm_watcher_stop() + join first. */
+    atomic_store(&w->stopped, 1);
     cbm_mutex_lock(&w->projects_lock);
     cbm_ht_foreach(w->projects, free_state_entry, NULL);
     cbm_ht_free(w->projects);
--- a/src/watcher/watcher.c
+++ b/src/watcher/watcher.c
@@ -53,6 +53,10 @@ struct cbm_watcher {
     CBMHashTable *projects; /* name → project_state_t* */
     cbm_mutex_t projects_lock;
     atomic_int stopped;
+    /* Deferred-free list: freed after the next poll_once. */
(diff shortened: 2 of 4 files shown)


Golevka2001:
Mind if I add a pending-free list? Unwatch defers the free there, poll_once drains it under lock before snapshotting—avoids the UAF without refcounting.

Golevka2001:
I have formatted the file with `clang-format`, please re-trigger the CI check.

DeusData:
Thanks @Golevka2001 — unwatching deleted projects is the right fix and the deferred-free design is sound. One concern before merge: `cbm_watcher_free` also drains/frees `pending_free`, so if it can run while `cbm_watcher_poll_once` is mid-cycle there's a use-after-free / double-free window on the `pending_free` array. Please confirm the watcher thread is stopped/joined before `cbm_watcher_free` frees it (or guard the free with `projects_lock`). A small test exercising the delete→unwatch path would also help. 🙏

DeusData:
Thanks @Golevka2001 — real bug and a clean fix: `cbm_watcher_unwatch` existed but had no production caller, so deleting a project left the watcher polling and re-indexing a deleted DB. Wiring it into `handle_delete_project` and using a deferred-free against the lock-free poll loop is the right approach (the one-cycle deferral is sound for the single watcher thread).

Two things before it can land:
1. **Regression test**: please add one exercising the fix — register a watch, run the `delete_project` path, and assert `cbm_watcher_watch_count` drops to 0 (ideally also a poll-after-unwatch to guard the deferred-free). The existing watcher tests are a good template.
2. **CI hasn't run** on this PR yet (no checks present — likely first-time-contributor workflow approval); it needs to be green before merge.

Nice catch on the zombie-reindex — looking forward to the test.

Golevka2001:
Fixed. cbm_watcher_free now sets stopped=1 as a safety net, documented the precondition in the header, and added a test for the unwatch -> pending_free drain path. 😃

DeusData:
Thank you — this is a real bug with a thoughtful fix. We verified the zombie is worse than reported: the watcher's index_fn recreates the deleted .db, so deleted projects actually resurrect. Your deferred-free design is the right call given poll_once polls outside the lock (a naive immediate free would be a genuine cross-thread UAF). Two follow-ups we'll track separately: the UI DELETE /api/project path has the same zombie (no watcher access at all), and the realloc-failure fallback frees immediately (reintroducing the UAF in the OOM corner) — neither blocks this. Merging.

PR #538, opened 2026-06-20
fix(mcp): trace_path cross_service follows CROSS_* edges (#522)

### LINKED ISSUE #522: trace_path mode=cross_service does not traverse CROSS_* edges
A `CROSS_HTTP_CALLS` edge from a function to a provider is visible through `query_graph`, but `trace_path` with `mode=cross_service` from that same function does not surface the cross hop. So when cross-service edges exist, the cross-service trace does not follow them.

Repro (self-contained): https://gist.github.com/azuttre/21e5e85f7590261721e3af7aa079dbfb → `issue-2-trace_path-ignores-cross-edges.sh`. It inserts one `CROSS_HTTP_CALLS` edge in the format the cross-repo pass writes, then shows:

- A) `query_graph` with `MATCH (s)-[e:CROSS_HTTP_CALLS]->(t)` returns the edge.
- B) `trace_path` with `mode=cross_service` from the source function returns only in-repo callees, with no hop to the provider.

Expected: B surfaces the cross hop. Got: it does not.

Env: v0.8.1, macOS (arm64).

kirilklein:
## What does this PR do?

Fixes #522 — `trace_path` with `mode=cross_service` did not follow cross-repo edges, so cross-service traces silently dropped every cross hop even when the edge existed and was visible via `query_graph`.

**Root cause.** `resolve_trace_edge_types()` in `src/mcp/mcp.c` resolved the `cross_service` default edge-type list to `{HTTP_CALLS, ASYNC_CALLS, DATA_FLOWS, CALLS}` — none of the `CROSS_*` types the cross-repo pass actually writes. `cbm_store_bfs` filters edges with an exact `WHERE e.type IN (...)` clause, so any `CROSS_HTTP_CALLS` / `CROSS_ASYNC_CALLS` / … hop was filtered out. The traversal engine was fine; the defaults list was just missing the cross types.

**Fix.** Add the six `CROSS_*` types (`CROSS_HTTP_CALLS`, `CROSS_ASYNC_CALLS`, `CROSS_CHANNEL`, `CROSS_GRPC_CALLS`, `CROSS_GRAPHQL_CALLS`, `CROSS_TRPC_CALLS`) to the `cross_service` default list — exactly the set `pass_cross_repo.c` emits and `cross_repo_delete_edges()` cleans up. The count is now derived from the array size (`sizeof`) so it can't drift; the now-dead `MCP_N_DEFAULTS_4` constant is removed. The `trace_path` tool description is updated to match.

### Before / after (repro from the issue)
- **Before:** `query_graph` returns the `CROSS_HTTP_CALLS` edge, but `trace_path(mode=cross_service)` from the source function returns only in-repo callees.
- **After:** `trace_path(mode=cross_service)` surfaces the cross hop; `query_graph` still returns the edge.

### Test
Adds `integ_mcp_trace_path_cross_service`: seeds a `CROSS_HTTP_CALLS` edge between two indexed functions that have no `CALLS` relationship, then asserts `cross_service` surfaces the hop while `calls` mode does not (proving the cross edge specifically is what's followed). Verified to **fail before** the fix and **pass after**.

## Checklist

- [x] Tests pass locally (`make -f Makefile.cbm test`) — 5609 passed
- [x] Lint passes — `clang-format --dry-run --Werror` clean on changed files (full `lint-ci` deferred to CI)

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -29,7 +29,6 @@ enum {
     MCP_DEFAULT_LIMIT = 10,
     MCP_BFS_LIMIT = 100,
     MCP_N_DEFAULTS_2 = 2,
-    MCP_N_DEFAULTS_4 = 4,
     MCP_URI_PREFIX = 7,      /* strlen("file://") */
     MCP_CONTENT_PREFIX = 15, /* strlen("Content-Length:") */
     MCP_RETURN_2 = 2,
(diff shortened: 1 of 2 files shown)


azuttre:
Built this branch on macOS (arm64) and tested it: with a node reachable only through a CROSS_HTTP_CALLS edge, trace_path cross_service returns 0 callees on main and 1 on this branch. Confirmed working. Thanks for the quick turnaround.

prakashelango:
I have a kinesis stream for eventing and it couldn't recognize the cross service calls from upstream and downstream, by any chance will this change include the kinesis too?

PR #539, opened 2026-06-20
fix(pipeline): preserve ADR across full re-index (#516)fix(pipeline): preserve ADR across full re-index (#516)

### LINKED ISSUE #516: manage_adr data loss: ADRs in project_summaries are deleted during graph re-indexing
### Version

 0.8.1

### Platform

macOS (Apple Silicon)

### Install channel

npm

### Binary variant

standard

## Description
There is a data-loss bug in how the database dump pipeline handles the `project_summaries` table. When an Architecture Decision Record (ADR) is stored using the `manage_adr` tool, it correctly saves to the `project_summaries` table. However, any subsequent indexing operation (`index_repository` / pipeline execution) completely overwrites the SQLite database file and fails to preserve the `project_summaries` table, permanently deleting the ADR.

## Root Cause Analysis
The issue stems from the custom SQLite writer implementation used to dump the graph buffer to disk.

* In `src/pipeline/pipeline.c` (and `internal/cbm/sqlite_writer.c`), when the pipeline finishes, it opens the target `.db` file using `fopen(path, "wb")`, truncating the existing database.
* The monolithic `cbm_writer_finalize` (and `write_metadata_tables`) explicitly writes an empty B-tree for the `summaries_root`:
  ```c
  // internal/cbm/sqlite_writer.c
  *summaries_root = write_table_btree(w->fp, &w->next_page, NULL, NULL, NULL, 0, false);
  ```
* While the incremental pipeline attempts to preserve `file_hashes` by reopening the new DB and re-inserting them, no such preservation logic exists for `project_summaries`.

## Expected Behavior
The database dump pipeline should migrate or preserve the contents of the `project_summaries` table across re-indexing operations, ensuring that ADRs persist indefinitely across sessions as designed.

## Actual Behavior
The `project_summaries` table is wiped clean on every full dump, resulting in silent data loss.

## Steps to Reproduce
1. Index a repository to create the initial database.
2. Use the MCP tool to store an ADR:
   `manage_adr(project="<project_name>", mode="update", content="# My ADR...")`
3. Verify the ADR was stored:
   `manage_adr(project="<project_name>", mode="get")` (Returns the expected content).
4. Trigger a graph re-index (e.g., via `index_repository`).
5. Attempt to retrieve the ADR again:
   `manage_adr(project="<project_name>", mode="get")` (Returns `no_adr`).

### Logs

```text

```

_No response_

### Confirmations

RithvikReddy0-0:
## What does this PR do?

Fixes #516. ADRs stored via `manage_adr` were silently deleted on a full
re-index. Captures the ADR before the DB is deleted and restores it after the
rebuild. Only the full-dump path was affected (incremental never rewrites the DB).

Verified: index → store ADR → change/add files → re-index → ADR survives
(previously returned `no_adr`).

## Checklist

> Built and manually verified in WSL; ran clang-format. Full `make test` requires
> the Unix toolchain and timed out locally — happy to add a regression test if
> pointed to the right test file.

### DIFF
--- a/src/pipeline/pipeline.c
+++ b/src/pipeline/pipeline.c
@@ -97,6 +97,10 @@ struct cbm_pipeline {
     /* Committed graph size at dump time (-1 = dump did not run). #334 gate axis. */
     int committed_nodes;
     int committed_edges;
+
+    /* ADR (project_summaries) captured before a full-reindex DB delete, so it
+     * can be restored after the rebuild. NULL when no ADR existed. Issue #516. */
+    char *saved_adr;
 };
 
 /* ── Global pkgmap (one active pipeline at a time) ─────────────── */
@@ -810,6 +814,22 @@ static int try_incremental_or_delete_db(cbm_pipeline_t *p, cbm_file_info_t *file
         cbm_store_close(check_store);
     }
     cbm_log_info("pipeline.route", "path", "reindex", "action", "deleting old db");
+    /* Capture any ADR before deleting the DB so the full-reindex rebuild can
+     * restore it (project_summaries is otherwise lost). Issue #516. */
+    {
+        cbm_store_t *adr_store = cbm_store_open_path(db_path);
+        if (adr_store) {
+            cbm_adr_t existing;
+            if (cbm_store_adr_get(adr_store, p->project_name, &existing) == CBM_STORE_OK) {
+                if (existing.content) {
+                    free(p->saved_adr);
+                    p->saved_adr = strdup(existing.content);
+                }
+                cbm_store_adr_free(&existing);
+            }
+            cbm_store_close(adr_store);
+        }
+    }
     cbm_unlink(db_path);
     char wal[PL_WAL_BUF];
     char shm[PL_WAL_BUF];
@@ -869,6 +889,11 @@ static int dump_and_persist_hashes(cbm_pipeline_t *p, const cbm_file_info_t *fil
     cbm_store_t *hash_store = cbm_store_open_path(db_path);
     if (hash_store) {
         cbm_store_delete_file_hashes(hash_store, p->project_name);
+
+        /* Restore the ADR captured before the dump. Issue #516. */
+        if (p->saved_adr) {
+            cbm_store_adr_store(hash_store, p->project_name, p->saved_adr);
+        }
         for (int i = 0; i < file_count; i++) {
             struct stat fst;
             if (stat(files[i].path, &fst) == 0) {
@@ -895,6 +920,8 @@ static int dump_and_persist_hashes(cbm_pipeline_t *p, const cbm_file_info_t *fil
         cbm_store_close(hash_store);
         cbm_log_info("pass.timing", "pass", "persist_hashes", "files", itoa_buf(file_count));
     }
+    free(p->saved_adr);
+    p->saved_adr = NULL;
 
     /* Export persistent artifact if enabled */
     if (p->persistence) {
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks @RithvikReddy0-0 — the ADR save/restore around the delete/rebuild path is the right fix for #516.

Two things before merge:

1. **Add a reproduce-first test** (the checklist boxes are unchecked) — index a repo with an ADR, trigger a full re-index, assert the ADR survives. That's the regression guard the project requires for every fix.
2. Please double-check the lifetime of `p->saved_adr`: on an early-exit path (an error before `dump_and_persist_hashes`), is it freed? If the pipeline teardown doesn't `free(p->saved_adr)`, that's a leak — add the free on the appropriate path.

(For info: the doubled commit title is just a cosmetic git artifact — the diff itself is clean and not duplicated.) 🙏

DeusData:
Confirming the two concerns from a closer pass:

1. The `saved_adr` leak is real — on the `cbm_gbuf_dump_to_sqlite` failure early-return in `dump_and_persist_hashes`, `p->saved_adr` is freed neither there nor in `cbm_pipeline_free`, so it leaks on that error path. Free it on that path (or add it to `cbm_pipeline_free`).
2. `cbm_store_adr_store`'s return value isn't checked, so a failed restore silently drops the ADR — the same symptom as the original bug. Please surface that error.

Both can go in alongside the reproduce-first test. 🙏

RithvikReddy0-0:
Thanks — all three addressed in 2f434d4.

**Leak.** `free(p->saved_adr)` now lives in `cbm_pipeline_free`, so it's released on every exit path including the `cbm_gbuf_dump_to_sqlite` failure early-return. The successful-restore path still sets it to NULL after freeing in `dump_and_persist_hashes`, so there's no double-free.

**Unchecked restore.** `cbm_store_adr_store`'s return is now checked; a failed restore logs `pipeline.err phase=adr_restore` instead of silently dropping the ADR.

**Test.** Added `pipeline_adr_survives_full_reindex` (reproduce-first): index a repo, store an ADR, force a full re-index by adding files past the incremental threshold, assert the ADR is present and unchanged. Passes against the fix (`pipeline_adr_survives_full_reindex  PASS`).

One note from running the full suite locally: `5619 passed, 1 failed`, and the single failure is `incr_full_index` (`tests/test_incremental.c:302` — RSS delta 2560MB vs 2048MB ceiling on a 1100-file project). That's unrelated to this change — the fix only adds a `strdup` of an ≤8KB ADR string, which can't move RSS by ~512MB. It's the ASAN/UBSan-instrumented test binary inflating memory past a threshold tuned for a native non-instrumented run, and the file isn't touched by this PR. Flagging it for transparency rather than claiming a clean sweep.

DeusData:
Thanks @RithvikReddy0-0 — all three review items are addressed now (the `saved_adr` leak on the dump-failure path, the unchecked `cbm_store_adr_store` return, and the reproduce-first test), and CI is green. It just needs a rebase on latest `main` (currently conflicting). Rebase and I'll merge. 🙏

RithvikReddy0-0:
Rebased on latest main conflicts resolved (both were additive: kept the `committed_nodes/edges` fields alongside `saved_adr`, and both test registrations). Builds clean and the ADR test still passes. Ready when you are.

DeusData:
Thanks @RithvikReddy0-0 — good catch and a clean fix. A full re-index unlinks and rebuilds the DB from the graph buffer, which doesn't carry the `project_summaries` ADR, so stored ADRs were silently lost (returning `no_adr` afterward). Capturing the ADR before the unlink and restoring it after the rebuild is the right approach, and `pipeline_adr_survives_full_reindex` is properly reproduce-first. The error-path frees and the log-on-failed-restore are a nice touch. Merging — and thanks for #516.

PR #542, opened 2026-06-21
fix: honor git global core.excludesFile as an ignore source

### LINKED ISSUE #499: index_repository silently ignores git global core.excludesFile — globally-excluded files are indexed
### Version

latest (main, as of 2026-06-18)

### Platform

macOS (Apple Silicon)

### Install channel

### Binary variant

standard

**Actual behavior:** When indexing a repository, the tool reads only per-directory `.gitignore` files committed inside the repo. The global exclude file referenced by `git config core.excludesFile` (e.g. `~/.gitignore_global` on macOS/Linux, `%USERPROFILE%\.gitignore_global` on Windows) is silently ignored.

Files the user has told git to exclude globally — `.DS_Store`, `Thumbs.db`, `.idea/`, `*.pyc`, `__pycache__/`, IDE config directories — are **fully indexed** and appear in the knowledge graph.

**Expected behavior:** The indexer should honour all three layers of git's `--exclude-standard` exclude chain:
1. Per-directory `.gitignore` files ✓ (works today)
2. `$GIT_DIR/info/exclude` (tracked separately in #489)
3. Global `core.excludesFile` (this issue — currently unhandled)

`git ls-files --exclude-standard` already reads all three layers; the indexer's tree walk should match that behaviour.

### Reproduction

1. Configure a global gitignore and add a pattern:
   ```
   git config --global core.excludesFile ~/.gitignore_global
   echo "*.pyc" >> ~/.gitignore_global
   ```

2. Create a minimal test repo with a `.pyc` file **not** listed in any in-repo `.gitignore`:
   ```
   mkdir /tmp/testrepo && cd /tmp/testrepo && git init
   touch foo.py foo.pyc
   ```

3. Index the repo:
   `codebase-memory-mcp index_repository '{"project":"/tmp/testrepo"}'`

4. Query for the excluded file:
   `codebase-memory-mcp cli search_graph '{"project":"/tmp/testrepo","name_pattern":"foo.pyc"}'`

**Expected:** no results — `foo.pyc` should be excluded by the global gitignore.
**Actual:** `foo.pyc` appears as a graph node — `core.excludesFile` was not consulted.

**Cross-check:** `git -C /tmp/testrepo ls-files --exclude-standard --others` returns nothing for `foo.pyc`, confirming git itself respects the global exclude file at all three layers.

### Logs

```text

```

_No response_

### Confirmations

mvanhorn:
## What does this PR do?

Honors git's global `core.excludesFile` (including the XDG default `~/.config/git/ignore`) as a repo-root ignore source during discovery, matching git's own ignore precedence. Previously only in-repo `.gitignore` / `.git/info/exclude` were consulted, so files a user globally ignores were still indexed.

This resolves the half of #499 the maintainer separated from the in-flight `.git/info/exclude` work in #493: the global excludes chain (explicit `core.excludesFile`, then `$XDG_CONFIG_HOME/git/ignore`, then `~/.config/git/ignore`) is now read and applied at the repo root.

Closes #499

## Checklist

AI was used for assistance.

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -13,14 +13,18 @@
 
 #include "foundation/constants.h"
 #include "foundation/compat_fs.h"
+#include "foundation/platform.h"
 #ifdef _WIN32
 #include "foundation/win_utf8.h"
 #endif
+#include <ctype.h>
 #include <stdint.h> // int64_t
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h> // strdup
 #include <sys/stat.h>
+
+int cbm_gitignore_match_result(const cbm_gitignore_t *gi, const char *rel_path, bool is_dir);
 
 /* ── Hardcoded always-skip directories ──────────────────────────── */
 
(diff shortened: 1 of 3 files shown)


DeusData:
Hi @mvanhorn — thank you for this. Honoring git's global excludes (#499) is a real gap, and your layering (`~/.gitconfig`, XDG, plus `.cbmignore` negation) is the right shape. The tests are thorough, too.

Before we can land it, there's one security issue in the resolution order we need to sort out. `resolve_global_excludes_path()` reads `core.excludesFile` from the **repo-local `.git/config` first**, and `expand_git_path()` accepts any absolute path, which then gets opened by `cbm_gitignore_load()` → `fopen()` with no containment check.

Because cbm indexes arbitrary directories on request (an MCP client/agent can point it at any path, including an untrusted repo someone just cloned), a repository can ship a `.git/config` like:

```ini
[core]
    excludesFile = ~/.ssh/id_rsa
```

and simply indexing that repo will open and read that file. git itself is safe here because it only reads a repo's `.git/config` after you've deliberately cloned/init'd it — that trust step doesn't exist for an indexer pointed at an arbitrary tree.

The direction we'd suggest:

- **Drop the repo-local `.git/config` pass** and resolve `excludesFile` only from the user-owned global configs (`~/.gitconfig` and `$XDG_CONFIG_HOME/git/config`). Those are the files the user controls, and they're what #499 is actually asking for. A repo's own ignores are already handled by `.gitignore` / `.cbmignore`.
- If you do want to keep any repo-sourced path, **resolve it and verify it stays within the repository root before opening it**, and reject anything that escapes the tree.
- Add a guard test: a repo-local `.git/config` whose `excludesFile` points outside the repo (e.g. into `$HOME`) must be **ignored**, not read.

One note on timing: we think this is worth doing, but we're capacity-constrained right now, so even once the security fix is in it may take a while to land in a release. If others want it, a 👍 on #499 helps us prioritize. And if you'd rather not carry the rework, we're happy to build on top of your branch and credit you — just say the word.

Really appreciate the PR and the careful tests. 🙏

DeusData:
Follow-up after a closer security review — I want to **reframe** my earlier comment.

On reflection this isn't a blocking vulnerability: reading `core.excludesFile` from the repo-local `.git/config` mirrors git's own `--exclude-standard` behavior, the file is opened read-only, its contents aren't returned to the caller, and planting a malicious `.git/config` already implies local file access.

So please treat the earlier note as **recommended hardening / defense-in-depth, not a blocker**: restricting a repo-local `excludesFile` to paths within the repo (or only honoring the user-owned global configs) is still a nice-to-have given cbm can be pointed at untrusted trees — but it's optional, and the feature itself is fine. Thanks again, and apologies for the over-strong initial framing. 🙏

mvanhorn:
Done in 9f66e42.

- Dropped the repo-local `.git/config` pass entirely in `resolve_global_excludes_path()`. `core.excludesFile` now resolves only from the user-owned configs (`~/.gitconfig`, then `$XDG_CONFIG_HOME/git/config`), so an untrusted repo can no longer point it at something like `~/.ssh/id_rsa` and have cbm read it during indexing. A repo's own ignores stay handled by `.gitignore` / `.cbmignore`, as you said.
- Added a guard test (`discover_repo_local_excludesfile_is_ignored`): a repo-local `.git/config` whose `excludesFile` points into `$HOME` is ignored, and the file it tried to hide is still discovered.

Verified locally with the discover suite under ASan/UBSan (319 passing, including the new test). No rush on landing given your capacity note, and happy to leave it here as-is. Thanks for the careful review.

Issue #500, opened 2026-06-18
Please make ALWAYS_SKIP_DIRS configurable by user


zeronewb:
### What problem does this solve?

Rignt now codebase-memory-mcp excluding part of repo blindly, for example:
"nixl/src/plugins/obj","nixl/test/gtest/unit/obj" from https://github.com/ai-dynamo/nixl

### Proposed solution

Make it configurable like .gitignore

### Alternatives considered

Compiling by hand and replacing skip list not very fun.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
Thanks for raising this, @zeronewb 🙏 It's a fair point. Today `.cbmignore` (gitignore syntax) lets you add excludes, but it can't un skip a directory the built in list drops, which is exactly your case. That also overlaps with #497.

The clean design is to make the skip list an overridable default (a config key plus `.cbmignore` negation like `!obj/`) rather than a hardcoded constant. We'll treat #500 and #497 together as making the ignore and skip layer fully user controllable. Appreciate the report!

mvanhorn:
Implemented in https://github.com/DeusData/codebase-memory-mcp/pull/543

DeusData:
Cross-linking #403 as a concrete use case for the deferred config-key half here: there, the IDE install directory gets auto-indexed as a **root**, so the user override needs to be settable globally (in `~/.config/codebase-memory-mcp/config.json`) and consulted during auto-index **root selection**, not just during the subdirectory walk. The `.cbmignore` `!negation` work in #543 covers the in-walk half but not root selection.

Issue #497, opened 2026-06-18
Support .ignore files that act as a local override layer on top of .gitignore.


keeferwu:
### Version

codebase-memory-mcp 0.8.1

### Platform

Linux (x64)

### Install channel

npm

### Binary variant

standard

### What happened, and what did you expect?

Support .ignore files (using the same pattern syntax as .gitignore) that act as a local override layer on top of .gitignore. This is a well-established pattern used by tools like  fd, ag, and rg:

.gitignore — patterns to ignore (unchanged, respected by CodeGraph as today)
.ignore — patterns to override .gitignore with negation patterns (! prefix)

For example, if .gitignore contains:
customer/

And the user creates .ignore with:
!customer/

Then codebase-memory-mcp would include the customer/ directory in its index even though it's gitignored.

### Reproduction

none

### Logs

```text
none
```

### Confirmations

- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

DeusData:
Thanks for the suggestion, @keeferwu 🙏 We relabeled this from bug to enhancement, since it's asking for a new ignore layer convention rather than fixing broken behavior.

It's the same underlying need as #500: a way to un ignore paths the defaults drop. `.cbmignore` already exists (gitignore syntax) but is additive only today; honoring negation (`!customer/`) would give you exactly the `rg` and `fd` style `.ignore` override behavior, without introducing a new filename. Would `.cbmignore` negation cover your use case, or do you specifically need the `.ignore` filename for tool compatibility? Happy to fold this into the #500 work either way!

keeferwu:
Thanks for the detailed response! To answer your question directly: I do specifically need the .ignore filename, not just .cbmignore negation. Here's why:

1. .ignore is shared across major search tools

.ignore is a de-facto convention honored by rg, ag, fd. Each tool also has its own tool-specific ignore file (.rgignore, .agignore, .fdignore), but .ignore is the common one that lets a single config apply across all of them.

2. Avoiding ignore-file proliferation

As a user, I don't want to maintain a pile of overlapping ignore files for each tool. A single unified .ignore keeps things simple and easy to maintain.

3. Why negation is needed

[...]

PR #543, opened 2026-06-21
fix: .cbmignore negation overrides built-in skip directories

mvanhorn:
## What does this PR do?

Lets a `.cbmignore` negation pattern (e.g. `!obj/`) un-skip a directory that is in the built-in `ALWAYS_SKIP_DIRS` set, so users can opt a normally-skipped directory back into discovery. This ships the negation-override half of #500 the maintainer described; the joint config-key default-override (#497) is deferred.

Refs #500

## Checklist

AI was used for assistance.

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -419,6 +419,11 @@ typedef struct {
     int excluded_cap;
 } file_list_t;
 
+typedef struct {
+    cbm_gitignore_t *ignore;
+    cbm_gitignore_t *negation_overrides;
+} cbmignore_matchers_t;
+
 static void file_list_add_excluded(file_list_t *fl, const char *rel_path) {
     if (!rel_path || rel_path[0] == '\0') {
         return;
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks @mvanhorn — this reviewed clean and is ready (the negation-override logic is sound and well-tested). It now conflicts with `main` (`src/discover/discover.c`). Could you rebase on latest `main`? I'll merge once it's conflict-free. 🙏

DeusData:
Thanks @mvanhorn — this is clean and well-tested, and I appreciate the scope discipline (deferring the #497 half). It's now conflicting with `main` — could you rebase? I'll merge once it's conflict-free. 🙏

mvanhorn:
Thanks @DeusData! Rebased onto latest `main`. The conflicts were in `discover.c` and `test_discover.c` where your `global_gi` work and this PR's `cbmignore_negation_overrides` both extended the same `should_skip_directory`/`walk_dir` signatures and the discover test helpers. Resolved by threading both parameters through every signature and call site, and keeping both sets of tests (your `.git/info/exclude` tests and the negation tests) under the unified `discover_has_rel_path` helper. All discover tests pass locally and the touched files are clang-formatted. Should be conflict-free now.

DeusData:
Thanks @mvanhorn — the negation mechanism is well-built (the inverted-matcher approach preserves gitignore last-match-wins semantics) and nicely tested. One thing to resolve before this can land:

`.cbmignore` is loaded from the **indexed repo itself** (`src/discover/discover.c:811`, `<repo>/.cbmignore`), so this feature lets any repo we index re-include the builtin-protected directories in `ALWAYS_SKIP_DIRS` — including `.git`, `.claude`, `node_modules`, `.venv`/`venv`, `site-packages`, and `vendor`. That's a footgun:
- `!.claude/` / `!.git/` would walk agent-config and VCS internals;
- `!node_modules/` / `!site-packages/` can blow up indexing time/memory on real repos.

Since the file ships inside the (potentially third-party) repo, this is attacker-influenced, so we'd prefer a guard over relying on users. Could you:
1. Add a hard deny-subset that negation can **never** un-skip — at minimum VCS (`.git`/`.hg`/`.svn`), dependency dirs (`node_modules`/`.venv`/`venv`/`site-packages`/`vendor`), and agent/IDE config (`.claude`). Build/output dirs like `obj`/`dist`/`target` are the legitimate use case and can stay un-skippable.
2. Add tests asserting `!node_modules/` and `!.git/` remain skipped even with negation.

Minor: since this adds capability (refs #500) rather than fixing a defect, `feat:` fits better than `fix:`. Happy to merge once the deny-list is in — thanks again!

DeusData:
Thank you — the problem is real and your 5 tests are the most valuable part of this PR (anchored-nested and FAST-mode cases included); we'd like to carry them essentially verbatim. The implementation, through no fault of yours, has been overtaken by main: cbm_gitignore_match_result now exposes negation directly (returns -1 when the last matching rule is a negation), which reduces the ~110-line inverted-matcher to roughly five lines inside the built-in-skip branch of should_skip_directory. Concretely, what we'd love to see in a refreshed version:
1. Rebase onto current main and implement via cbm_gitignore_match_result — last-match-wins survives, and the walk-chain conflicts disappear.
2. Use cbm_fopen (or reuse cbm_gitignore_load) instead of raw fopen — raw fopen breaks non-ASCII repo paths on Windows and is against house policy.
3. Policy decision (decided): negation may un-skip ordinary built-in dirs (obj/, dist/, target/ etc.) but NOT a small non-negatable core — .git, node_modules, and worktree-internal dirs stay skipped regardless, since .cbmignore is repo-committed and those skips are also OOM/safety protections (see #489). Please keep a test asserting !.git/ stays skipped.
We're also about to add a .cbmignore how-to doc spelling out the full syntax including these negation semantics, and a follow-up issue to unify the exclusion predicate across the auxiliary walkers so your negation propagates beyond discovery. Thanks again — the tests and the direction are exactly right.


---

DeusData:
Thank you for this — the problem was real, the negation semantics you proposed were right, and your five test cases were the most valuable part of the PR: they were carried over essentially verbatim. Since the branch predated the 3-state matcher on main (which shrinks the whole implementation to a few lines on cbm_gitignore_match_result), we carried the work over the line in a distilled follow-up with you credited as co-author on the commit — merged as 416e4f23 (PR #807). It also implements the decided safety policy: negation can un-skip ordinary built-in dirs (obj/, dist/, target/…) but a small non-negatable core (.git, node_modules, worktree dirs) stays protected, with your tests guarding both directions. The full syntax incl. negation is now documented in docs/cbmignore.md. Closing in favor of the distill — thanks again, and #500 is fixed thanks to your push on it!

PR #544, opened 2026-06-21
feat(mcp): expose index freshness and relationship provenance

vvenegasv:
## What does this PR do?

This PR adds machine-readable freshness and provenance evidence to MCP responses, helping clients determine whether results come from a current index, a stale or partial index, direct source resolution, or heuristic inference.

Changes include:

* Persisting the indexed Git HEAD and file coverage metrics for each project.
* Extending `index_status` with additive evidence about indexed/current revisions, working tree state, freshness, coverage, and known limitations.
* Returning `edge_evidence` from `trace_path` and `trace_call_path` when relationship provenance is available.
* Preserving backward compatibility by returning unknown or unavailable values when older indexes do not contain the new metadata.

## Why is this needed?

Consumers of the MCP need factual and inspectable signals to decide whether a result can be trusted as-is, whether the repository should be re-indexed, or whether the result should be verified directly against source code.

Rather than exposing an opaque reliability score, this change provides the underlying evidence needed for clients to make that decision.

## Checklist

* [x] Every commit is signed off (`git commit -s`)
* [ ] Tests pass locally (`make -f Makefile.cbm test`)
* [ ] Lint passes (`make -f Makefile.cbm lint-ci`)
* [x] New behavior is covered by a test

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,17 @@ codebase-memory-mcp cli --raw search_graph '{"label": "Function"}' | jq '.result
 | `delete_project` | Remove a project and all its graph data. |
 | `index_status` | Check indexing status of a project. |
 
+`index_status` includes an additive `evidence` object. `evidence.index_snapshot`
+reports the timestamp and Git HEAD captured when the last successful graph
+snapshot was finalized, the current HEAD, tracked working-tree state, and a
+`freshness` value: `current` means the indexed HEAD equals current HEAD and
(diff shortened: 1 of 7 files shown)


DeusData:
Thanks @vvenegasv — closing this in favor of **#559**, which is the same index-freshness/provenance feature but with the new C code actually covered by tests (this version ships none). Nothing is lost — #559 is a superset of this. I've left a note over there about one small cleanup to split out. 🙏

PR #545, opened 2026-06-21
fix: persist call-site line number on CALLS edges (fixes #503)

### LINKED ISSUE #503: CALLS edges omit the call-site line number (CBMCall.start_line captured but never persisted)
### Version

codebase-memory-mcp dev

### Platform

Linux (x64)

### Install channel

Built from source

### Binary variant

standard

**What happens**

Call edges (`CALLS`, and the siblings emitted through the same path) are stored without the source line of the call site. The line is computed during extraction but never written to the edge, so no consumer (`query_graph`, `trace_path`, ...) can recover *where* a call happens, only which symbol/module it is in.

**What I expected**

Each call edge to carry the call-site line (it is already available at emit time), so "where exactly is X called" is answerable from the graph instead of falling back to grep.

**Reproduction (any repo, no private code needed)**

1. Index any repository, e.g. a small public one like `pallets/flask`: `index_repository {"repo_path": "/path/to/flask"}`
2. Query inbound call edges: `query_graph {"project":"<proj>","query":"MATCH ()-[r:CALLS]->() RETURN r.callee, r.line LIMIT 5"}`
3. `r.line` is `null` for every row. `get_graph_schema` confirms the `CALLS` edge properties are `source_id, target_id, args, callee, candidates, confidence, strategy, url_path, via`: there is no line/position field.

**Where it happens (main @ e599df1)**

- `CBMCall.start_line` ("1-based source line of the call") is captured in `internal/cbm/cbm.h`.
- `finalize_and_emit()` in `src/pipeline/pass_parallel.c` builds the edge props (appends `args`, closes the JSON) and emits the edge, but never writes
  `call->start_line`. The value is in scope at that point; it is just not persisted.

**Impact**

`query_graph` / `trace_path` cannot report a call-site line, so caller/impact analysis stops at module/function granularity and agents fall back to grep when asked for exact call sites.

**Proposed fix**

Persist the line as an edge property in `finalize_and_emit()` (right after args, before the closing brace), only when `start_line > 0`. About 8 lines, one function, no schema-table change (edge props are JSON, so `r.line` becomes queryable after a re-index). Surfacing it in the `trace_path` callers list is a larger follow-up (it touches the traversal structs in `src/store/`) and I would
keep that as a separate PR. I have this patch implemented and locally verified (every `CALLS` edge gets an accurate `line` after re-index) and can open a PR if you are open to it.

**Open question**: property name `line` vs `call_line`, and store it in the JSON props (zero schema change) vs a dedicated edge column?

### Reproduction

Using a self-contained dummy snippet (no network, no private code, fully runnable):

1. Code being indexed — minimal dummy, save as `/tmp/cbmrepro/m.py`:

   ```python
   def target():
       return 1

   def caller():
       return target()   # call is on line 5
Commands:

codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/cbmrepro"}'
codebase-memory-mcp cli query_graph '{"project":"tmp-cbmrepro","query":"MATCH ()-[r:CALLS]->(t) WHERE t.name="target" RETURN r.callee, r.line"}'

(Use whatever project slug index_repository reports if it differs.)

Output vs expected:

Actual: one row, r.callee = "target", r.line = null. The edge has no line property at all (get_graph_schema shows CALLS props as source_id, target_id, args, callee, candidates, confidence, strategy, url_path, via).
Expected: r.line = 5, the source line where target() is called.

### Logs

```text
N/A
```

```text
 5 / 5 / 1
```

### Confirmations

noahkhomer18:
Fixes #503

### DIFF
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -1182,6 +1182,11 @@ static void finalize_and_emit(cbm_gbuf_t *gbuf, int64_t src_id, int64_t tgt_id,
     if (n > 0 && (size_t)n < CBM_SZ_2K - PP_ESC_SPACE) {
         size_t pos = append_args_json(props, CBM_SZ_2K, (size_t)n, call);
         if (pos < CBM_SZ_2K - SKIP_ONE) {
+            // Persist call-site line number on CALLS edges (fixes issue #503)
+            if (call->start_line > 0 && pos < CBM_SZ_2K - 32) {
+                int wr = snprintf(props + pos, CBM_SZ_2K - pos, ",\"line\":%d", call->start_line);
+                if (wr > 0) pos += wr;
+            }
             props[pos] = '}';
             props[pos + SKIP_ONE] = '\0';
         }


DeusData:
Thanks @noahkhomer18 — this is the right idea for #503, but #508 (already merged ✅) covers it more completely: it persists the call-site line on **both** the parallel and serial/LSP resolution paths, names the constant, and ships an integration test.

Since #503 is now fixed on `main`, I'm closing this one. Appreciate the contribution! 🙏

PR #567, opened 2026-06-23
Local integrated

win4r:
## Checklist

- [ ] Every commit is signed off (`git commit -s`) — required, CI rejects
      unsigned commits ([DCO](../DCO), see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [ ] Tests pass locally (`make -f Makefile.cbm test`)
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`)
- [ ] New behavior is covered by a test (reproduce-first for bug fixes)

PR #568, opened 2026-06-23
M2 idiomatic kinds

win4r:
## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,13 @@
-# codebase-memory-mcp
+# codebase-memory-mcp-pro
+
+> **🔱 Fork notice** — `codebase-memory-mcp-pro` is a community fork of [**DeusData/codebase-memory-mcp**](https://github.com/DeusData/codebase-memory-mcp) (MIT License, © 2025 DeusData), maintained by [@win4r](https://github.com/win4r). It tracks upstream and integrates the following fixes ahead of their upstream merge:
+>
+> - **Incremental-reindex correctness** ([#528](https://github.com/DeusData/codebase-memory-mcp/pull/528)) — preserve inbound cross-file `CALLS` edges on incremental re-index; editing a file no longer orphans calls into its symbols.
+> - **Cypher / `query_graph`** — populate node properties carried through `WITH` aggregation ([#465](https://github.com/DeusData/codebase-memory-mcp/pull/465)); fix label-filtered traversal silently truncating at 10 rows ([#412](https://github.com/DeusData/codebase-memory-mcp/pull/412)).
+> - **MCP tools** — `detect_changes` honors `since` ([#464](https://github.com/DeusData/codebase-memory-mcp/pull/464)); definition-preferred name resolution with ambiguity reporting ([#466](https://github.com/DeusData/codebase-memory-mcp/pull/466)); valid UTF-8 in `get_code_snippet` ([#526](https://github.com/DeusData/codebase-memory-mcp/pull/526)).
+> - **Robustness / build** — stack-buffer-overflow fix in `append_args_json` ([#475](https://github.com/DeusData/codebase-memory-mcp/pull/475)); JSON control-character escaping ([#527](https://github.com/DeusData/codebase-memory-mcp/pull/527)); preserve ADRs across a full re-index ([#539](https://github.com/DeusData/codebase-memory-mcp/pull/539)); libgit2 ≥ 1.8 build fix ([#512](https://github.com/DeusData/codebase-memory-mcp/pull/512)).
+>
+> All credit for the original engine belongs to DeusData. License unchanged — see [LICENSE](LICENSE). The upstream README follows verbatim.
 
 [![GitHub Release](https://img.shields.io/github/v/release/DeusData/codebase-memory-mcp?style=flat&color=blue)](https://github.com/DeusData/codebase-memory-mcp/releases/latest)
 [![License](https://img.shields.io/badge/license-MIT-green)](LICENSE)
--- a/bench/BASELINE.md
+++ b/bench/BASELINE.md
@@ -1,0 +1,29 @@
+# Head-to-head baseline — cbm-pro (24e6784c) vs codegraph (0.9.9)
+
+Repo: LingoLearn-iOS-main (29 Swift files). Harness: `bench/headtohead.sh`. Date: 2026-06-21.
+Per "confirm the failure before fixing it" — this is the *before* state. Re-run after each WS to prove movement.
+
+## Structural
+| metric | cbm-pro | codegraph | M1 target |
+|---|---|---|---|
+| nodes | 663 | 338 | — |
+| edges | 1876 | 792 | — |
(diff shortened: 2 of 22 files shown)


DeusData:
Thanks @win4r — this is the cumulative branch (it includes #567 + #569 plus the Swift idiomatic type kinds Struct/Enum/Actor, which are a nice addition).

Same blockers as the other two: the README rebrand to a personal fork, the `bench/` harness with a hardcoded personal path, the empty description, and DCO not signed.

Rather than three stacked, fork-branded PRs, could you send the **useful** pieces as focused, signed-off PRs — the `explore` tool, the Swift fixes/kinds — dropping the README rebrand and `bench/`? We'd be glad to land those — the engineering is genuinely good; it just needs to arrive clean (no rebrand/bench) and with CI green (signed-off commits, lint and tests passing). 🙏

DeusData:
Reviewed as the top of the stack — these three are strict containment (#567 ⊂ #569 ⊂ #568), so this PR carries everything in both others. I have closed #567 (fully superseded) and left detailed notes on #569; please read those first. This covers what is unique to #568.

**Your extension-clobber catch is sharp, and it may well be a live bug on main.**

You spotted that a same-file `extension X` re-labels `struct X` back to `Class`, because the extension pushes a type def with the same qualified name and the `UNIQUE(project, qualified_name)` upsert makes it last-write-wins. Looking at current `main`: it still pushes a type def for `extension`, and since `declaration_kind` is not `struct` the label stays `Class` — so the same clobber mechanism appears to be present. That is a real graph-quality bug and worth its own PR with a repro.

Similarly, main only special-cases Swift `struct`. **`enum` and `actor` still lump to `Class`**, so your `swift_idiomatic_type_kinds` test would be red on main for those two. Genuine repro material.

**But the mechanism needs rebuilding before it can land.** This PR extends roughly ten hardcoded label allowlists — registry gates in `pass_definitions`, `pass_parallel`, `pipeline_incremental`, both `resolve_as_class` sites, four store SQL lists, `pass_configlink`, `pass_enrichment`, BM25 ranking. Since you branched, main centralised exactly this into **`cbm_label_is_type_like()`** (`helpers.c:160`), whose comment says it plainly: *"Single source of truth … adding a label here updates them all."* That helper exists precisely to kill the drift that scattered allowlists cause, so re-adding ten of them would move us backwards. A rebase here is a reimplementation, not a conflict resolution.

**And one piece is a taxonomy decision rather than a fix.** `Actor` would be a **new first-class label** in the graph schema — it is not in `cbm_label_is_type_like()` today. Adding a label has blast radius across every consumer, so that is the maintainer's call and I have escalated it alongside the Swift `Enum` question. Your instinct matches our stated doctrine (production emits the precise label, consumers get updated rather than the test relaxed) — it is the execution that has aged, not the direction.

**Something your work surfaced about main itself, which I am filing regardless of these PRs:** `store.c` still uses `label IN ('Function','Method','Class')` at four sites (arch boundaries, clusters, vector search). Main *already* emits `Struct` nodes for Rust, Go, Swift and D — and those nodes are being excluded from all four today. That is a bug on main that your allowlist audit exposed, and it would have gone unnoticed otherwise. Thank you for that.

**Practical asks for any resubmission**, all of which currently block CI regardless of merit: `4cd82fa` has no `Signed-off-by`, so DCO fails; the AI session `Co-Authored-By` trailers should be stripped; and one concern per PR.

One thing that needs independent verification if this concept is revived: the commit itself notes total edges dropping 1813 → 1689 and claims all of them were spurious. That is plausible given the clobber fix, but an edge-count drop is exactly the kind of claim we want measured against a current main rather than inherited.

**Leaving this open** pending the maintainer's taxonomy answer, rather than closing your work while a decision is outstanding.

DeusData:
Closing this, and I want to be clear that **part of your work has landed on `main`** rather than letting a close read as a rejection.

## What shipped, with your credit

**#1380 (`63eda54`, merged as `880b87a6a`)** carries `Co-Authored-By: win4r`. Your allowlist audit found that four SQL queries in `store.c` and the BM25 ranking in `mcp.c` hardcoded their own label lists, so `get_architecture` (boundaries, packages, clusters) and vector search were dropping every `Struct`, `Interface`, `Enum`, `Type` and `Trait` node in the project — which, for Rust, Go, Swift and D, means most of the codebase's types were missing from the architecture view. That was live on `main` and nobody else had spotted it.

Two changes from your version, both deliberate. It routes through the canonical `cbm_label_is_type_like()` set rather than widening the literals in place, and it is pinned by a test in both directions, so the next type-like label fails CI instead of quietly shrinking query results. And it does not include `Actor`, since `main` does not emit that label — including it would have been shipping new vocabulary through a bug fix.

Your audit also surfaced a second problem indirectly: while revert-checking that fix, the check *passed* when it should have failed, because the build did not rebuild on a header change. That is fixed in the same PR.

## Why the PRs themselves cannot land

Neither is mergeable in its current shape: both are `CONFLICTING` against a `main` that has moved a long way, `dco` and `lint` are red, and **#568 contains #569 with identical commit OIDs** — so merging #568 would silently also take the README rebrand to `codebase-memory-mcp-pro` and the `bench/` harness, which hardcodes a local path and shells out to an unpinned external binary. Those two cannot come upstream in any form.

The nine cherry-picked commits are all already on `main` under different SHAs, so they are now pure conflict noise.

## What is recorded, and still owed to you

Two further bugs you found are real, verified, and on the maintainer's list — they are **not** being dropped by this close:

1. **Swift enum and protocol members are double-emitted.** `push_class_body_children` scans body containers by child *type* and omits `enum_class_body` and `protocol_body`, while `find_class_body` locates them via the `body` *field* — which a child-type scan structurally cannot see. Every Swift enum static and protocol member is emitted twice, once as a `Method` and once as a spurious top-level `Function`. Your measurement of 38 duplicate nodes on a 29-file repo checks out, and your fix auto-merges cleanly onto current `main`.
2. **Swift `extension` clobbers the type it extends.** Because Swift qualified names bake in the file stem, `struct P {}` and `extension P {}` in `P.swift` both compute the same QN, and last-write-wins relabels `P` back to `Class` *and* relocates its `file_path` and line range onto the extension — so `get_code_snippet` points at the wrong code. Cross-file, `extension Date {}` mints a phantom `Class` node for a stdlib type.

The first is held only because it touches the same function as an open architecture decision on another PR — we do not want to answer the same prevention-versus-suppression question twice, in the same function, with different reasoning. When that lands it will carry your credit too.

## Left open for the maintainer

The `Enum` and `Actor` labelling and the `explore` tool are direction calls, not defects. `Enum` is low-risk since it already exists in the canonical set; `Actor` would be a genuinely new permanent label that every type-resolution consumer would silently drop unless added there as well.

## Thank you

Four real defects on `main`, one of which surfaced only because you audited allowlists nobody else thought to check, and another you reasoned out correctly from the UNIQUE upsert semantics. Your tests pin behaviour rather than asserting it. That is a better hit rate than most review rounds produce, and it is why one of these is already shipped and two more are queued rather than filed away.

PR #569, opened 2026-06-23
M1 surpass

win4r:
## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,13 @@
-# codebase-memory-mcp
+# codebase-memory-mcp-pro
+
+> **🔱 Fork notice** — `codebase-memory-mcp-pro` is a community fork of [**DeusData/codebase-memory-mcp**](https://github.com/DeusData/codebase-memory-mcp) (MIT License, © 2025 DeusData), maintained by [@win4r](https://github.com/win4r). It tracks upstream and integrates the following fixes ahead of their upstream merge:
+>
+> - **Incremental-reindex correctness** ([#528](https://github.com/DeusData/codebase-memory-mcp/pull/528)) — preserve inbound cross-file `CALLS` edges on incremental re-index; editing a file no longer orphans calls into its symbols.
+> - **Cypher / `query_graph`** — populate node properties carried through `WITH` aggregation ([#465](https://github.com/DeusData/codebase-memory-mcp/pull/465)); fix label-filtered traversal silently truncating at 10 rows ([#412](https://github.com/DeusData/codebase-memory-mcp/pull/412)).
+> - **MCP tools** — `detect_changes` honors `since` ([#464](https://github.com/DeusData/codebase-memory-mcp/pull/464)); definition-preferred name resolution with ambiguity reporting ([#466](https://github.com/DeusData/codebase-memory-mcp/pull/466)); valid UTF-8 in `get_code_snippet` ([#526](https://github.com/DeusData/codebase-memory-mcp/pull/526)).
+> - **Robustness / build** — stack-buffer-overflow fix in `append_args_json` ([#475](https://github.com/DeusData/codebase-memory-mcp/pull/475)); JSON control-character escaping ([#527](https://github.com/DeusData/codebase-memory-mcp/pull/527)); preserve ADRs across a full re-index ([#539](https://github.com/DeusData/codebase-memory-mcp/pull/539)); libgit2 ≥ 1.8 build fix ([#512](https://github.com/DeusData/codebase-memory-mcp/pull/512)).
+>
+> All credit for the original engine belongs to DeusData. License unchanged — see [LICENSE](LICENSE). The upstream README follows verbatim.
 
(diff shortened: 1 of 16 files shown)


DeusData:
Thanks @win4r — building on my note on #567: the genuinely new and useful parts here are the read-only `explore` MCP tool and the Swift duplicate-node fix — both look good and safe.

The blockers are the same as #567: the README rebrand to a personal fork, the `bench/headtohead.sh` harness (it hardcodes `/Users/charlesqin/...` and an external `codegraph` dependency, so it can't live upstream), the empty description, and unsigned commits.

Could you resubmit just the **useful** code — e.g. the `explore` tool as its own focused PR and the Swift fix as another — without the README rebrand or `bench/`, with signed-off commits (`git commit -s`) and a short description? Happy to review and merge those — the code is genuinely good; it just needs to come in clean (no rebrand/bench) and with CI green (signed-off commits, lint and tests passing). 🙏

DeusData:
Reviewed the whole set together, because these three are a **strict stack** rather than independent PRs: #567 ⊂ #569 ⊂ #568, confirmed by comparing commit OIDs. So this PR carries all of #567, and #568 carries all of this. Merging #568 would merge everything.

I have closed #567 — every cherry-pick in it landed upstream within days, and the last one patches code that has since been removed entirely. Details are there.

**The standout finding, and I would like it as its own PR: your Swift enum/protocol duplicate-node fix (`2e94cb4`).**

`push_class_body_children`'s body-container list is missing Swift's `enum_class_body` and `protocol_body`, so enum statics and protocol members get re-walked and double-emitted as top-level Functions — 38 duplicate nodes on a 29-file Swift repo, by your own measurement. I checked current `main`: `extract_defs.c:6343` **still lacks both**, the vendored Swift grammar still emits `enum_class_body`, and there is no Swift enum-dup test anywhere in the suite. So this looks genuinely still-open, and it is squarely our top priority — duplicate nodes are a graph-quality defect, not a cosmetic one.

Please send that as a small, atomic PR against current main: just the fix and its regression test, re-verified (five weeks of extraction rework and a grammar refresh have gone by, so a fresh repro is worth doing), with a DCO sign-off. Your regression test is already binding — it asserts the absence of the duplicate def, which is the right shape.

**The `explore` tool is a different kind of question, and it is the maintainer's rather than mine.**

It is ~300 lines adding a **permanent new MCP tool** — a one-way door — and the commit is explicit that it is built to match a competitor ("Matches codegraph's explore and exceeds it"). That is exactly the case where we ask whether a feature fits *this* project's worldview rather than another tool's. It also overlaps `get_code_snippet(include_neighbors)`, `trace_path` and `search_graph`, and it predates our move to the tree/cursor output format, so its markdown rendering is designed against an output philosophy we have since replaced.

I have put that question in front of the maintainer. **Please do not sink more work into the diff before there is an answer** — I would rather you opened a short proposal issue describing the caller-attribution and hotspot ideas, because those are worth discussing on their own merits.

Two smaller things about the bench harness, neither of which needs fixing unless the direction call goes your way: `bench/headtohead.sh:13` hardcodes a personal path (`/Users/charlesqin/.local/bin/codebase-memory-mcp`), so it silently defaults to a non-existent binary anywhere else; and it shells out to an external, unpinned `codegraph` binary plus `python3`. It is not wired into build or CI, so there is no supply-chain exposure — but we would be shipping a script whose behaviour depends on whatever `codegraph` happens to be installed.

One code note for whenever this resumes: `handle_explore` uses raw `strndup`, and we have `cbm_strndup` in `src/foundation/compat.h` precisely because the raw one is not portable. And the explore tests currently assert schema validity plus two error paths but no happy-path output, so a broken renderer would still pass.

**Leaving this open** pending the maintainer's answer on the tool, rather than closing it out from under you.

DeusData:
Closing this alongside #568, where the full reasoning is written out — but the short version belongs here too, because a bare "closed" would misrepresent what happened.

**Part of your work has landed on `main` with your credit.** #1380 (`63eda54`, merged as `880b87a6a`) carries `Co-Authored-By: win4r` and fixes the label-allowlist bug your audit found: four SQL queries in `store.c` plus the BM25 ranking in `mcp.c` hardcoded their own label lists, so `get_architecture` and vector search were silently dropping every `Struct`, `Interface`, `Enum`, `Type` and `Trait` node. For Rust, Go, Swift and D that meant most of the codebase's types were missing from the architecture view. It was live on `main`, and nobody else had found it.

This PR specifically cannot merge because **#568 contains it with identical commit OIDs** — deciding one decides both — and because it carries the README rebrand to `codebase-memory-mcp-pro` and the `bench/` comparison harness, which hardcodes a local path and shells out to an unpinned external binary. Neither of those can come upstream. The nine cherry-picked commits are all on `main` already under different SHAs, so they are now only conflict noise.

Two further bugs you found — the Swift enum/protocol double-emission and the `extension` clobbering the type it extends — are verified, recorded, and will land with your credit. Details are in the #568 close.

Thank you. Four real defects on `main` from one review round is a better hit rate than most, and it is why one is already shipped rather than filed away.

Issue #576, opened 2026-06-23
Ingest dbt manifest.json into Model/Source nodes + DEPENDS_ON lineage


alexisperinger-ux:
### What problem does this solve?

dbt computes a resolved dependency graph into `target/manifest.json`, but it is never ingested, so a project's model and source lineage is invisible to the index even though dbt already calculated it. Model-to-source lineage in particular cannot be reconstructed from source files alone.

Public test bed: `dbt-labs/jaffle_shop` (run `dbt compile` to produce `target/manifest.json`).

### Proposed solution

Add an explicit, opt-in MCP tool `ingest_dbt_manifest { project, manifest_path }` that runs after `index_repository` against an existing project store:

- Parse `manifest.json` with the vendored `yyjson`.
- Emit `model` / `seed` / `snapshot` as `Model` / `Seed` / `Snapshot` nodes and `source` as `Source` nodes, keyed by dbt `unique_id`.
- Emit each node's `depends_on.nodes` as `DEPENDS_ON` edges (model-to-model and model-to-source).

Zero schema change: reuses `cbm_store_upsert_node` / `cbm_store_insert_edge` and the existing `DEPENDS_ON` edge. Only lineage-bearing resources become nodes; `test` / analysis / `operation` are skipped. This is the authoritative dbt path, complementing source-level extraction (#575).

### Alternatives considered

- Auto-running ingestion inside `index_repository` when `target/manifest.json` is present: deferred so this adds only an explicit, opt-in tool and leaves default indexing untouched.
- Ingesting `test` / analysis nodes by default: skipped to keep the graph lineage-only.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
Thanks — ingesting the resolved `manifest.json` as the authoritative dbt lineage path (opt-in `ingest_dbt_manifest` tool; Model / Source / Seed / Snapshot nodes + `DEPENDS_ON`; skipping tests / operations) is the right complement to the source-level extraction in #575.

Implemented in your PR #583 — continuing there. Labeling `enhancement` / `parsing/quality`.

DeusData:
Closing this as not planned **in the form proposed** — ingesting `target/manifest.json` through a dedicated MCP tool — while the underlying goal has been met from the source side.

[...]

Issue #575, opened 2026-06-23
Extract dbt {% macro %} definitions from raw .sql models


alexisperinger-ux:
### What problem does this solve?

dbt `.sql` models are Jinja-templated (`{{ ref() }}`, `{{ source() }}`, `{% macro %}`), which the SQL grammar cannot parse, and dbt's manifest only exists after `dbt compile`. So in a raw checkout, models and their dependencies are invisible: a model file is only a generic `Module`, and a referenced model like `stg_users` is not even a node, so `ref()` lineage cannot form.

Public test bed: `dbt-labs/jaffle_shop` (index the raw `models/` without compiling).

### Proposed solution

Run an additive `tree-sitter-jinja2` pass on dbt-templated `.sql` (files containing `{{` / `{%`):

- `{{ ref('m') }}` / `{{ source('s','t') }}` become `USAGE` lineage edges.
- A dbt model (a `.sql` file with no macro defs) becomes a `Model` node keyed by file stem, so cross-file `{{ ref('that_model') }}` resolves into model-to-model lineage.
- `{% macro name(...) %}` becomes a `Macro`.

Zero schema change (freeform labels + existing `USAGE` edge). `Model` is emitted only on the `.sql` path, so a plain `.jinja` / `.j2` template is not treated as a model. These source-level `Model` nodes coexist with the manifest path (#576) without conflict.

Caveat: the vendored tree-sitter-jinja2 grammar models only {{ }} expressions (so ref() / source() are parsed from the AST). It has no rule for {% %} statements, so {% macro %} names are recovered with a small text scan until a statement-aware grammar is vendored.

### Alternatives considered

- Reusing the embedded-language (`<script>`-in-HTML) re-parse path: rejected because dbt Jinja is interleaved throughout the file, not a delimited sub-region; a full-file second parse is the right shape.
- Extending the grammar to parse `{% %}` statements: larger and touches grammar vendoring; deferred.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
Thanks — the raw-checkout gap (Jinja-templated models being invisible before `dbt compile`) is real, and the additive `tree-sitter-jinja2` pass emitting `ref()` / `source()` as `USAGE` edges and file-stem `Model` nodes is a clean approach that coexists with the manifest path (#576).

Implemented in your PR #584 — continuing there. Labeling `enhancement` / `parsing/quality`.

DeusData:
[...]

PR #582, opened 2026-06-23
feat(sql): first-class Table/View nodes and FROM/JOIN lineage edges

### LINKED ISSUE #574: SQL DDL is under-modeled: CREATE TABLE/VIEW become generic Variable nodes with no lineage

SQL DDL is under-modeled. `CREATE TABLE` / `CREATE VIEW` are extracted but labeled generic `Variable`, `CREATE MATERIALIZED VIEW` is dropped, and there is no table-to-table lineage. Every `CREATE` collapses to an untyped `Variable` node with zero lineage edges, so the graph cannot answer basic questions:

- "What tables/views does this view read from?"
- "What breaks downstream if I change this table?"

Public test bed: `lerocha/chinook-database` (pure `CREATE TABLE` / `CREATE VIEW` DDL).

### Proposed solution

- Label `create_table` as `Table`, and `create_view` / `create_materialized_view` as `View`.
- Emit FROM/JOIN `relation` references as `USAGE` lineage edges, resolved through the existing definition registry, so a view links to the tables/views it selects from.
- Zero schema change: freeform node labels + the existing `USAGE` edge type. Only relations that resolve against the registry emit an edge, so DML-only files fabricate nothing.

Scope: DDL labeling + FROM/JOIN lineage only.

### Alternatives considered

- A new `REFERENCES` edge type instead of reusing `USAGE`: avoided to add no edge type; `USAGE` already means "X refers to Y".
- Modeling columns as `Field` nodes: deferred to keep this small.

### Confirmations

alexisperinger-ux:
## Summary

SQL DDL was routed through the config-variable path, so `CREATE TABLE` / `CREATE VIEW` became generic `Variable` nodes with no table-to-table lineage: questions like "what does this view read from?" or "what breaks if I drop this table?" were unanswerable from the graph. This makes DDL first-class:

- `create_table` becomes a `Table` node; `create_view` / `create_materialized_view` become `View` nodes.
- FROM/JOIN relations become `USAGE` lineage edges, resolved against the Table/View registry, so view-to-table lineage is queryable.
- Schema-qualified DDL is named by the table, not the schema (`app.users` resolves as `users`).

Only relations that resolve against the registry emit an edge, so unresolved names and DML-only files fabricate no nodes or edges.

**No schema change:** node labels are freeform strings and `USAGE` is an existing edge type.

**Index mode:** the new nodes need a full-mode index; fast and moderate prune the directories where DDL typically lives.

**Tests:** `sql_ddl_node_labels`, `sql_view_lineage_usages`, `sql_schema_qualified_name` in `tests/test_extraction.c`.

Fixes #574.

## Related PRs

- **[#582](https://github.com/DeusData/codebase-memory-mcp/pull/582) (this PR)**: SQL DDL, first-class `Table` / `View` nodes + FROM/JOIN lineage ([#574](https://github.com/DeusData/codebase-memory-mcp/issues/574)).
- **[#584](https://github.com/DeusData/codebase-memory-mcp/pull/584)**: dbt Jinja from raw `.sql`, `Model` / `Macro` nodes + `ref()` / `source()` lineage ([#575](https://github.com/DeusData/codebase-memory-mcp/issues/575)).
- **[#583](https://github.com/DeusData/codebase-memory-mcp/pull/583)**: dbt manifest ingestion, `Model` / `Source` nodes + `DEPENDS_ON` lineage ([#576](https://github.com/DeusData/codebase-memory-mcp/issues/576)).

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -3020,11 +3020,72 @@ static bool extract_config_class_def(CBMExtractCtx *ctx, TSNode node, const char
     return true;
 }
 
+// Collect FROM/JOIN table references (tree-sitter-sql `relation` nodes) anywhere
+// under `node` and emit them as usages scoped to enclosing_qn. pass_usages then
+// resolves each ref_name to the referenced Table/View def and creates a USAGE
+// lineage edge (e.g. a view -> the tables it selects from). Emitting them here
+// (rather than via the generic identifier walker) sets the correct enclosing
+// scope and bypasses the is_definition_name suppression that drops them.
+static void collect_sql_relation_usages(CBMExtractCtx *ctx, TSNode node,
+                                        const char *enclosing_qn) {
+    if (strcmp(ts_node_type(node), "relation") == 0) {
+        TSNode nm = resolve_sql_func_name(node); // object_reference -> identifier
+        if (!ts_node_is_null(nm)) {
+            char *tname = cbm_node_text(ctx->arena, nm, ctx->source);
+            if (tname && tname[0]) {
+                CBMUsage usage;
+                usage.ref_name = tname;
+                usage.enclosing_func_qn = enclosing_qn;
+                cbm_usages_push(&ctx->result->usages, ctx->arena, usage);
+            }
+        }
+    }
+    uint32_t n = ts_node_child_count(node);
+    for (uint32_t i = 0; i < n; i++) {
+        collect_sql_relation_usages(ctx, ts_node_child(node, i), enclosing_qn);
+    }
+}
+
+// Handle SQL DDL relation defs: CREATE TABLE / VIEW / MATERIALIZED VIEW become
+// first-class Table/View nodes rather than generic Variable nodes. The relation
+// name sits on an object_reference child (the same shape create_function uses),
+// so resolve_sql_func_name locates it. Also emits FROM/JOIN dependencies as
+// usages so lineage edges form. Returns true if handled.
+static bool extract_sql_ddl_class_def(CBMExtractCtx *ctx, TSNode node, const char *kind) {
+    if (ctx->language != CBM_LANG_SQL) {
+        return false;
+    }
+    const char *label;
+    if (strcmp(kind, "create_table") == 0) {
+        label = "Table";
+    } else if (strcmp(kind, "create_view") == 0 ||
+               strcmp(kind, "create_materialized_view") == 0) {
+        label = "View";
+    } else {
+        return false;
+    }
+    TSNode name_node = resolve_sql_func_name(node);
+    if (ts_node_is_null(name_node)) {
+        return false;
+    }
+    char *name = cbm_node_text(ctx->arena, name_node, ctx->source);
+    if (!name || !name[0]) {
+        return false;
+    }
+    push_simple_class_def(ctx, node, name, label);
+    const char *qn = cbm_fqn_compute(ctx->arena, ctx->project, ctx->rel_path, name);
+    collect_sql_relation_usages(ctx, node, qn);
+    return true;
+}
+
 static void extract_class_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec) {
     CBMArena *a = ctx->arena;
     const char *kind = ts_node_type(node);
 
     if (extract_config_class_def(ctx, node, kind)) {
+        return;
+    }
+    if (extract_sql_ddl_class_def(ctx, node, kind)) {
         return;
     }
 
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -256,12 +256,31 @@ static TSNode resolve_ocaml_func_name(TSNode node) {
     return null_node;
 }
 
-// SQL: resolve create_function name from object_reference→identifier or direct identifier.
+// Last identifier (DFS pre-order) under `node`. For a schema-qualified
+// object_reference (schema.table) this is the table name; the schema prefix is
+// ignored. Leaves *found false and returns `best` unchanged if none is present.
+static TSNode sql_last_identifier(TSNode node, TSNode best, bool *found) {
+    if (strcmp(ts_node_type(node), "identifier") == 0) {
+        best = node;
+        *found = true;
+    }
+    uint32_t cc = ts_node_child_count(node);
+    for (uint32_t i = 0; i < cc; i++) {
+        best = sql_last_identifier(ts_node_child(node, i), best, found);
+    }
+    return best;
+}
+
+// SQL: resolve create_function / create_table / create_view name. The name sits
+// on an object_reference; for a schema-qualified name (schema.table) take the
+// last identifier (the table), not the first (the schema).
 static TSNode resolve_sql_func_name(TSNode node) {
     TSNode obj_ref = cbm_find_child_by_kind(node, "object_reference");
     if (!ts_node_is_null(obj_ref)) {
-        TSNode id = cbm_find_child_by_kind(obj_ref, "identifier");
-        if (!ts_node_is_null(id)) {
+        bool found = false;
+        TSNode empty = {0};
+        TSNode id = sql_last_identifier(obj_ref, empty, &found);
+        if (found) {
             return id;
         }
     }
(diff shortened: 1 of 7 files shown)


DeusData:

DeusData:
Reviewed all three of your PRs together, since they interlock — the full picture is here and the sibling threads point back to it. Thank you for the "reviewed as a set" note and the cross-links; that made this much easier to do properly.

**You diagnosed a real defect precisely.** `CREATE TABLE` and `CREATE VIEW` genuinely land as mislabeled `Variable` nodes on current main (`lang_specs.c`, `sql_var_types`), and the schema-qualified naming bug in `resolve_sql_func_name` — taking the *first* identifier of `object_reference`, so `app.users` resolves to `app` — is real and still live. Your last-identifier fix plus the `sql_schema_qualified_name` regression test is exactly the shape we want, and it fixes naming for SQL *functions* too, not just DDL.

**The resolution-aware design is the right instinct for this codebase.** Only relations that resolve against the definition registry emit edges, so unresolved names and DML-only files fabricate nothing. That matches our graph-accuracy-first doctrine better than most extraction PRs we see — the easy version of this feature would have invented nodes for every table name it saw.

All six new tests across the set are binding: they fail on main for the right reasons.

**Why this is not a straight merge: it is surface, not quality.**

`Table` and `View` are **new node labels**, and every label has to be carried by consumers forever. Right now they would be second-class citizens:

- `cbm_label_is_type_like()` (`helpers.c:160`) covers Class/Struct/Interface/Enum/Type/Trait only, so Table/View are type-like nowhere.
- `store.c` hardcodes label lists that exclude them — arch boundaries `:5279`, arch clusters `:6689`, hotspots `:5181`, and **vector search `:7742`**.

So the nodes would be findable through `search_graph` but **silently invisible to `get_architecture` and semantic search**. The PR does not update those lists or mention the exclusion. (For what it is worth, you are not alone in hitting this — a separate contributor's audit surfaced that main's own `Struct` nodes are excluded by the same lists today. Your work helped confirm it is systemic.)

**One thing that would bite on a naive rebase.** You patch the registry allowlists in `pass_definitions.c` and `pass_parallel.c`, but main has since refactored both to call `cbm_label_is_type_like()` — **and added a third keep-in-sync site at `pipeline_incremental.c:526` that this PR never touches**. Rebasing mechanically would register Table/View during a full index but *not* on incremental reindex, so lineage would silently degrade after the first file change. That is the kind of divergence that is very hard to notice later.

**One risk worth assessing before this lands**, independent of the direction call: Table and View short names enter the **shared global registry**, so very common names — `users`, `orders`, `data` — become resolution targets for unresolved short-name usages from *every* language in the project. That could produce cross-language false `USAGE` edges. Worth a look at `pass_usages` resolution ordering.

**Where it stands.** Whether cbm should carry a data-shape vocabulary (Table/View here, plus Model/Source/Macro in your sibling PRs) in a *code* knowledge graph is a product-direction question, and it is the maintainer's rather than mine. I have put it to him explicitly, along with the note that your linked issues #574/#575/#576 were self-filed and never got a direction answer — that silence is on us, and it is why you built three PRs' worth of work without knowing whether the concept was wanted.

**Please hold off reworking until there is an answer.** The anchors have moved enough that this needs a rework rather than a rebase, and I would rather not have you spend that twice.

One note regardless of the outcome: your `resolve_sql_func_name` fix is correct, small, and label-neutral — it stands on its own as a bug fix even if the lineage vocabulary is declined, and I have said so upstream.

krahnikblis:
@DeusData i just came here to try to upvote/comment this! i have been avoiding installing the various codebase graph MCP tools because of the lack of SQL support, and installed this main repo because it _claimed_ it had it! but then last week my indexing of SQL projects resulted in big `Variable` graphs centered on `coalesce()` and agents saying "nah i need to read the raw files plskthx" had me frustrated, but...

1. i merged @alexisperinger-ux 's PRs into the current main project pulled as of today. had to do some manual edits to resolve conflicts, wasn't too difficult.
2. i built from source, this was a bit more tedious since all this `c` stuff is well orthogonal to my world, but this is where i just let AI steer me. ended up with a built `.exe` which needed copying into a couple locations manually - the `install.ps1` script is not friendly to local builds!
3. connected the built version to VS Code and chatted the agent up, also consulted the UI - graph is now more rational in terms of data lineage, but it seems the tool descriptions/parameters available to the agent are still oriented toward code, not data; it kept trying to look for functional edges/nodes, so it seems optimization would need a bit of tool description/prompt nudging - i aim to solve this with a `SKILL.md` for the near term
4. overall it works! it's not as detailed as [SQL-Lineage](https://github.com/reata/sqllineage) (i.e., column lineage, CTEs, etc.), and probably could benefit from parsing assistance from something like [SQLGlot](https://github.com/tobymao/sqlglot) for various dialects, **but this is totally awesome for a SQL exploration tool!** 

so... if these PRs could be worked in to the main project, i'm an enthusiastic user!

DeusData:
Thank you for this — and thank you for your patience with how long it sat.

Your diagnosis was right and your design was right: DDL collapsing into untyped `Variable` nodes was a real modelling defect, resolving `FROM`/`JOIN` through the existing definition registry was the correct minimal mechanism, taking the last identifier of a schema-qualified `object_reference` was the correct fix (it also repaired `CREATE FUNCTION schema.fn` naming, which nobody had noticed), and your three tests were binding rather than decorative. That is a higher standard of contribution than we usually get, and it is why the feature is now on `main`.

**This has shipped — distilled in #1749, commit 7ba84e37 (merge 4bbc978c), with `Co-authored-by: alexisperinger-ux` on the commit.**

Let me be straight about why it landed that way instead of as a merge of this branch, because you deserve a real explanation rather than a form letter.

When I reviewed this in July I asked you to hold off reworking until I had answered the product-direction question — whether data-shape vocabulary (`Table`/`View`) belongs in a code knowledge graph at all. That answer took far longer than it should have, and it was mine to give. In the meantime the pipeline moved underneath the branch: registry seeding is now performed in three places (the full pass, the parallel pass, and the incremental re-resolve), and a straightforward rebase would have registered relations in only two of them, so an incremental reindex would have silently diverged from a clean full index. Beyond that, the direction answer came with conditions attached — a cross-language collision guard, incremental invalidation, and search/architecture integration — that together are larger than the change you originally sent.

Asking you to absorb all of that after a wait you did not cause seemed like the wrong way to treat the contribution, so I implemented it here and credited you on the commit. Concretely, what got added around your work:

- **`{0}`-initialisation of the lineage usage struct.** Your `CBMUsage` left `kind`, the resolution flags, and the byte spans as stack garbage, which meant nondeterministic edge classification and a guaranteed finding in our MSan lane. That one was merge-blocking on its own.
- **A single registry-membership predicate** (`cbm_label_is_registry_symbol`) replacing the three hand-synchronised label lists, so the rebase trap above cannot happen again to anyone.
- **A central relation veto in `cbm_registry_resolve`.** Table names such as `users`, `orders`, and `config` collide with ordinary identifiers in every language, and the shared registry would happily bind a Python `users()` call straight to your SQL table. Rather than add a label check at each consumer, the default resolve now refuses relation-labelled results outright, and only the SQL lineage path opts in through `cbm_registry_resolve_lineage`. Every present and future consumer is safe by construction. There is a test that fails on precisely that false `CALLS` edge when the veto is removed.
- **Incremental correctness**: `Table`/`View` now participate in the per-file surface hash, so renaming a table invalidates dependent SQL files instead of leaving stale lineage behind the early cutoff.
- **Findability**: relations rank with the type tier in search and appear in the architecture boundary, package, and cluster queries.

The verification you would want: 1291 tests green across every affected suite under ASan/UBSan, the O(n²) complexity guard confirming registry work stays linear, and the CI-image linters clean.

Your sibling PRs #583 and #584 (dbt manifest ingestion and Jinja lineage) are still open and under review — they raise their own direction question about how far into the data-tooling domain this project should reach, and I will come back to you on those threads specifically rather than letting them drift the way this one did.

A few pieces were deliberately deferred rather than dropped: vector-search embeddings for relation nodes, write-lineage (`INSERT INTO` / `UPDATE` as `WRITES` edges), and the tool-description wording that @krahnikblis raised above. If any of those interest you, I would be glad to have them from you.

Closing this one as shipped — with thanks.

PR #583, opened 2026-06-23
feat(dbt): add ingest_dbt_manifest tool for Model/Source nodes and DEPENDS_ON lineage

### LINKED ISSUE #576: Ingest dbt manifest.json into Model/Source nodes + DEPENDS_ON lineage

alexisperinger-ux:
## Summary

dbt computes a resolved dependency graph into `target/manifest.json`, but it was never ingested, so a project's model and source lineage stayed invisible to the index even though dbt had already calculated it. This adds an opt-in MCP tool `ingest_dbt_manifest {project, manifest_path}`:

- `model` / `seed` / `snapshot` become `Model` / `Seed` / `Snapshot` nodes; `source` becomes a `Source` node, each keyed by dbt `unique_id`.
- Each node's `depends_on.nodes` becomes a `DEPENDS_ON` edge (model-to-model and model-to-source).
- `test` and analysis resources are skipped, so the graph keeps only lineage-bearing nodes.

It runs after `index_repository` against an existing store, so it augments rather than fabricates a graph, and is independent of index mode.

**No schema change:** reuses `cbm_store_upsert_node` / `cbm_store_insert_edge` and the existing `DEPENDS_ON` edge type.

**Tests:** `tool_ingest_dbt_manifest_issue576` in `tests/test_mcp.c`, asserting 5 models + 2 sources + 5 `DEPENDS_ON` edges from a fixture manifest, with the `test` node skipped.

Fixes #576.

## Related PRs

- **[#582](https://github.com/DeusData/codebase-memory-mcp/pull/582)**: SQL DDL, first-class `Table` / `View` nodes + FROM/JOIN lineage ([#574](https://github.com/DeusData/codebase-memory-mcp/issues/574)).
- **[#584](https://github.com/DeusData/codebase-memory-mcp/pull/584)**: dbt Jinja from raw `.sql`, `Model` / `Macro` nodes + `ref()` / `source()` lineage ([#575](https://github.com/DeusData/codebase-memory-mcp/issues/575)).
- **[#583](https://github.com/DeusData/codebase-memory-mcp/pull/583) (this PR)**: dbt manifest ingestion, `Model` / `Source` nodes + `DEPENDS_ON` lineage ([#576](https://github.com/DeusData/codebase-memory-mcp/issues/576)).

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -4554,8 +4554,9 @@ static int64_t dbt_upsert_node(cbm_store_t *store, const char *project, const ch
         mat = yyjson_get_str(yyjson_obj_get(cfg, "materialized"));
     }
     char props[CBM_SZ_512];
-    snprintf(props, sizeof(props), "{\"source\":\"dbt\",\"package\":\"%s\",\"materialized\":\"%s\"}",
-             pkg ? pkg : "", mat ? mat : "");
+    snprintf(props, sizeof(props),
+             "{\"source\":\"dbt\",\"package\":\"%s\",\"materialized\":\"%s\"}", pkg ? pkg : "",
+             mat ? mat : "");
     cbm_node_t n;
     memset(&n, 0, sizeof(n));
     n.project = project;
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -55,6 +55,7 @@ enum {
 #include "foundation/log.h"
 #include "foundation/str_util.h"
 #include "foundation/dump_verify.h"
+#include "foundation/hash_table.h"
 #include "foundation/compat_regex.h"
 #include "pipeline/artifact.h"
 
@@ -271,6 +272,12 @@ typedef struct {
 } tool_def_t;
 
 static const tool_def_t TOOLS[] = {
+    {"ingest_dbt_manifest",
+     "Ingest a dbt target/manifest.json into the indexed project: adds Model/Source nodes and "
+     "DEPENDS_ON lineage edges from each node's ref()/source() dependencies. Index the "
+     "repository first, then call with the project name.",
+     "{\"type\":\"object\",\"properties\":{\"project\":{\"type\":\"string\"},\"manifest_path\":{"
+     "\"type\":\"string\"}},\"required\":[\"project\",\"manifest_path\"]}"},
     {"index_repository",
      "Index a repository into the knowledge graph. "
      "Special mode 'cross-repo-intelligence': skip extraction, only match Routes/Channels "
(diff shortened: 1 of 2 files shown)


DeusData:

DeusData:
Reviewed as part of your set — the direction discussion is on #582, and this note covers what is specific to #583. Unlike the other two, this one is file-disjoint (`mcp.c` + its test), so it could in principle land independently.

**The structure is clean.** A two-pass build — nodes first, then edges resolved through a uid→id hash table — is the right shape, and keying nodes by dbt's `unique_id` is a sound choice for stable identity. The edge insert reuses the existing `ON CONFLICT DO UPDATE` upsert, so re-ingesting the same manifest is idempotent; I verified that rather than assuming it.

**Four things need fixing before this could merge, independent of the direction call.** None of them is malicious — they are the ordinary hazards of parsing an untrusted file, and this is a path that reads one.

1. **Raw `fopen()` on a user-supplied path.** `dbt_read_file` uses it directly. Our rule is `cbm_fopen()` (UTF-8 → `_wfopen`), because raw `fopen` breaks non-ASCII paths on Windows — a `manifest_path` containing any non-ASCII character would fail there. The rest of `mcp.c` is consistent about this (`:6452`, `:6934`, `:7163` and others), so this reads as an oversight rather than a choice.

2. **JSON injection into stored node properties.** `dbt_upsert_node` `snprintf`s `package_name` and `materialized` — both attacker-controlled strings from the manifest — unescaped into `properties_json`:
   ```
   {"source":"dbt","package":"%s",...}
   ```
   A manifest value containing a double quote produces malformed or attacker-shaped JSON in the `nodes.properties` column, which downstream `json_extract` consumers then read. There is an escaping helper in the tree for exactly this. The same code path also truncates silently at 512 bytes.

3. **No size or parse budget.** The whole file is read into memory and handed to yyjson. Real dbt manifests reach hundreds of megabytes, and a hostile one is a straightforward memory-exhaustion vector against the server. A size cap with a clear error is enough.

4. **Missing tool-table entries.** A new MCP tool needs its row in the 4-field tool table plus an annotations entry, or it will not be advertised or annotated consistently with the others.

One behavioural caveat worth documenting in the tool description if this lands: the ingested nodes live in the store, so a subsequent full reindex wipes them and the manifest has to be re-ingested. Anyone using this will hit that, and finding out by surprise is unpleasant.

**And the larger question, which is the maintainer's:** `ingest_dbt_manifest` would be a **permanent new MCP tool** — a one-way door we support indefinitely. That is bound up with the same direction call as your other two PRs: whether cbm carries a data-lineage vocabulary at all. It is also worth noting that #584's source-extraction path may cover much of the same ground without a new tool, so if the answer is yes, the sequencing matters.

Please hold the rework until there is an answer on direction — I would rather you fixed these four things once, against a decision, than twice.

DeusData:
Thank you for this, and for the care that went into it — the two-pass build (nodes first, then edges through a `unique_id` map) is the right shape for manifest ingestion, and the upsert path really is idempotent as you described.

I'm closing it, and the reason is architectural rather than anything about the code quality.

**The decisive problem: ingested data would not survive.** The indexing pipeline publishes a generation by building a staging database and renaming it over the live one. Anything written into the store from outside the indexer is therefore erased by the next index run — and since watched projects auto-reindex in the background, that would happen at an unpredictable moment, with no error. The tool would report `"models": 5`, and the nodes would be gone by the time someone queried for them.

The one kind of externally-authored data that does survive is ADRs, and only because there is deliberate capture-and-restore machinery for it (`capture_existing_adr`, the `saved_adr` field, and a regression test pinning the behaviour). Making manifest ingestion real would mean either building the equivalent for dbt nodes, or moving ingestion into the pipeline as a pass — and once you're inside the pipeline, the source files are right there, which is the route we took.

**So the concept did land, from the other direction.** dbt lineage now ships in #1756, with `Co-authored-by: alexisperinger-ux`, built on your sibling PR #584's source-parsing approach: `{{ ref() }}` and `{{ source() }}` are read out of the repo's own `.sql` models during normal indexing, producing `Model` nodes and lineage edges that are rebuilt on every index like everything else in the graph. Details are on #584's closing comment.

**On the security findings**, since you may hit the same pattern elsewhere: the handler passed `manifest_path` straight to `fopen()` with no workspace-boundary check, so any file the server could read was ingestable — that bypasses hardening we added in #1464. Alongside that: raw `fopen` instead of our `cbm_fopen` wrapper (non-ASCII paths break on Windows), manifest strings interpolated unescaped into the stored properties JSON (a value containing a quote breaks out of the field), and no size cap on the read. Each is fixable in isolation; together they're what tends to happen when a write path grows outside the one that already has the safeguards.

**What the manifest genuinely offers that source parsing cannot** — resolved database and schema names, materialization strategy, seeds and snapshots, and the authoritative DAG after macro expansion — is real, and I don't want to pretend otherwise. If we come back to it, it will be as a pipeline pass that reads the manifest when one is present, so the data is rebuilt rather than injected. That's a bigger design conversation than a PR review, and it isn't fair to keep this open waiting on it.

Thank you for three substantial, well-argued PRs. Two of the three concepts have now shipped, both with your name on the commits.

PR #584, opened 2026-06-23
feat(dbt): extract dbt Jinja lineage and macros from raw .sql models

### LINKED ISSUE #575: Extract dbt {% macro %} definitions from raw .sql models

alexisperinger-ux:
## Summary

dbt `.sql` models are Jinja-templated, so without a compiled manifest a model and its dependencies are invisible to the graph: a referenced model like `stg_users` is not even a node. This indexes raw (uncompiled) dbt models:

- `{{ ref('m') }}` / `{{ source('s','t') }}` become `USAGE` lineage edges, via the vendored `tree-sitter-jinja2` grammar.
- A dbt model (a `.sql` file with no macro defs) becomes a `Model` node keyed by file stem, so cross-file `{{ ref('that_model') }}` resolves into model-to-model lineage.
- `{% macro name(...) %}` becomes a `Macro` node.

`Model` is emitted only on the `.sql` path, so a plain `.jinja` / `.j2` template is not treated as a model, and a macro-defining file is treated as a library, not a model.

**No schema change:** freeform `Model` / `Macro` labels and the existing `USAGE` edge type.

**Index mode:** macro extraction is gated to full mode, so the new nodes need a full-mode index.

**Tests:** `dbt_jinja_macro_defs`, `dbt_jinja_ref_lineage`, `dbt_sql_ref_lineage` in `tests/test_extraction.c`.

Complements the authoritative manifest path ([#583](https://github.com/DeusData/codebase-memory-mcp/pull/583)).

Fixes #575.

## Related PRs

- **[#582](https://github.com/DeusData/codebase-memory-mcp/pull/582)**: SQL DDL, first-class `Table` / `View` nodes + FROM/JOIN lineage ([#574](https://github.com/DeusData/codebase-memory-mcp/issues/574)).
- **This PR**: dbt Jinja from raw `.sql`, `Model` / `Macro` nodes + `ref()` / `source()` lineage ([#575](https://github.com/DeusData/codebase-memory-mcp/issues/575)).
- **[#583](https://github.com/DeusData/codebase-memory-mcp/pull/583)**: dbt manifest ingestion, `Model` / `Source` nodes + `DEPENDS_ON` lineage ([#576](https://github.com/DeusData/codebase-memory-mcp/issues/576)).

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -588,6 +588,9 @@ CBMFileResult *cbm_extract_file(const char *source, int source_len, CBMLanguage 
     cbm_extract_imports(&ctx);
     cbm_extract_unified(&ctx);
 
+    // dbt Jinja: lineage (ref/source) + macro defs from raw templated .sql models.
+    cbm_extract_dbt_jinja(&ctx);
+
     // Channel detection (Socket.IO / EventEmitter) — JS/TS only.
     cbm_extract_channels(&ctx);
 
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -584,6 +584,7 @@ void cbm_channels_push(CBMChannelArray *arr, CBMArena *a, CBMChannel ch);
 // --- Sub-extractor entry points ---
 
 void cbm_extract_definitions(CBMExtractCtx *ctx);
+void cbm_extract_dbt_jinja(CBMExtractCtx *ctx);
 void cbm_extract_imports(CBMExtractCtx *ctx);
 void cbm_extract_usages(CBMExtractCtx *ctx);
 void cbm_extract_semantic(CBMExtractCtx *ctx);
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -5619,3 +5619,242 @@ void cbm_extract_definitions(CBMExtractCtx *ctx) {
     // Extract module-level variables
     extract_variables(ctx, ctx->root, spec);
 }
+
+// ============================ dbt Jinja extraction ============================
+// dbt models/macros are Jinja-templated .sql. The vendored tree-sitter-jinja2
+// grammar parses {{ ... }} expressions but treats {% ... %} statements as opaque
+// text, so the two dbt constructs are recovered differently:
+//   - {{ ref('m') }} / {{ source('s','t') }} are extracted via tree-sitter as
+//     lineage usages (ref_name = the referenced model/table); pass_usages then
+//     resolves each to a matching Model node and emits a lineage edge.
+//   - {% macro name(...) %} defs are recovered with a small text scan, since the
+//     grammar does not model the {% %} statement block.
+// This runs as an additive pass over raw (uncompiled) .sql, complementing the
+// authoritative compiled-manifest path (the ingest_dbt_manifest MCP tool).
+
+// True if the source contains any Jinja delimiter ({{ or {%).
+static bool source_has_jinja(const char *s, int len) {
+    for (int i = 0; i + 1 < len; i++) {
+        if (s[i] == '{' && (s[i + 1] == '{' || s[i + 1] == '%')) {
+            return true;
+        }
+    }
+    return false;
+}
+
+// Strip a single pair of surrounding ' or " quotes from a jinja lit_string text.
+static char *jinja_unquote(CBMArena *a, char *s) {
+    if (!s) {
+        return NULL;
+    }
+    size_t n = strlen(s);
+    if (n >= 2 && (s[0] == '\'' || s[0] == '"') && s[n - 1] == s[0]) {
+        char *inner = cbm_arena_strdup(a, s + 1);
+        size_t m = strlen(inner);
+        if (m > 0) {
+            inner[m - 1] = '\0';
+        }
+        return inner;
+    }
+    return s;
+}
+
+// Return the last (rightmost) lit_string under `node` in DFS order. For a
+// fn_call this is the final string argument: ref('pkg','model') -> 'model',
+// source('src','tbl') -> 'tbl'. The function name is an identifier, never a
+// lit_string, so it is not picked up.
+static TSNode jinja_last_lit_string(TSNode node, TSNode best, bool *found) {
+    if (strcmp(ts_node_type(node), "lit_string") == 0) {
+        best = node;
+        *found = true;
+    }
+    uint32_t cc = ts_node_child_count(node);
+    for (uint32_t i = 0; i < cc; i++) {
+        best = jinja_last_lit_string(ts_node_child(node, i), best, found);
+    }
+    return best;
+}
+
+// Walk a jinja2 AST; emit ref()/source() calls as usages scoped to enclosing_qn.
+static void collect_jinja_ref_usages(CBMExtractCtx *ctx, TSNode node, const char *enclosing_qn) {
+    if (strcmp(ts_node_type(node), "fn_call") == 0) {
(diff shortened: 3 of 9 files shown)


DeusData:

DeusData:
Reviewed as part of your set — the direction discussion lives on #582 and this covers what is specific to #584.

**First, a structural thing you will want to know: this PR contains all of #582.** Its first four commits are re-cherry-picks of the same four originals (same messages and `cherry picked from` trailers, different OIDs), and every hunk of #582's diff appears verbatim here. They are not a stacked branch — both target `main` directly, so they duplicate rather than build on each other. As it stands the two cannot both land: either this one goes in and #582 becomes redundant, or #582 lands and this gets rebased down to dbt-only. Worth deciding which you would prefer.

(Minor related signal: the tip commit is "resolve cherry-pick conflict markers in test_extraction.c". The final diff is clean — I checked — but conflict markers were pushed at some point.)

**Security is clean, and I want to credit one choice specifically:** you used the `tree-sitter-jinja2` grammar **already vendored** on main rather than adding a dependency. That is the right instinct, and it means no `MANIFEST.md` obligation is triggered. The text scanner and AST walks are bounds-checked.

**Three things give me pause, and they are about precision rather than safety.**

1. **The Model heuristic has no dbt gate.** Any templated `.sql` file with zero macros becomes a `Model` node keyed by file stem. There is no `dbt_project.yml` detection, so an Airflow or Flyway repo — or any repo with Jinja-templated SQL — would grow false `Model` nodes. For a project whose first priority is graph accuracy, inventing nodes for non-dbt repos is the failure mode we care most about avoiding.

2. **Macro recovery is a raw text scan.** It is comment-blind, so a commented-out `{% macro %}` still emits a definition. In a tree-sitter codebase that is a precision outlier. I understand why — the grammar does not model `{% %}` blocks — but it is worth naming rather than leaving implicit.

3. **The nested second parse establishes a pattern.** Re-parsing a `.sql` file with a second parser inside `cbm_extract_file` is a new architectural shape for this codebase: a host language plus an embedded language. It works, and it is bounded. But patterns like this get copied, so if we want embedded-language support it deserves to be designed once deliberately rather than arriving ad hoc for one language pair. That is part of what I have put to the maintainer.

**What is genuinely good here:** raw dbt models really are code sitting in indexed repositories, they index today as broken SQL with zero lineage, and your design reuses the existing usage-resolution pass rather than inventing parallel machinery. The three dbt tests are binding. This is not a drive-by.

It also inherits #582's issues — the stale registry anchors and the **third keep-in-sync site at `pipeline_incremental.c:526`** that neither PR touches, which on a naive rebase would mean labels registered during a full index but not on incremental reindex. With three labels now (Table, View, Model) the shared-registry short-name pollution surface is correspondingly larger.

**Please hold off reworking until the maintainer answers the direction question on #582.** This is the PR that most depends on that answer, and rebasing 462 lines against moved anchors is a poor use of your time before the concept is settled.

DeusData:
Thank you for this one, and for the whole three-PR series — this was the PR that convinced me the concept belonged in the project.

**The direction question is answered: yes.** dbt models are source files in the repository, `ref()` and `source()` are genuine source-level dependencies, and reading them is exactly what this project does. The reason a dbt project's dependency graph was invisible to us is simply that the SQL grammar cannot see through Jinja — which is the gap you identified and closed.

**It has shipped as #1756**, with `Co-authored-by: alexisperinger-ux` on the commit. I'm closing this PR as superseded rather than merging it, and I want to explain both halves of that honestly.

**Why it could not merge as-is.** This branch carries seven commits, four of which are PR #582's SQL DDL work re-cherry-picked rather than stacked. That work landed on `main` last night in #1749 (also distilled from you, also credited), so those four commits are now both obsolete and conflicting. Roughly a third of this diff is a feature that already exists upstream in a different form — merging it would have re-litigated a decision that was already made.

**What changed in the implementation, and why.** The lineage half is your design; the differences are integration decisions that only became available after #1749:

- **`Model` is a *relation* label**, joining `Table` and `View` from #1749, rather than a standalone label registered in two of the three registry seed sites. That is one line in `cbm_label_is_relation`, and it earns registry seeding, the cross-language name veto, incremental surface hashing, search ranking, and the architecture queries all at once. It also means a model's `source('raw','customers')` resolves onto a `Table` declared in a plain DDL migration in the same repo — dbt lineage and SQL lineage are one graph rather than two.
- **A different gate.** Yours keyed the Model node on "`.sql` and contains Jinja", which mints a node for every Airflow-templated query too. This one requires an actual `ref()`/`source()` call in the file: the dbt builtins are the evidence. That is cheaper than a `dbt_project.yml` lookup and more precise, since it also excludes non-dbt templated SQL that happens to sit inside a dbt repo.
- **Macros are deliberately left out.** I went into this intending to keep them, and then read the grammar: vendored tree-sitter-jinja2 has no node types for `{% ... %}` statements at all, only `{{ }}` expressions. Your text scan was therefore the only option available, and it inherits two limits — it cannot tell a commented-out macro from a live one, and it reports every macro as a single-line span because there is no `{% endmacro %}` to close against. Rather than ship that, macro support is filed as a follow-up where it can be done with a real scanner.
- **`{0}`-initialised usage structs.** Both this PR and #582 left `kind`, the resolution flags, and the byte spans as uninitialised stack memory, which means nondeterministic edge classification and a guaranteed MSan finding. Same fix as in #1749.
- The pass lives in its own `internal/cbm/extract_dbt.c` rather than adding 240 lines to the 5,800-line `extract_defs.c`.

Verification, for what it's worth: 1296 tests green across every affected suite under ASan/UBSan, the clang-analyzer memory gate clean, and two of the claims above pinned by tests that are *proven* to fail when the thing they test is removed.

Your sibling PR #583 is being closed too, for a different and more specific reason — the manifest tool writes to the store outside the indexing pipeline, and our publish-by-rename would erase it on the next reindex. That is explained on that thread.

If macros interest you, the follow-up is yours for the taking, and you'd start from a codebase that already has the lineage layer you designed. Thank you again — three well-argued PRs, and two of them have now shipped in some form.

Issue #581, opened 2026-06-23
Memory leak: process grows to 50+ GB virtual memory over hours/days, crashes Windows


baoyu0:
## Description

`codebase-memory-mcp` v0.8.1 has a severe memory leak when running as a persistent MCP server (stdio mode). Over the course of hours to ~2 days, the process grows to consume **50+ GB** of virtual memory, eventually exceeding the system commit limit and crashing the entire Windows machine.

## Environment

- **OS**: Windows 10, 64-bit (build 10.0.26100)
- **codebase-memory-mcp version**: v0.8.1 (installed via the official Windows .zip release, 269 MB binary)
- **RAM**: 32 GB physical
- **Page file**: system-managed
- **Agent**: Hermes Agent (launches codebase-memory-mcp as a subprocess via MCP stdio config)
- **Single project indexed**: `C:/Users/zhaid/.pi/agent` (~55k nodes, ~113k edges, ~97 MB on disk)
- **Config**: `auto_index=false`, `auto_index_limit=50000`

## Evidence

Windows Resource-Exhaustion-Detector (Event ID 2004) logged twice before the crash:

**00:23:54** — 8 minutes before crash:
| Process | Virtual Memory (Commit Charge) |
|---------|-------------------------------|
| `codebase-memory-mcp.exe` (PID 52508) | **46.4 GB** |
| `python.exe` (Hermes companion) | **59.4 GB** |

**00:28:53** — 3 minutes before crash — still growing:
| Process | Virtual Memory |
|---------|--------------|
| `codebase-memory-mcp.exe` (PID 52508) | **54.5 GB** (+8 GB in 5 min) |
| `python.exe` (Hermes companion) | **59.3 GB** |

Combined commit of ~114 GB exceeded the system commit limit (~100 GB = 32 GB RAM + page file), triggering:
- `Volsnap` event 6: "The shadow copy of volume C: could not create a new page stack. The system may be low on virtual memory."
- `OmenCommandCenterBackground.exe` crashed with `System.OutOfMemoryException`
- `Everything.exe` crashed with stack overflow (0xc00000fd)
- `dwm.exe` (Desktop Window Manager) crashed 3 times
- `AUDIODG.EXE` (audio) crashed
- Eventually Kernel-Power 41: system rebooted uncleanly

## Recurrence

This has happened at least 3 times since installation (June 12):
- **Jun 13**: same pattern — the process consumed ~107 GB
- **Jun 15**: ~99 GB
[...]

Issue #580, opened 2026-06-23
Feature request: --max-memory flag for RAM-constrained hosts


darkhipo:
On an 8GB M1 Air indexing the MultiAgency codebase (~6000 files), codebase-memory-mcp v0.8.1 consumes 8-11GB RSS. Two concurrent instances pushed the host to swap death with malloc corruption and SIGSEGV.

The knowledge graph is built fully in memory. For large codebases on constrained hardware, this causes:
- Swap thrash (load averages > 20)
- malloc(3) failures ("pointer being freed was not allocated")
- Process crashes from corrupted allocations

Request: a `--max-memory=N` flag (in MB or GB) that limits the in-memory graph size. When exceeded, spill to disk or use a streaming/paged approach. This would let the binary run on 8GB machines by trading speed for RAM safety.

Workaround: kill and restart periodically. Not sustainable.

RithvikReddy0-0:
Took a look at how memory is currently bounded, in case it helps scope this.

There's already an RSS budget mechanism (`cbm_mem_init` in `foundation/mem.c`, currently `MAIN_RAM_FRACTION` = 50% of total RAM) that drives extraction back-pressure in `pass_parallel.c` it sheds per-file retention when the process goes over budget. So a `--max-memory=N` flag could feed straight into that path with little plumbing.

The caveat is that the existing budget bounds the *extraction* peak, not the final knowledge graph, which is held fully in memory before the dump to SQLite. On a ~6000-file repo that final graph is most of your 8-11GB. So a flag wired only into `cbm_mem_init` would tighten back-pressure but wouldn't actually keep the process under N on a large repo  the graph would still blow past it. Delivering the real "run safely on 8GB" behavior needs the paged/spill approach you described, which is the larger change.

Separately, the `malloc` corruption  `"pointer being freed was not allocated"` plus SIGSEGV — is worth treating as its own issue. That's heap corruption, not a clean allocation failure; OOM alone usually gives a NULL malloc or the OOM-killer, not a freed-pointer abort. It may be a latent double-free/overrun that only trips under pressure (or under the two-concurrent-instances case, if any state is shared via the cache dir). Might be worth splitting that out so it doesn't get lost behind the feature work.

Happy to help with the flag-into-`cbm_mem_init` piece if you decide a partial cap (with honest docs about what it does and doesn't bound) is worth shipping ahead of the full paged-graph work.

DeusData:
[...]

Issue #563, opened 2026-06-23
Index failed in large repo


jwu-au:
### Version

codebase-memory-mcp 0.8.1

### Platform

macOS (Apple Silicon)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

I was trying to index a large repo, trying to limit the workers to 4, but still `abort` in the end. Is there any workaround on this? 

### Reproduction

CBM_WORKERS=4 codebase-memory-mcp cli index_repository \
  '{"repo_path": "/my_repo_path", "mode": "fast", "persistence": true}'

### Logs

```text
level=info msg=mem.init budget_mb=24576 total_ram_mb=49152
level=info msg=pipeline.discover files=123856 elapsed_ms=8603
level=info msg=pipeline.route path=full
level=info msg=pass.start pass=structure files=123856
level=info msg=pass.done pass=structure nodes=181977 edges=156452
level=info msg=pass.timing pass=structure elapsed_ms=424
level=info msg=pipeline.mode mode=parallel workers=4 files=123856
level=info msg=parallel.extract.start files=123856 workers=4
level=info msg=parallel.mem.budget total_mb=24576 per_worker_mb=6144
level=info msg=parallel.extract.file.start pos=3 size_kb=4356 path=

....
....

level=info msg=parallel.extract.progress done=123760 total=123856
level=info msg=parallel.extract.progress done=123770 total=123856
level=info msg=parallel.extract.progress done=123780 total=123856
level=info msg=pkgmap.scan_repo manifests=2500
level=info msg=pkgmap.build entries=2588
level=info msg=parallel.extract.mem rss_mb=5548 peak_mb=15352 budget_mb=24576 per_worker_mb=6144
level=info msg=parallel.extract.done nodes=3488953 errors=72
level=info msg=pass.timing pass=parallel_extract elapsed_ms=124294
level=info msg=mem.collect phase=post_extract rss_mb=5548
level=info msg=parallel.registry.start files=123856
level=info msg=parallel.registry.done entries=3358155 defines=3488953 imports=1512960
level=info msg=pass.timing pass=registry_build elapsed_ms=21999
level=info msg=mem.phase phase=registry_build rss_mb=7772 peak_mb=15352
level=info msg=pass.timing pass=lsp_cross_prepare elapsed_ms=1388
level=info msg=mem.phase phase=lsp_cross_prepare rss_mb=8832 peak_mb=15352
[...]

PR #586, opened 2026-06-23
feat(mem): CBM_MAX_MEMORY_MB explicit memory-budget override (#580)

yangsec888:
## Summary

Adds a `CBM_MAX_MEMORY_MB` env var that overrides the `ram_fraction × total_RAM` memory budget with an explicit cap in MiB. Closes the `--max-memory` request in #580 and gives #563 a workaround on hosts without cgroup limits (macOS / bare metal).

## Motivation

The knowledge graph is built fully in memory and the budget is sized to 50% of detected RAM. Two gaps remain even after cgroup-aware detection landed:

1. **No cgroup on the host** (macOS in #580/#563, Windows in #581) → budget is 50% of *physical* RAM, which can still swap-thrash an 8 GB machine.
2. **Wanting a budget *below* the cgroup limit.** In a container the parent process and any sibling/subprocesses share the cgroup; `0.5 × cgroup` can still be too high. There was no way to pin a lower, explicit budget.

`CBM_MAX_MEMORY_MB` lets the operator set the in-memory budget directly, with the same precedence shape as the existing `CBM_WORKERS` override: explicit override > implicit detection.

## Changes

- `src/foundation/mem.c`
  - New pure helper `cbm_mem_resolve_budget(total_ram, ram_fraction, max_memory_mb)` — single source of truth for the budget number (no globals/env), testable.
  - `cbm_mem_init` reads `CBM_MAX_MEMORY_MB` via `cbm_safe_getenv`, applies it, logs `source=env|ram_fraction` on the `mem.init` line, and warns on `mem.max.invalid` (non-numeric/≤0) / `mem.max.clamped` (above total RAM).
- `src/foundation/mem.h` — declare `cbm_mem_resolve_budget`; document the env var.
- `README.md` — add `CBM_MAX_MEMORY_MB` to the env-var table.
- `tests/test_mem.c` — 6 cases over the pure resolver.

## Semantics

| `CBM_MAX_MEMORY_MB` | Resulting budget |
|---|---|
| unset / empty | `ram_fraction × total_ram` (unchanged default) |
| positive integer ≤ total | exactly that many MiB |
| positive integer > total | clamped to `total_ram` (+ `mem.max.clamped` warn) |
| `0`, negative, non-numeric | ignored, fall back to fraction (+ `mem.max.invalid` warn) |

`mem.init` now logs the chosen source, e.g.:

```
level=info msg=mem.init budget_mb=5632 total_ram_mb=12288 source=env
```

## Design notes

- The override is clamped to `total_ram` when known so it can never claim more than physical/cgroup RAM, but is honored as-is when detection returns 0 (so it also rescues hosts where RAM detection fails).
- Budget math is extracted into a pure function because `cbm_mem_init` is one-shot per process (atomic CAS) and can't be re-exercised in-process.

## Testing

- `tests/test_mem.c` resolver cases.
- Verified standalone against the compiled `cbm_mem_resolve_budget` (12/12 checks pass).
- `clang-format --dry-run --Werror` clean on all changed files.
- Manual: `CBM_MAX_MEMORY_MB=512 codebase-memory-mcp …` → `mem.init … source=env`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,6 +444,7 @@ codebase-memory-mcp config reset auto_index              # reset to default
 | `CBM_DIAGNOSTICS` | `false` | Set to `1` or `true` to enable periodic diagnostics output to `/tmp/cbm-diagnostics-<pid>.json`. |
 | `CBM_DOWNLOAD_URL` | *(GitHub releases)* | Override the download URL for updates. Used for testing or self-hosted deployments. |
 | `CBM_LOG_LEVEL` | `info` | Set the minimum log level. Accepted values (case-insensitive): `debug`, `info`, `warn`, `error`, `none` — or their numeric equivalents `0`–`4` matching the internal enum. Logs go to stderr; stdout is reserved for MCP JSON-RPC. |
+| `CBM_MAX_MEMORY_MB` | *(50% of detected RAM)* | Explicit memory budget in MiB, overriding the default `ram_fraction × total RAM`. Caps the in-memory graph budget on RAM-constrained hosts, and lets containers pin a budget *below* the detected cgroup limit to leave headroom for sibling processes. Clamped to physical/cgroup RAM; non-positive/invalid values are ignored with a warning. |
 | `CBM_WORKERS` | *(detected)* | Override the parallel-indexing worker count returned by `cbm_default_worker_count`. Useful inside containers where `sysconf(_SC_NPROCESSORS_ONLN)` reports host CPUs rather than the cgroup's effective quota. Range 1–256; invalid values are ignored with a warning. |
 
 ```bash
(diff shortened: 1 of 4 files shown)


DeusData:

DeusData:

DeusData:
Thank you for this — the engineering was genuinely good (a pure, testable budget resolver with clamp-to-total-RAM and source logging). The reason we're closing it: main already has an equivalent explicit override, **CBM_MEM_BUDGET_MB** (checked first in cbm_mem_init, before any RAM-fraction logic, so it wins over the tiered default). Adding CBM_MAX_MEMORY_MB alongside it would give two env vars for one purpose with subtly different clamp semantics. If you'd like, your clamp + `source=` logging ideas would be a welcome small hardening of the *existing* CBM_MEM_BUDGET_MB path (extract its inline logic into your pure resolver, add the clamp + a README row it currently lacks) — happy to review that. Closing in favor of the existing knob; thanks for the solid work.

PR #605, opened 2026-06-24
docs: remove stale language and test counts

CooperSheroy:
## Summary
- remove stale fixed test-count wording from CONTRIBUTING.md
- avoid hard-coded language-count references in contributor docs and related comments

## Validation
- rg -n "2040 cases|64 languages|CBM_SZ_64" CONTRIBUTING.md src/discover/discover.h tests/test_language.c
- git diff --check

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ The binary is output to `build/c/codebase-memory-mcp`.
 scripts/test.sh
 ```
 
-This builds with ASan + UBSan and runs all tests (~2040 cases). Key test files:
+This builds with ASan + UBSan and runs the full C test suite. Key test files:
 - `tests/test_pipeline.c` — pipeline integration tests
 - `tests/test_httplink.c` — HTTP route extraction and linking
 - `tests/test_mcp.c` — MCP protocol and tool handler tests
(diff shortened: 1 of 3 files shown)


DeusData:

DeusData:
Merged — thank you, and sorry this sat for five weeks over four lines.

Documentation gets the same fact-check as code here, because a confidently wrong doc is worse than a stale one. Every number you removed was independently verified as genuinely stale, and by a wide margin:

- **Test count.** The doc claimed "~2040 cases". The suite is at roughly **6,855** today across 118 files. Stale by more than 3×.
- **Language count.** The doc said 64. The `CBMLanguage` registry has **162** concrete entries, with 160 vendored grammar directories.
- **`CBM_SZ_64`.** This one is the nicest catch: that comment in `discover.h` read as though `CBM_SZ_64` were a language count, but it is a generic 64-byte buffer-size constant used all over the tree — `pass_infrascan.c` and `pipeline_internal.h` among others. So the old comment was not merely stale, it was actively misleading, and `CBMLanguage` is exactly the right thing to point at instead.

**The reason this merges rather than getting a "please update the number" round trip is that you removed the numbers entirely.** Number-free wording cannot rot. Every hard-coded count in a doc is a small future lie waiting to happen, and this repo has several — replacing them with a description of *what* the thing is, rather than *how many*, is the durable fix.

Also worth noting for the record: two of the three touched files are source files, but both hunks are comment-only, so there is zero behavioural risk. All checks green including DCO.

Thank you for taking the time on documentation — it is the kind of contribution that rarely gets opened and always gets read.

Issue #384, opened 2026-05-27
Security: predictable /tmp filename in cbm-code-discovery-gate (symlink attack vector)


rileydev:
## Issue

The PreToolUse hook script `cbm-code-discovery-gate` (installed by `codebase-memory-mcp install` for Claude Code) uses a predictable filename in `/tmp`:

```bash
GATE=/tmp/cbm-code-discovery-gate-$PPID
find /tmp -name 'cbm-code-discovery-gate-*' -mtime +1 -delete 2>/dev/null
if [ -f "$GATE" ]; then
    exit 0
fi
touch "$GATE"
```

This is vulnerable to a symlink attack: an attacker who can predict `$PPID` (a small integer) and has write access to `/tmp` could pre-create a symlink at the predictable path pointing at an arbitrary target file. When the hook's `touch "$GATE"` runs, it would update the mtime of (or create) the target — potentially confusing other tools, or in shared environments writing into paths the user didn't intend.

The bare `[ -f "$GATE" ]` check also follows symlinks, so a malicious symlink would silently allow the gate to pass without the script actually creating its expected state file.

## Practical risk

Low for single-user dev containers (the `$PPID` prediction window is narrow and `/tmp` typically has the sticky bit set). Higher in shared/multi-tenant environments where the script could run with predictable PPIDs alongside untrusted users.

## Suggested fix

Two reasonable directions, paraphrased from an automated security review:

**(a) `mktemp` + per-session env var:**

```bash
: "${CBM_GATE_FILE:=$(mktemp -p "${XDG_RUNTIME_DIR:-/run/user/$(id -u)}" cbm-gate.XXXXXX)}"
GATE="$CBM_GATE_FILE"
```

**(b) User-private state directory:**

```bash
STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/cbm"
install -d -m 700 "$STATE_DIR"
GATE="$STATE_DIR/gate-$PPID"
```

Either approach, plus harden the file operations to refuse symlinks:

```bash
[ -e "$GATE" ] && [ ! -L "$GATE" ] && exit 0
install -m 600 /dev/null "$GATE"
```

And tighten the cleanup `find` to the chosen state dir rather than walking all of `/tmp`:

```bash
find "$STATE_DIR" -maxdepth 1 -type f -name 'gate-*' -mtime +1 -delete
```

## Context

[...]

PR #616, opened 2026-06-24
feat: OpenCode feature parity — plugin + skills + hooks

### LINKED ISSUE #585: feat: Full OpenCode feature parity with Claude Code — pre-tool hook, session-start context, and skill installation
## Summary

`codebase-memory-mcp install` configures OpenCode with only bare MCP server registration and a static `AGENTS.md` file. Claude Code receives three additional integration layers — a **PreToolUse hook** that auto-injects graph context during Grep/Glob, a **SessionStart reminder hook**, and a **consolidated skill** (`codebase-memory`) — that make the tool proactive rather than passive. OpenCode has equivalent hook primitives (`tool.execute.before`, `experimental.chat.system.transform`, plugin system) but the installer does not wire them up.

This issue requests full feature parity for OpenCode so that `codebase-memory-mcp install` delivers the same proactive integration depth as Claude Code.

### What Claude Code gets

| Integration | Mechanism | Location |
|---|---|---|
| MCP server | `.claude/.mcp.json` | — |
| Consolidated skill | `~/.claude/skills/codebase-memory/SKILL.md` | `cbm_install_skills()` — quick decision matrix, exploration/tracing workflows, quality analysis, 14 tool reference, Cypher examples, edge types, gotchas |
| PreToolUse hook | `settings.json` → `hooks.PreToolUse` with matcher `Grep|Glob` | `cbm_upsert_claude_hooks()` — runs `cbm-code-discovery-gate` shim which delegates to `hook-augment` subcommand; non-blocking (exit 0 on all failure paths); never intercepts `Read` |
| SessionStart hook | `~/.claude/hooks/cbm-session-reminder` | Injects one-line reminder to prioritize `search_graph`, `trace_path`, `get_code_snippet`, `query_graph`, `get_architecture`, `search_code` over raw grep |

### What OpenCode gets today

| Integration | Mechanism | Location |
|---|---|---|
| MCP server | `opencode.json` | `cbm_upsert_opencode_mcp()` |
| Static instructions | `~/.config/opencode/AGENTS.md` | Static file, no dynamic behavior |
| PreToolUse hook | **Missing** | — |
| SessionStart hook | **Missing** | — |
| Skill | **Missing** | — |

The OpenCode install path (`cli.c:3094`) calls only `install_generic_agent_config()` with `cbm_upsert_opencode_mcp`. Compare this to the Claude Code path which additionally calls `cbm_install_skills()`, `cbm_upsert_claude_hooks()`, and installs the `cbm-session-reminder` and `cbm-code-discovery-gate` shim scripts.

### 1. Pre-tool hook: `tool.execute.before` (equivalent to Claude Code PreToolUse)

OpenCode supports a `tool.execute.before` hook via its plugin system. This should be wired to intercept `grep` and `glob` tool calls (never `read` — same invariant as Claude Code) and inject structured graph context from `search_graph` as additional context, identical to what the `cbm-code-discovery-gate` shim does for Claude Code.

**Implementation approach:**

- Create a small plugin file (e.g. `~/.config/opencode/plugins/cbm-augment.ts` or `.js`) that exports a `tool.execute.before` hook
- The hook inspects the incoming tool name; if it matches `grep` or `glob`, extract the search token, call the `codebase-memory-mcp` binary in CLI mode (`codebase-memory-mcp cli search_graph '{"name_pattern": "..."}'`), and inject results as additional context
- Must be **non-blocking**: all failure paths (binary missing, no indexed project, query timeout) must return silently without preventing the original tool call
- Register the plugin in `opencode.json` under the `"plugin"` key

**Reference:** Claude Code equivalent is `cbm_upsert_claude_hooks()` in `cli.c` which registers the `cbm-code-discovery-gate` shim under `hooks.PreToolUse` with matcher `"Grep|Glob"`.

### 2. Session-start context injection (equivalent to SessionStart hook)

OpenCode supports `experimental.chat.system.transform` which can modify the system prompt sent to the LLM. This should inject the same codebase-memory reminder that `cbm-session-reminder` provides for Claude Code — instructing the agent to ALWAYS use `codebase-memory-mcp` tools first for code exploration, listing the specific tools (`search_graph`, `trace_path`, `get_code_snippet`, `query_graph`, `get_architecture`, `search_code`), and advising `grep`/`glob`/`read` for text/configs/non-code files.

**Implementation approach:**

- Either via the same plugin file (adding an `experimental.chat.system.transform` hook) or via a separate mechanism
- The injected text should be the same content as `cbm-session-reminder` (the session reminder script content defined in `cli.c`)
- Alternatively, this could be folded into the `AGENTS.md` file dynamically, but the system transform hook is more reliable since it runs on every session regardless of whether the agent reads AGENTS.md

**Reference:** Claude Code equivalent is `cbm-session-reminder` script installed to `~/.claude/hooks/` and registered under `hooks.SessionStart` in `settings.json`. Codex, Gemini CLI, and Antigravity all get a similar SessionStart reminder — OpenCode is the only hook-capable agent that does not.

### 3. Skill installation (equivalent to Claude Code `codebase-memory` skill)

Install the consolidated `codebase-memory` skill to OpenCode's skills/instructions directory. The skill content is already defined in the `skill_content` static variable in `cli.c` and includes:

- Quick Decision Matrix (question to tool mapping)
- Exploration Workflow (list_projects, get_graph_schema, search_graph, get_code_snippet)
- Tracing Workflow (search_graph, trace_path, detect_changes)
- Quality Analysis (dead code, fan-out, fan-in)
- 14 MCP Tools reference
- Edge Types reference
- Cypher query examples
- Gotchas and tips

**Implementation approach:**

- Install the skill to the appropriate OpenCode location (e.g. `~/.config/opencode/skills/codebase-memory/` or wherever OpenCode discovers instruction files)
- The skill content (`SKILL.md` with YAML frontmatter) can be reused as-is from the Claude Code skill — the tool names and workflows are agent-agnostic
- Add a call to `cbm_install_skills()` (or equivalent) in the OpenCode branch of `install_cli_agent_configs()`

**Reference:** Claude Code equivalent is `cbm_install_skills()` which writes `SKILL.md` to `~/.claude/skills/codebase-memory/`. The old 4-skill layout (`codebase-memory-exploring`, `codebase-memory-tracing`, `codebase-memory-quality`, `codebase-memory-reference`) was consolidated into a single skill (`CBM_SKILL_COUNT = 1`).

## Proposed Multi-Agent Support Table Update

The README table currently shows:

| Agent | MCP Config | Instructions | Hooks |
|---|---|---|---|
| Claude Code | `.claude/.mcp.json` | 4 Skills | PreToolUse (Grep/Glob graph augment, non-blocking) |
| OpenCode | `opencode.json` | `AGENTS.md` | — |

After this change:

| Agent | MCP Config | Instructions | Hooks |
|---|---|---|---|
| Claude Code | `.claude/.mcp.json` | 1 Skill (`codebase-memory`) | PreToolUse (Grep/Glob graph augment, non-blocking) |
| OpenCode | `opencode.json` | 1 Skill (`codebase-memory`) + `AGENTS.md` | `tool.execute.before` (grep/glob graph augment, non-blocking) + `experimental.chat.system.transform` (session reminder) |

## Technical Notes

- OpenCode's plugin system loads TypeScript/JavaScript files specified in `opencode.json` under the `"plugin"` key. A plugin exports a function receiving a context object and returns an object with hook implementations.
- The `tool.execute.before` hook receives `(input, output)` where `output.args` can be mutated before the tool runs. For graph augmentation, the hook should add context to the tool input or system context, not block the tool.
- The `experimental.chat.system.transform` hook allows modifying the system prompt — this is the correct place to inject the session-start reminder.
- All hooks must be **structurally non-blocking** (same design principle as Claude Code's `cbm-code-discovery-gate` which always exits 0).
- The `hook-augment` subcommand of the `codebase-memory-mcp` binary already exists and is agent-agnostic — it takes a search token and returns graph context. The OpenCode plugin can call this directly via `BunShell` (`$`) or `child_process`.

## Uninstall

`codebase-memory-mcp uninstall` should also remove:
- The plugin file from `~/.config/opencode/plugins/`
- The plugin entry from `opencode.json` `"plugin"` array
- The skill from the OpenCode skills directory

## Environment

- OS: Linux (but should work cross-platform)
- OpenCode: latest, global config `~/.config/opencode/opencode.json`
- codebase-memory-mcp: v0.8.x

## Related

- #437 — OpenCode MCP registration bugs (closed, config shape fixed)
- #221 — Windows install broken for OpenCode (open)
- The README claims "11 agents, one command — configures MCP entries, instruction files, and pre-tool hooks for each" but OpenCode does not receive hooks or skills, making this claim inaccurate for OpenCode.

ZeR020:
## Summary

Closes #585. Adds full OpenCode feature parity with Claude Code:

1. **TypeScript plugin** (`~/.config/opencode/plugins/cbm-augment.ts`) — auto-discovered, no `opencode.json` entry needed
2. **Consolidated `codebase-memory` skill** (`~/.config/opencode/skills/codebase-memory/SKILL.md`) — same skill content as Claude Code
3. **Uninstall path** — removes plugin, skills, MCP config, and AGENTS.md

## Design Decision: `tool.execute.after` (not `.before`)

Issue #585 proposed `tool.execute.before` for context injection. However, source verification against `anomalyco/opencode` proved that `.before` is **non-functional** for this purpose:

**`tool.execute.before`** (`packages/opencode/src/session/tools.ts:82`):
```ts
yield* plugin.trigger("tool.execute.before",
  { tool: item.id, sessionID, callID },  // input — NO args field
  { args },                                // output — fresh wrapper object
)
const result = yield* item.execute(args, ctx)  // uses ORIGINAL args
```
- Input does NOT include `args`. Output `{ args }` is a fresh object; mutation has zero effect — the tool uses the original `args` variable, not the mutated output.
- Runs BEFORE the tool produces a result — there's nothing to append to.

**`tool.execute.after`** (`packages/opencode/src/session/tools.ts:92`):
```ts
const output = { ...result, attachments: [...] }
yield* plugin.trigger("tool.execute.after",
  { tool, sessionID, callID, args },  // input — INCLUDES args
  output,                               // output — actual tool result
)
return output  // model sees this
```
- `output.output` (type: `string`) is model-visible tool result text.
- Mutating `output.output` in place IS model-visible.
- `input.args` available (has `pattern` for grep/glob).

This PR uses `.after` so that graph context is appended to the tool result the model actually sees.

### Plugin (`cbm-augment.ts`)

**`tool.execute.after` hook:**
- Checks if `input.tool` is `"grep"` or `"glob"` (OpenCode uses lowercase tool names)
- Maps to capitalized `"Grep"`/`"Glob"` for the agent-agnostic `hook-augment` binary
- Spawns `<binary> hook-augment`, writes `{"tool_name":"Grep","tool_input":{"pattern":"..."}}` to stdin
- Parses `hookSpecificOutput.additionalContext` from stdout
- Appends to `output.output` with `\n\n` separator
- **All failures silently swallowed** — never blocks or throws (consistent with Claude Code's `cbm-code-discovery-gate`)

**`experimental.chat.system.transform` hook:**
- Pushes the session reminder string to `output.system` (rebuilt fresh each turn, no accumulation)
- Message consistent with `CMM_SESSION_REMINDER_CMD` used by Codex/Gemini/Antigravity

### Install/Uninstall

Install path (`install_cli_agent_configs`):
- Calls `cbm_install_skills(skills_dir, true, dry_run)` — reuses existing function
- Calls `cbm_upsert_opencode_plugin(home, binary_path, dry_run)` — writes plugin with embedded binary path
- Plan mode (`--plan`) records `skills` + `plugin` entries without mutating

Uninstall path (`uninstall_cli_agents`):
- Calls `cbm_remove_skills(skills_dir, dry_run)`
- Calls `cbm_remove_opencode_plugin(home, dry_run)` — unlinks plugin file (unlike Claude Code which leaves inert scripts, OpenCode auto-discovers plugins so the file must be deleted)

### Security

- **TOCTOU-safe**: `fchmod(fileno(f), ...)` before `fclose(f)` (same pattern as `cbm_install_hook_gate_script`)
- **Quote rejection**: binary paths containing `"` are rejected (prevents injection into the TypeScript `BIN = "%s"` string)
- **No predictable temp files** (issue #384)
- **Non-blocking**: all plugin failures silently swallowed

## Tests

5 new tests in `tests/test_cli.c`:
- `cli_upsert_opencode_plugin_fresh` — verifies plugin file written with correct binary path, hook names, and `satisfies Plugin`
- `cli_upsert_opencode_plugin_idempotent` — verifies re-install overwrites cleanly with new path
- `cli_upsert_opencode_plugin_rejects_quote` — verifies security: binary paths with `"` are rejected
- `cli_remove_opencode_plugin` — verifies install → remove → file gone, and idempotent remove
- `cli_opencode_skills_installed` — verifies SKILL.md written to correct path with correct content

All 5 pass. 5698 total tests pass. 1 pre-existing failure (`incr_full_index` in `test_incremental.c`) is unrelated — an RSS memory threshold assertion that's environment-dependent.

## Files Changed

- `src/cli/cli.c` (+149 lines): `opencode_plugin_content` static string, `cbm_upsert_opencode_plugin()`, `cbm_remove_opencode_plugin()`, install/uninstall path modifications
- `src/cli/cli.h` (+9 lines): function declarations
- `tests/test_cli.c` (+130 lines): 5 new tests + `RUN_TEST` registrations


DeusData:

DeusData:
Reviewed alongside the two other agent-install PRs in the queue, since all three occupy the same area. Half of this has shipped; the other half is a direction question.

**The skills half is now on main.** Commit `a3903ca` (12 July) rebuilt the client-surface machinery and installs the consolidated skill for OpenCode at `~/.config/opencode/skills` plus an agent profile. So that part is covered.

**The plugin half is genuinely novel** — `git log -S 'opencode/plugins'` on main is empty, so nothing equivalent exists. And the design work behind it is good: your PR body proves `.before` is non-functional upstream by reference to the actual OpenCode source, which is exactly the kind of evidence that makes a reviewer's job easy. The `tool.execute.after` → `hook-augment` → `additionalContext` path is a sensible shape.

**Why the plugin is a maintainer call rather than a merge:**

- It ships a **persistent executable artifact** into a user's config directory that spawns our binary on every grep and glob. That is the point of the feature, but it is also precisely the class of artifact our recent antivirus work has been trimming, so it needs a deliberate yes.
- It binds us to OpenCode's **`experimental.`** namespace (`experimental.chat.system.transform`) and to `tool.execute.after` output-mutation semantics verified against one upstream revision. Both can shift under us, and once shipped we support it indefinitely.

I have put that to the maintainer. The instruction content itself is clean — I checked specifically, since shipped agent-facing text is content an agent will follow, and yours is product-consistent steering with no goal manipulation and no leaked personal context.

**Two defects to fix if it does go ahead:**

1. **Raw `fopen()` in `cbm_upsert_opencode_plugin`.** Our rule is `cbm_fopen()` (UTF-8 → `_wfopen`) for user-home paths — raw `fopen` breaks on a non-ASCII Windows home directory.
2. **Backslash is not escaped in the plugin template.** You correctly reject binary paths containing `"`, but a Windows path like `C:\Users\...` gets embedded into the TypeScript string literal `BIN = "%s"` and produces mangled escapes. The quote check alone is not sufficient there.

Smaller: under `--dry-run` the plugin and skills success lines print unconditionally, and the `if (!dry_run)` wrapper around `cbm_upsert_opencode_plugin` is redundant since the function already handles dry-run itself.

**Your tests are good** — five round-trips covering fresh install, idempotent overwrite, quote rejection, and remove-then-remove-again. Worth saying, because install code is easy to leave untested.

Please hold off reworking until the direction answer arrives; the whole block needs re-basing onto the `a3903ca` structure and I would rather you did that once.

DeusData:
Thank you for this, and I am sorry it took as long as it did to come back with a real answer.

**The decision: OpenCode stays an MCP-only integration, so we are not taking the plugin.** The reasoning is about what this project ships rather than about the quality of your work, and I want to give you the whole of it.

## Why

`codebase-memory-mcp` is an MCP server, and today it ships exactly one kind of artifact into a user's config: text. Markdown, JSON, `.sh`, `.cmd`, `.ps1`. **Zero `.ts` or `.js` files** — I checked before deciding, rather than assuming. Taking this would make it the first, and it would be executable, auto-discovered by OpenCode with no user action, `chmod 0755`, and interpreted by a third-party runtime. Once one client gets a shipped JS artifact the others reasonably ask for one too, and that is a maintenance surface we do not want to own.

There is a second reason, specific to the mechanism: the plugin binds us to OpenCode's **`experimental.chat.system.transform`** and to in-place mutation of `output.output`, which you reverse-engineered from their source rather than from a contract. Every failure mode of that design is *silent* — if they rename or re-wrap either, the augmentation simply stops, with no error and no signal, and we find out through a support report. Owning a compatibility surface against another product's experimental API, on a default install path, is a real ongoing cost.

## What OpenCode users already get

This is the part that made the decision easier, and it is worth stating plainly: **OpenCode already has the full tool surface.** `cbm_upsert_opencode_mcp` registers the server in `~/.config/opencode/opencode.json`, so all fifteen tools are available today. Main also installs `AGENTS.md`, the skill, and three deny-by-default read-only agents.

So what your plugin adds is not capability — it is the *automatic* grep/glob reflex. Real value, but a narrower gap than it first appears, and the way we deliver that elsewhere is through each client's own native hook configuration (Claude Code, Codex, Copilot, Gemini, Factory each have one). OpenCode does not have an equivalent, which is exactly why you had to write a plugin. That is a gap in OpenCode's extension model, not something we want to route around by shipping a runtime artifact.

## Two things you should know regardless

**The skills half of your PR already shipped.** Main installs the OpenCode skill and its uninstall path, with tests, and went further with an agent profile — `mode: subagent`, `edit: deny`, `bash: deny`. That happened independently while this was open, which is on our queue, not on you.

**And the plugin does not currently run.** I checked empirically rather than assuming: the payload sends `{"tool_name":"Grep",...}` with no `hook_event_name`, and current `hook-augment` requires that field and accepts Grep/Glob only under `PreToolUse` — so it produces **zero bytes** of output. This is bit-rot rather than an error on your part: at your merge base the hook keyed on `tool_name` alone, and main's multi-dialect hardening changed the contract while the PR waited. Worth knowing so you do not spend time debugging something that was working when you wrote it.

One more that would have bitten a Windows user: the `const BIN = "%s"` template rejects `"` but not `\`, so a path like `C:\Users\urs\…` produces a hard `SyntaxError` in an auto-loaded plugin directory — which is worse than the feature not working, because it can break OpenCode's plugin subsystem at session start. Mentioning it in case the code lives on in your own fork.

## If the ground shifts

If OpenCode ever ships a declarative hook configuration like Claude Code's or Codex's, this becomes a straightforward addition and we would want it — I will come back to this PR and to you if that happens.

Thank you again. Reverse-engineering the hook points from their source, honouring abort signals, and scanning stdout for the last parseable JSON line rather than trusting the first are all the instincts of someone who has debugged this class of integration before.

DeusData:
Following up because the decision changed.

**The OpenCode augmentation is being implemented — #1392, with `Co-Authored-By: ZeR020` on the commit.**

When I closed this, the objection was shipping a `.ts` artifact from our repository. The maintainer has since asked for the capability on the condition that no `.ts` file lives in our tree, so the plugin is now **generated at install time** and written into `~/.config/opencode/plugins/cbm-augment.ts` as a marked block — which keeps a user's own module in that auto-loaded directory intact.

Your hook-point research is what made this possible at all. Working out that `tool.execute.after` is where output can be augmented, from their source rather than from documentation, is the part that would have taken me longest.

**Two defects from the original are closed by construction, and both are pinned by tests** — mentioning them because they are worth knowing if the code lives on in your own fork:

- **The payload omitted `hook_event_name`.** Current `hook-augment` requires it and accepts Grep/Glob only under `PreToolUse`, so the plugin emitted **zero bytes** — it had been a silent no-op since main's multi-dialect hardening changed the contract underneath you. The generated payload carries the field.
- **The path template rejected `"` but not `\`.** A home like `C:\Users\urs\bin` produced an invalid unicode escape, so the whole plugin failed to parse — worse than not working, since a broken module in an auto-loaded directory can take out the plugin subsystem at session start. The generator escapes backslash, quote, newline and CR, and refuses to emit anything rather than produce a truncated literal.

One thing I recorded rather than papered over: `tool.execute.after`'s ability to modify output is **not** in OpenCode's documented plugin contract — only `tool.execute.before`'s argument mutation is. So if they change it, augmentation stops silently. We are accepting that knowingly, and it is written in a comment at the emitter so nobody rediscovers it as a mystery later. Your instinct to reverse-engineer it was sound; the risk is inherent to the extension point, not to your implementation.

Thank you for the original work, and for the care in it.

Issue #554, opened 2026-06-22
C++ out-of-line method definitions: CALLS edge source falls back to Module (file-level) instead of the enclosing Method


spwlyzx:
### Version

0.8.1

### Platform

macOS (Apple Silicon)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

For C++ out-of-line method definitions (`void Foo::Bar() { ... }` in a `.cpp` file), CALLS edges emitted from call sites inside the method body have **source = Module (file-level node)** instead of **source = the enclosing Method node**. This causes `trace_path(direction=inbound)` to return file paths rather than specific caller method names, and prevents multi-hop method-to-method BFS traversal.

**Expected:** CALLS edges should have `source = Method("Foo::Bar")`, matching how in-class definitions are handled.

**Actual:** CALLS edges have `source = Module("path/to/file.cpp")`. The `calls_find_source()` fallback to `__file__` fires because `call->enclosing_func_qn` does not match the graph node's qualified name after PR #428 promoted out-of-line definitions from Function to Method (changing their QN).

### Reproduction

**Reproduction:**

1. Create a minimal C++ project:

```cpp
// include/factory.h
#pragma once
class TileDataFactory {
public:
    void CreateTileData(int level);
};

// src/factory.cpp
#include "factory.h"
void TileDataFactory::CreateTileData(int level) {}  // out-of-line

// src/consumer.cpp
#include "factory.h"
class DataProvider {
public:
    void ConvertBlob();   // in-class declaration
};

void DataProvider::ConvertBlob() {   // out-of-line definition
    TileDataFactory f;
    f.CreateTileData(1);   // <-- this call site
}
```

2. Index and query:

```bash
codebase-memory-mcp cli index_repository '{"repo_path": "/tmp/repro"}'
codebase-memory-mcp cli query_graph '{"query": "MATCH (src)-[:CALLS]->(dst {name: \"CreateTileData\"}) RETURN src.name, src.label, dst.name, dst.label", "project": "..."}'
```

3. **Actual output:** `src.name = "src/consumer.cpp"`, `src.label = "Module"`
4. **Expected output:** `src.name = "ConvertBlob"`, `src.label = "Method"` (with `parent_class = DataProvider`)

---

**Root cause (from source inspection):**

In `src/pipeline/pass_calls.c` L320–333, `calls_find_source()` first tries `enclosing_func_qn`:

```c
[...]

PR #621, opened 2026-06-25
fix(extract): attribute calls inside out-of-line C++ methods to the method

KerseyFabrications:
Follow-up to #463 (out-of-line attribution gap reported there).

### Problem
A call inside a C++ out-of-line method definition attributes to the **File** node instead of the enclosing method — including when the definition is
wrapped in a `namespace {}` block.

### Root cause
Not namespace context (which never enters the C++ QN scheme — `namespace_name` is null and the namespace appears in no QN), but a **dropped class
qualifier**. The call-side enclosing-QN computation produced `project.path.Bar` instead of `project.path.Foo.Bar` for out-of-line defs (no enclosing class
AST node / no class scope on the walk stack). The pipeline resolves a call's source by exact QN (`calls_find_source` → `find_by_qn`), so the mismatch falls
straight back to the `__file__` node. Identical for the global (`using namespace`) and namespace-block forms.

### Fix
The out-of-line class resolution lived in the defs extractor but had two divergent, incomplete copies on the call side — the same drift behind #438.
Consolidated rather than adding a fourth copy:
- `cbm_cpp_out_of_line_parent_class` — promoted from `extract_defs.c` into `helpers`.
- `cbm_cpp_out_of_line_method_qn` — new shared helper building the class-scoped QN, used by both `compute_func_qn` (CALLS, unified walk) and
`cbm_enclosing_func_qn` (cached path: USAGES/THROWS/CONFIGURES/type-assigns).

Corrects attribution for **every** edge type inside out-of-line methods, not just CALLS.

### Tests
- `cpp_out_of_line_enclosing_qn` — extraction-level QN match for global / namespace-block / nested-namespace forms.
- `pipeline_cpp_out_of_line_call_attribution` — full index of a header + `using namespace` .cc + `namespace {}` .cc; both `Baz()` calls source from the
`Method` node, not the file.

Full suite green apart from a pre-existing, unrelated ASan RSS-budget flake.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -458,56 +458,6 @@ static TSNode resolve_func_name_fp(TSNode node, CBMLanguage lang, const char *ki
     return null_node;
 }
 
-// C++/CUDA: out-of-line method definitions name the function with a qualified
-// declarator (`Foo::bar`, or `ns::Foo::bar`). Return the immediate enclosing
-// class name (the scope segment directly left of the function name, e.g. "Foo"),
-// or NULL when the declarator is unqualified (a plain free function). Without
-// this, an out-of-line definition — whose class body lives declaration-only in a
-// header — would be recorded as a free Function with no link to its class.
-static char *cpp_out_of_line_parent_class(CBMArena *a, TSNode node, const char *source) {
-    // Descend the declarator chain to its qualified_identifier, if any.
-    TSNode qid = {0};
-    TSNode decl = ts_node_child_by_field_name(node, TS_FIELD("declarator"));
-    for (int depth = 0; depth < DECLARATOR_DEPTH_LIMIT && !ts_node_is_null(decl); depth++) {
-        const char *dk = ts_node_type(decl);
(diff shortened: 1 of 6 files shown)


KerseyFabrications:
This was converted to draft until #463 is merged.

KerseyFabrications:
All conflicts resolved. All tests should pass.

DeusData:

KerseyFabrications:
Superseded by the `cbm_resolve_func_name` / shared-helper extraction refactor now on `main`, which implements the same out-of-line C++ method call-attribution fix in `compute_func_qn` (class-scoped enclosing QN via `cbm_cpp_out_of_line_parent_class`, same #554/#621 reference). The core fix here is therefore redundant — closing.

Two bits from this PR are *not* covered by that refactor and will follow as a small separate PR: (1) the same out-of-line reconstruction on the **cached** enclosing-QN path (`cbm_enclosing_func_qn`), so usages/throws/configures/type-assigns inside an out-of-line method also attribute to the method rather than the file; and (2) a pipeline-level integration test for out-of-line call attribution (existing coverage is extraction-level only).

Issue #507, opened 2026-06-18
Support Markdown Architectural Decision Records


Arimals:
### What problem does this solve?

Currently, only a single source of truth is supported for ADR, you cannot maintain a structure and have multiple records.

### Proposed solution

1. Support multiple ADR record(s) by assigning a numeric value to each record and specify keywords/phrases that can be picked up in MCP queries.
2. Follow MADR specifications: https://adr.github.io/madr/decisions/0013-use-yaml-front-matter-for-meta-data.html
3. Allow records to be queried based on direct ADR tool calls and or query/searches for code that may be a result of an ADR record.

### Alternatives considered

Reference MADR or individual ADR records already present in the repository. The lack of true MADR and metadata tracking limits the exposure and potential of current ADR support implementation. Doing code queries or ADR queries will not yield the desired ADR record leveraged in context searches but rather make the agent have to go and query each record mentioned in the main ADR of the indexed project.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
Great addition, will check as soon as there is capacity

DeusData:
Linking #504 here — "import existing ADR docs already present in the repo" is a subset of this MADR proposal and should be covered by the same work. Consolidating the two so they're tracked together.

Issue #504, opened 2026-06-18
Make the tool import existing ADR docs in a repository


rikbosch:
### What problem does this solve?

Most repositories contain existing ADR documentation that might be of interest codebase-memory.

### Proposed solution

It would be nice if the initial project onboarding would import those docs as well.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
Thanks so much for this, @rikbosch 🙏 It fits naturally with the existing `manage_adr` tool, which already stores ADRs in the graph; we just don't bootstrap from what's already on disk yet. For a first cut we'd detect a conventional ADR location (`docs/adr/`, `adr/`, or MADR style `NNNN_*.md` filenames) and import them so `get_architecture` and `manage_adr` surface them.

Two quick questions: (1) auto detect, or a config key like `adr_path`? (2) one time import on first index, or keep them in sync via the watcher? If you have a preferred ADR tool (MADR or log4brains), let us know which to target first. PRs very welcome!

rikbosch:
Maybe a combination? Scavenge most user paths line docs/adr or other known paths like **/adr/*.md?

I'd keep them in sync just like the rest of the code. The adr documents should be immutable, but they can be superseded by newer adr documents.

DeusData:
This overlaps heavily with #507, which proposes the broader MADR multi-record support — including importing existing ADR records already present in a repo. Marking this `duplicate` and consolidating the discussion under #507 so the "import existing docs" piece is tracked as part of that work. Leaving this open for visibility for now. Thanks for raising it.

DeusData:
Re-checked this under the stricter duplicate rule we are using now: this is related to #507, but it is not a 100% duplicate.

#507 is the broader Markdown/MADR support track. This issue is specifically about importing existing ADR documents during onboarding/indexing, so I am keeping it open as a distinct acceptance criterion and removing the `duplicate` label. We should still implement it in coordination with #507.

Issue #256, opened 2026-04-15
`manage_adr` (MCP/CLI) and the UI `/api/adr` endpoints use separate storage backends


bschrib:
The MCP tool `manage_adr` writes/reads ADRs as files at `<root>/.codebase-memory/adr.md`. The HTTP UI's `/api/adr` endpoints read/write SQLite via `cbm_store_adr_get` / `cbm_store_adr_store`. Writes via either interface are invisible to the other.

**Reproduction (v0.6.0, macOS arm64):**

```
# write via CLI
codebase-memory-mcp cli manage_adr '{"project":"<name>","mode":"update","content":"## PURPOSE\ntest\n"}'

# CLI roundtrips
codebase-memory-mcp cli manage_adr '{"project":"<name>","mode":"get"}'   # -> returns content
cat <repo>/.codebase-memory/adr.md                                         # -> content present

# UI endpoint
curl http://localhost:9749/api/adr?project=<name>                          # -> {"has_adr": false}
```

The UI's `+ ADR` button stays `+ ADR` because `hasAdr` comes from that `/api/adr` response (`graph-ui/src/components/StatsTab.tsx:111`).

**Code pointers:**

- `src/mcp/mcp.c:3498-3555` — `handle_manage_adr`, file-backed at `<root>/.codebase-memory/adr.md`.
- `src/ui/http_server.c:449-559` — `handle_adr_get` / `handle_adr_save`, SQLite-backed via `cbm_store_adr_*` (`src/store/store.c:4537,4560`).

**Expected:** one canonical backend. The file-on-disk path is likely the right target since it's committable and lives with the repo; the SQLite-backed UI handlers would need to switch to reading/writing the file.

**Impact:** ADRs created by agents via MCP don't surface in the graph UI. ADRs created by humans in the UI aren't readable by agents.

DeusData:
Thanks — clean diagnosis. The two interfaces did use separate backends: `manage_adr` read/wrote `<root>/.codebase-memory/adr.md` (a file), while the UI `/api/adr` GET/POST used the SQLite `project_summaries` table via `cbm_store_adr_get`/`cbm_store_adr_store`. Writes on one side were invisible to the other.

[...]

PR #623, opened 2026-06-25
fix(pipeline): preserve ADRs across re-indexing (#516)

### LINKED ISSUE #516: manage_adr data loss: ADRs in project_summaries are deleted during graph re-indexing

yasku:
## What does this PR do?

Fixes #516 — **`manage_adr` ADRs are silently lost on re-index.**

`manage_adr` stores ADRs in the `project_summaries` SQLite table (the store backend unified in #256). The custom B-tree writer (`internal/cbm/sqlite_writer.c`) serializes a fresh database with an **empty** `project_summaries` btree on every dump, so any re-index that rewrites the DB silently wipes the stored ADR — `manage_adr(mode="get")` then returns `no_adr`.

### Root cause — the loss happens at **two** dump sites, not one

The issue report (and its confirming comment) identified only the full-reindex path. Tracing the pipeline showed the incremental path loses the ADR too:

| Path | Function | ADR? |
|------|----------|------|
| Incremental, no file changes | early-return, no dump | ✅ survives |
| Incremental, a file changed/added | `dump_and_persist` (`pipeline_incremental.c`) | ❌ wiped |
| Full re-index | `dump_and_persist_hashes` (`pipeline.c`) | ❌ wiped |

Both dump functions already re-persist `file_hashes` after the rewrite but had **no equivalent step for `project_summaries`** — that asymmetry is the bug. Fixing only the full path (as the report suggested) would have left the bug live on any single-file edit.

### The fix — mirror the existing `file_hashes` preservation

- Capture the ADR from the old DB **before** it is rewritten, in `try_incremental_or_delete_db()` — the single choke point that opens the old DB for both paths.
- Carry it on the pipeline (`carried_adr`, exposed to the incremental translation unit via a `cbm_pipeline_carried_adr()` accessor, matching the project's existing opaque-struct accessor pattern e.g. `cbm_pipeline_repo_path()`).
- Re-insert it with `cbm_store_adr_store()` after each dump, right next to the `file_hashes` restore.

Behavior is otherwise unchanged (absent DB → skip, healthy → reindex, corrupt → preserved as `.corrupt`). Memory: `carried_adr` is `strdup`'d, freed once in `cbm_pipeline_free`, NULL-initialized via the existing `calloc`.

## How it was researched (provenance)

The root cause and fix shape were grounded in the repo's own history and source, not assumed:

- **#256** (`manage_adr` storage) — closed by commit `70e3ed7` *"fix(adr): unify manage_adr onto the SQLite store shared with the UI"*. Confirms the current backend is the SQLite `project_summaries` table (`cbm_store_adr_get/store`), i.e. the right place to fix.
- **`internal/cbm/sqlite_writer.c`** — the writer emits `project_summaries` as an empty btree by design (`// Autoindex for project_summaries — empty (0 rows)`), confirming the wipe is structural, not incidental.
- **`file_hashes` preservation** (`pipeline.c` / `pipeline_incremental.c`) — the established precedent this fix mirrors: capture-before-rewrite, re-insert-after-reopen.
- **#125** (prior `manage_adr` use-after-free, closed) — reviewed to confirm the ADR code path is sensitive; this PR does not touch `mcp.c` ADR handling.
- **Forward-compat note:** #504 / #507 propose multiple/MADR ADRs. The current schema is one row per project (`project TEXT PRIMARY KEY`); this fix matches that. If multi-ADR lands later, the carry generalizes to all `project_summaries` rows.
- **SQLite architecture check:** SQLite's native cross-DB copy mechanisms (Online Backup API, `ATTACH` + `INSERT … SELECT`) require both sides to be engine-managed databases. The dump here is a hand-serialized B-tree file (`sqlite_writer.c`), so those native mechanisms do not apply to it — the correct approach is the app-level capture/re-insert via the public store API, which is exactly the existing `file_hashes` pattern.

## Checklist

- [x] Every commit is signed off (`git commit -s`) — DCO verified locally via `scripts/check-dco.sh`
- [x] Tests pass locally (`make -f Makefile.cbm test`) — **5687 passed**, 0 failures
- [x] Lint passes — `clang-format` clean; `cppcheck` (CI suppress flags) exits 0 on all changed files
- [x] New behavior is covered by a test (reproduce-first) — see below

## Tests added

Both regression tests live in `tests/test_pipeline.c` and assert the ADR is retrievable after a re-index (store an ADR, re-index, `cbm_store_adr_get` must still return it):

- `pipeline_full_reindex_preserves_adr_issue516` — adds files to force the **full** dump path.
- `pipeline_incremental_reindex_preserves_adr_issue516` — modifies one file (same file count) to force the **incremental** dump path.

Also verified end-to-end via the CLI: index → `manage_adr update` → change a source file → re-index → `manage_adr get` returns the stored ADR (returned `no_adr` before the fix).

## Validation

All checks were run locally and pass:

| Check | Command | Result |
|-------|---------|--------|
| Full test suite | `make -f Makefile.cbm test` | ✅ 5687 passed |
| Build (prod) | `make -f Makefile.cbm cbm` | ✅ no warnings (`-Werror`) |
| Formatting | `clang-format` (changed files) | ✅ clean |
| Static analysis | `cppcheck` w/ CI suppress flags (changed files) | ✅ exit 0 |
| No-skips policy | `scripts/check-no-test-skips.sh` | ✅ OK |
| Security audit | `scripts/security-audit.sh` | ✅ passed |
| DCO | `scripts/check-dco.sh` | ✅ signed |

### Test environment

| | |
|-----|-----|
| Machine | Apple M1, 8-core (4P + 4E), 8 GB RAM |
| OS | macOS 26.5 (build 25F71), arm64 (Darwin 25.5.0) |
| Compiler | Apple clang 21.0.0 (clang-2100.1.1.101) |
| clang-format | 22.1.8 |
| cppcheck | 2.21.0 |

> Note: the one `unreadVariable` finding cppcheck surfaced was a pre-existing unused `rc` assignment in the unrelated `usages_creates_edges` test that the larger file exposed; this PR drops that dead assignment so the changed file stays lint-clean. No other unrelated code is touched.

## Scope

Four files, focused on #516:
- `src/pipeline/pipeline.c` — struct field, accessor, capture, full-path restore.
- `src/pipeline/pipeline_incremental.c` — incremental-path restore.
- `src/pipeline/pipeline_internal.h` — accessor declaration.
- `tests/test_pipeline.c` — two regression tests (+ the dead-assignment cleanup noted above).

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -789,20 +789,42 @@ static cbm_store_t *resolve_store(cbm_mcp_server_t *srv, const char *project) {
     project_db_path(project, path, sizeof(path));
     srv->store = cbm_store_open_path_query(path);
     if (srv->store) {
-        /* Check DB integrity — auto-clean corrupt databases */
+        /* Check DB integrity — preserve corrupt databases instead of
+         * silently deleting them so the user can inspect the evidence and
+         * recover or report the issue (#557). Rename to .corrupt.<ts> so
+         * the path is obvious in the cache directory. */
         if (!cbm_store_check_integrity(srv->store)) {
-            cbm_log_error("store.auto_clean", "project", project, "path", path, "action",
-                          "deleting corrupt db — re-index required");
             cbm_store_close(srv->store);
             srv->store = NULL;
-            /* Delete the corrupt DB + WAL/SHM files */
-            cbm_unlink(path);
+            /* Rename the corrupt DB + WAL/SHM to preserve evidence */
             char wal_path[MCP_FIELD_SIZE];
             char shm_path[MCP_FIELD_SIZE];
+            char corrupt_path[MCP_FIELD_SIZE];
+            char corrupt_wal[MCP_FIELD_SIZE];
+            char corrupt_shm[MCP_FIELD_SIZE];
+            long long now_s = (long long)time(NULL);
             snprintf(wal_path, sizeof(wal_path), "%s-wal", path);
             snprintf(shm_path, sizeof(shm_path), "%s-shm", path);
-            cbm_unlink(wal_path);
-            cbm_unlink(shm_path);
+            snprintf(corrupt_path, sizeof(corrupt_path), "%s.corrupt.%lld", path, now_s);
+            snprintf(corrupt_wal, sizeof(corrupt_wal), "%s-wal.corrupt.%lld", path, now_s);
+            snprintf(corrupt_shm, sizeof(corrupt_shm), "%s-shm.corrupt.%lld", path, now_s);
+            if (rename(path, corrupt_path) != 0) {
+                cbm_unlink(path); /* fallback: delete if rename fails */
+            }
+            /* best-effort: sidecars may not exist */
+            if (rename(wal_path, corrupt_wal) != 0) {
+                cbm_unlink(wal_path);
+            }
+            if (rename(shm_path, corrupt_shm) != 0) {
+                cbm_unlink(shm_path);
+            }
+            cbm_log_error("store.auto_clean", "project", project, "path", path, "action",
+                          "corrupt db preserved", "preserved_as", corrupt_path);
+            (void)fprintf(stderr,
+                          "ERROR corrupt project database: %s\n"
+                          "ERROR   preserved as: %s\n"
+                          "ERROR   inspect the .corrupt file, delete it when done, and re-index.\n",
+                          path, corrupt_path);
             return NULL;
         }
 
--- a/src/pipeline/pipeline.c
+++ b/src/pipeline/pipeline.c
@@ -97,6 +97,11 @@ struct cbm_pipeline {
     /* Committed graph size at dump time (-1 = dump did not run). #334 gate axis. */
     int committed_nodes;
     int committed_edges;
+
+    /* ADR (project_summaries) content read from the old DB before a rewrite,
+     * re-inserted after the dump so manage_adr survives re-indexing (#516).
+     * NULL when no ADR was stored. Owned by the pipeline; freed in _free. */
+    char *carried_adr;
 };
 
 /* ── Global pkgmap (one active pipeline at a time) ─────────────── */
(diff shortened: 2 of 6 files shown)


yasku:
Closing this PR. Issue #516 has since been addressed on `main` via #539, so this alternative implementation is no longer needed. I'm also moving my remaining work to a private repository. Thanks for the reviews.

Issue #349, opened 2026-05-14
index_status returns project-not-found for project returned by list_projects


edecarvalhoreis:
## Summary
`codebase-memory-mcp cli index_status` can return `project not found or not indexed` for a project name that is returned by `list_projects` in the same environment.

## Version
- `codebase-memory-mcp 0.6.1`

## Reproduction
1. `codebase-memory-mcp cli list_projects`
2. Pick the returned project name, for example `sample-project`.
3. `codebase-memory-mcp cli index_status '{"project":"sample-project"}'`

## Actual
`index_status` fails with:
```json
{"error":"project not found or not indexed","hint":"Use list_projects to see all indexed projects, then pass the project name.","available_projects":["sample-project"],"count":1}
```

## Expected
`index_status` should accept the same project id/name returned by `list_projects` and return status metadata.

## Additional signal
`index_repository` may fail with `pipeline.err phase=dump` and a generic pipeline hint, which makes diagnosis difficult.

## Impact
Breaks automation paths that depend on `index_status` for health/selector checks.

## Suggested fix direction
- Align selector parsing for `index_status` with `list_projects` output.
- Surface concrete dump-phase failure reason (DB path/permission/state) instead of a generic hint.

DeusData:
Thanks — and the giveaway was your own log: the error lists `available_projects:["sample-project"]` while claiming the same name isn't found. That happens because the two code paths disagree:

- `list_projects` (and the error helper) enumerate by **opening the `.db` files directly**.
- `index_status`/`search_graph`/etc. go through `resolve_store`, which first runs the name through `cbm_validate_project_name` (a security gate that allows only `[A-Za-z0-9._-]`).

The project name is derived from the repo path by `cbm_project_name_from_path`, which only replaced `/` and `:` with `-` — leaving **spaces, `@`, `+`, unicode, parentheses, etc.** intact. So a repo like `/home/u/my project` became the name `home-u-my project`: indexing created `home-u-my project.db` and `list_projects` showed it, but `resolve_store` rejected the space and returned `project not found`.

Fixed in commit 5595ce4: `cbm_project_name_from_path` now maps **every** character the validator would reject to `-` (plus collapses `..` and strips leading dots), so a derived name always round-trips through `resolve_store`. Regression test `project_name_always_validator_safe_issue349` asserts the invariant for paths with spaces, `@`, `+`, `..`, leading dots, and non-ASCII bytes.

[...]

PR #624, opened 2026-06-25
fix: preserve non-ASCII (CJK) path segments in auto-generated project name

### LINKED ISSUE #571: Project name strips non-ASCII (CJK) characters from path, resulting in truncated/unrecognizable names
## Bug Description

## Steps to Reproduce

## Expected Behavior

## Actual Behavior

## Impact

## Environment

- codebase-memory-mcp v0.8.1
- macOS (Apple Silicon)

mvanhorn:
## What does this PR do?

Fixes #571: `cbm_project_name_from_path()` derived an auto project slug by mapping
every byte outside `[A-Za-z0-9._-]` to `-` and then collapsing dashes. For a path
such as `/Users/yunxin/Desktop/开发/后端`, every UTF-8 byte of the CJK segments is
unsafe, so each segment collapsed to a single `-` that was then trimmed away,
yielding `Users-yunxin-Desktop`. Two different non-ASCII directories under the same
ASCII parent collapsed to the same name, so the identifying information was
silently lost and distinct repos collided. Users with non-Latin directory names
saw indexing create a DB under a truncated, unrecognizable name.

This change percent-encodes non-ASCII bytes instead of discarding them, so
non-ASCII paths keep a distinct, recoverable name. It also keeps the slug
generator in agreement with `cbm_validate_project_name()` -- the same invariant
that motivated the `#349` space fix (a name that indexes but `resolve_store`
later rejects reports project-not-found).

Changes:

- `src/pipeline/fqn.c` -- `cbm_project_name_from_path()` now percent-encodes any
  byte `>= 0x80` as uppercase `%HH` into a new, larger buffer, while ASCII
  separators / unsafe bytes still map to `-`. ASCII slugs are byte-identical to
  before (e.g. `/home/u/my project` -> `home-u-my-project`). The existing
  dash/dot collapse and leading/trailing trim are preserved. Because the slug is
  later used as `<cache>/<name>.db`, long slugs are bounded to 200 bytes with a
  deterministic FNV-1a hash suffix so distinct long paths still produce distinct
  names and stay within the OS filename-component limit. Includes overflow guards
  and correct memory management.
- `src/foundation/str_util.c` -- `cbm_validate_project_name()` additionally
  accepts `%`. It remains filesystem-safe; the existing `..`, `/`, `\`, and
  leading-`.` rejections are unchanged.
- `tests/test_fqn.c` -- new coverage for CJK percent-encoding, distinctness, the
  long-path bound, the `%`-accepting validator, and an unchanged-ASCII
  regression.

Example: `/Users/yunxin/Desktop/开发/后端` now yields
`Users-yunxin-Desktop-%E5%BC%80%E5%8F%91-%E5%90%8E%E7%AB%AF` (distinct, valid)
instead of `Users-yunxin-Desktop`.

## Checklist

- [x] Every commit is signed off (`git commit -s`) -- required, CI rejects
- [x] Tests pass locally (`make -f Makefile.cbm test`) -- 5689 passed, 0 failed
- [x] Lint passes (`make -f Makefile.cbm lint-ci`) -- clang-format clean on
      touched files

Fixes #571

### DIFF
--- a/src/foundation/str_util.c
+++ b/src/foundation/str_util.c
@@ -282,10 +282,10 @@ bool cbm_validate_project_name(const char *name) {
     /* Reject leading dot (hidden files / relative refs) */
     if (name[0] == '.')
         return false;
-    /* Allow only alphanumeric, dash, underscore, dot */
+    /* Allow only alphanumeric, dash, underscore, dot, percent */
     for (const char *p = name; *p; p++) {
         if (!(((*p >= 'a') && (*p <= 'z')) || ((*p >= 'A') && (*p <= 'Z')) ||
-              ((*p >= '0') && (*p <= '9')) || *p == '-' || *p == '_' || *p == '.')) {
+              ((*p >= '0') && (*p <= '9')) || *p == '-' || *p == '_' || *p == '.' || *p == '%')) {
             return false;
         }
     }
--- a/src/pipeline/fqn.c
+++ b/src/pipeline/fqn.c
@@ -11,12 +11,22 @@
 #include <stdbool.h>
 #include <stddef.h> // NULL
 #include <stdio.h>
+#include <stdint.h> // SIZE_MAX
 #include <stdlib.h>
 #include <string.h> // strdup
 
 /* Maximum path segments in a FQN (CBM_SZ_256 slots total, -2 for project + name) */
 #define FQN_MAX_PATH_SEGS 254
 #define FQN_MAX_DIR_SEGS 255
+
+/* Cap the derived project name well under the common 255-byte filename-component
+ * limit (the name is later used as "<cache_dir>/<name>.db"). Percent-encoding
+ * non-ASCII bytes can triple the byte length, so a deep CJK path could otherwise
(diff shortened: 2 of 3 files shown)


DeusData:

DeusData:
Thank you @mvanhorn — closing this in favor of a distilled follow-up.

The core of #571 (preserving CJK path segments by hex-encoding non-ASCII bytes) already landed on `main` in `3f754cf`, which is why this PR now conflicts. But your PR caught a genuinely-missing piece that fix lacked: **without a length cap, a deep CJK / non-ASCII path hex-encodes past the filesystem's 255-byte filename-component limit, making the `.db` un-openable (ENAMETOOLONG).**

I've distilled exactly that improvement — a name-length cap with a hash suffix, added **on top of** the existing hex scheme so there's no project-name drift for already-indexed projects — into **#745** (`db20e53`), **co-authored with you**. It's guarded by a reproduce-first test (`project_name_length_capped_issue624`), and the full suite is green.

Really appreciate the careful catch on the filename length limit. 🙏

PR #625, opened 2026-06-25
Gate git watcher behind auto_watch config (contained indexing)

Andy11-cpu:
## Summary

Split from the original bundled PR per review feedback. This PR covers **auto_watch** only; sibling PRs handle the other changes:

- #752 — RAM budget tiering (25% / 35% / 50%)
- #753 — Install hooks opt-in (`--hooks`)
- #754 — Project index deduplication by git/path identity (**merge before this PR**)

### This PR

- Add `auto_watch` config (default **false**)
- Do not register the git watcher on MCP `initialize` unless `auto_watch` is enabled
- Re-index only via `index_repository`, or when `auto_watch` is turned on after a manual index
- `index_repository` registers the watcher only when `auto_watch` is enabled

## Motivation

On multi-project setups, registering the git watcher on every MCP connect caused background re-index storms even when `auto_index` was false.

## Test plan

- [x] `cli_config_get_bool` covers `auto_watch` default false
- [x] `mcp_auto_watch_disabled_skips_watcher_on_connect` in `test_mcp.c`
- [ ] `make -f Makefile.cbm test` passes (CI)
- [ ] MCP connect with existing index: `watcher.skip` (not `watcher.watch`) when `auto_watch=false`

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -175,6 +175,7 @@ GRAPH_BUFFER_SRCS = src/graph_buffer/graph_buffer.c
 # Pipeline module (new)
 PIPELINE_SRCS = \
     src/pipeline/fqn.c \
+    src/pipeline/project_resolve.c \
     src/pipeline/path_alias.c \
     src/pipeline/registry.c \
     src/pipeline/pipeline.c \
@@ -332,7 +333,7 @@ TEST_DISCOVER_SRCS = \
 
 TEST_GRAPH_BUFFER_SRCS = tests/test_graph_buffer.c
 
-TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c
+TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_project_resolve.c
 
 TEST_WATCHER_SRCS = tests/test_watcher.c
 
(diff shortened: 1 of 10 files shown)


DeusData:

DeusData:
Thanks. This bundles several independent behavior changes: `auto_watch`, project deduplication, RAM tiering, and hook opt-in defaults. Please split these or link focused tracking issues for each. Also please remove generated/co-author/session attribution trailers from non-merge commits. Because this changes MCP initialization/watchers and install defaults, each behavior needs focused tests and CI evidence.

Andy11-cpu:
Thanks for the review — addressed all three points:

**1. Split into focused PRs**
- #752 — RAM budget tiering (25% / 35% / 50%)
- #753 — Install hooks opt-in (`--hooks`)
- #754 — Project index deduplication by git/path identity
- #625 (this PR, updated) — `auto_watch` gate only; stacks on #754

**2. Removed co-author / session attribution trailers**
Non-merge commits now have a single `Signed-off-by` line only (no `Co-authored-by: Cursor`).

**3. Focused tests per behavior**
- RAM tiers: `mem_ram_fraction_*` in `test_mem.c` (#752)
- Hooks opt-in: `cli_install_plan_hooks_opt_in_default` in `test_cli.c` (#753)
- Project dedup: `test_project_resolve.c` (#754)
- `auto_watch`: `mcp_auto_watch_disabled_skips_watcher_on_connect` + config default test (#625)

CI should run on each PR independently. Happy to adjust merge order if you prefer a different stack.

DeusData:
Thank you — the auto_watch gate was the right idea for multi-project containment, and your instinct to gate the watcher registration was sound. We carried the auto_watch half over the line as **54efcef7** (PR #849) with you credited as co-author, defaulting the key to **true** (opt-out) so every existing user keeps background auto-sync exactly as before — only users who explicitly want containment flip it off. We deliberately left the bundled project-dedup commit out of this one: it's a substantial feature that deserves its own focused review, and it stays with #754 where it can get that. Closing this in favor of the distilled auto_watch gate — thanks again, and please do keep #754 moving on its own!

PR #637, opened 2026-06-26
fix: use _wfopen on Windows to support non-ASCII (CJK/U...) file paths

### LINKED ISSUE #636: Windows: tree-sitter definitions pass fails on ALL files when repo_path contains non-ASCII (CJK / Cyrillic / Arabic / accented Latin / etc.) characters
## Bug Description

On Windows, when the `repo_path` passed to `index_repository` contains Chinese (CJK) characters in directory or file names, the **definitions pass** (tree-sitter AST parsing) fails on **every file** — `defs=0`, `calls=0`, `imports=0`, `errors=N` for all files. The resulting knowledge graph contains only File/Folder/Project nodes with no Function, Class, Method, or Variable nodes, making the index effectively useless for code intelligence.

This is **more severe** than issue #571 (project name truncation for CJK paths) — that issue is cosmetic, while this one prevents all code-level analysis entirely.

## Steps to Reproduce

1. On Windows, create or index a project whose path contains Chinese characters, e.g.:
   ```
   F:/项目资产-ProjectAssets/智能体/智能体原型/
   ```
2. Run:
   ```
   codebase-memory-mcp cli index_repository '{"repo_path":"F:/项目资产-ProjectAssets/智能体/智能体原型"}'
   ```
3. Observe the pipeline log — the definitions pass shows:
   ```
   pass=definitions defs=0 calls=0 imports=0 errors=41
   ```
4. `get_architecture` returns only File/Folder nodes, zero Function/Class/Method nodes

## Expected Behavior

Files at paths with Chinese characters should be parsed by tree-sitter identically to ASCII-only paths. A 41-file TypeScript/JS/CSS project should produce, e.g., `defs=200+` with `errors=0`.

## Actual Behavior (Verbose Pipeline Log)

```
pass=discover files=41
pass=structure nodes=60 edges=58           ← File/Folder tree works fine
pass=definitions defs=0 calls=0 imports=0 errors=41  ← ALL files fail
pass=calls total=0 resolved=0 unresolved=0 errors=41
pass=semantic inherits=0 decorates=0 implements=0 errors=41
```

## Root Cause Verified

Identical file content, copied to an ASCII-only path, indexes **correctly**:

| Test | Path | defs | calls | errors |
|------|------|------|-------|--------|
| ASCII path | `C:/tmp/core-test.js` | 30 | 103 | **0** ✅ |
| Chinese path | `F:/...智能体.../core.js` | 0 | 0 | **41** ❌ |

The 417-line JS file is byte-identical. The only variable is whether the parent path contains CJK characters. Languages tested: JavaScript, TypeScript (.ts/.tsx), Python, CSS — all fail identically when the path has CJK characters.

## Environment

- **OS:** Windows 10 (64-bit)
- **codebase-memory-mcp version:** v0.8.1
- **Filesystem:** NTFS (UTF-8)
- **Shell tested:** git-bash (MSYS2) — same result via direct Python subprocess call

## Suspected Cause

Likely a Windows-specific encoding issue. The structure pass (directory enumeration via `FindFirstFileW`/`FindNextFileW`) works because it natively uses wide-char UTF-16 paths. But the definitions pass probably reads file paths as plain `char*` and passes them to tree-sitter via `fopen` or similar, which on Windows uses the ANSI codepage (e.g., GBK on zh-CN systems) rather than UTF-8. This causes tree-sitter to receive garbled paths or fail to open file contents for AST parsing.

On Linux/macOS, `char*` paths in UTF-8 work natively because the filesystem encoding is UTF-8 — this explains why similar issues (#571) were reported but the definitions-pass failure was not caught.

maxniu1:
## Summary

On Windows, `fopen()` uses the active ANSI codepage (e.g. GBK on zh-CN, Shift-JIS on ja-JP, CP1251 on ru-RU) rather than UTF-8. When a repository path contains characters outside ASCII-7 — **any non-English language**: Chinese, Japanese, Korean, Cyrillic, Arabic, accented Latin (é/ü/ñ), Thai, Hebrew, Devanagari, etc. — every file's tree-sitter definitions pass fails (`defs=0`, `errors=N` for all files). The resulting knowledge graph contains only File/Folder nodes with zero code intelligence.

The existing directory-enumeration code in `compat_fs.c` already handles this correctly via `FindFirstFileW`/`_wmkdir`/`_wunlink` with wide-character paths. This PR applies the same pattern to file **reading** for tree-sitter parsing.

## Changes

| File | Change |
|------|--------|
| `src/foundation/compat_fs.h` | Declare `cbm_fopen()` |
| `src/foundation/compat_fs.c` | Implement `cbm_fopen()`: `_wfopen` on Windows, `fopen` on POSIX |
| `src/pipeline/pass_parallel.c` | Replace `fopen` → `cbm_fopen` |
| `src/pipeline/pass_calls.c` | Replace `fopen` → `cbm_fopen` |
| `src/pipeline/pass_definitions.c` | Replace `fopen` → `cbm_fopen` |

All three pipeline files had a duplicated `read_file()` static function. Each now calls `cbm_fopen()` instead of raw `fopen()`.

### Before (v0.8.1)
```
pass=definitions defs=0 calls=0 imports=0 errors=41   ← ALL files fail
```

### After (this PR)
```
pass=definitions defs=215 calls=259 imports=18 errors=0  ← all pass
```

Tested on a 41-file TypeScript/JS/Python/CSS project with CJK directory names (`F:/.../智能体/智能体原型/`). Full index: 245 nodes, 347 edges, 197ms.

## Related

- Fixes #636 (this specific bug)
- Related to #571 (project name truncation for CJK paths — cosmetic, already fixed)
- This is the **functional** fix that makes the tool work for non-English Windows users

### DIFF
--- a/src/foundation/compat_fs.c
+++ b/src/foundation/compat_fs.c
@@ -130,6 +130,22 @@ int cbm_pclose(FILE *f) {
     return _pclose(f);
 }
 
+/* Open a file with wide-char _wfopen for UTF-8 path support on Windows. */
+FILE *cbm_fopen(const char *path, const char *mode) {
+    wchar_t *wpath = cbm_utf8_to_wide(path);
+    if (!wpath) {
+        return NULL;
+    }
+    wchar_t wmode[8] = {0};
+    if (mbstowcs(wmode, mode, 7) == (size_t)-1) {
+        free(wpath);
+        return NULL;
+    }
+    FILE *f = _wfopen(wpath, wmode);
+    free(wpath);
+    return f;
+}
+
 bool cbm_mkdir_p(const char *path, int mode) {
     (void)mode;
     wchar_t *wpath = cbm_utf8_to_wide(path);
@@ -260,6 +276,10 @@ FILE *cbm_popen(const char *cmd, const char *mode) {
 
 int cbm_pclose(FILE *f) {
     return pclose(f);
+}
+
+FILE *cbm_fopen(const char *path, const char *mode) {
+    return fopen(path, mode);
 }
 
 bool cbm_mkdir_p(const char *path, int mode) {
--- a/src/foundation/compat_fs.h
+++ b/src/foundation/compat_fs.h
@@ -56,4 +56,8 @@ int cbm_rmdir(const char *path);
  * POSIX: fork() + execvp(). Windows: _spawnvp(). */
 int cbm_exec_no_shell(const char *const *argv);
 
+/* Open a file for reading. On Windows, uses wide-char _wfopen to support
+ * non-ASCII (UTF-8) paths. On POSIX, delegates to fopen(). */
+FILE *cbm_fopen(const char *path, const char *mode);
+
 #endif /* CBM_COMPAT_FS_H */
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -19,6 +19,7 @@ enum { PC_RING = 4, PC_RING_MASK = 3, PC_SIG_SCAN = 15, PC_REGEX_GRP = 2 };
 #include "graph_buffer/graph_buffer.h"
 #include "foundation/log.h"
 #include "foundation/compat.h"
+#include "foundation/compat_fs.h"
 #include "foundation/str_util.h"
 #include "cbm.h"
 #include "service_patterns.h"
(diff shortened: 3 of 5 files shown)


DeusData:

DeusData:
Thanks, this is a useful Windows path fix. Before review can continue, please add a DCO `Signed-off-by` trailer to the non-merge commit and include the Windows test command/output you used for the non-ASCII path case.

DeusData:
Thank you @maxniu1 — you correctly root-caused #636: on Windows, non-ASCII (CJK) paths go through the ANSI `fopen`, so `_wfopen` (UTF-8→UTF-16) is the right fix.

We're landing the fix via #703, which applies the same approach across **all** the file-read passes. This PR converts 3 of them, which leaves `pass_semantic` still going through ANSI `fopen` — so semantic extraction would keep failing on CJK paths (your own issue log shows `pass=semantic errors=41`). Rather than expand this PR to match #703, we're taking #703 and closing this as a duplicate.

Full credit for the diagnosis — the fix ships in the next release. 🙏

Issue #627, opened 2026-06-25
Crash when calling query_graph


zbynekwinkler:
### Version

codebase-memory-mcp 0.8.1

### Platform

Windows (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

After indexing I tasked my agent to find dead code. It started calling the mcp and it has crashed almost on the first useful query.

### Reproduction

It is a private codebase. I had to check the box that it is not. I am doing so in the hopes that it might help anyway.

`codebase-memory-mcp cli query_graph '{"project": "YYY", "query": "MATCH (f:Function) WHERE NOT f.file_path CONTAINS ''ext'' AND NOT f.file_path CONTAINS ''Tests'' AND NOT f.file_path CONTAINS ''examples'' AND NOT f.name = ''main'' OPTIONAL MATCH (c)-[:CALLS]->(f) WITH f, c WHERE c IS NULL RETURN f.name, f.qualified_name, f.file_path, f.start_line ORDER BY f.file_path LIMIT 50"}'`

no usefull output, just a crash dump appearing in `AppData\Local\CrashDumps\`

### Logs

```text
PS C:\Users\XXX\AppData> codebase-memory-mcp cli query_graph '{"project": "YYY", "query": "MATCH (f:Function) WHERE NOT f.file_path CONTAINS ''ext'' AND NOT f.file_path CONTAINS ''Tests'' AND NOT f.file_path CONTAINS ''examples'' AND NOT f.name = ''main'' OPTIONAL MATCH (c)-[:CALLS]->(f) WITH f, c WHERE c IS NULL RETURN f.name, f.qualified_name, f.file_path, f.start_line ORDER BY f.file_path LIMIT 50"}'
level=info msg=mem.init budget_mb=32596 total_ram_mb=65192
PS C:\Users\XXX\AppData>
```

### Project scale (if relevant)

Nodes: 29,361 Edges: 59,940

### Confirmations

- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

zbynekwinkler:
I think I have it crashing also under wsl.

<img width="927" height="271" alt="Image" src="https://github.com/user-attachments/assets/2899d5a4-cabe-41d2-8ed0-ee943dd35eb0" />

zbynekwinkler:
That is actually a local build of current main.

[...]

Issue #618, opened 2026-06-25
hook_augment PreToolUse augmenter is a structural no-op on Windows (POSIX-only path guards)


catsmonster:
### Summary

`codebase-memory-mcp hook-augment` (the non-blocking Claude Code **PreToolUse** Grep/Glob augmenter) is a **structural no-op on Windows**. It runs, inits, and exits 0 with no `additionalContext` for *every* input, because the path guards in `src/cli/hook_augment.c` require POSIX-style `/`-prefixed absolute paths, which a Windows drive-letter `cwd` (`C:\...` / `C:/...`) can never satisfy.

Net effect: on Windows the Grep/Glob graph augmentation never fires, even for an exactly-indexed symbol. (The "never block a tool" guarantee is intact — it just never contributes anything.)

- **Version:** `codebase-memory-mcp 0.8.1`
- **Platform:** Windows (x64)
- **Install channel:** install.ps1
- **Binary variant:** standard

### What happened, and what did you expect?

**Expected:** running a `Grep` on an indexed symbol fires the PreToolUse hook and injects matching graph symbols as `additionalContext`.

**Actual:** `hook-augment` emits 0 bytes of stdout for any Grep/Glob payload on Windows; no context is ever added.

### Reproduction

Platform-level, no proprietary code needed — reproduces with any indexed project on Windows.

1. On Windows, index any repo (e.g. a clone of a public repo) so it appears in `list_projects` with a `root_path` like `C:/Users/me/proj`.
2. Invoke the augmenter exactly as the installed PreToolUse hook does, with a realistic Claude Code payload:

   ```
   echo '{"hook_event_name":"PreToolUse","tool_name":"Grep","cwd":"C:/Users/me/proj","tool_input":{"pattern":"someIndexedSymbol"}}' \
     | codebase-memory-mcp hook-augment
   ```

3. **Result:** empty stdout (no `hookSpecificOutput`). **Expected:** a `{"hookSpecificOutput":{"hookEventName":"PreToolUse","additionalContext":"[codebase-memory] N graph symbol(s) match ..."}}` payload.

`search_graph` for the same project/token returns correct hits, confirming the index and project name are fine — only `hook-augment`'s path handling fails.

### Root cause

`src/cli/hook_augment.c` — two POSIX-only guards:

1. **`cbm_cmd_hook_augment`, the `_WIN32` branch (~L330):**
   ```c
[...]

Issue #548, opened 2026-06-21
D:\ drive and custom path cannot be selected in server UI


ethernidee:
### Version

0.8.1

### Platform

Windows (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

ui

### What happened, and what did you expect?

The UI project directory selection dialog:

1) Does not allow to write custom path.
2) Does not show all available drives on Windows, when selecting "/" (root).

Thus it's not possible to select project directory in UI mode at all.

### Reproduction

Try to select D:\ drive path on server UI page on Windows or try to enter arbitrary path in GUI directory selection dialog.

### Logs

```text

```

### Confirmations

- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

DeusData:
Hey, thx for reporting. Will check ASAP

DeusData:
This issue now has a first-class, durable **reproduction** in the cumulative bug-reproduction suite landing in #667 — a test that pins the exact failing behaviour, so it's tracked and guarded against silent regression. Where #667 also lands the root-cause fix, that's called out in the PR; otherwise the reproduction makes the eventual fix straightforward and verifiable.

The reproduction goes live once #667 merges. (Posted from the release branch's QA pass.)

DeusData:
Fixed on main (drive/root selection now served via the roots API — GetLogicalDrives + append_roots_json in http_server.c) and now covered by a green Windows regression guard, tests/windows/test_ui_drive_listing.py (added in #639, passing on Windows CI). Closing.

PR #639, opened 2026-06-27
test(windows): add red regressions for platform-specific failures

### LINKED ISSUE #700: Bug: File parsing fails silently on Windows when project path contains accented/non-ASCII characters
### Version

0.8.1

### Platform

Windows (x64)

### Install channel

### Binary variant

ui

When codebase-memory-mcp indexes a repository on Windows, if the absolute path to the repository contains accented characters (e.g., C:\Users\...\Proyectos De César\...), the indexer correctly identifies and creates File nodes for all files in the project, but fails silently to extract any functions, classes, or AST-based nodes via tree-sitter.

The files appear in the graph (e.g. searching for label: File returns them), but searching for label: Function returns 0 results because the internal file reading/parsing phase fails.

codebase-memory-mcp should be able to open and parse the contents of files even if their absolute paths contain non-ASCII characters on Windows.

The files are registered in the index as File nodes (likely via directory traversal), but their contents are never successfully parsed by tree-sitter, leaving the knowledge graph completely devoid of semantic structural nodes (Functions, Classes, Methods, etc.) without throwing an explicit fatal error during indexing.

Because codebase-memory-mcp is a static C binary, this is likely an encoding issue with how Windows standard I/O handles file paths. Standard functions like fopen on Windows expect ANSI paths by default unless the program explicitly uses wide-character APIs (e.g., _wfopen) or the UTF-8 code page is enforced. When the C backend attempts to open the file using its UTF-8 path string, Windows fails to find the file, causing the parser to silently skip it.

### Reproduction

1.- On a Windows machine, create a directory with an accented character, e.g., mkdir C:\test-césar
2.- Create a dummy Python file inside it: echo "def hello(): pass" > C:\test-césar\main.py
3.- Run the indexer on this path via the MCP tool or CLI.
4.- Query the graph for files: MATCH (n:File) RETURN n (Returns the main.py file).
5.- Query the graph for functions: MATCH (n:Function) RETURN n (Returns 0 results).
6.- Repeat the steps in a path without accents (e.g., C:\test-cesar), and the function hello is correctly extracted.

### Logs

```text

```

```text

```

261 nodes / 249 edges / 232 files 

### Confirmations

Flipper1994:
## Summary

Windows-only integration tests that drive the product surface (a real
`codebase-memory-mcp.exe` over JSON-RPC stdio / CLI / HTTP UI, real SQLite DB),
found during a native-Windows red-test campaign. Deterministic, standard-library
Python 3 only, and pass on Linux/macOS.

**Updated after review + rebased onto current `main`.** Three of the four
originally-red findings were fixed upstream since the branch was cut at `b075f05`,
so they are now **green regression guards**; the fourth is a genuine, still-open
Windows bug kept as a **known red**. This PR still contains **no production fixes**.

| Test | Issue | Status |
|---|---|---|
| `test_non_ascii_path.py` | #636 / #357 | **GREEN guard** — fixed by #700 (`cbm_fopen` routing in the pass readers) |
| `test_hook_augment.py` | #618 | **GREEN guard** — fixed by #619 (`cbm_is_walkable_abs_path` accepts `X:/`) |
| `test_ui_drive_listing.py` | #548 | **GREEN guard** — fixed (drives via the `roots` field); test rewritten |
| `test_cli_non_ascii_arg.py` | #423 / #20 | **RED (open)** — `main()` is still narrow-argv, no wide command line |

Re-verified on this Windows host (11 Pro 26200, drives `C:`/`D:`/`E:`): the three
guards pass green, the keeper stays red.

## What changed since the first review

- **Rebased** onto current `main` (the only conflict was `.gitignore`; resolved).
- **`test_non_ascii_path.py` / `test_hook_augment.py`** kept as-is functionally
  (they already pass on fixed code) and re-framed as green guards for #700 / #619.
- **`test_ui_drive_listing.py` rewritten.** The #548 fix intentionally exposes
  drives through a new `roots` array, not the `dirs` array the original test
  asserted — so the old assertion would stay red against fixed code. It now asserts
  every fixed drive appears in `roots` and is browsable (`GET /api/browse?path=X:/`).
  Meaningful on a single-drive machine too, so it also gates single-drive CI.
- **`test_cli_non_ascii_arg.py`** unchanged — still genuinely red (`main()` narrow argv).
- **`scripts/test-windows.ps1`** now splits **green guards** (must stay green, gate
  CI) from **opt-in known-reds** (expected red, never gate CI); adds `-GuardsOnly`.
  Indexing runs in-process (`CBM_INDEX_SUPERVISOR=0`) so a guard reflects the
  path/hook/drive fix under test, not the orthogonal index-supervisor worker path.
- **CI wiring** (`.github/workflows/_test.yml`): new `test-windows-guards` job builds
  the product+UI binary (`scripts/build.sh --with-ui`) and runs the guards with
  `-GuardsOnly`, so #700 / #619 / #548 stay enforced on Windows. This is separate
  from the existing `test-windows` sanitizer C-suite job.
- **`RED_TEST_ANALYSIS.md`** refreshed to record the landed fixes and status.

### `test_non_ascii_path.py` — #636 / #357 (fixed by #700)
Byte-identical TypeScript fixtures under non-ASCII parent paths (Latin-1 accents,
Cyrillic, CJK, Greek) must extract the same definitions as the ASCII baseline.
Before #700, native Windows extracted only `File`/`Folder` nodes for non-ASCII
paths (the pass readers `fopen()`'d a UTF-8 path in the ANSI code page); #700
routed those reads through `cbm_fopen` → `_wfopen`. Guard fails if that regresses.

### `test_hook_augment.py` — #618 (fixed by #619)
`hook-augment` (the PreToolUse Grep/Glob augmenter) must emit `hookSpecificOutput`
for a drive-letter `cwd`. Before #619 it bailed on `cwd[0] != '/'`; #619 added
`cbm_is_walkable_abs_path` (accepts `X:/`). Guard fails if that regresses.

### `test_ui_drive_listing.py` — #548 (fixed)
The UI directory picker must let the user reach every logical drive. `handle_browse`
now appends `"roots":["C:/","D:/",...]` (via `GetLogicalDrives`) to every
`/api/browse` response. Needs a UI build (`cbm-with-ui`), which the CI job builds.

### `test_cli_non_ascii_arg.py` — #423 / #20
`cli index_repository` rejects a non-ASCII `repo_path` ("repo_path is required")
because `main()` (`src/main.c`) is `int main(int argc, char **argv)` with no
`wmain` / `GetCommandLineW`, so on Windows argv arrives in the ANSI code page.
Opt-in only; never gates CI. Real MCP clients pass `repo_path` inside byte-clean
JSON-RPC over stdio, so this affects the documented `cli` entrypoint and the
hook/install flows that shell out to it.

## How to run

```powershell
pwsh -File scripts/test-windows.ps1            # build (with UI) + guards + known-reds
pwsh -File scripts/test-windows.ps1 -GuardsOnly # only the green guards (the CI gate)
```

Each test exits `0` (green), `1` (red), or `2` (precondition/setup). Full
environment, commands, ruled-out seed areas, and fix references are in
`tests/windows/RED_TEST_ANALYSIS.md`.

## Runner scope / isolation

- The three green guards are wired into Windows CI (the `test-windows-guards` job)
  so the landed fixes stay enforced; the opt-in known-red is never run in CI.
- No production changes — the diff is limited to `tests/windows/`,
  `scripts/test-windows.ps1`, and the CI job in `.github/workflows/_test.yml`.
- The harness launches only the locally built `codebase-memory-mcp.exe` (JSON-RPC
  stdio / CLI) and, for the drive-picker guard, the project's own embedded HTTP UI
  bound to loopback; it makes no network calls and writes only to temp dirs.

## Related issues

Guards fixes: #700 (→ #636/#357), #619 (→ #618), #548. Reproduces (open): #423,
#20. Also documents ruled-out seed areas (#513/#530.x/#635, #422/#348, #274,
#371/#137, #627, #272, #511, #331, #185/#406, #227/#367) and cross-platform
out-of-scope items (#571, #530.2/.4/.5, #394 umbrella) in the analysis.

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -124,3 +124,43 @@ jobs:
         run: scripts/test.sh CC=clang CXX=clang++ ${{ matrix.os == 'windows-11-arm' && 'SANITIZE=' || '' }}
         env:
           CBM_SKIP_PERF: ${{ inputs.skip_perf && '1' || '' }}
+
+  # Windows product-surface regression guards. Distinct from test-windows above
+  # (the sanitizer C suite): these drive a real product binary + embedded HTTP UI
+  # over stdio / CLI / HTTP and fail if a Windows bug already fixed on main comes
+  # back -- non-ASCII repo paths dropping definitions (#636/#357, fixed by #700),
+  # the PreToolUse hook augmenter no-op on drive-letter cwd (#618, fixed by #619),
+  # and the UI directory picker not enumerating drives (#548, roots field). The
+  # still-open narrow-argv repro (#423/#20) is opt-in and excluded via -GuardsOnly.
+  test-windows-guards:
+    runs-on: windows-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - uses: msys2/setup-msys2@66cd2cce69caa17b53920067426061ca1de3a884 # v2
+        with:
+          msystem: CLANG64
+          path-type: inherit
+          install: >-
+            mingw-w64-clang-x86_64-clang
+            mingw-w64-clang-x86_64-zlib
+            make
+            git
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: "22"
+
+      - name: Build product binary with embedded UI
+        shell: msys2 {0}
+        # --with-ui builds the frontend (npm) and embeds it, so the drive-picker
+        # guard's HTTP UI is available. Functional gate only (no sanitizers).
+        run: scripts/build.sh --with-ui CC=clang CXX=clang++
+
+      - name: Windows regression guards (#636/#357, #618, #548)
+        shell: pwsh
+        # -GuardsOnly runs the three green guards and gates on them; the runner
+        # runs indexing in-process (CBM_INDEX_SUPERVISOR=0) so a guard reflects the
+        # path/hook/drive fix under test, not the orthogonal index-worker spawn.
+        run: ./scripts/test-windows.ps1 -GuardsOnly -Binary build/c/codebase-memory-mcp.exe
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,7 @@ soak-results/
 
 # Local npm cache
 graph-ui/.npm-cache-local/
+
+# Python bytecode from tests/windows/ harness
+__pycache__/
+*.pyc
--- a/scripts/test-windows.ps1
+++ b/scripts/test-windows.ps1
@@ -1,22 +1,50 @@
 <#
 .SYNOPSIS
-    Run the native-Windows red-test suite for codebase-memory-mcp.
+    Run the native-Windows product-surface test suite for codebase-memory-mcp.
 
 .DESCRIPTION
-    Builds the production binary (build/c/codebase-memory-mcp.exe) if it is not
-    already present, then runs the deterministic Windows red tests under
-    tests/windows/. These tests reproduce platform-specific failures at the
-    product surface (real MCP process, real stdio, real SQLite DB).
+    Builds the product binary (build/c/codebase-memory-mcp.exe) if it is not
+    already present, then runs the deterministic Windows integration tests under
+    tests/windows/ against a real codebase-memory-mcp.exe (real stdio / CLI /
+    HTTP UI, real SQLite DB).
 
-    The unit/invariant C suite is built and run via Makefile.cbm. On native
-    Windows the MinGW/LLVM toolchain ships no libasan/libubsan, so the sanitizer
-    flags must be disabled for the local build (SANITIZE=). Where the toolchain
-    *does* provide AddressSanitizer/UBSan (Linux containers, WSL), prefer
-    scripts/test.sh which keeps the sanitizers on.
+    Two categories of test:
+
+      GUARDS      - regression guards for Windows bugs already fixed on main.
+                    They must stay GREEN (exit 0); a RED (exit 1) means the fix
+                    regressed and fails this runner.
+                      * test_non_ascii_path.py    guards #636/#357 (fixed by #700)
+                      * test_hook_augment.py      guards #618      (fixed by #619)
+                      * test_ui_drive_listing.py  guards #548      (roots field)
+
+      KNOWN REDS  - genuine, still-open Windows bugs reproduced at the product
+                    surface. They are EXPECTED to be RED (exit 1) and are opt-in
+                    (never gate CI). If one turns GREEN the underlying bug was
+                    fixed and it should be promoted to a guard.
+                      * test_cli_non_ascii_arg.py reproduces #423/#20 (narrow
+                        argv main() - no wide command line)
+
+    Determinism: indexing runs in-process (CBM_INDEX_SUPERVISOR=0). These tests
+    exercise path / hook / drive handling, not the index-supervisor subprocess
+    path; the pass-level readers (#700's cbm_fopen routing) run in-process either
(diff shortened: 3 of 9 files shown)


DeusData:

DeusData:
Thanks for putting these Windows repros together. Before this can proceed, please remove generated/session attribution from the PR text and make sure the red-test runner remains outside the normal green test path. Since the harness launches local binaries and an HTTP server, maintainers will review the runner scope carefully.

Flipper1994:
Thanks for the feedback — both points are addressed:

- **Attribution:** removed the generated/session footer from the PR description.
- **Runner scope:** added a "Runner scope / isolation" section to the description spelling out that the runner is opt-in (`scripts/test-windows.ps1` only), is referenced by no `Makefile` / `make test` target / `.github/` workflow, ships no production changes, and only launches the locally built binary and the project's own loopback HTTP UI (no network, temp-only writes).

Happy to adjust the runner further if you'd like it gated or relocated differently.

DeusData:
Thank you for this — the mcp_stdio.py harness fills a real gap (we've wanted a Windows product-binary harness for a while), and the tests are deterministic and well-engineered. The catch: the branch was cut at b075f05 and main has moved — 3 of the 4 reds are stale: the hook-augment case was fixed by #619 (issue #618 is closed), the non-ASCII pass readers were fixed by #700 (cbm_fopen everywhere), and the drive-listing test asserts drives appear in the dirs array while the landed fix intentionally exposes them via the roots field — so it stays red against fixed code, which is as misleading as a false green. The keeper is test_cli_non_ascii_arg.py: main() is still narrow argv with no wide entry point, so that red is genuine. Could you: rebase + re-run on Windows; convert the two now-green tests into permanent green guards (ideally wired into the windows CI job so #619/#700 stay enforced); rewrite the drive-listing test against the roots-field/user-level invariant; and refresh or drop the analysis doc (it snapshots claims that are no longer true of main)? With that, this becomes a great addition. Thanks!

Flipper1994:
Thanks — this was exactly right, and I've reworked the PR accordingly.

Rebased onto current `main` (only `.gitignore` conflicted). Re-ran on Windows (11 Pro, drives C:/D:/E:) and confirmed your read on all four:

- **`test_non_ascii_path.py` (#636/#357 → #700)** and **`test_hook_augment.py` (#618 → #619)** are green on main now. Kept both as green regression guards and wired them into a new `test-windows-guards` CI job so #700/#619 stay enforced.
- **`test_ui_drive_listing.py` (#548)** rewritten. You were right that it asserted drives in `dirs` while the fix exposes them via `roots` — the old assertion stayed red against fixed code. It now asserts every fixed drive is in `roots` and is browsable. Green on main (`roots=['C:/','D:/','E:/']`), and meaningful on a single-drive box too, so it also gates the single-drive CI runner.
- **`test_cli_non_ascii_arg.py` (#423/#20)** is the genuine keeper — `main()` is still narrow-argv with no wide command line. Opt-in, never gates CI.

One implementation note: the runner sets `CBM_INDEX_SUPERVISOR=0` so the guards index in-process — the passes under test (#700's `cbm_fopen`) run in-process either way, and it keeps the guards independent of the index-supervisor's separate worker process. Happy to switch to the supervised path if you'd prefer.

Refreshed `RED_TEST_ANALYSIS.md` to record the landed fixes. Thanks again for the careful review.

DeusData:
Thank you — and thanks especially for the rebase: converting the three now-fixed reds into green regression guards (non-ASCII paths via #703, hook-augment via #619, drive-listing via the roots API) while keeping the CLI-argv case as an opt-in known-red is exactly the right shape for a platform-guard board. Verified each against main. Merged as 73889219; the new Windows-guards CI job is a welcome addition. Two small follow-ups on our side: the non-ASCII fix is #703 (not #700), and #423/#20 need reopening since the CLI narrow-argv bug your known-red guards is genuinely still present. Appreciated!

PR #646, opened 2026-06-27
fix(pipeline): stop creating Route nodes from URLs in config files

### LINKED ISSUE #521: Route nodes created from URL strings in config / non-source files
Indexing a repo that contains only config files (no source, no HTTP route definitions) creates `Route` nodes for arbitrary URL-like strings in those files. In a three-file repro I get four `Route` nodes:

- `https://app.terraform.io` (a Dependabot/Terraform registry URL, from `dependabot.yaml`)
- `https://auth.example.com/.well-known/jwks.json` (from `config.yaml`)
- `http://order-service:8080/v2/orders/{id}` (an upstream config value, from `config.yaml`)
- `curl --fail http://localhost:9000/ || exit 1` (a healthcheck shell command, from `compose.yaml`)

None of these is a route the service serves. They inflate the route set that `get_architecture` and cross-repo route matching rely on, which makes that downstream matching noisier.

Repro (self-contained, generic): https://gist.github.com/azuttre/21e5e85f7590261721e3af7aa079dbfb → `issue-1-config-urls-as-routes.sh`

Expected: no `Route` nodes. Got: the four strings above.

Env: v0.8.1, macOS (arm64).

mvanhorn:
## What does this PR do?

Indexing a repo of only config files produced spurious `Route` nodes from arbitrary URL-like strings. The infra-route extractor (`cbm_pipeline_extract_infra_routes` in `src/pipeline/pipeline.c`) harvested any `CBM_STRREF_URL` string literal from any `.yaml`/`.yml`/`.tf`/`.hcl`/`.toml` file, regardless of whether the host file actually defines service routes.

This restricts the loose URL string-ref harvesting to genuine Infrastructure-as-Code files (Terraform / HCL) and additionally requires the value to be a bare URL:

- `cbm_is_infra_route_source_file()` — only `.tf`, `.tf.json`, `.hcl` are route sources. Generic config (`config.yaml`), dependency manifests (`dependabot.yaml`), container orchestration (`compose.yaml`), and Kubernetes / Kustomize manifests are excluded.
- `cbm_is_bare_endpoint_url()` — rejects command strings that merely embed a URL (e.g. a `curl ... || exit 1` healthcheck), while still accepting query-string URLs.

Structured topic→endpoint bindings still flow through `cbm_pipeline_process_infra_bindings()`, so real infrastructure endpoints (Cloud Scheduler / Pub/Sub targets) continue to produce Route nodes.

Why this matters: per #521, a three-file repro (`dependabot.yaml` + `config.yaml` + `compose.yaml`) yielded four bogus routes — a Terraform registry URL, a JWKS discovery URL, an upstream service host, and a healthcheck shell command. None is a route the service serves; they inflate the Route set that `get_architecture` and cross-repo route matching depend on, making downstream matching noisier.

## Checklist

### Testing notes

- Built the indexer and reproduced the issue: before the fix the three-file config repro produced the four `Route` nodes from #521; after the fix it produces zero. A Terraform `.tf` endpoint URL still produces an infra Route.
- Added unit tests in `tests/test_pipeline.c`:
  - `infra_route_source_file_gate` — Terraform/HCL accepted; dependabot/config/compose/k8s/kustomize/toml and `.tfvars` rejected.
  - `infra_bare_endpoint_url_gate` — bare URLs accepted; healthcheck/command strings rejected.

Fixes #521

### DIFF
--- a/src/pipeline/pipeline.c
+++ b/src/pipeline/pipeline.c
@@ -485,15 +485,65 @@ static void cbm_pipeline_process_infra_bindings(cbm_gbuf_t *gbuf, const cbm_file
     }
 }
 
-static bool is_infra_file(const char *fp) {
-    return fp != NULL &&
-           (strstr(fp, ".yaml") != NULL || strstr(fp, ".yml") != NULL ||
-            strstr(fp, ".tf") != NULL || strstr(fp, ".hcl") != NULL || strstr(fp, ".toml") != NULL);
+/* Basename of a path: the segment after the final '/' or '\\'. */
+static const char *infra_basename(const char *fp) {
+    const char *base = fp;
+    for (const char *p = fp; *p != '\0'; p++) {
+        if (*p == '/' || *p == '\\') {
+            base = p + 1;
+        }
+    }
+    return base;
+}
+
+/* True when `name` ends with `suffix`. */
+static bool infra_ends_with(const char *name, const char *suffix) {
+    size_t nl = strlen(name);
+    size_t sl = strlen(suffix);
+    return nl >= sl && strcmp(name + nl - sl, suffix) == 0;
+}
+
+/* True for Infrastructure-as-Code files whose URL string literals denote real
+ * service endpoints: Terraform / HCL module sources, backends and provider
+ * endpoints. Generic application config (config.yaml), dependency manifests
+ * (dependabot.yaml), container-orchestration (compose.yaml) and Kubernetes /
+ * Kustomize manifests are NOT route sources — the URL-like strings they hold
+ * (package registries, JWKS discovery endpoints, upstream hosts, healthcheck
+ * shell commands) are not routes the service serves, and harvesting them
+ * inflated the Route set that get_architecture and cross-repo matching rely on
+ * (issue #521). Structured topic->endpoint bindings in config files still flow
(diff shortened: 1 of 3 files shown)


DeusData:

DeusData:
Thanks, this is nicely scoped to #521. Before merge, please rewrite the commit metadata to remove generated/co-author/session attribution. The `curl`/`wget` strings appear to be test fixtures, not network execution.

DeusData:
Thank you @mvanhorn — clean, well-scoped work, and you diagnosed #521 correctly (config-file URLs like the terraform-registry / JWKS / healthcheck endpoints were being harvested as bogus Route nodes).

It turns out #521 was already fixed on `main` by a different approach that landed right after you opened this — **#667** (`04697c6`) uses a "deny-wins-by-value" denylist (jwks / registry / healthcheck / upstream keys + a two-pass dedup) while keeping the broad file gate, whereas this PR takes a file-type allowlist (`.tf`/`.hcl` only). Both are sound; since the denylist version is already shipped, this PR now conflicts with `main` and a naive resolution would regress the dedup. So I'm closing it as superseded rather than re-architecting the shipped fix.

Really appreciate the careful scoping and the repro — genuinely useful. 🙏

PR #650, opened 2026-06-27
fix: trace_path returns full caller set across .d.ts-duplicated symbol nodes

### LINKED ISSUE #546: trace_path / reverse-dependency returns an INCOMPLETE caller set when a symbol is duplicated by an ambient `.d.ts` declaration (callers silently split by import style)
---

### Version
`codebase-memory-mcp` 0.8.1 (CLI)

### Summary
When an exported symbol is defined in real source **and** also re-declared (signature only, no body) in an ambient `.d.ts` file, the indexer creates **two distinct `Function` nodes** for the same logical symbol. `CALLS` edges from consumers are then **partitioned across the two nodes depending on how each consumer imports the symbol**:

- consumers importing via the **path alias** that the `.d.ts` types → edge points to the **`.d.ts` stub node**
- consumers importing via the **relative path** to the real file → edge points to the **implementation node**

As a result, `trace_path(name, mode=calls)` resolves the name to **one** of the two nodes and returns **only that node's callers**, silently omitting the rest. There is no warning that the symbol is fragmented, so reverse-dependency / impact analysis under-reports consumers.

### Environment (where this naturally occurs)
A TypeScript monorepo where an app consumes a workspace package through a `tsconfig` path alias, and the app ships a hand-written ambient declaration file to type that alias. This is a common pattern, so the duplicate-node split is easy to hit in real repos.

### Minimal reproduction (synthetic)
```
repo/
  packages/widget/src/scroll.ts
      export function alignToEdge(el: HTMLElement): () => void { /* real impl */ }
  packages/widget/src/internalConsumer.ts
      import { alignToEdge } from './scroll'        // relative import
      alignToEdge(node)
  app/types/widget-shim.d.ts
      export function alignToEdge(el: HTMLElement): () => void   // signature only, no body
  app/src/externalConsumer.ts
      import { alignToEdge } from '@widget'         // path-alias import, typed by widget-shim.d.ts
      alignToEdge(node)
  # tsconfig paths: "@widget": ["packages/widget/src"]   (alias typed by the .d.ts shim)
```
Index the repo, then:
```
query_graph: MATCH (n {name:'alignToEdge'}) RETURN n.qualified_name
  -> 2 nodes: packages/widget/src/scroll.ts  AND  app/types/widget-shim.d.ts

trace_path(function_name='alignToEdge', mode='calls')
  -> returns callers of ONE node only (e.g. externalConsumer), missing internalConsumer

query_graph: MATCH (n)-[:CALLS]->(t {qualified_name:'<impl node>'}) RETURN n   -> internalConsumer
query_graph: MATCH (n)-[:CALLS]->(t {qualified_name:'<.d.ts node>'}) RETURN n  -> externalConsumer
```
Neither query alone yields the full `{internalConsumer, externalConsumer}` set.

### Expected
Reverse-dependency / `trace_path(mode=calls)` for a logical symbol should return **all** of its callers, regardless of whether each caller imported via the real path or via an alias typed by an ambient `.d.ts`. Either:
- a body-less ambient `.d.ts` declaration should not create a `CALLS`-target node separate from the implementation (merge / alias the two), **or**
- `trace_path` and reverse queries should union nodes that share `name` + `signature`, **or**
- at minimum, the result should flag that the symbol resolves to multiple nodes so callers know the set is partial.

### Actual
Two separate `Function` nodes; callers partitioned by import style; `trace_path` returns a partial set with no indication it is incomplete.

### Impact
This directly undercuts the impact-analysis / blast-radius use case the tool is built for: "who depends on this symbol if I change it?" returns a confidently partial answer, which is worse than no answer for change-safety decisions.

### Question
Is the duplicate node from a body-less ambient `.d.ts` declaration intended? If so, is there a recommended query pattern to retrieve the unified caller set across alias-typed and real-path consumers? Happy to provide more detail.

mvanhorn:
## What does this PR do?

Root cause is in the `trace_path` handler in `src/mcp/mcp.c`: `cbm_store_find_nodes_by_name` returns all same-named nodes, then `pick_resolved_node` selects exactly one (`sel`) via `node_resolution_score` = label_rank*weight + (end_line - start_line). A body-less `.d.ts` declaration has span 0, so it always scores below the real implementation, loses the pick, is NOT flagged as a tie, and its callers are dropped — BFS runs only from `nodes[sel].id`. Fix per the issue's option 2 ("union nodes that share name + signature"): after `pick_resolved_node` chooses the canonical node, gather the sibling node ids that represent the same logical symbol — same `name`, callable label (`Function`/`Method`), and body-less span (start_line == end_line, the ambient-declaration signal) — and run the inbound/outbound BFS from each, merging `cbm_traverse_result_t` sets deduplicated by node id before serialization.

When an exported symbol is defined in real source AND re-declared (signature only, no body) in an ambient `.d.ts` file, the indexer creates two distinct callable nodes for the same logical symbol. `CALLS` edges from consumers get partitioned across the two nodes by import style: consumers importing via the `tsconfig` path alias point at the `.d.ts` stub node, consumers importing via the relative path point at the implementation node. `trace_path(function_name, mode=calls)` then resolves the name to a single node and returns only that node's callers, silently omitting the rest. Reverse-dependency / blast-radius analysis (the tool's core use case) thus returns a confidently partial answer with no warning.

Fixes #546

## Checklist

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -2202,6 +2202,205 @@ static yyjson_mut_val *bfs_to_json_array(yyjson_mut_doc *doc, cbm_traverse_resul
     return arr;
 }
 
+static bool trace_is_callable_node(const cbm_node_t *n) {
+    return n && n->label && (strcmp(n->label, "Function") == 0 || strcmp(n->label, "Method") == 0);
+}
+
+static bool trace_is_bodyless_declaration(const cbm_node_t *n) {
+    return n && n->start_line == n->end_line;
+}
+
+static int trace_label_rank(const cbm_node_t *n) {
+    if (!n || !n->label) {
+        return 0;
+    }
+    if (strcmp(n->label, "Function") == 0 || strcmp(n->label, "Method") == 0) {
+        return 3;
+    }
+    if (strcmp(n->label, "Route") == 0) {
+        return 2;
+    }
+    if (strcmp(n->label, "Class") == 0 || strcmp(n->label, "Interface") == 0) {
+        return 1;
+    }
+    return 0;
+}
+
+static int trace_node_resolution_score(const cbm_node_t *n) {
+    int span = n && n->end_line > n->start_line ? n->end_line - n->start_line : 0;
+    return trace_label_rank(n) * 1000 + span;
+}
+
+static int pick_trace_resolved_node(const cbm_node_t *nodes, int node_count) {
+    int sel = 0;
+    int best = trace_node_resolution_score(&nodes[0]);
+    for (int i = 1; i < node_count; i++) {
+        int score = trace_node_resolution_score(&nodes[i]);
+        if (score > best) {
+            best = score;
+            sel = i;
+        }
+    }
+    return sel;
+}
+
+static int trace_real_callable_count(const cbm_node_t *nodes, int node_count) {
+    int count = 0;
+    for (int i = 0; i < node_count; i++) {
+        if (trace_is_callable_node(&nodes[i]) && !trace_is_bodyless_declaration(&nodes[i])) {
+            count++;
+        }
+    }
+    return count;
+}
+
+static char *trace_node_signature_dup(const cbm_node_t *n) {
+    if (!n || !n->properties_json || n->properties_json[0] == '\0') {
+        return NULL;
+    }
+
+    yyjson_doc *doc = yyjson_read(n->properties_json, strlen(n->properties_json), 0);
+    if (!doc) {
+        return NULL;
+    }
+    yyjson_val *root = yyjson_doc_get_root(doc);
+    yyjson_val *sig = root && yyjson_is_obj(root) ? yyjson_obj_get(root, "signature") : NULL;
+    const char *sig_str = sig && yyjson_is_str(sig) ? yyjson_get_str(sig) : NULL;
+    char *copy = sig_str && sig_str[0] ? heap_strdup(sig_str) : NULL;
+    yyjson_doc_free(doc);
+    return copy;
+}
+
+static bool trace_signatures_compatible(const cbm_node_t *a, const cbm_node_t *b) {
+    char *sig_a = trace_node_signature_dup(a);
+    char *sig_b = trace_node_signature_dup(b);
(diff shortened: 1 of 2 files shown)


DeusData:

DeusData:
Thanks, this is nicely focused. Since it changes `trace_path` resolution and response behavior, please add a short compatibility note describing the new ambiguous/fragmented-symbol behavior and confirm existing clients still receive the current `callers`/`callees` fields unchanged.

DeusData:
Thank you @mvanhorn — this is a genuinely sharp piece of work, and your diagnosis of #546 is exactly right (a `.ts` impl + a body-less ambient `.d.ts` stub become two callable nodes, and `trace_path` traces only one → a partial caller set).

Two things happened in parallel. First, #546's *under-report* was independently fixed on `main` in `3f754cf` (`bfs_union_same_name`) a few hours before this PR, which is why it now shows `CONFLICTING` — your `trace_`-prefixed helpers collide 1:1 with the landed `bfs_union_same_name` / `pick_resolved_node` / `node_resolution_score`, so it can't merge as-is.

Second — and this is the valuable part — reviewing your PR surfaced that the *shipped* fix has a real flaw your approach avoids: `bfs_union_same_name` unions **every** same-named node unconditionally, and `pick_resolved_node` only flagged ambiguity on an exact score tie. So two genuinely-different same-named functions with different body lengths were being **over-unioned** into one confidently-conflated caller set (the mirror of #546). Your key insight — a body-less `.d.ts` stub is a *fragment* to union, but 2+ real definitions are *genuine ambiguity* — is precisely the fix.

I've distilled exactly that into **#748** (`7d6d52b`), **co-authored with you**, as a minimal gate on the existing code (2+ real callable defs → the existing disambiguation path; body-less stubs still union with their impl), guarded by two reproduce-first tests derived from your scenarios — one proving distinct symbols no longer conflate (RED→GREEN), one guarding that the `.d.ts` stub+impl union still works. Full suite green.

Closing this as superseded by the distilled follow-up, with full credit to you. The separate #546 issue is being closed as already-fixed by `3f754cf`. Really appreciate the careful work. 🙏

PR #671, opened 2026-06-28
fix(pipeline): preserve non-ASCII (UTF-8) project names

### LINKED ISSUE #571: Project name strips non-ASCII (CJK) characters from path, resulting in truncated/unrecognizable names

ouzsrcm:
Fixes #571

## What does this PR do?

Non-ASCII path components (e.g. CJK) were mapped to '-' in
cbm_project_name_from_path and rejected by cbm_validate_project_name, so a repo
like /home/u/我的项目 produced a truncated, unrecognizable project name.

This allows UTF-8 bytes (>=0x80) through in both functions, so Unicode/CJK names
survive. Traversal safety is unchanged: "..", "/", "\" and leading-dot are still
rejected (all ASCII < 0x80). Added regression test project_name_preserves_unicode;
the existing project_name_always_validator_safe_issue349 invariant still passes.

## Checklist

### DIFF
--- a/src/foundation/str_util.c
+++ b/src/foundation/str_util.c
@@ -283,10 +283,14 @@ bool cbm_validate_project_name(const char *name) {
     /* Reject leading dot (hidden files / relative refs) */
     if (name[0] == '.')
         return false;
-    /* Allow only alphanumeric, dash, underscore, dot */
+    /* Allow alphanumeric, dash, underscore, dot, and any non-ASCII (UTF-8)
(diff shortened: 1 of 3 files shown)


ouzsrcm:
Looks like this is already fixed on main via hex transliteration of non-ASCII bytes — closing as obsolete.

PR #672, opened 2026-06-28
fix(git): fix canonical_root for worktrees and subdirectory projects

### LINKED ISSUE #659: detect_changes: impacted_symbols always empty for projects inside git worktrees (canonical_root miscalculated)
## Summary

`detect_changes` correctly returns `changed_files` but always returns `impacted_symbols: []` / `impacted_count: 0` when the indexed project lives inside a **git worktree** (not the worktree root itself).

The root cause is that `canonical_root` is computed one level too high, causing a path prefix mismatch between git diff output and graph file paths.

## Environment

- codebase-memory-mcp: v0.8.1
- OS: macOS 25.5.0 (Darwin)
- Project layout: git monorepo at `/workspace/`, with sub-projects indexed from subdirectories (e.g., `/workspace/scripts/`, `/workspace/runtime/solar-monitor/`)

## Repository Structure

```
/workspace/                         ← actual git root (.git lives here)
├── scripts/
│   └── solar_daily_report_v1.py   ← indexed as project "workspace-scripts"
├── runtime/
│   └── solar-monitor/              ← indexed as project "workspace-runtime-solar-monitor"
```

## Steps to Reproduce

1. Index a project whose `root_path` is a **subdirectory** of the git worktree root (e.g., `workspace/scripts/`)
2. Call `detect_changes` with `scope: "impact"` and `since: "HEAD~1"`

```json
{
  "project": "Users-openclaw-.openclaw-workspace-scripts",
  "since": "HEAD~1",
  "scope": "impact",
  "depth": 2
}
```

3. Observe: `changed_files` is correctly populated; `impacted_symbols` is `[]`

## Expected Behavior

`impacted_symbols` should list all symbols in the changed files plus their transitive callers.

## Actual Behavior

```json
{
  "changed_files": ["scripts/solar_daily_report_v1.py", ...],
  "changed_count": 120,
  "impacted_symbols": [],
  "impacted_count": 0,
  "depth": 2
}
```

## Root Cause

`list_projects` reveals the miscalculation:

```json
{
  "name": "Users-openclaw-.openclaw-workspace-scripts",
  "root_path": "/Users/openclaw/.openclaw/workspace/scripts",
  "git": {
    "worktree_root": "/Users/openclaw/.openclaw/workspace",
    "git_common_dir": "../.git",
    "canonical_root": "/Users/openclaw/.openclaw/workspace/.."
  }
}
```

| Field | Value | Notes |
|-------|-------|-------|
| Actual git root | `/Users/openclaw/.openclaw/workspace/` | confirmed via `git rev-parse --show-toplevel` |
| `canonical_root` | `/Users/openclaw/.openclaw/` | **one `..` too many** |
| git diff path | `scripts/solar_daily_report_v1.py` | relative to actual git root |
| Expected path for matching | `workspace/scripts/solar_daily_report_v1.py` | relative to wrong canonical_root |

The path prefix that `detect_changes` strips from git diff paths (`workspace/scripts/`) does not match what the graph stored (`solar_daily_report_v1.py` relative to `root_path`), so no symbols are ever matched.

The symbol **is** present in the graph (verified via `search_graph`):
```json
{
  "name": "solar_daily_report_v1.py",
  "qualified_name": "Users-openclaw-.openclaw-workspace-scripts.solar_daily_report_v1",
  "label": "Module",
  "file_path": "solar_daily_report_v1.py",
  "out_degree": 81
}
```

The `canonical_root` appears to be computed as `worktree_root + "/" + git_common_dir + "/.."`  which for `git_common_dir = "../.git"` resolves to `workspace/../.git/..` = `.openclaw/` instead of the correct `workspace/`.

## Workaround

Currently falling back to manual `trace_path` + `search_graph` to assess blast radius for shared library changes.

## Additional Notes

- `changed_files` detection works correctly in all cases
- Projects indexed from the worktree root directly are unaffected
- Standalone git repos (not worktrees) are unaffected — `detect_changes` works correctly there
- This affects all `Users-openclaw-.openclaw-workspace-*` projects (anything inside the workspace)

anivaryam:
## Summary

Fixes #659 — `detect_changes` returns empty `impacted_symbols` for projects indexed inside git worktrees or as repo subdirectories because `canonical_root` was computed incorrectly.

**Root cause:** `git rev-parse --git-common-dir` outputs a path relative to the directory passed via `-C` (the indexed path, `input_path`), not to `worktree_root`. Two errors in `derive_canonical_root()`:

1. The relative `git_common_dir` (e.g. `"../.git"`) was joined with `worktree_root` instead of `input_path`. For a project at `workspace/scripts/`, `worktree_root` is `workspace/` but the relative path is relative to `scripts/` — joining it with `workspace/` produced `workspace/../.git`, which strips to `workspace/..`.
2. Even if joined correctly (`scripts/../.git`), stripping the `/.git` suffix leaves `scripts/..` — a path with unresolved `..` components. Without normalization this is stored as-is and never matches the project's stored root path.

**Fix (`src/git/git_context.c`):**
- Pass `input_path` (the original indexed directory) as the join base for relative `git_common_dir` paths.
- Call `realpath()` (POSIX) / `_fullpath()` (Windows) on the joined `.git` path _before_ stripping the suffix. The `.git` directory always exists for a valid repository, so the call succeeds and collapses any `..` components.

**Tests (`tests/test_git_context.c`):**
Three new tests in `suite_git_context`:
- `canonical_root_repo_root` — normal repo indexed from its root (regression guard)
- `canonical_root_subdir` — project indexed from a subdirectory of the repo (issue #659)
- `canonical_root_linked_worktree` — project in a linked git worktree (issue #659 primary case)

## Test plan

- [ ] `canonical_root_repo_root` passes
- [ ] `canonical_root_subdir` passes (was failing before fix)
- [ ] `canonical_root_linked_worktree` passes (was failing before fix)
- [ ] Full `make test` suite has no regressions


DeusData:

DeusData:
Thanks, this looks focused on #659. Before merge, please add DCO sign-off to the non-merge commit. Also, because the regression test shells out to `git`, please confirm the command inputs are fixed/test-controlled and note expected CI platform coverage.

DeusData:
Thank you @anivaryam — this is a correct and well-diagnosed fix. You nailed the root cause: `git rev-parse --git-common-dir` is relative to the `-C` dir, and the old code joined it against `worktree_root` and stripped `/.git` textually, leaving an unresolved `..` that never matched the stored root.

I've taken your `git_context.c` fix **verbatim** into a distilled follow-up, **#749** (`354befe`), **co-authored with you**. The only change is the test suite: your `canonical_root_repo_root` / `canonical_root_subdir` tests called `make_git_repo()` (which shells out to `git`) unconditionally, so the Windows CI leg failed with *"failed to init git repo"* (`ci-ok` red) — while the linked-worktree test was already `SKIP`ed there. In the follow-up all three git-based tests `SKIP_PLATFORM` on Windows (matching that existing convention), so the suite is green cross-platform, and the linked-worktree case is kept as the genuine reproduce-first guard (it's the one that reliably produces a relative `--git-common-dir` and fails on the unfixed code).

Closing this as superseded by the distilled PR, with full credit to you. Really appreciate the careful work — the fix ships in the next release. 🙏

PR #674, opened 2026-06-28
Use /usr/bin/env bash for widely compatible shebang

SuperSandro2000:
## What does this PR do?

Fix the shebangs of scripts to work on eg NixOS.

## Checklist

### DIFF
--- a/scripts/benchmark-search-graph.sh
+++ b/scripts/benchmark-search-graph.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # benchmark-search-graph.sh — Time search_graph name_pattern= queries against a
 # codebase-memory-mcp binary to measure the regex / LIKE pre-filter performance.
 #
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # build.sh — Clean build of production binary (standard or with UI).
 #
 # Usage:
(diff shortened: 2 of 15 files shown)


DeusData:

DeusData:
Thanks. This looks like a focused portability change. Since it touches build scripts and generated hook scripts, please link the tracking issue or add a short rationale for the NixOS compatibility need, and we’ll need normal CI validation before merge. One security-hygiene nit: please change the fake `ghp_...` URL fixture to a deliberately non-token-looking placeholder so secret scanners don’t treat it like a leaked token shape.

DeusData:
Thanks for converging the remaining holdouts — we verified the repo already uses env-form on ~20 newer scripts, so this brings full consistency, and MSYS2 provides /usr/bin/env so the Windows legs are safe. Three small asks and this merges: (1) the two open questions from earlier (a rationale/tracking link, and the ghp_-shaped fixture string); (2) please keep at least one '#!/bin/bash' case in the parser tests — the current sweep removed the only coverage for absolute-path shebang extraction in infra_parse_shell; (3) FYI the hook-script hunks change runtime resolution on every user's machine (PATH-resolved bash instead of pinned /bin/bash) — standard practice and fine by us, just worth a sentence in the PR body. Thanks!


---

DeusData:
Thanks for converging the shebang holdouts — NixOS users will appreciate it. We carried it over the line in 010ac9b2 (PR #812) with you credited as co-author, with two small deltas from the review: the absolute-path shebang parser fixtures keep one #!/bin/bash case (so that extraction path stays covered), and the ghp_-shaped fixture string became an obviously-fake placeholder that still trips the secret detector. A third hook heredoc had appeared since your branch (SubagentStart) — converted too, so the tree is now fully consistent at 0 raw /bin/bash shebangs outside fixtures. Closing in favor of the distill — thanks again!

PR #683, opened 2026-06-29
fix(discover): honor info/exclude and core.excludesFile in linked worktrees

### LINKED ISSUE #682: index_repository ignores .git/info/exclude and core.excludesFile inside linked git worktrees
### Version

`codebase-memory-mcp 0.8.1` (also reproduces on `main` @ 9169f46, built from source)

### Platform

macOS (Apple Silicon)

### Install channel

GitHub release archive / install.sh (0.8.1); Built from source (main)

### Binary variant

standard

#489 and #499 taught the indexer to honor `.git/info/exclude` and the global `core.excludesFile`, but only for an ordinary checkout. Inside a **linked git worktree** both are dropped again, so directories you excluded there get walked and indexed.

The reason is that in a linked worktree `<repo>/.git` is a regular *file* (a gitlink, `gitdir: <path>`), not a directory. `cbm_discover_ex()` gates every per-clone ignore source on `.git` being a directory:

```c
bool is_git_repo = wide_stat(gi_path, &gi_stat) == 0 && S_ISDIR(gi_stat.st_mode);
```

When that check is false, `info/exclude` and `config` are never read. Committed `.gitignore` still applies (the #510 path loads `<repo>/.gitignore` unconditionally), which is why only the non-committed excludes break and the gap is easy to overlook.

This is the same class of problem #489 was opened for (agent/sandbox worktrees blowing up the index), just from the other side: #489 indexes the *main* repo whose worktrees sit under an `info/exclude` path; this is indexing the *worktree itself*, which is exactly what branch-per-worktree agent workflows do. I hit it on a Zephyr firmware worktree where indexing pulled in ~31k nodes of `build/` output that `git check-ignore` reports as ignored.

Expected: a worktree honors `info/exclude` and `core.excludesFile` the same way an ordinary checkout does.

### Reproduction

```bash
T=$(mktemp -d); cd "$T"
git init -q main && cd main
echo 'def included(): return 1' > main.py
git add -A && git -c user.email=t@t -c user.name=t commit -qm init

# add a linked worktree, then exclude build/ ONLY via the shared info/exclude
git worktree add -q ../wt -b feat
echo 'build/' > .git/info/exclude
mkdir -p ../wt/build
echo 'def build_artifact(): return 2' > ../wt/build/gen.py

# git agrees build/ is ignored inside the worktree:
( cd ../wt && git check-ignore build/gen.py )   # -> build/gen.py
head -1 ../wt/.git                                # -> gitdir: .../worktrees/wt   (a file, not a dir)

codebase-memory-mcp cli index_repository "{\"repo_path\":\"$T/wt\"}"
```

Actual: the index summary has no `build` in `excluded`, and `build/gen.py` is indexed:

```
"nodes":8        # main.py + build/gen.py
# search_code build_artifact -> "file":"build/gen.py","total_results":1
```

Expected: `build/` excluded, only `main.py` indexed (matching `git check-ignore`).

For contrast, the identical layout in a normal checkout (`info/exclude` in a non-worktree `.git/`) excludes `build/` correctly, so the divergence is specific to the worktree gitlink.

### Confirmations

- [x] I searched existing issues and this is not a duplicate. (Closest are #489 and #499, both closed and both fixed for ordinary checkouts only; this is the linked-worktree follow-up.)

Naam:
## What does this PR do?

Closes #682.

In a linked git worktree, `<repo>/.git` is a regular file containing `gitdir: <path>`, not a directory. `cbm_discover_ex()` gated every per-clone ignore source on `.git` being a directory:

```c
bool is_git_repo = wide_stat(gi_path, &gi_stat) == 0 && S_ISDIR(gi_stat.st_mode);
```

So when you index a worktree directly, `.git/info/exclude` and the global `core.excludesFile` are skipped, and the walker descends into directories you excluded there (`build/`, `twister-out/`, and so on). The committed `.gitignore` still applies because the #510 path loads `<repo>/.gitignore` unconditionally, which is why only the non-committed excludes broke. This is the worktree side of #489 and #499, both of which fixed the ordinary-checkout case only.

The fix resolves the git common dir up front. When `.git` is a gitlink file, it parses `gitdir:`, then follows `<gitdir>/commondir` to the shared directory where git keeps `info/exclude` and `config` (falling back to the gitdir itself when there is no `commondir` file). `info/exclude` and `config` are read from that common dir, so a worktree behaves like an ordinary checkout. A plain `.git` directory resolves to itself, so ordinary repos are unaffected.

Reproduce-first test `discover_worktree_info_exclude` fails before the change (`count==2`, `build/` leaks) and passes after. `discover_worktree_committed_gitignore` guards the committed-`.gitignore`-in-worktree path.

## Checklist

- [x] Every commit is signed off (`git commit -s`), verified locally with `scripts/check-dco.sh`
- [x] Tests pass locally (`make -f Makefile.cbm test`): 5714 passed, 0 failed under ASan/UBSan; discover suite 84/84
- [x] Lint passes (`lint-format`, `lint-no-suppress`); `lint-cppcheck` not run locally (cppcheck not installed), left to CI

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -737,6 +737,107 @@ static void walk_dir(const char *dir_path, const char *rel_prefix, const cbm_dis
 
 /* ── Public API ───────────────────────────────── */
 
+static bool discover_path_is_absolute(const char *path) {
+    if (!path || !path[0]) {
+        return false;
+    }
+    if (path[0] == '/' || path[0] == '\\') {
+        return true;
+    }
+#ifdef _WIN32
+    return isalpha((unsigned char)path[0]) && path[1] == ':';
+#else
+    return false;
+#endif
+}
+
+/* Resolve the shared "common" git directory for repo_path.
+ * Handles three layouts:
+ *   1. <repo>/.git is a directory    - ordinary repo; common_dir == <repo>/.git
+ *   2. <repo>/.git is a regular file - linked worktree gitlink "gitdir: <path>";
+ *      the common dir is read from <git_dir>/commondir (git stores info/exclude +
+ *      config there, shared across worktrees). Falls back to git_dir when no
+ *      commondir file exists.
+ *   3. neither - not a git repo.
+ * Returns true when a git dir was resolved. Fixes the worktree case where
+ * .git/info/exclude and core.excludesfile were silently dropped because the old
+ * check required .git to be a directory (issue #489 only covered ordinary repos). */
+static bool resolve_git_common_dir(const char *repo_path, char *common_dir, size_t cd_sz) {
+    char dot_git[CBM_SZ_4K];
+    snprintf(dot_git, sizeof(dot_git), "%s/.git", repo_path);
+    struct stat st;
+    if (wide_stat(dot_git, &st) != 0) {
+        return false;
+    }
+    if (S_ISDIR(st.st_mode)) {
+        snprintf(common_dir, cd_sz, "%s", dot_git);
+        cbm_normalize_path_sep(common_dir);
+        return true;
+    }
+    if (!S_ISREG(st.st_mode)) {
+        return false;
+    }
+
+    /* Linked worktree: parse "gitdir: <path>" from the gitlink file. */
+    FILE *f = fopen(dot_git, "r");
+    if (!f) {
+        return false;
+    }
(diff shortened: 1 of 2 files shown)


DeusData:

PR #685, opened 2026-06-29
perf(index): bound parallel source retention for low RAM

nguyentamdat:
## Summary
- bound parallel indexing source retention by memory budget instead of retaining every source file unconditionally
- fall back to short-lived source reads during fused cross-LSP resolution when a file was not retained
- add regression coverage for Python cross-file LSP resolution when `retain_sources=false`

## Verification
- `make -f Makefile.cbm test` -> `5728 passed`

## Notes
Split out from #684 so the low-RAM indexing change is reviewed independently from the JVM callgraph fix.

### DIFF
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -40,15 +40,6 @@ enum {
 #define PP_FIELD_HINT_CONF 0.85
 enum { PP_CSHARP_M_PREFIX_LEN = 2 };
 
-/* Source-retention caps for the parallel pipeline. The extract worker
- * copies source bytes into result->arena so the fused cross-file LSP
- * step in resolve_worker can run without re-reading from disk. Bound
- * peak RSS with a per-file cap (skip retention for pathological huge
- * generated files) and a total project-wide cap (skip when budget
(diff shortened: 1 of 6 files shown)


DeusData:

DeusData:

DeusData:
Thank you — your core design was right and we carried it over the line: budget-derived retention bounds plus the bounded re-read fallback so a file over the cap still gets its cross-file resolution instead of silently losing edges. Landed as **6404747b** (PR #854) with you credited as co-author. Two refinements from a rust-analyzer memory-model study went in on top: the cap is source-text-specific and modest (a cap is a floor as well as a ceiling — a large RAM-derived cap would just *hold* that much), with CBM_RETAIN_TOTAL_MB / CBM_RETAIN_PER_FILE_MB env knobs; and while building it we found your original Python red test was a false guard (the per-file py_lsp already resolved those calls, so it passed without the fix) and replaced it with a genuine Java↔Kotlin cross-file fixture that's actually red on the unfixed path. Closing in favor of the distill — this is the retention layer of the #832 memory work; thanks for the solid foundation!

PR #693, opened 2026-06-29
fix(search): use NUL-separated filelist + xargs -0 for paths with spaces

### LINKED ISSUE #687: fix(search): search_code returns 0 matches when repo path contains spaces on macOS
## Bug

`search_code` returns 0 matches for all patterns when the indexed repository path contains a space character on macOS. This is a common scenario for users with directories like `/Users/name/My Project/`.

## Reproduction

1. Create or clone a repo into a path with spaces, e.g. `/tmp/test repo/`
2. Run `index_repository` on that path
3. Run `search_code` with any pattern that should match
4. Observe 0 results despite the graph being indexed

## Expected

search_code should return matches regardless of spaces in the repo path.

## Environment

- macOS (Darwin)
- codebase-memory-mcp v0.8.1+

## Related

- #670 (similar 0-match issue on macOS arm64)
- #636 (Windows path issue with spaces)

lg320531124:
## Problem

`search_code` returns 0 matches when the indexed repo path contains spaces (e.g. `/Users/name/My Project/`).

Root cause: `write_scoped_filelist()` writes newline-separated paths, and `xargs` splits on whitespace — so a path like `/tmp/test repo/src/main.c` becomes two arguments: `/tmp/test` and `repo/src/main.c`.

## Fix

- `write_scoped_filelist()`: write NUL-separated paths via `fwrite()` instead of `fprintf()` with newline separators
- `build_grep_cmd()`: add `-0` flag to `xargs` so it reads NUL-terminated entries

Non-scoped mode was already safe (`root_path` is single-quoted in the grep command: `grep -rn ... '%s'`).

## Verification

Indexed and searched a repo at `/tmp/test repo/` — `search_code` now returns correct results:

```json
{"results":[{"node":"main","qualified_name":"tmp-test-repo.src.main.main","label":"Function","file":"src/main.c","start_line":2,"end_line":5,"match_lines":[3]}],"total_grep_matches":1,"total_results":1}
```

Previously: 0 matches.

Closes #687

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -3821,10 +3821,10 @@ static void build_grep_cmd(char *cmd, size_t cmd_sz, bool use_regex, bool scoped
     const char *flag = use_regex ? "-E" : "-F";
     if (scoped) {
         if (file_pattern) {
-            snprintf(cmd, cmd_sz, "xargs grep -Hn %s --include='%s' -f '%s' < '%s' 2>/dev/null",
+            snprintf(cmd, cmd_sz, "xargs -0 grep -Hn %s --include='%s' -f '%s' < '%s' 2>/dev/null",
                      flag, file_pattern, tmpfile, filelist);
         } else {
-            snprintf(cmd, cmd_sz, "xargs grep -Hn %s -f '%s' < '%s' 2>/dev/null", flag, tmpfile,
+            snprintf(cmd, cmd_sz, "xargs -0 grep -Hn %s -f '%s' < '%s' 2>/dev/null", flag, tmpfile,
                      filelist);
         }
     } else {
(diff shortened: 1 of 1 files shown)


DeusData:

DeusData:
Thank you @lg320531124 — you correctly root-caused #687, and the fix concept is exactly right: a newline-separated filelist + plain `xargs` splits on spaces, while NUL-separated records + `xargs -0` keep each path a single argument (and you correctly kept newline for the Windows PowerShell `Get-Content -LiteralPath` path). No command injection either — filenames flow through the filelist to `xargs`, never a shell.

One issue kept it from merging as-is: the filelist writer builds each record in a fixed `char buf[CBM_SZ_2K]` (2048) with unbounded `memcpy`s of `<root>` + `/` + `<file>` + separator. Indexed entries are **absolute** paths, and on Linux `PATH_MAX` is 4096 — so a long workspace root + nested path exceeds 2048 and overflows the stack buffer (a memory-safety regression that's actually worse than the 0-matches bug, and green CI didn't catch it because no test exercises a long path).

I've distilled your fix into **#751** (`3dd2471`), **co-authored with you**: same NUL / `xargs -0` approach + your Windows newline branch, but the record is written piece-by-piece with `fwrite`/`fputc` (no fixed buffer, safe for any length), plus a reproduce-first test that indexes a repo under `"<tmp>/my project"` and asserts the match is found (RED before → GREEN after). Full suite green.

Closing as superseded by the distilled PR, with full credit to you. Really appreciate the correct diagnosis. 🙏

PR #695, opened 2026-06-29
fix(discover): honor .git/info/exclude and core.excludesFile in linked worktrees

### LINKED ISSUE #689: fix(git): index_repository ignores .git/info/exclude and core.excludesFile inside linked git worktrees
## Bug

When a repository is a linked git worktree, `index_repository` does not honor:

1. `.git/info/exclude` — the per-repo local exclude rules
2. `core.excludesFile` — the global git config pointing to a custom ignore file

This causes files that should be ignored (build artifacts, IDE configs, etc.) to be indexed, bloating the graph and returning irrelevant search results.

## Context

Linked worktrees have a `.git` file (not directory) pointing to the main worktree's git dir. The exclude resolution path differs from regular repos:

- Regular: `.git/info/exclude` → `.gitignore` → `core.excludesFile`
- Worktree: `.git` → read pointer → `.git/worktrees/<name>/info/exclude`

## Related

- #572 (recently fixed `core.excludesFile` support for regular repos)
- #659 (detect_changes also broken in worktrees)

lg320531124:
## What Problem This Solves

Linked git worktrees have a `.git` *file* (not directory) containing `gitdir: <path>` pointing to the worktree-specific git directory under the main repo's `.git/worktrees/<name>/`. The discover module checked `S_ISDIR` on `.git`, which returned `false` for worktrees, causing two issues:

1. **`.git/info/exclude` not honored** — per-repo local exclude rules were skipped
2. **`core.excludesFile` not evaluated** — git config check also used wrong path

## Why This Change Was Made

Added `resolve_git_dir_from_dotgit_file()` to parse the `gitdir:` pointer and resolve the actual git directory. Both `info/exclude` and `config` lookups now use the resolved path.

## Evidence

- Built and tested on macOS arm64 with a linked git worktree
- Before fix: `local_only.txt` (in `info/exclude`) was indexed
- After fix: correctly excluded; only `main.c` indexed
- Regular repos unaffected

Fixes #689

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -585,6 +585,76 @@ static int wide_stat(const char *path, struct stat *st) {
 #endif
 }
 
+/* Resolve the real git directory from a linked worktree's .git file.
+ * Linked worktrees have a .git *file* (not directory) containing
+ * "gitdir: <absolute-path>" pointing to the worktree-specific git dir
+ * under the main repo's .git/worktrees/<name>/.
+ * Returns true and writes the resolved path into `out` on success.
+ * Fixes issue #689: .git/info/exclude and core.excludesFile were ignored
+ * inside linked worktrees because the .git-is-directory check failed. */
+static bool resolve_git_dir_from_dotgit_file(const char *repo_path, char *out, size_t out_sz) {
+    char dotgit_path[CBM_SZ_4K];
+    snprintf(dotgit_path, sizeof(dotgit_path), "%s/.git", repo_path);
+
+    struct stat st;
+    if (wide_stat(dotgit_path, &st) != 0 || !S_ISREG(st.st_mode)) {
+        return false;
+    }
+
+    FILE *f = fopen(dotgit_path, "r");
+    if (!f) {
+        return false;
+    }
+
+    char line[CBM_SZ_4K];
+    if (!fgets(line, sizeof(line), f)) {
+        fclose(f);
+        return false;
+    }
+    fclose(f);
+
+    /* Trim trailing newlines/carriage returns. */
+    size_t len = strlen(line);
+    while (len > 0 && (line[len - 1] == '\n' || line[len - 1] == '\r')) {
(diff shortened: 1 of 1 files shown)


DeusData:

DeusData:

lg320531124:
Done — rewrote the commit metadata to drop the `Co-Authored-By` trailer (head now `5d7315a`). The `Signed-off-by` DCO line and the rest of the message are unchanged. Ready for merge.

DeusData:
Thanks, confirmed the metadata cleanup. Current checks and DCO are green; this is back in the maintainer review queue.

DeusData:
Thank you @lg320531124 — you correctly diagnosed #689 (a linked worktree's `.git` is a gitlink *file*, so `info/exclude` / `core.excludesFile` were skipped), and the gitlink-parsing approach is the right one.

The catch is timing + completeness: #689 is the same bug as #682, which was already fixed on `main` in **#683** (`resolve_git_common_dir`) and hardened in **#750** (the reads switched to `cbm_fopen` for non-ASCII Windows paths) — which is why this now shows `CONFLICTING`.

Two things the merged version does that this PR doesn't:
- It follows the **`commondir`** indirection. For a linked worktree `GIT_DIR` is `.git/worktrees/<name>`, but the shared `info/exclude` and `config` live in `GIT_COMMON_DIR` (the main `.git`), reached via the `<gitdir>/commondir` file. This PR reads `<gitdir>/info/exclude` directly, so it would look in `.git/worktrees/<name>/info/exclude` (which doesn't exist) and still miss the shared rule that #682's own repro places in the main repo's `.git/info/exclude`.
- It uses `cbm_fopen` (the UTF-8→`_wfopen` wrapper) rather than raw `fopen`, so non-ASCII repo paths open on Windows.

So I'm closing this as already-fixed by #683/#750 rather than merging. Genuinely appreciate the correct diagnosis and the clean, bounded implementation. 🙏

Issue #668, opened 2026-06-28
Indexing large SQL file crashes native parser (Tree-sitter SQL assertion / stack overflow)


georgelichen:
## Summary

`index_repository` can crash the native process while indexing a large `.sql` file. In a full repository run I observed a Tree-sitter assertion:

```text
Assertion failed: symbol < self->token_count, file internal/cbm/vendored/ts_runtime/src/./language.c, line 79
```

I then narrowed this down to a single large SQL file. I cannot attach the original file because it contains private/customer database schema and data scripts, but the failure is reproducible with that single file copied into an otherwise empty temporary repository.

## Environment

- OS: Windows
- Shell: PowerShell
- Binary: `C:/Users/lichen/.local/bin/codebase-memory-mcp.exe`
- `codebase-memory-mcp --version`: `codebase-memory-mcp 0.8.1`
- MCP `initialize` serverInfo reported: `codebase-memory-mcp` `0.10.0`
- Command mode: `codebase-memory-mcp cli index_repository ...`

## Reproduction Shape

Full repository indexing crashed while scanning a repository with many SQL scripts:

```powershell
codebase-memory-mcp.exe cli index_repository '{"repo_path":"C:/path/to/repo","mode":"full"}'
```

The full run reached SQL files under `Src/Database` and then aborted in the Tree-sitter runtime.

To reduce the input, I copied one SQL file into a new empty directory and indexed only that directory:

```powershell
$root = "$env:TEMP/cbm-sql-repro"
New-Item -ItemType Directory -Force -Path $root | Out-Null
Copy-Item C:/path/to/CreateDB.sql "$root/CreateDB.sql" -Force
codebase-memory-mcp.exe cli index_repository "{`"repo_path`":`"$($root.Replace('\','/'))`",`"mode`":`"full`"}"
```

The single file is about 2.6 MiB and contains T-SQL database creation/schema/data script content.

## Actual Result

Single-file repro exits with native crash code `-1073741571` (`0xC00000FD`, stack overflow on Windows). The output stops during the definitions pass:

```text
level=info msg=mem.init budget_mb=32767 total_ram_mb=65534
level=info msg=pipeline.discover files=1 elapsed_ms=1
level=info msg=pipeline.route path=full
level=info msg=pass.start pass=structure files=1
level=info msg=pass.done pass=structure nodes=2 edges=1
level=info msg=pass.timing pass=structure elapsed_ms=0
[...]

PR #698, opened 2026-06-29
fix(extract): guard against large SQL file stack overflow (#691)

### LINKED ISSUE #691: fix(extract): indexing large SQL files crashes native parser (tree-sitter SQL assertion / stack overflow)
## Bug

Indexing a repository that contains large SQL files (e.g. schema dumps >10K lines, stored procedures) causes the native parser to crash with either:

- A tree-sitter SQL assertion failure
- A stack overflow in the C parser

This crashes the entire indexing pipeline rather than gracefully degrading for that file.

## Expected behavior

- Large SQL files that exceed the parser's capacity should be skipped with a warning, not crash the pipeline.
- Alternatively, increase the stack size for the SQL tree-sitter grammar or add a file-size/line-count guard before parsing.

## Suggested fix

Add a pre-parse guard in the extraction pipeline:

```c
if (line_count > MAX_SQL_PARSE_LINES) {
    log_warn("skipping large SQL file: %s (%d lines)", path, line_count);
    return SKIP;
}
```

## Related

- #668 (original report)

lg320531124:
## Summary

The tree-sitter SQL grammar (39 MB `parser.c`) uses deeply recursive non-terminals that overflow the C stack on files with many statements (e.g. schema dumps >10K lines, stored procedures). The stack overflow kills the thread **before** any timeout callback can fire, so a pre-parse guard is the only safe mitigation.

## Changes

- Add a line-count guard (`CBM_SQL_MAX_LINES = 5000`) in `cbm_extract_file()` that returns `has_error` for SQL files exceeding the threshold
- The pipeline skips the file gracefully instead of crashing

## Why line-count, not byte-size?

- tree-sitter's recursive descent parser pushes a stack frame per statement
- A 500KB SQL dump with 10K `CREATE TABLE` lines overflows the default 512KB macOS thread stack
- A 2MB compressed binary blob with few newlines does NOT overflow
- Line count is the correct proxy for SQL stack depth

## Why 5000 lines?

- Default macOS thread stack: 512 KB
- tree-sitter SQL frame size: ~100 bytes per recursion
- 5000 lines → ~500 KB stack usage (safe margin)
- Captures virtually all crash-inducing workloads

## Why in `cbm_extract_file()`, not the discover layer?

- The discover layer only checks byte size and knows nothing about SQL
- The guard must be where tree-sitter is invoked — that is where the crash happens
- Keeps the change minimal and colocated with the dangerous call

## Testing

- All 5720 existing tests pass
- Small SQL files (<5000 lines) parse normally
- Large SQL files (6000 lines) are skipped gracefully with `has_error=true`

Fixes #691

Related: #668 (original report)

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -515,6 +515,31 @@ CBMFileResult *cbm_extract_file(const char *source, int source_len, CBMLanguage 
         return result;
     }
 
+    /* Guard: skip parsing large SQL files that crash the tree-sitter SQL
+     * grammar.  The SQL grammar (39 MB parser.c) uses deeply recursive
+     * non-terminals that overflow the C stack on files with many statements
+     * (e.g. schema dumps >10K lines, stored procedures).  The stack overflow
+     * kills the thread before any timeout callback can fire, so we must
+     * reject the file *before* calling ts_parser_parse_with_options().
+     * Fixes #691 — large SQL files crash native parser. */
+    if (language == CBM_LANG_SQL) {
+        enum { CBM_SQL_MAX_LINES = 5000 };
+        int lines = 0;
+        for (const char *p = source; *p; p++) {
+            if (*p == '\n') {
+                lines++;
+                if (lines >= CBM_SQL_MAX_LINES) {
+                    break;
+                }
+            }
+        }
+        if (lines >= CBM_SQL_MAX_LINES) {
+            result->has_error = true;
+            result->error_msg = cbm_arena_strdup(a, "SQL file too large for tree-sitter parsing");
+            return result;
+        }
+    }
+
     // Get tree-sitter language
     const TSLanguage *ts_lang = cbm_ts_language(language);
     if (!ts_lang) {


DeusData:

DeusData:
Thank you @lg320531124 — you're right that a large `.sql` file crashing the whole index run (#668, of which #691 is a duplicate) is a real, high-priority problem, and reaching for a pre-parse guard is the right instinct.

After analysis we're not going to take the **line-count** cap, because line count is the wrong signal for this crash: a recursive/GLR parser overflows on *depth*, not size — a 20-line but thousands-deep-nested statement still crashes and sails under 5000 lines, while a legitimately large-but-flat 6000-line migration parses fine and would be dropped by this cap. It happens to stop the one reported schema dump, but by coincidence (that file being both big and a real crasher), not by addressing the mechanism.

Instead we've planned a proper, general fix (reproduce-first): (1) root-cause the actual tree-sitter SQL crash and fix it at the grammar/scanner level so it stops crashing on **all** platforms; and (2) a general resilience layer so that **any** file that fails during indexing — any language, parse or LSP — is skipped-and-reported (per-run logfile + surfaced via MCP/CLI) instead of aborting the run, with a cross-platform process supervisor so a hard crash can never take the whole run down. That resolves #668 and the whole class of similar issues.

Closing this in favor of that effort — genuinely appreciate the report and the push on it. 🙏

PR #699, opened 2026-06-29
fix(complexity): receiver-aware self-recursion detection (#599)

### LINKED ISSUE #592: task: Parsing & extraction quality — language/format coverage gaps
## Scope

Umbrella tracker for **graph extraction quality** — language/format coverage gaps, missing node/edge kinds, and extraction false positives. These are not crashes (see the stability/performance umbrellas) but cases where indexing *succeeds* yet the resulting graph is shallow, mistyped, or wrong.

### Language coverage / LSP depth
- [ ] #382 — Java: `@Annotation`, signatures, and AST properties missing from graph nodes
- [ ] #405 — Rust LSP (hybrid-LSP tier resolution)
- [ ] #535 — Hybrid LSP support for Julia
- [ ] #415 — Index inner declarations of factory/setup callbacks (Vue Pinia setup stores, composables, React hooks)

### Format / IaC / config extraction
- [ ] #450 — GitHub Actions semantic extraction (workflows, jobs, steps, uses/needs edges)
- [ ] #451 — Python class-field nodes, type-annotation edges, enum members
- [ ] #452 — Terraform module composition edges + block-type labels
- [ ] #454 — Helm nested `.Values` linkage, template→kind typing, image/env + hook/RBAC edges

### Correctness / false positives
- [ ] #495 — cfg-gated twin functions collapse into one node; `get_code_snippet` returns inactive branch
- [ ] #521 — Route nodes created from URL strings in config / non-source files

### Content / docs indexing (BM25)
- [ ] #490 — Index documents as well
- [ ] #518 — Section nodes don't index body text — BM25 can't search markdown content
- [ ] #519 — META.yaml / frontmatter description values not indexed for BM25

### Tracked via open PR (not part of this task's open work)
- #438 / #554 — cross-file & C++ out-of-line CALLS edges resolve to Module instead of the enclosing function — PR #463
- #459 — Perl LSP-tier semantic resolution — PR #461
- #462 — ObjectScript (InterSystems IRIS) language support — PR #467 / #590
- #440 — cross-repo Maven library dependency links — PR #442
- #574 — SQL DDL Table/View nodes + FROM/JOIN lineage — PR #582
- #575 — dbt lineage/macros from raw `.sql` — PR #584
- #576 — dbt `manifest.json` ingest + DEPENDS_ON — PR #583

## Acceptance

Per item: the missing nodes/edges are emitted (or the false positive is suppressed), with a reproduce-first test on a public fixture, or the item is closed with rationale (out of scope / by design).

## Why one task

These share the extraction pipeline (tree-sitter queries + hybrid-LSP resolvers + node/edge emission). Triaging them together keeps language/format coverage coherent instead of scattered one-off queries.

### Language coverage / LSP depth
- [ ] #669 — Oracle PL/SQL support
- [ ] #694 — C/C++ `CALLS` edges attributed to File/Module instead of enclosing function

### Routes, cross-repo, and protocol extraction
- [ ] #686 — Go Fiber route paths empty and gRPC not captured; cross-repo edges stay zero
- [ ] #678 — FastAPI `Route.path` empty and router prefix stripped
- [ ] #652 — NATS subject detection for cross-repo `ASYNC_CALLS` edges
- [ ] #523 — cross-repo-intelligence returns zero edges for byte-identical call/route
- [ ] #440 — cross-repo Maven library dependency links

### Framework extraction and false positives
- [ ] #636 — Windows non-ASCII repo paths make tree-sitter definitions fail on every file
- [ ] #606 — PHP framework/core method calls become false-positive project `CALLS` edges
- [ ] #603 — Django REST Framework ViewSet routes miss `HANDLES` mappings
- [ ] #600 — statement-scoped `orm_call_in_loop` signal for N+1 detection
- [ ] #599 — recursion detector matches by bare name and flags `super()` / wrapper calls
- [ ] #598 — route extractor over-fits GraphQL, config, and regex literals as Route nodes
- [ ] #700 — duplicate/evidence case: Windows accented path creates File nodes but no Function/Class extraction (see #636)

## Added 2026-06-30 (triage)
- [ ] #710 — Python deep chained call expression hangs definitions pass
- [ ] #706 — caller-side URL constants prevent `HTTP_CALLS` / cross-repo edges
- [ ] #721 — VB6 `.frm` / `.cls` extension collisions misclassify files as FORM/Apex
- [ ] #718 — VB/VB6 function bodies collapse to declaration-only spans

## Added 2026-07-01 (triage)
- [ ] #734 — Spring class-level `@RequestMapping` prefix dropped from emitted Route paths
- [ ] #733 — Kinesis stream producer/consumer detection for cross-repo async edges
- [ ] #730 — TypeScript `paths` aliases unresolved, causing fake dead-code results

## Added 2026-07-09 (triage)
- [ ] #759 — Next.js App Router: app/**/route.ts handlers produce 0 Route nodes
- [ ] #769 — Angular component sibling files (ts/html/scss) merged into one File node; search_code silently incomplete
- [ ] #871 — trace_path returns 0 callers for functions called via CommonJS require()
- [ ] #952 — PHP/Laravel: routes inside Route::prefix()->group() produce no Route nodes
- [ ] #953 — cross-repo-intelligence: gRPC route identities are project/namespace-prefixed, CROSS_GRPC_CALLS can never match
- [ ] #961 — C extractor: #ifdef-split-brace functions dropped (partially addressed: #960 fixed cbm_path_within_root; #949 in review recovers callables via the preprocessor pass; coverage signal #963 now detects the class)

(#946 C++ parse → PR #949; #964 include→IMPORTS → PR #983 — tracked in their PRs.)

## Added 2026-08-03 (triage)
- [ ] #1417 — semantic_query ranking lets nonsense outrank a relevant symbol
- [ ] #1410 — Vue `<script setup>` produces an empty symbol graph
- [ ] #1409 — Java 21 record-pattern instanceof extraction gap
- [ ] #1407 — C# overloads collapse because method identity omits signature/arity
- [ ] #1406 — fast mode silently excludes first-class source under scripts/
- [ ] #1405 — Gobra language-support request
- [ ] #1402 — multilingual semantic-query quality and fallback strategy
- [ ] #1401 — non-ASCII BM25 quality and missing docstring search content
- [ ] #1354 — TypeScript cross-file method calls lose CALLS edges

### LINKED ISSUE #599: [bug] `self_recursive` / `unguarded_recursion` match calls by bare name → `super().save()` and same-named wrappers are false positives
**Version:** codebase-memory-mcp v0.8.1
**Project:** real-world repo, 93,315 nodes

## Summary
The recursion detector appears to flag a function as self-recursive whenever its body contains a call whose **bare name** equals the function's own name — ignoring the receiver. This makes `super().method()` overrides and same-named delegating wrappers false positives. In our repo, **every** `unguarded_recursion=true` node also has `self_recursive=true`, and a large fraction are not recursive at all.

## Repro
`MATCH (f:Function) WHERE f.unguarded_recursion = true RETURN f.name, f.qualified_name, f.file_path, f.self_recursive`
(and the same for `:Method`). Representative false positives:

| node | actual body | why it's not recursion |
|---|---|---|
| `ConsumerAddress.save`, `ProductPriceHistory.save`, `PurchaseOrderV2.save` | `super().save(*args, **kwargs)` | calls the **parent** class method |
| `MPApiError.__init__`, `CloudAIExhausted.__init__` | `super().__init__(...)` | parent ctor |
| `StoreAdmin.get_queryset`, `MerchantCompanyAdmin.get_queryset` | `super().get_queryset(...)` | parent method |
| `*.setUp`, `DecimalEncoder.default` | `super().setUp()` / `super().default(o)` | parent method |
| `ApiClient.get/post/put` (TS) | `return this.request('GET', …)` / `axios.get(…)` | different receiver |
| `console`-style `error`/`warn` wrappers | `console.error(…)` | different receiver |
| minified bundle ids `e`,`Vo`,`Ii` (lines=1) | minified | not analyzable |

## Minimal example
```python
class Order(models.Model):
    def save(self, *a, **k):
        self.full_clean()
        super().save(*a, **k)   # parent call — flagged as self-recursion today
```
```ts
class ApiClient {
  get(url: string) { return axios.get(url); }   // flagged as self-recursion today
}
```

## Expected
- A call is self-recursion only when it resolves to the **same** function/method (same receiver / same class binding), not by bare name.
- Specifically: `super().X()` is a parent-class call and must never count as recursion of `X`.
- Calls on a different object (`axios.get`, `this.request`, `console.error`) must not count.

## Bonus (likely same subsystem)
`transitive_loop_depth` saturates to an impossible **174** on many unrelated functions (e.g. `whatsapp_webhook`, several `main`s) — looks like the interprocedural propagation doesn't break cycles and saturates. Worth checking alongside the recursion resolver.

## Impact
`unguarded_recursion` and `transitive_loop_depth` are currently unusable as-is for hot-path/risk triage on Python/TS codebases (super() overrides are everywhere).

lg320531124:
## Problem

Closes #599. Refs #592 (umbrella epic — not auto-closed).

The recursion detector flagged a function/method as self-recursive whenever its body contained a call whose **bare name** matched the function name — ignoring the receiver. This made `super().method()` overrides and same-named delegating wrappers false positives. In the reporter's 93k-node repo, **every** `unguarded_recursion=true` node also had `self_recursive=true`, and a large fraction were not recursive at all.

False positives from #599:

| node | body | why not recursion |
|---|---|---|
| `Order.save`, `ProductPriceHistory.save` | `super().save(*a, **k)` | calls the **parent** class method |
| `MPApiError.__init__` | `super().__init__(...)` | parent ctor |
| `ApiClient.get` (TS) | `return this.request(...)` / `axios.get(...)` | different receiver |
| `*.error` / `*.warn` wrappers | `console.error(...)` | different receiver |

## Root cause

`internal/cbm/cbm.c` stripped the callee to its short name with `strrchr(callee_name, '.')` and compared only that against the enclosing def name. `super().save()` → callee_name `"super.save"` → short `"save"` == def name `"save"` → flagged, even though `super` is the parent class, never `self`.

## Fix

Add `is_self_receiver()`: treat a call as self-recursion only when the callee resolves to the same instance/class.

- **bare names** (free function calling itself) → self-recursion (preserves prior behavior)
- **`self` / `this` / `cls` / `@self` receivers** → self-recursion (same object)
- **`super` receiver** → NOT self (parent class)
- **any other receiver** (`axios`, `console`, `this.request` on a different instance) → NOT self

The check is applied where the name comparison already happens, so all three signals stay in sync: `is_recursive`, `recursion_in_loop`, `unguarded_recursion`. No blanket suppression of the recursion flag — just receiver-aware classification, exactly as the maintainer specified in #599.

## Regression test

`complexity_receiver_aware_recursion` covers the maintainer-named cases:

1. `super().save()` inside `save` → **not** recursive (parent call); `self.save()` in the same method still trips it
2. `axios.get()` inside `get` → **not** recursive (different receiver)
3. `self.recur()` inside `recur` → recursive (same object), guarded → not unguarded

```
complexity_recursion_in_loop_unguarded   PASS
complexity_guarded_recursion             PASS
complexity_receiver_aware_recursion      PASS   ← new
5721 passed, 0 failed
```

## Scope

Two files, +90/-2. No changes to the call-graph extraction (callee_name already carries the receiver from `extract_calls.c`); only the classification predicate changes.

## Maintainer feedback (addressed)

Per @DeusData's review, four asks were addressed in commit `3718e09`:

1. **`Closes` → `Refs #592`** — #592 is the umbrella epic tracking several sub-issues, not a duplicate of #599. The PR body now reads `Refs #592 (umbrella epic — not auto-closed)` so merging this PR will not auto-close #592.
2. **Super()-only red-first fixture** — Added a `super().save()`-only fixture as the first assertion block of `TEST(complexity_receiver_aware_recursion)` asserting `ASSERT_FALSE(d->is_recursive)`. Verified red-first: reverting `internal/cbm/cbm.c` to `origin/main` (the unfixed name-only comparison at line 780) makes this assertion fail at `tests/test_extraction.c:3012` — the test is a real guard for the fix, not a tautology. Restoring the fixed `cbm.c` turns it green again.
3. **Stray `}` in Python fixture** — Removed the leftover `"}\n"` from the existing `super_call.py` fixture so the fixture source is well-formed.
4. **Go `s.save()` limitation acknowledged** — Go method-call receivers (`s.save()` inside `func (s *T) save()`) are not yet classified as self-recursion: the Go extractor does not surface the receiver on `callee_name` the way the Python/TS extractors do. This is documented here as a known limitation (receiver extraction for Go is future work) and locked with a Go-method fixture in `complexity_receiver_aware_recursion` that asserts the **current** behavior so any future improvement is a deliberate, test-visible change rather than a silent drift.

Full suite on the fixed code: **5736 passed, 0 failed** (exit 0). The only failure observed during development was the single targeted `complexity_receiver_aware_recursion` assertion on the deliberately-reverted unfixed `cbm.c`, confirming the test catches the bug.

Signed-off-by: lg320531124 <lg320531124@users.noreply.github.com>

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -438,6 +438,36 @@ static bool is_alloc_name(const char *n) {
     return name_in_set(n, set);
 }
 
+// Whether a callee expression targets the same instance/class as the enclosing
+// method, i.e. counts as genuine self-recursion rather than a same-named call on
+// a different receiver.  callee_name may be bare ("recur") or qualified
+// ("self.recur", "this.recur", "cls.recur", "super.recur", "axios.get", ...).
+//
+// Bare names have no receiver → assume self-call (preserves prior behavior and
+// the common case of a free function calling itself by bare name).
+// Qualified names: only self/this/cls/@self receivers are same-object; super()
+// is the parent class (never self), and any other receiver (axios, console,
+// this.request on a different instance, etc.) is a different target.  See #599.
+static bool is_self_receiver(const char *callee_name) {
+    if (!callee_name || !callee_name[0]) {
+        return false;
+    }
+    const char *dot = strchr(callee_name, '.');
+    if (!dot) {
+        return true; // bare name → self-recursion candidate
+    }
+    // Receiver is everything before the first '.'.
+    size_t rlen = (size_t)(dot - callee_name);
+    static const char *const self_receivers[] = {"self", "this", "cls", "@self", NULL};
+    for (int i = 0; self_receivers[i]; i++) {
+        size_t sl = strlen(self_receivers[i]);
+        if (rlen == sl && strncmp(callee_name, self_receivers[i], sl) == 0) {
+            return true;
+        }
+    }
+    return false; // super / axios / console / any other receiver
+}
+
 // Count parameters from a signature string like "(int a, Foo* b, cb (*)(int,int))".
 // Fallback for languages where param_names isn't populated (e.g. C keeps only the
 // signature text). Counts commas at the top paren level; treats "()"/"(void)" as 0.
@@ -771,12 +801,17 @@ CBMFileResult *cbm_extract_file(const char *source, int source_len, CBMLanguage 
             continue;
         }
         CBMDefinition *d = &result->defs.items[best];
-        // callee_name may be bare ("recur") or qualified ("pkg.recur", "self.recur")
+        // callee_name may be bare ("recur") or qualified ("self.recur",
+        // "super.recur", "axios.get").  Only treat as self-recursion when the
+        // callee resolves to the same instance/class — bare names or
+        // self/this/cls receivers.  super().X() and other-receiver calls
+        // (axios.get, console.error) must not count even when the method name
+        // matches (#599).
         const char *dot = strrchr(c->callee_name, '.');
         const char *callee_short = dot ? dot + 1 : c->callee_name;
         bool in_loop = c->loop_depth > 0;
 
-        if (strcmp(callee_short, d->name) == 0) {
+        if (strcmp(callee_short, d->name) == 0 && is_self_receiver(c->callee_name)) {
             // Direct self-recursion. The call graph omits self-edges (pass_calls
             // skips source==target), so detect it here; seeds "recursive".
             d->is_recursive = true;
--- a/tests/test_extraction.c
+++ b/tests/test_extraction.c
@@ -2993,18 +2993,40 @@ TEST(complexity_guarded_recursion) {
  * a method also named get) must not count either.  Bare names and self/this
  * receivers still count.  Guards the regression maintainer DeusData named. */
 TEST(complexity_receiver_aware_recursion) {
-    /* super().save() inside a method named save: parent-class call, NOT self. */
+    /* RED-FIRST GUARD: super().save() ONLY (no self.save()).  On the unfixed
+     * code the bare strcmp(callee_short, name) match flagged any same-named
+     * callee, so super().save() inside save() wrongly set is_recursive=true.
+     * super() targets the PARENT class — never self-recursion — so this must
+     * be false.  This is the headline guard the maintainer (DeusData) asked
+     * for: it fails on the unfixed codebase and only passes once the receiver
+     * whitelist is applied. */
     CBMFileResult *r = extract("class Repo:\n"
-                               "    def save(self):\n"
-                               "        if self.id:\n"
-                               "            super().save()\n"
-                               "        else:\n"
-                               "            self.save()\n"  /* genuine self-call */
-                               "}\n",
-                               CBM_LANG_PYTHON, "t", "super_call.py");
+                "    def save(self):\n"
+                "        if self.id:\n"
+                "            super().save()\n",
+                CBM_LANG_PYTHON, "t", "super_only.py");
     ASSERT_NOT_NULL(r);
     ASSERT_FALSE(r->has_error);
     const CBMDefinition *d = find_def(r, "save");
+    ASSERT_NOT_NULL(d);
+    ASSERT_FALSE(d->is_recursive);       /* super().save() is parent-class, NOT self */
+    ASSERT_FALSE(d->unguarded_recursion);
+    cbm_free_result(r);
+
+    /* Combined fixture: super().save() AND a genuine self.save() in the same
+     * method.  The self.save() branch is what triggers is_recursive here; the
+     * super().save() branch must NOT.  Kept as a complementary check to the
+     * super-only guard above. */
+    r = extract("class Repo:\n"
+                "    def save(self):\n"
+                "        if self.id:\n"
+                "            super().save()\n"
+                "        else:\n"
+                "            self.save()\n",  /* genuine self-call */
+                CBM_LANG_PYTHON, "t", "super_call.py");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    d = find_def(r, "save");
     ASSERT_NOT_NULL(d);
     ASSERT_TRUE(d->is_recursive);        /* self.save() still triggers it */
     ASSERT_FALSE(d->unguarded_recursion); /* both calls guarded by `if/else` */
@@ -3035,6 +3057,27 @@ TEST(complexity_receiver_aware_recursion) {
     ASSERT_NOT_NULL(d);
     ASSERT_TRUE(d->is_recursive);
     ASSERT_FALSE(d->unguarded_recursion); /* guarded by `if n > 0` */
+    cbm_free_result(r);
+
+    /* KNOWN LIMITATION — Go methods: receiver `s` in `func (s *Store) save()`
+     * is the Go-method receiver, the Go analogue of `self`/`this`.  But the
+     * self-receiver whitelist is {self, this, cls, @self} and does not include
+     * arbitrary receiver names like `s`, so `s.save()` inside `save()` is NOT
+     * flagged as self-recursion.  This test LOCKS the current (limiting)
+     * behavior so the trade-off is explicit; relaxing it later would require
+     * tracking the enclosing method's receiver name.  See PR #699. */
+    r = extract("package store\n\n"
+                "type Store struct{}\n\n"
+                "func (s *Store) save() {\n"
+                "    s.save()\n"
+                "}\n",
+                CBM_LANG_GO, "t", "go_method.go");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    d = find_def(r, "save");
+    ASSERT_NOT_NULL(d);
+    ASSERT_FALSE(d->is_recursive);      /* 's' is not in {self,this,cls,@self} */
+    ASSERT_FALSE(d->unguarded_recursion);
     cbm_free_result(r);
     PASS();
 }
--- a/tests/test_extraction.c
+++ b/tests/test_extraction.c
@@ -2987,6 +2987,58 @@ TEST(complexity_guarded_recursion) {
     PASS();
 }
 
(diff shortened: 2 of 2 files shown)


lg320531124:
Ready for review — all 14 CI checks green across the full platform matrix (ubuntu-latest/arm, macos-14/15-intel, windows CLANG64), plus dco/analyze/lint/CodeQL/security gates.

Closes #599 (and #592 dup). Implements the receiver-aware call classification you specified: `super().save()` and same-named wrappers on other receivers (`axios.get`, `console.error`) no longer trip `self_recursive`/`unguarded_recursion`, while genuine self-calls (bare names, `self`/`this`/`cls` receivers) still do. Regression test `complexity_receiver_aware_recursion` covers the `super()` case you named. Local: 5721 passed, 0 failed.

DeusData:

lg320531124:
Thanks for the acknowledgment — totally understand the review backlog. No rush on my end; happy to rebase or expand the test matrix whenever you're ready to take a look.

DeusData:
Thank you — the receiver-aware semantics are exactly what #599 asked for, and your axios.get test is a genuine red-first guard. Four small asks before merge:
1. Please change 'Closes #592' to 'refs #592' — #592 is the umbrella epic that merely lists #599; merging as-is would auto-close the whole epic.
2. The headline super() test is currently vacuous: the fixture contains both super().save() and self.save(), so is_recursive is true either way and it passes on unfixed code. A super()-only fixture asserting !is_recursive makes it a real guard.
3. There's a stray trailing '}' in the Python fixture (C-style artifact) — harmless but worth removing.
4. Trade-off to acknowledge (and ideally test): Go methods — func (s *Store) save() { s.save() } — extract as s.save; since 's' isn't in the whitelist, genuine self-recursion that main flags today stops being flagged. If per-language receiver-name comparison is out of scope here, a code comment + a test documenting the accepted trade-off is fine.
The approach itself is right — happy to merge with these addressed. Thanks!


---

lg320531124:
Thanks for the careful review, @DeusData — all four points addressed in commit `3718e09` (pushed to this branch). Test suite is green: **5736 passed, 0 failed**.

Point-by-point:

1. **#592 umbrella vs. duplicate** — Done. The PR body now reads `Closes #599. Refs #592 (umbrella epic — not auto-closed)`, so merging will not auto-close #592.

2. **Super()-only red-first fixture** — Added as the first assertion block of `TEST(complexity_receiver_aware_recursion)`: a `super().save()`-only fixture asserting `ASSERT_FALSE(d->is_recursive)` (`tests/test_extraction.c:3012`). Verified it is a real red-first guard: reverting `internal/cbm/cbm.c` to `origin/main` (the name-only comparison, no `is_self_receiver` guard) makes this assertion fail; restoring the fix turns it green. The test is not a tautology — it fails on the unfixed code.

3. **Stray `}` in Python fixture** — Removed. The `super_call.py` fixture is now well-formed.

4. **Go `s.save()` limitation** — Acknowledged. The Go extractor does not yet surface the receiver on `callee_name` the way the Python/TS extractors do, so `s.save()` inside `func (s *T) save()` is not currently classified as self-recursion. Documented as a known limitation in the PR body, and locked with a Go-method fixture in the same test asserting the **current** behavior — so any future improvement is a deliberate, test-visible change.

Full run on the fixed code: `complexity_receiver_aware_recursion PASS`, 5736 passed, 0 failed, exit 0. Appreciate the review.

DeusData:
Thank you — the receiver-aware semantics were exactly what #599 needed, and your axios.get test was a genuine red-first guard that carried over. We took it over the line as dac31d24 (PR #815) with you credited as co-author, with the review deltas folded in: receiver comparison now uses the whole chain via the last dot (so self.obj.recur() classifies correctly), the Go-method false negative got a real fix (the enclosing definition's receiver turned out to be reachable, so 's.save() inside save' is dynamically whitelisted rather than documented as a gap), a super()-only fixture that genuinely fails on the old code, and the epic reference kept as a plain ref so #592 stays open. Closes #599. Closing this in favor of the distill — thanks again!

Issue #670, opened 2026-06-28
search_code returns 0 matches for all patterns on macOS arm64 v0.8.1 (graph index healthy)


materwelonDhruv:
## Summary

On macOS arm64, `search_code` returns `total_grep_matches: 0` and an empty result for every pattern, including universal strings like `return` that appear thousands of times in any codebase. The graph tools (`search_graph`, `query_graph`, `index_status`, `list_projects`) all work correctly on the same index. The failure is silent, there is no error, just an empty result.

This looks like a regression of #250, the same silent-0 symptom on v0.6.x (root cause was `rg` invoked without an explicit path, so it read from stdin), which was fixed in v0.7.0.

## Environment

- codebase-memory-mcp 0.8.1 (darwin-arm64 static binary)
- macOS arm64, Darwin 25.5.0
- MCP host Claude Code, stdio transport

## Repro

1. Index a repository and confirm `index_status` reports ready with a real node count, for example `{ "status": "ready", "nodes": 9909, "edges": 18879 }`.
2. Confirm `search_graph` returns correct symbol results for that project.
3. Call `search_code` with any pattern that certainly exists, for example `return`, `Button`, `asChild`, or a known symbol name.
4. Every call returns `{ "total_grep_matches": 0, "raw_match_count": 0, "files": [] }`.
5. A plain `grep -rn "return" <repo>` returns thousands of hits, so the content is there.

## Expected

`search_code` returns matching lines with their graph context, consistent with what the graph tools confirm about the indexed content.

## Actual

`search_code` always returns `{ "total_grep_matches": 0, "raw_match_count": 0, "files": [] }`, with no error surfaced.

## Notes

- #250 (closed) tracked the identical symptom on v0.6.x and was fixed in v0.7.0 by giving the grep step an explicit indexed-file list instead of letting `rg` read stdin. The 0.8.1 text-search path appears to have lost that on macOS.
- #629 (open) is a separate problem, the 0.8.1 darwin-arm64 binary aborting with a dyld error on load. Here the binary loads fine and the graph tools run, only the text-search path is broken.

DeusData:
Thanks for the report. I’ve triaged this under #396 as a `search_code` backend regression separate from graph-index health. Next step is a darwin-arm64 repro that proves the grep/search path receives an explicit repository path and cannot silently return zero for universal patterns.

DeusData:
[...]

PR #701, opened 2026-06-29
fix(search): use xargs -0 for null-delimited filelist to handle space…

naman616:
…s in path (#670)

write_scoped_filelist wrote newline-delimited absolute paths; BSD xargs split each entry on whitespace, so a space in root_path (e.g. /Users/John Smith/...) caused grep to receive broken path fragments, silently producing zero matches.

Fix: write null-delimited paths (fputc '\0') and invoke xargs -0 so each path is passed as a single argument regardless of embedded spaces or metacharacters. Also add -H to the non-scoped grep -rn fallback so BSD grep always emits the filename prefix that collect_grep_matches requires.

Regression test: search_code_space_in_path_issue670 creates a project rooted at a directory containing a space, inserts an indexed node, and asserts that search_code returns a match — reproducing the exact reported failure scenario.

## Checklist

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -3663,18 +3663,19 @@ static void build_grep_cmd(char *cmd, size_t cmd_sz, bool use_regex, bool scoped
     const char *flag = use_regex ? "-E" : "-F";
     if (scoped) {
         if (file_pattern) {
-            snprintf(cmd, cmd_sz, "xargs grep -Hn %s --include='%s' -f '%s' < '%s' 2>/dev/null",
+            snprintf(cmd, cmd_sz, "xargs -0 grep -Hn %s --include='%s' -f '%s' < '%s' 2>/dev/null",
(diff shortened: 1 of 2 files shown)


DeusData:

DeusData:

DeusData:
Thank you for digging into #670 — your diagnosis of the newline-delimited filelist breaking on spaced paths was correct. While this was open, the same class got fixed on main via 3dd2471 (#751) plus 32bb56e (-Hn), with one crucial platform split your CI run surfaced the hard way: Windows' PowerShell Get-Content path can't consume NUL-delimited lists, which is why main keeps newline delimiting there — and why this branch's own Windows job went red. Since the Unix fix is already on main and the unconditional NUL can't ship as-is, we're closing this as superseded. Your instinct on the root cause was right, and we appreciate the regression test shape — the merged version guards the same scenario. Thanks again!

PR #703, opened 2026-06-29
fix(windows): use _wfopen for non-ASCII project paths

### LINKED ISSUE #700: Bug: File parsing fails silently on Windows when project path contains accented/non-ASCII characters

ShauryaaSharma:
Fixes #700

## What does this PR do?

On Windows, `fopen` uses the ANSI/system codepage and silently returns `NULL` when the file path contains non-ASCII characters (accented letters, CJK, etc.). The indexer's directory traversal already uses wide Win32 APIs (`FindFirstFileW`), so `File` nodes are created correctly, but the subsequent file-read/parse phase calls `fopen` with the UTF-8 path, which
fails silently. tree-sitter never receives the source, so no `Function`/`Class`/`Method` nodes are extracted for any file whose
absolute path contains a non-ASCII character.

Add `cbm_fopen(path, mode)` to `compat_fs`:
- **Windows**: converts the UTF-8 path to UTF-16LE via the existing  `cbm_utf8_to_wide` helper (`win_utf8.h`) and delegates to `_wfopen`.
- **POSIX**: plain `fopen` passthrough, no behaviour change.

Replace every `fopen` call that receives a user project file path:
`pass_parallel`, `pass_definitions`, `pass_calls`, `pass_semantic`,
`pass_usages`, `pass_lsp_cross`, `pass_pkgmap`, `pass_k8s`,
`pass_envscan`, `path_alias` (×2), `artifact`, `discover/language`,
`discover/gitignore`, `discover/discover`, `discover/userconfig`,
`mcp/mcp.c` (`read_file_lines` and `adr.md` reader), 18 call sites
across 17 files.

## Checklist

- [x] Every commit is signed off (`git commit -s`) : required, CI rejects unsigned commits.
- [ ] Tests pass locally (`make -f Makefile.cbm test`) : full suite requires ASan + 64-bit toolchain unavailable in this environment; the fix was verified with a targeted harness: created a real `cbm-test-césar\hello.py` fixture via Win32 wide-char APIs, confirmed plain `fopen` returns `NULL` for that path (bug reproduced), then confirmed `cbm_fopen` opens and reads it correctly. 7 cases including NULL path, missing file, write+round-trip, and non-ASCII in both directory and filename, all passing.
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) : clang-tidy unavailable in this environment; please run in CI
- [ ] New behavior is covered by a test (reproduce-first for bug fixes) : no test added to `tests/`; the affected code paths are Windows-only and not currently covered by the test suite

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -245,7 +245,7 @@ static bool expand_git_path(const char *path, char *out, size_t out_sz) {
 }
 
 static bool read_core_excludes_file(const char *config_path, char *out, size_t out_sz) {
-    FILE *f = fopen(config_path, "r");
+    FILE *f = cbm_fopen(config_path, "r");
     if (!f) {
         return false;
     }
--- a/src/discover/gitignore.c
+++ b/src/discover/gitignore.c
@@ -12,6 +12,7 @@
  *   - patterns with / are rooted (anchored to base)
  */
 #include "foundation/constants.h"
+#include "foundation/compat_fs.h"
 
 enum { GI_INIT_CAP = 16, GI_CHAR_IDX1 = 1, GI_CHAR_IDX2 = 2, GI_SKIP3 = 3 };
 #include "discover/discover.h"
@@ -287,7 +288,7 @@ cbm_gitignore_t *cbm_gitignore_load(const char *path) {
         return NULL;
     }
 
-    FILE *f = fopen(path, "r");
+    FILE *f = cbm_fopen(path, "r");
     if (!f) {
         return NULL;
     }
--- a/src/discover/language.c
+++ b/src/discover/language.c
@@ -11,6 +11,7 @@
 #include "cbm.h" // CBMLanguage, CBM_LANG_*
 
 #include "foundation/constants.h"
+#include "foundation/compat_fs.h"
 
 enum { LANG_SCAN_PASSES = 2 };
 #define SLEN(s) (sizeof(s) - 1)
@@ -998,7 +999,7 @@ CBMLanguage cbm_disambiguate_m(const char *path) {
         return CBM_LANG_MATLAB;
     }
 
-    FILE *f = fopen(path, "r");
+    FILE *f = cbm_fopen(path, "r");
     if (!f) {
         return CBM_LANG_MATLAB;
     }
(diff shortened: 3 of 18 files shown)


DeusData:

Issue #705, opened 2026-06-29
scripts/build.sh / make fail under the Nix flake dev shell on macOS; cross-arch builds unsupported


kriswill:
### Version

codebase-memory-mcp dev (built from source, main @ 7824e50)

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

### What happened, and what did you expect?

In the Nix flake dev shell (`nix develop`) on macOS, `scripts/build.sh` aborts before compiling anything: it prints `WARNING: clang (…) is x86_64 only — cannot build native arm64` and exits 1 — even though the Nix `clang` targets `arm64-apple-darwin` natively and builds fine. I expected `scripts/build.sh` (the "Build from Source" flow in CONTRIBUTING) to produce `build/c/codebase-memory-mcp`. The same false negative hits `test.sh`/`lint.sh`/`repro.sh` (shared `scripts/env.sh`).

Separately, a cross-architecture build (`scripts/build.sh --arch x86_64`) does not produce an x86_64 binary, and once the compiler check is bypassed the build fails to compile with `unknown type name 'git_allocator'`.

`nix build` works only because the flake's `buildPhase` already bypasses `build.sh` and calls `make` directly (with a comment acknowledging the `file`-check failure), so this is specific to the documented developer workflow, not the packaged build.

### Reproduction

Public repo: this repository (`DeusData/codebase-memory-mcp` @ `main`). Host: macOS, Apple Silicon. Toolchain: nixpkgs `clang-wrapper` 21.x in the flake dev shell (`CC=clang CXX=clang++`); `clang --version` → `Target: arm64-apple-darwin`.

1. `nix develop`
2. `scripts/build.sh`
   - Actual: aborts with `WARNING: clang (…) is x86_64 only — cannot build native arm64`, exit 1.
   - Expected: builds `build/c/codebase-memory-mcp` (native arm64).
3. Cross-arch: `scripts/build.sh --arch x86_64`
   - Actual: no working x86_64 binary (compiler check, then link failure — see Logs).
4. Bypass the check the way the flake does — `make -f Makefile.cbm cbm`:
   - Actual: compile error `internal/cbm/cbm.c:347: unknown type name 'git_allocator'` (only when `libgit2` is on the pkg-config path, i.e. inside the Nix shell).

### Logs

```
# 1) scripts/build.sh
=== build.sh: os=darwin arch=arm64 cores=10 cc=clang cxx=clang++ ===
[...]

PR #722, opened 2026-06-30
fix(build): include <git2/sys/alloc.h> for git_allocator

kriswill:
## What does this PR do?

Adds the missing `#include <git2/sys/alloc.h>` in `internal/cbm/cbm.c`.

The file uses `git_allocator` (to bind libgit2's allocator to mimalloc via `GIT_OPT_SET_ALLOCATOR`) but includes only `<git2.h>`, which does **not** pull in `<git2/sys/alloc.h>` where `git_allocator` is declared — it's a `sys/` header. When libgit2 is on the pkg-config path, the production build fails to compile:

```
internal/cbm/cbm.c:347: error: unknown type name 'git_allocator'
```

It was latent because the block is guarded by `CBM_BIND_TS_ALLOCATOR` (prod-only) **and** `HAVE_LIBGIT2`, so it compiles out unless libgit2 is present — which happens on a dev machine / the Nix dev shell with libgit2 installed, but not in CI. One-line fix.

Part of #705 (the macOS/Nix build report); split into its own narrow PR at the maintainer's request. `Refs #705` rather than `Fixes`, since #705 also covers the build-script and cross-arch work landing in separate PRs.

## Checklist

> **On the last three items:** the changed block is `CBM_BIND_TS_ALLOCATOR`-guarded (prod-only) and `HAVE_LIBGIT2`-gated, so it is compiled out of the ASan unit build — there is no runtime behavior to unit-test and `make test` does not exercise it. The fix is verified by compiling the **prod** binary with libgit2 on the pkg-config path (`scripts/build.sh` inside `nix develop`, or any host with `libgit2-dev`); `clang-format` is clean. No CI job currently builds with libgit2, which is why this regressed unseen — happy to add a libgit2 build-matrix entry in a follow-up if that would be useful.

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -20,7 +20,8 @@
 #if defined(CBM_BIND_TS_ALLOCATOR) && CBM_BIND_TS_ALLOCATOR
 #include "sqlite3.h" // sqlite3_mem_methods, sqlite3_config, SQLITE_CONFIG_MALLOC — bind sqlite to mimalloc
 #if defined(HAVE_LIBGIT2)
-#include <git2.h> // git_allocator, git_libgit2_opts, GIT_OPT_SET_ALLOCATOR — bind libgit2 to mimalloc
+#include <git2.h>           // git_libgit2_opts, GIT_OPT_SET_ALLOCATOR — bind libgit2 to mimalloc
+#include <git2/sys/alloc.h> // git_allocator — not pulled in by <git2.h> (it's a sys/ header)
 #endif
 #endif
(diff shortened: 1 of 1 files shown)


DeusData:

DeusData:
Thank you for the clean, correctly-commented fix — and for splitting it out of #705 so promptly. Verified against upstream libgit2 history: git_allocator has indeed always lived in the sys/ plumbing header, so your comment text is exactly right. This unblocks every libgit2-enabled build (Nix, Homebrew, MacPorts). Merging. The libgit2 ≥1.7 version-floor + CI-leg idea from your PR body is great — follow-up issue coming, and we'd be glad to have you on it.

PR #727, opened 2026-06-30
fix(graph-ui): keep filter sidebar visible when all nodes are filtered out

rarepops:
## Summary

In the Graph view, clicking **None** in the Filters panel (disable all filters) made the entire view, **including the filter sidebar itself** — collapse to a centered "All nodes filtered out / Reset Filters" message. With the sidebar gone, the only way back was **Reset Filters**, which re-enables *everything*; there was no way to toggle individual filters back on.

Root cause: an early `return` in `GraphTab` fired whenever `filteredData.nodes.length === 0` and replaced the whole layout (left sidebar included).

## Fix

- Narrow the early return to the genuinely-empty / no-data case (no data, or the project has zero nodes — where there are no filters to interact with anyway).
- Render the "All nodes filtered out" state **inside the graph area** instead, so the `FilterPanel` + `Sidebar` stay mounted. Users can now re-enable individual filters (or hit Reset Filters) without losing the panel.
- Side effect: a project that genuinely has zero nodes now correctly reads **"No nodes in this project"** rather than the misleading "All nodes filtered out".

## Before / After

**Before** — clicking *None* hides the sidebar and replaces the whole view:

Before_Fix

**After** — the filter sidebar stays; the message shows only in the graph area:

After_Fix

## Testing

- Added `graph-ui/src/components/GraphTab.filters.test.tsx`: disables all filters via the **None** shortcut and asserts the Filters sidebar (plus the All/None controls) remain in the document.
- `npx vitest run` → 9/9 passing.
- `tsc -b` → clean.

### DIFF
--- a/graph-ui/src/components/GraphTab.filters.test.tsx
+++ b/graph-ui/src/components/GraphTab.filters.test.tsx
@@ -1,0 +1,63 @@
+/* @vitest-environment jsdom */
+import "@testing-library/jest-dom/vitest";
+import { fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { GraphTab } from "./GraphTab";
+import type { GraphData } from "../lib/types";
+
+/* GraphScene renders a WebGL <Canvas> which jsdom can't run — stub it out. */
+vi.mock("./GraphScene", () => ({
+  GraphScene: () => null,
+  computeCameraTarget: () => null,
+}));
+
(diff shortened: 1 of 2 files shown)


DeusData:

DeusData:
Thank you — clean root-cause fix. The early return replacing the entire layout (sidebar included) was exactly the bug, and relocating the empty state into the graph pane while distinguishing 'no nodes in project' from 'all filtered out' is an accuracy improvement on top. Precise regression test too. Merging.

PR #729, opened 2026-06-30
fix(foundation): open UTF-8 repo paths with _wfopen on Windows

GrenAnHao:
## Summary

- Add portable `cbm_fopen()` in `compat_fs`: on Windows, UTF-8 paths are converted to UTF-16 and opened via `_wfopen`; on POSIX it delegates to `fopen`.
- Route project source file reads through `cbm_fopen` in the indexing hot path (`pass_parallel`, `pass_definitions`, `pass_semantic`, `pass_envscan`) and language disambiguation (`language.c`).

## Problem

On Windows with a non-UTF-8 ANSI code page (e.g. GBK / ACP 936), `fopen()` interprets path bytes in the system code page. When the repo absolute path contains non-ASCII characters (e.g. a CJK directory name), UTF-8 path bytes are misinterpreted and `fopen` fails.

Directory discovery already uses wide-char APIs (`cbm_opendir` / `FindFirstFileW`), so the file tree is built. Extraction passes, however, use `fopen` in `read_file()` and silently skip every source file. The resulting index contains only `File` / `Folder` / `CONTAINS_*` nodes — no symbols, no `IMPORTS` / `CALLS` edges, and `file_hashes` stays empty.

## Verification

Reproduced and fixed on Windows 11 (ACP 936):

| Metric | Before (`fopen`) | After (`cbm_fopen`) |
|--------|------------------|---------------------|
| Nodes | 61 (structure only) | 426 |
| Edges | 59 (`CONTAINS_*` only) | 1095 (`IMPORTS`, `CALLS`, `DEFINES`, …) |
| File→file IMPORTS edges | 0 | 12+ |
| File→file CALLS edges | 0 | 103 |

Minimal repro: `fopen(utf8_path_to_cjk_file)` returns NULL; `_wfopen(wide_path)` opens and reads successfully.

Built with official toolchain: MSYS2 CLANG64, `scripts/build.sh CC=clang CXX=clang++`.

## Test plan

- [x] Manual repro on Windows (ACP 936): CJK repo path indexes with symbols
- [x] `gcc -fsyntax-only` on all changed files (MSYS2 clang 22.1.7)
- [x] Production binary build via `scripts/build.sh`
- [ ] CI `scripts/test.sh` (maintainer / Linux CI)

## Notes

This is a focused bug fix (7 files, ~40 lines). No API surface or pipeline algorithm changes.

Integrators calling the CLI on Windows with non-ASCII paths in JSON args may still need ASCII `\uXXXX` escaping for argv (separate from this fix); MCP/stdio paths are unaffected.

### DIFF
--- a/src/discover/language.c
+++ b/src/discover/language.c
@@ -11,6 +11,7 @@
 #include "cbm.h" // CBMLanguage, CBM_LANG_*
 
 #include "foundation/constants.h"
+#include "foundation/compat_fs.h" // cbm_fopen
 
 enum { LANG_SCAN_PASSES = 2 };
 #define SLEN(s) (sizeof(s) - 1)
@@ -998,7 +999,7 @@ CBMLanguage cbm_disambiguate_m(const char *path) {
         return CBM_LANG_MATLAB;
     }
 
-    FILE *f = fopen(path, "r");
+    FILE *f = cbm_fopen(path, "r"); /* wide-open on Windows: handles non-ASCII (CJK) paths */
     if (!f) {
         return CBM_LANG_MATLAB;
     }
--- a/src/foundation/compat_fs.c
+++ b/src/foundation/compat_fs.c
@@ -188,6 +188,23 @@ int cbm_exec_no_shell(const char *const *argv) {
     return (int)_spawnvp(_P_WAIT, argv[0], argv);
 }
 
+FILE *cbm_fopen(const char *path, const char *mode) {
+    wchar_t *wpath = cbm_utf8_to_wide(path);
+    if (!wpath) {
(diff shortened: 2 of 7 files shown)


DeusData:

DeusData:
Thank you @GrenAnHao — same correct diagnosis and the same `_wfopen` approach (and honestly the cleanest, most locale-independent mode conversion of the three PRs).

We're landing #703, which converts **every** file-read site. This one covers 5 but misses `pass_calls`, so that path would still break on non-ASCII. Closing as a duplicate in favor of the complete fix — really appreciate the contribution. 🙏

Issue #720, opened 2026-06-30
py_lsp: py_eval_expr_type infinite recursion — Python missed in 9f981a1 depth-guard fix


Rakesh-207:
## Bug

`py_eval_expr_type` in `py_lsp.c` enters unbounded recursion on certain Python expression patterns, hanging the indexer at 100% CPU indefinitely. The process never terminates and must be killed manually.

## Root Cause

Commit **9f981a1** (*"Bound LSP resolve recursion depth"*) added depth guards to three LSP resolvers to prevent exactly this class of bug:

| Language | File | Guard |
|----------|------|-------|
| Java | `java_lsp.c` | `walk_depth` cap 512 |
| C/C++ | `c_lsp.c` | `walk_depth` cap 512 |
| TypeScript | `ts_lsp.c` | `member_depth` cap 64 |
| **Python** | **`py_lsp.c`** | **— none —** |

Python was missed. The commit message itself notes: *"The expression-eval guards (eval_depth) did not cover these walks."*

## Stack Trace (from `sample`)

The worker thread spins in a self-referential call chain with no depth cap:

```
extract_worker
  → cbm_extract_file
    → cbm_run_py_lsp
      → py_lsp_process_file
        → py_resolve_calls_in
          → py_process_statement
            → py_eval_expr_type          ← recurses here
              → py_eval_expr_type
                → py_eval_expr_type
                  → py_eval_expr_type     ← 10+ levels, never terminates
```

The main thread is blocked on `_pthread_join` waiting for this worker. Memory grows to ~8 GB from the deep recursion stack. No log output after the file starts extracting — it just burns CPU forever.

## Trigger

A single ~6,000-line Python file with complex runtime code (PyTorch tensors, MediaPipe, GStreamer WebRTC, Modal decorators, deeply chained method calls). The file is valid, production Python. I'm not able to share the source, but the pattern that triggers it involves deeply nested method-call chains on dynamically-typed objects where the type resolver can't find a base case.

## Reproduction

1. Index a repo containing a large Python file (~6K lines) with complex expression nesting
2. The parallel extraction phase reaches the file and hangs
3. `sample <pid>` shows the recursive `py_eval_expr_type` chain above
4. Process must be killed manually

## Environment

- **Version**: 0.8.1 (latest, darwin/arm64)
- **OS**: macOS 26.1 (Apple Silicon)
- **Memory budget**: 4 GB (default), 8 GB RAM system
- **Workers**: 8

[...]

PR #732, opened 2026-07-01
Fix(memorization) issue 710 py lsp memoization

### LINKED ISSUE #710: Python parser hangs on deeply chained method calls during definitions pass
## Describe the bug

When indexing a Python project, the `index_repository` tool hangs indefinitely during the `definitions` pass. The process consumes massive CPU (800+ seconds) and memory (~1.4 GB) without producing any result, eventually hitting the MCP request timeout (300s).

## To Reproduce

Create a Python file containing a deeply chained method call like the following (LangGraph `StateGraph` builder pattern):

```python
def build_graph(saver):
    graph = (
        StateGraph(state_schema=State)
        .add_node("node_a", handler_a)
        .add_node("node_b", handler_b)
        .add_node("node_c", handler_c)
        .add_node("node_d", handler_d)
        .add_node("node_e", handler_e)
        .add_node("node_f", handler_f)
        .add_node("node_g", handler_g)
        .add_node("node_h", handler_h)
        .add_node("node_i", handler_i)
        .add_node("node_j", handler_j)
        .add_node("node_k", handler_k)
        .add_node("node_l", handler_l)
        .add_node("node_m", handler_m)
        .add_node("node_n", handler_n)
        .add_edge(START, "node_a")
        .add_conditional_edges(
            "node_a", route_fn,
            {
                "node_b": "node_b",
                "node_c": "node_c",
                "node_d": "node_d",
                "node_error": "node_error",
            })
        .add_conditional_edges(
            "node_b", interrupt_fn,
            {
                "node_interrupt": "node_interrupt_b",
                "node_next": "node_e",
            })
        .add_conditional_edges(
            "node_c", interrupt_fn,
            {
                "node_interrupt": "node_interrupt_c",
                "node_next": "node_f",
            })
        .add_conditional_edges(
            "node_d", interrupt_fn,
            {
                "node_interrupt": "node_interrupt_d",
                "node_next": "node_g",
            })
        .add_edge("node_interrupt_b", "node_b")
        .add_edge("node_interrupt_c", "node_c")
        .add_edge("node_interrupt_d", "node_d")
        .add_edge("node_e", END)
        .add_edge("node_f", END)
        .add_edge("node_g", END)
        .add_edge("node_error", END)
        .compile(checkpointer=saver, debug=True)
    )
    return graph
```

Then run:

```
codebase-memory-mcp cli index_repository '{"repo_path": "/path/to/project"}'
```

The pipeline output stops at:

```
level=info msg=pass.start pass=definitions files=N
```

and never proceeds to the next pass.

## Binary-search isolation

I isolated the issue via binary search on the project files:

| Test case | Result |
|-----------|--------|
| Single-file Python project (`def hello()`) | 170 ms, OK |
| 5 small Python files (<100 lines each) | 508 ms, OK |
| 8 Python files (excluding the problematic file) | 616 ms, OK |
| Adding the file with `build_graph` (14 chained calls + 4 conditional edges) | **Hangs at definitions pass** |
| Same file, first 50 lines only (no `build_graph`) | OK |
| Same file, first 120 lines (includes `build_graph`) | **Hangs** |
| Same file, `build_graph` body replaced with a comment | OK |

The hang is specifically triggered by the deeply nested method chain pattern (`StateGraph(...).add_node(...).add_node(...).....add_conditional_edges(...).add_edge(...).....compile(...)`) with ~65 levels of nesting.

## Expected behavior

The parser should either:
1. Handle deeply chained method calls without hanging, or
2. Detect the timeout/complexity and skip the file with a warning rather than blocking the entire indexing pipeline.

## Environment

- **codebase-memory-mcp version**: 0.8.1
- **OS**: Windows 10/11 (x64)
- **Total RAM**: 16 GB
- **Memory budget reported by server**: 8055 MB

## Additional context

- The `structure` pass completes fine (nodes and edges are extracted correctly).
- Only the `definitions` pass hangs — suggesting the tree-sitter Python parser enters a pathological state on this AST pattern.
- The same project indexes successfully when the problematic function body is removed or the chain is broken into separate assignment statements.
- Workaround: break the chain into intermediate variables, e.g. `g = g.add_node(...)` repeated on separate lines.

LA-10:
## What does this PR do?

Fixes #710 index_repository hangs during the definitions pass on Python files with deeply chained method calls, consuming 800+ s and ~1.4 GB with no progress.

py_eval_expr_type (internal/cbm/lsp/py_lsp.c) recursively evaluates expression types with no memoization. Chained calls cause the same sub-chains to be re-evaluated at every level roughly O(2^N) in chain depth

Memoize py_eval_expr_type, keyed by node start byte. Each sub-expression is typed once, collapsing chain evaluation to O(depth). The original body is renamed _uncached and wrapped by a thin memoizing entry results identical, only redundant recomputation removed. Cache is arena-allocated per file (freed automatically).

Added pylsp_issue710_deep_call_chain_no_hang to tests/test_py_lsp.c a ~40-link chain that hangs on the old code and passes on the fix.

## Checklist

### DIFF
--- a/internal/cbm/lsp/py_lsp.c
+++ b/internal/cbm/lsp/py_lsp.c
@@ -1330,10 +1330,10 @@ static void py_type_cache_grow(PyLSPContext *ctx) {
     CBMPyTypeCacheEntry *old_entries = ctx->type_cache;
 
     int new_cap = old_cap ? old_cap * 2 : PY_TYPE_CACHE_INITIAL_CAP;
-    CBMPyTypeCacheEntry *new_entries =
-        (CBMPyTypeCacheEntry *)cbm_arena_alloc(ctx->arena, (size_t)new_cap * sizeof(CBMPyTypeCacheEntry));
+    CBMPyTypeCacheEntry *new_entries = (CBMPyTypeCacheEntry *)cbm_arena_alloc(
+        ctx->arena, (size_t)new_cap * sizeof(CBMPyTypeCacheEntry));
     if (!new_entries)
-        return; 
+        return;
     memset(new_entries, 0, (size_t)new_cap * sizeof(CBMPyTypeCacheEntry));
 
     ctx->type_cache = new_entries;
@@ -1396,8 +1396,6 @@ static const CBMType *py_eval_expr_type(PyLSPContext *ctx, TSNode node) {
     py_type_cache_insert(ctx, key, result);
     return result;
 }
-
-
 
 static void py_process_statement(PyLSPContext *ctx, TSNode node) {
     if (!ctx || ts_node_is_null(node))
@@ -1877,8 +1875,8 @@ static void py_emit_call_for(PyLSPContext *ctx, TSNode call_node) {
                     // Skip if mod is already rooted under the project to avoid
                     // "<root>.<root>.mod".
                     if (!(strncmp(mod, ctx->module_qn, root_len) == 0 && mod[root_len] == '.')) {
-                        char *qual_mod = (char *)cbm_arena_alloc(ctx->arena, root_len + 1 +
-                                                                                strlen(mod) + 1);
+                        char *qual_mod =
+                            (char *)cbm_arena_alloc(ctx->arena, root_len + 1 + strlen(mod) + 1);
                         if (qual_mod) {
                             memcpy(qual_mod, ctx->module_qn, root_len);
                             qual_mod[root_len] = '.';
--- a/internal/cbm/lsp/py_lsp.c
+++ b/internal/cbm/lsp/py_lsp.c
@@ -26,6 +26,7 @@
 // Forward decls
 static void py_resolve_calls_in(PyLSPContext *ctx, TSNode node);
 static const CBMType *py_eval_expr_type(PyLSPContext *ctx, TSNode node);
+static const CBMType *py_eval_expr_type_uncached(PyLSPContext *ctx, TSNode node);
 static void py_process_statement(PyLSPContext *ctx, TSNode node);
 static const CBMRegisteredFunc *py_lookup_attribute(PyLSPContext *ctx, const char *type_qn,
                                                     const char *member_name);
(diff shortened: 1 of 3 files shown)


DeusData:

DeusData:
Thank you — your diagnosis is the correct one for #710, and we verified it independently: py_eval_expr_type evaluates each attribute receiver twice (container special-case + general attribute path), giving O(2^N) re-evaluation on chained calls; the recently-merged walker caps (#779) did not and could not address this. Memoization is the right generalized fix. One blocking correctness issue before merge:
- The cache key (node start byte) is not unique — every node shares its start byte with its leftmost descendants (a, a.b, a.b(), a.b().c() all collide). Insert is first-writer-wins, so the leftmost leaf's type permanently occupies the slot and composite results are silently dropped → wrong or missed CALLS edges on ordinary chained calls, repo-wide and silently. Please key on node identity (TSNode.id) or (start_byte, end_byte, symbol).
- Smaller: if arena alloc persistently fails, the insert probe loop can spin once the table is 100% full — a bounds check avoids a hang-path inside a hang fix.
- Test ask: add a heterogeneous-receiver chain (client.session().execute()-style) asserting the resolved QN — exactly the case the key collision breaks; your current fixture can't see it because G and G() evaluate to the same type.
With those, this is the primary fix for #710 (and quite possibly #720) and we'd like to land it first in the py-eval series. Thanks again for the rigorous write-up.


---

DeusData:
Thank you — your diagnosis WAS the fix for #710: we independently verified the double receiver evaluation gives O(2^N) re-evaluation on chained calls, and memoization is the right cure. We carried it over the line as 90f59c51 (PR #809) with you credited as co-author, with one correction from the review: the cache key is node identity (TSNode.id) rather than start byte — leftmost descendants share start bytes, and a probe experiment confirmed the collision silently mis-resolves heterogeneous chains like client.session().execute(). The distill also adds a scope-generation flush (lambda re-walks under different bindings would otherwise serve stale types), the depth/steps guards from #758, and never-caches truncated results. #710 closes — it would be great if you could retest on main. Closing this in favor of the distill — thanks again for the rigorous write-up!

LA-10:
Confirmed resolved on main (90f59c5). Built from source and re-ran the build_graph StateGraph chain repro through index_repository. On the pre-fix release binary the definitions pass pins a core and runs to timeout (killed at 6.4 CPU-s); on the fixed build the same input completes cleanly 
pass=definitions returns in ~0.1s CPU and node/edge counts as expected. Verified on Intel macOS. Thanks for carrying it over!

PR #666, opened 2026-06-28
Fix watcher stale root cleanup

pcristin:
## Summary
- Track consecutive missing project roots in the watcher.
- Preserve projects through transient missing-root checks, then prune after three consecutive misses.
- Delete cached project DB/WAL/SHM files when sustained absence is confirmed.
- Add a regression test covering transient miss preservation and sustained cleanup.

Fixes #286.

## Validation
- `make -f Makefile.cbm build/c/test-runner` passed.
- `git diff --check` passed.

PR #738, opened 2026-07-01
fix(watcher): prune sustained missing roots

### LINKED ISSUE #286: Feature Request: Automatic deletion of project when underlying folder is deleted
Using work-trees with this tool has been fabulous because the AI can easily just add the worktree as a project and start using it, including any changes then that it makes.
But, as one uses work-trees and changes and then removes them those projects stay requiring manual deletion. 

My feature request is to also allow automatic deletion when the watcher determines that the root folder for a project is gone so that when work trees are cleaned up, the projects are also automatically cleaned up.

Thx.

pcristin:
## What does this PR do?

Implements the watcher stale-root cleanup discussed in #286:

- Track consecutive missing project roots in the watcher.
- Preserve projects through transient missing-root checks, then prune after sustained absence.
- Delete cached project DB/WAL/SHM files when sustained absence is confirmed.
- Cover the transient-miss and sustained-cleanup behavior with a regression test.

Fixes #286.
Supersedes #666 after refreshing the branch/title/commit metadata to match the contribution guide.

## Checklist

- [x] Every commit is signed off (`git commit -s`) — required, CI rejects unsigned commits.
- [ ] Tests pass locally (`make -f Makefile.cbm test`).
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`).
- [x] New behavior is covered by a test (reproduce-first for bug fixes).

## Validation

- `bash scripts/check-no-test-skips.sh`
- `git diff --check`
- `build/c/test-runner` was run after rebasing; the watcher regression passed. The full local runner still had unrelated local/environment failures.

### DIFF
--- a/src/watcher/watcher.c
+++ b/src/watcher/watcher.c
@@ -22,6 +22,7 @@
 #include "foundation/compat.h"
 #include "foundation/compat_thread.h"
 #include "foundation/compat_fs.h"
+#include "foundation/platform.h"
 #include "foundation/str_util.h"
 
 #include <stdio.h>
@@ -30,6 +31,7 @@
 #include <time.h>
 #include <stdatomic.h>
 #include <sys/stat.h>
+#include <errno.h>
 
 /* ── Per-project state ──────────────────────────────────────────── */
 
@@ -39,6 +41,7 @@ typedef struct {
     char last_head[CBM_SZ_64]; /* git HEAD hash */
     bool is_git;               /* false → skip polling */
     bool baseline_done;        /* true after first poll */
+    int missing_root_count;    /* consecutive polls where root_path was absent */
     int file_count;            /* approximate, for interval calc */
     int interval_ms;           /* adaptive poll interval */
     int64_t next_poll_ns;      /* next poll time (monotonic ns) */
@@ -69,6 +72,8 @@ struct cbm_watcher {
 #define POLL_BASE_MS 5000
 #define POLL_FILE_STEP 500 /* add 1s per this many files */
 #define POLL_MAX_MS 60000
+/* Sustained absence window before deleting an irreversible cached DB. */
+#define MISSING_ROOT_DELETE_AFTER ((3 * 60 * CBM_MSEC_PER_SEC) / POLL_BASE_MS)
 
 /* Sleep chunk for responsive shutdown (ms) */
 #define SLEEP_CHUNK_MS 500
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for reopening this as a clean replacement for #666. DCO and CI are green, and the PR now has a cleaner title/body/history. I’ve added the same triage labels and moved this into the maintainer review queue.

DeusData:
Thank you — this is the right mechanism for #286 and complements #537 (just merged) perfectly: event-driven unwatch for explicit deletion, your polling prune for silently vanished roots; neither covers the other's case. Two changes before merge, both about the delete being irreversible:
1. Count only ENOENT/ENOTDIR as 'missing' — today any stat failure qualifies, so an EACCES/EIO blip (network mount, macOS TCC revocation, ejected volume) is indistinguishable from deletion.
2. Lengthen the window: 3 consecutive polls can be ~15 seconds, and the .db contains the user-authored ADR — unrecoverable content. #286's thread explicitly asked for tolerance of transient unmounts; a minutes-scale sustained-absence threshold would honor that.
Also heads-up: you'll need a small rebase (test_watcher.c moved under #537). Optionally route the prune through #537's new deferred-free list for a single freeing discipline. Thanks again — the reset-on-reappearance handling is nicely done.


---

pcristin:
Thanks for the detailed review. I pushed the requested update in 0d1f60d:

- rebased onto current main after #537
- only ENOENT/ENOTDIR now count as missing; inaccessible/unavailable roots reset the missing streak and do not prune
- raised the prune threshold to a 3-minute sustained absence window (36 base polls)
- routed prune removal through the #537 deferred-free path
- added/updated watcher regression tests for sustained missing roots and inaccessible roots

Local validation is green for `git diff --check`, `bash scripts/check-no-test-skips.sh`, and `make -f Makefile.cbm build/c/test-runner`; the watcher section passes in the full runner.

DeusData:
Thank you — this was the right mechanism for #286 and the perfect complement to the unwatch-on-delete fix: polling prune is the only way to catch silently vanished roots. We carried it over the line as 836f90de (PR #817) with you credited as co-author, with the data-loss hardening from the review folded in: only ENOENT/ENOTDIR count as missing (an EACCES/EIO blip on a network mount resets the streak instead of deleting the DB), pruning additionally requires a 10-minute sustained-absence grace window (env-tunable), and the prune's entry removal now routes through the deferred-free discipline the unwatch fix introduced — one consistent freeing model. Your restore-reset behavior carried over as-is. Closes #286. Closing this in favor of the distill — thanks again!

PR #739, opened 2026-07-01
fix(graph-ui): fix Windows folder navigation in the index picker

rarepops:
## Summary

Fixes several issues with the repository folder picker (the **New Index** modal) on **Windows**. All changes are frontend-only (`graph-ui`) and Windows-scoped, so POSIX (Linux/macOS) navigation is left byte-for-byte unchanged, guarded by an `isWinPath` check.

On Windows the picker was effectively unusable: clicking breadcrumb segments errored with *"not a directory"*, switching drives didn't update the listing, and the `/` quick-jump button led to an empty view.

Screenshot 2026-07-01 160803

## Changes

- **Breadcrumb navigation.** On a Windows drive path (`C:/Users/x`) the breadcrumb built click targets as `"/" + segments` (`/C:/Users`), which the backend rejects as *"not a directory"*. It now builds drive-aware targets (`C:/`, `C:/Users`) and drops the meaningless leading `/` crumb on Windows.
- **Typed-path refresh.** Typing a path (e.g. `D:/`) updated the input and breadcrumb but never re-fetched the directory listing, so switching drives by typing showed the previous drive's folders. Added a debounced, silent `/api/browse` refresh (plus Enter-to-navigate), scoped to Windows drive paths. This matters because older backends don't enumerate drives, making typing the only way to switch drives.
- **Drive quick-jump.** On Windows the POSIX `/` quick-jump root is meaningless (browsing it returns an empty listing). It now offers drive roots instead, always including the current drive (derived from the browsed path), so it works even when the backend returns `roots: ["/"]`.

## Testing

- `graph-ui` unit tests: **12/12 passing**, including new regression tests for Windows breadcrumb navigation, typed-drive refresh, `/` to drive-root replacement, and a POSIX-negative test asserting a typed POSIX path does **not** trigger a re-browse.
- Manually verified in the running UI on Windows: breadcrumb clicks, typing `D:/`, and the drive-root quick-jump all list the correct folders.

## Notes

- Backend (`src/ui/http_server.c`) is unchanged; these fixes work against existing/older builds that return `roots: ["/"]`.
- The **"Project name" field removal** that was previously part of this PR has been split into a separate PR (#805) per review request, so this PR is now scoped purely to the four Windows navigation commits.

### DIFF
--- a/graph-ui/src/components/StatsTab.test.tsx
+++ b/graph-ui/src/components/StatsTab.test.tsx
@@ -158,6 +158,35 @@ describe("StatsTab index modal", () => {
     expect(await screen.findByText("projects")).toBeInTheDocument();
     expect(screen.queryByText("Documents")).not.toBeInTheDocument();
   });
+
+  it("replaces the meaningless '/' root with the drive on Windows", async () => {
+    const fetchMock = mockProjectsFetch((url) => {
+      if (url.startsWith("/api/browse")) {
+        const m = /[?&]path=([^&]*)/.exec(url);
+        const path = m ? decodeURIComponent(m[1]) : "C:/Users/rap";
+        return new Response(JSON.stringify({
+          path,
+          parent: "C:/",
+          dirs: ["Documents"],
+          roots: ["/"], // older backend: no drive enumeration
+        }), { status: 200, headers: { "Content-Type": "application/json" } });
+      }
+      return undefined;
+    });
+
+    render(<StatsTab onSelectProject={() => {}} />);
+    fireEvent.click(await screen.findByRole("button", { name: "Index your first repository" }));
+
+    /* The bogus "/" quick-jump is gone; the current drive root is offered. */
+    expect(await screen.findByRole("button", { name: "Browse C:/" })).toBeInTheDocument();
+    expect(screen.queryByRole("button", { name: "Browse /" })).not.toBeInTheDocument();
+
+    /* Clicking it browses to the drive root, not "/". */
+    fireEvent.click(screen.getByRole("button", { name: "Browse C:/" }));
+    await waitFor(() => {
+      expect(fetchMock).toHaveBeenCalledWith("/api/browse?path=C%3A%2F");
+    });
+  });
 });
 
 describe("IndexProgress", () => {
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the Windows picker fix. Triage: this is a Windows graph UI indexing bug, priority high, because drive/root navigation can block indexing from the UI.

For review, we will focus on drive-root breadcrumb targets, typed-path refresh, `/` behavior on Windows, preserving POSIX behavior, and the regression tests around the index picker/i18n changes. This is in the maintainer review queue.

DeusData:
Thank you — the Windows navigation work is exactly what remained of #548 after the backend drive-enumeration landed: drive-aware breadcrumbs, typed-drive refresh, roots handling, and the POSIX-negative test is appreciated. One structural ask: commit 61884ca removes the 'Project name' field entirely — that's a cross-platform feature removal and an API-shape/product decision (the backend still accepts project_name), and it sits in a PR whose body says the changes are Windows-scoped. Your rationale for removing it is genuinely reasonable, but it deserves its own PR where it can be discussed as a product question rather than riding along. Could you split that commit out and keep the four navigation commits here? Small heads-up: you may need to rebase the test-file import line once the indexing-error PR lands (it touches the same line). With the split, this merges quickly — thanks again!


---
**Update:** to get the navigation fixes into the next release, if the split doesn't happen in the next couple of days we'll cherry-pick the four navigation commits ourselves (your authorship preserved on each) and leave the project-name-field question to its own discussion. Whichever is easier for you!

rarepops:
@DeusData sure no problem! I didn't think it was needed honestly, would be good to have a rename for a project or have the name different from the path etc. I will make the change :)

rarepops:
https://github.com/DeusData/codebase-memory-mcp/pull/805

DeusData:

DeusData:
Thank you for the quick split — exactly as hoped, and the four navigation commits here are the ones our review greenlit. One last mechanical step: the branch now conflicts with main on the StatsTab test file (the indexing-error surfacing landed yesterday and touched the same import line). A quick rebase/merge of main and this is ready to merge. The product question about the Project-name field continues in #805 on its own merits. Thanks again!

DeusData:
Thank you — and thanks for the fast turnaround on both the split and the rebase. The four navigation fixes are exactly what remained of the Windows picker work: drive-aware breadcrumbs, typed-path refresh (correctly scoped to drive paths), and the root handling, with the POSIX-negative test giving confidence nothing changes off-Windows. Merging. The Project-name discussion concluded as keep-but-relabel (landing separately, credited to you) — thanks for pushing that conversation with #805.

PR #740, opened 2026-07-01
feat(graph-ui): URL routing, filter panel UX, and camera fly-to fix

rarepops:
## Summary

Improvements to the graph UI: the left filter panel is reworked, the active tab and selected project are now reflected in the URL, the camera fly-to no longer snaps back to centre, and a few confusing labels are clarified.

All changes are scoped to `graph-ui/`. No server or protocol changes.

### Filter panel and folder tree
- Cap the filter panel height and move the node/edge chips into a scroll area, so a long list of types no longer pushes the folder tree off screen.
- Regroup the chips under **Node types** and **Relationships**, and pin the **Show labels** toggle as a footer.
- Add a **Folders** heading (en/zh) above the folder tree so that section is clearly labelled.

### URL routing
- The active tab and selected project are stored in the URL query string (`?tab=...&project=...`), so the view survives refreshes and can be bookmarked or shared. Back/forward is handled via `popstate`.
- Query params are used rather than path segments because the embedded HTTP server only serves `index.html` at `/` (path routes would 404 on refresh).

### Graph tab and HUD
- The **Graph** tab is disabled until a project is selected, and opening the **Projects** tab clears the active project.
- The empty-state hint now points at the **Projects** tab (it previously referenced a "Stats" tab that no longer exists).
- The ambiguous **Clear** button in the graph HUD is renamed **Clear selection** so it reads distinctly from **Refresh**.

### Camera fly-to fix
- Focusing a node or cluster only moved the camera and called `lookAt`, leaving the `OrbitControls` pivot at the origin. `OrbitControls` re-centred the view on the next frame, snapping the camera back once the animation ended.
- The controls target is now lerped to the focus point as well, so the view stays on the clicked node and orbits around it afterwards.

## Screenshots

- **Filter panel (grouping):**
image

- **URL routing (address bar showing ?tab / project):** image

## Testing

- `npm run build` (tsc typecheck + vite build): passing
- `npm test` (vitest): 8/8 passing
- Manual: deep-link `/?tab=graph&project=<name>` restores the view; the Projects tab clears the selection and disables the Graph tab; browser back/forward round-trips; the camera stays on the focused node.

## Follow-up (not in this PR)

- Persisting the active filter selection (disabled node/edge types plus show-labels) in the URL.

### DIFF
--- a/graph-ui/src/App.tsx
+++ b/graph-ui/src/App.tsx
@@ -1,14 +1,62 @@
-import { useState } from "react";
+import { useCallback, useEffect, useState } from "react";
 import { GraphTab } from "./components/GraphTab";
 import { StatsTab } from "./components/StatsTab";
 import { ControlTab } from "./components/ControlTab";
 import type { TabId } from "./lib/types";
 import { useUiMessages } from "./lib/i18n";
 
+const TAB_IDS: TabId[] = ["graph", "stats", "control"];
+
+interface RouteState {
+  tab: TabId;
+  project: string | null;
+}
+
+/* Read the active tab + selected project from the URL query string so the
+ * current view survives refreshes and can be bookmarked or shared. */
+function readRoute(): RouteState {
+  const params = new URLSearchParams(window.location.search);
+  const rawTab = params.get("tab");
+  const tab = TAB_IDS.includes(rawTab as TabId) ? (rawTab as TabId) : "stats";
+  const project = params.get("project");
+  return { tab, project: project ? project : null };
+}
+
+/* Build the canonical URL for a route, preserving the path and hash. */
+function routeUrl(tab: TabId, project: string | null): string {
+  const params = new URLSearchParams();
+  params.set("tab", tab);
+  if (project) params.set("project", project);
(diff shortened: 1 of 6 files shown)


DeusData:
Thanks for the graph UI work. Triage: this is an enhancement/UX PR, priority normal, scoped to `graph-ui/`.

For review, the main things we will check are URL/query-state behavior across refresh and browser back/forward, selected-project encoding, i18n coverage for the new labels, and whether the camera fly-to / filter-panel changes have regression coverage. This is in the maintainer review queue now.

DeusData:
Thank you — this is exactly the kind of graph-ui polish that makes the next release feel better: dependency-free URL routing (native URLSearchParams + history, correctly whitelisting the tab param — the right call for the embedded SPA rather than pulling in react-router), the filter-panel scroll/grouping cleanup, the honest label fixes, and the camera fly-to lerp fix with a clear root-cause note. Clean per-commit split, green on every platform incl. Windows, no new deps. Merged as 9987a4ea. (If you ever feel like peeling the camera fix into its own PR next time it keeps the history even cleaner, but the split here made it trivial to review.) Appreciated!

PR #502, opened 2026-06-18
Add Mojo language support

Tokarzewski:
## What does this PR do?

Adds **Mojo** (Modular's Python-superset systems language) wiring — the enum, extraction spec, language entry, and tests — so the integration point is ready for the grammar to be re-vendored after provenance audit.

> **Provenance (grammar — not yet vendored in this PR):**
> - **Grammar:** [lsh/tree-sitter-mojo](https://github.com/lsh/tree-sitter-mojo)
> - **Pinned commit:** `33193a99afe6`
> - **License:** MIT (forked from tree-sitter-python)
> - **ABI version:** 15 (compatible with current runtime ceiling)
> - **Scanner type:** C (no libstdc++ dependency)
> - **Registries:** Not listed in nvim-treesitter or Helix — community grammar
> - **Regeneration:** `tree-sitter generate` from the pinned commit (standard workflow)
> - **Tracking issue:** #737

### Spec design
The grammar's node types mirror Python's, so the spec **reuses the `py_*` arrays** and overrides only the class types (same reuse pattern already used for CFScript→`js_*`). Mojo-specific divergences:

- `fn`/`def` → `function_definition`
- `struct`/`class` → `class_definition`; `trait` and `__extension` get their own nodes (`trait_definition` / `extension_definition`), so traits map to the **Interface** label
- compile-time `alias NAME = value` has no dedicated grammar node — the upstream grammar recovers it as an `assignment` (name still captured)

### What's in this PR (wiring only)
- `CBM_LANG_MOJO` enum (appended, no renumbering of persisted DBs)
- Extraction spec in `lang_specs.c`
- `.mojo` and `.🔥` extensions + `LANG_NAMES` in `language.c`
- `scripts/new-languages.json` entry
- Tests: `test_grammar_regression.c` case (`fn`→Function, `struct`→Class, `trait`→Interface) and the matching `test_grammar_labels.c` golden
- `MANIFEST.md` entry (marked PENDING — vendored files removed pending audit)
- `THIRD_PARTY.md` and README language count

### What's NOT in this PR (deferred)
- Vendored grammar files (`internal/cbm/vendored/grammars/mojo/`) — removed from this PR per maintainer request. To build, copy the C parser + scanner from the pinned commit above into that directory and re-add `grammar_mojo.c` to the build.

### Verification (original branch, before removal)
Indexed real Mojo corpora end-to-end:
[...]

PR #744, opened 2026-07-01
chore(grammar): vendor Mojo tree-sitter parser

DeusData:
Refs #502

## Summary
- vendor the Mojo tree-sitter C parser/scanner from `lsh/tree-sitter-mojo` at `33193a99afe6`
- add the normal one-translation-unit wrapper so `tree_sitter_mojo` can link when Mojo support is wired
- document Mojo provenance, MIT license, ABI 15 status, and the Helix source caveat in the grammar manifest

## Security / provenance
- vendored only `parser.c`, `scanner.c`, `tree_sitter/*.h`, and `LICENSE`
- did not vendor package manager hooks, workflows, generated lockfiles, or prompt/agent instruction files
- local scan found no file, process, or network API usage in the vendored C surface

## Validation
- `build/c/test-runner` was attempted locally; this checkout is not fully green locally. The retained failures are outside this vendor change: existing Kotlin label/C-LSP read-only guards plus local HTTP listener and DNS clone setup failures in the sandbox.

### DIFF
--- a/THIRD_PARTY.md
+++ b/THIRD_PARTY.md
@@ -26,7 +26,7 @@ The core runtime headers in `internal/cbm/vendored/common/tree_sitter/`
 
 ## Tree-sitter Grammars
 
-160 pre-generated parsers are vendored in `internal/cbm/vendored/grammars/<lang>/`
+159 pre-generated parsers are vendored in `internal/cbm/vendored/grammars/<lang>/`
 (generated `parser.c` plus `scanner.c` where applicable, compiled statically).
 Each grammar is the work of its upstream authors and each grammar directory
 contains the upstream `LICENSE` file.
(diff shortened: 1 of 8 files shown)


DeusData:
Maintainer security/provenance note after merge:

- The vendored Mojo grammar is pinned to `lsh/tree-sitter-mojo@33193a99afe6d0dbe865d56f6e7514c4087f87b2`.
- Upstream reports MIT licensing, and the vendored copy includes the MIT license/provenance metadata.
- CI license/provenance gates passed, including the vendored dependency integrity checks.
- Static review of the vendored parser sources did not find process execution, outbound network behavior, environment exfiltration, sensitive-path file IO, or telemetry-style code paths. The only URL-like reference found was a docs URL used in scanner handling.
- This is still a community grammar rather than an official Modular grammar, so future grammar refreshes should go through the same license, provenance, and static security review before merge.

Based on those checks, I consider this acceptable for vendoring in the project.

PR #747, opened 2026-07-01
docs: add real-world bash/ops-tooling repo example to Performance

saddestmartian:
## What does this PR do?

Adds a small "Real-World Example" subsection to the Performance section of the README.

The existing Performance benchmarks (Linux kernel, Django) are measured on large polyglot application codebases. Bash already scores in the "Excellent" parsing tier under Language Support, but there's no worked example of what that translates to for a different repo shape: infra/ops-tooling repos that are mostly shell scripts, YAML/JSON config, and Markdown docs rather than a typical multi-language application.

This adds one real session's measurement on an internal dev-tooling repo (bash + JSON + Markdown, ~14.9k indexed nodes / ~20.5k edges): a single `search_code` call reproducing a config-file consumer inventory in ~375ms vs. a grep-based sub-agent fan-out taking ~131s across 12 tool calls for the same result. It's explicitly labeled as one real-world data point, not a controlled multi-trial benchmark, to keep it honest alongside the hardware-benchmarked numbers above it.

Docs-only change — no source touched.

## Checklist

- [x] Tests pass locally (`make -f Makefile.cbm test`) — N/A, docs-only change, no C source touched
- [x] Lint passes (`make -f Makefile.cbm lint-ci`) — N/A, docs-only change, no C source touched
- [x] New behavior is covered by a test (reproduce-first for bug fixes) — N/A, not new behavior, a README addition

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,14 @@ Benchmarked on Apple M3 Pro:
 
 **Token efficiency**: Five structural queries consumed ~3,400 tokens via codebase-memory-mcp versus ~412,000 tokens via file-by-file grep exploration — a **99.2% reduction**.
 
+### Real-World Example: A Bash/Config-Heavy Ops-Tooling Repo
+
+The benchmarks above are measured on large polyglot application codebases (Linux kernel, Django). It's worth showing what "Excellent"-tier Bash parsing (see [Language Support](#language-support)) translates to for a different repo shape: infra/ops-tooling repos that are mostly shell scripts, YAML/JSON config, and Markdown docs — not a typical multi-language application.
+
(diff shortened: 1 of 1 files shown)


DeusData:
Hey, looks good. Can you remove the last line? Basically embedd the insights naturally :)

DeusData:
Thank you, @saddestmartian — a genuinely thoughtful contribution, and you framed it exactly right (honest that it's a single real-world session, not a controlled benchmark). The technical points hold up too: Bash is an Excellent-tier language here, and shell functions are indexed as first-class graph nodes.

We're keeping the README **Performance** section limited to benchmarks we can reproduce and stand behind end-to-end (the M3 Pro figures, the measured token-efficiency comparison) — readers treat that section as authoritative, and we can't reproduce a third-party session's exact numbers. That's the *only* reason we're not merging; it's about keeping that section reproducible, not the quality of your work.

If you'd like to share the data point, a GitHub Discussion would be a great home for it and genuinely useful to teams evaluating the tool for ops/bash-heavy repos. Really appreciate the effort. 🙏

PR #752, opened 2026-07-02
Tier RAM budget by host size

Andy11-cpu:
## Summary

- Use 25% of physical RAM on hosts with <=16GB, 35% on <=32GB, and 50% above
- Add `cbm_mem_ram_fraction_for_total()` with unit tests

## Motivation

Split from #625 — reduces default memory reservation on 16GB machines without changing indexing behavior.

## Test plan

- [x] `mem_ram_fraction_*` tier tests in `test_mem.c`

### DIFF
--- a/src/foundation/mem.c
+++ b/src/foundation/mem.c
@@ -106,6 +106,21 @@ static void check_pressure(size_t rss) {
 
 /* ── Public API ────────────────────────────────────────────────── */
 
+#define RAM_FRACTION_DEFAULT 0.5
+#define RAM_FRACTION_16GB 0.25
+#define RAM_FRACTION_32GB 0.35
+#define RAM_BYTES_PER_GB (1024ULL * 1024 * 1024)
+
+double cbm_mem_ram_fraction_for_total(size_t total_ram_bytes) {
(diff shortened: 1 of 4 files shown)


DeusData:
Thanks for the RAM-budget split. Triage: high-priority stability/performance work from the memory-pressure track.

Review will check that the tiered default is conservative on 16GB machines, does not regress larger-host throughput unexpectedly, and composes correctly with explicit memory-budget overrides. Please keep any benchmark or before/after memory observations in the PR description as it evolves.

DeusData:

DeusData:
Thank you — this is a clean, well-judged fix: tiering the RAM fraction by host size (25% ≤16GB / 35% ≤32GB / 50% above) so small and containerized hosts don't over-commit, and it correctly reads the cgroup-aware total so a memory-limited container gets the right tier. Atomic, green on all platforms, with genuine boundary tests. Merged as a960d3d2. It composes cleanly with the CBM_MEM_BUDGET_MB explicit override (override still wins) and the retention layer's re-read fallback covers the smaller-budget case with no lost edges. Appreciated — and thanks for splitting this out from the earlier bundle so it could land on its own merits!

PR #755, opened 2026-07-02
fix(mcp): prefilter scoped search_code paths

yutianyu111602-glitch:
## Summary
- Apply `path_filter` while writing the indexed file list used by scoped `search_code` grep.
- Keep the existing post-grep path filter as a defensive fallback.
- Normalize indexed Windows paths before regex matching so filter semantics stay aligned with result paths.

## Verification
- `git diff --check`
- `wsl.exe bash -lc 'cd /mnt/c/code/githubstar/codebase-memory-mcp && make -f Makefile.cbm cbm -j2'`

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -4259,9 +4259,12 @@ static void classify_all_grep_hits(grep_match_t *gm, int gm_count, cbm_store_t *
     }
 }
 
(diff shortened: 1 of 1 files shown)


yutianyu111602-glitch:
Superseded by #756, which carries the same patch with a DCO Signed-off-by trailer.

PR #756, opened 2026-07-02
fix(mcp): prefilter scoped search_code paths

yutianyu111602-glitch:
## Summary
- Apply `path_filter` while writing the indexed file list used by scoped `search_code` grep.
- Keep the existing post-grep path filter as a defensive fallback.
- Normalize indexed Windows paths before regex matching so filter semantics stay aligned with result paths.

## Verification
- `git diff --check origin/main...HEAD`
- `wsl.exe bash -lc 'cd /mnt/c/code/githubstar/codebase-memory-mcp && make -f Makefile.cbm cbm -j2'`

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -4259,9 +4259,12 @@ static void classify_all_grep_hits(grep_match_t *gm, int gm_count, cbm_store_t *
     }
 }
 
-/* Write indexed file list for scoped grep. Returns true if scoped. */
+/* Write indexed file list for scoped grep. Returns true if scoped.
+ * When path_filter is provided, apply it before grep so large indexed projects
+ * do not scan unrelated files only to discard them later. */
 static bool write_scoped_filelist(cbm_mcp_server_t *srv, const char *project, const char *root_path,
-                                  const char *filelist) {
+                                  const char *filelist, bool has_path_filter,
+                                  cbm_regex_t *path_regex) {
     cbm_store_t *pre_store = resolve_store(srv, project);
     if (!pre_store) {
         return false;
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks for the scoped `search_code` path-filter fix. Triage: UX/search correctness bug.

This PR is currently conflicting, so the first step is a rebase on current `main`. After that, review will check that prefiltering and the existing post-filter agree across POSIX and Windows-normalized paths, and that scoped searches do not accidentally widen access.

DeusData:
Thanks — we verified this is correct-by-construction: the prefilter applies the same compiled regex to the same relative path string as the post-grep filter, so it can only remove work, never change results, and nothing new reaches the shell. Two asks: (1) rebase onto current main — the filelist loop moved under 3dd2471 (#751), so your filter-continue now belongs at the top of the fwrite loop; (2) two small tests: a scoped search whose hit survives the prefilter, and a path_filter matching zero indexed files returning a clean zero result (that second one grazes the old 'grep reads stdin on empty input' class and deserves an explicit guard). Then we're happy to merge — this is genuinely useful on large indexed projects. Thanks!


---

DeusData:
Thank you for this — the prefilter idea was correct-by-construction (same compiled regex, same relative path as the post-grep filter, so it can only remove work, never change results), and it's genuinely valuable on large indexed projects. Since the branch predated the #751 filelist rewrite and had no tests, we carried it over the line in a distilled follow-up rebased onto the new writer, with regression tests and a deterministic empty-filelist guard (grep now skipped entirely when the prefilter yields zero entries — measured 6ms→0ms), crediting you as co-author on the commit. Merged as 185451ad (PR #819). Closing this one in its favor — thanks again for the contribution!

PR #757, opened 2026-07-02
fix(mcp): tighten ingest traces schema for strict clients

### LINKED ISSUE #731: Request: Add support for Visual Studio GitHub Copilot for codebase-memory-mcp
## Problem Description

When using `codebase-memory-mcp` as an MCP server with GitHub Copilot in **Visual Studio** (not VS Code), the server is recognized and appears in the MCP list, but any query that invokes it fails with a schema validation error. The error indicates that the function `codebase_memory_ingest_traces` does not include `"additionalProperties": false` as required by Visual Studio's Copilot client.

This is a **compatibility issue specific to Visual Studio**. The same server works perfectly with other MCP clients such as Trae and Cursor, so it is likely that Visual Studio's Copilot enforces a stricter JSON Schema validation.

## Steps to Reproduce

1. Install `codebase-memory-mcp` (e.g., via `npm install -g codebase-memory-mcp` or using the Windows installer).
2. Configure the MCP server in Visual Studio using the built-in MCP configuration UI:
   - Go to **Tools → Options → MCP Servers** (or use the MCP Server Register menu in Copilot Chat).
   - Add a new server with the following settings:
     - **Server ID**: `codebase-memory`
     - **Type**: `stdio`
     - **Command**: `codebase-memory-mcp` (this works if the binary is in your PATH)
   - The UI generates the following configuration automatically:
{
"inputs": [],
"servers": {
"codebase-memory": {
"type": "stdio",
"command": "codebase-memory-mcp",
"args": [],
"env": {}
}
}
}
- Alternatively, you can manually create a `.vs/mcp.json` file in your solution root with the same content.
3. Open a solution and start a Copilot Chat session in **Agent mode**.
4. Ask a question that would trigger the MCP tools, e.g., "Index this repository" or "Show me the call hierarchy of class X".
5. Observe the error in the Copilot output window.

## Actual Result

Copilot Chat returns an error and the chat becomes unusable. The detailed error log shows:
Microsoft.VisualStudio.Copilot.Service: CopilotFailure
exception_type: Microsoft.VisualStudio.Copilot.Diagnostics.Exceptions.CopilotApiException
exception_message: Invalid schema for function 'codebase_memory_ingest_traces': In context=(properties',traces','items'), 'additionalProperties' is required to be supplied and to be false.
http_status_code: 400
inner_exception: System.ClientModel.ClientResultException
inner_exception_message: HTTP 400 (: invalid_request_body) Invalid schema for function 'codebase_memory_ingest_traces': In context=(properties',traces','items'), 'additionalProperties' is required to be supplied and to be false.

## Expected Result

The MCP server should function correctly with Visual Studio's GitHub Copilot, allowing users to query codebase memory without errors. The schema of all exposed functions should comply with Visual Studio's validation requirements.

## Additional Context

- **Works on**: Trae, Cursor, and likely VS Code (with appropriate extension).
- **Fails on**: Visual Studio 2022 (version 17.12 or later, with MCP support enabled).
- The error is specific to the function `codebase_memory_ingest_traces` – possibly other functions also need the same fix.
- The error suggests that Visual Studio requires `"additionalProperties": false` to be explicitly set in the schema for object properties, which may not be required by other MCP clients.
- The configuration above is generated directly by Visual Studio's UI, not manually written, so the configuration format itself is correct.

## Suggested Fix

Update the tool definitions (especially the schema for `codebase_memory_ingest_traces`) to include `"additionalProperties": false` in the appropriate locations (e.g., within the `properties` object for `traces` and its `items`). This would make the server fully compliant with Visual Studio's stricter validation.

## Environment

- **OS**: Windows 11
- **Visual Studio**: 2022 (Enterprise/Community) with MCP support
- **GitHub Copilot**: Latest version (built-in)
- **codebase-memory-mcp**: v0.8.1 (latest as of writing)

## Additional Notes

If you need more logs or testing, I am happy to provide them. This is a significant blocker for users who rely on Visual Studio as their primary IDE. Thank you for maintaining this great project!

### Proposed solution

We kindly request that codebase-memory-mcp be made compatible with Visual Studio's GitHub Copilot, so that users of that IDE can also benefit from this tool.

### Alternatives considered

_No response_

### Confirmations

dpersek:
## Summary
- Declare supported `ingest_traces` item fields in the generated MCP schema.
- Set `traces.items.additionalProperties` to `false` for strict clients such as Visual Studio Copilot.
- Add a parsed `tools/list` regression test covering the nested schema path.

## Why
Fixes #731.

Visual Studio Copilot rejects `codebase_memory_ingest_traces` because the nested trace item object schema does not explicitly set `additionalProperties: false`. The schema now closes that object while still advertising the existing trace payload shape: `caller`, `callee`, and optional `count`.

## Tests
- `build/c/test-runner mcp` - passed, 122 tests.
- `git diff --check` - passed.
- `make -f Makefile.cbm test` - built and ran, but failed outside the MCP path with local listener/offline clone environment failures.
- `make -f Makefile.cbm lint-ci` - blocked locally because `cppcheck` is not installed.

## Scope
- No runtime trace ingestion behavior changes.
- No tool names or handler dispatch changes.
- No manual Visual Studio smoke test was performed.

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -491,7 +491,8 @@ static const tool_def_t TOOLS[] = {
 
     {"ingest_traces", "Ingest traces", "Ingest runtime traces to enhance the knowledge graph",
      "{\"type\":\"object\",\"properties\":{\"traces\":{\"type\":\"array\",\"items\":{\"type\":"
-     "\"object\"}},\"project\":{\"type\":"
+     "\"object\",\"properties\":{\"caller\":{\"type\":\"string\"},\"callee\":{\"type\":\"string\"},"
+     "\"count\":{\"type\":\"integer\"}},\"additionalProperties\":false}},\"project\":{\"type\":"
      "\"string\"}},\"required\":[\"traces\",\"project\"]}"},
 };
 
--- a/tests/test_mcp.c
+++ b/tests/test_mcp.c
@@ -247,6 +247,62 @@ TEST(mcp_tools_array_schemas_have_items) {
     PASS();
 }
 
+TEST(mcp_ingest_traces_items_disallow_additional_properties_issue731) {
+    char *json = cbm_mcp_tools_list();
+    ASSERT_NOT_NULL(json);
+
+    yyjson_doc *doc = yyjson_read(json, strlen(json), 0);
+    ASSERT_NOT_NULL(doc);
+    yyjson_val *root = yyjson_doc_get_root(doc);
+    ASSERT_NOT_NULL(root);
+    yyjson_val *tools = yyjson_obj_get(root, "tools");
+    ASSERT_NOT_NULL(tools);
+    ASSERT_TRUE(yyjson_is_arr(tools));
+
+    yyjson_val *tool;
+    yyjson_arr_iter iter;
+    yyjson_arr_iter_init(tools, &iter);
+    yyjson_val *ingest_traces = NULL;
+    while ((tool = yyjson_arr_iter_next(&iter)) != NULL) {
+        yyjson_val *name = yyjson_obj_get(tool, "name");
+        if (name && yyjson_is_str(name) && strcmp(yyjson_get_str(name), "ingest_traces") == 0) {
+            ingest_traces = tool;
+            break;
+        }
+    }
+    ASSERT_NOT_NULL(ingest_traces);
+
+    yyjson_val *input_schema = yyjson_obj_get(ingest_traces, "inputSchema");
+    ASSERT_NOT_NULL(input_schema);
+    yyjson_val *properties = yyjson_obj_get(input_schema, "properties");
+    ASSERT_NOT_NULL(properties);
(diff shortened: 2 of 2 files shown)


DeusData:
Thanks for the strict-client schema fix for #731. Triage: MCP/editor integration compatibility bug.

Review will check the generated `tools/list` schema shape for strict clients and ensure the actual `ingest_traces` runtime behavior is unchanged. Nested `additionalProperties:false` is useful, but it needs to stay aligned with the advertised trace fields.

DeusData:
Thanks for the precise fix — we verified ingest_traces was the only bare propertyless object schema in the entire TOOLS array, which matches VS complaining exactly there, and your parsed tools/list test is a genuine guard. The declared caller/callee/count fields now become the contract for the future real implementation, which is a nice bonus. Merging — we'll ask the reporter on #731 to verify in VS.

PR #758, opened 2026-07-02
fix(py-lsp): cap expression evaluation depth

### LINKED ISSUE #720: py_lsp: py_eval_expr_type infinite recursion — Python missed in 9f981a1 depth-guard fix

dpersek:
## Summary
- Add a Python expression-evaluation depth guard with graceful `unknown` fallback.
- Track evaluator recursion depth in `PyLSPContext`.
- Add a fork-protected regression for deeply nested Python expressions.

## Why
Fixes #720.

Python was missing the eval-depth protection already present in sibling LSP implementations. Deeply nested Python expressions now stop at a bounded depth instead of continuing recursive type evaluation indefinitely.

## Tests
- `build/c/test-runner stack_overflow` - passed, 14 tests.
- `build/c/test-runner py_lsp py_lsp_stress` - passed, 102 tests.
- `git diff --check` - passed.

## Scope
- No redesign of Python type inference or call resolution.
- No changes to other language LSP caps.
- Full test/lint were not rerun; local broad proof is environment/tooling-limited as noted above.

### DIFF
--- a/internal/cbm/lsp/py_lsp.c
+++ b/internal/cbm/lsp/py_lsp.c
@@ -23,9 +23,12 @@
  * only from py_lsp.c, never compiled standalone. */
 #include "py_builtins.c"
 
+#define PY_LSP_MAX_EVAL_DEPTH 64
+
 // Forward decls
 static void py_resolve_calls_in(PyLSPContext *ctx, TSNode node);
 static const CBMType *py_eval_expr_type(PyLSPContext *ctx, TSNode node);
+static const CBMType *py_eval_expr_type_inner(PyLSPContext *ctx, TSNode node);
 static void py_process_statement(PyLSPContext *ctx, TSNode node);
 static const CBMRegisteredFunc *py_lookup_attribute(PyLSPContext *ctx, const char *type_qn,
                                                     const char *member_name);
@@ -693,6 +696,18 @@ static const CBMType *py_iterable_element_type(PyLSPContext *ctx, const CBMType 
 }
 
 static const CBMType *py_eval_expr_type(PyLSPContext *ctx, TSNode node) {
+    if (!ctx || ts_node_is_null(node))
+        return cbm_type_unknown();
+    if (ctx->eval_depth >= PY_LSP_MAX_EVAL_DEPTH)
+        return cbm_type_unknown();
+
+    ctx->eval_depth++;
+    const CBMType *result = py_eval_expr_type_inner(ctx, node);
+    ctx->eval_depth--;
+    return result ? result : cbm_type_unknown();
+}
+
+static const CBMType *py_eval_expr_type_inner(PyLSPContext *ctx, TSNode node) {
     if (!ctx || ts_node_is_null(node))
         return cbm_type_unknown();
 
--- a/internal/cbm/lsp/py_lsp.h
+++ b/internal/cbm/lsp/py_lsp.h
@@ -69,6 +69,9 @@ typedef struct {
     int dict_literal_count;
     int dict_literal_cap;
 
+    // Expression evaluator recursion depth guard.
+    int eval_depth;
+
     // Debug mode (CBM_LSP_DEBUG env, shared across all language LSPs).
     bool debug;
 } PyLSPContext;
--- a/tests/test_stack_overflow.c
+++ b/tests/test_stack_overflow.c
@@ -487,6 +487,27 @@ TEST(lsp_cpp_deep_expression_no_crash) {
     PASS();
 }
 
+TEST(lsp_python_deep_expression_no_crash) {
+    /* Deep parenthesized expressions force repeated py_eval_expr_type
+     * recursion through assignment RHS inference. The evaluator should hit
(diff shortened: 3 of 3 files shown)


DeusData:
Thanks for the Python depth guard. Triage: high-priority parser stability fix for #720.

Review will check that deeply nested expressions degrade to `unknown` without recursion blowup, that the depth counter is reset correctly across evaluations, and that sibling LSP behavior remains unchanged. The fork-protected regression is the right shape for this class of bug.

DeusData:
Thanks — the gap is real and distinct from #779: the expression evaluator is reachable at unbounded depth regardless of the walker cap, and Python is indeed the only LSP without an eval guard. Wrapper/inner shape matches the house pattern nicely. Three changes before merge:
1. Cap 64 is too low: legit chains cost ~2 levels per link, so #710's ~65-link real-world chain needs ~130 — everything beyond the cap degrades to unknown even after memoization lands. c_lsp uses 256; please match it.
2. The test is currently a false guard: DEPTH=256 nested parens ≈ ~512 stack frames and passes on unfixed main. The sibling deep-nesting tests in the same file all use DEPTH=30000 — please match, so the test is red-first.
3. Optional but ideal: port the other half of the C guard too (per-file eval_steps budget) — the depth cap alone doesn't stop breadth blowups.
Sequencing: #732's memoization is the actual hang fix and both PRs rewrap the same function, so we'd like to land #732 first and have this rebase on top as the outer guard (and please make sure truncated results are never cached). We'll ask the #720 reporter to retest once both are in. Thanks!


---

dpersek:
Thanks, this approach works for me. I’m happy for #809 to carry the distilled fix forward with the co-author credit.

DeusData:
Thanks — the gap was real and distinct from the walker caps: the expression evaluator was the one Python path with no guard, and your wrapper/inner shape matched the house pattern. We carried it over the line as 90f59c51 (PR #809) with you credited as co-author, with the review deltas: cap raised to 256 (matching c_lsp — 64 truncated legitimate ~65-link chains), a per-file eval-steps budget (the other half of the C guard), the deep-expression test at DEPTH=30000 so it genuinely fails on unguarded code, and composition with the memoization from #732 such that truncated results are never cached. #720 closes — a retest on main would be very welcome. Closing in favor of the distill — thanks again!

Issue #308, opened 2026-05-01
TypeScript path alias resolution (`tsconfig.json` paths) produces zero graph edges in monorepo


muczynski-persefoni:
## Environment
- **codebase-memory-mcp version**: 0.6.0
- **Codebase**: TypeScript monorepo (~10,400 files) with Cloudflare Workers, Electron app, shared packages
- **Package manager**: bun
- **OS**: macOS

## Summary

When indexing a TypeScript monorepo that uses `tsconfig.json` path aliases (e.g., `@first-recon/worker-shared`), cross-package imports produce **zero CALLS and IMPORTS edges** in the graph. This makes the tool return confidently wrong results for any cross-package dependency analysis.

### Setup
Our monorepo has a shared package at `packages/worker-shared/` with a `tsconfig.json` path alias `@first-recon/worker-shared`. Multiple workers import from it:

```typescript
// In workers/chat-worker/circuit-breaker.ts
import { generateCorrelationId } from '@first-recon/worker-shared';
```

### Expected Behavior
```cypher
MATCH (f:Function)-[:CALLS]->(g:Function {name: 'generateCorrelationId'})
RETURN f.name, f.qualified_name
```
Should return callers from all packages that import and call this function.

### Actual Behavior
**0 results.** Zero CALLS edges, zero IMPORTS edges cross package boundaries.

Verified with ripgrep: `rg "generateCorrelationId" --type ts -l` finds **14 files** across 8+ workers that import and use this function. The graph sees none of them.

### Confirming query
```cypher
MATCH (m:Module)-[:IMPORTS]->(n) WHERE n.name CONTAINS 'correlation' RETURN count(*)
```
→ 0 results.

## Impact

This is not an edge case — in a monorepo architecture, cross-package imports via path aliases are the **default pattern**. Every worker importing from shared packages is invisible to the graph.

Practical consequences:
- **Impact analysis** reports zero consumers for widely-used shared utilities
- **Dead code detection** would flag active code as unused
- An AI agent using this tool for impact analysis would systematically undercount callers and could recommend breaking changes to shared interfaces

## Additional Issues Found During Evaluation

While testing, we also found these related issues:

### 1. `n.file` and `n.path` properties return empty strings
File location must be reverse-engineered from `n.qualified_name` by stripping the project prefix and replacing dots with `/`. This is undocumented and breaks for files with dots in their names (e.g., `some.config.ts`).

### 2. Barrel file re-exports not followed
[...]

PR #766, opened 2026-07-02
fix(pipeline): resolve ../ climbs in tsconfig path alias targets

### LINKED ISSUE #730: codebase-memory graph it doesn't resolve TypeScript aliases like @packages/shared
### Version

codebase-memory-mcp 0.8.1

### Platform

macOS (Apple Silicon)

### Install channel

### Binary variant

standard

The codebase-memory graph resolves imports by direct file path, but it doesn't resolve TypeScript aliases like @packages/shared → ../../packages/shared/src/index.ts. This generates a massive amount of fake dead code for everything imported via alias. I did the tests using codebase-memory-mcp from opencode

### Reproduction

The codebase-memory graph resolves imports by direct file path, but it doesn't resolve TypeScript aliases like @packages/shared → ../../packages/shared/src/index.ts. This generates a massive amount of fake dead code for everything imported via alias. I did the tests using codebase-memory-mcp from opencode

### Logs

```text

```

```text

```

_No response_

### Confirmations

apappas1129:
## What does this PR do?

Closes #730: tsconfig.json `paths` aliases that climb out of their config's
directory (e.g. a tsconfig at `apps/web/tsconfig.json` mapping `@shared/*`
to `../../packages/shared/src/*` — the standard monorepo layout) resolved
to a broken repo-relative path.

`resolve_target_relative()` in `src/pipeline/path_alias.c` only stripped a
single leading `"./"` and otherwise concatenated `dir_prefix` and `target`
verbatim, so a `../../` target left literal `".."` segments in the
resolved path (e.g. `apps/web/../../packages/shared/src/*`). That string
never matches a real module's FQN, since `cbm_pipeline_fqn_module`
tokenizes on `/` without collapsing `..` — so every cross-package import
through such an alias was invisible to the graph and looked like dead
code.

This is also why #308 regressed: the original fix (v0.7.0) handled the
simple `"./src"` case but never normalized `../` climbs, which is the
pattern monorepos actually use for cross-package aliases.

The fix makes `resolve_target_relative()` walk the target's path segments
and pop a directory off `dir_prefix` for each `..`, mirroring the
normalization `cbm_pipeline_resolve_relative_import` already does for
plain relative imports in `fqn.c`.

## Checklist

      `test_incremental.c` (2550MB vs 2048MB limit) on this machine
      environment; relying on CI
- [x] New behavior is covered by a test: added
      `path_alias_loader_monorepo_dotdot_climb` in `tests/test_path_alias.c`,
      an end-to-end test with a real tsconfig containing a `../../` alias
      target, asserting the resolved path is correct

### DIFF
--- a/src/pipeline/path_alias.c
+++ b/src/pipeline/path_alias.c
@@ -63,27 +63,74 @@ static char *strip_resolved_ext(char *path) {
     return path;
 }
 
-/* If target starts with "./" and dir_prefix is non-empty, prepend dir_prefix.
- * Returns heap-allocated repo-relative target. */
+/* Join dir_prefix with target, collapsing "." and ".." segments so aliases
+ * that climb out of their tsconfig's directory (the common monorepo
+ * pattern: a tsconfig at apps/web/tsconfig.json pointing an alias at a
+ * wildcard target like "../../packages/shared/src/" + wildcard) resolve
+ * to a real repo-relative path. Naive concatenation left literal ".."
+ * components in the target, which never match a module's FQN since
+ * cbm_pipeline_fqn_module tokenizes on '/' without collapsing them
+ * (#730). A trailing '/' on target (the usual case right before a
+ * wildcard) is preserved so the caller's later wildcard-substring
+ * concat still lines up. Returns heap-allocated
+ * repo-relative target. */
 static char *resolve_target_relative(const char *dir_prefix, const char *target) {
     if (!target) {
         return NULL;
     }
-    const char *t = target;
-    if (t[0] == '.' && t[1] == '/') {
-        t += 2;
-    }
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the tsconfig path-alias fix for #730. Triage: TypeScript import-resolution correctness.

Review will check that `../` normalization is repo-bounded, cross-platform, and does not allow alias targets to escape into unintended roots. The monorepo alias fixture described in the PR is the right acceptance shape.

DeusData:
Thank you — correct fix at the correct layer. We verified the segment-walk clamps at repo root (no traversal possible) and that the resolved string only ever feeds graph-QN resolution, never filesystem I/O; the allocation sizing is tight but exact. The red-first monorepo test seals it. Merging. Your issue write-ups on this series are a pleasure to review.

Issue #767, opened 2026-07-02
Wildcard tsconfig path alias resolves a same-prefix scoped npm package as a local folder (no import-target label check)


apappas1129:
### Version

codebase-memory-mcp 0.8.1

### Platform

_(pick yours — mine was Linux (x64))_

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

A wildcard `tsconfig.json` path alias (`"@lib/*": ["./src/lib/*"]`) matches *any* import specifier that shares its string prefix — including an unrelated scoped npm package (`@lib/external-pkg`, resolved normally from `node_modules`) that happens to start with the same `@lib/` prefix. When the substituted candidate doesn't correspond to a real, already-indexed file, resolution doesn't just fail cleanly: a later fallback strategy re-tries a *shortened* version of the import path, which matches the tsconfig's other, bare (non-wildcard) alias entry for the same prefix — and that one points at a real directory. The result is an `IMPORTS` edge from the importing file straight to that directory's **Folder** node, not to any file that was ever indexed.

Expected: the alias candidate should only "win" if it resolves to a real, indexed file — exactly how `tsc` itself handles `paths`: it tries the mapped candidate first, and if that path doesn't resolve to an actual file, it falls through to normal `node_modules` resolution. Right now this engine has no such fallback/verification step, so any scoped npm package sharing a prefix with a local alias is silently misresolved.

### Reproduction

1. Minimal dummy repo:

`tsconfig.json`:
```json
{
  "compilerOptions": {
    "paths": {
      "@lib": ["./src/lib"],
      "@lib/*": ["./src/lib/*"]
    }
  }
}
```

`src/lib/thing.ts`:
```ts
export const Thing = {};
```

`src/consumer.ts`:
```ts
import { ClientC } from '@lib/external-pkg';

export function useClient() {
  return new ClientC();
}
```

Note there is **no** `src/lib/external-pkg.ts` — `@lib/external-pkg` is meant to be a real external package (e.g. published as `@lib/external-pkg` on npm), not a local module. The bug reproduces whether or not it's actually present in `node_modules`, since the resolver never gets that far.

2. Commands:
```
codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/lib-demo"}'
[...]

PR #770, opened 2026-07-02
fix(graph): key IMPORTS edge dedup on local_name, not just source/target

### LINKED ISSUE #768: Multiple named imports from one specifier produce only a single IMPORTS edge — sibling symbols are invisible to the graph
### Version

codebase-memory-mcp 0.8.1

### Platform

_(pick yours — mine was Linux (x64))_

### Install channel

### Binary variant

standard

A single `import { A, B } from './lib'` statement — two named symbols, one specifier, zero ambiguity, no aliasing involved at all — produces exactly **one** `IMPORTS` edge in the graph, carrying only one of the two `local_name` values. The other imported (and actually used) symbol has no edge at all: it's as if that half of the import statement was never parsed.

Expected: the graph should represent both `A` and `B` as being imported by the consumer — either as two separate `IMPORTS` edges (one per symbol) or as a single edge whose `local_name`/equivalent property is a list of all symbols imported from that specifier. Either way, no imported-and-used symbol should be silently dropped.

This causes real correctness problems downstream: "who imports X" / caller / dead-code / impact-analysis queries return **zero results** for whichever symbol got dropped, even though the symbol is genuinely imported and called in the file. Worse, it's not limited to those queries — see "Root cause" below: the same dropped edge also breaks cross-file *call* resolution for that symbol, not just the import edge itself.

### Reproduction

1. Minimal dummy repo (no tsconfig paths/aliases needed — plain relative import, to isolate this from any alias-resolution issue):

`src/lib.ts`:
```ts
export const A = {
  method: () => 'value',
};

export const B = {
  check: () => true,
};
```

`src/consumer.ts`:
```ts
import { A, B } from './lib';

export function run() {
  return B.check() && A.method();
}
```

2. Commands:
```
codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/multi-import-demo"}'
codebase-memory-mcp cli query_graph '{"project":"multi-import-demo","query":"MATCH (f)-[r:IMPORTS]->(t) WHERE f.file_path CONTAINS \"consumer.ts\" RETURN t.file_path, r.local_name"}'
```
(adjust `project` to whatever name the indexer assigns your repo)

3. Result vs expected:
   - **Actual (reproduced and confirmed against the built binary):** exactly 1 row returned — `[["src/lib.ts","B"]]`. `A` has zero representation as an `IMPORTS` edge, even though it's imported and called in the file.
   - **Expected:** 2 rows (or 1 row with both names captured) — both symbols visible.

Real-world instance this was distilled from: a production TypeScript monorepo, `import { A, B } from '@lib'` in a consumer file, resolving cleanly to `src/app/lib` (no alias ambiguity — `@lib` is a bare, non-wildcard tsconfig entry). Graph kept the `B` edge; `A` had zero representation, even though `search_graph`/grep both confirm it's imported and called in that file.

### Root cause (verified by reading the edge-writer and reproducing against the built binary)

`cbm_gbuf_insert_edge()` in `src/graph_buffer/graph_buffer.c` dedups every edge purely on `(source_id, target_id, type)` — see `make_edge_key()` (~line 138) and its use inside `cbm_gbuf_insert_edge()` (~line 914). Two named imports from the same specifier resolve to the same `(source_file, target_file)` pair, so the second `IMPORTS` edge collides on that key with the first. On collision the function does not merge or reject — it **replaces** the existing edge's `properties_json` outright:

```c
cbm_gbuf_edge_t *existing = cbm_ht_get(gb->edge_by_key, key);
if (existing) {
    /* Merge properties (just replace for now) */
    if (properties_json && strcmp(properties_json, "{}") != 0) {
        free(existing->properties_json);
        existing->properties_json = heap_strdup(properties_json);
    }
    return existing->id;
}
```

The `(just replace for now)` comment suggests this was a known stub, never finished.

This has a bigger blast radius than just "who imports X" queries: `src/pipeline/pass_calls.c`, `pass_usages.c`, `pass_semantic.c`, and `pass_lsp_cross.c` all build their `local_name → resolved module QN` maps for **cross-file call resolution** by walking each file's `IMPORTS` edges and parsing exactly one `local_name` out of each edge's `properties_json` (`strstr(props, "\"local_name\":\"")`, single value, no list handling). So the dropped symbol's calls also fail to resolve cross-file — this bug silently degrades call-graph accuracy, not just import-edge visibility.

### Suggested fix direction

Key `IMPORTS` edges specifically on `(source_id, target_id, type, local_name)` instead of `(source_id, target_id, type)`, so two different symbols imported from the same specifier become two distinct edges. This requires no changes to any of the four downstream consumers listed above — they already iterate edges and read one `local_name` per edge; they will simply see both edges once dedup stops collapsing them. Other edge types (`CALLS`, `HTTP_CALLS`, `DEFINES`, ...) should keep their current `(source, target, type)` key unchanged, since collapsing multiple call sites of the same callee into one edge there is presumably intentional.

### Logs

_(none — pure graph-construction/dedup issue, not a crash/perf issue)_

_N/A_

Reproduces on a 2-file dummy repo; also observed on a ~3,300-node / ~7,600-edge production repo (multiple files affected, not isolated to one).

### Confirmations

- [x] I searched existing issues and this is not a duplicate. (Related but distinct: #180, closed, covers *zero* IMPORTS edges from unresolved package-name specifiers entirely — a different failure mode from this one, where resolution succeeds but only one of several names from the same specifier is kept.)

apappas1129:
## What does this PR do?

Closes #768: a single `import { A, B } from './lib'` statement produced exactly
one `IMPORTS` edge instead of two — whichever symbol got written second
silently overwrote the first.

`cbm_gbuf_insert_edge()` (`src/graph_buffer/graph_buffer.c`) dedups edges on
`(source_id, target_id, type)` only. Two named imports from the same specifier
resolve to the same `(source_file, target_file)` pair, so the second
`IMPORTS` edge collides with the first on that key. On collision the function
doesn't merge or reject — it replaces `properties_json` outright (the
existing `/* Merge properties (just replace for now) */` comment suggests
this was a known stub).

This has a bigger blast radius than "who imports X" queries: `pass_calls.c`,
`pass_usages.c`, `pass_semantic.c`, and `pass_lsp_cross.c` all build their
`local_name -> resolved module QN` maps for cross-file call resolution by
parsing exactly one `local_name` out of each `IMPORTS` edge's properties. So
the dropped symbol's calls also failed to resolve cross-file, not just the
import edge itself.

`make_edge_key()` now folds `local_name` into the dedup key specifically for
`IMPORTS` edges, so two different symbols imported from the same specifier
become two distinct edges. A re-inserted identical import (e.g. an idempotent
re-index) still dedups to one edge. No changes were needed in any of the four
downstream consumers — they already iterate edges and parse one `local_name`
per edge; they simply see both edges now that dedup stops collapsing them.
Other edge types (`CALLS`, `HTTP_CALLS`, `DEFINES`, ...) keep their original
`(source, target, type)` key, unchanged.

## Checklist

      `test_incremental.c` (~2.6GB vs 2048MB limit) on this machine
      environment; relying on CI
- [x] New behavior is covered by a test: added `gbuf_imports_multi_symbol_dedup`
      in `tests/test_graph_buffer.c`, asserting two same-target imports with
      different `local_name` produce distinct edges while a repeated
      identical import still dedups to one

### DIFF
--- a/src/graph_buffer/graph_buffer.c
+++ b/src/graph_buffer/graph_buffer.c
@@ -1056,7 +1056,8 @@ int cbm_gbuf_delete_edges_by_type(cbm_gbuf_t *gb, const char *type) {
         cbm_gbuf_edge_t *e = gb->edges.items[i];
         if (strcmp(e->type, type) == 0) {
             char key[EDGE_KEY_BUF];
-            make_edge_key(key, sizeof(key), e->source_id, e->target_id, e->type, e->properties_json);
+            make_edge_key(key, sizeof(key), e->source_id, e->target_id, e->type,
+                          e->properties_json);
             const char *ekey = cbm_ht_get_key(gb->edge_by_key, key);
             cbm_ht_delete(gb->edge_by_key, key);
             free((void *)ekey);
--- a/src/graph_buffer/graph_buffer.c
+++ b/src/graph_buffer/graph_buffer.c
@@ -135,7 +135,31 @@ static void make_id_key(char *buf, size_t bufsz, int64_t id) {
     snprintf(buf, bufsz, "%lld", (long long)id);
 }
 
-static void make_edge_key(char *buf, size_t bufsz, int64_t src, int64_t tgt, const char *type) {
+/* IMPORTS edges carry exactly one imported symbol's local_name (#768): two
+ * named imports from the same specifier resolve to the same (source,
+ * target) pair but are distinct symbols. Key on local_name too so the
+ * second import doesn't dedup-collide with and overwrite the first --
+ * every pass that walks IMPORTS edges (pass_calls.c, pass_usages.c,
+ * pass_semantic.c, pass_lsp_cross.c) expects one local_name per edge, so
+ * losing an edge here silently breaks cross-file call resolution for
+ * whichever symbol got dropped, not just "who imports X" queries. Other
+ * edge types keep the plain (source,target,type) key: collapsing repeat
+ * edges of the same type between the same two nodes (e.g. multiple call
+ * sites) into one is the existing, intended dedup behavior there. */
+static void make_edge_key(char *buf, size_t bufsz, int64_t src, int64_t tgt, const char *type,
+                          const char *properties_json) {
+    if (properties_json && strcmp(type, "IMPORTS") == 0) {
+        static const char local_name_key[] = "\"local_name\":\"";
+        const char *ln = strstr(properties_json, local_name_key);
+        if (ln) {
+            ln += sizeof(local_name_key) - 1;
+            const char *end = strchr(ln, '"');
+            size_t ln_len = end ? (size_t)(end - ln) : strlen(ln);
+            snprintf(buf, bufsz, "%lld:%lld:%s:%.*s", (long long)src, (long long)tgt, type,
+                     (int)ln_len, ln);
+            return;
+        }
+    }
     snprintf(buf, bufsz, "%lld:%lld:%s", (long long)src, (long long)tgt, type);
 }
 
@@ -244,7 +268,7 @@ static void remove_node_from_ptr_array(node_ptr_array_t *arr, int64_t node_id) {
 static void unindex_edge(cbm_gbuf_t *gb, const cbm_gbuf_edge_t *e) {
     char key[EDGE_KEY_BUF];
 
-    make_edge_key(key, sizeof(key), e->source_id, e->target_id, e->type);
+    make_edge_key(key, sizeof(key), e->source_id, e->target_id, e->type, e->properties_json);
     const char *ekey = cbm_ht_get_key(gb->edge_by_key, key);
     cbm_ht_delete(gb->edge_by_key, key);
     free((void *)ekey);
@@ -919,7 +943,7 @@ int64_t cbm_gbuf_insert_edge(cbm_gbuf_t *gb, int64_t source_id, int64_t target_i
 
     /* Check for dedup */
     char key[EDGE_KEY_BUF];
-    make_edge_key(key, sizeof(key), source_id, target_id, type);
+    make_edge_key(key, sizeof(key), source_id, target_id, type, properties_json);
 
     cbm_gbuf_edge_t *existing = cbm_ht_get(gb->edge_by_key, key);
     if (existing) {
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the IMPORTS dedup fix for #768. Triage: parsing/import graph correctness.

Review will focus on whether edge identity including `local_name` is sufficient without surprising other edge types, and whether the tests cover multiple named imports from the same specifier plus existing consumers of edge dedup behavior.

DeusData:
Thank you — the diagnosis is exactly right (main keys IMPORTS dedup on (src,tgt,type) and silently replaces properties, dropping sibling imports and breaking downstream local_name consumers), and your gbuf test is a genuine red-first guard at that tier. One blocking completeness issue we found while verifying end-to-end: the edge-uniqueness contract lives in three places, and this PR changes only one. Both persistence paths still declare UNIQUE(source_id, target_id, type) — the runtime store (store.c DDL + the ON CONFLICT upsert) and the raw dump writer (which emits that DDL and hand-builds the unique autoindex mechanically from all edges). With this change alone, any repo containing import { A, B } from './x' produces a DB that violates its own declared unique constraint — PRAGMA integrity_check reports non-unique index entries, and later upserts against that pair have ambiguous semantics. (CI stayed green only because no fixture has a multi-symbol import.) Could you extend the PR to change the contract coherently in all three places — e.g. include local_name in IMPORTS uniqueness or redefine the index — plus one end-to-end test: index a two-symbol import, dump, run integrity_check, query both edges back? Also worth bumping EDGE_KEY_BUF or documenting the truncation behavior for very long local names. With that, this is a clean merge — thanks again for the excellent issue write-up.


---

DeusData:
Thank you — your diagnosis was exactly right (dedup keyed on (src,tgt,type) silently replacing properties and dropping sibling imports), and your gbuf test was a genuine red-first guard at that tier. While verifying end-to-end we found the uniqueness contract lives in three places (graph buffer key, the store's UNIQUE DDL + upsert conflict target, and the raw dump writer's hand-built autoindex) — changing only one ships DBs that violate their own declared constraint, so we carried your fix over the line with the other two sites co-updated, a schema-compat probe for older DBs, and an end-to-end integrity_check test, crediting you as co-author. Merged as 541ff613 (PR #814), closing #768. Closing this in favor of the distill — your issue write-ups on this series (#767, #768, #730) have been a genuine pleasure to work with. Thanks again!

Issue #734, opened 2026-07-01
Java/Spring: class-level @RequestMapping prefix dropped from Route nodes — breaks cross-repo-intelligence for prefixed controllers


maehrlae2:
Version: v0.8.1 (Windows amd64)

## Summary
For Spring MVC/Boot REST controllers that put a base path on the class via `@RequestMapping("/prefix")` and per-endpoint paths on `@GetMapping`/`@PostMapping`/etc., the extracted `Route` node only stores the method-level path fragment — the class-level prefix is silently dropped. Every `Route` in the graph ends up with the wrong (truncated) path, so `cross-repo-intelligence` can never produce a `CROSS_HTTP_CALLS` edge for these routes: the caller's real URL (which does include the prefix) can never string-match the truncated route.

## Repro
```java
@RestController
@RequestMapping("/api/corporate/mcc")
public class MccRequirementController {

    @GetMapping("/corporates/{corporateId}/requirements")
    public ResponseEntity<?> getCorporateRequirements(@PathVariable long corporateId) { ... }
}
```

Expected `Route`: `GET /api/corporate/mcc/corporates/{corporateId}/requirements`
Actual `Route` in the graph (via `search_graph(label="Route", name_pattern=".*mcc.*")`): `GET /corporates/{corporateId}/requirements` — the `/api/corporate/mcc` prefix is missing.

Not an isolated case: on a real ~45k-node Spring Boot codebase, sampling all 120 extracted `Route` nodes, none carry their controller's class-level `@RequestMapping` prefix, across several distinct controllers/base paths.

## Impact
`index_repository(mode="cross-repo-intelligence")` against a real consumer (a Laravel/PHP frontend calling these Java REST endpoints over HTTP with fully-qualified, correct paths) returns `total_cross_edges: 0` for every category — consistent with #523, #678, #686, which look like the same underlying bug class (router/class-level prefix stripped) across different frameworks (FastAPI, Go Fiber). This appears to be the Java/Spring instance of that same class of bug, which I didn't see filed separately.

## Environment
- OS: Windows 11
- codebase-memory-mcp: v0.8.1
- Language: Java 23 / Spring Boot 3.5.5, `@RestController` + class-level `@RequestMapping` + method-level `@GetMapping`/`@PostMapping`/etc.

DeusData:
[...]

PR #772, opened 2026-07-02
fix(extract): compose Spring controller route prefixes

jonghyunlee-29cm:
## What does this PR do?

Distilled from a real Spring/Kotlin route-indexing failure: reindexing still missed actual controller API names because Spring method routes were indexed without their class-level `@RequestMapping` prefix.

## Bug/Background

Spring annotation extraction can already produce method-level routes such as `/orders`, but `push_method_def` only inspected the method node. Class-level mappings such as `@RequestMapping("/api")` or Kotlin's `@RequestMapping(value = ["/internal/v1/api"])` were not folded into the method route before Route/HANDLES emission.

That leaves the graph with partial Route node names like `/orders` instead of the actual endpoint `/api/orders`, which makes API discovery and route search miss the production route shape even though a HANDLES edge exists.

## Fix

- Compose Java/Kotlin class-level Spring mapping prefixes with method-level mapping paths while extracting method definitions.
- Walk bounded annotation-argument subtrees and choose the first path-shaped string literal, covering named and array-valued annotation arguments while ignoring non-route literals such as media types.
- Preserve the existing single-`route_path` extraction model.

## Reproduce-first

- Strengthened the Spring Java `edge_types_probe` fixture to assert exact Route node names: `/api/orders` and `/api/orders/{id}`.
- Strengthened the Spring Kotlin fixture with `@RequestMapping(value = ["/internal/v1/api"])` and exact Route node assertions.
- Assert the exact Route node set so stale partial routes such as `/orders` cannot remain alongside the full route.

## Validation

- `git diff --check`
- `GIT_CONFIG_GLOBAL=/dev/null build/c/test-runner edge_types_probe` -> `53 passed`
- `GIT_CONFIG_GLOBAL=/dev/null make -f Makefile.cbm test` -> `5747 passed`
- GitHub CI `lint / lint` -> passed (`scripts/lint.sh --ci`, which provisions cppcheck 2.20.0 and clang-format-20 before running the `lint-ci` path)

## Known limitation

This keeps the existing one-route-per-definition behavior. Multi-path annotations still select the first path-shaped literal rather than expanding into multiple Route nodes.

## Checklist

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -1323,35 +1323,58 @@ static TSNode find_decorator_args(TSNode call_node) {
     return args;
 }
 
+static bool is_route_string_kind(const char *kind) {
+    return strcmp(kind, "string") == 0 || strcmp(kind, "string_literal") == 0 ||
+           strcmp(kind, "interpreted_string_literal") == 0;
+}
+
+static const char *route_path_from_string_node(CBMArena *a, TSNode node, const char *source) {
+    if (!is_route_string_kind(ts_node_type(node))) {
+        return NULL;
+    }
+    char *path = cbm_node_text(a, node, source);
+    if (!path) {
+        return NULL;
+    }
+    int plen = (int)strlen(path);
+    if (plen >= PAIR_CHARS && (path[0] == '"' || path[0] == '\'')) {
+        path = cbm_arena_strndup(a, path + SKIP_CHAR, (size_t)(plen - PAIR_CHARS));
+    }
+    return (path && path[0] == '/') ? path : NULL;
+}
+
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the Spring route-prefix fix. Triage: parsing/route extraction bug for #734.

Review will check class-level plus method-level composition across Java and Kotlin annotation shapes, including `value=[...]` forms, and ensure emitted Route paths still preserve method/HANDLES behavior. A small controller fixture is the right guard here.

DeusData:
Thank you — verified root cause (no class-prefix folding existed anywhere in push_method_def), and the fix handles both Java and Kotlin through the shared modifiers path with strictly bounded recursion. The exact Route-set assertions are the strongest test hygiene in this review batch — they prevent stale partial routes from silently co-existing. Merging.

PR #774, opened 2026-07-02
Sync MCP server version with binary version

minorityre:
## What does this PR do?

Fixes the MCP initialize response so `serverInfo.version` comes from the same version source used by the binary instead of a hard-coded `0.10.0` string.

This also initializes the shared CLI version state from `CBM_VERSION` during startup, so release builds stamped with `scripts/build.sh --version ...` report the same version through `--version`, update checks, and MCP `initialize`.

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) - not run: this Windows environment has no `make`/`gcc`, and WSL fails to mount its distro VHDX.
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) - not run for the same local toolchain limitation.

Local checks run:

- `git diff --check -- src/main.c src/mcp/mcp.c tests/test_mcp.c`

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -521,6 +521,7 @@ int main(int argc, char **argv) {
      * below opens sqlite early), else sqlite3_config returns SQLITE_MISUSE and
      * the bind is silently ignored. No-op in the test build. */
     cbm_alloc_init();
+    cbm_cli_set_version(CBM_VERSION);
     cbm_profile_init(); /* reads CBM_PROFILE env var, gates all prof macros */
     /* CBM_LOG_LEVEL support — distilled from #414 (closes #413). Apply before
      * the first log statement so the configured level governs all output. */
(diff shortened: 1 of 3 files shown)


DeusData:
Thanks for the version-reporting fix. Triage: MCP/editor integration bug.

Review will check that `serverInfo.version`, CLI `--version`, update checks, and release-stamped builds all read from the same source without hard-coded drift. The compatibility constraint is that `initialize` response shape stays otherwise unchanged for MCP clients.

DeusData:
Thank you — this turned out to fix more than the title says. Besides serverInfo reporting a nonexistent 0.10.0, nothing ever called cbm_cli_set_version, so release builds self-identified as "dev" and the update-check compared the latest tag against 0.0.0 — meaning the 'Update available' nag fired on every release build. Your one-liner fixes both, and the test is a genuine red-first guard. Merging.

PR #776, opened 2026-07-02
PCP fork improvements: index-time importance ranking, token-budgeted repo-map tool, memory-ceiling backpressure, MCP param hardening, and pre-existing test-suite fixes

petercoxphoto:
This PR contributes a batch of improvements developed on the `petercoxphoto` fork. They fall into four independent areas plus a test-suite hardening pass; each was built test-first and reviewed. Happy to split into separate PRs if you'd prefer to take them piecemeal.

## 1. Index-time importance scoring (P3)
Computes and persists a per-symbol importance score at index time, so ranking is available without a query-time pass. Includes a canonical `cbm_is_test_path` classifier so real-corpus test files are penalised consistently in ranking.

## 2. Token-budgeted, seed-aware repo-map tool (P4)
A new `repo_map` MCP tool returning a ranked, token-bounded structural map of a repository, optionally seeded by anchor files/symbols (for orientation use-cases where you want the most relevant slice within a fixed token budget).

## 3. Memory-ceiling backpressure
Enforces an RSS memory ceiling with a per-file parse cap so a large repo can't balloon the indexer's resident set unbounded (we saw a full index reach ~15 GB before this). Also fixes `cbm_mem_rss()` under-counting on Linux, which had been blinding the ceiling/backpressure logic.

## 4. MCP parameter hardening
`search_code` and the project-reading tools now distinguish a *missing* required parameter from an *unknown* project (previously both returned a misleading "project not found"), and accept `project_id` as an alias for `project`.

## 5. Pre-existing test-suite fixes (hardening pass)
Fixes three latent defects that made the default test suite unreliable:

- **A real heap-use-after-free in the parallel indexer** (not a test bug): `extract_worker`'s final teardown called `cbm_slab_destroy_thread()` assuming the thread parser was already destroyed. A worker reaching that point with a live cached parser had its lexer's `included_ranges` slab page freed underneath it, causing a use-after-free on the next `ts_lexer_goto`. Fixed by destroying the thread parser unconditionally before the slab teardown, mirroring the two already-correct sites (whose comments document the exact invariant). Reproduced ~3 runs in 4 before the fix; clean across 20 consecutive runs after.
- **LeakSanitizer leaks** from unfreed `cbm_node_t` fields in a test, which had forced an `LSAN_OPTIONS=exitcode=0` workaround to keep the suite green.
- **An unguarded divide-by-zero** (SIGFPE on a starved store) and a `setenv`/`HOME` cleanup leak on assertion failure that could poison later tests.

After this pass the full suite exits 0 under default sanitizer options (ASan + UBSan) with no leak-suppression workaround.

## Known issue not addressed here (deliberately)
A separate, pre-existing flaky assertion in `tests/test_convergence_probe.c` (an LSP convergence-probe timing assertion) still fails intermittently — roughly 1 run in 10. It is memory-clean and unrelated to the fixes above; we left it untouched rather than fold an unrelated change into this PR.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -199,6 +199,7 @@ PIPELINE_SRCS = \
     src/pipeline/pass_similarity.c \
     src/pipeline/pass_semantic_edges.c \
     src/pipeline/pass_complexity.c \
+    src/pipeline/pass_importance.c \
     src/pipeline/pass_cross_repo.c \
     src/pipeline/artifact.c \
     src/pipeline/pass_pkgmap.c
@@ -326,7 +327,7 @@ TEST_DISCOVER_SRCS = \
 
 TEST_GRAPH_BUFFER_SRCS = tests/test_graph_buffer.c
 
-TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c
+TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_importance.c
 
 TEST_WATCHER_SRCS = tests/test_watcher.c
 
--- a/src/foundation/constants.h
+++ b/src/foundation/constants.h
@@ -48,6 +48,21 @@ enum {
     CBM_PERCENT = 100,
 };
 
+/* ── Per-file parse-size cap ──────────────────────────────────── */
+/* Default per-file source-read cap (MB), env-overridable via CBM_MAX_FILE_MB
+ * (see cbm_max_file_bytes() in system_info.c, same clamp shape as
+ * cbm_default_worker_count()/CBM_WORKERS). 10 MB clears a hand-authored
+ * amalgamation source like sqlite3.c (~8 MB) while still bounding a single
+ * worker's per-file parse working set. Previously expressed 7x as the
+ * misused CBM_PERCENT (100) * CBM_SZ_1K * CBM_SZ_1K "100 MB" literal —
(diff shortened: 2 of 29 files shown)


DeusData:
Hey @petercoxphoto, thx will check

DeusData:
Thank you for this fork's improvements — there's real value in here. We've started landing the genuine stability fixes as focused, atomic PRs: the **Linux `cbm_mem_rss()` undercount fix** (your commit 132460f5) just merged as **610531ad**, credited to you — it's high-leverage, since on Linux mimalloc's current_rss was reporting the committed-page counter (~few MB) instead of true RSS, which had been blinding backpressure/ceiling and the new host-tiered budgets. Because this PR bundles ~6 concerns (that fix + heap-UAF/leak fixes + the importance-ranking feature + the token-budgeted repo_map tool + a render-cap change), we can't merge it as one unit — could you **split it into atomic PRs**? Specifically: the remaining memory-safety fixes (heap-UAF teardown, LSan leaks) as their own bug PRs, and the **importance ranking** + **repo_map tool** as separate feature PRs (features are on hold for a moment while we finish the memory/stability release, but they're welcome after). Happy to fast-track the fix PRs. Closing this bundle in favor of the split; the rss fix is already in. Really appreciate the work.

petercoxphoto:
Thanks Martin, and thanks for pulling the RSS fix in, glad it was useful.

I've split the rest into two feature PRs:

- #879 — index-time importance scoring (weighted-degree)
- #880 — the token-budgeted `repo_map` tool (stacked on #879, since it ranks by the importance key)

On the memory-safety fixes: they're not in the split because you've already fixed them yourself. We'd branched off the v0.8.1 tag rather than current `main`, so we independently hit the same bugs and fixed them locally before realising you'd already covered them: the project alias in #714, the leaks in #581/#675, and the slab-teardown UAF in the #859 aligned-pages work. No sense duplicating any of that.

Both feature branches are rebased onto current `main` and green under ASan/UBSan (the importance suite plus the full mcp suite), so they're ready to go if and when you want them. Taking them now or leaving them until after the stability release is entirely your call, no expectations either way.

PR #780, opened 2026-07-02
Normalize project paths for ADR lookups

SS-42:
## Summary

This fixes several ADR lookup edge cases that can make an indexed project appear to have no ADR:

- report `adr_present` from the SQLite-backed ADR store, with the legacy file as fallback
- canonicalize existing `repo_path` values before indexing, so `repo_path="."` derives the same project key as the absolute path
- normalize path-like `project` arguments through the same project-key derivation used by indexing, including symlink paths
- return the existing rich project-list error for `manage_adr` and `detect_changes` when the project is missing
- update the ADR creation hint to use `manage_adr(mode='update')`

Observed environment: macOS 26.5.2 (25F84), Darwin 25.5.0 arm64, Apple clang 21.0.0. The path normalization issue is not expected to be macOS-specific; the symlink case is POSIX-path dependent.

## Test plan

- `ASAN_OPTIONS=detect_leaks=0 ./build/c/test-runner mcp`
- `ASAN_OPTIONS=detect_leaks=0 ./build/c/test-runner fqn pipeline`

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -60,6 +60,8 @@ enum {
 #include "pipeline/artifact.h"
 
 #ifdef _WIN32
+#include <direct.h>
+#include <io.h>
 #include <process.h>
 #define getpid _getpid
 #else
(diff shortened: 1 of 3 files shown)


DeusData:
Thanks for the ADR path-normalization fix. Triage: UX/path-identity bug for ADR/project lookup behavior.

This overlaps conceptually with #794 because relative `repo_path` handling must not create a second identity or stale lookup path. Review will check that canonicalization does not drift existing project names unnecessarily and that missing-project errors remain helpful.

DeusData:
Thank you — this is mostly root-cause work, not symptom patching: canonicalizing at index time plus name-derivation is 'canonicalize once at registration', and the lookup-side normalization is genuinely needed because users pass path-shaped project args at query time regardless. The store-backed adr_present fix closes a real omission (main only stat'd the legacy adr.md). One heads-up we verified: projects previously indexed via a symlinked root will derive a new name after this (one-time orphaned .db) — worth a release-note line, not a blocker. Merging.

PR #781, opened 2026-07-02
Exit promptly after parent death

SS-42:
## Summary

This makes the parent-death watchdog exit immediately from the watchdog thread when the launching process disappears.

The previous path requested graceful shutdown from the watchdog thread. That can leave a stdio server alive long enough to keep the pipe open, and it also runs non-async-safe shutdown work from a thread whose only job is parent-death detection. The new path writes a short diagnostic directly to stderr and calls `_exit(0)`.

The regression test now waits until startup reaches the watchdog-ready point before killing the wrapper, increases the timeout, and treats a zombie child as already exited.

Observed environment: macOS 26.5.2 (25F84), Darwin 25.5.0 arm64, Apple clang 21.0.0. This code path is POSIX-only; Windows remains under the existing `#ifndef _WIN32` guard.

## Test plan

- `make -f Makefile.cbm build/c/codebase-memory-mcp CC=clang CXX=clang++`
- `tests/test_parent_watchdog.sh`

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -50,6 +50,9 @@ enum {
 #include <string.h>
 #include <signal.h>
 #include <stdatomic.h>
+#ifndef _WIN32
+#include <unistd.h>
+#endif
 
 #ifndef CBM_VERSION
 #define CBM_VERSION "dev"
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for working on the parent-death path. Triage: high-priority editor/stdio lifecycle stability.

Review will focus on whether `_exit(0)` is the right boundary for this watchdog thread, whether stderr diagnostics remain short and safe, and whether the regression is deterministic across supported POSIX runners. Windows staying behind the existing guard is fine for this PR scope.

DeusData:
Thanks for this — textbook watchdog hygiene. Verified: no atexit handlers exist in src/, the pipeline lock is an in-process atomic (nothing persists past death), and SQLite WAL is crash-safe, so _exit(0) from the watchdog thread is exactly right where a stdio flush to a dead pipe could previously wedge the orphan. The test's mem.init marker placement checks out too. Merging.

PR #783, opened 2026-07-02
Isolate smoke-test dry-run state

SS-42:
## Summary

This isolates smoke-test dry-run subcommands from the developer's real home directory and cache.

Phase 6 now runs `install --dry-run`, `uninstall --dry-run`, `update --dry-run`, and the config set/get/reset round trip under a temporary `HOME`, `XDG_CONFIG_HOME`, `APPDATA`, `LOCALAPPDATA`, `CBM_CACHE_DIR`, and private `PATH`. The temporary tree is removed by the existing cleanup trap.

Observed environment: macOS 26.5.2 (25F84), Darwin 25.5.0 arm64, Apple clang 21.0.0. The isolation is shell/environment based and should apply across supported platforms.

## Test plan

- `make -f Makefile.cbm build/c/codebase-memory-mcp CC=clang CXX=clang++`
- `scripts/smoke-test.sh build/c/codebase-memory-mcp`

### DIFF
--- a/scripts/smoke-test.sh
+++ b/scripts/smoke-test.sh
@@ -12,11 +12,12 @@ set -euo pipefail
 BINARY="${1:?usage: smoke-test.sh <binary-path>}"
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")/.." && pwd)"
 TMPDIR=$(mktemp -d)
+DRYRUN_HOME=""
 # On MSYS2/Windows, convert POSIX path to native Windows path for the binary
 if command -v cygpath &>/dev/null; then
     TMPDIR=$(cygpath -m "$TMPDIR")
 fi
-trap 'rm -rf "$TMPDIR"' EXIT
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks for tightening the smoke-test dry-run isolation. Triage: UX/test-infra safety change touching scripts.

Because this changes script behavior around HOME, XDG, APPDATA, PATH, and cache directories, review will check that dry-run commands cannot mutate a developer real config and that cleanup is reliable on failure. Please keep the test evidence in the PR description current after any updates.

DeusData:
Thanks — this fixes a real hygiene bug (the smoke battery was mutating the developer's actual config via 'config set'). Env-only isolation into a script-owned mktemp home, byte-identical assertions, clean teardown. One non-blocking question we'll keep an eye on: whether Windows resolves the profile via USERPROFILE/known-folder APIs in any of these paths — worst case isolation is incomplete there, never worse than today. Merging.

PR #784, opened 2026-07-02
Reject oversized self-update ZIP entries

SS-42:
## Summary
- reject self-update ZIP entries whose declared compressed size exceeds the remaining archive bytes before inflating
- parse ZIP local-header sizes through explicit little-endian helpers and size_t allocation bounds
- remove unreachable null checks in Cypher IS NULL / IS NOT NULL evaluation after the existing null guard

Observed locally on macOS 26.5.2 (Darwin 25.5.0, arm64) with Cppcheck 2.21.0 and Homebrew clang-format 22.1.8.

## Tests
- red-first: with the old `header_end + (int)comp_size > data_len` check restored, `ASAN_OPTIONS=detect_leaks=0 ./build/c/test-runner cli` fails `cli_extract_binary_from_zip_rejects_truncated_deflate_size_over_int_max` because the truncated deflated entry is accepted
- `scripts/lint.sh --ci`
- `ASAN_OPTIONS=detect_leaks=0 ./build/c/test-runner cli cypher`

### DIFF
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -2566,7 +2566,7 @@ unsigned char *cbm_extract_binary_from_zip(const unsigned char *data, int data_l
                        (data[pos + ZIP_OFF_EXTRALEN + CLI_SKIP_ONE] << BYTE_SHIFT));
 
         int header_end = pos + ZIP_HDR_SZ + name_len + extra_len;
-        if (header_end + (int)comp_size > data_len) {
+        if (header_end > data_len || comp_size > (uint32_t)(data_len - header_end)) {
             break;
         }
 
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -80,6 +80,7 @@ enum {
 #endif
 #include <errno.h>  // EEXIST
 #include <fcntl.h>  // open, O_WRONLY, O_CREAT, O_TRUNC
+#include <limits.h> // UINT_MAX
 #include <stdint.h> // uintptr_t
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


DeusData:
Thanks for the cppcheck cleanup. Triage: bug/static-analysis cleanup touching CLI ZIP parsing and Cypher null handling.

Review will check that the ZIP size/bounds changes have regression coverage for truncation and allocation overflow, and that the Cypher changes preserve the intended NULL semantics. This looks focused, but it still needs normal test/lint evidence because it changes parser/evaluator behavior.

DeusData:
Thanks — all four hunk groups check out as genuine fixes: the LE helpers are behavior-identical, the size_t widening is sound, and we verified the cypher IS NULL guards are truly unreachable (eval_condition returns early on !actual just above). Two asks before merge, both about the third hunk — which is much more than lint cleanup: the old int-cast truncation check is genuinely bypassable for comp_size >= 2^31, letting inflate read past the downloaded archive buffer. That's a real bounds-check fix in the self-update path and deserves to be visible, not buried: (1) please retitle/split so the security-relevant fix is called out (its own commit at minimum, ideally its own PR line in the body); (2) the current regression test passes on unfixed code — comp_size=65536 against a 49-byte buffer already fails the old check, and it never triggers the >=2^31 overflow nor the DEFLATE path. Please make it red-first: comp_size >= 0x80000000 (e.g. 0xFFFF0000) with method=DEFLATE, so it fails on main and passes only with the corrected check. Then we're glad to merge. Thanks!


---

DeusData:
Thank you — every hunk here was a genuine fix, and the ZIP truncation change was much more than lint cleanup: the int-cast bypass for comp_size >= 2^31 was a real OOB-read in self-update extraction, and we've retitled it accordingly so it gets the visibility it deserves. Special thanks for amending the regression test to the 0xFFFF0000 case after the review — we verified it's a genuine red-first guard and carried it as-is. Merged as 010ac9b2 (PR #812) with you credited as co-author. Closing in favor of the distill — thanks again!

PR #789, opened 2026-07-02
feat(graph-ui): dead-code filtering, node code preview & GitHub deep-links

brishtiteveja:
## Summary

Adds dead-code discovery and code navigation to the graph UI, plus a couple of sidebar fixes and a build fix.

### Dead-code filtering + status coloring
- `layout3d` classifies each node and emits `status` (`dead | single | entry | test | exported | normal | structural`) and `in_calls`.
- **Dead** = Function/Method with zero incoming CALLS **and** zero USAGE that is not an entry point, route, test, or exported symbol.
- Caller counts come from the **full graph** via `cbm_store_batch_count_degrees` (not the sampled render edges, which would falsely mark sampled-in functions as uncalled), batched in chunks of 500 to stay within the store's bound-IN-clause capacity.
- FilterPanel gains a **Dead code** section: color-by-status, show-only-dead, hide-entry-points, hide-tests, and a live dead count. GraphTab recolors nodes by status when enabled.

### Node code preview + GitHub deep-links
- `layout3d` emits `qualified_name`, `start_line`, `end_line` per node.
- New `GET /api/repo-info?project=` returns the indexed repo's git remote normalized to a web base + `blob_base` (`https://host/org/repo/blob/<branch>`), read via libgit2. Handles scp-style (`git@host:org/repo.git`), `ssh://`, and `https://` remotes.
- Hover tooltip enriched with `file:lines`, dead-code status, and caller count.
- Node detail panel: **Show code** (fetches the snippet via the existing `/rpc get_code_snippet` tool) and **Open on GitHub** deep-link to the exact lines.

### Sidebar fixes
- Search is now a **case-insensitive regex** (matches name / file_path / qualified_name), falling back to literal substring on invalid regex, with an inline hint. Result cap raised 50 → 500.
- **Whole left sidebar scrolls as one region**; the tree no longer nests its own scroll so the (now taller) filter panel can't clip the search/tree.

### Fixes
- `restore 50k node render limit` — the prior 2000-node cap collapsed stellar degree-coloring to uniform red and dropped nearly all edges. Restored to 50000 (tunable via `CBM_UI_MAX_RENDER_NODES`), which also makes dead-code classification meaningful.
- `include git2/sys/alloc.h for git_allocator` — the `git2.h` umbrella header does not pull in `sys/` headers, breaking the build under libgit2 1.9.

## Verification
Built `cbm-with-ui` clean. Verified over `/api/*` on a 78k-node project: `in_calls` matches graph ground truth exactly, `repo-info` yields the correct GitHub base, layout nodes carry qn+lines, and `get_code_snippet` returns real source for a node's qualified_name.


phpmac:
v12 pls audit

DeusData:
Thanks for the graph UI work. Triage: graph UI/UX enhancement, but this is larger than a pure frontend change because it also touches UI server/layout code and `.gitignore`.

Review will focus on bounded file/code preview behavior, dead-code classification against full graph degrees, path/link generation, and making sure the UI changes do not introduce unbounded reads or confusing external-link behavior. Please keep the scope tight around graph UI discovery/navigation.

DeusData:
Thank you, @brishtiteveja — this is a genuinely strong set of graph-UI improvements. The **dead-code filtering**, **node code preview**, and **GitHub deep-links** all shipped: we distilled them onto current `main` in #863, merged as `b0b7d510`. The one piece we intentionally left out is the node render-cap revert — keeping the cap protects the browser from locking up on very large graphs — but if you'd like to revisit that specific behavior, a focused follow-up PR would be very welcome. Closing this in favor of the shipped distill; full credit to you for the features and the UX work. Thanks so much! 🙏

PR #793, opened 2026-07-03
fix(pipeline): honor discovery exclusions in pkgmap/envscan repo walks

### LINKED ISSUE #792: pkgmap/envscan repo walks bypass gitignore exclusions - 15+ min apparent hang on repos with large ignored artifact dirs
### Summary

`index_repository` takes **15–18 minutes** (appearing completely hung) on a repo that actually indexes in **~3 seconds**, because the auxiliary filesystem walks in `pkgmap` and `envscan` do not honor the project's gitignore-derived exclusions. They descend into a **1.6-million-file** gitignored artifact directory that the discovery pass correctly excludes.

### Environment

- codebase-memory-mcp built from main @ `dcf98dc` (also reproduced on release v0.8.1)
- Windows 11 Pro (26200), NTFS, Ryzen 7 8700F
- Repo: ~1,000 source files (TS/React + Python), plus a gitignored `map-import/output/` containing **1,638,518** generated map-tile files

### Symptoms

- `parallel.extract.progress` reaches `done=1020 total=1021` and then **no log output for 15+ minutes** while the process burns ~100% of one core, RAM slowly grows, disk I/O ≈ 0. Looks exactly like a deadlock/hang. (In reality extraction has finished — 1021 isn't a multiple of 10, so the final progress tick never prints.)
- The wall time is finite but enormous; the process eventually completes if you wait long enough.

### Root cause

At the extraction join, `merge_pkg_entries()` in `src/pipeline/pass_parallel.c` runs:

```c
/* Supplement with a repo-wide filesystem walk so manifests filtered
 * by the main discoverer (package.json, composer.json — in
 * IGNORED_JSON_FILES) still feed pkgmap. */
cbm_pkgmap_scan_repo(ctx->repo_path, &pkg_entries[0]);
```

`pkgmap_walk_dir()` (`src/pipeline/pass_pkgmap.c`) only skips `cbm_should_skip_dir()` — the hardcoded name list (`node_modules`, `.git`, …). It never consults the project's gitignore-derived exclusion list, which discovery itself honors (index_status correctly reports `map-import/output` in `excluded.dirs`). A directory named `output` matches nothing in the hardcoded list, so the walk enumerates all 1.6M files, single-threaded, invisibly between two log lines.

`pass_envscan.c` has the same defect with its own private `is_ignored_dir()` hardcoded list.

### Diagnosis evidence

Sysinternals `handle64 -p <pid>` on the "hung" process, captured mid-stall:

```
  7D0: File          C:\...\tacops\map-import\output\camp_pendleton\0\248
  A94: File          C:\...\tacops\map-import\output
  B18: File          C:\...\tacops\map-import\output\camp_pendleton\0
  CAC: File          C:\...\tacops\map-import\output\camp_pendleton
```

A/B confirmation: a `git clone` of the same repo (which lacks the untracked tile tree) full-indexes in **3.0 s**; the original working copy takes 15–18 min. With the exclusion fix applied to both walks (PR to follow), the original working copy full-indexes in **3.0 s** and an incremental no-op takes 0.4 s.

### Repro

1. Any repo with a large gitignored directory (ML checkpoints, tile pyramids, media assets):
   ```
   mkdir big && git init
   echo "bigdir/" > .gitignore
   # generate a few hundred thousand small files under bigdir/
   ```
2. `index_repository` it. Discovery excludes `bigdir/` (visible in `index_status`), but wall time scales with the total file count of `bigdir/` anyway; on NTFS this is minutes for ~10^6 files.

### Possibly related

The symptom profile (extraction "stalls" near the end, RAM growth, no I/O, hours of apparent hang on repos with heavy artifact directories) may explain parts of #775 and the large-repo reports under #593.

### Suggested fix

Thread the pipeline ctx's excluded-dirs list into `cbm_pkgmap_scan_repo()` and the envscan walker and prefix-match relative paths before descending — PR incoming from this account.

Sowiedu:
Fixes #792.

### Problem

The auxiliary filesystem walks in `pkgmap` (`cbm_pkgmap_scan_repo`, called from `merge_pkg_entries()` at the parallel-extract join) and `envscan` only skip hardcoded directory names (`cbm_should_skip_dir()` / a private `is_ignored_dir()` list). They never consult the project's gitignore-derived exclusion list that discovery honors. On a repo with a large gitignored artifact tree this walk dominates the entire pipeline: with a 1.6M-file gitignored tile directory, `index_repository` takes **15–18 minutes** (silent between two log lines, looks like a hang) instead of ~3 seconds.

### Fix

- New helper `cbm_pipeline_relpath_is_excluded()` (in `path_alias.c`) that prefix-matches a repo-relative path against the pipeline ctx's excluded-dirs list.
- `cbm_pkgmap_scan_repo()` now takes the ctx's `excluded_dirs`/`excluded_count` and consults the helper before descending, in addition to the existing `cbm_should_skip_dir()` and Windows reparse-point checks.
- The envscan walker does the same via `cbm_scan_project_env_urls_excluded()`; the original `cbm_scan_project_env_urls()` remains as a wrapper (NULL exclusions) so existing callers are unaffected.

Hardcoded skip lists are kept as a fast path; the exclusion check only adds prefix comparisons on directory descent.

### Verification

On the affected repo (~1,000 source files + 1,638,518-file gitignored `map-import/output/`, Windows 11/NTFS):

| | before | after |
|---|---|---|
| full index | 15–18 min | **3.0 s** |
| incremental no-op | ~7 min | **0.4 s** |

Synthetic check: repo with a gitignored `bigdir/` of ~20k files — the walk no longer descends into it (verified via timing and open-handle inspection with Sysinternals `handle64`). Graph output is unchanged on repos without large ignored dirs (node/edge counts identical before/after on the same tree).

Diagnosis details, handle dumps, and the elimination path are in #792.

### DIFF
--- a/src/pipeline/pass_envscan.c
+++ b/src/pipeline/pass_envscan.c
@@ -363,16 +363,22 @@ static int scan_env_file(const char *full_path, const char *rel, file_type_t ft,
 /* Process a single directory entry for env scanning. Returns bindings added. */
 static int process_env_entry(cbm_dirent_t *ent, const char *dir_path, const char *root_path,
                              cbm_env_binding_t *out, int max_out, char path_stack[][CBM_SZ_512],
-                             int *stack_top) {
+                             int *stack_top, char **excluded_dirs, int excluded_count) {
     char full_path[CBM_SZ_512];
     snprintf(full_path, sizeof(full_path), "%s/%s", dir_path, ent->name);
+    const char *rel = full_path + strlen(root_path);
+    while (*rel == '/') {
+        rel++;
+    }
 
     struct stat st;
     if (stat(full_path, &st) != 0) {
         return 0;
     }
     if (S_ISDIR(st.st_mode)) {
-        if (!is_ignored_dir(ent->name) && *stack_top < CBM_SZ_256) {
+        if (!is_ignored_dir(ent->name) &&
+            !cbm_pipeline_relpath_is_excluded(rel, excluded_dirs, excluded_count) &&
+            *stack_top < CBM_SZ_256) {
             strncpy(path_stack[*stack_top], full_path, sizeof(path_stack[0]) - 1);
             path_stack[*stack_top][sizeof(path_stack[0]) - SKIP_ONE] = '\0';
             (*stack_top)++;
@@ -386,14 +392,11 @@ static int process_env_entry(cbm_dirent_t *ent, const char *dir_path, const char
     if (ft == FT_UNKNOWN) {
         return 0;
     }
-    const char *rel = full_path + strlen(root_path);
-    while (*rel == '/') {
-        rel++;
-    }
     return scan_env_file(full_path, rel, ft, out, max_out);
 }
 
-int cbm_scan_project_env_urls(const char *root_path, cbm_env_binding_t *out, int max_out) {
+int cbm_scan_project_env_urls_excluded(const char *root_path, cbm_env_binding_t *out, int max_out,
+                                       char **excluded_dirs, int excluded_count) {
     if (!root_path || !out || max_out <= 0) {
         return 0;
     }
@@ -418,9 +421,13 @@ int cbm_scan_project_env_urls(const char *root_path, cbm_env_binding_t *out, int
         cbm_dirent_t *ent;
         while ((ent = cbm_readdir(d)) && count < max_out) {
             count += process_env_entry(ent, dir_path, root_path, out + count, max_out - count,
-                                       path_stack, &stack_top);
+                                       path_stack, &stack_top, excluded_dirs, excluded_count);
         }
         cbm_closedir(d);
     }
     return count;
+}
+
+int cbm_scan_project_env_urls(const char *root_path, cbm_env_binding_t *out, int max_out) {
+    return cbm_scan_project_env_urls_excluded(root_path, out, max_out, NULL, 0);
 }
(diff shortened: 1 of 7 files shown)


phpmac:
v12 pls audit

DeusData:
Thanks for the focused fix. Triage: high-priority performance/stability PR for #792.

Review will focus on whether pkgmap/envscan now share the same exclusion model as discovery, and whether the regression fixture proves a large ignored artifact tree is skipped rather than merely making this one repo faster. This should stay scoped to exclusion propagation and progress/degradation behavior.

DeusData:
Thank you — of the several ignore-semantics PRs open right now, this is the one closest to the root cause: reusing discovery's materialized exclusion decisions instead of re-implementing matching per walker is exactly the right foundation, and the boundary-correct prefix match checks out. Three asks before merge:
1. A small regression test (repo rule: reproduce-first) — e.g. an excluded dir containing a package.json that pkgmap must not scan.
2. A body correction: cbm_scan_project_env_urls currently has no production callers (the envscan half keeps tests compiling but can't have contributed to the hang), and the PR also fixes the path-alias walk, which deserves to be named — it's likely essential to your 3.0s result.
3. If cheap for you: the incremental path (pipeline_incremental.c's ctx lacks excluded_dirs, so merge_pkg_entries still walks unexcluded on real incremental reindexes) — otherwise we'll file it as an immediate follow-up so the 15-minute walk can't survive there.
With #1 this merges quickly — it's priority/high for good reason. Thanks again for the thorough diagnosis in #792.


---

DeusData:
Thank you — of all the ignore-semantics work in flight, yours was the one closest to the root cause: reusing discovery's materialized exclusion decisions instead of re-implementing matching per walker was exactly right, and your boundary-correct prefix predicate carried over verbatim. We took it over the line with the missing regression tests (control-first so they can't pass vacuously), the incremental-path plumbing (your fix's one gap — the ctx used by incremental reindexes didn't carry the exclusions, tracked as #804), and a small narrative correction (envscan's walker has no production callers). Merged as 2e7c64b3 (PR #811) with you credited as co-author — closes #792 and #804. Thanks again for the excellent diagnosis in #792!

PR #796, opened 2026-07-03
graph-ui: auto-rotation toggle + silent auto-refresh

MicheleColella:
## Summary

Three small, self-contained UX changes to the 3D graph UI (`graph-ui/`):

1. **Idle auto-rotation → explicit toggle.** Today the graph starts spinning on its own after 60s of inactivity (`IdleAutoRotate`). This removes that component and replaces it with an **off-by-default** `Auto-rotate` toggle button in the graph toolbar, wired to `OrbitControls.autoRotate`. Rotation now only happens when the user asks for it.

2. **Silent auto-refresh.** The graph only updated on a manual **Refresh**. This adds a background poll every `GRAPH_AUTO_REFRESH_MS` (15s) that silently re-fetches the layout so the view reflects re-index changes on its own. The poll skips the loading spinner and the error banner, so it never blanks the graph and keeps the last good render on a transient fetch failure.

3. **Filter preservation across refresh.** The filter-init effect previously reset the user's label/edge-type selection on every `data` change — which the new auto-refresh would trigger every 15s. It now re-initializes only when the graph's label/edge-type *universe* changes (new project / schema change), so a refresh preserves the current filter selection.

## Notes

- Auto-rotation is off by default, so default behavior changes only in that the graph no longer self-rotates when left idle.
- `OrbitControls` already carried `autoRotateSpeed={0.4}`; only the `autoRotate` flag is now controlled.

## Test plan

- [x] `npm run build` (tsc -b + vite build) — clean
- [x] `npm test` (vitest) — 8/8 pass
- [ ] Manual: toggle button flips rotation on/off; graph refreshes ~15s without spinner; filters survive a refresh

### DIFF
--- a/graph-ui/src/components/GraphScene.tsx
+++ b/graph-ui/src/components/GraphScene.tsx
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect, useCallback } from "react";
+import { useState, useRef, useEffect } from "react";
 import { Canvas, useThree, useFrame } from "@react-three/fiber";
 import { OrbitControls } from "@react-three/drei";
 import type { OrbitControls as OrbitControlsImpl } from "three-stdlib";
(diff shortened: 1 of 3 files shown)


MicheleColella:
Closing: opened prematurely, not intended for upstream. Sorry for the noise.

Issue #640, opened 2026-06-27
get_architecture and query tools always return "project not found" on Linux — store never opened (strace confirms no .db openat)


bengtfrost:
### Version

codebase-memory-mcp dev and Release version (binary)

### Platform

Linux (x64)

### Install channel

Built from source

### Binary variant

standard

### What happened, and what did you expect?

What happened:
get_architecture (and all other query tools) always return "project not found or not indexed" even though list_projects correctly lists the project. strace confirms that query tools never call openat on any .db file — they open the cache directory, build the available_projects list from filenames, then immediately return the error without ever attempting to open a project database.
index_repository also never creates a central codebase-memory.db registry file. The cache directory after indexing contains only the per-project .db files and _config.db.

What I expected:
get_architecture '{"project_name": "home-username-Utveckling-Rust-Repo-goose-core"}' to return architecture data for the indexed project.
Reproduction steps:

### Reproduction

1. codebase-memory-mcp cli index_repository '{"repo_path": "/path/to/repo"}'
2. codebase-memory-mcp cli list_projects → correctly lists the project
3. codebase-memory-mcp cli get_architecture '{"project_name": "<name from step 2>"}' → "project not found"
4. strace -e trace=openat codebase-memory-mcp cli get_architecture '...' → no openat on any .db file, only on the cache directory

Same result via MCP JSON-RPC over stdio. Both release and self-compiled dev build affected.

Environment:

Fedora 44, x86_64
codebase-memory-mcp dev (also reproduced on latest release)
auto_index=false, local embedding server configured in _config.db

Key observation:
list_projects opens each per-project .db file directly via directory scan and works correctly. Query tools (get_architecture, presumably search_graph etc.) appear to use a different code path to resolve the project — one that never reaches openat on the .db file. The two code paths diverge somewhere after the directory scan that populates available_projects.

### Logs

```text
❯ codebase-memory-mcp cli index_repository '{"repo_path": "/home/username/Utveckling/Rust/Repo/goose-core"}'
level=info msg=mem.init budget_mb=7877 total_ram_mb=15754
level=info msg=pipeline.discover files=612 elapsed_ms=6
level=info msg=pipeline.route path=incremental stored_hashes=612
level=info msg=incremental.classify changed=0 unchanged=612 deleted=0 mode_skipped=0
level=info msg=incremental.noop reason=no_changes
{"project":"home-username-Utveckling-Rust-Repo-goose-core",...,"status":"indexed"}

[...]

PR #805, opened 2026-07-03
fix(graph-ui): remove misleading 'Project name' field from index picker

rarepops:
Split out from #739 at the maintainer's request, so this cross-platform change can be discussed as a standalone product / API-shape decision rather than riding along with the Windows navigation work.

## What

Removes the optional **Project name** input from the index picker. Project names now always derive from the selected folder path.

## Why

The field was labelled *"Optional display name"*, but its value became the project's permanent identity, not a cosmetic label:

- the `<name>.db` filename,
- the `projects.name` primary key,
- the `project.` prefix on every node's `qualified_name`.

Since a project cannot be renamed after indexing and the name is always derivable from the folder path, the field was misleading.

## Scope

- Removes the input, its React state, and its submit payload.
- Drops the now-dead `projectName` / `projectNamePlaceholder` i18n strings (EN + ZH).
- Updates the picker test accordingly.

The backend still accepts the `project_name` alias (#640), so this only removes the UI affordance; the request/response shape is unchanged. Happy to discuss whether the field should stay, be relabelled, or be removed as done here.


rarepops:
Personally I would love this to stay and be properly implemented... I should be able to rename projects since I am not a fan of the default names they get now

DeusData:

DeusData:
Thank you for the clean split and the rigorous rationale — and for the patience while we weighed it. Decision: we're keeping the field but fixing the dishonesty you identified, rather than removing it. You're completely right that 'Optional display name' silently becoming permanent, un-renameable identity is a footgun; but removing the field costs the only human-friendly naming affordance in the UI, and the derived fallback names are the full flattened absolute path — rough in every dropdown and query scope. So the field stays with an honest label ('Project ID — permanent, cannot be renamed later') and matching help text, which lands shortly crediting you as co-author for driving the UX analysis. If a project-rename feature ships later, we'll revisit true cosmetic naming. Closing this PR per that decision — your #739 navigation work merges separately as soon as its rebase lands. Thanks again!

Issue #726, opened 2026-06-30
UI /api/layout drops connection for large graph when max_nodes is high


sharif-smj:
## Summary

The graph UI fails for a large indexed private repository. Indexing succeeds and CLI/MCP graph queries work, but opening the project in the browser UI causes `/api/layout` to hang and then drop the connection without a JSON error response.

This appears isolated to the browser UI layout endpoint, not the core indexer.

## Environment

- `codebase-memory-mcp --version`: `0.8.1`
- Binary: UI release asset for macOS arm64
- OS: macOS arm64
- UI URL: `http://127.0.0.1:9749`

## Project size

The indexed private project is roughly:

- 23k nodes
- 96k edges

`list_projects` and `get_architecture` work for this project.

## Reproduction

After indexing a large project, open it in the graph UI, or call the layout endpoint directly:

```bash
curl -v 'http://127.0.0.1:9749/api/layout?project=<large-private-project>&max_nodes=50000'
```

Observed result:

```text
curl: (52) Empty reply from server
```

In Chrome, the graph page shows:

```text
Failed to fetch
```

The request hangs for roughly 20-35 seconds before the connection is dropped. The local UI process then restarts if supervised externally.

## Threshold observed locally

These returned successfully:

```text
max_nodes=10000
max_nodes=12000
max_nodes=14000
```

These failed or caused the UI server connection to be dropped:

```text
max_nodes=16000
max_nodes=20000
max_nodes=50000
```

The bundled frontend appears to request `max_nodes=50000` when loading a graph.

## Expected behavior

For large projects, the UI should ideally do one of the following:

- render a bounded layout that the server can handle,
- automatically lower `max_nodes`,
- stream/progressively load layout data,
- or return a structured JSON error instead of dropping the connection.

## Notes

I intentionally redacted the repository name, local paths, and project-specific identifiers because this was reproduced against a private codebase. Happy to provide more diagnostics if there is a safe way to collect them without exposing source or project details.

DeusData:
Thanks for the careful report and for keeping the private repo details out. This is a valid UI/server resilience bug: the layout endpoint needs a bounded default and a structured failure path instead of letting a large graph drop the connection. I’m tracking it under the large-repo performance epic and marking it high priority for the graph UI path. The thresholds you included are already actionable.

rudi193-cmd:
## Data point — willow-2.0 graph size (v0.8.1, linux-amd64)

[...]

Issue #498, opened 2026-06-18
3D graph visualization freezes browser/system on repos with >10K nodes


push-rush:
## Bug Description

When clicking "3D Visualization" on the codebase graph page for a medium-to-large repository, the browser tab loads indefinitely and eventually freezes the entire system. On a Windows 11 machine (64GB RAM, RTX 4090), the visualization becomes unresponsive after ~30 seconds of loading, requiring a hard kill of the browser process.

## Reproduction Steps

1. Index a medium-to-large codebase (e.g. MMORPG server with 43K+ nodes, 102K+ edges)
2. Open the codebase graph page
3. Click "3D Visualization"
4. Observe: loading spinner persists, then browser becomes unresponsive, eventually OS freezes

## Environment

- **Repo scale**: 43,729 nodes, 102,719 edges
- **OS**: Windows 11 Pro 10.0.26200
- **Hardware**: 64GB RAM, RTX 4090
- **Browser**: Chrome / Edge (latest)
- **MCP version**: latest

## Technical Analysis

The 3D visualization uses a force-directed layout, which is O(n²) per iteration. With 43K nodes:

- **Layout computation**: Each iteration calculates repulsion between all node pairs → ~1.9B pairwise calculations per frame
- **WebGL rendering**: Tens of thousands of geometry primitives + text labels saturate GPU
- **Memory pressure**: Graph data + spatial index + physics state exceed available memory → swap → OS freeze

Typical web-based 3D force graphs (d3-force, vis-network, sigma.js) become unusable beyond ~1,000–2,000 nodes. This repo is 40× that threshold.

### Option A (quickest): UI-based node cap
Add a config/env variable (e.g. `VIZ_MAX_NODES=2000`) that caps rendered nodes in 3D mode. Show a warning banner: "Graph truncated: showing 2,000 of 43,729 nodes. Use filters to narrow down."

### Option B (better UX): Progressive loading with pre-filtering
- Show a query dialog before rendering: select node types (Functions only / Files only / Entry points only)
- Allow filtering by complexity threshold or file path pattern
- Default to entry points + top-level structure (~500–1,000 nodes)

### Option C (architecture): 2D hierarchy as primary view
Provide a 2D treemap / Sankey / sunburst diagram as the default visualization, reserving 3D for filtered subgraphs under 500 nodes.

## Workaround (for other users hitting this)

Temporarily reduce index scope before visualizing:

```json
{
  "index_options": {
[...]

Issue #402, opened 2026-06-01
UI binary crashes (clean exit) while rendering 3D graph


sandeepshekhawat:
### Description
The `ui` variant binary crashes with a clean exit (no error message, no stack trace) 
while the 3D graph is rendering in the browser. The process simply terminates silently.

### Environment
- **Version:** 0.7.0
- **OS:** macOS 26.2 (Build 25C56)
- **Arch:** darwin arm64
- **RAM:** 16 GB
- **Variant:** ui (installed via `install.sh --ui`)

### Steps to Reproduce
1. Install the UI variant: `curl -fsSL .../install.sh | bash -s -- --ui`
2. Run: `tail -f /dev/null | codebase-memory-mcp` (to keep stdin open)
3. Open http://localhost:9749
4. Wait for 3D graph to render

### Observed Behavior
Process exits cleanly mid-render. Logs only show:
level=info msg=mem.init budget_mb=8192 total_ram_mb=16384

No crash dump, no error, no signal info.

### Expected Behavior
Graph renders without crashing the process.

### Additional Notes
- The MCP stdio tools (`search_graph`, `trace_path`, etc.) work fine
- Crash is reproducible every time the graph attempts to render
- `KeepAlive` in LaunchAgent does not restart the process after this crash

DeusData:
Thanks for the report. A *clean* exit (no signal, no stack) mid-render points to a deliberate exit path rather than a SIGSEGV — the UI server runs alongside the MCP stdio loop, so the whole process goes down if either side exits. I can't reproduce it without a bit more:

1. **Exit code** — capture it right after it dies (e.g. `…; echo "exit=$?"`).
2. **Verbose stderr up to the exit** — paste the last ~30 lines (the only line so far is `mem.init`).
3. **Graph size** — the `nodes`/`edges` from the `index_repository` response for the project you're viewing (you're on 16 GB / budget 8192 MB, so I want to rule out a size/budget exit).
4. Does it reproduce with a **small** project (a few files), or only on a large graph?
5. After the exit, do the MCP stdio tools still respond, or is the whole process gone?

A repro project (or just its approximate node/edge count) would help most.

DeusData:
One material change since my last comment: **v0.8.1** replaced the UI's entire HTTP transport layer with a new implementation. Worth retesting the 3D graph view there — could you give v0.8.1 a try?

[...]

PR #821, opened 2026-07-03
fix(ui): guard Barnes-Hut octree against unbounded recursion on coincident nodes

Ljove02:
## What & why

`octree_insert` in `src/ui/layout3d.c` never terminates when two bodies occupy
the same (or nearly the same) position: it keeps pushing them into the same
child octant, halving `half_size` each level, until it allocates octree cells
without bound. On large graphs — where coincident/near-coincident positions are
common — this exhausts memory and freezes / SIGKILLs the UI process.

Root cause behind:
- #498  3D graph visualization freezes browser/system on repos with >10K nodes
- #726  UI /api/layout drops connection for large graphs when max_nodes is high
- #402  UI binary crashes (clean exit) while rendering the 3D graph

## The fix

Add a depth / minimum-cell-size floor to `octree_insert`. When reached (points
are effectively coincident) we stop subdividing and fold the body into the cell
as a mass-weighted aggregate — how Barnes-Hut already treats a far cluster. This
restores the intended O(n log n) time / O(n) memory for any input.

With the runaway gone, the caps that were masking it are lifted:
- server `HARD_MAX_NODES` 10000 -> 200000 (`DEFAULT_MAX_NODES` 2000 -> 60000)
- UI `GRAPH_RENDER_NODE_LIMIT` 2000 -> 60000

## Verification

A real 45,511-node / 66,839-edge project: `/api/layout` returns all 45,511 nodes
in ~2.9s, memory flat, no freeze; renders through the existing `InstancedMesh`
unchanged. Previously this froze the machine (34 GB swap) during layout.

The cap values are a policy choice — happy to gate behind `CBM_UI_MAX_RENDER_NODES`
or split into a separate commit if you'd prefer the octree guard alone.

### DIFF
--- a/graph-ui/src/hooks/useGraphData.test.ts
+++ b/graph-ui/src/hooks/useGraphData.test.ts
@@ -15,7 +15,7 @@ describe("fetchLayout", () => {
 
     await fetchLayout("large-project");
 
-    expect(GRAPH_RENDER_NODE_LIMIT).toBe(2000);
+    expect(GRAPH_RENDER_NODE_LIMIT).toBe(60000);
     expect(fetchMock).toHaveBeenCalledTimes(1);
     const calls = fetchMock.mock.calls as unknown as Array<[string]>;
     const [url] = calls[0];
--- a/graph-ui/src/hooks/useGraphData.ts
+++ b/graph-ui/src/hooks/useGraphData.ts
@@ -9,7 +9,7 @@ interface UseGraphDataResult {
   fetchDetail: (project: string, centerNode: string) => void;
 }
 
-export const GRAPH_RENDER_NODE_LIMIT = 2000;
+export const GRAPH_RENDER_NODE_LIMIT = 60000;
 
 export async function fetchLayout(
   project: string,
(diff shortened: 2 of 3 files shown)


DeusData:

DeusData:
Thank you — the octree recursion guard was exactly right, and the coincident-point analysis (unbounded subdivision once half_size underflows the coordinate ULP) precisely explained the 34GB-swap freezes. We carried the guard half over the line as **2e8f0f02** (PR #848) with you credited as co-author: the depth+half_size floor with the mass-weighted centroid fold, and the HARD_MAX_NODES raise so power users of CBM_UI_MAX_RENDER_NODES get the higher ceiling. We deferred the default render-cap change (2000→60000) — that's a real UX/perf policy shift for every user (multi-MB layout JSON, ~3s server layout), and it deserves its own PR and discussion; please do open that as a follow-up, it's a reasonable ask on its own merits. Refs #498/#726/#402 (awaiting reporter confirmation). Closing in favor of the distill — thanks again for a sharp fix!

PR #828, opened 2026-07-03
fix(pipeline): stable USAGE-edge ownership for directory-module languages (#787)

### LINKED ISSUE #787: Nondeterministic Java USAGE-edge misattribution since #667 — edges attach to wrong source files on every fresh index
## Summary

Since the merge of #667, fresh indexing attaches Java `USAGE` edges to the **wrong source files**, nondeterministically — every fresh index of the same unchanged repo produces a different set of incorrect edges. Queries then return "users" that contain no reference to the target symbol at all, while real users are missing. Not present in v0.8.1 or at the #528 merge commit; present from the #667 merge through current HEAD (dcf98dc). **This is not in any released binary yet** — flagging it now so it doesn't ship in the next release.

## Repro (spring-petclinic, macOS arm64, source builds via `scripts/build.sh`)

```bash
git clone --depth 1 https://github.com/spring-projects/spring-petclinic.git
# ground truth: 7 files reference OwnerRepository besides its own definition
grep -rl "OwnerRepository" spring-petclinic/src --include="*.java"
```

Index the repo (stdio MCP `index_repository`, fresh `CBM_CACHE_DIR` each run), then:

```cypher
MATCH (c {name: 'OwnerRepository'})<-[r:USAGE]-(m) RETURN m.file_path
```

**HEAD (dcf98dc), 5 fresh indexes — 5 different wrong answers:**

| run | bogus "users" returned | real users missing |
|---|---|---|
| 0 | EntityUtils.java, PetValidatorTests.java | ClinicServiceTests, OwnerControllerTests, PetControllerTests, VisitControllerTests |
| 1 | PetType.java | OwnerControllerTests, PetControllerTests |
| 2 | PetValidator.java | OwnerControllerTests, PetControllerTests |
| 3 | EntityUtils.java, OwnerRepository.java, PetTypeFormatterTests.java | ClinicServiceTests, OwnerControllerTests, PetControllerTests, VisitControllerTests |
| 4 | PetTypeFormatter.java, PetValidatorTests.java | OwnerControllerTests, PetControllerTests, VisitControllerTests |

None of the bogus files contain the string `OwnerRepository`.

**Controls (same machine, same build script, same query):** v0.8.1 tag and the #528 merge commit (be3e038) each return exactly the 7 real users with zero bogus entries, deterministically (2/2 runs each).

On a larger private Spring Boot service (~1.3k Java files, 142 `@Service`/`@Component` beans) the effect is severe: v0.8.1 resolves 100% of grep-verified bean users (757/757); HEAD drops to ~16% with the remainder misattributed.

## Bisect

`git bisect --first-parent` between be3e038 (good) and dcf98dc (bad), each step = full build + 3 fresh-index probes:

- good: 66b3e30 (#614), b075f05 (#633)
- **first bad: 36d83280 — merge of #667** ("QA: reproduce-first bug suite + LSP/grammar extraction fixes")

## Notes on the suspected area

#667 switches Java/Go module-QN derivation to directory-based (`cbm_pipeline_fqn_module_dir`; `pu_module_is_dir` / `pp_module_is_dir` / `pxc_module_is_dir` copies that the comments say MUST match `cbm_lang_module_is_dir`). With Java packages, many files share a directory-derived module QN, so if USAGE source resolution keys on module QN it can pick an arbitrary same-package file — consistent with the observed behavior (misattributed sources; varies run to run; still wrong, though less so, with `CBM_WORKERS=1`).

Environment: macOS 26.5 arm64 (Darwin 25.5.0), Apple clang, plain `scripts/build.sh`.

spde:
## What does this PR do?

Fixes #787: since #667, fresh indexes of the same Java tree attached `USAGE` edges to different (wrong) source files on every run.

Reproduce-first, per the triage note on the issue: the new `tests/repro/repro_issue787.c` asserts stable USAGE source ownership across repeated indexes — exact distinct-source set, 5 sequential re-indexes, and a >50-file variant that forces the parallel merge path. All three cases fail on unfixed HEAD (the parallel case surfaces a random filler file as a USAGE source, which is the race made visible).

### Root cause

#667's directory-based Java/Go module QNs collide with the package's **Folder node** QN (or **Project** for root files), while every file still emits a `Module` def with that QN. Three code paths then corrupted source attribution on the shared node:

1. `cbm_gbuf_upsert_node` (graph_buffer.c) — the #667 guard kept the Folder/Project *label* on a colliding Module def but still clobbered `name`/`file_path`/line-range, so the shared node's `file_path` became whichever same-package file was processed last.
2. `merge_update_existing` (graph_buffer.c) — the worker-gbuf merge had no guard at all: unconditional "src wins", including relabelling the Folder node to `Module`. Worker merge order varies per run — this is the nondeterminism.
3. The three source-node finders (`find_enclosing_node` / `find_source_node` / `calls_find_source`) look up class-level references' enclosing QN directly, landing on that one shared directory node — so all same-package class-level USAGE edges (e.g. `@Autowired` fields) collapsed onto a single source with an arbitrary `file_path`.

### Fix

- Both graph-buffer paths now skip the update entirely when a Module def collides with a Folder/Project node (the merge path keeps the ID remap so worker edges still redirect to the canonical node).
- New `cbm_pipeline_node_is_dir_container()` predicate; the three finders treat a directory-container hit as a miss and fall through to the per-file File node.

### Verification

- New repro: 3/3 green, stable across 10 consecutive runs (was 0/3 on HEAD).
- Full gating suite: 5748 passed, 0 failed.
- spring-petclinic, 3 independent fresh indexes: `OwnerRepository` inbound USAGE sources = exactly the 7 ground-truth files, identical every run (HEAD gave 4–6 with run-varying false positives).
- A private Spring Boot service (~1.3k Java files, 142 beans): grep-verified bean-user recall back to 757/757 = 100% (HEAD: ~16%).
- Incremental reindexing behavior from #528 preserved (inbound USAGE survives an edit+reindex cycle).

Two adjacent direct enclosing-QN lookups were left out of scope (env-access CONFIGURES sources and `resolve_file_throws`); noted here rather than folded in, happy to follow up if you want them handled the same way.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -437,6 +437,7 @@ TEST_REPRO_SRCS = \
     tests/repro/repro_new_cypher_limit_zero.c \
     tests/repro/repro_issue363.c \
     tests/repro/repro_issue581.c \
+    tests/repro/repro_issue787.c \
     tests/repro/repro_invariant_calls.c \
     tests/repro/repro_invariant_graph.c \
     tests/repro/repro_invariant_breadth.c \
--- a/src/graph_buffer/graph_buffer.c
+++ b/src/graph_buffer/graph_buffer.c
@@ -587,25 +587,31 @@ int64_t cbm_gbuf_upsert_node(cbm_gbuf_t *gb, const char *label, const char *name
     /* Check if node already exists */
     cbm_gbuf_node_t *existing = cbm_ht_get(gb->node_by_qn, qualified_name);
     if (existing) {
+        /* Don't let a per-file "Module" def touch a structural directory node
+         * ("Project" root or "Folder"). In a directory-based-module language
+         * (Go/Java) a file's module_qn equals its directory QN: a root file →
+         * the project name (== the "Project" node's QN); a file in pkg/ →
+         * proj.pkg (== the "pkg/" Folder node's QN). Its always-emitted Module
+         * def collides here; the directory node is the package/module container
+         * and must keep its structural label AND its own name/file_path/range.
+         * Updating those in place set the shared node's file_path to whichever
+         * same-package file happened to be processed LAST (worker-order
+         * dependent) — the nondeterministic file attribution behind #787 — and
+         * left the Folder node exposed to delete-nodes-by-file on incremental
+         * reindex of that file. Skip the update entirely. (Both the sequential
+         * upsert and the parallel local-gbuf merge route through this function.) */
+        if (existing->label && label && strcmp(label, "Module") == 0 &&
+            (strcmp(existing->label, "Project") == 0 ||
+             strcmp(existing->label, "Folder") == 0)) {
+            return existing->id;
+        }
         /* Update in-place. name/properties are strdup'd BEFORE freeing old ones
          * (callers may pass existing->name as an argument). label/file_path are
          * interned: gb_intern returns a stable pool pointer (idempotent even when
          * label == existing->label), so the old value is replaced, never freed. */
         char *new_name = heap_strdup(name);
         char *new_props = properties_json ? heap_strdup(properties_json) : NULL;
-        /* Don't let a per-file "Module" def downgrade a structural directory node
-         * ("Project" root or "Folder"). In a directory-based-module language
-         * (Go/Java) a file's module_qn equals its directory QN: a root file →
-         * the project name (== the "Project" node's QN); a file in pkg/ →
-         * proj.pkg (== the "pkg/" Folder node's QN). Its always-emitted Module
-         * def collides here; the directory node is the package/module container
-         * and must keep its structural label. (Both the sequential upsert and the
-         * parallel local-gbuf merge route through this function.) */
-        if (!(existing->label && label && strcmp(label, "Module") == 0 &&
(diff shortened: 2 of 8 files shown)


DeusData:
Thanks for the stable USAGE-edge ownership fix for #787. Triage: high-priority parsing correctness.

Review will focus on the reproduce-first fixture across repeated indexes and the graph-buffer merge/upsert changes. The important invariant is deterministic source ownership for directory-module languages without regressing the #667 package/module behavior.

DeusData:
Thank you — this was a precise, well-diagnosed fix. All three mechanisms you identified (the label-only guard leaving name/file_path/props clobbered, the unguarded merge path, and class-level refs landing on the shared directory node) were verified exactly as you described, and your 3-part reproduction (exact source set / 5-run stability / >50-file parallel) was model reproduce-first work. We carried it over the line as **896aa914** (PR #844) with you credited as co-author — the only changes were adding DCO sign-off (your commits lacked the trailer) and one clang-format fix. Closes #787. Closing in favor of the distill — thanks again for the excellent write-up!

PR #829, opened 2026-07-03
fix(build): gate libgit2 allocator support

### LINKED ISSUE #801: build: libgit2 version floor (git_allocator needs >=1.7) + libgit2-enabled CI leg
Follow-up from #722/#705. internal/cbm/cbm.c binds the 3-member git_allocator (gmalloc/grealloc/gfree), which upstream introduced in libgit2 1.7.0 — on distros shipping older libgit2 (Debian 12 = 1.5.1, Ubuntu 22.04 = 1.1.0) the positional initializer mis-wires the struct (clang >=16: hard error; older GCC: warning + silently corrupt allocator).

Proposed:
- Version floor: `#if LIBGIT2_VER_MAJOR == 1 && LIBGIT2_VER_MINOR >= 7` around the binding, or `pkg-config --atleast-version=1.7 libgit2` in Makefile.cbm (clear error either way).
- A libgit2-enabled CI matrix leg so the HAVE_LIBGIT2 path stops being CI-invisible (it was never built in CI, which is how the missing include shipped). @kriswill volunteered in #722.

dpersek:
## Summary
- Require libgit2 >= 1.7.0 before enabling the optional `HAVE_LIBGIT2` build path.
- Add `REQUIRE_LIBGIT2=1` so CI/proof builds fail clearly if the intended libgit2 path is missing or too old.
- Add a defensive header-version guard before binding `git_allocator`.
- Run one Ubuntu PR test matrix leg with `libgit2-dev` installed and required.

## Why
Fixes #801.

The allocator binding uses the libgit2 1.7+ shape, but the build previously enabled `HAVE_LIBGIT2` for any libgit2 pkg-config result. PR CI also did not intentionally cover the optional libgit2 path.

## Tests
- `make -f Makefile.cbm cbm REQUIRE_LIBGIT2=1`
- `make -f Makefile.cbm build/c/test-runner REQUIRE_LIBGIT2=1`
- Missing pkg-config fallback probe with `PKG_CONFIG=/bin/false`
- Fake libgit2 1.6.9 pkg-config probe: default fallback works, required mode fails clearly
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/_test.yml")'`
- `git diff --check`

## Scope
- Keeps missing or too-old libgit2 as a normal fallback unless `REQUIRE_LIBGIT2=1` is set.
- Does not vendor or build libgit2 from source.
- Does not expand every OS matrix leg.
- Does not change git-history parsing behavior.

## Caveats
- Full `scripts/test.sh`, `make -f Makefile.cbm test`, and `make -f Makefile.cbm lint-ci` were not run locally.
- `actionlint` was unavailable locally, so workflow validation used YAML parsing plus manual review.
- Old-version behavior is proven with pkg-config simulation and header-macro compatibility review, not by compiling against a real old libgit2 package.

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -33,7 +33,7 @@ jobs:
           BROAD: ${{ inputs.broad_platforms }}
         run: |
           CORE_UNIX='[
-            {"os":"ubuntu-latest","cc":"gcc","cxx":"g++"},
+            {"os":"ubuntu-latest","cc":"gcc","cxx":"g++","libgit2":true},
             {"os":"ubuntu-24.04-arm","cc":"gcc","cxx":"g++"},
             {"os":"macos-14","cc":"cc","cxx":"c++"},
             {"os":"macos-15-intel","cc":"cc","cxx":"c++"}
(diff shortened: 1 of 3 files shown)


DeusData:
Thanks for the libgit2 build-gating fix for #801. Triage: build/CI correctness PR.

Because this touches `Makefile.cbm` and `.github/workflows/_test.yml`, review needs the usual build-infra caution: verify fallback behavior when libgit2 is absent/old, required-mode failure when requested, and that the new CI leg proves the intended optional path without making all platforms depend on libgit2.

DeusData:
Thank you — this closes #801 more completely than the issue asked: the header-level guard correctly handles a future libgit2 2.x (the issue's own suggestion would have rejected it), and making the CI leg loud via REQUIRE_LIBGIT2=1 means silent fallback can never fake coverage again. The exact regression class that let the missing include ship (#722's backstory) is now structurally impossible. Merging.

PR #830, opened 2026-07-03
fix(ui): unwatch projects after UI delete

### LINKED ISSUE #803: ui: DELETE /api/project leaves a zombie watch (deleted project resurrects)
Follow-up from the #537 review. #537 fixed the MCP delete_project path (now unwatches, with deferred free). The UI HTTP server has a second production deletion path — DELETE /api/project in src/ui/http_server.c (~:993-1024) unlinks the .db but has no watcher access at all — so a project deleted via the UI still gets auto-reindexed by its live watch entry, recreating the DB (the resurrection bug), same class as #537.

Fix shape: route the UI delete through the same unwatch (+ deferred free) that handle_delete_project now uses. Also from the #537 review: the realloc-failure fallback in cbm_watcher_unwatch frees immediately, reintroducing the cross-thread UAF in the OOM corner — worth hardening in the same pass.

dpersek:
## Summary
- Wire the UI HTTP server to the existing watcher so project lifecycle actions can update subscriptions.
- Unwatch projects when `DELETE /api/project` successfully deletes the DB or finds a valid project DB already missing.
- Add focused HTTP regressions for success, already-missing DB, no-watcher behavior, missing name, invalid name, and unlink failure.
- Harden watcher unwatch bookkeeping so pending-free allocation failure leaves the watch entry intact.

## Why
Fixes #803.

The UI delete path removed project database files but left the watcher entry alive, so a later watch cycle could reindex the deleted project and recreate it. The MCP delete path already handled watcher cleanup; this applies the same lifecycle boundary to the UI HTTP route.

## Tests
- `build/c/test-runner httpd watcher` (`92 passed`)
- `git diff --check`

## Scope
- Keeps `400` for missing names and no watcher cleanup.
- Keeps invalid project names as `404` without watcher cleanup.
- Keeps primary DB unlink failures as `500` without watcher cleanup.
- Does not change project discovery, indexing, or MCP delete behavior.

## Caveats
- The watcher out-of-memory path is review-covered rather than fault-injection tested to avoid adding public test-only hooks.
- Full `make -f Makefile.cbm test` and `make -f Makefile.cbm lint-ci` were not run locally; focused affected C suites were rebuilt and passed.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -637,6 +637,7 @@ int main(int argc, char **argv) {
     if (ui_cfg.ui_enabled && CBM_EMBEDDED_FILE_COUNT > 0) {
         g_http_server = cbm_http_server_new(ui_cfg.ui_port);
         if (g_http_server) {
+            cbm_http_server_set_watcher(g_http_server, g_watcher);
             if (cbm_thread_create(&http_tid, 0, http_thread, g_http_server) == 0) {
                 http_started = true;
             }
--- a/src/ui/http_server.c
+++ b/src/ui/http_server.c
@@ -19,6 +19,7 @@
 #include "ui/layout3d.h"
 #include "mcp/mcp.h"
 #include "store/store.h"
+#include "watcher/watcher.h"
 #include "cli/cli.h"
 /* pipeline.h no longer needed — indexing runs as subprocess */
 #include "foundation/log.h"
@@ -31,6 +32,7 @@
 #include <sqlite3/sqlite3.h>
 #include <yyjson/yyjson.h>
 
+#include <errno.h>
 #include <math.h>
 #include <stdatomic.h>
 #include <stdio.h>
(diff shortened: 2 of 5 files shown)


DeusData:
Thanks for the UI delete/watch cleanup for #803. Triage: UX/project-lifecycle bug.

This PR is currently conflicting, so the first step is a rebase on current `main`. After that, review will check that UI delete mirrors the MCP delete lifecycle boundary, and that watcher unwatch bookkeeping stays correct on success, already-missing DBs, invalid names, and unlink failures.

DeusData:
Thank you — this was exactly what #803 needed, and your red-first test (main's UI delete handler provably had zero watcher access) made the review easy. We carried it over the line as e8d28327 (PR #843) with you credited as co-author, with one structural improvement your branch predated: the OOM hardening moved into the shared defer_state_free helper (which now returns success and only removes the watch entry when the defer actually succeeded), so both the unwatch path AND the root-prune path from the watcher-prune work get the same UAF-free discipline. Closes #803. Closing this in favor of the distill — thanks again, and the libgit2 gate in #829 was merged as-is earlier today!

Issue #832, opened 2026-07-04
Memory not released after indexing: 20GB+ RSS for 5MB of indexed data


fxfxfx123:
## Summary

After indexing a small project (65 files, 1.3MB, 2509 nodes), codebase-memory-mcp retains 20GB+ RSS on a 32GB Windows machine. Memory grows monotonically and is never released to the OS.

## Environment

- OS: Windows 11, i7-12700 (20 cores), 32GB RAM
- CBM Version: 0.8.1
- Project: 65 Markdown files, 1.3MB, 2509 nodes

## Root Cause (from source code)

1. mimalloc abandoned-thread arenas not purged (mem.c missing mi_option_abandoned_thread_purge)
2. Default worker count uses all 20 cores, each with its own mimalloc arena
3. DEFAULT_RAM_FRACTION=0.5 means 16GB budget, no pressure to release

## Observed memory

Without fix: 20GB+ RSS, growing 600MB/min, never releases
With workaround (CBM_WORKERS=1): 13MB stable

## Fix

PR #831: https://github.com/DeusData/codebase-memory-mcp/pull/831

- Enable mi_option_abandoned_thread_purge in mem.c
- Lower DEFAULT_RAM_FRACTION from 0.5 to 0.25
- Cap initial worker count at 8 in system_info.c

## Workaround

Set CBM_WORKERS=1 in MCP env config and run: codebase-memory-mcp config set auto_index false

DeusData:
Thanks for the Windows memory-retention report and the before/after numbers. Triage: high-priority stability/performance issue under #593, with Windows impact.

This overlaps the memory-pressure work in #752/#685 but is more specific about post-index RSS retention and mimalloc worker-thread arenas. I see #833 is open against this; review should check that memory is actually returned after indexing without just hiding the issue by reducing worker parallelism too aggressively.

DeusData:
v0.9.0 includes memory-retention and RSS accounting fixes plus the indexing supervisor work, so this memory-retention report needs a fresh pass. Could you retest on v0.9.0 with the same project shape and share peak/live RSS if it still grows? Marking awaiting-reporter.

fxfxfx123:
## v0.9.0 Memory Test Report

Clean install of v0.9.0 (UI variant, Windows amd64), same project structure as before.

### Environment
- **OS**: Windows 10 (amd64)
- **Binary**: codebase-memory-mcp-ui-windows-amd64.zip (v0.9.0)
[...]

PR #833, opened 2026-07-04
fix: memory not released after indexing (20GB+ RSS for 5MB data)

fxfxfx123:
## Problem

After indexing a small project (65 files, 1.3MB, 2509 nodes), codebase-memory-mcp retains 20GB+ RSS on a 32GB Windows machine. Memory grows monotonically and is never released to the OS.

## Root Cause (from source code)

1. mimalloc arenas from exited worker threads not purged - mem.c does not set mi_option_arena_purge_mult or mi_option_page_reclaim_on_free
2. Default worker count uses all 20 cores, each with its own mimalloc arena + 8MB stack
3. DEFAULT_RAM_FRACTION=0.5 means 16GB budget, no pressure to release

## Fix

| File | Change | Effect |
|------|--------|--------|
| src/foundation/mem.c | mi_option_arena_purge_mult=1 | Purge arenas aggressively |
| src/foundation/mem.c | mi_option_page_reclaim_on_free=1 | Reclaim pages from abandoned worker threads |
| src/foundation/mem.c | DEFAULT_RAM_FRACTION 0.5 to 0.25 | Lower memory budget |
| src/foundation/system_info.c | Cap initial workers at 8 | Reduce peak memory 60pct |

## Benchmark

| Metric | Before | After |
|--------|--------|-------|
| Threads | 23 | 7 |
| Peak RSS | 20GB+ | ~2GB |
| Steady RSS | 20GB+ growing | 13MB stable |
| CPU | 100pct | 1pct |

Tested on Windows 11, i7-12700, 32GB RAM, 65-file project.

### DIFF
--- a/src/foundation/mem.c
+++ b/src/foundation/mem.c
@@ -122,8 +122,10 @@ void cbm_mem_init(double ram_fraction) {
     mi_option_set(mi_option_arena_eager_commit, 0);
     mi_option_set(mi_option_purge_decommits, SKIP_ONE);
     mi_option_set(mi_option_purge_delay, 0); /* immediate purge, no 1s delay */
-    mi_option_set(mi_option_arena_purge_mult, 1); /* purge arenas aggressively, not 10x delay */
-    mi_option_set(mi_option_page_reclaim_on_free, 1); /* reclaim pages from abandoned worker threads */
+    /* Purge arenas aggressively (default multiplier is 10x purge_delay). */
+    mi_option_set(mi_option_arena_purge_mult, 1);
+    /* Reclaim pages from exited worker thread heaps on free. */
+    mi_option_set(mi_option_page_reclaim_on_free, 1);
 
     /* CBM_MEM_BUDGET_MB env override (memory analogue of CBM_WORKERS).
      * Lets users cap the budget directly without an enclosing cgroup —
--- a/src/foundation/mem.c
+++ b/src/foundation/mem.c
@@ -13,7 +13,7 @@
 #include "foundation/constants.h"
 
 #define MAX_RAM_FRACTION 1.0
-#define DEFAULT_RAM_FRACTION 0.5
+#define DEFAULT_RAM_FRACTION 0.25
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the Windows memory-retention fix for #832. Triage: high-priority stability/performance PR.

Review will check the mimalloc options and worker-count change separately: we need post-index RSS to return to sane levels, but we also need to avoid an over-broad throughput regression or conflicting with the explicit memory-budget work in #752/#685. Please keep the before/after memory evidence current in the PR.

fxfxfx123:
Yes, still active and waiting for review. All CI checks pass. Happy to adjust if there are concerns about the worker-count change — the mimalloc arena purge (mem.c) is the core fix, and the worker cap (system_info.c) can be dropped if it conflicts with #752/#685.

DeusData:

DeusData:
Thank you — your diagnosis found the right lever. The core of the #832 fix is `mi_option_page_reclaim_on_free=1`: mimalloc v3 flipped the reclaim default so a long-lived thread no longer reclaims pages abandoned by exited worker threads → the RSS ratchet you reported. That exact option is folded into the keystone memory fix, merged as **7884ccb9** (PR #853), with you credited. The other hunks were dropped after analysis: `arena_purge_mult=1` is inert at our `purge_delay=0`, the `DEFAULT_RAM_FRACTION` change was dead code (all call sites pass an explicit fraction), and the worker-cap belonged to a different layer (host-tiered budgets, since landed as #752).

Being straight with you: **#832 isn't fully closed yet.** The keystone also routes the two background index paths through a subprocess (so the kernel returns 100% of each cycle's RSS on exit), and #854 added budget-derived retention — so 'RSS won't come back' is fixed. But the *trigger* you hit on Windows (the watcher re-indexing on every poll even when nothing changed) is tracked separately as #841 and still to come. So I'll leave #832 open until that lands, but your page_reclaim finding is shipped. Really appreciate the sharp report — closing this PR in favor of the folded fix.

Issue #835, opened 2026-07-04
Withdrawn by author


tecnos-okazaki:

DeusData:
Thanks for the measured TS/JS call-resolution report. Triage: high-priority parsing/call-graph correctness under #592.

The failure mode is clear: unresolved receivers should not fall through to weak same-name strategies and fabricate CALLS edges to unrelated project symbols. I see #836 already targets this. Review should verify the TS/JS guard suppresses only weak member-call matching and does not drop resolved LSP-backed calls.

DeusData:
Fixed on main via the resolver change that suppresses weak short-name strategies for TS/JS member calls with unresolved receivers — so a call like re.test() no longer fabricates a CALLS edge to an unrelated same-named symbol (e.g. SalesforceRestClient.test), and high-fan-in 'magnet' symbols stop poisoning get_architecture/trace_path. LSP-resolved calls are unaffected. Full credit to @tecnos-okazaki, whose work landed via a recovery PR (#840) after the original fork was deleted mid-merge. A retest on main/next release would be welcome. Note: genuinely cross-file inherited TS method resolution remains a separate gap, now tracked with a red reproduction.

PR #836, opened 2026-07-04
Withdrawn by author

tecnos-okazaki:


DeusData:
Thanks for opening the TS/JS resolver fix for #835. Triage: high-priority parsing/call-graph correctness.

Since this is currently a draft, we will wait for it to be marked ready before full review. The review focus will be narrow: suppress weak short-name strategies for unresolved TS/JS member receivers without dropping resolved LSP-backed calls or bare calls.

DeusData:
Thank you — this is exactly the graph-quality work we want: the drop-list construction is provably right (a keep-list would have killed lsp_ts_method), the emit-point gating spares every Route/HTTP/ASYNC branch by structure, and disclosing the discarded draft that lost 399 HTTP_CALLS edges is model engineering transparency. The ts/S6 probe flip is approved — its old green was a false guard by our own annotation; we're adding a RED reproduction for the TS cross-file inheritance gap on our side so the gap stays visible. Merging.

DeusData:
Heads-up for the record: your fork was deleted while we were merging, which made GitHub auto-close this PR in a way that cannot be reopened. No action needed — we recovered your exact commits (authorship and sign-offs intact) from the preserved PR ref and they are landing via the recovery PR referenced above. Thank you again for excellent work — this closes #835 as intended, fully credited to you.

PR #839, opened 2026-07-04
test(graph-ui): fix locale-dependent number-format assertion in GraphTab.test.ts

### LINKED ISSUE #825: test(graph-ui): GraphTab number-format assertion is locale-dependent
GraphTab.test.ts asserts a formatted count of `2,000`, but `Intl`/toLocaleString under a German locale renders `2.000` — the test is green on en_US CI runners and red on any de_* developer machine (observed 2026-07-03 during unrelated work; suite green under LC_ALL=en_US.UTF-8).

Fix: pin the locale in the assertion or format via an explicit locale (e.g. `toLocaleString('en-US')`) in the component/test so the suite is environment-independent.

sahil-mangla:
## What does this PR do?

Pins the number formatting in `formatGraphLimitNotice` to the `'en-US'` locale. 

Previously, the use of `toLocaleString()` without specifying a locale caused numbers to format according to the host machine's system locale (e.g., `2.000` under German locales like `de-DE` instead of `2,000`). This caused the unit test in `GraphTab.test.ts` to fail on any developer machine configured with a non-`en_US` locale. Explicitly pinning this to `'en-US'` ensures consistent, environment-independent formatting for tests and the English UI message.

Fixes #825

## Checklist

- [x] Tests pass locally (`make -f Makefile.cbm test` / `npm run test` in `graph-ui`)
- [x] Lint/Build passes (`npm run build` in `graph-ui`)

### DIFF
--- a/graph-ui/src/components/GraphTab.tsx
+++ b/graph-ui/src/components/GraphTab.tsx
@@ -31,7 +31,7 @@ interface GraphTabProps {
 
 export function formatGraphLimitNotice(data: GraphData | null): string | null {
   if (!data || data.total_nodes <= data.nodes.length) return null;
-  return `Showing ${data.nodes.length.toLocaleString()} of ${data.total_nodes.toLocaleString()} nodes. Use filters to narrow.`;
+  return `Showing ${data.nodes.length.toLocaleString("en-US")} of ${data.total_nodes.toLocaleString("en-US")} nodes. Use filters to narrow.`;
 }
 
 export function GraphTab({ project }: GraphTabProps) {


DeusData:
Thanks for the focused locale-stability fix for #825. Triage: graph UI test/UX bug, backlog priority.

Review will check that pinning `en-US` is intentional for this English UI notice, and that the test remains stable on non-English developer machines without masking future i18n work. This looks nicely scoped.

DeusData:
Thank you — this is the right fix: pinning the format to `toLocaleString("en-US")` in the component makes the count deterministic regardless of the viewer's host locale, rather than just patching the test assertion. The surrounding notice is English-only so the digit grouping stays consistent. Merged as 093eaf17; closes #825. Appreciated!

PR #857, opened 2026-07-04
fix(extract-rust): mark #[test]/#[tokio::test] fns is_test so they are filtered

### LINKED ISSUE #855: fix(extract-rust): inline test fns (#[test]/#[tokio::test]) are not marked is_test — leak into graph/agent context
## Summary

Inline Rust test functions — `#[test]`, `#[tokio::test]`, `#[async_std::test]`, `#[actix_rt::test]`, and `#[cfg(test)] mod tests { … }` bodies — that live inside a **regular** source file (e.g. `src/lib.rs`) are **not** detected as tests. They are indexed as ordinary `Function` nodes with `is_test = false`, so they are **not** filtered by the `is_test != 1` clause in `store.c` and leak into graph queries / agent context, polluting it with test code.

## Root cause

Test detection is **file-path-based only**:

- `cbm_is_test_file(rel_path, lang)` (`internal/cbm/helpers.c:292`) marks a *file/module* as a test only by path. For Rust it is `return has_suffix(base, "_test.rs") || has_prefix(base, "test_");` (with the code comment *"Rust tests are typically mod tests inside the file, but test files too"*). A file like `src/lib.rs` is therefore never a test file, regardless of its contents.
- `CBMDefinition.is_test` (`internal/cbm/cbm.h:213`) is the per-*function* flag. It is **read** by `src/pipeline/pass_definitions.c` and `src/pipeline/pass_parallel.c` (emitted into node properties) but is **never assigned** in the extraction layer — the only `is_test` write is `mod.is_test = ctx->result->is_test_file;` (`internal/cbm/extract_defs.c:6316`), which is module/file-level. So `def->is_test` stays at its zero-initialized default (`false`).

Because functions are never marked, the `store.c` filter (`json_extract(properties, '$.is_test') != 1`) only ever drops whole test-*files*, never inline test functions.

## Repro

Index any repo containing, in a non-test-named file:

```rust
// src/lib.rs
pub fn real_fn() {}

#[cfg(test)]
mod tests {
    #[tokio::test]
    async fn something_works() {}
}
```

`cbm cli --json search_graph '{"project":"…","name_pattern":"something_works"}'` returns `something_works` as a `Function` with `is_test` absent/false — i.e. it is not filtered and appears alongside `real_fn` in graph context.

## Proposed fix (Rust-scoped)

In the Rust function-extraction path in `internal/cbm/extract_defs.c`, scan each function's `attribute_item` / `meta_item` children for test macros and set `def->is_test = true`. Initial attribute set:

- bare `#[test]`
- `#[tokio::test]` (the case that surfaced this — async tests)
- `#[async_std::test]`
- `#[actix_rt::test]`
- `#[test_case::case]` (optionally)

Plus a regression test in `tests/test_extraction.c` asserting a Rust snippet with `#[tokio::test] async fn …` yields `def->is_test == true`. `make -f Makefile.cbm test` + `scripts/lint.sh` clean.

## Scope

Rust only in the follow-up PR. Rust is a high-value first target because inline `#[cfg(test)] mod tests { … }` is idiomatic, so these unmarked test functions leak into the graph for *any* typical Rust crate — not an edge case. Go / Python / JS / Java / C# can follow in separate PRs per the one-issue-per-PR guidance.

## Ask

**Status:** PR #857 implements this (Rust-scoped, with a regression test; `scripts/test.sh` green, red→green verified). Feedback on the attribute set and the extraction-hook point welcome there or here — happy to adjust. (Filed for alignment first; went ahead since it's a focused correctness fix.)

---
*Citations verified against `main` at `37c00d1`.*

moofone:
Closes #855.

## Summary
- Inline Rust test functions — `#[test]`, `#[tokio::test]`, `#[async_std::test]`, `#[actix_rt::test]`, `#[test_case::case]` — inside a regular `.rs` file are now marked `is_test = true` on the `Function` node, so the `store.c` `is_test != 1` filter excludes them from graph/agent context. Previously they leaked in because test detection was file-path-based only (`cbm_is_test_file`: `*_test.rs` / `test_*`).
- Adds `rust_def_is_test(decorators)` (mirrors `rust_cfg_qualified_name`'s decorator scan), set on `def` in `extract_func_def`'s Rust branch. `extract_decorators` stores `attribute_item` text in its bracketed form (`"#[test]"`, `"#[tokio::test]"`, `"#[tokio::test(...)]"`), so the matcher is substring-based and robust to argument lists.
- Rust-scoped per the issue. Free functions are covered, including `#[cfg(test)] mod tests { #[tokio::test] async fn … }`. `impl`/`trait` methods can follow in a separate change.

## Verification
- `scripts/test.sh` → **5831 passed** (clean ASan + UBSan build). New regression `extract_rust_test_attr_marks_is_test_issue855` covers `#[test]`, `#[tokio::test]`, and a negative (non-test) case.
- Red → green: with detection disabled, the new test fails exactly at `ASSERT(sync == 1 && "#[test] fn is_test")` and is the only failure (5830 passed, 1 failed); restored → green.
- `scripts/lint.sh` (clang-tidy / cppcheck / clang-format) is not available in my local environment, so CI is the authority for it; the new code mirrors the surrounding `rust_cfg_qualified_name` style.

## Notes
- DCO-signed. Happy to adjust the attribute set or the extraction-hook point per maintainer feedback on #855.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -1869,6 +1869,35 @@ static const char **extract_decorators(CBMArena *a, TSNode node, const char *sou
  * first and one branch is lost (#495). Fold the cfg predicate into the QN so
  * each cfg-gated twin gets a DISTINCT, predicate-encoding QN. Returns the
  * (possibly suffixed) QN; the original QN when no cfg attribute is present. */
+/* Rust: mark a function as a test when it carries a test attribute (#855).
+ * cbm's test detection is otherwise file-path-based (cbm_is_test_file:
+ * *_test.rs / test_*), so inline #[test]/#[tokio::test] functions inside a
+ * regular .rs file are indexed as ordinary Functions (is_test=false) and leak
+ * past the store.c `is_test != 1` filter into graph/agent context. The
+ * attribute_item text extract_decorators stores is the bracketed form
+ * ("#[test]", "#[tokio::test]", "#[tokio::test(...)]", ...). */
+static bool rust_def_is_test(const char *const *decorators) {
+    if (!decorators) {
+        return false;
+    }
+    for (int i = 0; decorators[i]; i++) {
+        const char *d = decorators[i];
+        /* Path-qualified async/param test macros (substring match, robust to the
+         * optional argument list and the surrounding #[ ]). */
+        if (strstr(d, "tokio::test") || strstr(d, "async_std::test") ||
+            strstr(d, "actix_rt::test") || strstr(d, "test_case::case")) {
+            return true;
+        }
+        /* Bare #[test] / #[test(...)]: match the bracketed path exactly so we do
+         * NOT match the unrelated #[test_case::case] (handled above) or a
+         * hypothetical #[test_crate]. */
+        if (strstr(d, "#[test]") || strstr(d, "#[test(")) {
+            return true;
+        }
+    }
+    return false;
+}
+
 static const char *rust_cfg_qualified_name(CBMArena *a, const char *base_qn,
                                            const char *const *decorators) {
     if (!decorators) {
@@ -3192,6 +3221,7 @@ static void extract_func_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec 
     // predicate into the QN so both branches survive the graph upsert (#495).
     if (ctx->language == CBM_LANG_RUST) {
         def.qualified_name = rust_cfg_qualified_name(a, def.qualified_name, def.decorators);
+        def.is_test = rust_def_is_test(def.decorators);
     }
 
     // Docstring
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you — this is a correct, well-scoped fix: detecting #[test]/#[tokio::test]/#[async_std::test]/etc. attributes and marking those functions is_test, which flows end-to-end into the node properties and the store queries that exclude tests from graph/agent context. The decorator substring-matching mirrors the existing rust_cfg pattern, and your test (on a non-test path so is_test can only come from the attribute detection) is a genuine non-vacuous guard. Merged as bf3657f7; closes #855. Deferring impl/trait test methods to a follow-up was the right call. Appreciated!

Issue #867, opened 2026-07-05
Incremental reindex costs ≈ a full index, especially after artifact bootstrap


moofone:
## Problem

After artifact bootstrap (and to a lesser degree on steady-state incremental runs), an incremental re-index ends up re-doing almost the work of a full index. Three independent causes, verified against `main` (`74b27f3`):

### P1 — Artifact bootstrap doesn't actually avoid the full-index cost

`try_artifact_bootstrap` (`src/mcp/mcp.c`) byte-copies the shared artifact DB into the local cache; the pipeline then routes to `cbm_pipeline_run_incremental` because hash rows exist (`src/pipeline/pipeline.c`, the `pipeline.route path=incremental` branch).

`classify_files` (`src/pipeline/pipeline_incremental.c`) classifies changed/unchanged **only by `mtime_ns + size`**. The imported `file_hashes` rows carry the **exporter's** mtimes; a fresh clone stamps every file with checkout time → essentially every file mismatches → full extract + resolve + post-passes. Bootstrap saves almost nothing on run 1 — the exact run it exists to make cheap.

Two building blocks already exist but are dead/unused in production:
- `cbm_artifact_commit()` (`src/pipeline/artifact.c`) reads the git commit from `artifact.json` — never called in production.
- `cbm_store_upsert_file_hash_batch()` (`src/store/store.c`) — exists, never called in production.

### P2 — `persist_hashes` is row-at-a-time and double-stats every file

`persist_hashes` (`src/pipeline/pipeline_incremental.c`) re-`stat()`s every discovered file (classify already statted them all) and calls `cbm_store_upsert_file_hash` per row → one implicit SQLite write transaction per file.

**Correctness bonus:** re-statting *after* the run records the post-run mtime, so a file edited mid-index is stamped "current" and the **next** run misses the edit. Using the classify-time stat is both faster and more correct.

### P3 — Artifact re-export on every incremental run

`dump_and_persist` (`src/pipeline/pipeline_incremental.c`): if an artifact exists, every incremental run — even a 1-file edit — reads the whole DB, zstd-compresses it, and rewrites `.codebase-memory/graph.db.zst` (+ metadata), churning the user's git working tree and burning I/O/CPU proportional to the whole graph.

## Proposed fix (3 focused changes)

I'd like to submit these as **three small, stacked PRs** referencing this issue, each under ~500 lines and independently green:

[...]

PR #868, opened 2026-07-05
fix(artifact): reconcile imported hashes against git on bootstrap (1/3)

moofone:
> Tracking issue: **#867** (Refs #867). This is **part 1 of 3** stacked PRs addressing incremental-reindex cost. Review only commit `fix(artifact): reconcile imported hashes against git on bootstrap`. Parts 2 (#869) and 3 (#870) stack on top of this one.

## What & why

After artifact bootstrap, `try_artifact_bootstrap` byte-copies a teammate's DB into the local cache, but the imported `file_hashes` rows carry the **exporter's** `mtime_ns`. A fresh clone stamps every file with checkout time, so `classify_files` (mtime+size only) marks ~every file changed and the first incremental run re-parses the whole repo — the exact run artifact bootstrap exists to make cheap.

## Change

Add `cbm_artifact_reconcile_hashes()` in `src/pipeline/artifact.c` (+ `artifact.h`), called from `try_artifact_bootstrap` immediately after a successful `cbm_artifact_import`. It re-stamps the `file_hashes` rows of files **git reports unchanged** between the artifact's commit and the local working tree, using local `stat()` values. The existing, untouched `classify_files` logic then classifies correctly. **Zero changes to the incremental pipeline itself.**

Reuses two pieces of existing dead code: `cbm_artifact_commit()` (never called in production) and `cbm_store_upsert_file_hash_batch()` (existed, never called in production — reconciliation persists all restamped rows in one transaction).

### Trust gate (the design point flagged in #867)

`cbm_artifact_export` now writes an **optional** `"reconcile_basis":"git-clean-head"` field to the **existing** `artifact.json`, only when export can prove the DB matches a clean checked-out tree at `commit`:
1. `commit` is a valid hex OID (40/64-char) from `git rev-parse HEAD`;
2. the working tree has no tracked/staged/untracked changes outside `.codebase-memory/` (`:(exclude)` pathspec);
3. every `file_hashes` row's on-disk `mtime_ns + size` matches the stored stamp (belt-and-suspenders that catches a stale/swapped DB even when the tree looks clean).

Reconciliation requires the marker and the commit to be present + hex-valid + resolvable locally, and **skips (returns -1)** on any doubt: no git, untrusted/missing marker, non-hex/unknown commit, shallow clone, or any popen/parse uncertainty. A skip leaves rows foreign → today's slow-safe full incremental. **No `schema_version` bump** (older binaries ignore the unknown field).

### Security

`commit` is hex-validated before any command construction; `repo_path` is shell-validated via `cbm_validate_shell_arg` (same pattern as `git_context.c` / `watcher.c`); git output is parsed as NUL-delimited (`-z`) directly (never line-oriented), so paths with newlines/quotes are handled. A non-empty `-z` buffer not ending in NUL is treated as corrupt → skip. New `cbm_popen` call site added to `scripts/security-allowlist.txt` with justification.

## Tests (`tests/test_artifact.c`, following the existing suite)

- `artifact_export_marks_clean_basis` — marker set on a clean export, dropped when the tree is dirty.
- `artifact_reconcile_restamps_unchanged` — full-index → export → clone → modify 2 + add 1 → import → reconcile restamps unchanged rows to local mtime, leaves changed rows foreign (`restamped == 3`).
- `artifact_reconcile_skips_untrusted_metadata` — strip the marker → `-1`, rows untouched (regression guard for old/manual/dirty/stale artifacts).
- `artifact_reconcile_skips_unknown_commit` — valid-hex-but-absent commit → `-1`.
- `artifact_reconcile_skips_without_git` — no `.git` → `-1`.
- null-arg safety on `cbm_artifact_reconcile_hashes`.

All fixtures generated at runtime in tmpdirs; no committed fixtures. `scripts/test.sh` (ASan+UBSan) green locally — full suite passes.

## Benchmark (S2 = bootstrap + first incremental, K = 7 changed)

| N | baseline wall / `changed` | candidate wall / `changed` |
|---|---|---|
| 500 | 635 ms / 503 | **308 ms / 7** |
| 2000 | 4561 ms / 2003 | **735 ms / 7** |
| 8000 | 53691 ms / 8003 | **2563 ms / 7** |

Candidate S2 wall is **4.8 %** of baseline at N=8000 (~21× faster). After S2, candidate node/edge counts exactly equal a from-scratch full index of the same tree (500/2000/8000). Full matrix + raw CSV kept outside the repo.

## Checklist

- [x] C only; conventional commit; DCO signed on every commit.
- [x] `scripts/test.sh` (ASan/UBSan) green.
- [x] New `popen`/`cbm_popen` site justified + added to `scripts/security-allowlist.txt`.
- [ ] `scripts/lint.sh` (clang-tidy/cppcheck/clang-format) — not installed locally; verified ≤100-col + style by hand, CI is the gate.

Seeking maintainer sign-off on the `artifact.json` trust-marker reuse (point (a) in #867). Happy to rework if a different convention is preferred.

### DIFF
--- a/scripts/security-allowlist.txt
+++ b/scripts/security-allowlist.txt
@@ -41,7 +41,7 @@ src/pipeline/pass_githistory.c:popen:via cbm_popen wrapper call
 # ── Pipeline: artifact persistence (git HEAD hash, merge driver config) ────
 src/pipeline/artifact.c:cbm_popen:git rev-parse HEAD for artifact metadata (hardcoded cmd)
 src/pipeline/artifact.c:cbm_popen:git config merge.ours.driver for gitattributes (hardcoded cmd)
-src/pipeline/artifact.c:cbm_popen:git diff/ls-files/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
+src/pipeline/artifact.c:cbm_popen:git diff/ls-files/ls-tree/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
 src/pipeline/artifact.c:popen:via cbm_popen wrapper calls
 
 # ── UI: HTTP server process management ─────────────────────────────────────
--- a/scripts/security-allowlist.txt
+++ b/scripts/security-allowlist.txt
@@ -41,6 +41,7 @@ src/pipeline/pass_githistory.c:popen:via cbm_popen wrapper call
 # ── Pipeline: artifact persistence (git HEAD hash, merge driver config) ────
 src/pipeline/artifact.c:cbm_popen:git rev-parse HEAD for artifact metadata (hardcoded cmd)
 src/pipeline/artifact.c:cbm_popen:git config merge.ours.driver for gitattributes (hardcoded cmd)
+src/pipeline/artifact.c:cbm_popen:git diff/ls-files/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
 src/pipeline/artifact.c:popen:via cbm_popen wrapper calls
 
 # ── UI: HTTP server process management ─────────────────────────────────────
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -3246,7 +3246,13 @@ static void try_artifact_bootstrap(const char *project_name, const char *repo_pa
     project_db_path(project_name, db_buf, sizeof(db_buf));
     if (cbm_file_size(db_buf) < 0 && cbm_artifact_exists(repo_path)) {
         cbm_log_info("index.artifact_bootstrap", "project", project_name);
-        cbm_artifact_import(repo_path, db_buf);
+        if (cbm_artifact_import(repo_path, db_buf) == 0) {
+            /* Reconcile imported (foreign-mtime) hash rows against the local
+             * working tree so the first incremental run classifies by actual
+             * git diff instead of re-parsing ~everything. Best-effort: a -1
+             * return leaves rows foreign and falls back to today's behavior. */
+            (void)cbm_artifact_reconcile_hashes(repo_path, db_buf, project_name);
+        }
     }
 }
 
--- a/src/pipeline/artifact.c
+++ b/src/pipeline/artifact.c
@@ -973,6 +973,10 @@ static void reconcile_add_nul_entries(CBMHashTable *ht, const char *buf, size_t 
  * incremental run then classifies by actual git diff instead of re-parsing
  * ~every file (whose stored mtime_ns is the exporter's, not local).
  *
+ * Only rows tracked at the artifact commit are eligible: files git ignores can
+ * still be indexed (.cbmignore negations, #500), and git cannot vouch for their
+ * content — they stay foreign and are re-parsed.
+ *
  * Returns the number of rows re-stamped, or -1 when reconciliation was skipped
  * (no git / untrusted metadata / unknown or non-hex commit / shallow clone /
  * any popen or parse uncertainty). Best-effort: never fails the import — a -1
@@ -1009,15 +1013,25 @@ int cbm_artifact_reconcile_hashes(const char *repo_path, const char *cache_db_pa
         return CBM_NOT_FOUND;
     }
 
-    /* 3. Build the changed set relative to the artifact commit. NUL-delimited
-     *    (-z) output is parsed directly; line-oriented parsing cannot handle
-     *    paths containing newlines or quotes. */
+    /* 3. Build the changed set relative to the artifact commit, plus the set of
+     *    paths tracked AT that commit. NUL-delimited (-z) output is parsed
+     *    directly; line-oriented parsing cannot handle paths containing
+     *    newlines or quotes. The tracked set exists because git diff/ls-files
+     *    are blind to files git ignores: a gitignored-yet-indexed file (e.g. a
+     *    .cbmignore negation un-skipping a generated dir, #500) would otherwise
+     *    be restamped as "unchanged" while its content is not under git's
+     *    control at all. Only rows tracked at the artifact commit are eligible
+     *    for restamping; everything git can't vouch for stays foreign. */
     char diff_args[CBM_SZ_256];
     snprintf(diff_args, sizeof(diff_args), "diff -z --name-only --no-renames %s --", commit);
+    char lstree_args[CBM_SZ_256];
+    snprintf(lstree_args, sizeof(lstree_args), "ls-tree -r -z --name-only %s --", commit);
     char *diff_out = NULL;
     size_t diff_len = 0;
     char *ls_out = NULL;
     size_t ls_len = 0;
+    char *tracked_out = NULL;
+    size_t tracked_len = 0;
     if (git_capture_full(repo_path, diff_args, &diff_out, &diff_len) != 0) {
         free(commit);
         return CBM_NOT_FOUND;
@@ -1028,14 +1042,22 @@ int cbm_artifact_reconcile_hashes(const char *repo_path, const char *cache_db_pa
         free(commit);
         return CBM_NOT_FOUND;
     }
+    if (git_capture_full(repo_path, lstree_args, &tracked_out, &tracked_len) != 0) {
+        free(diff_out);
+        free(ls_out);
+        free(commit);
+        return CBM_NOT_FOUND;
+    }
     /* Parse-invariant guard: git -z NUL-terminates every entry including the
      * last; a non-empty buffer not ending in NUL means truncated/corrupt output
      * → untrusted → skip rather than risk misclassifying a changed file as
      * unchanged (graph corruption). */
     if ((diff_len > 0 && diff_out[diff_len - 1] != '\0') ||
-        (ls_len > 0 && ls_out[ls_len - 1] != '\0')) {
+        (ls_len > 0 && ls_out[ls_len - 1] != '\0') ||
+        (tracked_len > 0 && tracked_out[tracked_len - 1] != '\0')) {
         free(diff_out);
         free(ls_out);
+        free(tracked_out);
         free(commit);
         return CBM_NOT_FOUND;
     }
@@ -1044,10 +1066,13 @@ int cbm_artifact_reconcile_hashes(const char *repo_path, const char *cache_db_pa
     reconcile_add_nul_entries(changed, diff_out, diff_len);
     reconcile_add_nul_entries(changed, ls_out, ls_len);
     int changed_count = (int)cbm_ht_count(changed);
+    CBMHashTable *tracked = cbm_ht_create(CBM_SZ_64);
+    reconcile_add_nul_entries(tracked, tracked_out, tracked_len);
 
-    /* 4. Restamp every row whose rel_path is NOT in the changed set with local
-     *    stat() values. Changed rows stay foreign → re-parsed; rows whose file
-     *    is missing locally stay foreign → find_deleted_files purges them. */
+    /* 4. Restamp every row whose rel_path IS tracked at the artifact commit and
+     *    is NOT in the changed set, with local stat() values. Changed or
+     *    untracked rows stay foreign → re-parsed; rows whose file is missing
+     *    locally stay foreign → find_deleted_files purges them. */
     cbm_store_t *store = cbm_store_open_path(cache_db_path);
     int restamped = 0;
     int skipped = 0;
@@ -1061,7 +1086,8 @@ int cbm_artifact_reconcile_hashes(const char *repo_path, const char *cache_db_pa
             if (batch) {
                 int batch_n = 0;
                 for (int i = 0; i < stored_count; i++) {
-                    if (cbm_ht_get(changed, stored[i].rel_path)) {
+                    if (!cbm_ht_get(tracked, stored[i].rel_path) ||
+                        cbm_ht_get(changed, stored[i].rel_path)) {
                         continue;
                     }
                     char abs[CBM_SZ_4K];
@@ -1097,8 +1123,10 @@ int cbm_artifact_reconcile_hashes(const char *repo_path, const char *cache_db_pa
     }
 
     cbm_ht_free(changed);
+    cbm_ht_free(tracked);
     free(diff_out);
     free(ls_out);
+    free(tracked_out);
     free(commit);
 
     cbm_log_info("artifact.reconcile", "restamped", itoa_buf(restamped), "changed",
--- a/src/pipeline/artifact.c
+++ b/src/pipeline/artifact.c
@@ -12,6 +12,7 @@ enum {
     ART_ZSTD_BEST = 9,
     ART_RATIO_SCALE = 10, /* multiply ratio by 10 for integer logging */
     ART_NUL = 1,          /* NUL terminator byte */
+    ART_NS_PER_SEC = 1000000000, /* nanoseconds per second (mtime_ns helper) */
 };
 #define ART_BYTES_PER_MB ((size_t)1024 * 1024)
 
@@ -21,6 +22,8 @@ enum {
 #include "foundation/compat_fs.h"
 #include "foundation/compat.h"
 #include "foundation/log.h"
+#include "foundation/str_util.h" /* cbm_validate_shell_arg */
+#include "foundation/hash_table.h" /* CBMHashTable (reconcile changed-set) */
 
 #include "zstd_store.h"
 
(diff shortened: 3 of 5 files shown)


DeusData:
Thanks for splitting this out from #867. Triage: performance/correctness bug, review first in the #868 -> #869 -> #870 stack.

Because this touches artifact import and the security allowlist file scripts/security-allowlist.txt, review will be a bit cautious: git trust boundaries, path handling, unchanged-file detection, and the necessity/scope of the allowlist update all need to be checked. The focused test coverage in this PR is the right shape.

moofone:
Thanks for the triage focus areas — here's a map of each to the code, plus one hardening commit pushed in response.

**New commit `2e49906`** — while re-auditing "unchanged-file detection" against your list, I found one hole and closed it: a file that is **gitignored yet indexed** (possible via a `.cbmignore` negation un-skipping a gitignored dir — the #500 feature) appears in `file_hashes` but is invisible to both `git diff` and `ls-files --others --exclude-standard`, so it would have been restamped as "unchanged" even though git can't vouch for its content. Reconciliation now also captures `git ls-tree -r -z --name-only <commit>` and only restamps rows **tracked at the artifact commit** and not in the changed set; everything git can't vouch for stays foreign → re-parsed. New regression test `artifact_reconcile_skips_untracked_rows` covers exactly this shape. The allowlist line was extended to mention `ls-tree` (same call path, same validation).

Concern → where it's handled:

**Git trust boundaries**
- Repo-controlled metadata never reaches a git command unless: the `reconcile_basis` marker equals the sole trusted value AND `commit` passes `is_hex_oid` (40/64-char hex) — both hard gates at the top of `cbm_artifact_reconcile_hashes`.
- `repo_path` is the only other interpolated component; it goes through `cbm_validate_shell_arg` (rejects `'" ;|&$<>` backticks, newlines, and `\` on POSIX) in `build_git_cmd`, mirroring the existing `git_context.c`/`watcher.c` pattern, plus a `%!^` rejection on Windows.
- Shallow-clone guard: `cat-file -e <commit>^{commit}` must succeed before any diff is trusted.
- Every uncertainty path (no git, popen failure, non-zero exit, OOM, truncated output) returns -1 → rows stay foreign → today's slow-safe full incremental. Tests: `artifact_reconcile_skips_untrusted_metadata`, `_skips_unknown_commit`, `_skips_without_git`, null-arg safety.

**Path handling**
- All three git listings are parsed as NUL-delimited (`-z`) buffers directly — no line-oriented parsing, so paths with newlines/quotes are handled.
- Parse-invariant guard: a non-empty `-z` buffer not ending in NUL is treated as truncated/corrupt → skip (-1), never a partial parse.
- Per-row `snprintf` overflow of `repo_path + rel_path` skips that row (stays foreign), counted in the logged `skipped` field.

**Unchanged-file detection**
- Changed set = `diff -z --name-only --no-renames <commit>` (tracked changes, staged or not, renames shown as add+delete) ∪ `ls-files -z --others --exclude-standard` (untracked). Rows in the changed set stay foreign; rows missing locally stay foreign and are purged by `find_deleted_files`; and per the new commit, rows not tracked at the commit stay foreign.
- The exporter side only emits the trust marker when the tree is clean outside `.codebase-memory/` AND every hash row's on-disk mtime+size matches the DB (`db_hashes_match_disk`) — so a stale/swapped DB can't be marked reconcilable even from a clean tree. Test: `artifact_export_marks_clean_basis`.

**Allowlist necessity/scope**
- One line, covering the three (now four) read-only git subcommands on the single new call path, all funneled through `build_git_cmd`'s validation. No new process-spawn mechanism — same `cbm_popen` wrapper as the two existing artifact.c entries.

Also filed #885 for the cross-path bootstrap limitation noted in #867's deferred list (pre-existing, caps this PR's team-sharing benefit until fixed).

Full ASan/UBSan suite green with the new commit. Happy to adjust the trust-marker convention if you'd prefer a different shape for point (a).

DeusData:
Thanks — this is an excellent writeup and the design (trust-gated reconcile, skip-on-doubt, -z parsing, shell-validated paths) reads exactly right. Two things block the merge:

1. **CI: the new tests fail on all three unix/windows legs** — `git commit` exits 128 inside the test fixtures (`artifact_reconcile_restamps_unchanged`, `artifact_reconcile_skips_untracked_rows`). The bare CI runners have no git identity, so any `git commit` in a tmpdir fixture needs it set explicitly, e.g. `git -C "$repo" -c user.email=test@test -c user.name=test commit ...` (or a one-time `git config` after init). Locally this passes because of your global identity.
2. **Branch staleness** — the branch predates several large merges on main (kernel-perf batch, coverage signal). Please update/rebase onto current main so CI revalidates against reality.

One more nit while you're in there: `runf(...) == 32768` in the assertion output suggests the helper reports raw `system()` status — if it's cheap, decoding with `WEXITSTATUS` would make future failures readable.

Happy to merge once green — parts 2/3 (#869, #870) will follow after this one lands.

DeusData:
Thanks @moofone — reviewed the whole P1→P2→P3 stack (#868/#869/#870) together. The security posture is clean across all three (careful git-shell hygiene: hex-validated OIDs, `cbm_validate_shell_arg`, NUL-delimited parsing, and the security-allowlist entry), and the design is sound and clearly #867-scoped. Three things before any of it can land:

1. **CI is red on Linux + Windows (all three PRs), and it's a test-harness bug — not your production code.** `artifact_reconcile_restamps_unchanged` and `artifact_reconcile_skips_untracked_rows` fail with git exit 128 ("committer identity unknown"): the `clone_to_b` helper in `tests/test_artifact.c` clones repo B but never sets `user.email`/`user.name` in it (only repo A gets `git_init`). Your laptop has a global git identity so it passes locally; CI runners don't. The LeakSanitizer failure is a knock-on (the `proj` string leaks on the early `ASSERT_EQ` abort). Please set a git identity in repo B in the fixture (and free `proj` on the early-return paths) so the P1 guards are honestly green.

2. **All three are CONFLICTING against current `main`** (the kernel-perf work rewrote `pipeline.c`/`pipeline_incremental.c` since you branched). They'll need a rebase, and they must land as an ordered stack (#868 → #869 → #870) since the branches are chained by merge commits.

3. **Two direction decisions are the maintainer's call** and I'm raising them now so there are no surprises after the rebase: (a) #868 persists a `reconcile_basis:"git-clean-head"` trust marker into `artifact.json` and adds a git-reconcile step on every bootstrap — you explicitly asked for sign-off on that convention, and we'll confirm it; (b) #870 changes the default so incremental runs *skip* artifact re-export below thresholds, adding two permanent env-var knobs (`CBM_ARTIFACT_EXPORT_MIN_CHANGES`, `CBM_ARTIFACT_EXPORT_MAX_AGE_S`) plus a drift heuristic — a default-behavior + surface change we'll weigh. #869 (P2, the batched single-stat persist) is clean and in-scope on its own.

Really appreciate the depth here — let's get the harness green + rebased and we'll take it from there.

DeusData:
Reviewed all three PRs in this stack together. First, an apology that matters more than the review: **you asked for design sign-off in #867 on 5 July and have been waiting roughly four weeks.** You did exactly the right thing — raised the trust-model question *before* building on it — and we left you without an answer while main moved underneath you. That is our failure, not yours.

**This is one of the strongest external stacks we have received.** Concretely: a real, measured problem (bootstrap re-parsing nearly everything it just imported), a tracking issue asking for sign-off before landing, and security hygiene on every git shell-out that is better than a lot of our own code — `cbm_validate_shell_arg` on `repo_path`, a hard `is_hex_oid()` gate on the `commit` value from `artifact.json` *before* any command construction, NUL-delimited `-z` parsing with an explicit truncation guard so filenames containing newlines or quotes are handled. You also reused two existing dead APIs instead of inventing new ones, caught a #500-shaped hole yourself, and fixed it with its own regression test. The ~21× bootstrap win with exact graph-equivalence is the kind of evidence that makes a reviewer's job easy.

**The stack structure, confirmed by commit-OID comparison:** these literally stack. #869 contains all of #868's commits, and #870 contains all of both. Since all three target `main`, merging #870 would merge the entire stack at once. So the order is forced — 868, then 869, then 870 — with a re-diff after each.

---

**Two decisions are the maintainer's, and both are genuinely yours to have asked about.**

**1. The `reconcile_basis` trust marker (#868).** This is the one you raised in #867, and the review agrees it is a real trust-model question rather than a code-quality one. The marker is *self-attested by the artifact producer* and unverifiable by the importer. Today a tampered artifact gets accidentally scrubbed on first run, because every file re-parses on the mtime mismatch. With reconcile, files git reports as unchanged are not re-parsed — so poisoned graph rows in a crafted artifact would survive in exactly those files. It stays within the existing artifact trust model (importing already byte-copies a foreign DB from the same trust domain as the repo), and your tracked-set and local-git-diff gates are sound. But it removes an accidental safety property, and that trade — scrubbing versus the bootstrap win — is the maintainer's call.

**2. The debounce policy and env surface (#870).** `CBM_ARTIFACT_EXPORT_MIN_CHANGES` and `CBM_ARTIFACT_EXPORT_MAX_AGE_S` would be permanent public configuration, and the 10 / 24 h defaults are themselves a policy choice about acceptable staleness in a shared artifact.

I have put both in front of the maintainer explicitly. **Please do not spend time rebasing until those are answered** — the stack's fate hangs on the first one, and it would be unfair to have you resolve four weeks of drift for a design that might not land.

---

**Meanwhile, three things worth knowing.**

**Your CI red is not your production code.** `clone_to_b()` does not set a git identity in the clone, so `git commit` exits 128 on identity-less runners, failing two reconcile tests — and the assert-fail path then leaks `proj` (68 bytes), which reddens the ASan leg a second way. Both are two-line test fixes.

**Main moved under you, and it changes #869.** Commit `b79c6da` made hash persistence all-or-nothing: `persist_hashes` now returns `int`, and any failed upsert rejects the whole staging generation, because atomic publication makes preserving the complete previous generation safer than installing partial metadata. Your row-at-a-time fallback was written against the old "partial preservation beats total loss" contract and now points backwards. The irony is that your single-transaction batch fits the *new* model better than what you wrote — the right rework keeps the classify-time stamps and the batch upsert, returns the batch return code, and **drops the fallback loop entirely**. I verified the transactionality concern I most cared about, incidentally: a crash mid-batch rolls back, old stamps survive, and the next run re-parses. That is the safe direction and it converges.

**#870 needs re-targeting**: the incremental re-export moved to `export_after_publish` in `pipeline.c`.

One test note for #869 when it does get reworked: `incremental_persist_batch_roundtrip` passes unmodified on main, because without a mid-run edit the post-run re-stat produces the same mtime. It is a fine guard for the batch path, but the headline claim — that the old code masks a mid-run edit — currently ships untested. A seam that mutates the file between classify and persist would bind it.

**And an offer.** Given all three conflict with a heavily-reworked main, if you would rather not carry that rebase, we can distill the approved design points onto main ourselves with `Co-Authored-By` credit to you. Entirely your preference — the work is yours either way, and I would rather ask than assume.

Answers on #867 coming.

DeusData:
Decision on the trust question, which is what this PR has been waiting on: **accepted — go ahead with `reconcile_basis` as designed.**

The reasoning, since you raised the concern honestly rather than hoping nobody noticed: importing an artifact already means adopting somebody else's entire prebuilt graph — every node, every edge, every hash row. If that artifact is hostile, declining to re-stat a handful of files is nowhere near the most interesting thing it can do to you. The marker doesn't create a new trust boundary; it operates inside one we crossed the moment we supported artifact import at all. Requiring the importer to independently verify it would mean content-hashing every file, which is precisely the work this PR exists to avoid.

What I do want is for that to be **stated rather than implied**. Please add a line to the import path's documentation (and ideally a comment at the reconcile site) saying plainly: an imported artifact is trusted input, and `reconcile_basis` lets it assert that git-unchanged files need no re-scan. Someone auditing this in a year should find the trade-off written down, not have to infer it from the absence of a check.

Two other things before this can land, neither of them about the design:

- **CI is red on Linux and Windows.** From the logs it looks like the test fixture clones a repo without setting a git identity, so the commit fails and an early return leaks — a fixture bug rather than production code, but it needs fixing before I can read the matrix.
- **The branch conflicts with `main`.** Now that the design question is settled, please go ahead and rebase — sorry for asking you to hold, but I didn't want you redoing a rebase against a decision that might have gone the other way.

Decisions on the rest of your stack are on their respective PRs (#869 needs rework against a persist-contract change that landed on `main`; #870's debounce is accepted with tighter defaults). They still need to merge in order, so this one goes first.

Thank you for the follow-on issue you filed for the gitignored-but-indexed gap, and for the extra regression test you added mid-review without being asked. The tests here genuinely exercise the export → clone → modify → import → reconcile path rather than mocking around it, which is why this was reviewable at all.

DeusData:
@moofone — thank you for this, and I owe you an apology for the review latency. Seven weeks is far too long, especially after you replied to the first round within a day and pushed hardening commit `2e49906` for a hole you found yourself, unprompted.

Taking you up on the offer I made earlier: I'm landing this design on current main myself, with `Co-authored-by:` credit to you. The reason is drift, not quality — since your branch point the six files you touched have taken **9,652 insertions across 60 commits** (`pipeline_incremental.c` alone +2,385, `mcp.c` +8,081), and with merge-commit history `rebaseable` is false. Asking you to absorb that after a wait that was my fault isn't a reasonable thing to do.

**Three things you should know, because two of them are my errors.**

**1. The Windows red is a production bug, and I misdiagnosed it twice as your fixture.** It isn't the git identity in `clone_to_b`. The single-quoted pathspecs — `':(exclude).codebase-memory'` and `cat-file -e '%s^{commit}'` — don't survive `cmd.exe`, which `cbm_popen` uses on Windows, and `^` is its escape character. This file's own comment at `artifact.c:218-222` states the rule I failed to apply when reviewing: *"Callers then use DOUBLE quotes (honored by both POSIX sh and cmd.exe, unlike single quotes on cmd.exe)."* The tell was there all along — Windows CI fails on the **production** assert in `artifact_export_marks_clean_basis` while Linux and macOS pass it. The fix I prescribed would not have turned Windows green, and you'd have burned a cycle finding that out.

**2. One real defect worth naming**, since it's subtle and the rest of the code is careful: `cbm_ht_create()` is unchecked at the two call sites around `artifact.c:1064`/`:1068`. If `changed` comes back NULL while `tracked` succeeds, every tracked row gets restamped as unchanged — **including genuinely changed files**. It's the one place the code inverts its own skip-on-doubt invariant: in `classify_files` a dropped entry means "re-parse" (safe), here it means "unchanged" (not). Landing it fail-closed.

**3. Your trust analysis was right and I'm writing it down.** The protection today isn't just "we trust the artifact anyway" — foreign mtimes force a re-parse that *auto-scrubs* a poisoned artifact at an unpredictable time, and `reconcile_basis` converts that transient, self-healing exposure into a durable one gated on a producer-written string. The acceptance still stands, but that reasoning belongs in a doc comment rather than in a review thread, which is what you asked for.

I'll close this citing the merge SHA once it lands. Your tests are being ported first, before the production code — they encode nuance that's easy to lose in a rewrite.

DeusData:
Landed. **On main as of `3de05cd6`** (PR #1905), with `Co-authored-by:` credit to you.

@moofone — thank you, and here is a straight account, including two things I got wrong on this thread.

**I misdiagnosed the Windows failure twice as your fixture, and it was neither that nor what I then claimed.** The single-quoted git pathspecs were a genuine production bug — `cbm_popen` runs `cmd.exe /c` on Windows, which does not honour single quotes — and that is fixed. But fixing it did **not** turn Windows green, because there was a second, unrelated defect underneath: `file_hashes.mtime_ns` is stamped from Windows `FILETIME` (100-nanosecond ticks) while the trust check re-read the same files with `_wstat64` and encoded whole seconds. The two can only agree when a write lands exactly on a second boundary, so `reconcile_basis` was never written on Windows. Same symptom, different cause, invisible on POSIX where both paths read nanoseconds.

**A third defect made the feature inert everywhere, not just on Windows.** `file_hashes` carries synthetic rows under `.codebase-memory/.semantic-input/` with no file on disk; the trust check `stat()`ed them and always failed. So `reconcile_basis` was never written on *any* platform. That is main-drift your PR predates — and it means the thread's "passes on Linux and macOS" reading was itself misleading.

Also fixed: `cbm_ht_create()` was unchecked at two sites, and the polarity mattered. If the `changed` table came back NULL while `tracked` succeeded, every tracked row would be restamped as unchanged — **including genuinely changed files**. It was the one place the code inverted its own skip-on-doubt invariant. It now fails closed.

**Your trust analysis was right and is now written into the code**, at both the import path and the reconcile site: the protection today is not merely "we trust the artifact anyway" — foreign mtimes force a re-parse that auto-scrubs a poisoned artifact at an unpredictable time, and `reconcile_basis` converts that transient, self-healing exposure into a durable one gated on a producer-written string. That belonged in the source, not a review thread.

**One limit you should know about**, because it caps what this buys today: `cbm_project_name_from_path` derives the project name from the *entire absolute path*, so a teammate's clone at a different path matches zero artifact rows and this reconcile finds nothing to do. That is your own #885, and the ~21x figure is in-benchmark until it is fixed. This lands the mechanism so #885 becomes a small unlock rather than a large one.

Diagnosing this also turned up a broader problem, filed as **#1917**: the same encoding split means the incremental classifier compares whole-second `stat` values against FILETIME-tick stored values, so **every incremental reindex on Windows re-parses the whole project**. It fails safe, which is why it went unnoticed — but Windows has effectively never had incremental indexing. Your PR is what surfaced it.

#869 and #870 were closed separately: #869's idea shipped independently and more strongly on main, and #870's design is preserved as #1884. Thank you for all three.

PR #869, opened 2026-07-05
refactor(incremental): batch + single-stat persist_hashes (2/3)

moofone:
> Tracking issue: **#867** (Refs #867). This is **part 2 of 3**, stacked on #868. Review only commit `refactor(incremental): persist hashes in one batch from classify-time stamps` (the P1 commit is tracked by #868).

## What & why

`persist_hashes` (`src/pipeline/pipeline_incremental.c`) re-`stat()`ed every discovered file **after** the run and upserted hash rows one at a time (one implicit SQLite write transaction per file). Two problems:

1. **Wasted work** — `classify_files` had already `stat()`ed every file with a stored hash.
2. **Correctness bug** — re-stat'ing *after* the run records the post-run mtime, so a file edited **mid-index** is stamped "current" and the **next** run misses the edit. Using the classify-time stat is both faster and more correct.

## Change

- `classify_files` now outputs a `cbm_file_stamp_t` array (one `stat()` per file at classify time, including for new files).
- `persist_hashes` builds one `cbm_file_hash_t` array from `files[]+stamps[]` plus `mode_skipped[]` and upserts it via the **existing-but-unused** `cbm_store_upsert_file_hash_batch` (single `BEGIN…COMMIT`).
- On batch failure (it rolls back on first row error) it falls back to a row-at-a-time loop so one bad row still can't lose all persistence — preserving the documented "partial preservation beats total loss" contract.
- `stamps` flows `classify_files → cbm_pipeline_run_incremental → dump_and_persist → persist_hashes` (all static, file-local signature changes), freed at every return path (noop / load_db_failed / normal).

## Tests (`tests/test_artifact.c`)

- `incremental_persist_batch_roundtrip` — exercises the batch path via the real pipeline route (full index → modify → incremental reindex → noop), asserts the persisted mtime equals the classify-time mtime and the subsequent run classifies no-change.
- The existing incremental suite (`incr_modify_file` / `incr_add_file` / `incr_delete_file` / `incr_noop_reindex`, ~5900 cases) is the regression guard.

`scripts/test.sh` (ASan/UBSan) green locally.

## Benchmark (S3 = steady-state incremental, K = 5 modified)

| N | baseline wall (export ran) | candidate wall (export skipped*) |
|---|---|---|
| 500 | 297 ms | 226 ms |
| 2000 | 673 ms | 540 ms |
| 8000 | 2290 ms | 1755 ms |

Candidate ≤ baseline at every size. (*the export-skip is P3, tracked by part 3; P2's contribution is the batched, single-stat persist.) S4 (noop) and S1 (full) unchanged within noise.

## Checklist

- [x] C only; conventional commit; DCO signed.
- [x] `scripts/test.sh` (ASan/UBSan) green; no regressions in the incremental suite.
- [x] No new popen/system/network calls (no security-allowlist change needed).
- [ ] `scripts/lint.sh` — not installed locally; verified by hand, CI is the gate.

Commit message notes the mid-run-edit correctness fix.

### DIFF
--- a/scripts/security-allowlist.txt
+++ b/scripts/security-allowlist.txt
@@ -41,7 +41,7 @@ src/pipeline/pass_githistory.c:popen:via cbm_popen wrapper call
 # ── Pipeline: artifact persistence (git HEAD hash, merge driver config) ────
 src/pipeline/artifact.c:cbm_popen:git rev-parse HEAD for artifact metadata (hardcoded cmd)
 src/pipeline/artifact.c:cbm_popen:git config merge.ours.driver for gitattributes (hardcoded cmd)
-src/pipeline/artifact.c:cbm_popen:git diff/ls-files/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
+src/pipeline/artifact.c:cbm_popen:git diff/ls-files/ls-tree/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
 src/pipeline/artifact.c:popen:via cbm_popen wrapper calls
 
 # ── UI: HTTP server process management ─────────────────────────────────────
--- a/scripts/security-allowlist.txt
+++ b/scripts/security-allowlist.txt
@@ -41,6 +41,7 @@ src/pipeline/pass_githistory.c:popen:via cbm_popen wrapper call
 # ── Pipeline: artifact persistence (git HEAD hash, merge driver config) ────
 src/pipeline/artifact.c:cbm_popen:git rev-parse HEAD for artifact metadata (hardcoded cmd)
 src/pipeline/artifact.c:cbm_popen:git config merge.ours.driver for gitattributes (hardcoded cmd)
+src/pipeline/artifact.c:cbm_popen:git diff/ls-files/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
 src/pipeline/artifact.c:popen:via cbm_popen wrapper calls
 
 # ── UI: HTTP server process management ─────────────────────────────────────
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -3246,7 +3246,13 @@ static void try_artifact_bootstrap(const char *project_name, const char *repo_pa
     project_db_path(project_name, db_buf, sizeof(db_buf));
     if (cbm_file_size(db_buf) < 0 && cbm_artifact_exists(repo_path)) {
         cbm_log_info("index.artifact_bootstrap", "project", project_name);
-        cbm_artifact_import(repo_path, db_buf);
+        if (cbm_artifact_import(repo_path, db_buf) == 0) {
+            /* Reconcile imported (foreign-mtime) hash rows against the local
+             * working tree so the first incremental run classifies by actual
+             * git diff instead of re-parsing ~everything. Best-effort: a -1
+             * return leaves rows foreign and falls back to today's behavior. */
+            (void)cbm_artifact_reconcile_hashes(repo_path, db_buf, project_name);
+        }
     }
 }
 
--- a/src/pipeline/artifact.c
+++ b/src/pipeline/artifact.c
@@ -973,6 +973,10 @@ static void reconcile_add_nul_entries(CBMHashTable *ht, const char *buf, size_t 
  * incremental run then classifies by actual git diff instead of re-parsing
  * ~every file (whose stored mtime_ns is the exporter's, not local).
  *
+ * Only rows tracked at the artifact commit are eligible: files git ignores can
+ * still be indexed (.cbmignore negations, #500), and git cannot vouch for their
+ * content — they stay foreign and are re-parsed.
+ *
  * Returns the number of rows re-stamped, or -1 when reconciliation was skipped
  * (no git / untrusted metadata / unknown or non-hex commit / shallow clone /
  * any popen or parse uncertainty). Best-effort: never fails the import — a -1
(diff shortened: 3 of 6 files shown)


DeusData:
Thanks for keeping part 2 scoped. Triage: performance/correctness bug in the incremental path, stacked after #868.

Review should verify that classify-time stamps really close the mid-index edit miss without widening the race window, and that the batch write path keeps failure behavior conservative. Since this also carries the #868 base changes, we should review/merge in stack order.

moofone:
On the two review questions from triage:

**Does the classify-time stamp close the mid-index miss without widening the race window?** Yes — the race can only narrow, by construction. The stamp for each file is taken *before* its content is parsed. So for any edit relative to the stamp time:

- **Edit before the stat** → the parse (later) sees the edited content, and the stamp reflects the edited mtime. Consistent.
- **Edit after the stat** (mid-index, the old bug's window) → the parsed content is *newer than the stamp*, so on the next run the file's actual mtime differs from the stored stamp → re-parse. The edit is caught one run later, never lost.

The old post-run re-stat had the opposite property: it stamped the *post-edit* mtime against *pre-edit* parsed content, so the next run compared equal and the edit was silently missed. There is no input ordering under which the new scheme misses an edit that the old one caught; the possible outcomes are "consistent" or "conservative re-parse".

**Does the batch path keep failure behavior conservative?** The batch upsert is a single `BEGIN…COMMIT` that rolls back on the first row error. On any non-OK return, `persist_hashes` logs `incremental.persist_batch_failed` and falls back to `persist_hashes_row_by_row` — the historical loop, preserved verbatim including its per-scope warnings — so one bad row still cannot lose all hash persistence (the documented "partial preservation beats total loss" contract). The same fallback covers batch-array OOM. A file that couldn't be stat'ed at classify time is skipped in both paths → re-parsed next run (conservative direction).

Test `incremental_persist_batch_roundtrip` drives the batch path through the real pipeline route and asserts the persisted mtime equals the classify-time mtime; the existing incremental suite (`incr_modify_file`/`incr_add_file`/`incr_delete_file`/`incr_noop_reindex`) is the regression guard.

Branch has been updated with the #868 base changes merged in (this PR's reviewable commit is unchanged: `e9d4190`).

DeusData:
Reviewed as part of the stack — the full write-up is on #868, since these three literally stack (this PR contains all of #868's commits, and #870 contains both). Please read that one first; this note covers what is specific to #869.

**The core idea is right and still wanted.** Statting every file once and emitting a `cbm_file_stamp_t` array, then writing one batch through the previously-dead `cbm_store_upsert_file_hash_batch`, turns 2N stats into N and N implicit write transactions into one. And your reasoning about *classify-time* stamps is the part I want to credit specifically: stamping at classify time rather than post-run means an edit landing between the stat and the parse yields a stamp older than the file, so the next run re-parses. That is over-parsing rather than staleness — the safe direction. The bug you identified in the old code is real: a post-run re-stat stamps content that was never parsed.

I went looking for the failure mode that would make this dangerous — a crash mid-batch leaving hashes that claim content the graph never got — and it cannot happen. `cbm_store_upsert_file_hash_batch` does a real `BEGIN`, rolls back on the first error, and commits otherwise; a crash rolls back, the old stamps survive, and the next run re-classifies those files as changed. `dump_and_persist` also still dumps the graph before persisting hashes, so a crash cannot produce a hash newer than the content it describes.

**What changed under you, and it is the one real rework.** Commit `b79c6da` on main replaced the persist contract: `persist_hashes` now returns `int`, and any failed upsert rejects the entire staging generation, because atomic publication makes preserving the complete previous generation safer than installing partial metadata. Your row-at-a-time fallback was written against the old contract and now moves against that design — it would reintroduce exactly the partial hash-row states main just eliminated.

The good news is that your batch fits the new model *better* than the fallback did. The rework is: keep the classify-time stamps and the batch upsert, return the batch's return code, and **delete the fallback loop**.

**One test gap worth closing in the same pass.** `incremental_persist_batch_roundtrip` passes unmodified on current main — without a mid-run edit, the post-run re-stat produces the same mtime and run 3 is still a no-op. It is a perfectly good regression guard for the batch path, but the headline correctness claim (that the old code masks a mid-run edit) has no failing test behind it. A seam that mutates the file between classify and persist would bind it, and this repo's opt-in test-seam pattern is the natural fit.

**Please hold the rebase for now** — #868 carries a trust-model question the maintainer needs to answer first, and since this PR contains #868, its fate follows. Details and the distillation offer are on #868.

DeusData:
The stack's design question is resolved — #868's `reconcile_basis` is accepted (details on that thread), so this one is unblocked and you can rebase.

Two things need rework here, both caused by `main` moving rather than by anything wrong in your original reasoning:

**1. The row-at-a-time fallback is now stale.** Since you wrote this, a change landed making hash persistence atomic — all-or-nothing. Your fallback loop retries individual rows when the batch fails, which under the new contract means retrying after a rollback that already reverted everything, and can leave a partially-persisted state the new design specifically rules out. Please drop the fallback and propagate the batch's return code instead. Simpler, and it matches the invariant the persist layer now guarantees.

**2. The test doesn't yet bind the headline claim.** `incremental_persist_batch_roundtrip` passes on unmodified `main` too, because nothing in it mutates a file mid-run — so it exercises the batching but not the bug that motivated the change. The interesting claim in this PR is that the *old* post-run re-stat could miss an edit that happened while indexing was in flight; the test that would prove it needs a seam to modify a file between classify and persist. The repo has test-seam precedent for exactly this shape (`CBM_ENABLE_TEST_SEAMS` / the incremental test API), so it shouldn't need new machinery.

That second point is a compliment as much as a request: the batching win is real and measurable on its own, but the correctness half is the more valuable claim and right now it's unproven. Worth having a test that fails without your change, so the next person who touches stamp timing finds out immediately.

Merge order is still #868 → #869 → #870, so this waits on #868's CI going green.

DeusData:
@moofone — closing this one, and I want to be precise about why, because **your reasoning was right and it was not rejected.**

Main solved this independently while your PR was waiting, and in two places:

- The batch upsert you argued for is in production at `src/pipeline/pipeline.c:1763` (`cbm_store_upsert_file_hash_batch(store, generation->manifest, …)`), with `tests/test_store_nodes.c:1045` documenting the migration.
- The mid-index-edit correctness problem you identified — the one you caught by reasoning about post-run re-stat masking edits — is fixed *more strongly* than either of us proposed, by `semantic_manifest_hash_file` (`pipeline_incremental.c:164-200`): a content SHA-256 with a stat→read→stat torn-read guard.

You spotted a real correctness gap and were ahead of the fix. That it arrived by another route is my fault for the seven-week wait, not a comment on the analysis.

The reason I can't merge it as-is: `persist_hashes` no longer exists in that file, and the diff sets `batch[n].sha256 = ""` with the comment "column not yet populated" — which was true when you wrote it and isn't now. Merged today it would **blank real content hashes**, so this would regress main rather than improve it.

Thank you for it, genuinely. #868's design is being landed with `Co-authored-by:` credit to you.

PR #870, opened 2026-07-05
feat(incremental): debounce artifact auto re-export (3/3)

### LINKED ISSUE #867: Incremental reindex costs ≈ a full index, especially after artifact bootstrap

moofone:
> Tracking issue: **#867** (Closes #867). This is **part 3 of 3**, stacked on #868 and #869. Review only commit `feat(incremental): debounce artifact auto re-export` (the P1/P2 commits are tracked by those PRs). Merging this completes the set from #867.

## What & why

`dump_and_persist` (`src/pipeline/pipeline_incremental.c`) re-exported the artifact on **every** incremental run — even a 1-file edit read the whole DB, zstd-compressed it, and rewrote `.codebase-memory/graph.db.zst` (+ metadata), churning the user's git working tree and burning O(graph) I/O/CPU.

## Change

Export only when:
- `changed_total` (changed + deleted) `>= CBM_ARTIFACT_EXPORT_MIN_CHANGES` (default **10**), OR
- the artifact's `indexed_at` is older than `CBM_ARTIFACT_EXPORT_MAX_AGE_S` (default **24 h**).

- Age comes from `artifact.json`'s `indexed_at` via a new `cbm_artifact_age_seconds()` helper in `artifact.c` — parsed with a portable UTC epoch computation (days-from-civil; no libc `timegm`/timezone dependency). **Not** the `.zst` filesystem mtime, because checkout/touch can make an old artifact look fresh.
- Env override `CBM_ARTIFACT_EXPORT_MIN_CHANGES=0` restores per-run export (today's behavior).
- The skip is logged (`artifact.export_skipped changed=<n> age_s=<a>`) so staleness stays observable — "no silent caps".

**Safety argument for staleness:** with P1's git reconcile (#868) in place, a stale artifact merely means bootstrap reconciles a slightly larger git diff — cost degrades gracefully, correctness is unaffected.

## Tests (`tests/test_artifact.c`)

- `artifact_export_debounced` — 1 change (below threshold) + fresh artifact → the `.zst` inode mtime is bit-identical and `artifact.export_skipped` is logged.
- `artifact_export_forced_by_env` — `CBM_ARTIFACT_EXPORT_MIN_CHANGES=0` → the `.zst` is re-written.

Both drive a real incremental run via the pipeline route. `scripts/test.sh` (ASan/UBSan) green locally.

## Benchmark (artifact export flag, K = 5)

| Scenario | baseline `artifact_export` | candidate `artifact_export` |
|---|---|---|
| S3 incremental @ 500/2000/8000 | **1** (always re-exported) | **0** (debounced) |

Candidate S3 wall drops accordingly (8000: 2290 ms → 1755 ms; the rest is P2's batch persist). S4 (noop) and S1 (full) unaffected.

## Checklist

- [x] C only; conventional commit; DCO signed.
- [x] `scripts/test.sh` (ASan/UBSan) green.
- [x] No new popen/system/network calls.
- [ ] `scripts/lint.sh` — not installed locally; verified by hand, CI is the gate.

Commit message notes the staleness-observability line. With all three PRs merged, the full before/after matrix from #867 holds (S2 ~21× faster at N=8000, graph-equivalence exact).

### DIFF
--- a/scripts/security-allowlist.txt
+++ b/scripts/security-allowlist.txt
@@ -41,7 +41,7 @@ src/pipeline/pass_githistory.c:popen:via cbm_popen wrapper call
 # ── Pipeline: artifact persistence (git HEAD hash, merge driver config) ────
 src/pipeline/artifact.c:cbm_popen:git rev-parse HEAD for artifact metadata (hardcoded cmd)
 src/pipeline/artifact.c:cbm_popen:git config merge.ours.driver for gitattributes (hardcoded cmd)
-src/pipeline/artifact.c:cbm_popen:git diff/ls-files/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
+src/pipeline/artifact.c:cbm_popen:git diff/ls-files/ls-tree/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
 src/pipeline/artifact.c:popen:via cbm_popen wrapper calls
 
 # ── UI: HTTP server process management ─────────────────────────────────────
--- a/scripts/security-allowlist.txt
+++ b/scripts/security-allowlist.txt
@@ -41,6 +41,7 @@ src/pipeline/pass_githistory.c:popen:via cbm_popen wrapper call
 # ── Pipeline: artifact persistence (git HEAD hash, merge driver config) ────
 src/pipeline/artifact.c:cbm_popen:git rev-parse HEAD for artifact metadata (hardcoded cmd)
 src/pipeline/artifact.c:cbm_popen:git config merge.ours.driver for gitattributes (hardcoded cmd)
+src/pipeline/artifact.c:cbm_popen:git diff/ls-files/cat-file for bootstrap hash reconciliation (commit hex-validated via is_hex_oid, repo path shell-validated via cbm_validate_shell_arg)
 src/pipeline/artifact.c:popen:via cbm_popen wrapper calls
 
 # ── UI: HTTP server process management ─────────────────────────────────────
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -3246,7 +3246,13 @@ static void try_artifact_bootstrap(const char *project_name, const char *repo_pa
     project_db_path(project_name, db_buf, sizeof(db_buf));
     if (cbm_file_size(db_buf) < 0 && cbm_artifact_exists(repo_path)) {
         cbm_log_info("index.artifact_bootstrap", "project", project_name);
-        cbm_artifact_import(repo_path, db_buf);
+        if (cbm_artifact_import(repo_path, db_buf) == 0) {
+            /* Reconcile imported (foreign-mtime) hash rows against the local
+             * working tree so the first incremental run classifies by actual
+             * git diff instead of re-parsing ~everything. Best-effort: a -1
+             * return leaves rows foreign and falls back to today's behavior. */
+            (void)cbm_artifact_reconcile_hashes(repo_path, db_buf, project_name);
+        }
     }
 }
 
--- a/src/pipeline/artifact.c
+++ b/src/pipeline/artifact.c
@@ -973,6 +973,10 @@ static void reconcile_add_nul_entries(CBMHashTable *ht, const char *buf, size_t 
  * incremental run then classifies by actual git diff instead of re-parsing
  * ~every file (whose stored mtime_ns is the exporter's, not local).
  *
+ * Only rows tracked at the artifact commit are eligible: files git ignores can
+ * still be indexed (.cbmignore negations, #500), and git cannot vouch for their
+ * content — they stay foreign and are re-parsed.
+ *
  * Returns the number of rows re-stamped, or -1 when reconciliation was skipped
  * (no git / untrusted metadata / unknown or non-hex commit / shallow clone /
  * any popen or parse uncertainty). Best-effort: never fails the import — a -1
@@ -1009,15 +1013,25 @@ int cbm_artifact_reconcile_hashes(const char *repo_path, const char *cache_db_pa
         return CBM_NOT_FOUND;
     }
 
-    /* 3. Build the changed set relative to the artifact commit. NUL-delimited
-     *    (-z) output is parsed directly; line-oriented parsing cannot handle
-     *    paths containing newlines or quotes. */
+    /* 3. Build the changed set relative to the artifact commit, plus the set of
+     *    paths tracked AT that commit. NUL-delimited (-z) output is parsed
+     *    directly; line-oriented parsing cannot handle paths containing
+     *    newlines or quotes. The tracked set exists because git diff/ls-files
+     *    are blind to files git ignores: a gitignored-yet-indexed file (e.g. a
+     *    .cbmignore negation un-skipping a generated dir, #500) would otherwise
+     *    be restamped as "unchanged" while its content is not under git's
+     *    control at all. Only rows tracked at the artifact commit are eligible
+     *    for restamping; everything git can't vouch for stays foreign. */
     char diff_args[CBM_SZ_256];
     snprintf(diff_args, sizeof(diff_args), "diff -z --name-only --no-renames %s --", commit);
+    char lstree_args[CBM_SZ_256];
+    snprintf(lstree_args, sizeof(lstree_args), "ls-tree -r -z --name-only %s --", commit);
     char *diff_out = NULL;
     size_t diff_len = 0;
     char *ls_out = NULL;
     size_t ls_len = 0;
+    char *tracked_out = NULL;
+    size_t tracked_len = 0;
     if (git_capture_full(repo_path, diff_args, &diff_out, &diff_len) != 0) {
         free(commit);
         return CBM_NOT_FOUND;
@@ -1028,14 +1042,22 @@ int cbm_artifact_reconcile_hashes(const char *repo_path, const char *cache_db_pa
         free(commit);
         return CBM_NOT_FOUND;
     }
+    if (git_capture_full(repo_path, lstree_args, &tracked_out, &tracked_len) != 0) {
+        free(diff_out);
+        free(ls_out);
+        free(commit);
+        return CBM_NOT_FOUND;
+    }
     /* Parse-invariant guard: git -z NUL-terminates every entry including the
      * last; a non-empty buffer not ending in NUL means truncated/corrupt output
      * → untrusted → skip rather than risk misclassifying a changed file as
      * unchanged (graph corruption). */
     if ((diff_len > 0 && diff_out[diff_len - 1] != '\0') ||
-        (ls_len > 0 && ls_out[ls_len - 1] != '\0')) {
+        (ls_len > 0 && ls_out[ls_len - 1] != '\0') ||
+        (tracked_len > 0 && tracked_out[tracked_len - 1] != '\0')) {
         free(diff_out);
         free(ls_out);
+        free(tracked_out);
         free(commit);
         return CBM_NOT_FOUND;
     }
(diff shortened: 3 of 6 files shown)


DeusData:
Thanks for the final incremental-cost piece. Triage: performance/correctness bug under #867/#593, stacked after #868 and #869.

Review focus: debounce thresholds need predictable defaults and override behavior, artifact age parsing needs portable time handling, and re-export suppression must not leave stale artifacts after materially changed indexes. We will review this after the earlier stack layers.

moofone:
Responding to the three triage focus areas — the staleness one prompted a hardening commit.

**New commit `bbdb97e` — no stale artifacts after materially changed indexes.** You're right that a per-run threshold alone has a hole: many consecutive small runs (each < 10 changes) could leave the artifact stale for up to the 24h age cap despite a materially changed index. The debounce now also computes **cumulative drift** — the number of discovered files whose classify-time mtime is newer than the artifact's `indexed_at` — from the stamps already in hand (zero extra I/O; the data is the #869 classify-time stamp array). Export triggers when *this run's* changes ≥ threshold, OR **cumulative drift ≥ threshold**, OR age ≥ max. Files touched since the last export keep local mtimes newer than `indexed_at`, so drift accumulates across runs and is stateless (no new persisted counters). Over-counting (touched-but-unchanged files) only exports more often — the safe direction. Test `artifact_export_forced_by_cumulative_drift` drives two consecutive below-threshold runs and asserts the second one exports.

**Predictable defaults and overrides.**
- Defaults: `MIN_CHANGES = 10`, `MAX_AGE = 24h`, both named enum constants.
- `CBM_ARTIFACT_EXPORT_MIN_CHANGES=0` restores today's per-run export; any non-numeric/negative value falls back to the default (strict `strtol` parse).
- New in this commit: `CBM_ARTIFACT_EXPORT_MAX_AGE_S` override with the same parse rules, so both knobs are tunable symmetrically. Test `artifact_export_forced_by_max_age_env` covers it.
- Every skip logs `artifact.export_skipped changed=<n> drift=<d> age_s=<a>` so staleness stays observable.

**Portable time handling.** Age comes from `artifact.json`'s `indexed_at` (not the `.zst` fs mtime — checkout/touch can make an old artifact look fresh), parsed by `parse_iso8601_utc`: a fixed-format `sscanf` + days-from-civil epoch computation, so no `timegm` (nonstandard), no `mktime` timezone dependence, identical behavior on POSIX and Windows. Range-checked fields; any mismatch → -1 → the age trigger simply doesn't fire (conservative: falls back to the change/drift triggers).

Worth noting the safety asymmetry: with #868's reconcile, a stale artifact is a *cost* issue (slightly larger git diff to reconcile at bootstrap), never a *correctness* issue — the importing side classifies by actual git state, not by artifact freshness.

Full ASan/UBSan suite green with the new commit.

DeusData:
Reviewed as part of the stack — the full write-up is on #868, since these three literally stack (this PR contains all of #868's and #869's commits, so merging it would merge the whole stack at once). Please read that one first; this note covers what is specific to #870.

**Mechanically the code is fine.** The debounce logic itself is sound, and the deliberate bounded-staleness framing in your description is honest about what it trades.

**What holds it is policy, not implementation**, and it needs the maintainer rather than me:

- `CBM_ARTIFACT_EXPORT_MIN_CHANGES` and `CBM_ARTIFACT_EXPORT_MAX_AGE_S` would be **permanent public configuration** — env knobs are a one-way door here, and the 10 / 24 h defaults are themselves a policy choice about how stale a shared artifact may be.
- More fundamentally: do we want a debounced shared artifact at all, versus always-fresh? That is a product question about what teammates pulling an artifact are entitled to assume.

I have put both in front of the maintainer along with the trust-model question from #868.

**Two technical notes for whenever it resumes:**

1. **The gate needs re-targeting.** Incremental re-export moved to `export_after_publish` in `pipeline.c` since you branched, so the current hook point no longer exists in the same shape.
2. **The drift counter is deletion-blind.** Deletions do not currently count toward the change threshold, so a change set consisting mostly of removed files can sit under the debounce longer than intended. At minimum worth documenting; better would be counting deletions into a drift-equivalent.

**Please do not rebase yet.** Three PRs' worth of conflict resolution against a heavily-reworked main is a lot of work to spend before the underlying design questions are answered — and #868's trust marker decides whether any of the stack lands. There is also an offer on #868 to distill the approved parts ourselves with `Co-Authored-By` credit if you would rather not carry the rebase.

One thing worth saying plainly: the reason this sat is our review latency, not any deficiency in the work. You asked the right question on #867 before building, and we owed you an answer four weeks ago.

DeusData:
Policy decision on the debounce: **accepted in principle, with tighter defaults and one gap closed.**

You're right about the underlying waste — re-dumping the entire database on every run to produce an artifact nobody asked for is real cost, paid repeatedly, for no benefit in the common case. Bounding that by change count, cumulative drift and age is the right shape.

Two changes I'd like before it lands:

**1. `CBM_ARTIFACT_EXPORT_MAX_AGE_S` should default to about an hour, not 24.** The artifact's whole purpose is that someone *else* picks it up — a teammate cloning, or CI seeding from it. A 24-hour ceiling means a colleague can pull an artifact that predates a day's work with nothing in the output telling them so. An hour keeps the "don't dump on every run" win almost entirely (the expensive case is many runs in quick succession) while making the staleness window short enough that nobody reasons from a stale graph for long. If you measured something that makes an hour materially worse than 24, I'd rather see those numbers than my guess — say so and I'll take the data.

**2. Deletions must count toward the drift threshold.** As written the drift accounting is deletion-blind, so a repo where files are only *removed* can sit below the change threshold indefinitely and keep exporting an artifact that still contains them. That's the worst kind of staleness, because the artifact isn't merely old — it asserts the existence of things that are gone. This needs a test too; the existing drift test is a good template.

One thing I'd also like, cheaply: when an export is **skipped**, the reason should be visible somewhere a human will look. You already log it; my concern is someone wondering why their artifact is older than their last index and having no obvious answer. A line in the run summary naming the threshold that held it back would save that confusion.

The env knobs themselves are fine as permanent surface — they're operator dials with sane defaults, not a new API. Please document both in the README with their defaults and what they trade off.

Your four tests look binding, and the cumulative-drift trigger you added mid-review in response to triage feedback was the right instinct — it closes the case where many small runs never individually cross the threshold.

Merge order remains #868 → #869 → #870; #868's design question is now settled, so the stack is moving again.

DeusData:
@moofone — the design here is accepted and is **not** being dropped; I've re-filed it as #1884 so it stops being blocked on a branch that main has moved out from under.

What changed: export left `dump_and_persist` and now lives in `export_after_publish` (`src/pipeline/pipeline.c:2529-2544`), whose signature has no access to `changed_total`, `file_count` or `stamps[]` — the three things the debounce needs, all local to `cbm_pipeline_run_incremental`. That's new cross-module plumbing, i.e. a design task rather than a rebase, which isn't fair to hand you as a "please rebase" after a seven-week wait that was mine.

#1884 carries your accepted shape, including the cumulative-drift trigger you added mid-review, plus the changes I'd asked for (MAX_AGE ~1h, deletions counting toward drift with a test, a run-summary line naming the blocking threshold, README docs). You're credited there.

One defect worth flagging wherever this lands: `parse_iso8601_utc` (`artifact.c:244`) range-checks month through second but never the **year**, and the result feeds `(exported_epoch + 1) * CBM_NS_PER_SEC` at `pipeline_incremental.c:800`. Since `INT64_MAX / 1e9` ≈ year 2262, an `indexed_at` of `9999-01-01T00:00:00Z` overflows signed int64 — reachable on the first index after a clone, and the UBSan diag lane traps on it.

Closing this in favour of #1884. Thank you for the work and for the patience — the review latency here was entirely on my side.

PR #879, opened 2026-07-05
feat(pipeline): index-time importance scoring (weighted-degree)

petercoxphoto:
## Index-time importance scoring (weighted-degree, Aider-style)

Split out of #776 (now closed) as an atomic feature PR, per @DeusData's request to break that bundle into single-concern PRs. The RSS/backpressure stability fix from that bundle already landed as 610531ad; the memory-safety fixes are superseded by upstream's own leak/slab work, so this PR carries only the **importance-ranking feature**. The `repo_map` tool is a separate stacked PR that builds on this.

### What it does

Adds a 7th predump pass (`pass_importance.c`, registered after all edge-producing passes) that persists a numeric `importance` key per Function/Method/Class node on `properties_json` (append pattern from `pass_complexity.c`). It does **not** touch `compute_search_score` — it's index-time only.

```
importance = sqrt(num_refs) x priv x generic x distinct x test_penalty
  num_refs      = incoming CALLS + USAGE edges
  priv     x0.1   leading-underscore
  generic  x0.1   name defined in >=5 distinct files
  distinct x10    snake_case|camelCase and len>=8
  test_penalty x0.1 when the graph's own canonical cbm_is_test_path()
                    flags the file, OR the symbol is the TARGET of an
                    incoming TESTS edge (prod helpers in non-test files)
```

Weighted-degree only; PageRank deferred (measured decision).

### Test-penalty note

The original spec used an edge-only TESTS/TESTS_FILE penalty. Real-corpus verification showed that under-fires (TESTS_FILE had ~15 edges; `is_test` stamped on <2% of nodes), leaving 33/40 top-ranked symbols as test scaffolding. Switched to the graph's canonical `cbm_is_test_path()` classifier (the one `pass_tests` uses) OR an incoming TESTS edge — **not** a naive `strstr` (it leaves `testutil_helpers`-style prod helpers unpenalised). After re-index: top-40 = 0 test-file symbols (was 33).

### Tests

`tests/test_importance.c` — 10 cases (registration/dispatch, the formula factors, and the test-penalty inversion). Rebased onto current `main`; the only merge work was test-harness wiring (`RUN_SELECTED_SUITE`).

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -211,6 +211,7 @@ PIPELINE_SRCS = \
     src/pipeline/pass_similarity.c \
     src/pipeline/pass_semantic_edges.c \
     src/pipeline/pass_complexity.c \
+    src/pipeline/pass_importance.c \
     src/pipeline/pass_cross_repo.c \
     src/pipeline/artifact.c \
     src/pipeline/pass_pkgmap.c
@@ -345,7 +346,7 @@ TEST_DISCOVER_SRCS = \
 
 TEST_GRAPH_BUFFER_SRCS = tests/test_graph_buffer.c
 
-TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_index_resilience.c
+TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_index_resilience.c tests/test_importance.c
 
 TEST_WATCHER_SRCS = tests/test_watcher.c
 
--- a/src/pipeline/pass_importance.c
+++ b/src/pipeline/pass_importance.c
@@ -14,12 +14,28 @@
  *     priv     0.1  if the name is private (leading underscore)
  *     generic  0.1  if the name is DEFINED in >= 5 distinct files
  *     distinct 10   if the name is snake_case or camelCase AND len >= 8
- *     test_penalty 0.1 if the symbol is test scaffolding, detected via the
- *                  TESTS / TESTS_FILE EDGES (never a filename substring):
- *                  the symbol is the TARGET of an incoming TESTS edge
- *                  (create_tests_edges: test-fn -> prod symbol), OR it lives
- *                  in a file that is the SOURCE of a TESTS_FILE edge
- *                  (create_tests_file_edges: test-file -> prod-file).
+ *     test_penalty 0.1 if the symbol is test scaffolding:
+ *                  - the symbol is the TARGET of an incoming TESTS edge
+ *                    (create_tests_edges: test-fn -> prod helper living in a
+ *                    NON-test file, e.g. a fixture in testutil.go), OR
+ *                  - the symbol lives in a test file per the graph's canonical
+ *                    cbm_is_test_path() classifier (the SAME one pass_tests
+ *                    uses).
+ *
+ * REAL-DATA DEVIATION FROM THE SPEC'S "edge-based TESTS/TESTS_FILE" mechanism
+ * (surfaced to Peter, documented in builder-notes.md): on the real connectors
+ * graph the TESTS/TESTS_FILE edges alone do NOT achieve AC1's exclusion goal.
+ * TESTS_FILE is emitted only when a test file maps to a resolvable production
+ * file (15 edges total on connectors); TESTS edges never point at a symbol that
+ * lives in a test file (create_tests_edges excludes tgt_is_test); and the node
+ * is_test property is stamped on <2% of nodes and on NONE of the test-resident
+ * fixtures/classes. So edge-only left the entire top-40 as test scaffolding
+ * (session_dir, FakeConnectorClient, ...). cbm_is_test_path() is the graph's
+ * own precise classifier (test_ prefix, _test.<ext> suffix, /tests/ dir, ...) —
+ * it is NOT the naive strstr("test") the spec warned against: it correctly
+ * leaves e.g. src/testutil_helpers.go UNpenalised (test-plan row #6 inversion).
+ * The TESTS-edge leg is retained so prod helpers in non-test files that are
+ * exercised by tests are still demoted.
  *
  * Weighted-degree only. PageRank (transitive importance) is a measured
  * refinement deliberately NOT built here — the spike showed the weighted
@@ -119,44 +135,6 @@ static int name_distinct_file_count(const cbm_gbuf_t *gb, const char *name) {
     return distinct;
 }
 
-/* Set of file paths that are the SOURCE of a TESTS_FILE edge (= test files). */
-typedef struct {
-    const char **paths;
-    int count;
-} cbm_test_file_set_t;
-
-static void test_file_set_build(const cbm_gbuf_t *gb, cbm_test_file_set_t *set) {
-    set->paths = NULL;
-    set->count = 0;
-    const cbm_gbuf_edge_t **edges = NULL;
-    int ne = 0;
-    if (cbm_gbuf_find_edges_by_type(gb, "TESTS_FILE", &edges, &ne) != 0 || ne == 0) {
-        return;
-    }
-    set->paths = malloc((size_t)ne * sizeof(*set->paths));
-    if (!set->paths) {
-        return;
-    }
-    for (int i = 0; i < ne; i++) {
-        const cbm_gbuf_node_t *src = cbm_gbuf_find_by_id(gb, edges[i]->source_id);
-        if (src && src->file_path) {
-            set->paths[set->count++] = src->file_path;
-        }
-    }
-}
-
-static bool test_file_set_contains(const cbm_test_file_set_t *set, const char *path) {
-    if (!path) {
-        return false;
-    }
-    for (int i = 0; i < set->count; i++) {
-        if (set->paths[i] && strcmp(set->paths[i], path) == 0) {
-            return true;
-        }
-    }
-    return false;
-}
-
 static int incoming_edge_count(const cbm_gbuf_t *gb, int64_t id, const char *type) {
     const cbm_gbuf_edge_t **edges = NULL;
     int ne = 0;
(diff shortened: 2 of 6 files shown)


DeusData:
Thanks for splitting this out of #776. Triage: normal-priority feature work for ranking/index quality.

Review will focus on score semantics, index-time overhead, properties_json mutation safety, and whether the tests prove stable ordering without making the ranking formula too rigid. This is the right size for a standalone review.

DeusData:
Reviewed this and #880 together, since #880 is stacked on it — verified at commit level: your four commits here appear verbatim in #880's head, so merge order is forced and #880 cannot land without this.

**The engineering discipline here is genuinely excellent, and I want to say so before the problems.** Red-first test boundaries with the failing run captured, positive discriminators so a dispatch-level error cannot fake a pass, exact-value asserts on every multiplier and boundary, and a round-trip persistence plus legacy-migration test. That is a model for how we would like feature PRs to arrive.

**The test-penalty work is the highlight.** You had a spec'd mechanism, real-corpus data showed it under-fired (33 of 40 top-ranked symbols were test scaffolding), and instead of shipping the spec you switched to the graph's own canonical classifier, proved the inversion case so `testutil_helpers` stays unpenalized, documented the deviation openly, and re-verified at 0/40. Measuring, finding your own design wanting, and saying so is exactly the standard this project wants.

**Two things block it as written.**

**1. A quadratic-pattern blocker at our scale.** `name_distinct_file_count` (`pass_importance.c:135-163`) does an O(k²) pairwise-strcmp dedup over a same-name group — and it is called *per node* from `:223`, so the group costs **O(k³) with no memoisation**. The base loop is fine (O(N), hash-indexed, single pass, no iteration), but names like `get`, `init`, `main`, `__init__` and `toString` plausibly reach k in the thousands on an 8.5M-node corpus. At k=3,000 that is ~2.7×10¹⁰ string compares for **one name** — minutes — and a handful of such names would add tens of minutes to a ~260-second index.

This is the "lookup-in-registration-loop" pattern we have a standing checklist for. The fix is straightforward: one O(N) prepass building a name→distinct-file-count hash map with hash-set dedup, then O(1) per node.

**2. It never runs on an incremental reindex.** The pass is registered only in the full pipeline; `pipeline_incremental.c`'s `run_postpasses` does not call it and the PR does not touch that file. So after any incremental run: symbols in changed files are purged and re-created **with no importance key at all**, while unchanged nodes keep **stale** scores even when their in-degree moved.

That compounds badly with #880 — see my note there, but the short version is that freshly-edited symbols score 0.0, and the seeded boost is multiplicative, so 0 × 50 = 0. **The files you are actively working on are exactly the ones that would silently drop out of the map.** That is the worst possible set to degrade on, and no test catches it because the migration test simulates via manual upserts rather than driving the real incremental path.

**On cost, for the record** (this part is fine): roughly 22–25 bytes per scored node, 3.4–5.1M scored nodes, so ~75–130 MB added to properties — real but modest against tens of GB. **No schema change**: it is a JSON key inside the existing properties column, legacy indexes keep working, and the forcing function to re-index is #880's explicit `unscored` error, which is *advertised* rather than silent. A release note would still be wanted.

**And the larger question, which is the maintainer's.** This introduces a *persisted importance-scoring concept* into the graph — your own description calls it an Aider-worldview idea. Whether cbm wants ranking persistence at all is a direction call, and rejecting the concept automatically rejects #880 as designed. I have put it upstream rather than deciding it in review.

There is also a consumer question worth an explicit answer: a full reindex recomputes scores, so `properties_json` churns on nodes in **untouched files** whenever call-graph degrees shift. That needs a decision about `detect_changes` semantics for property-only changes.

**Please hold off reworking until the direction answer arrives.** Both blockers are fixable and neither is a design flaw in your approach — but I would rather you fixed them once, against a decision.

DeusData:
Direction call: **yes, importance scoring belongs in the graph.** Thank you for waiting on it.

Ranking is a graph-quality problem, not a presentation one — "which of these 400 functions matters" is exactly the sort of question the graph should be able to answer rather than leaving every consumer to guess from name length. Weighted in-degree with distinctiveness and test-path discounts is a sound first cut, and storing it as a property rather than a schema column keeps it cheap to change when the formula inevitably gets refined.

Two things have to be fixed before it can land, and both are about scale rather than design:

**1. `name_distinct_file_count` is called per node inside the scoring loop with no memoisation.** On a repo with many same-named symbols — `get`, `init`, `run`, `handle` — that's a pairwise comparison per occurrence, which goes super-linear exactly where it hurts most. The Linux kernel has names with thousands of definitions; this would dominate the pass. The fix is a prepass: build a name → distinct-file-count map once, then look it up. That turns the whole thing into a single linear sweep.

Worth knowing that this project has a dedicated guard for precisely this class of bug — `build/c/test-runner complexity` runs the full pipeline at k and 2k replicated modules and asserts counter *ratios* rather than wall time. If your pass is linear the ratios stay ~2; a files×corpus coupling shows up as ~4 and fails. Please run it, and consider it the acceptance bar rather than my opinion.

**2. The pass never runs on the incremental path.** `pipeline_incremental.c` isn't touched, so importance is computed on full reindex only. The result is a graph where freshly-edited symbols carry no score at all while untouched ones carry scores computed against a graph that has since changed. Silent staleness in a ranking signal is worse than no ranking, because nothing looks wrong. It needs wiring into the incremental path, and a test that edits a file and asserts the affected nodes get re-scored.

There's a related question I'd like your view on rather than deciding by fiat: writing `importance` into `properties_json` on every full reindex means the property churns even when nothing structural changed, which affects what `detect_changes` reports. If the score is stable that's harmless; if it wobbles on every run it turns into noise. Have you looked at whether the values are stable across two identical reindexes?

Your ten tests are good — they check each multiplier in isolation and the ordering they produce together, plus the migration case where old indexes have no score. That's the shape that makes a formula safe to change later.

Your stacked #880 is a separate answer; details on that thread.

DeusData:
Following up on the green light from the 20th — and I owe you one correction to what I said there, which is better to have before you start reworking than after.

**The complexity suite cannot actually catch this pass.** I named it as the acceptance bar for the O(k³) fix, and that was wrong as the suite currently stands. `tests/test_complexity.c` gates deterministic *work counters* — nodes, edges, per-file registry defs — and asserts their ratios at k versus 2k. This pass emits no nodes and no edges and never touches `perfile_defs`, so its cost is pure `strcmp` work that no gated ratio observes. Wall time is deliberately non-gating, so a regression here would go green.

For the bar to bind, the rework needs to **expose a work counter for the name scan** and have the complexity suite assert on it. There is precedent to copy: the tail-match scan counters in `pass_parallel.c`, which `test_complexity.c` already references around line 64. With that in place, the O(N) prepass is provably linear rather than just believed to be.

So the rework list, consolidated:

1. **O(N) prepass for `name_distinct_file_count`.** As written it does O(k²) pairwise `strcmp` and is called per node, so it is O(k³) per name group — and on kernel-scale corpora `init`/`get`/`main` reach k in the thousands. One pass building a name → distinct-file-count map fixes it.
2. **A gated work counter for that scan** (the correction above), so the suite can prove it.
3. **The incremental path.** `pipeline_incremental.c` has no importance references, so after an incremental reindex fresh nodes are unscored while untouched nodes keep stale scores. With #880's multiplicative boost that means actively-edited files quietly drop out of the map — the worst direction for this to fail in. It needs wiring plus a test that drives the real incremental path rather than simulating it with manual upserts.
4. **Score stability across two identical reindexes** — still an open question from my side. If scores churn, `detect_changes` sees property deltas on untouched nodes.
5. Small one: test row #6's title and comments still describe the superseded edge-only `TESTS_FILE` design. The assert passes through `cbm_is_test_path` instead, so the inserted edge is inert — worth correcting so it does not mislead whoever reads it next.
6. The branch needs a rebase; the `passes[]` table and the test harness have both moved since July.

None of this is a complaint about the work. The red-first discipline here is genuinely good — the boundary pairs, the distinct-files-versus-nodes case, and an ordering test whose expected value distinguishes both misordering failure modes are all the right shape, and they are why this was reviewable at all.

On timing, so you are not guessing: I would like this to land in the **next two weeks or so**. If it goes quiet past that, we will implement the approved design ourselves with you credited as co-author and close this pointing at the commit — not to take it off you, but so the work ships rather than expires. I would much rather you carried it, and I am happy to answer anything on the counter shape or the incremental seam as you go.

Thanks for splitting this out of #776 when I asked, and for waiting through a long silence on my end to get the direction answer.

DeusData:
Landed. **Index-time importance scoring is on main as of `1130f577`** (PR #1883), with `Co-authored-by:` credit to you.

@petercoxphoto — thank you. The design is yours and it was accepted on its merits; what follows is what changed on the way in, because you should see it rather than diff it out of the tree.

**The blocker was real and worse than the estimate.** `name_distinct_file_count` was cubic in the size of a same-name group. The kernel hides that — its worst C name is `main` at k≈1064 — but Java does not. Measured on elasticsearch, one `-O2` binary, cold index, identical node counts across legs:

| leg | time |
|---|---|
| pass disabled | 85 s |
| **linear (shipped)** | **80 s** |
| cubic (original) | **483 s** |

**≈6.0x**, against a 3.7x estimate taken from nine names alone. Fixed by memoising the distinct-file count per distinct name and deduplicating paths through a hash set — Σk³ per group becomes Σk.

**Three traps closed, each proved by revert rather than by argument:**
- `PREDUMP_PASS_COUNT` is now derived structurally (`sizeof(passes)/sizeof(passes[0])`). Main had already grown a 7th pass, so a hand-written `= 7` beside an 8-entry table would have left importance — registered last — **silently never running, with every test green**.
- The property append is idempotent. Rehydrated nodes already carry `"importance"`, so an append-only write produced `...,"importance":1.0,"importance":1.0}`. The key is located in *key position*, not by `strstr`, which would also match text inside a string value.
- A **gated** complexity counter, not a recorded one — the existing gates structurally cannot observe this pass, so a green from them proved nothing about it.

**One bug found after the distill, in my own wiring rather than yours.** On the preferred closure-repair warm path the graph buffer holds only the delta, so importance was computed over a partial graph and those near-zero scores were persisted over correct ones. The equivalence test makes it concrete: editing `c0.py`'s *body* adds a call to `helper` in *unchanged* `hot.py` — full=√13, incremental=√12. Fixed by recomputing over the staging store after the delta merges, with the scoring rule kept as a single C function reached from SQL via a UDF so the two paths cannot drift.

That correctness costs **+17.0 s** on a kernel warm reindex (48.5 → 65.5 s, measured, 8.5M nodes), accepted deliberately and documented. Follow-up filed as #1889: the cost is a 1.13M-row UPDATE rather than the scoring itself.

**Stated plainly: nothing reads this score yet.** #880 is the consumer. This lands the production of the signal, not its use — so the cost arrives before the benefit, which is worth knowing.

Two things from your PR I deliberately left alone and filed instead: `cbm_is_test_path()` still lacks a `_test.c`/`_test.h` case, so the test-penalty under-fires on C, and `append_complexity_props()` carries the same append-only shape and would inherit the duplicate-key bug if ever wired into the incremental post-passes.

Thank you for the design and for the patience.

PR #880, opened 2026-07-05
feat(mcp): token-budgeted, seed-aware repo_map query tool

petercoxphoto:
## Token-budgeted, seed-aware `repo_map` MCP tool

Split out of #776 (now closed) as an atomic feature PR, per @DeusData's request. **Stacked on the importance-ranking PR** (#879) — `repo_map` ranks by the persisted `importance` key that PR adds, so please review/merge that first. The diff here shown against `main` includes the importance commits; the repo_map-only delta is the last two commits.

### What it does

Adds a `repo_map` MCP tool: a token-budgeted repository map of the most important symbols (persisted importance score), rendered one per line as `file: signature` — no bodies.

- **Two modes.** Pass `seed_anchors` (symbol names, qualified names, or file paths) to boost the seed neighbourhood (1-hop CALLS/USAGE, widened for weakly-connected seeds) to the top; omit / pass unresolvable seeds for the global importance-ranked map.
- **Output always fits `token_budget`** (default 1600, chars/4 estimate; must be positive).
- **Gated** on an importance-scored index — returns an explicit `unscored` error if the project needs re-indexing, rather than emitting a meaningless map.

### Adapted to current `main`

Upstream added a `title` field to the MCP tool struct (`{name, title, description, schema}`) since v0.8.1; the tool registration here was updated to `{"repo_map", "Repo map", ...}` to match.

### Tests

`repo_map_*` cases in `tests/test_mcp.c` (26 cases): registration/dispatch, token-budget fit (default/tight/zero/oversized), seed resolution + boost, gate errors, cross-project isolation, determinism, and leak-surface stability.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -211,6 +211,7 @@ PIPELINE_SRCS = \
     src/pipeline/pass_similarity.c \
     src/pipeline/pass_semantic_edges.c \
     src/pipeline/pass_complexity.c \
+    src/pipeline/pass_importance.c \
     src/pipeline/pass_cross_repo.c \
     src/pipeline/artifact.c \
     src/pipeline/pass_pkgmap.c
@@ -345,7 +346,7 @@ TEST_DISCOVER_SRCS = \
 
 TEST_GRAPH_BUFFER_SRCS = tests/test_graph_buffer.c
 
-TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_index_resilience.c
+TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_index_resilience.c tests/test_importance.c
 
 TEST_WATCHER_SRCS = tests/test_watcher.c
 
--- a/src/main.c
+++ b/src/main.c
@@ -507,9 +507,9 @@ static void print_help(void) {
     printf("  Claude Code, Codex CLI, Gemini CLI, Zed, OpenCode,\n");
     printf("  Antigravity, Aider, KiloCode, Kiro\n");
     printf("\nTools: index_repository, search_graph, query_graph, trace_path,\n");
-    printf("  get_code_snippet, get_graph_schema, get_architecture, search_code,\n");
-    printf("  list_projects, delete_project, index_status, detect_changes,\n");
-    printf("  manage_adr, ingest_traces\n");
+    printf("  get_code_snippet, get_graph_schema, get_architecture, repo_map,\n");
+    printf("  search_code, list_projects, delete_project, index_status,\n");
+    printf("  detect_changes, manage_adr, ingest_traces\n");
 }
 
 /* ── Main ───────────────────────────────────────────────────────── */
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -5959,8 +5959,8 @@ static char *handle_repo_map(cbm_mcp_server_t *srv, const char *args) {
         if (cbm_store_find_node_by_id(store, onehop[i], &nb) != CBM_STORE_OK) {
             continue;
         }
-        bool is_module = nb.label && (strcmp(nb.label, "Module") == 0 ||
-                                      strcmp(nb.label, "File") == 0);
+        bool is_module =
+            nb.label && (strcmp(nb.label, "Module") == 0 || strcmp(nb.label, "File") == 0);
         if (is_module && nb.file_path && nb.project && strcmp(nb.project, project) == 0) {
             cbm_node_t *fnodes = NULL;
             int fcount = 0;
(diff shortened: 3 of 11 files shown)


DeusData:
Thanks for splitting the repo_map tool into its own PR. Triage: normal-priority editor/MCP integration feature, stacked on #879.

Because this adds a new MCP tool surface, review will be intentionally cautious: token-budget enforcement, project-boundary handling, seed resolution, no unexpected file bodies beyond the stated output shape, and local-only behavior all need to be verified. Please keep this stacked behind #879 for review/merge order.

DeusData:
Reviewed with #879 — this PR is **stacked on it** (verified at commit level: #879's four OIDs appear verbatim in this head), so merge order is forced and this cannot land without it. It also depends on it at runtime: `repo_map` errors with `unscored` on any index `pass_importance` did not write.

**The build quality is high**, and I said so at length on #879 — red-first boundaries, positive discriminators, a determinism contract with a forced-tie byte-identity test, and a dedicated null-guard commit. The care is visible.

**The finding I most want you to see, because it is subtle and it defeats the tool's purpose.**

The seeded boost is **multiplicative**, applied against #879's importance score. And #879 does not run on incremental reindexes — so a symbol in a file you just edited comes back with importance **0.0**.

0 × 50 = 0.

Which means **an explicitly requested seed can be silently absent from its own map**. The files you are actively working on are precisely the ones that would vanish, and nothing in the output says anything is missing. For a tool whose job is orienting an agent in a codebase, silently omitting the working set is close to the worst available failure mode. Fixing #879's incremental gap closes it, which is another reason the two are entangled.

**Two more things block it as written:**

1. **Per-call full-table scans.** `store.c:961-1000` and `:932-959` scan all symbol rows with per-row `json_extract` and an unindexed `ORDER BY` — seconds per call on 8.5M nodes, on every invocation. It needs an indexed representation of the score rather than a scan.
2. **Truncation is inferable rather than advertised.** There is no `truncated` flag, no cursor, no pagination — and the output is a flat-string dialect rather than the tree-format-plus-cursor conventions main has since moved to. That matters more here than usual: a "token-budgeted" tool *by definition* drops content to fit, so the caller has to be told when that happened. An agent receiving a partial map that looks complete will reason confidently from it.

That last point is also the direction question I have put to the maintainer. `repo_map` would be a **permanent new MCP tool** — a one-way door — whose organising principle is a token budget, in a project whose stated priority is findability over token thrift. It also partially overlaps `get_architecture`'s hotspots. That is a product call, not a reviewer's.

**Two smaller things:**

- The `src/main.c` help hunk targets code that `0e1b453` removed, and the "15 graph tools" comment is stale.
- The `clang-format-20` commit is worth undoing on rebase: our formatter is the **Homebrew LLVM** build, and a standalone `clang-format-20` produces false whole-file drift here. It reflowed unrelated tests in `test_mcp.c` — that is the 36 deletions — which is scope noise and guaranteed rebase friction.

**Please hold off reworking until the direction answer on #879 arrives.** Both PRs turn on the same decision, and between the two blockers here and the two there, that is a lot of work to spend before the concept is settled.

DeusData:
Thank you for this, and I'm sorry the answer is no — but I owe you the actual reason rather than a vague "not right now", because you built it carefully and the reason is specific.

**`repo_map` is declined. Its stacked sibling #879 is accepted** (with two scale fixes; details on that thread), so the importance scoring you wrote is going to land.

The problem is the premise. `repo_map` is organised around a token budget: given N tokens, return the most valuable subset. That is a reasonable design for a lot of tools — but this project has deliberately gone the other way. Findability comes first and token thrift second, on the reasoning that an agent which cannot find the right symbol will burn far more tokens flailing than it ever saved on a compact answer. A tool whose primary parameter is `token_budget` bakes the opposite priority into permanent surface, and once agents in the wild call it, that contract is ours forever.

There's an overlap problem underneath it too: `get_architecture` already answers "give me the shape of this repo". Two tools that answer nearly the same question with different ranking rules is a worse outcome than one good tool — agents pick semi-randomly between them, and we get to keep both correct indefinitely.

The concrete concerns, for completeness, and because they'd need solving even if the premise held:

- **Cost per call.** `SELECT ... ORDER BY json_extract(properties,'$.importance')` is an unindexed scan of every symbol row. At the 8.5M-node scale this project targets, that's seconds per call for a tool designed to be called casually.
- **Silent omission.** Because the seed boost multiplies the persisted score, and #879 currently doesn't run on the incremental path, a symbol you just edited can score 0.0 and vanish from a map you explicitly seeded with it — with no `truncated` flag and no cursor to reveal that anything was dropped. An answer that quietly omits the thing you asked about is the worst failure mode a discovery tool can have.

None of that is careless work. Twenty-eight tests covering budget edges, seed resolution and widening, the unscored gate, cross-project isolation and determinism is more rigour than most feature PRs arrive with, and the `unscored` error gate — refusing to answer rather than returning a misleading empty map — is exactly the right instinct.

If you want the underlying capability, the door I'd point at is different: make **existing** tools rank better using #879's scores, rather than adding a tool that ranks. `search_graph` returning the most important matches first is the same value with no new surface and no second answer to "what does this repo look like". If you'd like to take that on once #879 lands, I'd review it gladly.

Two small things from this PR worth carrying over regardless: the stale `src/main.c` help hunk and the stale tool-count comment are real and should be fixed by someone — and if you rebase anything from here, drop the `clang-format-20` commit. This repo pins clang-format to the Homebrew LLVM build; a standalone `clang-format-20` produces whole-file drift that isn't real, which is where those 36 reflowed lines came from.

Closing this one. #879 stays open and is the half I want.

PR #886, opened 2026-07-05
fix(win): harden artifact.c git shell-outs against unquoted/unvalidated repo paths

Flipper1994:
## Summary

Follow-up to #881 (split out per review). Hardens the **git shell-outs in `pipeline/artifact.c`** — a *different* bug class from #881's `CreateProcess` argv-quoting: this is `popen`/shell command-string quoting.

## The bug

`artifact.c` shells out to git twice via `cbm_popen`, interpolating the repo path with **single quotes** and **no validation**:

```c
git -C '%s' rev-parse HEAD 2>/dev/null                 // git_head_hash
git -C '%s' config merge.ours.driver true 2>/dev/null  // ensure_gitattributes
```

- **`cmd.exe` does not honor single quotes**, so on Windows a repo path containing a **space** breaks argument grouping — the artifact commit hash and merge-driver config silently fail for any spaced repo path.
- An embedded quote or shell metacharacter in `repo_path` could break out of the intended argument.

This was the only unvalidated git shell interpolation left in the tree (every other git shell-out — `git_context.c`, `pass_githistory.c`, `watcher.c`, `mcp.c` — already validates and double-quotes).

## Fix

- Validate `repo_path` with `cbm_artifact_repo_path_is_shell_safe()` and switch to **double quotes** + a per-platform null device, matching `git_context.c` (the best-hardened git shell-out).
- `cbm_validate_shell_arg` rejects quote / backslash / substitution metacharacters; on Windows we additionally reject the `cmd.exe` expansion metacharacters `% ! ^`.
- Double quotes are honored by **both** POSIX `sh` and `cmd.exe`, so a legitimate **spaced** repo path now works — the concrete regression the single-quote form caused.

## Test

New guard suite in `tests/test_artifact.c` for the exposed `cbm_artifact_repo_path_is_shell_safe()`:

- plain and **spaced** paths accepted (the regression),
- quote / `;` / `$()` / backtick / pipe injection rejected,
- `cmd.exe` metacharacters `% ! ^` rejected on Windows (allowed on POSIX, where they are literal inside double quotes).

## Scope

Atomic to the artifact.c shell-quoting hardening. No behavior change on valid paths except that spaced paths now work on Windows.

Refs #881.

### DIFF
--- a/src/pipeline/artifact.c
+++ b/src/pipeline/artifact.c
@@ -21,6 +21,7 @@ enum {
 #include "foundation/compat_fs.h"
 #include "foundation/compat.h"
 #include "foundation/log.h"
+#include "foundation/str_util.h" /* cbm_validate_shell_arg — git shell-out hardening */
 
 #include "zstd_store.h"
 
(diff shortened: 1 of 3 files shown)


DeusData:
Merged — thank you, @Flipper1994! 🙏 This is a clean, well-targeted hardening: gating both git shell-outs on `cbm_artifact_repo_path_is_shell_safe`, switching single→double quotes (the real cross-platform fix), the null-device redirect, and the snprintf truncation guards for parity with `git_context.c`. The unit tests across plain / spaced / Windows paths and the injection cases are exactly what we want to see on a shell-out path. Really appreciate the careful work here — and on the CreateProcess argv-quoting fix too.

PR #889, opened 2026-07-05
fix(subprocess): NUL-terminate cbm_build_win_cmdline buf on overflow (post-review, follow-up to #881)

Flipper1994:
## Summary

Small post-review tidy — follow-up to #881 (already merged). Pure polish: no functional change for any real input.

## The one substantive fix

`cbm_build_win_cmdline()` (`src/foundation/subprocess.c`) returns `false` from inside its loop on buffer overflow **without NUL-terminating `buf`**, although its header contract in `subprocess.h` says the buffer holds a string. There is **no live bug** — both call sites (`cbm_run_win` and the UI `http_server` index spawn) ignore `buf` when the function returns `false` — but a future caller trusting the documented contract could read an unterminated (possibly uninitialized) buffer. This sets `buf[0] = '\0'` on the overflow path so `buf` is always a valid string, and corrects the header comment to match.

## Also (same files, same review pass)

- Fix the stale `cbm_cmdline_put` comment — it returns `pos` **unchanged** on overflow; the overflow is signalled via `*ovf`, not via an advancing `pos`.
- Bounds-guard the reference `CommandLineToArgvW` parser in the round-trip test (`tests/test_subprocess.c`) so a hypothetical future over-long test argument fails safely instead of overflowing its 256-byte row.

These came out of an adversarial self-review of #881 after it had already merged; landing them per the boy-scout rule rather than leaving a contract inaccuracy in a public function.

Refs #881.

### DIFF
--- a/src/foundation/subprocess.c
+++ b/src/foundation/subprocess.c
@@ -135,8 +135,8 @@ static bool cbm_tail_log(const char *log_file, long *tail_pos, cbm_proc_log_cb c
 /* ── Windows command-line quoting (pure; unit-tested on every platform) ─────── */
 
 /* Append char `c` to buf[cap], reserving the final byte for a NUL terminator.
- * Sets *ovf on overflow and stops writing; pos keeps advancing so the caller
- * still detects the overflow after the loop. */
+ * On overflow: sets *ovf, stops writing, and returns pos UNCHANGED — callers detect
+ * the overflow via the *ovf flag (not via the return value). */
 static size_t cbm_cmdline_put(char *buf, size_t cap, size_t pos, char c, bool *ovf) {
     if (pos + 1 >= cap) {
         *ovf = true;
@@ -213,6 +213,7 @@ bool cbm_build_win_cmdline(char *buf, size_t cap, const char *const *argv) {
     for (int i = 0; argv[i]; i++) {
         pos = cbm_cmdline_append_arg(buf, cap, pos, argv[i], i == 0, &ovf);
         if (ovf) {
+            buf[0] = '\0'; /* overflow: leave buf a valid (empty) string, never unterminated */
             return false;
         }
     }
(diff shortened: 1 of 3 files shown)


DeusData:
This is great, @Flipper1994 — the fix is spot on: `buf[0] = '\0'` on the overflow path (and I see the real `'\0'`, not the `' '` from your note 👍), the header doc now matches, and the `cbm_cmdline_put` comment correction is accurate — it does `return pos` unchanged on overflow, so the old "pos keeps advancing" line was genuinely stale. The `PUTO` bounds-guard on the reference parser is a nice touch too.

One thing before we merge, to match our reproduce-first convention: this still needs a **test that is RED on current `main` and green with the fix**. The `PUTO` guard hardens the reference parser but doesn't exercise the actual contract (overflow → `buf` terminated). Something like:

```c
TEST(win_cmdline_overflow_leaves_empty_string) {
    char buf[8];
    const char *const argv[] = {"averylongprogramname", "x", NULL};
    ASSERT_FALSE(cbm_build_win_cmdline(buf, sizeof(buf), argv)); /* overflows cap */
    ASSERT_EQ(buf[0], '\0'); /* RED on unfixed: buf[0] holds a partial byte, not NUL */
    PASS();
}
```

On the pre-fix code `buf[0]` is the first quoted byte (not `'\0'`), so it fails; with your fix it passes — that's the guard that proves the contract and locks it in. Register it in the suite and this is good to go. Everything else looks clean.

Flipper1994:
Done — added `win_cmdline_overflow_leaves_empty_string` exactly as you sketched and registered it in the suite. Verified the reproduce-first property locally: on the pre-fix code it's **RED** (`buf[0] == 34`, i.e. the leading `"`, not `'\0'`), and it goes **green** with the `buf[0] = '\0'` fix. Kept both comment corrections in this PR as you asked. Pushed — thanks for the sharp review 🙏

DeusData:
Merged (`fea48a50`) — thank you! Clean overflow-contract fix with a genuine reproduce-first guard. 🙏

Issue #20, opened 2026-03-05
Issue with windows paths?


mattepiu:
From cmd:

```
C:\Users\m>C:\Users\m\bin\codebase-memory-mcp.exe cli index_repository '{"repo_path": "C:\Temp\codebase-memory-mcp"}'
error: invalid arguments: invalid character '\'' looking for beginning of value

C:\Users\m>C:\Users\m\bin\codebase-memory-mcp.exe cli index_repository '{"repo_path": "C:\\Temp\\codebase-memory-mcp"}'
error: invalid arguments: invalid character '\'' looking for beginning of value
```

from git bash always ends a scan with:
```
Indexed "C:-Temp-codebase-memory-mcp": 2912 nodes, 10333 edges
  indexed_at: 2026-03-05T14:04:06Z
  db: C:\Users\m\.cache\codebase-memory-mcp\C:-Temp-codebase-memory-mcp.db

```
then if I check ~/.cache/codebase-memory-mcp directory:
```
$ ls -lahtr ~/.cache/codebase-memory-mcp/
total 4.0K
drwxr-xr-x 1 m 4096 0 Mar  5 15:04 ../
drwxr-xr-x 1 m 4096 0 Mar  5 15:04 ./
-rw-r--r-- 1 m 4096 0 Mar  5 15:04 C

> /c/Temp/codebase-memory-mcp (main)
$ cat ~/.cache/codebase-memory-mcp/C  

[NO OUTPUT AS FILE IS EMPTY]

```

xberry1231:
This'll give you C--Users-project. Could use regexp [/:]+ or change the replace to avoid the double --

DeusData:
This should be addressed in v0.4.1 — `ProjectNameFromPath()` now sanitizes colons and backslashes from Windows paths, and collapses consecutive dashes. For example `C:\Users\project` now correctly produces `C-Users-project` instead of `C:-Users-project`.

Could you update and test again? Let me know if you still encounter any issues.

DeusData:
Correction: the fix with the unified version management is in **v0.4.4** (v0.4.3 was built before the final commit). Please update to v0.4.4.

PR #891, opened 2026-07-05
fix(win): deliver non-ASCII CLI paths end-to-end (wide argv + CreateProcessW)

### LINKED ISSUE #423: codebase-memory-mcp cli index_repository asking for path but path is specified (Windows)

Flipper1994:
## Summary

Closes #423, #20 -- the still-open Windows narrow-argv bug flagged in the #639 review as "genuinely still present". A non-ASCII repo path passed to `codebase-memory-mcp cli index_repository "<json>"` was mangled on Windows and the command failed with "repo_path is required" instead of indexing the real directory. POSIX is unaffected (argv is already UTF-8).

## The bug has TWO boundaries (both fixed)

A self-review caught that fixing only argv is insufficient — the default runtime re-mangles the path a second time:

1. **argv read.** `main()` took only the narrow `int main(int argc, char **argv)`, so the CRT handed it argv in the active ANSI code page. On Windows `main()` now rebuilds argv from the wide command line (`GetCommandLineW` + `CommandLineToArgvW`) and converts each element to UTF-8 (`cbm_wide_to_utf8`).
2. **worker spawn.** With the supervisor at its **default (enabled)** setting, the worker is spawned via `CreateProcessA`, which re-interprets the UTF-8 command line through the ANSI code page — re-mangling the path at the parent→worker boundary even after (1). Both spawn sites (`subprocess.c` `cbm_run_win` and the UI `http_server.c` index spawn) now build a wide command line (`cbm_utf8_to_wide`) and use `CreateProcessW`.

Without (2) the default path stays broken while only `CBM_INDEX_SUPERVISOR=0` (in-process) works — so this is the complete end-to-end fix.

## Test — reproduce-first, promoted guard, exercises the DEFAULT path

`tests/windows/test_cli_non_ascii_arg.py` (the opt-in **known-red** from #639) is promoted to a **green guard** (`$knownReds` → `$guards`, gated by `test-windows-guards`). Critically it now **drops** the suite's `CBM_INDEX_SUPERVISOR=0`, so it crosses the real supervisor→worker spawn boundary — otherwise a green guard would mask the spawn half of the bug.

Verified on native Windows (Win11, CLANG64):
- **RED** before the fix; **GREEN** after (a `café_日本語_repo` indexes, `nodes>0`) on the **supervised** path.
- Full `-GuardsOnly` suite green.

Also: `-lshell32` added explicitly (`main.c` now depends on `CommandLineToArgvW`), and `RED_TEST_ANALYSIS.md` refreshed from known-red to green-guard.

Closes #423, #20.


DeusData:
One housekeeping ask before we merge, @Flipper1994: could you drop `tests/windows/RED_TEST_ANALYSIS.md` from this PR (`git rm` it — deleting it from the repo is fine)? We're keeping analysis/tracking markdown out of the tree as a policy. The code fix and the promoted `test_cli_non_ascii_arg.py` guard are exactly what we want — just the `.md` needs to go.

The fix itself looks great: clean wide-argv `main()` (GetCommandLineW → CommandLineToArgvW → UTF-8, with fallback) + `CreateProcessW` at both spawn sites, and `-lshell32` wired in. Once the `.md` is removed I'll re-diff and merge. Thanks! 🙏

Flipper1994:
Done -- `git rm`'d `tests/windows/RED_TEST_ANALYSIS.md` and updated the single reference to it in `test-windows.ps1` so nothing dangles. Just the code fix + the promoted `test_cli_non_ascii_arg.py` guard remain now. Pushed -- ready for your re-diff. Thanks! 🙏

DeusData:
Merged (`a5351d3a`) — thank you! Solid end-to-end non-ASCII CLI fix (wide argv via GetCommandLineW→CommandLineToArgvW + CreateProcessW at both spawn sites), and thanks for dropping the `.md` and the clean commit message. 🙏

Issue #802, opened 2026-07-03
discover: unify the exclusion predicate across all filesystem walkers


DeusData:
Umbrella follow-up from the #517/#543/#793 review. Today each walker owns a private skip predicate: main discovery honors built-ins + .gitignore stack + .cbmignore; the pkgmap walk honors only hardcoded built-ins; the path-alias walk has its own hardcoded list; envscan's walker is currently uncalled. Result: contradictory semantics (a dir un-skipped by negation would still be vetoed in pkgmap but descended by path_alias).

Target design:
- ONE shared exclusion predicate: built-in skips + gitignore stack + .cbmignore incl. negation (last-match-wins via cbm_gitignore_match_result).
- Non-negatable safety core: .git, node_modules, worktree-internal dirs stay skipped regardless of negation (repo-committed .cbmignore must not defeat OOM/safety skips, see #489).
- Transport: #793's materialized excluded-roots list; per-walker hardcoded lists demoted to delegating fast paths so negation overrides propagate.
- Incremental-path plumbing: pipeline_incremental's ctx lacks excluded_dirs, so merge_pkg_entries walks unexcluded on incremental reindexes (residue of #792, see #793 review).

Related: #543 (negation, being re-implemented on cbm_gitignore_match_result), #793, #510 (parent-gitignore half still open), #489.

DeusData:
Thanks for consolidating this from the exclusion-related PR reviews. Triage: discovery/watcher correctness bug under stability/performance. v0.9.0 improved ignore handling, but this issue should stay open until all filesystem walkers delegate to one shared exclusion predicate with the non-negatable safety core preserved.

PR #912, opened 2026-07-06
Add CBM_EXTRA_SKIP_DIRS env var for site-specific always-skip directories

chethanuk:
## What does this PR do?

Adds a `CBM_EXTRA_SKIP_DIRS` environment variable so callers can skip extra,
site-specific directory names during indexing without a rebuild.

**Motivation.** Prior AI-tooling output directories (`.codegraph/`,
`graphify-out/`, `.wiki/`, `.understand-anything/`) get walked and parsed as
source during indexing, since none are in the universal skip list
(`ALWAYS_SKIP_DIRS` in `src/discover/discover.c`) or in target repos' own
`.gitignore`. Hardcoding these four names would fix only this case; the general
problem is broader (a `.dbt` target dir, a Terraform state dir, etc.) and a
rebuild per new name does not scale.

**What changed.**
- `src/discover/discover.c`: `cbm_should_skip_dir` now also skips names listed
  in `CBM_EXTRA_SKIP_DIRS` (comma-separated, whitespace-trimmed), checked
  alongside the hardcoded list and applied in every mode. The truly universal
  `ALWAYS_SKIP_DIRS` list is left untouched. The value is read through
  `cbm_safe_getenv` into a local buffer — matching the rest of the codebase — so
  it is thread-safe (raw `getenv`'s pointer can be invalidated by a concurrent
  `setenv`/`putenv`) and correct for UTF-8 environment values on Windows. The
  CSV scan is non-mutating, so the copied buffer is never altered in place.
- `tests/repro/repro_invariant_discovery_fqn.c`: adds table-driven
  `invariant_discovery_env_extra_skip_dirs` — a name is discovered by default,
  skipped once `CBM_EXTRA_SKIP_DIRS` lists it (including a whitespace-padded CSV
  entry to exercise trimming), and discovered again after the var is cleared,
  proving the behavior is genuinely env-gated. Uses the portable
  `cbm_setenv`/`cbm_unsetenv` wrappers so it also compiles on Windows.
- `docs/cbmignore.md`: documents the new opt-in. `CBM_EXTRA_SKIP_DIRS` is set by
  the *caller* invoking indexing, not by the indexed repo, so it does not weaken
  the "not overridable from a repo-local ignore file" guarantee.

## How verified

- `str_in_env_csv` / `cbm_should_skip_dir` in `src/discover/discover.c` compile
  clean (`cc -fsyntax-only`); the repro test translation unit compiles clean.
- Extracted the CSV parser into a standalone harness and asserted the
  baseline/skip/restore and whitespace-trimming cases (listed name skipped,
  padded entries trimmed, substrings not matched, unset/empty env not skipped).
- Full `make -f Makefile.cbm test` requires the project toolchain; the new
  behavior is covered by the table-driven repro test above.

## Checklist

### DIFF
--- a/docs/cbmignore.md
+++ b/docs/cbmignore.md
@@ -69,7 +69,12 @@ a path wins. For directories:
 1. **Built-in skip list** — `.git`, `node_modules`, `dist`, `target`,
    `vendor`, tool caches, etc. (60+ names; the fast/moderate index modes add
    more, e.g. `docs`, `examples`, `testdata`). Not overridable from any
-   ignore file today.
+   ignore file today. The hardcoded list is intentionally fixed across all
+   repos, but you can extend it per invocation with the `CBM_EXTRA_SKIP_DIRS`
+   environment variable — a comma-separated list of extra directory names to
+   skip in every mode (e.g. `CBM_EXTRA_SKIP_DIRS=".dbt,.terraform-state"`).
+   Useful for site- or repo-specific artifact directories that don't belong
+   in the universal list. Whitespace around each name is trimmed.
 2. **Repo `.gitignore`** — `<repo>/.gitignore` merged with
    `<git-common-dir>/info/exclude` (worktree-aware); later patterns win on
    conflict. Honored even when the indexed directory is not a git repo root.
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -334,6 +334,44 @@ static bool resolve_global_excludes_path(char *out, size_t out_sz) {
     return false;
 }
 
+/* ── Env-var-configurable extra skip dirs ──────────────────────────
+ * CBM_EXTRA_SKIP_DIRS: comma-separated directory names to skip, on top of
+ * the hardcoded ALWAYS_SKIP_DIRS above. Lets a caller add repo-specific or
+ * site-specific artifact dirs (e.g. ".dbt", a Terraform state dir, a prior
+ * AI-tooling output dir) per invocation, without a rebuild. Applies in
+ * every mode, same as ALWAYS_SKIP_DIRS. Not cached across calls (the list
+ * is short and this keeps the function safe to call from any thread). */
+static bool str_in_env_csv(const char *s, const char *csv) {
+    if (!s || !csv || !csv[0]) {
+        return false;
+    }
+    size_t slen = strlen(s);
+    const char *p = csv;
+    while (*p) {
+        const char *comma = strchr(p, ',');
+        size_t seg_len = comma ? (size_t)(comma - p) : strlen(p);
+
+        const char *seg = p;
+        size_t len = seg_len;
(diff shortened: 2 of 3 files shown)


DeusData:
Thanks for the site-specific skip-dir proposal. Triage: backlog performance/config enhancement.

Review should keep this conservative: documented local-only environment behavior, no weakening of built-in safety skips, predictable parsing of the comma list, and tests proving it applies consistently across discovery modes.

DeusData:
Thanks — a site-specific always-skip knob is a reasonable ask and the invariant test placement is right. Two blockers:

1. **Lint fails** — the `src/discover/discover.c` hunks fail clang-format (see the lint job); repo style is the checked-in `.clang-format` via Homebrew LLVM.

2. **Rebase + integrate with the new by-design coverage classifier** — `discover.c` was substantially rewritten on main this week (#974): per-file skips now flow through `file_skip_reason()` which returns a reason string ("gitignore", "skip-list", …) that gets persisted and reported as the "purposely not indexed" class (`index_status.not_indexed`, `not_indexed_files` in the index response). Your `CBM_EXTRA_SKIP_DIRS` check should slot into that framework: directories skipped by the env var should surface as excluded subtrees exactly like the built-in skip list does today (they already will if you hook `should_skip_directory`), and please add a note in `docs/cbmignore.md` about how they appear in the coverage report. While rebasing you'll hit conflicts in the skip machinery — the reason-classifier refactor moved the file-level checks.

One design question to answer in the PR: why an env var rather than `.cbmignore` (which already covers per-repo excludes and is committable)? The README/docs should state the intended use case (site-wide policy across many repos without touching each one, presumably) so users know which mechanism to reach for.

DeusData:
Thank you @chethanuk — this is cleanly implemented (safe CSV parsing via cbm_safe_getenv, correct whitespace trimming, a genuine reproduce-first test, and it can only ever skip *more* directories, never un-skip the safety core). After review we've decided not to add it, for a direction reason rather than any problem with the code:

- **The per-repo case is already covered** by `.cbmignore` (gitignore syntax, read from the repo root) — `.codegraph/`, `graphify-out/`, `.wiki/`, etc. go there today (see docs/cbmignore.md).
- **The per-user / per-site 'always skip these everywhere' case** is already covered by git global excludes (`~/.config/git/ignore` / `core.excludesFile`), which we consult in the exclusion stack.
- Adding `CBM_EXTRA_SKIP_DIRS` as a new permanent env var that only the main discovery walker honors (not the pkgmap / path-alias / envscan walkers) would reintroduce exactly the per-walker fragmentation that **#802** exists to remove — it's unifying ALL exclusion onto one shared predicate.

If there's a real gap for you that `.cbmignore` + global git excludes don't cover (e.g. a per-user skip for a *non-git* target), that's worth raising on #802 so it lands in the one unified predicate rather than as a standalone knob. Really appreciate the contribution and the clean test.

PR #927, opened 2026-07-07
fix(pipeline): classify native fetch() as HTTP_CALLS, not local shadows

### LINKED ISSUE #856: HTTP client detection misses native `fetch()` — only matches hardcoded library allowlist (axios/node-fetch/ofetch/etc)
## Summary

`cbm_service_pattern_match()` in `internal/cbm/service_patterns.c` classifies a
call as an HTTP client call (`HTTP_CALLS` edge) only when the *resolved*
qualified name contains a substring from a hardcoded allowlist
(`http_libraries[]`, service_patterns.c ~L30-70): `axios`, `node-fetch`,
`undici`, `ofetch`, `requests`, `httpx`, etc.

The native/global Fetch API (`fetch(url, opts)` used with **no import
statement**, as in any browser or modern Node runtime) is not in this list.
Since it has no import, its resolved QN never contains any library
substring, so `cbm_service_pattern_match()` returns `CBM_SVC_NONE` and no
`HTTP_CALLS` edge is ever emitted — even when the call's first argument is a
literal path that byte-matches a real backend `Route` node.

## Repro

Vue frontend (`frontend/src/api.js`):
```js
export async function drawLottery(drawId) {
  const res = await fetch('/lottery/draw', {
    method: 'POST',
    headers: { 'Content-Type': 'application/json', Authorization: `Bearer ${auth.token}` },
    body: JSON.stringify({ draw_id: drawId }),
  })
  ...
}
```

FastAPI backend (`backend/app/draw.py`):
```python
@router.post("/lottery/draw")
def draw(body: DrawRequest, ...):
    ...
```

- `trace_path(function_name="drawLottery", mode="cross_service")` → no link found.
- `MATCH (a)-[r]->(b:Route {qualified_name:'__route__POST__/lottery/draw'})` → 3 inbound edges, all backend-side (`CALLS` from the module, `CALLS` from a test, `HANDLES` from the handler itself). Zero edges from the frontend caller.
- Project-wide: 8 of 9 frontend API functions in `api.js` call bare `fetch()` directly and get **zero** `HTTP_CALLS` edges. `cross-repo-intelligence` / `trace_path(mode=cross_service)` can never link them to their backend routes.

## Secondary oddity (found while investigating, not fully root-caused)

Two functions (`requestOtp`, `verifyOtp`) that call a **local wrapper**
(`postJson(path, body)` → itself calls bare `fetch`) *do* get an `HTTP_CALLS`
edge, despite `postJson` not resolving to anything in `http_libraries[]`,
`route_reg_libraries[]`, or matching `method_suffixes[]`/`route_reg_suffixes[]`
(checked all three tables). Whatever mechanism creates these edges also emits
a **duplicate `Route` node** with method `ANY` (`__route__ANY__/auth/otp`)
alongside the real FastAPI-registered one (`__route__POST__/auth/otp`), and
the edge lands on the `ANY` node — which has no `HANDLES` edge from the real
handler. So even the "working" case points at an orphan node, not the real
route.

## Suggested fix

Add the bare Fetch API as a recognized "library" for HTTP classification —
e.g., treat an unresolved/global `fetch` identifier (no import binding) as an
implicit HTTP client call when its first argument is a URL/path-shaped string
literal, the same way `emit_http_async_edge` already gates on
`first_string_arg[0] == '/' || strstr(url, "://")`. This is likely a common
false-negative for any modern frontend codebase that doesn't wrap `fetch` in
axios/ofetch/etc.

## Environment

- codebase-memory-mcp 0.8.1 (linux-amd64-portable)
- Reproduced on a Vue 3 + FastAPI monorepo, `index_repository(mode=full)`

apappas1129:
## What does this PR do?

Closes #856: native `fetch()` calls (Node 18+'s built-in, and the identical browser/WHATWG API — no static way to tell them apart, and no need to: both make a real outbound HTTP request) were invisible to cross-repo intelligence. A bare `fetch(url)` has no import and typically no local definition, so registry resolution comes back empty and the call was silently dropped instead of becoming an `HTTP_CALLS` edge.

The obvious fix — add `"fetch"` to the `http_libraries` substring table in `service_patterns.c` — is wrong: that table is consulted by an early, unconditional check in **both** `pass_calls.c` and `pass_parallel.c` that runs before/regardless of whether the callee resolves to a real local definition (by design, so `axios.get(url)` still classifies even when the registry mis-binds bare `get` to an unrelated local method — the #523 bypass). That's safe for `"axios"`/`"requests"` because nobody names their own function that; it is **not** safe for `"fetch"`, which collides with a plausible local identifier (`function fetch(){}` or `const fetch = () => {}`).

Instead, `cbm_service_pattern_is_global_fetch()` is a new, narrow, exact-match check consulted only in the **empty-resolution fallback** in both files — the existing #523 path for unindexed external libraries. By the time that branch runs, the registry has already had its chance to resolve `"fetch"` to a local/imported definition; only a genuine miss reaches the new check. This also means a member call like `repo.fetch()` is naturally excluded — its `callee_name` is `"repo.fetch"`, not the bare `"fetch"` this checks for.

`pass_parallel.c`'s empty-resolution branch calls the low-level `emit_http_async_service_edge()` directly rather than going through `emit_service_edge()`, which re-derives its own classification from `res->qualified_name` via `cbm_service_pattern_match()` — a call with `"fetch"` would come back `CBM_SVC_NONE` there and silently fall through to a plain `CALLS` edge, quietly losing the fix.

Three new tests in `test_pipeline.c`:
- bare `fetch()` → `HTTP_CALLS`, sequential path (< 50 files)
- bare `fetch()` → `HTTP_CALLS`, forced through the parallel resolver (≥ 50 files) — the two call-resolution implementations are independent, so both need direct coverage
- a local `function fetch(){}` shadowing the global → plain `CALLS` to the local definition, zero `HTTP_CALLS` (the false-positive this PR must not introduce), bundled with a `repo.fetch()` member-call check in the first test

Refs #592.

## Checklist

- [x] Tests pass locally (`make -f Makefile.cbm test`) — full suite: 5936 passed, 1 skipped (pre-existing, unrelated), 0 failed
- [x] Lint passes (`make -f Makefile.cbm lint-ci`) — clang-format clean on all touched files; `lint-cppcheck`'s pre-existing failures (`extract_defs.c`, `compat_regex.c`) reproduce identically on unmodified `main` (verified via stash/pop), confirming they predate this change and aren't introduced by it

### DIFF
--- a/internal/cbm/service_patterns.c
+++ b/internal/cbm/service_patterns.c
@@ -782,6 +782,10 @@ void cbm_service_patterns_init(void) {
     /* No-op — tables are static const */
 }
 
+bool cbm_service_pattern_is_global_fetch(const char *callee_name) {
+    return callee_name != NULL && strcmp(callee_name, "fetch") == 0;
+}
+
 cbm_svc_kind_t cbm_service_pattern_match(const char *resolved_qn) {
     if (!resolved_qn || !resolved_qn[0]) {
         return CBM_SVC_NONE;
--- a/internal/cbm/service_patterns.h
+++ b/internal/cbm/service_patterns.h
@@ -11,6 +11,8 @@
 #ifndef CBM_SERVICE_PATTERNS_H
 #define CBM_SERVICE_PATTERNS_H
 
+#include <stdbool.h>
+
 /* Edge type returned by pattern match. */
 typedef enum {
     CBM_SVC_NONE = 0,      /* Not a service pattern — use normal CALLS */
@@ -31,6 +33,16 @@ void cbm_service_patterns_init(void);
  * Matches on library name substrings in the QN (e.g., "requests" in
  * "project.venv.requests.api.get"). Import-alias transparent. */
 cbm_svc_kind_t cbm_service_pattern_match(const char *resolved_qn);
+
+/* True for a bare, unqualified call to the native `fetch()` API. Deliberately
+ * NOT part of the substring tables above: those are matched unconditionally
+ * against the raw callee name before/regardless of registry resolution
+ * (the #523 external-library bypass), and "fetch" — unlike "axios" or
+ * "requests" — collides with a plausible local identifier. Callers must
+ * only consult this after registry resolution has come back empty, so a
+ * locally resolvable `function fetch(){}` / `const fetch = () => {}` is
+ * classified via its real resolved QN instead and never reaches this check. */
+bool cbm_service_pattern_is_global_fetch(const char *callee_name);
 
 /* Per-worker TLS cache for cbm_service_pattern_match results. The
  * pattern matcher runs once per resolved CALL edge in emit_service_
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -506,8 +506,15 @@ static int resolve_single_call(cbm_pipeline_ctx_t *ctx, CBMCall *call,
          * HTTP_CALLS/ASYNC_CALLS edge directly (target is a synthesized route
          * node, not the absent library). Without this the call is dropped and
          * cross-repo matching finds no edge to match (#523). The parallel path
-         * has the equivalent empty-resolution fallback in resolve_file_calls. */
+         * has the equivalent empty-resolution fallback in resolve_file_calls.
+         *
+         * Native `fetch()` (#856) belongs here too, not in the substring
+         * tables above: it only counts as the global API once resolution has
+         * already failed to find a local/imported `fetch` definition. */
         cbm_svc_kind_t esvc = cbm_service_pattern_match(call->callee_name);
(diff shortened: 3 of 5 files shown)


DeusData:
Thanks for the native fetch fix for #856. Triage: parsing/quality bug, normal priority.

Review focus is the shadowing boundary: bare global fetch should classify as HTTP_CALLS, but local functions or variables named fetch must not become outbound HTTP edges. The split between service pattern classification and local resolution is the important part here.

DeusData:
Merged — thank you! This is how we love fixes to arrive: you understood WHY the obvious substring-table approach was wrong, placed the check precisely in the empty-resolution fallback on BOTH resolver paths, and tested the shadow false-positive alongside the happy paths. Excellent work.

PR #939, opened 2026-07-07
Add Mojo language wiring

aaiyer:
## Summary

Adds the missing Mojo language wiring now that a Mojo tree-sitter grammar is vendored in the C backend.

## Root Cause

The repository includes `tree_sitter_mojo`, but the language was not reachable through the normal discovery/indexing path:

- no `CBM_LANG_MOJO` enum entry
- no `.mojo` extension mapping
- no user config language-name mapping
- no `CBMLangSpec` row for Mojo parser/query node types

As a result, `.mojo` files did not produce Mojo symbols during indexing.

## Changes

- Adds `CBM_LANG_MOJO`.
- Maps `.mojo` files to Mojo.
- Adds `mojo` to user-config language name lookup.
- Wires the vendored `tree_sitter_mojo` parser into `lang_specs` with function, class/trait, call, import, branch, assignment, and module node types.
- Adds language-extension and grammar-regression coverage.

## Validation

```sh
build/c/test-runner language grammar_regression
```

Result: `218 passed`.

I also verified the patched backend against a small `.mojo` sample locally: it indexed `fn`, `struct`, and method symbols, and outbound trace found a method calling a top-level function.


DeusData:
Thanks for picking up Mojo wiring. Triage: draft language-support PR, backlog priority.

Before review can proceed, please remove or fully justify the unrelated vendored/nomic/code_vectors_blob.S change. Mojo language wiring should not include an unrelated vendored/generated blob. Once that is gone, review can focus on the expected pieces: extension mapping, language spec wiring, user-config alias, and grammar regression coverage.

aaiyer:
Thanks for the triage. Addressed by force-pushing the PR branch back to the Mojo-only commit `222e97a0760ae320488a8228eea2a6521b3ec4dc`.

`vendored/nomic/code_vectors_blob.S` is no longer in the PR diff. The current diff is limited to the Mojo language wiring pieces: extension mapping, language spec wiring, user-config alias, and grammar/language regression coverage.

DeusData:
Merged — thank you! Clean wiring of the already-vendored grammar with sensible Python-family node mappings, and extending the label/regression guards to cover the new language is exactly right. Mojo users, rejoice.

PR #492, opened 2026-06-17
Fix Windows artifact refresh in write_file_atomic

Ayush7Ranjan:
## Summary

  Fixes Windows artifact persistence for `index_repository` with
  `persistence=true`.

  On Windows, `.codebase-memory/graph.db.zst` was created on the
  first run, but subsequent runs did not refresh/replace the
  artifact. The root cause was `src/pipeline/
  artifact.c::write_file_atomic()` using `rename(tmp, path)`,
  which does not reliably replace an existing destination file
  on Windows.

  This PR updates the atomic replace path so:

  - Windows uses `MoveFileExA(tmp, path,
  MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)`
  - POSIX keeps the existing `rename(tmp, path)` behavior

  ## Verification

  Built on Windows with MSYS2:

  ```powershell
  C:\msys64\usr\bin\bash.exe -lc "export HOME=/tmp; export
  PATH=/mingw64/bin:/usr/bin:$PATH; cd /c/Users/aayus/OneDrive/
  Desktop/opensource/codebase-memory-mcp; scripts/build.sh
  CC=gcc CXX=g++"

  Ran the Issue #400 reproduction twice with:

  $env:CBM_CACHE_DIR = ".\.codebase-memory"
  $env:CBM_PERSIST_EXPORTS = "true"

  Result:

  First run:
  artifact_present=true
  graph.db.zst exists=true
  size=7586364
  lastWriteUtc=2026-06-17T18:58:02.0340409Z

  Second run:
  artifact_present=true
  graph.db.zst exists=true
  size=9914129
  lastWriteUtc=2026-06-17T18:58:07.9280511Z

  timestamp_changed=True

  ## Fixes

  Fixes #400

PR #940, opened 2026-07-07
Fix SQLite writer atomic cache publish

### LINKED ISSUE #897: Installing a fresh DB file leaves the old -wal/-shm behind — SQLite replays the stale WAL over the new file
**Version:** codebase-memory-mcp 0.8.1 (built from source)
**Platform:** macOS (Apple Silicon) · **Install channel:** Built from source · **Binary variant:** standard

Three code paths install a brand-new DB file at a path where a previous DB generation lived, and none of them removes an existing `<path>-wal` / `<path>-shm` sidecar:

1. the pipeline dump — `cbm_gbuf_dump_to_sqlite` → the custom page writer opens the final cache path with `fopen(path, "wb")`;
2. `cbm_artifact_import` — renames the imported file over the cache path (it cleans up the *tmp* file's sidecars, not the destination's);
3. `cbm_store_dump_to_file` — whose comment asserts the old sidecars "become stale and get recreated by the next reader".

That assertion is wrong: SQLite decides whether to replay a WAL purely from the sidecar's own header/checksums, not from any tie to the main file. A leftover WAL from a crashed or never-closed previous session is therefore recovered **on top of** the freshly installed file at the next open, splicing old-generation pages into it — short indexes, `btreeInitPage` failures, or silently resurrected stale rows depending on which frames land. The full pipeline path has no pre-dump sidecar cleanup at all (the incremental path happens to unlink them first, so the two pipelines disagree).

Expected: any code path that installs a fresh DB file deletes the destination's `-wal`/`-shm` first.

### Reproduction

1. Open a store at `<path>`, insert a few hundred rows; while the connection is still open, copy the hot `<path>-wal` aside. Close cleanly (checkpoint removes the WAL).
2. Restore the copied `-wal` next to `<path>` — simulating a crashed session's leftover.
3. Rebuild `<path>` from scratch (re-index the project, or any of the three paths above).
4. Open the DB: the stale WAL is replayed over the new file. `PRAGMA quick_check` fails and/or row counts show the previous generation's data.

### Cascade

An independent producer of the same corruption class as #895 (torn fast export); like it, the damage is then hidden by #896 (scans swallow `SQLITE_CORRUPT`). Fourth companion issue (malformed-JSON emission) linked in a comment once created.

### Maintainer note

I have a working private fix for this (with reproduce-first regression tests) that I can submit as a PR. Before I do, I'd prefer you groom this issue — confirm the framing, tag, and prioritize — so the PR lands against an agreed shape.

### Confirmations

WarGloom:
## Summary

- write direct SQLite dumps to a sibling temp file and atomically publish them only after finalization succeeds
- discard temp DBs on writer errors instead of replacing the live cache with partial output
- remove destination `-wal` and `-shm` sidecars when installing the freshly rebuilt DB

## Root Cause

The custom SQLite page writer opened the final cache DB with `fopen(path, "wb")`. In shared MCP sessions, that can truncate a live DB while SQLite readers still have the old file mapped, and it can leave old-generation WAL/SHM sidecars beside the rebuilt DB. The observed crash reached SQLite `btreeInitPage` during `cbm_store_delete_file_hashes` after the dump/reopen step.

## Validation

- `git diff --check`
- `ASAN_OPTIONS=detect_leaks=0 build/c/test-runner sqlite_writer`
- `ASAN_OPTIONS=detect_leaks=0 build/c/test-runner pipeline`
  - passes 217 tests
  - note: existing unrelated UBSAN warning remains in `src/pipeline/pass_similarity.c:152`
- `make -f Makefile.cbm build/c/codebase-memory-mcp`
- `CBM_CACHE_DIR=/tmp/cbm-wiki-mcp-pr-full-20260707-1 build/c/codebase-memory-mcp cli --index-worker index_repository '{"repo_path":"/home/nikita/work/Projects/ai/wiki-mcp"}' --response-out /tmp/cbm-wiki-mcp-pr-full-20260707-1.response`
  - completed with `status:"indexed"`, `nodes:2256`, `edges:12934`

Fixes #897.

### DIFF
--- a/internal/cbm/sqlite_writer.c
+++ b/internal/cbm/sqlite_writer.c
@@ -27,6 +27,16 @@
 #include <stdint.h>
 #include <stdbool.h>
 
+#ifdef _WIN32
+#include <io.h>
+#include <process.h>
+#include <windows.h>
+#define cbm_writer_getpid _getpid
+#else
+#include <unistd.h>
+#define cbm_writer_getpid getpid
+#endif
+
 #define CBM_PAGE_SIZE 65536
 
 /* SQLite reserves the page containing the 1 GiB file offset (the "pending byte"
@@ -1707,6 +1717,8 @@ static uint32_t build_node_index_sorted(FILE *fp, uint32_t *next_page, CBMDumpNo
 typedef struct {
     FILE *fp;
     uint32_t next_page;
+    char final_path[CBM_SZ_4K];
+    char temp_path[CBM_SZ_4K];
     const char *project;
     const char *root_path;
     const char *indexed_at;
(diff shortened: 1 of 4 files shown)


DeusData:
Thanks for the SQLite writer publish fix. Triage: draft high-priority cache/store integrity bug.

Review will be cautious because this touches direct DB publication: temp-file atomicity, cleanup on failure, destination WAL/SHM sidecar removal, and live-reader behavior all need explicit tests. Please also rename the PR title before it is ready for review so it describes the change directly.

WarGloom:
Addressed.

- Renamed the PR to describe the change directly: "Fix SQLite writer atomic cache publish".
- Added explicit regression coverage for the remaining publish-integrity cases:
  - `sw_publish_failure_removes_temp_output` verifies a failed destination replacement discards the temporary DB output.
  - `sw_publish_preserves_live_reader` verifies a POSIX reader with an open transaction keeps seeing the old DB while a new connection sees the atomically published DB after replacement.
- Kept the existing coverage for pre-finalize non-truncation and destination `-wal`/`-shm` sidecar cleanup.

Local validation after rebasing onto current `main`:

- `ASAN_OPTIONS=detect_leaks=0 build/c/test-runner sqlite_writer`
- `bash scripts/check-no-test-skips.sh`
- `git diff --check`

DeusData:
This fixes a real and nasty class — truncating the live DB under concurrent readers (#897's btreeInitPage crash) — and temp-write + atomic publish + discard-on-error is the right design. The tests and the WAL/SHM sidecar cleanup are welcome. Three changes requested:

1. **Windows non-ASCII paths**: the new code opens the temp file with raw `fopen` and publishes via `MoveFileExA` (ANSI). On Windows, both break for non-ASCII cache paths (e.g. a user profile with umlauts — the cache DB name derives from the repo path). Repo policy is `cbm_fopen` (UTF-8 → `_wfopen`) for any user/cache path; please use it for the temp + sidecar writes here, and use `MoveFileExW` with a UTF-8→wide conversion for the publish (yes, `artifact.c` has the same ANSI call from #492 — that's a latent bug we'd rather not duplicate; feel free to extract a shared `cbm_rename_replace()` helper and we'll fix artifact.c in a follow-up).

2. **Durability of the publish**: there's no `fflush` + `fsync` before the rename on the POSIX side (Windows at least passes `MOVEFILE_WRITE_THROUGH`). Without it, a power loss can atomically publish an incompletely-written file — the exact corruption the temp+rename dance is meant to prevent. `fflush(fp); fsync(fileno(fp));` before `fclose` in the finalize path closes it.

3. **Sidecar removal ordering**: please confirm (and comment) that the destination `-wal`/`-shm` removal happens only on the success path relative to the rename — removing the WAL of a DB an old reader still has open is fine on POSIX (unlinked-but-open) but fails or misbehaves on Windows if a handle is held; a note on the expected ordering/failure mode would help the next reader.

With those, this is exactly the writer we want.

Issue #946, opened 2026-07-08
bug for parse c++


chen135790:
### Version

codebase-memory-mcp v0.8.1

### Platform

Linux (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

When a C++ file contains `#ifdef` / `#else` / `#endif` blocks around function signatures, the tree-sitter C++ parser silently skips all code from that point until a much later line, resulting in massive gaps in the knowledge graph.

### Reproduction

**File:** `SurfaceFlinger.cpp` (11,657 lines)

**Trigger pattern (line 2064):**
```cpp
#ifdef FLYME_GRAPHICS_EXTEND_LUMARGB
status_t SurfaceFlinger::addRegionSamplingListener(const Rect& samplingArea,
                                                   const sp<IBinder>& stopLayerHandle,
                                                   const sp<IRegionSamplingListener>& listener,
                                                   const bool rgbSample) {
#else
status_t SurfaceFlinger::addRegionSamplingListener(const Rect& samplingArea,
                                                   const sp<IBinder>& stopLayerHandle,
                                                   const sp<IRegionSamplingListener>& listener) {
#220 

### Logs

```text
## Bug Description

When a C++ file contains `#ifdef` / `#else` / `#endif` blocks around function signatures, the tree-sitter C++ parser silently skips all code from that point until a much later line, resulting in massive gaps in the knowledge graph.

## Reproduction

**File:** `SurfaceFlinger.cpp` (11,657 lines)

**Trigger pattern (line 2064):**

#ifdef FLYME_GRAPHICS_EXTEND_LUMARGB
status_t SurfaceFlinger::addRegionSamplingListener(const Rect& samplingArea,
                                                   const sp<IBinder>& stopLayerHandle,
                                                   const sp<IRegionSamplingListener>& listener,
                                                   const bool rgbSample) {
#else
status_t SurfaceFlinger::addRegionSamplingListener(const Rect& samplingArea,
                                                   const sp<IBinder>& stopLayerHandle,
                                                   const sp<IRegionSamplingListener>& listener) {
#endif
Result:
- Methods indexed: lines 463–2065, then 10619–11637
- Gap: lines 2066–10618 (~8,552 lines) completely missing from graph
- Critical methods SurfaceFlinger::commit (line 2758) and SurfaceFlinger::composite (line 3007) are not indexed
Expected Behavior
[...]

PR #949, opened 2026-07-08
fix(cpp): remap recovered definitions to original lines

blankanswer:
## Summary

Follow-up to #1050 for #946/#961. The recovery already on `main` remains authoritative; this PR adds explicit simplecpp expanded-to-original line mapping and the broader regression fixtures requested in review.

## What changed

- Track original line and main-file ownership for every expanded output line.
- Remap candidate definitions before ERROR-region intersection and adoption.
- Fail closed for included-header spans, unmapped directives, macro-generated callables, and call-site-only name matches.
- Preserve the existing #1050 constraints: candidates must intersect a raw ERROR region, have no existing qualified name, and pass original-source validation.
- Add the real #946 SurfaceFlinger fixture for both preprocessor branches.
- Add the requested RED-to-GREEN include-shift fixture: a 40-line included header makes #1050 skip the recovered def without explicit mapping; the mapped result recovers original lines 3-12 and clears the fully covered parse range.

## Validation

- `./build/c/test-runner extraction`: all 215 functional extraction tests passed; the only suite failure was the pre-existing wide-flat wall-clock threshold under this 894 MiB/swap-constrained runner.
- `./build/c/test-runner parse_coverage`: 9 passed.
- `make -f Makefile.cbm lint-format`: passed.
- `git diff --check`: passed.

The previous six-commit implementation was backed up locally and replaced with this single additive commit on current `main`.

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -849,6 +849,106 @@ static bool cbm_line_contains(const char *src, int src_len, uint32_t line, const
     return false;
 }
 
+static bool cbm_identifier_char(char c) {
+    return isalnum((unsigned char)c) || c == '_';
+}
+
+/* Verify that the mapped original span contains callable-definition syntax,
+ * not merely the name at a macro invocation or call site. */
+static bool cbm_span_contains_callable_def(const char *src, int src_len, uint32_t start_line,
+                                           uint32_t end_line, const char *name) {
+    if (!src || src_len <= 0 || !name || !name[0] || start_line == 0 || end_line < start_line) {
+        return false;
+    }
+    int span_start = 0;
+    uint32_t line = 1;
+    while (span_start < src_len && line < start_line) {
+        if (src[span_start++] == '\n') {
+            line++;
+        }
+    }
+    if (line != start_line) {
+        return false;
+    }
+    int span_end = span_start;
+    while (span_end < src_len && line <= end_line) {
+        if (src[span_end++] == '\n') {
+            line++;
+        }
+    }
+    size_t name_len = strlen(name);
+    for (int pos = span_start; pos + (int)name_len <= span_end; pos++) {
+        if (strncmp(src + pos, name, name_len) != 0 ||
+            (pos > 0 && cbm_identifier_char(src[pos - 1])) ||
+            (pos + (int)name_len < src_len && cbm_identifier_char(src[pos + name_len]))) {
+            continue;
+        }
+        int before = pos;
+        while (before > span_start && isspace((unsigned char)src[before - 1])) {
+            before--;
+        }
+        if (before > span_start && strchr("(,=!?[.", src[before - 1])) {
+            continue;
+        }
+        int open = pos + (int)name_len;
+        while (open < span_end && isspace((unsigned char)src[open])) {
+            open++;
+        }
+        if (open >= span_end || src[open] != '(') {
+            continue;
+        }
+        int depth = 0;
+        int close = -1;
+        for (int i = open; i < span_end; i++) {
+            if (src[i] == '(') {
+                depth++;
+            } else if (src[i] == ')' && --depth == 0) {
+                close = i;
+                break;
+            }
+        }
+        if (close < 0) {
+            continue;
+        }
+        for (int i = close + 1; i < span_end; i++) {
+            if (src[i] == '{') {
+                return true;
+            }
+            if (src[i] == ';') {
+                break;
+            }
+        }
+    }
+    return false;
+}
+
+/* Remap an expanded-source definition back to the original input file. Every
+ * line in the definition must be attributable to the main file; generated
+ * macro bodies, included headers, and ambiguous spans fail closed. */
+static bool cbm_remap_preprocessed_def(CBMDefinition *def, const CBMPreprocessedSource *pp) {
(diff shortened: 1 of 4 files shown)


DeusData:
Thanks for the focused C++ parser fix. Triage: parser recovery bug for #946, normal priority. Review focus is ownership of recovered callables, avoiding duplicate definitions across #ifdef branches, and keeping unrelated extraction behavior unchanged.

DeusData:
This is the fix we most wanted from the 0.9.1-rc bug queue — the coverage signal we shipped this week (#963/#968) flagged exactly this class on our own codebase (#961: `cbm_path_within_root`, `handle_process_kill` — both in-function `#ifdef` splits), and recovering missing callables from the already-existing preprocessor second pass is the right systemic answer: raw parse stays primary, expanded AST only fills verified gaps, deduped by label+QN. The defs-only walker entry point is a clean seam.

Two things before merge:

1. **Line coordinates of recovered defs.** The merged callables carry EXPANDED-source line numbers, but `get_code_snippet` (and the UI's GitHub deep-links) read the ORIGINAL file at `start_line..end_line` — a drifted range shows the wrong code for exactly the methods this PR rescues. Please either remap to original coordinates where feasible, or mark recovered defs (e.g. a `recovered_from_preprocessed` property / reduced confidence) and document the drift — plus one test pinning whatever behavior you choose, so snippet consumers aren't silently wrong.

2. **Rebase onto current main** — the end-of-extraction region of `internal/cbm/cbm.c` changed in #968: the parse-coverage signal now computes error regions AFTER extraction and subtracts recovered definitions. Your merge hook lands in the same neighborhood (conflict likely), and there's a pleasant interaction to verify once rebased: callables you recover should automatically UN-flag the corresponding `parse_partial` ranges (a region fully covered by recovered defs is dropped). The regression test `tests/test_parse_coverage.c` shows the mechanics; a one-line assertion in your test that the fixture is no longer parse_partial-flagged would prove the two features compose.

Genuinely excited about this one — with the rebase + the line-coordinate decision it merges.

blankanswer:
Thanks for the detailed review — both requested items are addressed:

* Rebased onto current `main`.
* Recovered callables now map expanded lines back to original main-file coordinates and fail closed when mapping is unreliable.
* Raw definitions remain primary/fill-only; successfully remapped callables are used for parse-coverage, and the #946 regression verifies both coordinates and cleared `parse_incomplete`/error ranges.

I also added coverage for both preprocessor branches, remap failure, header ownership, call-site ambiguity, and more than 256 recoveries. All CI checks are green. Ready for another look.

DeusData:
An update you should see before investing more here, with an apology attached: while working the twin issue #961 (same root cause — `#ifdef`-split braces → ERROR region → dropped defs), I landed a minimal version of the recovery on main in #1050 (65a176e) without cross-linking it to this PR first. That overlap is on me — you had the fix candidate in flight for #946 and had just addressed review.

What #1050 does (deliberately minimal): re-runs the defs walk on the existing simplecpp second pass and adopts only defs that (a) intersect a raw-tree ERROR region, (b) have their name visible on the corresponding original source line, and (c) are absent by qualified name. It relies on simplecpp's same-file line alignment rather than explicit mapping. #946's SurfaceFlinger shape verifies fixed with it.

Reading your latest revision, two things look like genuine improvements **over** what landed:

1. **Explicit expanded→original line mapping with fail-closed semantics** — my name-on-line check rejects misaligned defs but can't *correct* them; your mapping potentially recovers defs whose lines shift (e.g. resolved includes above the def), which #1050 silently skips.
2. **Your fixture set** — broader than the two shapes #1050 pinned.

If you're willing, I'd welcome a rebase of exactly those two deltas on top of current main as the follow-up: the line-map hardening (with a case #1050 provably skips — that would be its RED test) plus your fixtures. Smaller PR, clearly additive, and it keeps your authorship on the part that outlives both versions. If you'd rather not re-cut it, say so and we'll handle the salvage ourselves with credit to this PR either way.

Sorry again for the collision — and thank you for the careful iteration on this; the fail-closed mapping idea is the right instinct.

blankanswer:
Re-cut as requested on current `main` in 99d3919.

This is now a single additive follow-up to #1050: explicit expanded→original line/file ownership mapping, fail-closed original-span validation, and the salvaged fixtures. The new include-shift test is the requested case #1050 provably skips: a 40-line resolved header shifts the expanded def outside the raw ERROR coordinates; mapping recovers it at original lines 3–12, excludes header defs, and clears the fully recovered parse range.

Also retained the real #946 SurfaceFlinger shape for both branches and a macro-generated-callable rejection guard. `parse_coverage` is 9/9 and all 215 functional extraction tests pass locally; only the unrelated wide-flat wall-clock guard failed under this 894 MiB/swap-constrained environment. Ready for another look.

DeusData:
Merged (ef14822). This is exactly the clean additive follow-up I hoped for — explicit expanded→original line/file mapping with fail-closed span validation, layered on top of #1050's recovery rather than replacing it. Verified: my #1050 `#961` guards still pass unchanged, your new include-shift test passes (the case #1050's name-on-line heuristic couldn't handle — where resolved includes shift the expanded line numbers), 249/249 extraction green, lint clean. Thanks for the graceful re-cut after the overlap; the fail-closed mapping was the right instinct and the codebase is better for it.

Issue #50, opened 2026-03-11
Multiple instances of codebase-memory-mcp.exe running and 2 databases in the same project.


SlippyDong:
# Multiple persistent processes and redundant indexing with codebase-memory-mcp

### Environment
*   **OS:** Windows 11
*   **Editor:** VS Code
*   **Extension:** Codex (or similar MCP-enabled extension)

---

### Description
I have noticed that the `codebase-memory-mcp` server spawns multiple instances that remain running even after work is finished. In one session involving a single chat and 4 prompts (where I didn't even explicitly call the tool), multiple processes were active:

![Image](https://github.com/user-attachments/assets/7df46a21-1548-4d66-8d2c-e7c65e38fe54)

### Redundant Indexing & `cwd` Behavior
After restarting VS Code, the MCP did not start automatically (correct). However, upon requesting a re-index, I discovered that two separate databases were being created for the same project: one for the root and one for the `/src` folder.

**Index Status Output:**
```text
Reindex completed with codebase_memory for your workspace root.

Project: d-My Data-Projects-Private projects-my_project
Root: d:\My Data\Projects\Private projects\my_project
Status: ready
Index type: incremental (after full reindex run)
Nodes: 12319 | Edges: 16072

Also present (session project):
Project: D-My Data-Projects-Private projects-my_project-src
Root: D:\My Data\Projects\Private projects\my_project\src
Status: ready
Nodes: 4624 | Edges: 6678
```

**Resulting Cache Files:**
* `d-My Data-Projects-Private projects-my_project.db`
* `D-My Data-Projects-Private projects-my_project-src.db`

I had assumed setting the `cwd` in `config.toml` would define the MCP's root, but it seems to be indexing both the workspace root and the `cwd`.

**My `config.toml`:**
```toml
[mcp_servers.codebase_memory]
command = "D:\\Dev\\MCP\\codebase-memory-mcp\\codebase-memory-mcp.exe"
cwd = "D:\\My Data\\Projects\\Private projects\\my_project\\src"
startup_timeout_sec = 20
tool_timeout_sec = 120
```

---

### Process Proliferation per Chat
Each new chat session appears to spawn a unique MCP process. While archiving a chat kills some processes, others remain orphaned.

**Process Ownership:**
| PID | Process | Parent PID | Owner |
| :--- | :--- | :--- | :--- |
| 9880 | codebase-memory-mcp.exe | 1096 | codex |
[...]

PR #955, opened 2026-07-08
fix(ui): enumerate codebase-memory-mcp processes on Windows

### LINKED ISSUE #950: [bug] /api/processes returns empty array on Windows — Control Panel "Active Processes" shows nothing
### Version

codebase-memory-mcp 0.8.1

### Platform

Windows (x64)

### Install channel

### Binary variant

ui

On Windows, the Control Panel → "Active Processes" section is always empty, even when multiple `codebase-memory-mcp.exe` processes are running. On macOS/Linux the same section correctly lists all running `codebase-memory-mcp` processes (with PID, CPU%, RSS, elapsed time, command).

**Expected:** Windows should enumerate all running `codebase-memory-mcp.exe` processes, just like macOS/Linux does.

**Actual:** The "Active Processes" panel is always blank and renders the `control.noProcesses` placeholder ("No processes found" / "未找到进程"), making process monitoring and the "Kill" action unusable on Windows.

### Steps

1. Install `codebase-memory-mcp` v0.8.1 (UI variant) on Windows 11 via the official release archive / `install.ps1`.
2. Start the server:
   ```powershell
   codebase-memory-mcp serve --ui=true --port=9749
   ```
3. Open `http://localhost:9749` in a browser and switch to the **Control Panel** tab.
4. Observe that the "Active Processes" section is empty (renders `control.noProcesses`).
5. Sanity-check the API directly:
   ```powershell
   curl http://localhost:9749/api/processes
   ```
   **Windows output (bug):**
   ```json
   {"self_pid":12345,"self_rss_mb":42.0,"self_user_cpu_s":0.1,"self_sys_cpu_s":0.0,"processes":[]}
   ```
   **macOS output (correct):**
   ```json
   {"self_pid":67890,"self_rss_mb":38.0,"self_user_cpu_s":0.2,"self_sys_cpu_s":0.0,"processes":[{"pid":67890,"cpu":0.2,"rss_mb":38.0,"elapsed":"00:42","command":"codebase-memory-mcp","is_self":true}]}
   ```

### Root cause (source-level)

`src/ui/http_server.c` → `handle_processes()` (around line 251).

The `#ifdef _WIN32` branch (lines 255–275) only fills self-metrics, then **hardcodes** `"processes":[]}`:

```c
pos += snprintf(buf + pos, sizeof(buf) - (size_t)pos,
                "{\"self_pid\":%d,\"self_rss_mb\":%.1f,"
                "\"self_user_cpu_s\":%.1f,\"self_sys_cpu_s\":%.1f,\"processes\":[]}",
                (int)_getpid(), (double)rss_bytes / (1024.0 * 1024.0), user_s, sys_s);
```

The `#else` branch (macOS/Linux, lines 276–336) correctly enumerates processes via:

```c
FILE *fp = popen("LC_ALL=C ps -eo pid,pcpu,rss,etime,comm 2>/dev/null"
                 " | grep '[c]odebase-memory-mcp'", "r");
```

No equivalent enumeration is implemented in the Windows branch — it is simply missing.

### Suggested fix

Replace the hardcoded `"processes":[]}` in the Windows branch with a `CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS, 0)` enumeration, matching image filenames containing `codebase-memory-mcp` via `GetProcessImageFileNameA` (already available via `<psapi.h>`), then reading `GetProcessMemoryInfo` + `GetProcessTimes` per PID. Emit the same JSON shape as the macOS branch (pid/cpu/rss_mb/elapsed/command/is_self).

### Related (separate bug, mentioning for visibility)

`handle_process_kill` in the same file wraps its child-PID whitelist check in `#ifndef _WIN32`, so on Windows any PID (including unrelated user processes) can be terminated via `/api/process-kill`. Worth filing as a separate security/correctness issue.

### Environment

- `codebase-memory-mcp` 0.8.1 (UI variant)
- Windows 11 x64
- Install channel: GitHub release archive / `install.ps1`
- Confirmed not a duplicate against existing open issues.

### Logs

```text
$ curl http://localhost:9749/api/processes
{"self_pid":12345,"self_rss_mb":42.0,"self_user_cpu_s":0.1,"self_sys_cpu_s":0.0,"processes":[]}

$ tasklist /FI "IMAGENAME eq codebase-memory-mcp.exe"
Image Name                     PID Session Name        Session#    Mem Usage
========================= ======== ================ =========== ============
codebase-memory-mcp.exe      12345 Console                    1     42,108 K
codebase-memory-mcp.exe      12501 Console                    1     18,420 K

(Server reports zero processes despite two running instances being clearly visible to the OS.)
```

```text

```

_No response_

### Confirmations

EightDoor:
## Problem

On Windows, the Control Panel's **Active Processes** tab always shows empty because `handle_processes` in `src/ui/http_server.c` only queries the current process via `GetProcessMemoryInfo`/`GetProcessTimes` and returns `"processes":[]`. The POSIX path uses `popen("ps")` to enumerate all `codebase-memory-mcp` instances.

Closes #950

## Solution

Added Windows process enumeration via `CreateToolhelp32Snapshot` + `Process32First`/`Process32Next` to find all `codebase-memory-mcp.exe` instances. For each matching process:

- Open via `OpenProcess` with `PROCESS_QUERY_INFORMATION | PROCESS_VM_READ`
- Read CPU time via `GetProcessTimes` (user + system)
- Read memory via `GetProcessMemoryInfo` (WorkingSetSize)
- Calculate elapsed time from creation timestamp
- Output in matching JSON format with `is_self` flag

## Changes

- `src/ui/http_server.c`: Added `#include <tlhelp32.h>` and ~70 lines of process enumeration logic

## Edge Cases Handled

- `OpenProcess` failure (skips process)
- `GetProcessMemoryInfo`/`GetProcessTimes` failure (zeros fallback)
- `CreateToolhelp32Snapshot` failure (returns empty array)
- Buffer overflow protection consistent with POSIX path

## Test Plan

- [ ] CI: `build-windows` / `build-windows-arm64` jobs compile successfully
- [ ] Manual: Open Control Panel on Windows with multiple `codebase-memory-mcp` instances, verify Active Processes tab shows all instances with CPU/memory/elapsed values

### DIFF
--- a/src/ui/http_server.c
+++ b/src/ui/http_server.c
@@ -50,6 +50,7 @@
 #include <windows.h>
 #include <process.h>
 #include <psapi.h> /* GetProcessMemoryInfo */
+#include <tlhelp32.h> /* CreateToolhelp32Snapshot, Process32First/Next */
 #else
 #include <sys/stat.h>
 #include <unistd.h>
@@ -501,7 +502,7 @@ static void handle_processes(cbm_http_conn_t *c) {
     int pos = 0;
 
 #ifdef _WIN32
-    /* Windows: GetProcessMemoryInfo + GetProcessTimes */
+    /* Windows: GetProcessMemoryInfo + GetProcessTimes for the current process */
     PROCESS_MEMORY_COUNTERS pmc;
     FILETIME ft_create, ft_exit, ft_kernel, ft_user;
     double user_s = 0, sys_s = 0;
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks for taking the Windows UI process-list gap. Triage: Windows local-UI bug for #950, normal priority. Review focus is access-denied handling, stable JSON shape, and avoiding exposure of more process detail than the UI needs.

DeusData:
The fix itself looks right — `CreateToolhelp32Snapshot` + `Process32First/Next` filtered on our exe name, with `OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ)` per match, mirrors what the POSIX `ps` path exposes (localhost-only UI, own-binary processes — no new information surface). Closes a long-standing gap (#50) and CI is otherwise green.

One blocker: **DCO fails** — the commit is missing its `Signed-off-by` trailer. `git commit --amend -s` (or `git rebase --signoff`) and a push will turn it green; happy to merge right after.

DeusData:
Reviewed — this is a clean, well-built fix. Using the Win32 toolhelp snapshot (`CreateToolhelp32Snapshot`/`Process32First`) instead of shelling out to `tasklist`/`wmic` is exactly right (no command-injection surface, no PATH reliance), `OpenProcess` is read-only, and handle cleanup is correct (`CloseHandle` per iteration + on the snapshot). It mirrors the existing POSIX `ps` enumeration and adds no new surface.

One correction needed before merge: the PR body and commit both say `Closes #50`, but #50 is an unrelated, already-closed issue. The bug this actually fixes is **#950** ("/api/processes returns empty array on Windows"). Could you update the reference to `Closes #950` so it links and auto-closes the right issue? Once that's fixed I'll merge (with a merge commit, to keep your DCO sign-off). Thanks @EightDoor!

EightDoor:
I've already updated the DCO, and the bug fix for #950 has also been implemented. I hope this can be merged as soon as possible. Thanks for your help. Thank you @DeusData

DeusData:
Thanks for following up, and sorry this sat after you completed the requested work. The DCO correction and #950 reference are present on current head `c5ebab6a15b4a6aa557c9c6deead7b09afa57d81`, and the check set is green. You have no remaining triage action; this is back in exact-head maintainer review.

DeusData:
Thank you for this, and sorry for the long wait — you fixed the DCO promptly and then the branch sat. That was on us.

Reviewed at your current head, and the work is good. One small thing stands between it and a merge.

**What checks out.** Using `CreateToolhelp32Snapshot` + `Process32First/Next` rather than shelling out to `tasklist` or `wmic` was the right instinct: it removes any command-injection or PATH-hijack surface entirely, and it is faster and more robust than parsing tool output. It is genuinely *safer* than the POSIX path we already ship, which runs `popen("ps … | grep")`.

The matching is exact rather than loose — `_stricmp` against `codebase-memory-mcp.exe`, case-insensitive, which is correct for Windows and tighter than the POSIX substring grep. I traced the caller before worrying about that: `/api/processes` is display-only (`ControlTab.tsx` renders cards and gauges), and there is no kill or terminate path anywhere in the server or the frontend, so the enumeration cannot act on anything.

Handle hygiene is correct throughout: every successful `OpenProcess` is paired with a `CloseHandle`, the snapshot handle is closed after the loop, there are no early returns inside the loop, and an `OpenProcess` failure skips that PID and keeps scanning instead of aborting. `pe.dwSize` is set before `Process32First`. Access rights are read-only. And every added line — including the `tlhelp32.h` include — sits inside the existing `#ifdef _WIN32` branches, so the POSIX code is byte-identical and Linux/macOS cannot change.

**The one blocker: an uninitialized read.**

```c
FILETIME ftc, fte, ftk, ftu;
```

These are uninitialized, and the elapsed-time block reads `ftc` **unconditionally**. If `GetProcessTimes` fails, that is an uninitialized read — undefined behaviour — producing a garbage `elapsed` string. The PR describes a "zeros fallback", but that only covers the CPU and RSS values, not this.

In practice it is close to unreachable: a handle with `PROCESS_QUERY_INFORMATION` will essentially always satisfy `GetProcessTimes`. But "essentially always" is not a guarantee, and we do not knowingly merge UB — it is the kind of thing that surfaces years later on someone else's machine under a sanitizer.

Either of these closes it:

```c
FILETIME ftc = {0}, fte = {0}, ftk = {0}, ftu = {0};
```

or, cleaner, move the elapsed computation inside the `GetProcessTimes` success branch.

Push that and this merges.

**Two optional notes, neither blocking:**

- A `now > start` guard on the elapsed subtraction would harden against clock skew — if creation time is marginally in the future, the unsigned subtraction wraps to something enormous.
- `PROCESS_QUERY_LIMITED_INFORMATION` with a fallback (the pattern this codebase already uses in `src/daemon/ipc.c`) would also list elevated instances that `PROCESS_VM_READ` cannot open. Current behaviour silently omits them, which is safe but slightly incomplete.

Thanks again — mirroring the POSIX JSON shape exactly, including `is_self`, meant the frontend needed no changes at all. That parity work is why this review was straightforward.

EightDoor:
Pushed 440c845 — moved the elapsed-time computation into the GetProcessTimes success branch (so ftc is only read when valid) and added a now > start guard against unsigned underflow on clock skew.  Thank you for your reply @DeusData

DeusData:
Merged — thank you, and thank you for staying with it through four rounds of review.

This one merged **as your commit**, not distilled, which is worth saying out loud: it's the first community PR in this backlog pass to land unchanged, and it earned that. The Toolhelp32 approach is strictly safer than the `popen("ps")` enumeration we still use on POSIX — read-only handles, a hardcoded exe-name comparison against kernel-supplied data rather than any request input, and list-only with no terminate path anywhere near it. If anything it makes the POSIX side look like the one due for attention.

You also took the uninitialised `FILETIME` finding seriously when it would have been easy to argue it was unreachable in practice, and picked up the clock-skew guard while you were in there. That's the difference between a fix that works and one that stays working.

Landed in commit 440c8450 (merge dfe67cc7). Closing #950 with it.

The one thing this doesn't have is automated coverage, which is on us rather than on you — a test would need a Windows runner driving the live endpoint with concurrent instances, and that harness doesn't exist yet. Noted as a gap on our side, not a condition on yours.

Issue #903, opened 2026-07-06
Windows: get_code_snippet returns "(source not available)" and search_code returns 0 matches for projects indexed under non-ASCII (CJK) paths — indexing itself succeeds


Rakua-Laqua:
## Environment

- Windows 10 Pro (10.0.19045), Japanese locale (ANSI code page 932)
- Build: `main-zip-2026-07-06` (includes the #881 worker-spawn quoting fix and the earlier `_wfopen` / wide-argv fixes #637 / #703 / #891)

## Summary

With current main, indexing a repository whose path contains non-ASCII (CJK) characters now works:
`index_repository` spawns workers cleanly, symbols (Class/Method/Function) are fully extracted, and
the project name encodes non-ASCII bytes as UTF-8 hex.

However, the tools that **re-read files from disk at query time** still fail for such projects:

| Tool | Result on CJK-path project | Result on same repo via ASCII junction |
|---|---|---|
| `index_repository` (full) | ✅ nodes/edges identical to junction | ✅ |
| `search_graph` / `query_graph` | ✅ symbols, signatures, complexity all present | ✅ |
| `get_code_snippet` | ⚠️ metadata OK (start/end line, signature), but `"source": "(source not available)"` | ✅ full source returned |
| `search_code` | ❌ 0 matches for any pattern | ✅ matches returned |

So extraction-time file reads work, but the query-time read path (snippet source load, grep) apparently
still goes through a narrow-char API and fails on CJK paths.

## Repro

```powershell
mkdir C:\tmp\repro-日本語
@'
def multiply(a, b):
    return a * b

class Calculator:
    def square(self, x):
        return multiply(x, x)
'@ | Set-Content C:\tmp\repro-日本語\util.py -Encoding utf8

codebase-memory-mcp cli index_repository --repo-path "C:/tmp/repro-日本語" --mode fast
# -> status:"indexed", project "C-tmp-repro-e697a5e69cace8aa9e", symbols present:
codebase-memory-mcp cli query_graph --args-file <(echo '{"query":"MATCH (c:Class) RETURN c.name","project":"C-tmp-repro-e697a5e69cace8aa9e"}')
# -> Calculator ... (extraction works)

codebase-memory-mcp cli get_code_snippet --qualified-name Calculator --project C-tmp-repro-e697a5e69cace8aa9e
# -> "source": "(source not available)"   <-- BUG (line range + signature are correct)

[...]

PR #956, opened 2026-07-08
fix(windows): read search_code temp files as UTF-8

Cosm1cAC:
﻿## Summary

`search_code` can miss matches on Windows when the indexed project is under a non-ASCII path, such as a CJK directory name.

Scoped search writes two temporary files: one for the pattern and one for the indexed file list. Both files are UTF-8. The Windows command builder was reading them back with plain `Get-Content` before handing each path to `Select-String -LiteralPath`.

That is fragile on Windows PowerShell 5.1. UTF-8 without a BOM is decoded with the local ANSI code page unless the encoding is explicit. A path like `C:/Users/.../移动版agent终端/src/file.js` is read back as mojibake, so `Select-String` receives a literal path that does not exist and `search_code` returns a clean zero-match result.

This patch makes the Windows `search_code` command builder read the pattern file and scoped file list with `Get-Content -Encoding UTF8 -LiteralPath`.

Related to #903.

## Reproduction

I reproduced this on Windows PowerShell 5.1 with a project rooted under:

`C:/Users/<user>/Documents/移动版agent终端`

and a source file containing `agentReachCommandForAction`.

Before the fix:

- `Select-String -LiteralPath <actual file> -Pattern agentReachCommandForAction` finds the expected lines.
- The scoped `search_code` pipeline reads the UTF-8 file list through plain `Get-Content`.
- The CJK path is decoded as mojibake, so `Test-Path -LiteralPath <decoded path>` is false.
- `search_code` returns zero grep matches.

After the fix:

- The same file list is read with `Get-Content -Encoding UTF8 -LiteralPath`.
- The decoded path exists.
- `Select-String` finds the expected matches.

## Verification

Local command-level check on Windows PowerShell 5.1:

- current pipeline: `current_matches=0`
- patched pipeline: `patched_matches=2`
- default-decoded filelist path exists: `False`
- UTF-8-decoded filelist path exists: `True`

Also ran:

- `git diff --check`

I added a Windows-only regression test that creates a scoped `search_code` fixture under a UTF-8 CJK project root.

I could not run the full native Windows C test suite locally because this machine does not have the MSYS2/CLANG64 GNU Make toolchain used by the Windows CI job. Visual Studio `cl` is installed, but that is not this project's Windows CI build path.

## Checklist

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -4497,7 +4497,8 @@ static void build_grep_cmd(char *cmd, size_t cmd_sz, bool use_regex, bool scoped
             snprintf(
                 cmd, cmd_sz,
                 "powershell -Command \"$pat = Get-Content -Encoding UTF8 -LiteralPath '%s'; "
-                "Get-Content -Encoding UTF8 -LiteralPath '%s' | ForEach-Object { Select-String -LiteralPath $_ -Pattern $pat%s "
+                "Get-Content -Encoding UTF8 -LiteralPath '%s' | ForEach-Object { Select-String "
+                "-LiteralPath $_ -Pattern $pat%s "
                 "-ErrorAction SilentlyContinue }"
                 " | Where-Object { $_.Path -like '*%s' }"
                 " | ForEach-Object { $_.Path + [char]9 + $_.LineNumber + [char]9 + $_.Line }\"",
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for covering the Windows UTF-8 temp-file path. Triage: Windows query/read bug, high priority because it affects non-ASCII projects after indexing. Review focus is PowerShell 5.1 encoding behavior, CJK path coverage, and no regression to POSIX temp-file reads.

DeusData:
Merged — thank you! Surgical fix with the right PS 5.1 root cause (BOM-less UTF-8 decoded as ANSI), and the test locks it in. The Windows non-ASCII path class keeps biting; this closes another corner of it. Much appreciated.

PR #967, opened 2026-07-08
docs: fix npm README language count

ketpatil77:
## Summary
- align the npm package README with the repository README and badges
- correct the language count from 159 to 158

## Testing
- not run (docs-only change)

### DIFF
--- a/pkg/npm/README.md
+++ b/pkg/npm/README.md
@@ -27,7 +27,7 @@ Restart your agent. Say **"Index this project"** — done.
 
 - **Extreme indexing speed** — Linux kernel (28M LOC, 75K files) in 3 minutes. RAM-first pipeline with LZ4 compression and in-memory SQLite.
 - **Plug and play** — single static binary for macOS (arm64/amd64), Linux (arm64/amd64), and Windows (amd64). No Docker, no runtime dependencies, no API keys.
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks for the docs cleanup. Triage: docs-only PR, normal priority, assigned to 0.9.2-rc. Review focus is keeping the npm README aligned with the main README and the current language count; no product-code risk in this diff.

DeusData:
Thank you — you were right, and the fix has shipped as #1765 with `Co-authored-by: ketpatil77` on the commit.

I'm closing this rather than merging it because it turned out to be half the problem. `159 languages` appears **twice** in that README (line 10 as well as the bullet you fixed), and the same file also claimed `14 MCP tools` when the actual count is 15 — I verified that against the `TOOLS[]` table in `src/mcp/mcp.c` rather than trusting either README. @nhtkylc91 had independently spotted the tool-count half in #1268. Since both PRs touched the same file, merging them in sequence would have needed a rebase and *still* left line 10 wrong, so they're combined into one complete fix with both of you credited.

The underlying cause is that the package README drifts from the root README because nothing checks they agree. That's worth fixing properly at some point, and it wouldn't have been visible without someone noticing the number was off in the first place.

Thanks for the catch — small correct fixes to stale docs are genuinely useful, and easy to not bother with.

Issue #388, opened 2026-05-29
Add an install plan receipt before mutating agent configs/hooks


caioribeiroclw-pixel:
## Context

`codebase-memory-mcp` is doing the useful but sensitive thing: one installer can configure MCP entries, instruction files, Skills, and pre-tool hooks across many coding agents.

The README already says the tool reads the codebase and writes agent configuration files. That is a good disclosure. The missing trust primitive I would look for as a user is a machine-readable install plan/receipt before those writes happen.

## Proposal

Add an `install --plan` / `install --dry-run` mode, or have `install` emit a pre-mutation receipt before applying `-y`, with fields like:

```json
{
  "type": "agent.install.plan.v1",
  "project_root": "<path or redacted>",
  "agents_detected": ["claude-code", "opencode", "openclaw"],
  "agents_selected": ["claude-code"],
  "config_files_planned": ["~/.claude/settings.json", ".claude/skills/..."],
  "instruction_files_planned": [".claude/skills/.../SKILL.md"],
  "hooks_planned": [
    {
      "agent": "claude-code",
      "event": "PreToolUse",
      "tools": ["Grep", "Glob"],
      "blocking": false,
      "command_source": "installed shim"
    }
  ],
  "backups_planned": ["~/.claude/settings.json.bak"],
  "writes_started": false,
  "network_after_install": false,
  "next_safe_command": "codebase-memory-mcp install -y"
}
```

The important part is `writes_started=false`: users and agents can inspect the plan before any config/hook mutation happens.

## Why this matters

This project crosses several boundaries at once:

- MCP server registration
- agent instruction/Skill installation
- repo or user config edits
- pre-tool hook installation
- optional indexing / local graph artifacts

For coding agents, those are not just setup details. They decide what tools become visible, what instructions get loaded, and what commands can run around future tool calls.

## Acceptance sketch

- `codebase-memory-mcp install --plan` prints planned agent/config/hook/instruction writes and exits without mutating files.
- `install -y` either prints the same receipt before mutation or writes it to a stable path/log.
- The plan avoids raw source code, secrets, full environment dumps, prompts, transcripts, and raw tool output.
- The final install summary maps actual writes back to planned writes.
- If auto-detection selects multiple agents, the receipt makes that explicit.

This would make the installer easier to trust from Claude Code/OpenClaw/Cursor-style agents, because the agent can prove what it is about to mutate before it mutates the user's toolchain.

DeusData:
[...]

PR #977, opened 2026-07-09
feat(codex): install codebase-memory skill

### LINKED ISSUE #976: feat: install codebase-memory skill for Codex
## Summary

Codex currently receives MCP config, .codex/AGENTS.md, and a SessionStart reminder, but it does not receive the consolidated codebase-memory Skill that Claude Code gets. Codex now supports user skills from HOME/.agents/skills, so the installer can reuse the existing embedded codebase-memory skill there.

## Proposed change

- Install the existing codebase-memory Skill to HOME/.agents/skills/codebase-memory/SKILL.md when Codex is detected.
- Preserve .codex/AGENTS.md as lightweight always-loaded guidance.
- Include the skill path in install --plan.
- Remove only the managed Codex skill on uninstall.
- Update tests, smoke/security install checks, and README multi-agent table.

This is separate from #715, which tracks dynamic Codex PreToolUse/PostToolUse graph augmentation.

agn-7:
## Summary

- install the existing consolidated codebase-memory Skill for Codex users under HOME/.agents/skills/codebase-memory/SKILL.md
- preserve .codex/AGENTS.md as the lightweight always-loaded guidance while using the Skill for progressive-disclosure workflows
- include the Codex skill path in install --plan and remove the managed skill during uninstall
- update smoke/security install checks and the README multi-agent table

Closes #976

## Testing

- ./build/c/test-runner cli
- make -f Makefile.cbm cbm
- bash -n scripts/security-install.sh
- bash -n scripts/smoke-test.sh
- fake HOME install --plan check for .agents/skills without mutation
- fake HOME install/uninstall cycle verifying Codex skill write/remove and existing config preservation
- bash scripts/security-install.sh build/c/codebase-memory-mcp

Note: make -f Makefile.cbm test currently exits nonzero in existing graph/edge regression suites with no graph DB; the touched CLI suite passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,8 +374,8 @@ Restart your agent. Verify with `/mcp` — you should see `codebase-memory-mcp` 
 
 | Agent | MCP Config | Instructions | Hooks |
 |-------|-----------|-------------|-------|
-| Claude Code | `.claude/.mcp.json` | 4 Skills | PreToolUse (Grep/Glob graph augment, non-blocking) |
-| Codex CLI | `.codex/config.toml` | `.codex/AGENTS.md` | SessionStart reminder |
+| Claude Code | `.claude/.mcp.json` | 1 Skill (`codebase-memory`) | PreToolUse (Grep/Glob graph augment, non-blocking) + SessionStart + SubagentStart |
+| Codex CLI | `.codex/config.toml` | `.codex/AGENTS.md` + 1 Skill (`codebase-memory`) | SessionStart reminder |
 | Gemini CLI | `.gemini/settings.json` | `.gemini/GEMINI.md` | BeforeTool (grep reminder) + SessionStart reminder |
 | Zed | `settings.json` (JSONC) | — | — |
 | OpenCode | `opencode.json` | `AGENTS.md` | — |
--- a/scripts/security-install.sh
+++ b/scripts/security-install.sh
@@ -49,6 +49,7 @@ find "$HOME" -type f > "$TMPDIR/created_files.txt" 2>/dev/null || true
 # Expected path patterns (relative to HOME):
 #   .config/*/mcp.json (or .mcp.json variants)
 #   .claude/skills/*
+#   .agents/skills/*
 #   .claude/settings.json
 #   .continue/config.yaml
 #   .codeium/config.json
@@ -56,6 +57,7 @@ find "$HOME" -type f > "$TMPDIR/created_files.txt" 2>/dev/null || true
 #   Various agent config dirs
 
 EXPECTED_PATTERNS=(
+    ".agents/"
     ".claude/"
     ".cursor/"
     ".config/"
@@ -121,10 +123,18 @@ fi
 echo ""
 echo "--- Auditing skill file content ---"
 
-SKILLS_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/skills"
-if [[ -d "$SKILLS_DIR" ]]; then
-    SKILL_ISSUES=0
+SKILL_DIRS=(
+    "${CLAUDE_CONFIG_DIR:-$HOME/.claude}/skills"
+    "$HOME/.agents/skills"
+)
+SKILL_ISSUES=0
+SKILL_DIRS_FOUND=0
 
+for SKILLS_DIR in "${SKILL_DIRS[@]}"; do
+    if [[ ! -d "$SKILLS_DIR" ]]; then
+        continue
+    fi
+    SKILL_DIRS_FOUND=1
     while IFS= read -r skill_file; do
         basename=$(basename "$skill_file")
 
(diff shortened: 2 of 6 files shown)


DeusData:
Thanks for the PR. I have triaged this with #976 into 0.9.2-rc as a Codex integration enhancement. The review focus will be install/uninstall ownership boundaries, preserving existing Codex config, and making sure managed skill files are clearly scoped to codebase-memory-mcp.

agn-7:
@DeusData should I resolve the conflicts?

DeusData:
Thank you for this — and I am not closing it yet, because your PR raises a question about **main** that needs answering first.

**The feature itself has shipped.** Commit `a3903ca` (12 July, three days after you opened this) rebuilt the client-surface machinery and now installs the consolidated skill for Codex. So the goal is covered.

**But you and main disagree about where Codex actually reads skills from.** You install to **`~/.agents/skills`**; main installs to **`~/.codex/skills`**. If Codex only honours `~/.agents/skills`, then main has a real bug and *your* path choice was the correct one — in which case Codex users are getting nothing today. That is worth settling before anything is closed, so I have flagged it for the maintainer rather than assuming main is right because main is newer. If you know the answer from using Codex, that would genuinely help.

**Two things in your diff are still missing from main and are worth having on their own merits**, entirely independent of the skill feature:

1. **`scripts/security-install.sh` still uses a predictable temp path.** Lines 141-145 on main write to `/tmp/sec_skill_urls`. Your change to `$TMPDIR/sec_skill_urls` is a genuine hardening fix and it is not on main.
2. **The security audit does not cover `~/.agents/skills`** — `EXPECTED_PATTERNS` has no `.agents/` entry, even though main *now writes there* for Zed, OpenHands and Warp. So we are auditing a path set that no longer matches what we install. Your PR extends the audit to cover it.

Both are small, both are real, and I would like them regardless of what happens to the rest. If you would rather not carry a fresh PR for them, we can distill them with `Co-Authored-By` credit to you — your call.

**On the work itself:** the discipline here is exactly right. Zero new instruction content — you reused the already-reviewed embedded skill verbatim rather than writing a variant, all writes go through the existing machinery with `force` semantics preserved, and your `cli_codex_skill_force_overwrite` test proves stale content survives *without* `--force`. The plan-mode test asserting `.agents/skills` is absent after `--plan` shows you read the #388 contract properly. That is a higher standard than most install PRs reach.

Leaving this open until the directory question is answered.

agn-7:
Thanks for holding this open. I checked current Codex behavior from Codex CLI 0.146.0 and the current Codex manual.

The public manual documents personal skills at `$HOME/.agents/skills` and repo skills at `.agents/skills`. That is the basis for the path used in this PR.

At the same time, current Codex/app-server examples and this active Codex installation also show skills under `$CODEX_HOME/skills` (`~/.codex/skills` by default). In this session, Codex is exposing skills from both `~/.agents/skills` and `~/.codex/skills`.

So I would not claim the main branch `$CODEX_HOME/skills` path is a no-op for current Codex. I think the feature part of this PR can be treated as superseded by `a3903ca`, rather than resolving the conflicts here.

For the two remaining `scripts/security-install.sh` items you called out, I will open a fresh minimal PR from current `main`: add `.agents/` to the audited expected path set, and move the temporary URL scratch file under the `mktemp` directory.

agn-7:
Opened the focused follow-up PR here: #1373.

I left this PR open for maintainer disposition because it still contains the original feature discussion and directory-path context, but the remaining actionable security-install changes are now isolated in the smaller PR.

DeusData:
Taking the disposition decision you left us: **closing this as superseded** — and thank you for handling it the way you did, because it is genuinely the most useful way a contributor has answered a question like this.

I asked whether Codex honours `~/.agents/skills` or `$CODEX_HOME/skills`, and flagged that if only the former worked then `main` had a real bug and your choice was right. You went and checked Codex CLI 0.146.0 and the current manual, found that the public documentation specifies `$HOME/.agents/skills` **but** that current Codex builds also surface skills from `$CODEX_HOME/skills`, and reported that both were live in your session. Then you drew the conclusion that was against your own PR: you declined to claim `main`'s path was a no-op, and said the feature half here should be treated as superseded rather than conflict-resolved.

That is the hard version of the answer. It would have been very easy to report only the manual and let your PR land.

You then split the two items that were still missing from `main` into #1373 — the predictable-temp-path hardening in `security-install.sh` and extending `EXPECTED_PATTERNS` to cover `.agents/` — and **that merged on 31 July**, so the part of this work with independent value is already on `main` under your name. Nothing actionable remains in this diff.

**One thing your research leaves open, which I want on the record rather than lost in a closed PR.** `main` writes Codex skills to `$CODEX_HOME/skills` only; `.agents/skills` is used for Zed, OpenHands and Warp. That works today because Codex reads both. But the *documented* location is the one this PR chose, so if Codex ever narrows to its documented behaviour, our Codex skill install goes silently dead — the failure mode being that the skill simply never appears, with no error. Writing to both locations would be cheap insurance. If you would like to take that on, I would review it gladly; if not, it is captured here and we will pick it up.

The discipline in the original PR is worth restating too, since it is what made it quick to assess: no new instruction text at all — you rendered the existing embedded `skill_content` through the existing `cbm_install_skills` machinery rather than authoring a parallel copy, preserved `force` semantics, and respected the plan-mode contract by recording the path instead of writing it. That is exactly how a client surface should be added here.

Sorry this sat for a month after you had already done everything asked of you. Thank you for the research, for the split, and for leaving the call to us rather than just closing it.

Issue #875, opened 2026-07-05
Python: CALLS / trace_path blind to import aliases (import scan_bash as _scan_bash)


rudi193-cmd:
## Summary

When a Python module imports a function under an alias (`from … import scan_bash as _scan_bash`) and calls the alias (`_scan_bash(command)`), **`trace_path` inbound callers for the original symbol are empty** (or under-reported), even though production call sites exist.

This nearly caused a false “scanner is dead scaffolding” finding in our security audit — the graph reported test-only callers for `scan_bash` because production sites call `_scan_bash`.

## Environment

- **Binary:** `codebase-memory-mcp 0.8.1`
- **Project:** `home-sean-campbell-github-willow-2.0` (public)
- **Related audit:** `docs/audits/KART_SCANNER_BWRAP_GAP_AUDIT_2026-06-15.md` Finding 2

## Reproduction

**Import / call site** (`willow/fylgja/events/pre_tool.py`):

```python
from willow.fylgja.security_scan import (
    scan_bash as _scan_bash,
    ...
)
...
issues = _scan_bash(command)  # line ~845, PreToolUse Bash branch
```

**Graph query:**

```
trace_path(function_name="scan_bash", direction="inbound", depth=2)
```

Production caller via `_scan_bash` is not attributed as an inbound caller of `scan_bash`. Grep finds the alias call; graph CALLS resolution does not bridge alias → canonical name.

Same class as hook-layer heavy `import X as _X` style (security scanner, MCP shims).

## Expected

CALLS edges (or trace_path traversal) should resolve through Python import aliases so inbound callers of the **definition** are visible when the call site uses the **alias** binding.

## Impact

- Dead-code / blast-radius audits under-report production usage.
- “Zero production callers” conclusions require grep verification for any aliased-import codebase (common in hooks and facades).

## Related issues

Not the same as [#763](https://github.com/DeusData/codebase-memory-mcp/issues/763) (file-node fallback) or [#871](https://github.com/DeusData/codebase-memory-mcp/issues/871) (CommonJS `require`), but same symptom: **empty inbound callers despite real call sites**.

## Confirmations

- [x] Searched open issues for Python import alias / `as _` — no exact duplicate.
- [x] Public OSS reproduction (willow-2.0).

DeusData:
[...]

PR #979, opened 2026-07-09
fix(registry): resolve aliased direct-symbol imports in import_map (#875)

LA-10:
## What does this PR do?

`trace_path` returns no inbound caller for a function called through an import alias (`from m import f as g`; `g()`). No CALLS edge is emitted at all.

**Root cause** (`registry.c`, `resolve_import_map`): `build_import_map` stores `key = local_name` → `value = the IMPORTS edge's target QN`. Plain imports target the **Module** node (value = module QN); aliased direct-symbol imports target the **Function** node (value = full symbol QN). The bare-callee branch assumes the former and always appends:

```c
snprintf(candidate, sizeof(candidate), "%s.%s", resolved, prefix);
```

For `_scan_bash` that yields `...security_scan.scan_bash._scan_bash` matches nothing, edge dropped. Plain imports work because `module + "." + name` reconstructs the real QN. That asymmetry is why only aliases break, and why `resolve_import_map_bare_function` (module-valued `vals[]`) never caught it.

**Fix:** if `resolved` is itself a registered symbol, it *is* the target return it before appending. Cannot regress plain/dotted imports: `pass_definitions.c:337` registers only never Module, so module-valued entries can't match. `resolve_import_map` is shared by the sequential and parallel pipelines, so one change covers both.

Issue #875 

## Verification

`tests/test_registry.c::resolve_import_map_bare_alias` — RED on unmodified clone, GREEN with fix. Existing import_map tests stay green.

End-to-end on `willow-2.0` (the issue's public repro), `trace_path(scan_bash, inbound)`:

- **Before:** `_scan_shell_fragment`, `check_kart_task` (both `core/kart_task_scan.py`)
- **After:** adds `main` (`willow/fylgja/events/pre_tool.py`) — the production caller

Call-site identity preserved, per the triage note:
```
src=main  callee=_scan_bash  strategy=import_map  tgt=...security_scan.scan_bash
```

---

## Checklist


LA-10:
On where the fix lives, I went with making the resolver handle the value shape build_import_map already produces. The other option would be normalizing at construction instead (store the module QN plus the original leaf name, so the append rule holds for everything). That's arguably cleaner, but it means touching build_import_map in both pass_calls.c and pass_parallel.c and changing the contract at pipeline.h:180. Happy to redo it that way if you'd rather.
Also worth mentioning: while instrumenting resolve_file_calls I noticed these calls only reach the registry because the Python LSP misses them entirely lsp=MISS for _scan_bash/_scan_write/_scan_worst, but lsp=HIT with a correct callee_qn for plain scan_bash. So the LSP isn't tracking import ... as ... bindings. Doesn't overlap with this fix, but happy to file it separately if that's useful.

DeusData:
Thanks for digging into the import-map shape. I have triaged this as a 0.9.1-rc bug fix for #875. During review we will compare the local resolver fix with the alternative of normalizing build_import_map, and we should keep the LSP alias miss as a separate follow-up so this PR stays focused.

DeusData:
Merged — thank you! The asymmetry analysis (module-valued vs symbol-valued import-map entries) was spot-on, the four-line fix lands exactly at the root cause, and the end-to-end verification on the issue's repro repo is the gold standard. Much appreciated.

PR #985, opened 2026-07-09
fix(graph-ui): clear ADR content on delete

ShiroKSH:
﻿## Summary

Fixes ADR deletion in the graph UI. The delete action cleared React state and then saved through the stale state value, so the old ADR content could be posted back instead of an empty body.

## Changes

- Allows the ADR save helper to receive the content that should be posted.
- Sends an explicit empty ADR body from the delete action.
- Wraps the normal save click handler so React mouse events are not treated as content.
- Adds a regression test for deleting an existing ADR.

## Validation

- npm test -- --run src/components/StatsTab.test.tsx
- npm test
- npm run build
- git diff --check

### DIFF
--- a/graph-ui/src/components/StatsTab.test.tsx
+++ b/graph-ui/src/components/StatsTab.test.tsx
@@ -207,6 +207,53 @@ describe("StatsTab index modal", () => {
     await new Promise((r) => setTimeout(r, 400));
     expect(browseCalls()).toBe(before);
   });
+
+  it("posts empty ADR content when deleting an existing ADR", async () => {
+    let saved: unknown = null;
+    mockProjectsFetch((url, init) => {
+      if (url === "/rpc") {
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the graph UI fix. I have triaged this into 0.9.1-rc as a normal-priority UX/behavior bug fix. Review focus will be the explicit empty delete body and making sure the UI event handling stays consistent with the ADR endpoints.

DeusData:
Merged. Verified end-to-end with vitest: the new "posts empty ADR content when deleting" test is a genuine guard — RED without your fix (the delete re-posts the stale "old decision text" because the save read through the cleared-but-not-yet-committed React state), GREEN with it. Passing the intended content explicitly to the save helper is the right fix for the stale-closure bug. All 11 StatsTab tests pass. Thank you — clean ADR data-safety fix.

PR #991, opened 2026-07-09
fix(cypher): support coalesce() and other multi-arg scalar functions in WHERE (#874)

### LINKED ISSUE #874: query_graph Cypher: coalesce() in WHERE rejected (unexpected operator)
## Summary

The Cypher subset used by `query_graph` rejects `coalesce()` in `WHERE` clauses. This blocks common null-safe numeric filters (e.g. complexity / hot-path audits over optional graph properties).

## Environment

- **Binary:** `codebase-memory-mcp 0.8.1` (linux-amd64)
- **Project:** `home-sean-campbell-github-willow-2.0` (public: willow-2.0)

## Reproduction

```cypher
MATCH (f:Function) WHERE coalesce(f.transitive_loop_depth, 0) >= 2 RETURN f.qualified_name LIMIT 10
```

**Result:** parser error — `unexpected operator at pos 33` (at `coalesce`).

Direct property comparisons work; only the `coalesce()` wrapper fails.

## Expected

`coalesce(expr, default)` in WHERE (or documented as unsupported with a clear error listing supported functions).

## Impact

Audit queries that need null-safe thresholds must issue one query per property or avoid arithmetic — easy to miss when porting Cypher from other engines.

## Related

Cypher hardening / subset documentation. Distinct from [#237](https://github.com/DeusData/codebase-memory-mcp/issues/237) (DISTINCT ordering) but same tool surface (`query_graph`).

## Confirmations

- [x] Searched issues for `coalesce` — no duplicate found.
- [x] Public OSS index (willow-2.0).

harshitaajoshi:
## What does this PR do?

Fixes #874

`query_graph` rejected `coalesce()` in WHERE clauses with a misleading "unexpected operator" error, because the WHERE parser only accepted `var`, `var.prop` or `var:Label` on the left side of a condition. Since `coalesce()` was already fully supported in RETURN, this PR reuses that existing machinery (`parse_multiarg_func_item()` / `eval_multiarg_func()`) in the WHERE path, so `coalesce`, `substring`, `replace`, `left` and `right` now work as condition LHS with identical semantics to RETURN.

As suggested in triage, unrecognized functions in WHERE now also fail loudly with a clear message instead of "unexpected operator":

```
unsupported function 'foo' in WHERE (supported: coalesce, substring, replace, left, right)
```

Changes:
- `src/cypher/cypher.h`: added `func` / `args` / `arg_count` to `cbm_condition_t` (moved the `cbm_func_arg_t` typedef above it)
- `src/cypher/cypher.c`: extracted `parse_condition_lhs()` to recognize multi-arg function calls, routed evaluation through the existing `eval_multiarg_func()`, and freed the new fields in `expr_free()` / `free_where()` and the parser error paths
- `tests/test_cypher.c`: 5 regression tests, written reproduce-first, including the exact repro query from the issue (`coalesce(f.transitive_loop_depth, 0) >= 2`) against nodes with and without the property, string fallback and first-arg-wins, NOT/AND composition, `substring` in WHERE, and the clear-error case

Verified end to end with the production binary: indexed a small fixture repo and confirmed the repro query returns only the matching function, and the error case surfaces the new message.

Note on local verification: the full suite (5987 tests) passes locally on a non-sanitizer build because ASan startup deadlocks on my macOS version (Apple/LLVM issue, unrelated to this change); clang-tidy on the touched files shows zero new findings vs main and clang-format is clean. Relying on CI for the ASan run and cppcheck.

## Checklist

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -760,6 +760,13 @@ static void expr_free(cbm_expr_t *e) {
                 safe_str_free(&cur->cond.in_values[i]);
             }
             free(cur->cond.in_values);
+            safe_str_free(&cur->cond.func);
+            for (int i = 0; i < cur->cond.arg_count; i++) {
+                safe_str_free(&cur->cond.args[i].variable);
+                safe_str_free(&cur->cond.args[i].property);
+                safe_str_free(&cur->cond.args[i].literal);
+            }
+            free(cur->cond.args);
         }
         if (cur->right) {
             if (top < EXPR_FREE_STACK) {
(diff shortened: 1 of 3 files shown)


DeusData:
Thanks for the Cypher parser fix. I have triaged this into 0.9.1-rc as a normal-priority bug fix. Review focus will be keeping WHERE function-call semantics aligned with RETURN handling and preserving clear errors for unsupported functions.

PR #995, opened 2026-07-10
fix(incremental): name File nodes by basename, matching the full pipeline

### LINKED ISSUE #994: Incremental re-index names File nodes by relative path; full build uses basename — same tree yields diverging graphs
### Summary

An incremental re-index (`index_repository` on an already-indexed project) names File nodes by their **repo-relative path**, while a full build names them by **basename**. Node upserts match by qualified name, so re-indexing a touched file **renames its File node in place** — the same working tree yields two different graphs depending on whether it was indexed fresh or incrementally, and the divergence grows with every edited file (both added and modified files are affected).

Because File nodes are USAGE-edge sources, this breaks name-based traversal: walking inbound `USAGE` transitively dead-ends at any touched file's node, silently dropping everything that reaches a target *through* a recently added/modified class.

### Reproduction (v0.9.0, macOS arm64; also present on current main)

```sh
BIN=./codebase-memory-mcp; T=$(mktemp -d); I=$(mktemp -d); F=$(mktemp -d)
mkdir -p "$T/src"
printf 'public class Bee { public void buzz() {} }\n' > "$T/src/Bee.java"
printf 'public class Aye { private final Bee bee = new Bee(); }\n' > "$T/src/Aye.java"
CBM_CACHE_DIR="$I" "$BIN" cli index_repository "{\"repo_path\":\"$T\"}"     # initial
printf 'public class Cee { private final Bee bee = new Bee(); }\n' > "$T/src/Cee.java"   # ADD
printf '// touched\n' >> "$T/src/Aye.java"                                                # MODIFY
CBM_CACHE_DIR="$I" "$BIN" cli index_repository "{\"repo_path\":\"$T\"}"     # incremental
CBM_CACHE_DIR="$F" "$BIN" cli index_repository "{\"repo_path\":\"$T\"}"     # full, same tree
# then per cache: query_graph "MATCH (n:File) RETURN n.name"
```

Result:

| cache | File node names |
|---|---|
| incremental | `Bee.java`, **`src/Cee.java`**, **`src/Aye.java`** |
| full (same tree) | `Cee.java`, `Aye.java`, `Bee.java` |

USAGE edges into `Bee` correspondingly come from `src/Cee.java` (incremental) vs `Cee.java` (full).

### Root cause

`pipeline_incremental.c` ("Create File nodes for changed files (purge deleted them)") passes `changed_files[i].rel_path` as the *name* argument of `cbm_gbuf_upsert_node`, whereas `pipeline.c` ("Create File node") computes the basename first. The incremental site also loses the `{"extension": ...}` props the full build sets. Introduced with the RAM-first incremental rework (80680ea); given the comment says it re-creates what purge deleted, this looks like an argument slip rather than intent.

### Fix

PR to follow: mirror the full pipeline's node shape at the incremental call site (basename + extension props), plus a regression test in the incremental suite (`incr_file_node_name_stays_basename`) — red without the fix, green with it; full incremental suite passes (162/162).

spde:
Fixes #994.

The incremental path passed `changed_files[i].rel_path` as the File node **name** (the full pipeline uses the **basename**). Upserts match by QN, so an incremental re-index renamed a touched file's File node in place — the same tree produced different graphs fresh vs incrementally, diverging further with every edited file, and name-based inbound-USAGE traversal dead-ended at touched files' nodes (details + minimal repro in #994).

**Change:** mirror the full pipeline's node shape at the incremental call site — basename as the name and the same `{"extension": ...}` props (previously `"{}"`).

**Test:** adds `incr_file_node_name_stays_basename` to the incremental suite: modify a file, incremental re-index, assert no File node name contains `/`. Red without the fix, green with it; full incremental suite 162/162 (macOS arm64, ASan/UBSan runner).

Commit is DCO signed-off.

### DIFF
--- a/src/pipeline/pipeline_incremental.c
+++ b/src/pipeline/pipeline_incremental.c
@@ -848,8 +848,17 @@ int cbm_pipeline_run_incremental(cbm_pipeline_t *p, const char *db_path, cbm_fil
     for (int i = 0; i < ci; i++) {
         char *file_qn = cbm_pipeline_fqn_compute(project, changed_files[i].rel_path, "__file__");
         if (file_qn) {
-            cbm_gbuf_upsert_node(existing, "File", changed_files[i].rel_path, file_qn,
-                                 changed_files[i].rel_path, 0, 0, "{}");
+            /* #994: the name must be the BASENAME with extension props,
+             * mirroring the full build's File node (pipeline.c) — upserts
+             * match by QN, so any other name here renames the node in place
+             * and the incremental graph diverges from a full build. */
+            const char *rel = changed_files[i].rel_path;
+            const char *slash = strrchr(rel, '/');
+            const char *basename = slash ? slash + SKIP_ONE : rel;
+            char props[CBM_SZ_256];
+            const char *ext = strrchr(basename, '.');
+            snprintf(props, sizeof(props), "{\"extension\":\"%s\"}", ext ? ext : "");
+            cbm_gbuf_upsert_node(existing, "File", basename, file_qn, rel, 0, 0, props);
             free(file_qn);
         }
     }
--- a/tests/test_incremental.c
+++ b/tests/test_incremental.c
@@ -964,6 +964,36 @@ static int resp_lacks_key(const char *resp, const char *key) {
 
 /* ── list_projects ─────────────────────────────────────────────── */
 
+TEST(incr_file_node_name_stays_basename) {
+    /* #994: File nodes must keep their basename NAME through an incremental
+     * re-index (the bug renamed touched files' nodes to rel_path, diverging
+     * from a full build). Modify a file, re-index, assert no File node name
+     * contains a path separator. */
+    char path[512];
+    snprintf(path, sizeof(path), "%s/fastapi/applications.py", g_repodir);
+    FILE *f = fopen(path, "a");
+    ASSERT(f != NULL);
+    fprintf(f, "\n# incr_file_node_name_stays_basename touch\n");
+    fclose(f);
+
+    char *resp = index_repo();
+    ASSERT(resp != NULL);
+    free(resp);
+
+    double ms;
+    char *r = call_tool_timed("query_graph", &ms,
+                              "{\"project\":\"%s\",\"format\":\"json\","
+                              "\"query\":\"MATCH (n:File) WHERE n.name CONTAINS '/'"
+                              " RETURN n.name LIMIT 5\"}",
+                              g_project);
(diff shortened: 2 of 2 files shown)


DeusData:
Thanks for the clean incremental-index fix. I have triaged it into 0.9.1-rc as the high-priority fix candidate for #994. Review focus will be matching the full-index File node shape and checking incremental hash and FTS behavior for unintended side effects.

DeusData:
Reviewed — the production fix is exactly right and ready; the test needs a one-line update before this can merge.

**Production half (pipeline_incremental.c): verified correct.** Byte-for-byte parity with the full build's File node site (same `strrchr` basename, same `{\"extension\":...}` props shape, same upsert argument order), and scope is exactly #994. Nice catch on the props loss too.

**Test half: it currently fails on main — through no fault of yours.** #1013 (merged after you branched) switched the default `query_graph` output from JSON to a compact token format, so `\"rows\":[]` never appears in the default response anymore. Your PR's CI only ran the pr-smoke legs (the full test workflow doesn't trigger for this path set), which is why it looked green. Verified locally against your branch merged onto current main:

- as-is: `incr_file_node_name_stays_basename` **fails** (161/162) — assertion never matches the new output shape
- with `\"format\":\"json\"` added to the `query_graph` args: **162/162**
- sanity: with the amended test but your production fix reverted, the test goes RED (rows contain `src/...` names) — so it stays a genuine guard under the new format

Concretely, in the `call_tool_timed` args:

```c
"{\"project\":\"%s\",\"format\":\"json\","
"\"query\":\"MATCH (n:File) WHERE n.name CONTAINS '/'"
" RETURN n.name LIMIT 5\"}",
```

(`format:\"json\"` is the stable escape hatch for programmatic consumers, so the rest of your assertion works unchanged.)

If you push that one-liner (and ideally rebase onto current main so the full workflow revalidates), this is good to go from my side.

spde:
Done — thanks for the thorough review, and for catching the #1013 format change (agreed, the pr-smoke-only CI legs hid it).

- Rebased onto current main (8c9ac98).
- Added `"format":"json"` to the `query_graph` args in `incr_file_node_name_stays_basename`, exactly as suggested.
- Re-verified locally on the rebased branch (ASan/UBSan runner): incremental suite **162/162**. Red-green still holds — reverting only the production hunk (name back to `rel_path`, props `{}`) makes the test go RED (`161/162`, rows contain `src/...` names), so it remains a genuine guard under the token-format default.

Production hunk unchanged from your verified version. Ready when you are.

Issue #937, opened 2026-07-07
Watcher can repeatedly re-index dirty repos and cause large disk write amplification


wymfly:
## Summary

We observed  background watcher processes causing very large sustained disk writes during normal MCP usage. On a macOS workstation using Codex with several indexed repos, host write counters exceeded 1 TB/day.

The issue appears to be write amplification from the watcher/indexing path, not application logs.

## What we observed

- Multiple  processes were active for different repos.
- Some processes wrote hundreds of MB within a short sampling window while idle from the user's perspective.
- Codex SQLite TRACE/DEBUG/INFO logging was ruled out: log inserts were blocked and the log WAL stopped growing.
- Re-indexing repeated even without commits, because the worktree stayed dirty/untracked.

## Suspected cause

The watcher treats a dirty/untracked worktree as changed on every poll. If the dirty state persists, each poll can trigger another re-index.

The incremental indexing path reduces some parsing work, but persistence still appears to rewrite/replace the project SQLite DB and may also rewrite the compressed artifact. For larger repos this can turn each watcher poll into hundreds of MB of writes.

This is amplified when multiple MCP clients/sessions run in the same repo or overlapping repos, since each process can independently trigger indexing and there does not appear to be a cross-process index lock/deduplication guard.

## Expected behavior

The watcher should not repeatedly re-index the same unchanged dirty state.

Possible mitigations:

- Trigger background watcher re-index only on Git HEAD changes, or
- Track a dirty-state signature and re-index only when that signature changes, and
- Add a cross-process lock/deduplication guard so multiple MCP processes cannot rebuild the same repo index concurrently.

Longer term, incremental indexing should ideally avoid near-full DB/artifact rewrites when only a small set of files changed.

## Related issues

This seems related to several existing reports, but combines them into one disk-write-amplification failure mode:

- #841
- #867
- #335
- #713
- #895
- #897
- #914

DeusData:
Thanks for the write-amplification report. Triage: high-priority watcher/performance bug under #593.

Persistent dirty or untracked state should not cause repeated full re-index/write cycles while the user is idle. A fix should reproduce a dirty-repo polling loop and prove the watcher coalesces or suppresses repeat work without missing real file changes.

Artemonim:
[...]

Issue #841, opened 2026-07-04
watcher: continuous auto-sync re-indexing churn on Windows (real driver of #832)


DeusData:
During the #833 review we found the reported 20GB RSS / 100% CPU / 600MB-per-minute growth on 65 markdown files (1.3MB) cannot be produced by a single index run — the reporter's own workaround (CBM_WORKERS=1 **plus auto_index false**) points at the auto-sync loop re-indexing continuously on Windows and leaking per cycle.

To investigate:
- Why does the watcher consider an unchanged directory dirty every poll on Windows (mtime granularity? path-normalization mismatch in the git-status hash? CRLF)?
- Per-cycle retention: what survives repeated index cycles (candidate: mimalloc page_reclaim_on_free=0 stranding pages of exited worker heaps — see #833's one plausible line).
Reproduction should loop N watcher cycles on an unchanged fixture on windows-latest and assert RSS returns to baseline after cbm_mem_collect.

Refs #832, #833.

DeusData:
v0.9.0 includes Windows watcher/runtime and memory-retention fixes that may change this continuous auto-sync churn path, and #832 is now awaiting a reporter retest on the same class of symptom. Could you retest the unchanged-directory watcher loop on v0.9.0 and confirm whether it still re-indexes continuously? Marking awaiting-reporter.

DeusData:
The linked reporter path has activity again, so I am removing awaiting-reporter here and keeping this in the 0.9.1-rc maintainer queue. The next step is still to verify the unchanged-directory watcher loop on Windows and separate watcher churn from any retained-memory behavior.

kevinkavi2001:
I can reproduce a deterministic Windows freshness false-positive that may help isolate the unchanged-directory churn tracked here.

### Environment

- codebase-memory-mcp 0.10.0
- Windows 10, amd64
- Binary: `C:\Users\Uzer\AppData\Local\Programs\codebase-memory-mcp\codebase-memory-mcp.exe`
- Project: `D:\SuperDevPro`
- Filesystem: NTFS
- Project is not a Git repository

### Reproduction

1. Delete the existing project index.
2. Run a clean full `index_repository`.
3. Do not modify the repository files.
4. Immediately call `check_index_coverage` for several exact paths.
5. Compare the reported freshness with each file's fractional modification timestamp:
   `os.stat(path).st_mtime_ns % 1_000_000_000`.
6. Reindex again and repeat the coverage check.

[...]

PR #1001, opened 2026-07-10
fix: stop CPU/memory blowup on non-git projects (#713, #937, #841)

### LINKED ISSUE #713: auto_index uses tens of GB of memory in a non-git folder containing code
### Version

codebase-memory-mcp 0.8.1

### Platform

Linux (x64)

### Install channel

### Binary variant

standard

With `auto_index=true`, opening an agent in a **non-git** folder that contains code indexes the whole tree and pushes RSS into the tens of GB. `auto_index_limit=50000` does not help.

The limit guard counts files with `git ls-files | wc -l` (`src/mcp/mcp.c:4953`). In a non-git directory that returns `0`, so the limit is never reached and auto-index proceeds with a full filesystem walk.

**Expected:** the limit applies to non-git roots as well, or auto-index skips a root that has no `.git`.

### Reproduction

The same 60,000 files are indexed when the directory is non-git, but skipped once it is a git repo:

```bash
mkdir -p /tmp/ab-nongit /tmp/ab-git
seq 1 60000 | xargs -P8 -I{} touch /tmp/ab-nongit/f{}.py
cp /tmp/ab-nongit/*.py /tmp/ab-git/
( cd /tmp/ab-git && git init -q && git add -A && git -c user.email=t@t -c user.name=t commit -qm x )

codebase-memory-mcp config set auto_index true
INIT='{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"t","version":"1"}}}'

# non-git root: a project .db is created in the cache dir (full index ran)
( cd /tmp/ab-nongit && printf '%s\n' "$INIT" | timeout 8 codebase-memory-mcp )

# git root, same files: no .db (auto-index skipped on the file-count limit)
( cd /tmp/ab-git && printf '%s\n' "$INIT" | timeout 8 codebase-memory-mcp )
```

### Logs

```text

```

```text

```

_No response_

### Confirmations

bethzyy:
Fixes #713. Also addresses the non-git half of #937 / #841 (watcher re-index churn).

Non-git folders currently hit two independent resource-exhaustion bugs. They share one root cause — every guard in these paths assumes `git` answers for the project — so this PR fixes both; happy to split it into two PRs if you prefer (the mcp/discover and watcher changes are fully independent).

## 1. Memory: auto-index OOM guard never fires on non-git roots (#713)

`maybe_auto_index` counted files with `git ls-files | wc -l`:

- In a **non-git folder** the count is 0, so `auto_index_limit` can never trigger and auto-index walks/indexes the entire tree — tens of GB of RAM on large code folders (#713).
- On **Windows** the pipe was broken for everyone: cmd.exe passes single quotes literally and has no `wc`, so the guard silently returned nothing even for git repos.

**Fix** (`src/mcp/mcp.c`, `src/discover/discover.{c,h}`):
- Count `git ls-files` lines in C — portable, no `wc`, double quotes.
- When the git count is 0 (non-git root, or nothing tracked), fall back to a new `cbm_discover_count_files(root, limit)`: a bounded native walk that honors the same skip dirs, suffix filters, and language gate as discovery and **stops at `limit + 1`**, so arbitrarily large trees cost O(limit) work. It deliberately does not load gitignore — over-counting only makes the guard more conservative, and the path must stay cheap.

## 2. CPU: watcher re-indexes forever when a tree stays dirty (#937/#841)

Two stacked defects in `src/watcher/watcher.c`:

- `is_git_repo()` used `git rev-parse --git-dir`, which **walks up the directory tree**. A plain non-git folder nested under any unrelated ancestor repo is misclassified as git — and since every file in it is untracked in that ancestor, `git status` never comes back clean.
- `check_changes()` treated *"working tree is dirty"* as *"changed"*, so a persistently dirty tree triggered a **full re-index on every poll cycle (5–60s), forever**. For the nested non-git case this is a guaranteed infinite loop with zero user action; #937 measured >1 TB/day of disk writes from this class of churn.

**Fix**:
- A folder with no `.git` entry of its own **and** zero tracked files is treated as non-git (watcher skips polling, per the module's design). An empty freshly-`git init`-ed root keeps its `.git`; a monorepo sub-package keeps its tracked files — both stay watched.
- Boolean dirty check replaced with a **status signature**: FNV-1a over `git status --porcelain --untracked-files=all -- .` output, mixing in the mtime+size of each listed path. A poll re-indexes only when the signature *moves*:
  - a tree that merely stays dirty hashes identically → no churn;
  - a repeated edit to an already-dirty file changes mtime/size → still caught;
  - `-- .` scopes status to the watched subtree, so a monorepo package no longer churns on sibling changes;
  - the signature is committed only after a successful re-index, so failed runs retry;
  - submodule status is folded into the hash on POSIX (same coverage as the old `submodule foreach` check).

## Tests

- `watcher`/`discover`/`gitignore`/`git_context` (173) and `mcp` (145) suites pass under ASan+UBSan on Linux.
- New: `watcher_nested_non_git_skips` (baseline logs `strategy=none` for a non-git folder under a git ancestor), `discover_count_files_bounded` (count, early exit, edge cases).
- Updated: `watcher_continued_dirty` previously asserted the buggy every-poll re-index; it now asserts stay-dirty → no re-trigger, further edit → re-trigger.
- E2E with the production binary: a 15-file non-git folder with `auto_index_limit=10` logs `autoindex.skip reason=too_many_files files=11 limit=10` (`files=11` = early exit at limit+1); with `limit=100` auto-index completes normally.

`scripts/security-allowlist.txt` updated for the renamed/added `cbm_popen` call sites in watcher.c (no new call classes).

### DIFF
--- a/scripts/security-allowlist.txt
+++ b/scripts/security-allowlist.txt
@@ -19,8 +19,9 @@ src/cli/cli.c:popen:sha256 checksum computation via shasum
 # ── Watcher: git status polling (repo paths validated via cbm_validate_shell_arg) ──
 src/watcher/watcher.c:system:git repo detection (is_git_repo)
 src/watcher/watcher.c:cbm_popen:git HEAD hash (git_head)
-src/watcher/watcher.c:cbm_popen:git working tree status (git_is_dirty)
+src/watcher/watcher.c:cbm_popen:git working tree status signature (git_status_signature)
 src/watcher/watcher.c:cbm_popen:git file count (git_file_count)
+src/watcher/watcher.c:cbm_popen:git toplevel resolution (git_show_toplevel)
 src/watcher/watcher.c:popen:via cbm_popen wrapper calls
 
 # ── Git context: git metadata resolution (repo paths validated via cbm_validate_shell_arg) ──
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -1080,6 +1080,93 @@ void cbm_discover_free_excluded(char **excluded, int count) {
     free(excluded);
 }
 
+/* ── Bounded file counting (#713) ─────────────────────────────────
+ * Count the indexable files under repo_path without building a result list,
+ * stopping as soon as the count exceeds `limit`. Serves as the auto-index
+ * OOM guard for non-git roots, where `git ls-files` cannot provide a count.
+ * Honors the same hardcoded skip dirs, suffix filters, and language gate as
+ * full discovery (mode FULL) but deliberately NOT gitignore: over-counting
+ * only makes the guard more conservative, and this path must stay cheap. */
+int cbm_discover_count_files(const char *repo_path, int limit) {
+    if (!repo_path || !repo_path[0] || limit < 0) {
+        return 0;
+    }
+
+    struct stat st;
+    if (wide_stat(repo_path, &st) != 0 || !S_ISDIR(st.st_mode)) {
+        return 0;
+    }
+
+    /* Dynamic stack of absolute directory paths (heap-owned). */
+    int cap = CBM_SZ_64;
+    char **stack = malloc((size_t)cap * sizeof(char *));
+    if (!stack) {
+        return 0;
+    }
+    stack[0] = strdup(repo_path);
+    if (!stack[0]) {
+        free(stack);
+        return 0;
+    }
+    int top = 1;
+
+    int count = 0;
+    while (top > 0 && count <= limit) {
+        char *dir = stack[--top];
+        cbm_dir_t *d = cbm_opendir(dir);
+        if (!d) {
+            free(dir);
+            continue;
+        }
+
+        cbm_dirent_t *entry;
+        while (count <= limit && (entry = cbm_readdir(d)) != NULL) {
+            if (entry->is_dir) {
+                if (cbm_should_skip_dir(entry->name, CBM_MODE_FULL)) {
+                    continue;
+                }
+                char sub[CBM_SZ_4K];
+                snprintf(sub, sizeof(sub), "%s/%s", dir, entry->name);
+                /* safe_stat skips symlinks / junctions so the walk cannot
+                 * cycle or escape the root — same policy as discovery. */
+                struct stat sst;
+                if (safe_stat(sub, &sst) != 0 || !S_ISDIR(sst.st_mode)) {
+                    continue;
+                }
+                if (top >= cap) {
+                    int new_cap = cap * PAIR_LEN;
+                    char **grown = realloc(stack, (size_t)new_cap * sizeof(char *));
+                    if (!grown) {
+                        continue; /* OOM: skip subtree — count stays a lower bound */
+                    }
+                    stack = grown;
+                    cap = new_cap;
+                }
+                char *copy = strdup(sub);
+                if (copy) {
+                    stack[top++] = copy;
+                }
+            } else {
+                if (cbm_has_ignored_suffix(entry->name, CBM_MODE_FULL)) {
+                    continue;
+                }
+                if (cbm_language_for_filename(entry->name) == CBM_LANG_COUNT) {
+                    continue;
+                }
+                count++;
+            }
+        }
+        cbm_closedir(d);
+        free(dir);
+    }
+
+    for (int i = 0; i < top; i++) {
+        free(stack[i]);
+    }
+    free(stack);
+    return count;
+}
+
 void cbm_discover_free_ignored(cbm_ignored_file_t *ignored, int count) {
     if (!ignored) {
         return;
--- a/src/discover/discover.h
+++ b/src/discover/discover.h
@@ -157,6 +157,14 @@ int cbm_discover_ex2(const char *repo_path, const cbm_discover_opts_t *opts, cbm
                      cbm_ignored_file_t **ignored_out, int *ignored_count_out,
                      int *ignored_total_out);
 
+/* Count indexable files under repo_path with an early exit: stops walking as
+ * soon as the count exceeds `limit` and returns limit + 1 (#713). Honors the
+ * hardcoded skip dirs, suffix filters, and language gate (mode FULL), but not
+ * gitignore — over-counting only makes a size guard more conservative. Used
+ * as the auto-index OOM guard for roots where `git ls-files` has no answer
+ * (non-git folders). Returns 0 on a missing/invalid root. */
+int cbm_discover_count_files(const char *repo_path, int limit);
+
 /* Free an array of file info results. NULL-safe. */
 void cbm_discover_free(cbm_file_info_t *files, int count);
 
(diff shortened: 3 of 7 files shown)


DeusData:
Thanks for tackling this resource-safety class. I have triaged it into 0.9.1-rc as high priority. Since it touches discovery, watcher behavior, MCP paths, and the security allowlist, review will focus on non-git guard behavior, watcher churn, the allowlist delta, and confirming there are no new file, process, or network side effects.

DeusData:
Thanks for the resource-safety work here — and a heads-up that the ground shifted under the watcher half while this sat.

**The two fixes have diverged in reviewability, so I'd like to take you up on the split you offered:**

**1. Memory / non-git auto-index OOM (#713)** — `discover.c`/`mcp.c`, the `git ls-files | wc -l` → real file count fix. This half is clean, independent, and still exactly right. If you split it into its own PR I can review and merge it quickly — it's the higher-value, lower-risk half.

**2. Watcher non-git churn (#937/#841 non-git half)** — this now **conflicts substantively** with #1045 (merged since you opened this), which reworked the git dirty-detection path: `git_is_dirty` is gone, replaced by a content signature (`git_dirty_signature` = porcelain entries + per-file size/mtime), and baselines now commit only after a successful reindex. Your allowlist edit renaming `git_is_dirty`→`git_status_signature` and the `watcher.c` changes collide directly with that. The **non-git** gap you're addressing may well still be real (a non-git folder has no HEAD and no porcelain, so #1045's signature doesn't cover it), but it needs rebasing on top of #1045's rewrite rather than replacing it — please rebase the watcher half onto current main and reconcile with the new signature approach, so we keep both the git and non-git coverage.

The `security-allowlist.txt` change itself is benign (just `cbm_popen` call-site descriptions), no concern there.

Net ask: **split #713 out for a fast merge; rebase + reconcile the watcher half against #1045 as a second PR.** Happy to look at the watcher design with you if the non-git path is trickier than a straight rebase. Appreciate the depth on this class.

DeusData:
Thanks for this, and apologies for the slow first response. Queued for review.

One thing you can act on now: GitHub reports the branch as CONFLICTING against `main`, which has moved a long way since you opened it. CI is fully green (20 checks), so the conflict is the only mechanical blocker. A rebase would help — I would rather review the diff you intend than a stale one.

Non-git project handling touching three linked issues is a useful consolidation; I want to give it a proper read rather than a rushed one.

DeusData:
Thank you for this, and I owe you a straight account of what happened, because the honest summary is: your diagnosis was right on both counts, and then main independently arrived at the same conclusions while this PR sat unreviewed. That is our failure of timing, not a failure of your work.

**Your diagnosis was correct, twice over.**

`git ls-files | wc -l` returns 0 on a non-git root — so the `auto_index_limit` guard could never fire there — and it was silently broken on Windows besides, thanks to cmd.exe quoting and `wc` not existing. And the watcher was answering "is the tree dirty" when the question it needed to answer was "did the tree change", which is why a persistently-dirty worktree re-indexed on every single poll, forever.

Main now carries both fixes. `9789a7c` (12 July, which closed #937 two days after you opened this) landed a dirty-state signature — porcelain output plus per-file size and mtime, zero reserved for clean, committed only after a successful re-index. That is the same design you arrived at, independently. And `83c137d` added `cbm_discover_count_bounded`, a git-independent native bounded count with full discovery-filter parity, wired into the auto-index guard. The overlap is convergent, not wasted: two independent analyses reaching the same root causes is good evidence both were right.

**Why I am not asking you to rebase.** Since then the watcher was rebuilt on a supervised argv-spawned git harness — no shell at all, with deadlines, output caps, `-z` NUL-separated parsing, nanosecond-resolution mtimes and fail-closed error states. The shell-popen foundation this PR patches no longer exists. A rebase would not be a rebase; it would be a rewrite against a different substrate, and it would be a lot of work to arrive where main already is.

**But three of your ideas are still not fixed on main, and they are real.** I would rather not lose them:

1. **Nested non-git misclassification.** Main's `git_repo_status` is still a bare `rev-parse --git-dir`, so a non-git folder sitting under an unrelated ancestor repo is still watched *as* that ancestor repo. It no longer loops forever — the signature settles — but the folder is tracked against the wrong repository. Your own-`.git`-or-tracked-files classification is the right answer.
2. **No subtree scoping.** Main's status invocation has no `-- .`, so a watched monorepo sub-package still re-indexes when a sibling package churns.
3. **Wrong stat base for nested watch roots.** Main's `sig_fold_path_stat` joins `root_path + "/" + rel`, but porcelain paths are *toplevel*-relative. When the watch root sits below the toplevel, every stat silently fails and the mtime/size fold is lost entirely. Deriving the toplevel via `rev-parse --show-cdup` — and your MSYS/Cygwin POSIX-path catch in `18c5d21` — is a genuinely sharp piece of work and a live gap today.

**So, your call, and I will follow whichever you prefer:**

- **You carry them**, as one or more small PRs against the current watcher. Happy to point at the exact functions.
- **We distill them**, with `Co-Authored-By` credit to you on the commits, if you would rather not rebuild against a substrate that changed under you. I will not do this without your say-so.

Either way I would like to leave this PR open until you have had the chance to answer, rather than closing it out from under you.

One last thing worth saying: rewriting `watcher_continued_dirty` so it asserts stay-dirty-stays-quiet *and* a further edit re-triggers — instead of asserting the buggy every-poll behaviour it previously encoded — is exactly the test discipline we want. And thank you for the upfront "happy to split this into two PRs, they are fully independent" note. You were right about that too; the two halves ended up being fixed by two different commits on main.

DeusData:
Following up on the offer from July rather than leaving this open indefinitely: **the three findings that were still live have shipped as #1770**, with `Co-authored-by: bethz` on the commit.

To be straight about what happened to the rest — `main` independently fixed your two headline problems while this sat unreviewed. The unbounded non-git file count got a bounded walk, and the per-poll churn got a dirty-state signature. That's on our review latency, not on you, and it means most of the diff here was overtaken by work you never saw.

What was **not** fixed, and what I verified was still live before reimplementing:

1. **Nested non-git misclassification.** `rev-parse --git-dir` walks up, so a folder merely sitting under an unrelated repo answered yes and inherited that repo's permanently-dirty state. Your `.git`-of-its-own-or-tracked-files test is the right discriminator and is what landed.
2. **Subtree `-- .` scoping.** `git status` reports the whole repository regardless of `-C`, so monorepo packages cross-triggered on each other.
3. **`--show-cdup` rather than `--show-toplevel`.** This is the one I'd have got wrong without your PR. Porcelain paths are repository-relative, so a subdirectory project stat'ed every path against the wrong root — invisible at a repo root, silently degrading the signature to text-only everywhere else. And your reason for preferring `--show-cdup` is exactly right: MSYS/Cygwin git returns a translated absolute path from `--show-toplevel` that won't join onto a native path.

It's a reimplementation rather than a rebase because the watcher was rewritten onto a supervised argv-spawn harness in the meantime, so the shell-`cbm_popen` foundation your patch built on no longer exists. Both new tests are verified binding — each reddens for its own bug and neither disturbs the other.

Closing this as superseded. Thank you for a genuinely well-diagnosed report — three distinct root causes, one of which needed real Windows knowledge to spot, and you offered to split the PR yourself before anyone asked.

PR #1004, opened 2026-07-10
fix(discovery): exclude minified assets in all modes

fernandofielmann:
## Summary
- exclude `.min.js` and `.min.css` files in full, moderate, and fast indexing modes
- report minified files as deliberate `ignored-suffix` exclusions instead of coverage failures
- add discovery and index-status regression coverage for JavaScript and CSS bundles

## Test plan
- [x] `build/c/test-runner discover index_resilience` (97 passed)
- [x] `make -j3 -f Makefile.cbm lint-ci`
- [x] `scripts/test.sh`
- [x] rebuild a private Angular + ASP.NET index and confirm the minified bundle is absent from architecture results
- [x] preserve all 20 Angular-to-ASP.NET route links in private validation


fernandofielmann:
Opened against the upstream repository by mistake; recreating it in the fork.

PR #1007, opened 2026-07-10
fix: compose JAX-RS method-level @Path with class-level prefix

### LINKED ISSUE #1005: Java/JAX-RS: method-level @Path sub-routes inconsistently composed with class-level prefix
**Version:** 0.9.0 (macOS arm64, official install.sh)
**Stack:** Java 21 / Quarkus (JAX-RS, jakarta.ws.rs)

## Observed

In the same indexed repository, two JAX-RS resources with the same annotation pattern (class-level `@Path` prefix + method-level `@Path` sub-paths, verb annotation before `@Path`) produce different Route extraction results:

- Resource A: all 10 sub-routes composed correctly, e.g. `__route__POST__/api/v1/things/{}/archive`, `__route__GET__/api/v1/things/{}`.
- Resource B: only the class-level route is emitted (`__route__GET__/api/v1/widgets`). Method-level sub-paths such as `@Path("/count")` and `@Path("/{id}/read")` produce no Route nodes at all.

The inconsistency is worse than a uniform miss: consumers cannot tell whether an absent route means "endpoint does not exist" or "extractor skipped it".

## Shape of the failing resource (sanitized)

```java
@Path("/api/v1/widgets")
@RequestScoped
public class WidgetResource {
    public WidgetResource(
            ListWidgetsUseCase list,
            @ConfigProperty(name = "app.widgets.default-limit") int defaultLimit,
            @ConfigProperty(name = "app.widgets.max-limit") int maxLimit) { ... }

    @GET
    @Produces(APPLICATION_JSON)
    public PageResponse<WidgetResponse> list(@QueryParam("limit") String limit) { ... }

    @GET
    @Path("/count")
    @Produces(APPLICATION_JSON)
    public CountResponse count() { ... }   // <- no Route node emitted
}
```

The resource that works has the same `@GET` / `@Path` ordering; notable differences in the failing one are constructor parameters annotated with `@ConfigProperty` and generic DTO return types (`PageResponse<T>`).

## Expected

Method-level `@Path` composed with the class-level prefix consistently, as it already happens for Resource A (`/{}` normalization there is great).

Possibly related to #734 (Spring class-level prefix dropped), though here the failure is intra-repo inconsistent rather than uniform.

Happy to run diagnostics builds or provide `CBM_DIAGNOSTICS=1` output.

CharlesQueiroz:
Fixes #1005.

## Root cause

`extract_route_from_annotations()` returned on the **first** mapping annotation. For JAX-RS methods the verb and the path live in two sibling annotations:

```java
@GET
@Path("/count")
public CountResponse count() { ... }
```

The `@GET` matched first (`annotation_route_method("GET")`), defaulted the path to `"/"`, and the sibling `@Path` was never read: every method-level `@Path` was silently dropped. Class-level `@Path` prefixes were never recognized either, because a lone `@Path` carries no verb, so the scan reported "no route" and `spring_class_route_prefix()` returned NULL.

This also explains the "inconsistent" symptom in #1005: resources whose endpoints are exercised by RestAssured tests got Route nodes from the tests' full-URL string literals, masking the missing annotation-derived routes; resources without such tests showed only the class-level route.

## Fix

New `scan_route_annotations()` collects the whole annotation set in one pass: the mapping verb (+ optional inline path, Spring style) and a separate JAX-RS `@Path` value. Then:

- **method-level** (`extract_route_from_annotations`): a verb annotation is still required (a lone `@Path` sub-resource locator is not an endpoint); path = inline mapping path, else `@Path`, else `"/"`.
- **class-level** (`spring_class_route_prefix`): inline mapping path (`@RequestMapping("/api")`), else class-level `@Path`.
- Bare `@HEAD`/`@OPTIONS` verbs are now recognized alongside GET/POST/PUT/DELETE/PATCH.

Spring extraction is unchanged: mapping annotations keep their inline path and the collecting scan preserves the first-mapping-annotation-wins order.

## Validation

- New regression test `extract_java_jaxrs_path_composition_issue1005` (class prefix alone, prefix + method `@Path`, verb preserved).
- Full suite: 5986 passed, 1 skipped, 0 failures.
- End-to-end fixture (sanitized copy of the failing resource shape from #1005, constructor `@ConfigProperty` params and generic DTO return types included) indexed with the patched binary now yields:
  - `__route__GET__/api/v1/widgets`
  - `__route__GET__/api/v1/widgets/count`
  - `__route__POST__/api/v1/widgets/{}/archive`
  - Spring control `__route__GET__/api/v2/things/all` unchanged.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -1296,7 +1296,8 @@ static const char *annotation_route_method(const char *name) {
     }
     /* JAX-RS bare-verb annotations (@GET/@POST/...) — path comes from @Path. */
     if (strcmp(name, "GET") == 0 || strcmp(name, "POST") == 0 || strcmp(name, "PUT") == 0 ||
-        strcmp(name, "DELETE") == 0 || strcmp(name, "PATCH") == 0) {
+        strcmp(name, "DELETE") == 0 || strcmp(name, "PATCH") == 0 || strcmp(name, "HEAD") == 0 ||
+        strcmp(name, "OPTIONS") == 0) {
         return name;
     }
     return NULL;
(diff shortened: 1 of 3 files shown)


DeusData:
Thanks for the targeted JAX-RS fix. I have triaged it into 0.9.1-rc as the fix candidate for #1005. Review focus will be class-plus-method path composition and making sure Spring-style annotation extraction does not regress.

DeusData:
Reviewed and verified locally — the diagnosis and fix are exactly right; same two mechanical blockers as your #1008, plus one recommendation.

**Verified** (your branch merged onto current main, db330bf):
- your unit test passes; extraction suite 214/214
- end-to-end on a real index: the JAX-RS resource now yields `/api/v1/widgets` **and** `/api/v1/widgets/count` (previously both methods collapsed onto a single `__route__GET__/` node — independently reproduced on main before applying your branch)
- Spring non-regression confirmed end-to-end: `@RestController` + class `@RequestMapping("/api/orders")` + `@GetMapping("/list")` still composes to `/api/orders/list`

**1. DCO (blocking):** commit `07704e1` lacks a `Signed-off-by` matching its author — `git commit --amend -s` + force-push.

**2. clang-format (blocking):** two lines in `extract_defs.c` (~1299–1300, the HEAD/OPTIONS condition) need wrapping — `make lint-format` locally shows the exact form.

**3. Recommendation (would take this from good to complete):** add an end-to-end Route-node test alongside the unit test — `tests/test_edge_types_probe.c` has `et_routes_exact`, which asserts the exact Route-node **name set** through the full pipeline. That's the tier that matches the reporter's actual symptom (collapsed/missing Route nodes and the dedup onto `__route__GET__/`), and it guards the class-prefix composition + emission dedup path your unit test can't see. Shape:
```c
static const char *routes[] = {"/api/v1/widgets", "/api/v1/widgets/count", NULL};
ASSERT_TRUE(et_routes_exact(f, 1, routes));
```
If you'd rather keep this PR minimal, say so and we'll add the probe-tier guard in a follow-up after merge.

Known gap we're both fine deferring: `@ApplicationPath` base-path composition (cross-class, like Spring's cross-file case) — worth its own issue once this lands.

With DCO + format in and CI green, this is good to merge from my side. Two solid, well-scoped fixes in a row — much appreciated.

CharlesQueiroz:
Applied the two blockers: sign-off added and the HEAD/OPTIONS condition in extract_defs.c is clang-format clean. Also took the recommendation rather than deferring it: added a handles_jaxrs_java probe to test_edge_types_probe.c asserting the exact Route-name set {/api/v1/widgets, /api/v1/widgets/count} through the full pipeline, guarding the class-prefix composition and the emission dedup path. Full suite passes locally.

Issue #998, opened 2026-07-10
pass_semantic_edges: SEMANTICALLY_RELATED edges are nondeterministic across identical runs (~50% churn)


DeusData:
Three identical full-mode indexes of the same flask checkout with the same binary produced SEMANTICALLY_RELATED counts of **74 / 59 / 126**, with only ~50% of the edge set shared between runs (HANDLES: 214/213/214 = fine; SIMILAR_TO: 18/18/18 byte-stable). On django: 79 vs 58 across runs.

The comment at `src/pipeline/pass_semantic_edges.c:74` shows one worker-scheduling dependence was already fixed ("per-node max_edges budget no longer depends on worker scheduling") — but the remaining selection still varies with scheduling order, most plausibly the per-node edge budget consuming LSH candidates in completion order.

**Impact assessment:** absolute scale is tiny (~0.02% of the graph; advisory edge class), so not release-blocking — but it violates the deterministic-output property the indexer otherwise holds (MT indexing is byte-identical since the kernel-perf work), makes A/B comparisons noisy, and churns incremental artifacts.

**Fix direction:** make candidate ordering deterministic before applying the per-node budget (sort candidates by (score, qn) instead of arrival order), same recipe as the earlier fix at :74.

harshitaajoshi:
Hey, I'd like to work on this one! The fix looks straightforward: in `cmp_deferred_edge_canonical` (pass_semantic_edges.c), the current sort key `(i, c)` uses the candidate's LSH arrival rank, which can vary with worker scheduling. Replacing it with `(i, score desc, j)` makes the per-node budget always admit the strongest edges first, with the partner func index (already sorted by qn in phase 1a) as a stable tie-break. PR coming shortly!

DeusData:
Re-measured on current main (db330bf) and this no longer reproduces — closing as fixed by intervening merges.

**Evidence:** six full-mode indexes of the same flask checkout (fresh store each run): three back-to-back runs plus `CBM_WORKERS=2/8/16` stress variants. All six produce a **byte-identical** SEMANTICALLY_RELATED edge set — 129 edges, identical SHA over the sorted `source_qn|target_qn` fingerprint — versus the 74/59/126 churn measured when this was filed.

Two findings from the investigation worth recording:

[...]

PR #1014, opened 2026-07-10
fix(semantic): sort deferred edges by (score desc, j) for deterministic admission (#998)

harshitaajoshi:
## What does this PR do?
Fixes the ~50% SEMANTICALLY_RELATED edge churn across identical runs reported in #998. The root cause is `cmp_deferred_edge_canonical` sorting deferred edges by `(i, c)` where `c` is the candidate's LSH arrival rank, which varies with worker scheduling. This PR replaces the sort key with `(i, score desc, j)` so the per-node `max_edges` budget always admits the highest-scored edges first, with the partner func index (sorted by qualified_name in phase 1a) as a deterministic tie-break.
## Checklist
- [x] Every commit is signed off (`git commit -s`) -- required, CI rejects
      `tests/repro/repro_parallel_determinism.c` already guards this class of fix

### DIFF
--- a/src/pipeline/pass_semantic_edges.c
+++ b/src/pipeline/pass_semantic_edges.c
@@ -1034,16 +1034,23 @@ static void phase5c_build_lsh_buckets(const uint64_t *signatures, int func_count
 /* Phase 6b: serialize deferred edges from all worker buffers into the graph
  * buffer (sequential because gbuf isn't thread-safe). */
 /* Canonical order for candidate pairs: ascending discovering-func index,
- * then ascending candidate rank — the order a sequential scoring loop over
- * canonically-sorted funcs would have produced. */
+ * then DESCENDING score so the per-node budget admits the strongest edges
+ * first, then ascending partner func index as a stable tie-break (func
+ * indices are sorted by qualified_name in phase 1a, so this is equivalent
+ * to a qn tie-break).  The previous (i, c) key used the candidate's LSH
+ * arrival rank which could vary with scheduling, producing ~50% edge-set
+ * churn across identical runs (#998). */
 static int cmp_deferred_edge_canonical(const void *pa, const void *pb) {
     const deferred_edge_t *a = pa;
     const deferred_edge_t *b = pb;
     if (a->i != b->i) {
         return a->i < b->i ? -1 : 1;
     }
-    if (a->c != b->c) {
-        return a->c < b->c ? -1 : 1;
+    if (a->score != b->score) {
+        return a->score > b->score ? -1 : 1; /* descending */
+    }
+    if (a->j != b->j) {
+        return a->j < b->j ? -1 : 1;
     }
     return 0;
 }


DeusData:
Thank you for digging into the admission ordering — before you invest more here, an important status update: **the nondeterminism this PR targets no longer reproduces on current main.**

I re-ran #998's measurement on main (db330bf): six full-mode indexes of the same flask checkout — three back-to-back plus `CBM_WORKERS=2/8/16` variants — and all six produce a **byte-identical** SEMANTICALLY_RELATED edge set (129 edges, same SHA over the sorted `src|dst` fingerprint), versus the 74/59/126 churn from the issue. The deferred-admission rework noted in the file header (removing the budget's dependence on worker-completion order) appears to have resolved it; #998 is closed with the full evidence.

On the mechanism this PR proposes: the candidate rank `c` is assigned when LSH buckets are walked in ascending func index *after* the buckets are built sequentially in phase5c — so given a fixed `funcs[]` ordering, `(i,c)` is already a deterministic total order; the comparator change would alter *which* edges win ties but not remove any scheduling dependence. For it to be actionable we'd need a RED case on current main — a reproducible input where two identical runs differ. If you can produce one (the fingerprint procedure in #998's closing comment is cheap), I'll gladly re-triage; otherwise I'd suggest we close this once you've had a chance to verify on main.

Either way — thanks for engaging with one of the gnarlier corners of the pipeline; determinism reports and fixes are exactly the kind of contribution this project needs.

DeusData:
Thanks for the focused deterministic-admission fix for #998. This is a normal-priority 0.9.1-rc parsing bug. Review will verify stable ordering across worker schedules, score-descending admission within the per-node edge budget, deterministic tie-breaking, and no regression in semantic edge recall.

DeusData:
Closing this against the closed #998 — thank you for the careful determinism work, and apologies the ground shifted under it.

As detailed in #998's closing comment: I re-measured on current main (six full-mode flask indexes, three back-to-back plus CBM_WORKERS=2/8/16) and the SEMANTICALLY_RELATED edge set is now byte-identical across all runs — the ~50% churn from the original report no longer reproduces. The deferred-admission rework noted in the pass_semantic_edges.c header (which removed the budget's dependence on worker-completion order) resolved it before this PR.

On the mechanism your PR changes: the candidate rank c is assigned when LSH buckets are walked in ascending func index, and the buckets are built sequentially in phase5c before scoring — so (i,c) is already a deterministic total order given a fixed funcs[] ordering. The comparator change would alter which edges win ties but wouldn't remove a scheduling dependence, because there isn't one at that layer anymore.

I'm closing rather than merging since there's no longer a repro to fix, but if you can produce a case on current main where two identical runs differ (the fingerprint procedure in #998's closing comment is cheap and decisive), please reopen and I'll re-triage immediately. Genuinely appreciate you engaging with one of the gnarlier corners of the pipeline — determinism reports are exactly the kind of contribution this project values.

Issue #871, opened 2026-07-05
trace_path returns 0 callers for functions called via CommonJS require() — CALLS edge not created


trinhan2106:
### Version

codebase-memory-mcp 0.8.1

### Platform

Windows (x64)

### Install channel

npm

### Binary variant

standard

### What happened, and what did you expect?

When a function is imported with CommonJS `const f = require("./f")` and called in another module, no `CALLS` edge is created — the require binding is indexed as a disconnected `Variable` node (`in_degree: 0, out_degree: 0`). As a result `trace_path` (direction=inbound) returns `"callers": []` even though real call sites exist. Expected: the calling module is reported as an inbound caller.

ES-module imports in the same repo resolve fine (a TS helper correctly shows ~100 inbound callers), so this is specific to CommonJS require bindings. A helper required across ~300 files produced ~300 disconnected Variable nodes and 0 callers. The miss is silent, so agents trust the empty caller list and blast-radius analysis under-reports.

### Reproduction

1. Code — dummy snippet, two files:

```js
// src/bs/doThing.js
module.exports = async function doThing({ id }) {
  return id;
};
```

```js
// src/mutations/doThing.js
const doThing = require("../bs/doThing");

module.exports = async (parent, args) => {
  return doThing({ id: args.id });
};
```

2. Commands:

```
codebase-memory-mcp cli index_repository '{"repo_path":"<repo>"}'
codebase-memory-mcp cli trace_path '{"project":"<project>","function_name":"doThing","direction":"inbound"}'
```

3. Result:

```json
{"function":"doThing","direction":"inbound","callers":[]}
```

Expected: the wrapper in `src/mutations/doThing.js` listed as an inbound caller.

Graph evidence: `query_graph` with `MATCH (n) WHERE n.name = 'doThing' RETURN n.qualified_name, labels(n)` shows the require binding in the calling file as a `Variable` node with no edges, alongside the real `Function` node of the definition.

### Logs

```text

```

### Diagnostics trajectory (memory / performance / leak issues)

```text

```

### Project scale (if relevant)

92,080 nodes / 263,791 edges / 16,169 files

### Confirmations

- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

DeusData:
[...]

Issue #773, opened 2026-07-02
SIGABRT (invalid free in ts_stack_delete via cbm_destroy_thread_parser) on second index_repository in one server process — server dies silently


ericsonjulio1:
## Summary

The MCP server crashes with `SIGABRT` (`malloc: pointer being freed was not allocated`) on the **second `index_repository` call within one server process**, whenever both indexed directories contain roughly 30+ source files. The process dies silently — nothing on stderr — so the MCP client just sees `MCP error -32000: Connection closed`, and all `codebase-memory` tools are gone for the rest of the client session.

This makes any long-lived session that indexes two projects (or re-indexes one) kill the server. It likely explains "server randomly disconnected" reports, since the first index works and the crash only hits the next one.

## Environment

- codebase-memory-mcp **v0.10.0**, prebuilt release binary
- macOS arm64 (Darwin 25.5.0, M-series), 8 GB RAM (`mem.init budget_mb=4096 total_ram_mb=8192`)
- Client: Claude Code via stdio transport (also reproduces with no client at all — see repro)

## Crash report (macOS .ips, faulting thread)

Five separate crashes across three days, all with this identical stack:

```
__pthread_kill
pthread_kill
abort
malloc_vreport
malloc_report
___BUG_IN_CLIENT_OF_LIBMALLOC_POINTER_BEING_FREED_WAS_NOT_ALLOCATED
ts_stack_delete
ts_parser_delete
cbm_destroy_thread_parser
extract_worker
cbm_parallel_for
cbm_parallel_extract
cbm_pipeline_run
handle_index_repository
cbm_mcp_server_handle
```

So an `extract_worker` in the second pipeline run tears down its thread parser and `ts_stack_delete` frees a pointer the allocator doesn't own — looks like a per-thread parser (or its `ts_stack`) surviving from the first run and being destroyed again, or being freed against the wrong allocator/arena after the first run's cleanup.

## Deterministic repro (no client needed, ~30 s)

```bash
#!/bin/bash
# Generates two unrelated dirs of tiny .py files, then indexes both in ONE server process.
# First index succeeds; the second crashes the server (5/5 runs on my machine).
BASE=$(mktemp -d)
mkdir -p "$BASE/dirA" "$BASE/dirB"
for i in $(seq 1 30); do
[...]

PR #1030, opened 2026-07-11
fix(pipeline): valid JSON for sequential-path service edges and Route nodes

### LINKED ISSUE #898: Sequential pipeline emits invalid JSON: unterminated "broker" on ASYNC_CALLS edges, raw non-JSON Route properties
**Version:** codebase-memory-mcp 0.8.1 (built from source)
**Platform:** macOS (Apple Silicon) · **Install channel:** Built from source · **Binary variant:** standard

Two JSON-emission bugs in `pass_calls.c` (the **sequential** pipeline path — small repos; the parallel path is correct):

1. The ASYNC_CALLS edge props format closes the `method` field with a quote but not `broker`, producing `"broker":"bullmq}`. The fixup block beneath it only repairs the truncation case (`props[plen-1] != '}'`), which never fires in the normal case — so **every brokered ASYNC_CALLS edge** gets malformed `properties`.
2. `create_svc_route_node` stores the **raw** method/broker string as the Route node's `properties` (e.g. literally `bullmq` instead of `{"broker":"bullmq"}`).

Consequence: any `json_extract` over those rows errors — including the `edges` generated columns (`local_name_gen`, the `url_path` index), so index-backed queries over such edges can fail, and `PRAGMA quick_check` (which recomputes generated columns) aborts with `malformed JSON` instead of completing. The parallel path's `build_service_route` / `emit_http_async_service_edge` handle both correctly, so the two pipelines produce different bytes for the same input.

Expected: the sequential path emits the same valid JSON as the parallel path.

### Reproduction

Index this snippet as a small repo (small file counts take the sequential path):

```js
const { Queue } = require("bullmq");
const mailQueue = new Queue("mail");
async function notify() { await mailQueue.add("welcome-email", {}); }
```

Then:

```sql
SELECT count(*) FROM edges WHERE json_valid(properties)=0;  -- expect >= 1
SELECT count(*) FROM nodes WHERE json_valid(properties)=0;  -- expect >= 1 (Route nodes)
PRAGMA quick_check;  -- aborts: "malformed JSON"
```

Observed field values: edges with `{"callee":"mailQueue.add","url_path":"welcome-email","broker":"bullmq}` and Route nodes whose `properties` column is the bare string `bullmq`.

### Cascade

Found while root-causing #895/#896: a `quick_check`-based integrity gate (the natural fix for #895) trips over these rows, so this needs fixing first or the gate must tolerate non-corruption `quick_check` aborts — otherwise projects with brokered async calls would re-index into the same state repeatedly. Companion: #897.

### Maintainer note

I have a working private fix for this (with reproduce-first regression tests) that I can submit as a PR. Before I do, I'd prefer you groom this issue — confirm the framing, tag, and prioritize — so the PR lands against an agreed shape.

### Confirmations

DeusData:
Closes #898.

Both sequential-path emission bugs from the report, fixed to mirror the parallel path byte-for-byte:
1. **ASYNC_CALLS edge props**: the single format string never closed the broker value's quote (`"broker":"bullmq}`) and the fixup beneath only handled truncation — the normal case was malformed on every brokered edge. Rebuilt as the parallel path's incremental construction.
2. **Synthesized Route nodes**: raw `bullmq`/`GET` stored as `properties` instead of `{"broker":"bullmq"}`/`{"method":"GET"}`.

These rows broke `json_extract`, the edges generated columns, and `PRAGMA quick_check` — and post-#1019, such caches would be **refused at artifact import**, upgrading this from cosmetic to data-loss-adjacent.

**Reproduce-first:** celery fixture through the real sequential pipeline — RED with exactly 1 malformed edge + 1 malformed Route node; GREEN now, with non-vacuous guards (≥1 ASYNC_CALLS edge must exist; broker json_extract-able).

**Side finding:** the report's bullmq JS repro no longer *classifies* on current main — CommonJS `require()`/ESM import resolution doesn't produce the resolutions the service matcher needs. That's precisely open #871 (probed and confirmed with three fixtures while building this).

Full suite 5,995/0, lint-ci clean. Fourth of kriswill's store/pipeline-integrity series — all now fixed.

### DIFF
--- a/src/foundation/compat_fs.c
+++ b/src/foundation/compat_fs.c
@@ -714,6 +714,51 @@ int cbm_exec_no_shell(const char *const *argv) {
 
 #endif /* _WIN32 */
 
+/* Canonicalize an EXISTING path (collapse `..`, resolve per-OS): realpath on
+ * POSIX; on Windows a wide-path GetFileAttributesW existence check +
+ * GetFullPathNameW. The previous callers used the ANSI CRT (_access/
+ * _fullpath) on UTF-8 input — locale-dependent by construction: on a CJK
+ * system codepage (e.g. Big5) the UTF-8 bytes of a CJK path re-decode into
+ * different characters and canonicalization corrupts the path (#973).
+ * Returns 0 when the path does not exist or cannot be resolved. */
+int cbm_canonical_path(const char *path, char *out, size_t out_sz) {
+    if (!path || !out || out_sz == 0) {
+        return 0;
+    }
+#ifdef _WIN32
+    wchar_t *wpath = cbm_utf8_to_wide(path);
+    if (!wpath) {
+        return 0;
+    }
+    if (GetFileAttributesW(wpath) == INVALID_FILE_ATTRIBUTES) {
+        free(wpath);
+        return 0;
+    }
+    enum { CANON_WIDE_MAX = 4096 };
+    wchar_t wfull[CANON_WIDE_MAX];
+    DWORD n = GetFullPathNameW(wpath, CANON_WIDE_MAX, wfull, NULL);
+    free(wpath);
+    if (n == 0 || n >= CANON_WIDE_MAX) {
+        return 0;
+    }
+    char *utf8 = cbm_wide_to_utf8(wfull);
+    if (!utf8) {
+        return 0;
+    }
+    size_t len = strlen(utf8);
+    if (len >= out_sz) {
+        free(utf8);
+        return 0;
+    }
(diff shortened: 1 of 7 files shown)


DeusData:
Heads-up for the history: the first commit on this branch accidentally bundled a stale-base delta (it would have reverted #1027/#1028 — caught before merge, auto-merge was disarmed). The second commit restores those files; the **net PR diff is exactly the intended two-file #898 fix** — please review the Files tab (net), not per-commit. Suite green (both the #898 and #773 guards pass), lint clean.

PR #1033, opened 2026-07-11
Feature/plsql support

ouzsrcm:
## Checklist


DeusData:
Thanks for the PL/SQL contribution. I verified the vendored parser against AndreasMaierDe/tree-sitter-plsql at commit 28aebef209be: it is MIT licensed and byte-identical across 353,744 lines except for the documented local include-path change. This is a 0.9.2-rc language enhancement; remaining review focuses on full build and sanitizer coverage, grammar limits, extraction quality, and representative PL/SQL fixtures.

DeusData:
Thank you — and I want to lead with this: **this is the most policy-complete grammar vendoring we have received from outside the project.** That is not a courtesy; it is why this review moved fast, and it is worth saying plainly given how long the PR has waited.

**What we independently verified, rather than took on trust.**

We fetched `AndreasMaierDe/tree-sitter-plsql` at the commit you pinned (`28aebef209be`) and diffed it byte for byte against what this PR vendors:

- `parser.c` — 353,744 lines — differs in **exactly one line**: `#include <tree_sitter/parser.h>` rewritten to `"tree_sitter/parser.h"`. That is precisely the local patch you already declared in `MANIFEST.md`. Nothing else differs anywhere.
- `tree_sitter/parser.h` — byte-identical.
- `LICENSE` — byte-identical (MIT, © 2022 AndreasMaierDe).

The MIT licensing checks out three independent ways, and all four of our vendoring requirements are met: LICENSE vendored with the code, `MANIFEST.md` updated with the pinned commit, the ABI, your include patch *and* the known limitation, `THIRD_PARTY.md` bumped 159→160, and the SBOM needs no edit because it is generated from the vendored directories at release time.

A grep of the entire 354k-line diff for the markers we care about — `http`, `curl`, `wget`, `base64`, `eval`, `system(`, `popen(`, `getenv`, `/etc/`, `.ssh` — returned exactly one hit: the provenance URL in `MANIFEST.md`. Upstream has no `scanner.c` at all, so there is no hand-written vendored lexer to audit. No build or CI file is touched; `grammar_plsql.c` is picked up by the existing wildcard and follows the per-grammar shim pattern exactly.

**The extraction work is equally solid.** The `fnc_name` / `prc_name` / `package_name` field resolvers are correctly scoped to `CBM_LANG_PLSQL`, so there is zero behavioural reach into any other language, and the package-qualified `ref_name_parent.ref_name` callee gives proper `UTIL_PKG.CALC_SALARY` call edges — exactly how our other language hooks are built. You also left `.sql` on generic SQL and `.prc` on FORM and asserted both, which is the right call and one people usually get wrong.

Special credit for the **honest negative test** on the upstream `CREATE TYPE … AS OBJECT` ERROR-node limitation. Capturing the repro fixture so a future grammar upgrade can clear the skip, instead of papering over it, is exactly our style.

**Where it stands.** Our standing rule is that vendored third-party source gets stripped from a contributor's PR and re-vendored by us after an independent audit. In this case the audit is already complete and came back clean, so the maintainer only has to decide *procedure*, not substance — plus one genuine question that is not about your work at all: upstream is a small personal project, dormant since February 2023, so adopting it means we effectively own a 353k-line parser if it goes away. That is his call and I have put it to him directly.

**Two small things you can fix meanwhile:**

1. The PR description is still the empty template. Our policy is that the PR text itself should state the upstream URL, the pinned commit, and the license — right now that provenance lives only in `MANIFEST.md` and the commit messages.
2. The `MANIFEST.md` category `vendor-maintained` is a misnomer here. That class means "the language vendor's own grammar", and AndreasMaierDe is a personal project rather than Oracle. A community/niche class, or a clarifying note, would be more accurate.

Thanks again for the care. Reviewing a 354k-line diff is normally a grim job; yours was not.

DeusData:
**The deep security review is complete and came back clean, so the direction here is accepted: PL/SQL is in, as the 160th vendored grammar.** Thank you for your patience — 353,744 lines of vendored parser from a dormant upstream is exactly the kind of contribution that deserves a slow, sceptical read rather than a fast one, and that is the only reason this took as long as it did.

Since the review's findings are more interesting than a one-word verdict, here is what it actually established.

**Prompt injection: none, and not in the "we looked and didn't find any" sense.** `parser.c` is 9,192,000 bytes containing **zero** bytes above 0x7F and no control characters other than newline. Zero-width, bidi-override, BOM and tag characters are therefore not absent-on-inspection but arithmetically impossible. The file contains **no comments at all** — no `//`, no `/* */`. Its 865 string literals are all distinct, and **not one of them contains a space**: 833 identifier-shaped grammar symbols, 29 punctuation terminals, and 3 accounted-for non-symbols (the include path, a warning pragma, and the empty string in `#pragma optimize("", off)`). English prose needs spaces; there is nowhere for it to hide.

**Reachable code: two functions.** A full C tokenisation covered all 9.19 MB with zero unmatched characters, so the enumeration is exhaustive rather than sampled. The file defines exactly `ts_lex` and `tree_sitter_plsql`, plus 15 canonical `static const` tables. The complete external-symbol surface is 83 identifiers, every one a tree-sitter macro, type, or a `uint16_t`/`uint32_t`/`NULL`. **No libc call of any kind.** No `system`, `exec*`, `fork`, `dlopen`, `getenv`, `fopen`, `socket`, `mmap`, `mprotect`, no inline asm, no function-pointer casts, no `__attribute__((constructor))`. There is **no external scanner** — `EXTERNAL_TOKEN_COUNT 0` and `.external_scanner` is unset — which matters because an external `scanner.c` is where hand-written code hides in a tree-sitter grammar. There is not a single hex or octal literal in the file, so no opaque blob is present either.

**Provenance — I closed this myself rather than leave it as a caveat.** The review could establish safety but not identity, so I fetched upstream at the pinned commit and compared bytes:

- `tree_sitter/parser.h` — git blob `2b14ac1046bb3b45543a62e41d8ee77d2cfae34c`, **byte-identical to upstream**, and independently byte-identical to the five copies already vendored and shipped here (`move`, `lean`, `jinja2`, `fennel`, `wolfram`).
- `parser.c` — applying **only** your declared `#include <tree_sitter/parser.h>` → `"tree_sitter/parser.h"` change to upstream's file reproduces git blob `55f15b314b1f9504220d56f6a7e12423722099e9`, which is exactly the blob in this PR. Byte-for-byte identity, proven by hash rather than by reading.
- `28aebef209be` is upstream's tip commit (last push 2023-02-17), MIT, `Copyright (c) 2022 AndreasMaierDe`, and the LICENSE body is byte-identical to the standard MIT text already vendored here.

So the vendored payload is upstream's, unmodified except for the one change you declared in `MANIFEST.md`. That fully answers the dormant-upstream concern that motivated the review: a generated LR parser has no expressive room for a payload, and this one measurably uses none. One incidental detail I enjoyed — the symbol table carries upstream's own typos (`aggrecate`, `commtted`, `hierachical`, `partiion`), which is a small authenticity signal a tampered table would not reproduce.

**On the grammar choice**, since we ask this of every vendored grammar: I checked for alternatives, and yours is the right pick. `njank/tree-sitter-plsql` and `vinicius-toshiyuki/tree-sitter-plsqloracle` are newer but carry **no license at all**, which disqualifies them outright under our vendoring policy, and `iliasaz/tree-sitter-sql-orcl` has zero stars. `AndreasMaierDe` is both the most-established and the only permissively-licensed real option.

**Repo hygiene is right too:** `MANIFEST.md` updated with upstream, commit, license, ABI 14, the no-external-scanner note and the known `CREATE TYPE ... AS OBJECT` limitation; `THIRD_PARTY.md` 159→160; the SBOM needs no manual edit because `generate-sbom.py` globs grammar directories. Gating is clean — `CBM_LANG_PLSQL` is appended before `CBM_LANG_COUNT` so nothing is renumbered, `lang_specs.c` uses a designated initializer, all extraction hooks are `lang == CBM_LANG_PLSQL`-guarded, and all 11 new extensions were checked against `EXT_TABLE` with zero collisions (`.prc` and `.sql` are untouched). No raw `fopen(`. No build, CI, dependency or install-hook change anywhere.

---

**What is blocking merge is entirely mechanical: four registration gaps, all of them our own tripwires firing correctly.** None is a design problem, and each has one obvious fix.

1. **`tests/repro/repro_language_registry.c:312`** — `invariant=capability_unset`. `LANGUAGE_CAPABILITIES` is deliberately default-free, so a new enum stays `CAP_UNSET` until audited. Given you set `plsql_call_types = {"ref_call"}` with the reference-vocabulary slots empty, the correct entry is **`CALL_WITHOUT_REFERENCE_VOCAB(PLSQL)`** — "call metadata but no reference node recognized by `extract_usages`" describes this exactly. That is a maintainer classification, so treat it as settled rather than a guess on your part.
2. **`tests/repro/repro_language_registry.c:474`** — `remaining_capability_explicit`, same root cause; the entry above resolves it.
3. **`tests/repro/repro_call_node_manifest.c:598`** — `invariant=live_to_table language=163 node=ref_call matches=0`. Needs a manifest row alongside the others, `ENTRY(CBM_LANG_PLSQL, "ref_call", DIRECT_CALL, PRIMARY_EXTRACTOR, true, true)`, plus a bump to `EXPECTED_HISTORICAL_CALL_NODE_TOTAL`.
4. **`security / security-static`** — the vendored-integrity check needs `scripts/security-vendored.sh --update` run and the refreshed hashes committed. Expected for any vendoring change; it is not reporting a problem with your bytes.

**Please rebase before doing this** — your base is from mid-July and the tree has moved a long way. Concretely, `EXPECTED_HISTORICAL_CALL_NODE_TOTAL` is now **219**, not the 189 your CI run was measuring against, so patching to the old number would just fail differently.

One question rather than a request: `extract_usages.c`'s `occurrence_specs[CBM_LANG_COUNT]` has no PL/SQL slot, so PL/SQL gets no occurrence/usage extraction. Plenty of languages are likewise unset, so this may well be deliberate — I would just rather it be a decision than an oversight. If you want occurrences, now is the cheap moment to add them.

Get those four in on a fresh rebase and I will merge it. Genuinely good contribution — thank you for the care in the vendoring paperwork, which made the review far easier than it could have been.

DeusData:
@ouzsrcm — PL/SQL support is on `main` as of 095acc5c. Thank you for bringing it; Oracle PL/SQL is a real gap for anyone indexing enterprise database code, and your extraction mapping is what shipped.

I want to be straight about how it landed. Rather than merge this branch, I re-vendored the grammar directly from upstream and re-applied your language wiring on top, then merged that as #1854. Both commits carry `Co-authored-by: Oğuz <oguzhan.saricam@gmail.com>` — the work is credited to you in the history.

**Why re-vendor rather than merge.** Our policy is that vendored bytes have to be *provable* against upstream, not merely plausible. A grammar copied from a working tree can't be checked that way, so re-vendoring from the pinned upstream commit was the only route that gets there. It is not a judgement about your copy — it is that the proof has to exist for anyone auditing the repo later. Concretely, what we can now assert about `AndreasMaierDe/tree-sitter-plsql` @ `28aebef2`:

- `LICENSE` and `tree_sitter/parser.h` are **byte-identical** to upstream's blobs
- `parser.c` reproduces upstream's git blob SHA exactly once the single documented local patch is reversed (`#include <tree_sitter/parser.h>` → `"..."`), which is a complete proof across 353,744 lines that nothing else was altered
- zero non-ASCII bytes, zero dangerous calls, no external scanner
- the licence audit passes with PL/SQL in the `IDENTICAL` bucket, needing no exemption entry at all

Your design decisions carried over intact: package spec/body → `Class`, procedures and functions → `Function`, file → `Module`, `ref_call` callee extraction including the package-qualified `pkg.name` form, and the extension set (`.pks .pkb .pck .pls .plb .plsql .fnc .trg .bdy .tps .tpb`) with `.sql` and `.prc` deliberately left alone — both now asserted negatively so a later change can't silently steal them.

One thing I added: upstream cannot parse `CREATE TYPE ... AS OBJECT` — it yields ERROR nodes. Rather than pin that absence vacuously, there is now a test asserting *positively* that the fixture parses incompletely, so a future grammar upgrade that fixes it turns the test RED and prompts us to claim the `Class` instead of the limitation silently persisting.

Closing this in favour of #1854. Thank you again — and if you spot anything in the shipped extraction that differs from what you intended for PL/SQL semantics, please say so; you know the language better than I do.

ouzsrcm:
Thanks for the detailed writeup. I went through the diff in #1854 and compared it against my original branch — the extraction logic (Class/Function mapping, ref_call with package-qualified callees) matches my intent, and the negative tests for the extension list are a nice addition I hadn’t thought to add myself.

The positive test for the CREATE TYPE ... AS OBJECT gap is also a good call — better than a silent limitation.

Overall this is in better shape than what I submitted. Thanks for putting in the extra work on provenance.

PR #1039, opened 2026-07-12
feat(ui): add configurable host binding

### LINKED ISSUE #229: Add a --host parameter for UI
Similar to --port we should be able to set the host interface

Blamp26:
## What does this PR do?

Adds a persisted `--host=<IPv4>` option for the embedded Graph UI.

The default remains `127.0.0.1`, preserving the existing localhost-only security posture and backward compatibility with older config files.

Supported examples:

    codebase-memory-mcp --ui=true --host=127.0.0.1 --port=9749
    codebase-memory-mcp --ui=true --host=192.168.88.26 --port=9749
    codebase-memory-mcp --ui=true --host=0.0.0.0 --port=9749

The implementation:

- accepts numeric IPv4 addresses only;
- rejects empty, malformed, and hostname values with exit status 2;
- persists `ui_host` alongside the existing UI settings;
- passes the configured host through the HTTP server/listener APIs;
- preserves port `0` support for tests;
- emits a prominent warning for non-loopback bindings;
- identifies `0.0.0.0` as all interfaces;
- keeps the existing CORS policy unchanged;
- updates documentation and the UI security audit.

Fixes #229

## Security

Non-loopback binding remains an explicit opt-in.

The warning explains that the UI:

- has no authentication;
- is reachable over the network;
- may expose codebase graph data and UI actions;
- should be protected with a firewall or trusted network.

The documentation recommends binding a specific LAN address instead of `0.0.0.0`.

## Validation

- [x] Every commit is signed off
- [x] `make -f Makefile.cbm test` — 6,012 passed, 1 platform-specific skip
- [x] `make -f Makefile.cbm lint-ci CPPCHECK=/opt/cppcheck/bin/cppcheck CLANG_FORMAT=clang-format-20`
- [x] `scripts/security-ui.sh`
- [x] `make -j$(nproc) -f Makefile.cbm cbm`
- [x] `make -j$(nproc) -f Makefile.cbm cbm-with-ui`
- [x] Focused UI/HTTP tests — 64 passed, 1 platform-specific skip
- [x] Default `127.0.0.1` binding tested manually
- [x] `0.0.0.0` binding and warning tested manually
- [x] Specific `192.168.88.26` binding tested manually
- [x] Persisted host reuse tested manually
- [x] Invalid and empty host rejection tested manually

## Portability note

Windows runtime testing was not available locally. The Windows Winsock and `AF_INET` paths compile through the project code, with additional Windows validation left to the repository CI matrix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,10 +114,18 @@ The UI ships as a separate `ui` build (it embeds the frontend). The default inst
 Then run it:
 
 ```bash
-codebase-memory-mcp --ui=true --port=9749
+codebase-memory-mcp --ui=true --host=127.0.0.1 --port=9749
 ```
 
 Open `http://localhost:9749` in your browser. The UI runs as a background thread alongside the MCP server — it's available whenever your agent is connected.
+
+The default host is `127.0.0.1` and is persisted with the UI port. To share the UI on a trusted LAN, bind a specific local address:
+
+```bash
+codebase-memory-mcp --ui=true --host=192.168.88.26 --port=9749
+```
+
+Use `--host=0.0.0.0` only when you need all interfaces. The UI has no authentication; a non-loopback bind is reachable over the network and may expose codebase graph data and UI actions. Protect it with a firewall or trusted network, and prefer a specific LAN IP over `0.0.0.0`.
 
 ### Auto-Index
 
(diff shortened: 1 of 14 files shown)


DeusData:
Thanks for making loopback the default and documenting the unauthenticated exposure risk. This is a security-sensitive 0.9.2-rc backlog enhancement. A non-loopback listener expands the product boundary even when explicitly configured, so merge is gated on a maintainer product-security decision. Any acceptable version must retain 127.0.0.1 as the default, require explicit opt-in, preserve strict CORS, and clearly warn that no authentication exists.

Blamp26:
Thanks for clarifying the product-security gate.

Before expanding this already-green PR, would built-in authentication materially change the product-security decision, or should authentication be handled as a separate design/PR?

I can prepare a minimal design with these constraints:

- `127.0.0.1` remains the default and keeps the current local behavior;
- non-loopback startup requires an explicitly configured authentication token;
- the token is supplied through an environment variable or token file, not persisted in plaintext configuration or exposed in command-line arguments;
- authentication protects every UI, asset, RPC, and API route;
- token comparison is constant-time;
- `Host` and unsafe-request `Origin` values are validated exactly;
- CORS remains allow-list based, never uses `*`, and never reflects arbitrary origins;
- failed authentication attempts are rate-limited;
- documentation still warns that plain HTTP provides no transport encryption and that untrusted networks require a VPN or TLS reverse proxy.

Would you prefer this to remain separate from #1039, or should the current PR be extended after design approval?

DeusData:
Thanks for thinking through the security requirements. We are closing the current PR because the graph UI remains intentionally loopback-only. Built-in authentication would create a separate permanent security surface and does not change the decision on this unauthenticated host-binding PR. Please do not extend this PR. We are not opening a follow-up implementation track for non-loopback access at this time.

itguy327:
I'm working headless with paseo and being able to bind would be very helpful

PR #1040, opened 2026-07-12
fix(cli): print help and require confirmation before destructive uninstall

### LINKED ISSUE #1038: `uninstall --help` performs a real uninstall — unknown flags silently ignored, no confirmation before destructive removal
## Summary

`codebase-memory-mcp uninstall --help` does not print help — it performs a **real, destructive uninstall**: it removes agent configurations (Claude Code, Codex, Gemini, VS Code, Cursor, …), skills, hooks, and deletes the installed binary, then exits 0.

This happened to me on a real machine: I ran `uninstall --help` to look up the flag for selective removal, and it wiped the tool from all five detected agents and deleted the binary.

## Environment

- codebase-memory-mcp 0.9.0 (release binary, UI variant)
- macOS (darwin/arm64)

## Reproduction (safe, uses a throwaway HOME)

```bash
FAKEHOME=$(mktemp -d)
mkdir -p "$FAKEHOME/.local/bin" "$FAKEHOME/.claude"
echo fake-binary > "$FAKEHOME/.local/bin/codebase-memory-mcp"
echo '{"mcpServers":{"codebase-memory-mcp":{"command":"x"}}}' > "$FAKEHOME/.claude/.mcp.json"

HOME="$FAKEHOME" codebase-memory-mcp uninstall --help
```

Output:

```
codebase-memory-mcp uninstall

Claude Code: removed 0 skill(s)
  removed MCP config entry
  removed PreToolUse + SessionStart + SubagentStart hooks
Removed .../.local/bin/codebase-memory-mcp

Uninstall complete.
```

Exit code 0. The fake binary is gone and `.claude/.mcp.json` is emptied — help was never printed.

## Root cause

Three compounding issues:

1. **Subcommand dispatch matches before global `--help`** — [`handle_subcommand` in src/main.c](https://github.com/DeusData/codebase-memory-mcp/blob/8c9ac98e30b48e63d0a4df39f86ae1cd029f5349/src/main.c#L540-L569) scans argv positionally, so in `uninstall --help` the subcommand matches at `i=1` and dispatches before the `--help` check ever sees `i=2`.
2. **`cbm_cmd_uninstall` silently ignores unknown flags** — [src/cli/cli.c#L4087-L4094](https://github.com/DeusData/codebase-memory-mcp/blob/8c9ac98e30b48e63d0a4df39f86ae1cd029f5349/src/cli/cli.c#L4087-L4094) only looks for `-y`/`-n` (via `parse_auto_answer`) and `--dry-run`; anything else — `--help`, `-h`, or a typo like `--dryrun` — falls through to a full uninstall. Note the typo case is arguably worse: `uninstall --dryrun` performs a *real* uninstall when the user explicitly asked for a dry run.
3. **No confirmation gate on the destructive path** — the agent-config removal and binary deletion at [src/cli/cli.c#L4104-L4137](https://github.com/DeusData/codebase-memory-mcp/blob/8c9ac98e30b48e63d0a4df39f86ae1cd029f5349/src/cli/cli.c#L4104-L4137) run unconditionally. The only `prompt_yn` is for index deletion. `prompt_yn` itself fails closed on non-TTY stdin (good), but nothing destructive is behind it — so even `uninstall -n` ("answer no to everything") removes all configs and the binary.

The same unknown-flag blindness affects the other mutating subcommands: `install --help` silently rewrites agent configs, and `update --help` downloads and replaces the binary.

## Expected behavior

- `uninstall --help` / `install --help` / `update --help` print usage and exit 0 without touching anything.
- `uninstall <unknown-flag>` errors out with usage instead of running.
- `uninstall` asks for confirmation (or requires `-y` when non-interactive) before removing configs and the binary; `-n` aborts.

## Proposed fix

I have a patch ready and will open a PR referencing this issue shortly:

- a `--help`/`-h` check at the top of `cbm_cmd_install`, `cbm_cmd_uninstall`, and `cbm_cmd_update` that prints per-command usage and exits 0;
- strict flag validation in `cbm_cmd_uninstall` (unknown flag → error + usage, exit 1);
- a `prompt_yn` confirmation gate in front of the destructive uninstall path (skipped for `--dry-run`; `-y` auto-confirms as before, matching the existing `scripts/smoke-test.sh` phase 6b invocation `uninstall --dry-run -y`);
- regression tests in `tests/test_cli.c`.

metehanulusoy:
Fixes #1038

## Problem

`uninstall --help` performed a **real, destructive uninstall** (removed all agent configs, skills, hooks, and deleted the binary, exit 0). Root causes, as detailed in the issue:

1. `handle_subcommand` in `main.c` matches `uninstall` positionally before the global `--help` check can see it.
2. `cbm_cmd_uninstall` silently ignored unknown flags — `--help`, `-h`, or a typo like `--dryrun` all fell through to a full uninstall.
3. The destructive path (agent-config removal + binary deletion) had no confirmation gate; the only `prompt_yn` guarded index deletion, so even `uninstall -n` removed everything.

## Fix

- `cmd_wants_help()` + per-command usage printers: `install`, `uninstall`, and `update` now print usage and exit 0 on `--help`/`-h` without touching anything.
- `cbm_cmd_uninstall` rejects unknown flags with an error + usage (exit 1).
- `cbm_cmd_uninstall` gates the destructive path behind `prompt_yn`:
  - `-y` auto-confirms — unchanged behavior for scripts;
  - `-n` aborts (previously it still removed configs and the binary);
  - non-TTY stdin without `-y` fails closed via the existing `prompt_yn` error;
  - `--dry-run` skips the prompt (mutates nothing), so the existing `scripts/smoke-test.sh` phase 6b invocation `uninstall --dry-run -y` behaves exactly as before.

No wrapper or script in the repo invokes bare `uninstall` non-interactively (checked `Formula/`, `pkg/`, `scripts/`, `install.sh`, `install.ps1`), so the new prompt breaks no automated path.

## Behavior after the fix

```
$ codebase-memory-mcp uninstall --help
Usage: codebase-memory-mcp uninstall [-y|-n] [--dry-run]
...                                          # exit 0, nothing touched

$ codebase-memory-mcp uninstall --dryrun
error: unknown uninstall option: --dryrun    # exit 1, nothing touched

$ codebase-memory-mcp uninstall </dev/null
error: interactive prompt requires a terminal. Use -y or -n flags.
Uninstall cancelled. No files were modified. # exit 1

$ codebase-memory-mcp uninstall -y
...
Uninstall complete.                          # unchanged
```

## Tests

Three regression tests in `tests/test_cli.c` (run under a temp `HOME`, using the existing `cbm_set_auto_answer_for_test` seam):

- `cli_uninstall_help_not_destructive` — `--help` exits 0 and leaves the fake install untouched
- `cli_uninstall_unknown_flag_rejected` — `--dryrun` typo exits 1 and leaves files untouched
- `cli_uninstall_confirmation_gate` — declined prompt aborts; `-y` proceeds and removes the binary

`scripts/test.sh` (ASan+UBSan) passes locally on macOS arm64. `clang-format --dry-run` reports zero warnings on both touched files; cppcheck is clean. clang-tidy was not available locally — relying on CI for that layer.

### DIFF
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -2964,6 +2964,49 @@ static bool prompt_yn(const char *question) {
     return (buf[0] == 'y' || buf[0] == 'Y') ? true : false;
 }
 
+/* ── Subcommand help ──────────────────────────────────────────── */
+
+/* Mutating subcommands (install/uninstall/update) must never run when the
+ * user asked for usage: `uninstall --help` used to perform a real uninstall
+ * because unknown flags were silently ignored (#1038). */
+static bool cmd_wants_help(int argc, char **argv) {
+    for (int i = 0; i < argc; i++) {
+        if (strcmp(argv[i], "--help") == 0 || strcmp(argv[i], "-h") == 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
+static void print_install_usage(void) {
+    printf("Usage: codebase-memory-mcp install [-y|-n] [--force] [--dry-run]\n");
+    printf("                                   [--plan] [--reset-indexes]\n\n");
+    printf("  -y, --yes        Answer yes to all prompts\n");
+    printf("  -n, --no         Answer no to all prompts\n");
+    printf("  --force          Overwrite existing skills and config entries\n");
+    printf("  --dry-run        Show what would change without modifying files\n");
+    printf("  --plan           Print machine-readable install plan (no changes)\n");
+    printf("  --reset-indexes  Delete existing indexes during install\n");
+}
+
+static void print_uninstall_usage(void) {
+    printf("Usage: codebase-memory-mcp uninstall [-y|-n] [--dry-run]\n\n");
+    printf("  -y, --yes   Answer yes to all prompts (required without a terminal)\n");
+    printf("  -n, --no    Answer no to all prompts (aborts the uninstall)\n");
+    printf("  --dry-run   Show what would be removed without modifying files\n");
+}
+
+static void print_update_usage(void) {
+    printf("Usage: codebase-memory-mcp update [-y|-n] [--force] [--dry-run]\n");
+    printf("                                  [--standard|--ui]\n\n");
+    printf("  -y, --yes     Answer yes to all prompts\n");
+    printf("  -n, --no      Answer no to all prompts\n");
+    printf("  --force       Update even if already on the latest version\n");
+    printf("  --dry-run     Show what would change without modifying files\n");
+    printf("  --standard    Select the standard variant\n");
+    printf("  --ui          Select the UI variant\n");
+}
+
 /* ── SHA-256 checksum verification ─────────────────────────────── */
 
 /* SHA-256 hex digest: 64 hex chars + NUL */
@@ -3748,6 +3791,10 @@ char *cbm_build_install_plan_json(const char *home, const char *binary_path) {
 }
 
 int cbm_cmd_install(int argc, char **argv) {
+    if (cmd_wants_help(argc, argv)) {
+        print_install_usage();
+        return 0;
+    }
     parse_auto_answer(argc, argv);
     bool dry_run = false;
     bool force = false;
@@ -4085,11 +4132,22 @@ static void uninstall_editor_agents(const cbm_detected_agents_t *agents, const c
 }
 
 int cbm_cmd_uninstall(int argc, char **argv) {
+    if (cmd_wants_help(argc, argv)) {
+        print_uninstall_usage();
+        return 0;
+    }
     parse_auto_answer(argc, argv);
     bool dry_run = false;
     for (int i = 0; i < argc; i++) {
         if (strcmp(argv[i], "--dry-run") == 0) {
             dry_run = true;
+        } else if (strcmp(argv[i], "-y") != 0 && strcmp(argv[i], "--yes") != 0 &&
+                   strcmp(argv[i], "-n") != 0 && strcmp(argv[i], "--no") != 0) {
+            /* Unknown flags must not fall through to a destructive uninstall:
+             * `uninstall --dryrun` (typo) used to perform a REAL uninstall (#1038). */
+            (void)fprintf(stderr, "error: unknown uninstall option: %s\n\n", argv[i]);
+            print_uninstall_usage();
+            return CLI_TRUE;
         }
     }
 
@@ -4100,6 +4158,15 @@ int cbm_cmd_uninstall(int argc, char **argv) {
     }
 
     printf("codebase-memory-mcp uninstall\n\n");
+
+    /* Removing agent configs and the binary is destructive: gate it on
+     * explicit confirmation. Non-interactive callers must pass -y;
+     * prompt_yn() fails closed without a terminal (#1038). */
+    if (!dry_run &&
+        !prompt_yn("Remove codebase-memory-mcp from all detected agents and delete the binary?")) {
+        printf("Uninstall cancelled. No files were modified.\n");
+        return CLI_TRUE;
+    }
 
     cbm_detected_agents_t agents = cbm_detect_agents(home);
     if (agents.claude_code) {
@@ -4382,6 +4449,10 @@ static bool check_already_latest(void) {
 }
 
 int cbm_cmd_update(int argc, char **argv) {
+    if (cmd_wants_help(argc, argv)) {
+        print_update_usage();
+        return 0;
+    }
     parse_auto_answer(argc, argv);
 
     bool dry_run = false;
--- a/tests/test_cli.c
+++ b/tests/test_cli.c
@@ -519,6 +519,137 @@ TEST(cli_uninstall_removes_skills) {
     PASS();
 }
 
+/* ── uninstall command safety (#1038) ─────────────────────────── */
+
+/* Test seam from cli.c: force prompt_yn() auto-answer (1=yes, -1=no, 0=prompt). */
+extern void cbm_set_auto_answer_for_test(int value);
+
+static bool test_path_exists(const char *path) {
+    struct stat st;
+    return stat(path, &st) == 0;
+}
+
+/* Create a fake install under home: binary + Claude MCP config.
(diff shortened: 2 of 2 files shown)


DeusData:
Thanks for addressing the destructive uninstall behavior with help, validation, confirmation, and regression tests. This is a high-priority 0.9.1-rc safety fix. The current Ubuntu, macOS, and Windows smoke checks are failing, so those failures need to be resolved before the review can complete; install, update, noninteractive, and dry-run behavior must remain fail-closed.

DeusData:
Good news — the fix itself is correct and complete: `-y|--yes` for non-interactive confirmation, and it fail-closes (auto-"n" → cancel) without a TTY, which is exactly the safety behavior this needs.

The CI red is purely branch staleness, not a bug in your change. I traced the smoke failure: Phase 9 (agent-config uninstall E2E) on your branch calls bare `uninstall`, so your new confirmation code sees no `-y` and no terminal → auto-cancels → the E2E assertion that expects an actual uninstall fails. Current `main` already updated those smoke call sites to `uninstall -y` (lines ~1338/1357/1358 in scripts/smoke-test.sh), so a **rebase onto current main resolves all three legs** — no code change needed from you.

Could you rebase onto current main and push? Once it's green I'll re-diff and merge — this is a high-priority safety fix (`uninstall --help` wiping everything is exactly the kind of thing that shouldn't sit in an rc), so I'd like to land it quickly. Thanks for the thorough fix + regression tests.

DeusData:
Thank you for this — the root-cause writeup is excellent and the bug is real. I need to be straight with you about one thing first, because I have been telling a lot of contributors tonight that their red CI was not their fault: **yours is.** It is a genuine, deterministic failure, and it points at a real design gap rather than a typo.

**Your diagnosis was right on all three counts**, and I verified each against current main:

- The dispatcher matches the subcommand *before* the global `--help` scan (`main.c:928-936`), so `uninstall --help` really did perform a real uninstall.
- Unknown flags fell through silently, so a typo like `--dryrun` wiped the machine.
- The destructive path had no confirmation at all.

Main has since shipped the unknown-flag rejection independently for all three commands, which is really just confirmation that your analysis was correct.

**The blocker: our smoke suite drives `uninstall -y -n`, and your gate reads that as "no".**

Smoke Phase 9 (and 9b-3, 9b-8) call `uninstall -y -n`, which in our idiom means *"yes, uninstall — no, keep the indexes"*. But `parse_auto_answer` is **last-flag-wins global state**, so `-y -n` resolves to auto-NO. Your new gate consumes that global answer, prints "n (auto)", and cancels the whole uninstall — producing `FAIL 9a-b` on all three platforms. That is exactly the three red `pr-smoke` legs plus the `ci-ok` aggregate. Everything else — the full 3-OS test matrix, TSan, lint, CodeQL, the security gates, DCO — is green, which is genuinely respectable for a PR this size.

Your audit checked that no wrapper invokes *bare* uninstall non-interactively, and that reasoning was sound as far as it went. What it could not have caught is that `-y -n` carries two different meanings for two different prompts.

**And that is the actual design point:** once `uninstall` has **two prompts with different scopes**, a single global yes/no channel cannot express both. This is not a rebase-fixable issue. The gate needs a scoped answer — an explicit `--keep-indexes` / `--delete-indexes` pair alongside `-y`, or a per-prompt answer mechanism — rather than sharing `prompt_yn`'s global state.

**One more thing worth fixing while you are in there: a Windows hang.** `prompt_yn`'s non-TTY fail-close is guarded by `#ifndef _WIN32`. On POSIX a closed stdin aborts cleanly, which is correct. But on Windows, a non-interactive caller holding stdin *open but silent* — a service wrapper, some CI shells — blocks in `fgets` **forever**. That property already existed, but your change extends its blast radius from a rare index-deletion prompt to **every bare `uninstall`**. Adding `_isatty(_fileno(stdin))` on the Windows side closes it.

**Your tests are the right shape** — temp-HOME isolation, the existing auto-answer seam, and asserting on-disk state rather than output. `cli_uninstall_confirmation_gate` covers ground main still does not. Two of the three have partially aged since main shipped flag rejection, so they will need revisiting on rebase.

**Where this stands.** #1038 is still open, and the maintainer's own comment there extended the acceptance criteria beyond what any single PR here covers. The pieces still genuinely wanted are: the friendly `--help` usage with exit 0, the confirmation gate with a scoped-answer design, and complete owned-hook-file cleanup. The uninstall function itself was substantially rewritten by the daemon and activation-transaction work, so this needs a from-scratch rebase rather than a conflict resolution.

If you would rather not carry that, say so and we can distill it with `Co-Authored-By` credit to you — the analysis is the valuable part and it is yours either way.

Small note: `print_update_usage` describes `update` as a variant-selecting self-updater, which is no longer true on main — it hands off to the install script now.

DeusData:
Closing this as superseded — **both halves are already on `main`**, and I verified each in the source rather than assuming.

**`uninstall --help` no longer uninstalls.** `cbm_cmd_uninstall` now checks for `--help`/`-h` and prints usage before doing anything, returning immediately. The help text itself warns "THIS IS DESTRUCTIVE" and points at `--dry-run`.

**The destructive path is confirmation-gated**, via `prompt_yn("Delete these indexes?")`.

What's worth showing you is *where* the help check sits, because it's the detail your PR and the upstream fix diverged on. The check runs **before** `parse_auto_answer`, with this comment beside it:

> Checked before `parse_auto_answer` so a `-y` sitting elsewhere on the line cannot auto-confirm the destruction we are trying to prevent.

That ordering is what avoids the regression the July review found in this PR: a single global confirmation state resolved `uninstall -y -n` — which our own smoke tests use, with two distinct prompts meaning "yes uninstall" and "no, keep indexes" — into an auto-NO that cancelled the whole uninstall on CI. Same goal, and the sequencing is what made the difference.

So there's nothing left to merge here, and it isn't a reflection on the work: you found a genuinely dangerous bug (#1038 — a `--help` flag that destroyed an installation is about as bad as CLI bugs get), and your three tests were the right shape, asserting on-disk state through a temp-HOME seam rather than trusting output text.

Apologies that our review latency meant it got fixed upstream without your patch landing. The report is what mattered most, and that part was yours.

PR #1043, opened 2026-07-12
fix(mcp): return structured compact results

WarGloom:
## What does this PR do?

Every MCP tool advertises an object `outputSchema`, but compact TOON and
plain-text results omitted `structuredContent`. Strict MCP clients therefore
rejected otherwise successful tool calls with `-32600`.

This change keeps JSON-object results unchanged and adds an object for every
other result:

- successful text and TOON payloads use `{"text":"..."}`
- error text uses `{"error":"..."}`

The existing text `content` block remains for model visibility and backwards
compatibility.

## Validation

- `ASAN_OPTIONS=detect_leaks=0 make -f Makefile.cbm test` — 6,009 passed,
  1 skipped
- `make -f Makefile.cbm lint-ci`
- wire-level coverage for compact `search_graph` responses plus unit coverage
  for plain-text and error results

## Checklist

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -258,16 +258,29 @@ char *cbm_mcp_text_result(const char *text, bool is_error) {
     yyjson_mut_arr_add_val(content, item);
     yyjson_mut_obj_add_val(doc, root, "content", content);
 
-    if (!is_error && text) {
+    bool has_structured_content = false;
+    if (text) {
         yyjson_doc *structured_doc = yyjson_read(text, strlen(text), 0);
         if (structured_doc) {
             yyjson_val *structured_root = yyjson_doc_get_root(structured_doc);
             if (yyjson_is_obj(structured_root)) {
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the compact structured-result fix. This is a normal-priority 0.9.1-rc MCP compatibility bug. Review will verify the JSON-RPC structuredContent contract, content fallback behavior for older clients, error responses, and that compact mode does not silently change existing tool payload semantics.

DeusData:
Merged. This closes a real protocol-compliance gap in the TOON compact-output work (#1013): every MCP tool advertises an object `outputSchema`, but the compact-TOON and plain-text result paths omitted `structuredContent`, so strict clients rejected otherwise-successful calls with -32600. Verified: the three assertions are genuine guards (RED without your change — `structuredContent` absent — GREEN with it), the value is a proper `{"text": ...}` object (no double-encoding), JSON-object results are unchanged, mcp suite 165/165, lint clean. Thank you — good catch on the strict-client rejection.

Issue #704, opened 2026-06-29
query_graph / search_graph return "project not found" inconsistently; DBs valid on disk but CLI looks up wrong inode


MatiasNAmendola:
# `query_graph` / `search_graph` return "project not found or not indexed" even with a registered, indexed project

## Environment

- **Tool**: `codebase-memory-mcp` v0.8.1 (binary download from release, macOS aarch64)
- **OS**: macOS 15 (Darwin), APFS filesystem (default case-insensitive, case-preserving)
- **Storage dir**: `~/.cache/codebase-memory-mcp/`
- **Project indexed**: MSBA-MSBA (3034 `.cs` files, 82,303 nodes / 269,108 edges per `list_projects`)

## Repro

1. Install cbm 0.8.1 per the official script (single static binary).
2. Index a repo: `codebase-memory-mcp cli index_repository '{"repo_path":"/path/to/MSBA-MSBA"}'` — succeeds, list_projects shows it.
3. **Within the same shell session** (or the same batch job that called index_repository), run:
   ```bash
   codebase-memory-mcp cli search_graph '{"name_pattern":"DTO$","label":"Class","limit":100,"project":"Users-...-MSBA-MSBA"}'
   ```
   → returns 35 results, exit=0.
4. Wait 10 minutes (or close/reopen the shell).
5. Run the **exact same** command again → `{"error":"project not found or not indexed", "available_projects":[...the same 4 entries...]}`

## What's odd

- `list_projects` keeps showing the project (full `nodes`/`edges`/`size_bytes` metadata present).
- `_config.db` is **empty** (`SELECT * FROM config` → 0 rows in `~/.cache/codebase-memory-mcp/_config.db`).
- The per-project SQLite DB (`Users-...-MSBA-MSBA.db`, ~282 MB) is still on disk, populated, the same content as before — `file` reports it as a valid `SQLite 3.x database`.
- Trying `codebase-memory-mcp config set project "<name>"` and `config reset project` does not change anything.
- Tried 8 variations of the `project` parameter naming (`project`, `project_name`, `project_id`, `projectName`, `projectPath`, `root_path`) — every form is rejected with the same error.

## Smell-test observations from a CI run on 2026-06-29

[...]

PR #1046, opened 2026-07-12
fix(mcp): ignore ::missed shadow rows when resolving a db's project name

### LINKED ISSUE #1044: Projects with a `::missed` shadow row vanish from list_projects and the graph UI
## Summary

Any project whose db carries a `<name>::missed` shadow row (created by the miss-graph feature for projects with at least one parse/index failure — i.e. most real-world repos) **disappears from `list_projects` and the graph UI**, even though the index is complete and directly-addressed tools (`search_graph`, `index_status`, `get_graph_schema`) keep working.

Observed on a real repo: a 704-file ASP.NET codebase indexed fine (70,410 nodes / 308,106 edges, `index_status` → `ready`), but `list_projects` returned `{"projects":[]}` and the UI showed "No indexed projects".

## Environment

- main @ 8c9ac98 (locally built), macOS darwin/arm64
- Also affects any release that contains both #704 and the miss-graph shadow project

## Root cause

Two features conflict:

1. The miss-graph pass inserts a **second row** into the per-db `projects` table — `<project>::missed` — because `nodes.project` has an FK to `projects(name)`. The comment there still claims this row is *"Invisible to list_projects, which scans the cache directory for .db files, not this table"* ([src/store/store.c#L1983-L1986](https://github.com/DeusData/codebase-memory-mcp/blob/8c9ac98e30b48e63d0a4df39f86ae1cd029f5349/src/store/store.c#L1983-L1986)) — that claim is stale since #704.
2. Since #704, `list_projects` (via `build_project_json_entry`) and `resolve_store_fallback_scan` key on the db's INTERNAL project name via `db_internal_project_name`, which requires the `projects` table to contain **exactly one row**: `n == 1` at [src/mcp/mcp.c#L1338](https://github.com/DeusData/codebase-memory-mcp/blob/8c9ac98e30b48e63d0a4df39f86ae1cd029f5349/src/mcp/mcp.c#L1338).

Once the shadow row lands, `n == 2` → `db_internal_project_name` returns false → the db is treated as "ghost / unreadable" and skipped from the listing, and the fallback-scan resolve path fails as well.

## Reproduction

```bash
# index any repo that produces >= 1 parse failure (or add the row manually):
sqlite3 ~/.cache/codebase-memory-mcp/<project>.db \
  "INSERT INTO projects(name, indexed_at, root_path) VALUES ('<project>::missed','2026-01-01T00:00:00Z','');"
codebase-memory-mcp cli list_projects
# → {"projects":[], "hint":"No projects indexed. ..."}  — while the db has 70k nodes
```

Deleting the `::missed` row makes the project reappear immediately.

## Expected behavior

Internal shadow projects (`*::missed`) are ignored when deriving the db's internal name; the primary project stays listed and resolvable.

## Proposed fix

In `db_internal_project_name`, filter out rows whose name contains `::` and require exactly one **primary** row, instead of `n == 1` over all rows. I have a patch with a reproduce-first regression test (modeled on the #704 test fixture) and will open a PR referencing this issue shortly.

metehanulusoy:
Fixes #1044

## Problem

Any project whose db carries a `<name>::missed` shadow row — created by the miss-graph pass for projects with at least one parse/index failure, i.e. most real-world repos — **vanished from `list_projects` and the graph UI**, and the fallback-scan resolve path failed. Directly-addressed tools kept working, which made the failure look like "nothing is indexed" while the index was fully present.

Root cause (details in the issue): since #704, `build_project_json_entry` and `resolve_store_fallback_scan` derive the db's internal name via `db_internal_project_name`, which required the per-db `projects` table to hold **exactly one** row (`n == 1`). The miss-graph pass adds a second row (`<name>::missed`, needed for the FK on `nodes.project`), so `n == 2` and the whole db was skipped as "ghost / unreadable". The comment in `store.c` still claims the shadow row is "invisible to list_projects" — that stopped being true when #704 made listing key on this table.

## Fix

`db_internal_project_name` now ignores internal shadow rows (names containing `::`) and requires exactly one **primary** row. Shadow rows keep satisfying the FK, never surface in listings, and no longer make the primary project unresolvable.

## Test

`tool_list_projects_ignores_missed_shadow_issue1044` in `tests/test_mcp.c`, modeled on the #704 reproduce-first fixture (isolated `CBM_CACHE_DIR`, db built via `issue704_make_db`, shadow row added exactly the way the miss-graph pass does):

- A: `list_projects` still advertises the primary project while the shadow row exists — **RED before this fix** (verified: `150 passed, 1 failed` on unpatched code), GREEN after (`151 passed`);
- B: the shadow name is never advertised;
- C: `search_graph` on the project still resolves and returns its node.

`scripts/test.sh` (ASan+UBSan) passes locally on macOS arm64. `clang-format --dry-run` reports zero warnings on the changed hunks (the 3 pre-existing warnings at `tests/test_mcp.c:967-968` are untouched, unrelated lines); cppcheck clean. clang-tidy was not available locally — relying on CI for that layer.

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -1335,10 +1335,23 @@ static bool db_internal_project_name(const char *full_path, char *name_out, size
     cbm_project_t *projs = NULL;
     int n = 0;
     bool ok = false;
-    if (cbm_store_list_projects(st, &projs, &n) == CBM_STORE_OK && n == 1 && projs[0].name &&
-        projs[0].name[0]) {
-        snprintf(name_out, name_sz, "%s", projs[0].name);
-        ok = true;
+    if (cbm_store_list_projects(st, &projs, &n) == CBM_STORE_OK) {
+        /* Ignore internal shadow projects ("<name>::missed" miss-graph rows):
+         * they share the db with the primary project and must not make it
+         * unresolvable — requiring n == 1 over ALL rows made every project
+         * with a miss graph vanish from list_projects and the UI (#1044). */
+        int primary = -1;
+        int primary_count = 0;
+        for (int i = 0; i < n; i++) {
+            if (projs[i].name && projs[i].name[0] && !strstr(projs[i].name, "::")) {
+                primary = i;
+                primary_count++;
+            }
+        }
+        if (primary_count == 1) {
+            snprintf(name_out, name_sz, "%s", projs[primary].name);
+            ok = true;
+        }
     }
     cbm_store_free_projects(projs, n);
     if (ok && out_store) {
--- a/tests/test_mcp.c
+++ b/tests/test_mcp.c
@@ -4134,6 +4134,81 @@ TEST(tool_resolve_store_by_internal_name_issue704) {
     PASS();
 }
 
+/* ── #1044: a "<name>::missed" shadow row must not hide the project ──
+ *
+ * The miss-graph pass inserts a second `projects` row ("<name>::missed") so
+ * its nodes satisfy the FK on nodes.project. db_internal_project_name
+ * required the projects table to hold EXACTLY ONE row, so any project with
+ * a miss graph vanished from list_projects and the graph UI, and the
+ * fallback-scan resolve path failed.
+ *
+ * RED on buggy code / GREEN on the fix:
+ *   A. list_projects still advertises "delta1044" while the shadow row exists.
+ *   B. the shadow name itself is never advertised.
+ *   C. search_graph(project="delta1044") still resolves and returns the node.
+ */
+TEST(tool_list_projects_ignores_missed_shadow_issue1044) {
+    char cache[256];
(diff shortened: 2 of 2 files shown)


tmonestudio:
Independent large-cache validation confirms this root cause: 500 valid project databases were present, 334 carried a derived `::missed` row, and the unpatched listing returned exactly 166 projects. Filtering the shadow restored all 500 primary projects without changing the databases. An equivalent local fix passed the MCP suite (145 passed, 6 platform-specific skips). Thanks for the focused fix..

DeusData:
Thanks for the focused fix for #1044. This is a high-priority 0.9.1-rc visibility bug. Review will verify exact base-project selection when one or more missed shadow rows exist, deterministic behavior regardless of row order, and unchanged results for databases without shadow rows.

PR #1060, opened 2026-07-12
fix(extract): populate body_tokens before MinHash gate; add OS/TS identifier node types

isc-tdyar:
**`body_tokens` was not populated for short functions; type-identifier node kinds were missing**

Two independent issues in `extract_body_ident_tokens` / `compute_fingerprint` that reduce semantic search quality:

### 1. `body_tokens` gated behind MinHash — short functions get no tokens

`def->body_tokens` was assigned *after* the `cbm_minhash_compute` early-return guard in `compute_fingerprint`. Any function below the MinHash length threshold (common for short methods, property accessors, constructors) had `body_tokens = NULL` even when the function body contained meaningful type identifiers. This degrades semantic search for the most common class of methods.

**Fix**: move `def->body_tokens = extract_body_ident_tokens(ctx, body)` to *before* the MinHash gate. Body tokens are now always populated when a body node exists, regardless of function length.

### 2. Type-identifier node kinds excluded from body-token extraction

The inclusion check in `extract_body_ident_tokens` matched `identifier` and `property_identifier` but not TypeScript/ObjectScript type annotation nodes. This meant type names in signatures and return types (e.g. `SerializedResult`, `MyModel`) were silently dropped from `body_tokens`, reducing semantic signal for typed languages.

**Fix**: add the following node kinds to the inclusion check:
- `type_identifier` (TypeScript, C, C++, Go, …)
- `class_name`, `method_name`, `routine_name` (ObjectScript UDL/routine)
- `objectscript_identifier_special`, `identifier_segment_immediate_special`, `quote_permitting_identifier` (ObjectScript)

### 3. Raise body-token buffer limits

`BT_MAX_IDENTS` (40→128), `BT_BUF` (512→2048), `BT_SEEN` (128→256), `BT_SEEN_MASK` (127→255) — the old limits silently truncated token lists for large function bodies.

### Regression test

`body_tokens_type_identifier` in `tests/test_extraction.c` — indexes a TypeScript function with `MyModel` and `SerializedResult` type annotations and asserts both appear in `body_tokens`.

### Impact

Observed on a 1,162-class ObjectScript codebase: before this fix, semantic similarity scores for structurally similar classes were negative or near-zero (`body_tokens` NULL for most methods). After: scores in the 0.86–0.87 range for known-similar class pairs.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -82,7 +82,7 @@ static bool try_append_ident(const char *source, uint32_t s, int len, uint32_t *
 /* Walk AST body, collect unique identifier text as space-separated string.
  * Returns arena-allocated string or NULL. */
 static char *extract_body_ident_tokens(CBMExtractCtx *ctx, TSNode body) {
-    enum { BT_STACK = 512, BT_BUF = 512, BT_MAX_IDENTS = 40, BT_SEEN = 128, BT_SEEN_MASK = 127 };
+    enum { BT_STACK = 512, BT_BUF = 2048, BT_MAX_IDENTS = 128, BT_SEEN = 256, BT_SEEN_MASK = 255 };
     TSNode bt_stack[BT_STACK];
     int bt_top = 0;
     bt_stack[bt_top++] = body;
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the focused extractor fix and test. This is a normal-priority 0.9.1-rc parsing PR. Review will verify that short functions now retain body tokens before the MinHash gate, the larger token limits remain bounded, and the additional ObjectScript and TypeScript identifier node kinds do not introduce duplicate or noisy tokens.

DeusData:
Merged (2080d82). Verified: `body_tokens_type_identifier` is a genuine guard — it fails RED without the fix (short functions whose `body_tokens` were assigned *after* the MinHash early-return got NULL tokens), GREEN with it. Both fixes are sound: moving the `body_tokens` assignment ahead of the `cbm_minhash_compute` gate so short functions still contribute to semantic search, and adding the missing OS/TS type-identifier node kinds. Small, well-scoped, lint clean. Thanks — improves semantic-search recall on short functions across the board, not just ObjectScript.

isc-tdyar:
Thanks!!

PR #1063, opened 2026-07-12
feat(pipeline): Ensemble production topology routing pass for ObjectScript/IRIS

### LINKED ISSUE #1061: Ensemble production topology routing pass for ObjectScript/IRIS
Following up on PR #467 (ObjectScript language support) — @DeusData asked in review to split the Ensemble routing pass into a separate PR. This issue tracks that work.

## What it does

`pass_ensemble_routing` is a pipeline pass that statically extracts **InterSystems Ensemble/IRIS Interoperability** production topology from ObjectScript `.cls` files at index time — no live IRIS instance required.

**Pass A — EnsembleItem nodes**: reads each `EnsembleProduction` class's `ProductionDefinition` XData block (XML), parses `<Item Name="..." ClassName="...">` entries, and creates `EnsembleItem` nodes in the graph for each production component (business services, processes, operations).

**Pass B — ROUTES_TO edges**: for each `EnsembleItem`, reads the source `.cls` file and extracts routing rules that reference other components by class name, emitting `ROUTES_TO` edges. Uses segment-anchored matching (`qn_ends_with_segment`) to avoid false positives.

**WorkMgr dispatch**: detects `.Queue("##class(X).method", ...)` calls and emits `CALLS` edges to the target method — the same pattern as task queues in Celery/Airflow.

## Why it's useful beyond IRIS

The pattern — statically extracting a runtime workflow topology from configuration XML embedded in source classes — is the same shape as:
- Celery task routing (`@app.task`, `task.apply_async(queue=...)`)
- Airflow DAG dependencies
- Spring Integration channel adapters

The `EnsembleItem` label and `ROUTES_TO` edge are the IRIS-specific names, but the pass structure is general enough to serve as a template for other workflow frameworks.

## Changes from the version reviewed in #467

Per review feedback:
- Pass is now gated: early-exits if the project contains no ObjectScript files (no unconditional O(all-method-files) scan for non-IRIS projects)
- Truncation logging added: `cbm_log_warn` fires when `MAX_ITEMS` or `MAX_SETTINGS` caps are hit so silent truncation is visible

## Ready to file PR on your signal

isc-tdyar:
Adds `pass_ensemble_routing` — a pipeline pass that statically extracts InterSystems Ensemble/IRIS Interoperability production topology from ObjectScript `.cls` files at index time. No live IRIS instance required.

Closes #1061. Follows up on #467 (ObjectScript language support) — split out at reviewer request.

## What it does

**Pass A — EnsembleItem nodes**: reads each `EnsembleProduction` class's `ProductionDefinition` XData block (XML), parses `<Item Name="..." ClassName="...">` entries, and creates `EnsembleItem` nodes for each production component (business services, processes, operations).

**Pass B — ROUTES_TO edges**: for each `EnsembleItem`, reads the source `.cls` file and extracts routing rules referencing other components by class name, emitting `ROUTES_TO` edges. Uses segment-anchored matching (`qn_ends_with_segment`) to avoid false positives from partial class name matches.

**WorkMgr dispatch**: detects `.Queue("##class(X).method", ...)` calls and emits `CALLS` edges to the target method — the same static dispatch pattern as task queues in Celery/Airflow.

## Changes from the version reviewed in #467

Per review feedback:
- **Language gate**: early-exits immediately if the project contains no ObjectScript files — zero overhead for non-IRIS projects
- **Truncation logging**: `cbm_log_warn` fires when `MAX_ITEMS` or `MAX_SETTINGS` caps are hit so silent truncation is visible in logs

## Files changed

- `src/pipeline/pass_ensemble_routing.{c,h}` — new pass
- `src/pipeline/pipeline.c` — wired into `run_predump_passes`
- `Makefile.cbm` — added to SRCS

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -287,6 +287,7 @@ PIPELINE_SRCS = \
     src/pipeline/pass_complexity.c \
     src/pipeline/pass_cross_repo.c \
     src/pipeline/artifact.c \
+    src/pipeline/pass_ensemble_routing.c \
     src/pipeline/pass_pkgmap.c
 
 # SimHash / MinHash module
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -356,6 +356,7 @@ PIPELINE_SRCS = \
     src/pipeline/pass_complexity.c \
     src/pipeline/pass_cross_repo.c \
     src/pipeline/artifact.c \
+    src/pipeline/pass_ensemble_routing.c \
     src/pipeline/pass_pkgmap.c
 
 # SimHash / MinHash module
--- a/src/pipeline/pass_ensemble_routing.c
+++ b/src/pipeline/pass_ensemble_routing.c
@@ -176,7 +176,7 @@ static ens_prod_def_t *parse_production_xml(const char *xml, const char *class_q
 }
 
 static char *read_file(const char *full_path) {
-    FILE *f = fopen(full_path, "rb");
+    FILE *f = cbm_fopen(full_path, "rb");
     if (!f)
         return NULL;
     fseek(f, 0, SEEK_END);
@@ -234,7 +234,7 @@ static int64_t find_entry_point(cbm_pipeline_ctx_t *ctx, const char *class_name)
         cbm_gbuf_find_by_name(ctx->gbuf, ENTRY_POINTS[ei], (const cbm_gbuf_node_t ***)&nodes,
                               &count);
         for (int ni = 0; ni < count; ni++) {
-            if (nodes[ni]->qualified_name && strstr(nodes[ni]->qualified_name, suffix))
+            if (nodes[ni]->qualified_name && strcmp(nodes[ni]->qualified_name, suffix) == 0)
                 return nodes[ni]->id;
         }
     }
@@ -422,6 +422,22 @@ static void collect_prod_defs(cbm_pipeline_ctx_t *ctx, ens_prod_def_t ***defs_ou
     *count_out = n;
 }
 
+/* True when haystack equals needle exactly, or ends with ".needle"
+ * (segment-anchored). Prevents "Ens" from matching "Ens.BusinessService". */
+static bool class_name_matches(const char *haystack, const char *needle) {
+    if (!haystack || !needle)
+        return false;
+    size_t hlen = strlen(haystack);
+    size_t nlen = strlen(needle);
+    if (nlen == 0)
+        return false;
+    if (hlen == nlen)
+        return strcmp(haystack, needle) == 0;
+    if (hlen > nlen && haystack[hlen - nlen - 1] == '.')
+        return strcmp(haystack + hlen - nlen, needle) == 0;
+    return false;
+}
+
 static bool method_belongs_to_production(const cbm_gbuf_node_t *method, const ens_prod_def_t *def) {
     if (!method->properties_json)
         return false;
@@ -429,7 +445,7 @@ static bool method_belongs_to_production(const cbm_gbuf_node_t *method, const en
     if (!jstr(method->properties_json, "parent_class", parent_class, sizeof(parent_class)))
         return false;
     for (int i = 0; i < def->n_items; i++) {
-        if (strstr(parent_class, def->items[i].class_name))
+        if (class_name_matches(parent_class, def->items[i].class_name))
             return true;
     }
     return false;
--- a/src/pipeline/pass_ensemble_routing.c
+++ b/src/pipeline/pass_ensemble_routing.c
@@ -1,0 +1,535 @@
+#include "pipeline/pass_ensemble_routing.h"
+#include "pipeline/pipeline_internal.h"
+#include "graph_buffer/graph_buffer.h"
+#include "foundation/log.h"
+#include "foundation/compat.h"
+#include "foundation/compat_fs.h"
+#include "foundation/constants.h"
+#include "foundation/str_util.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdbool.h>
+#include <ctype.h>
+
+#define CONF_LITERAL 0.95
+#define CONF_PROP 0.85
+
+#define MAX_ITEMS 256
+#define MAX_SETTINGS 8
+
+/* ── Language gate ───────────────────────────────────────────────── */
+
+/* Visitor state for ObjectScript detection. */
+typedef struct {
+    bool found;
+} ens_lang_check_t;
+
+static void check_objectscript_node(const cbm_gbuf_node_t *node, void *userdata) {
+    ens_lang_check_t *s = (ens_lang_check_t *)userdata;
+    if (s->found || !node->file_path)
+        return;
+    const char *fp = node->file_path;
+    size_t len = strlen(fp);
+    if (len >= 4 && (strcmp(fp + len - 4, ".cls") == 0 || strcmp(fp + len - 4, ".mac") == 0 ||
+                     strcmp(fp + len - 4, ".int") == 0)) {
+        s->found = true;
+    }
+}
+
+/* Return true when the graph buffer contains at least one node from an
+ * ObjectScript source file (.cls / .mac / .int). */
+static bool has_objectscript_nodes(cbm_gbuf_t *gbuf) {
+    ens_lang_check_t state = {false};
+    cbm_gbuf_foreach_node(gbuf, check_objectscript_node, &state);
+    return state.found;
+}
+
+static const char *TOPOLOGY_SETTINGS[] = {"TargetConfigName", "PatientHost", "ConformanceOperation",
+                                          NULL};
+
+static const char *ENTRY_POINTS[] = {"OnProcessInput", "OnMessage", "OnRequest", "OnTask", NULL};
+
+typedef struct {
+    char setting_name[CBM_SZ_256];
+    char value[CBM_SZ_256];
+} ens_setting_t;
+
+typedef struct {
+    char item_name[CBM_SZ_256];
+    char class_name[CBM_SZ_256];
+    bool enabled;
+    ens_setting_t settings[MAX_SETTINGS];
+    int n_settings;
+} ens_item_t;
+
+typedef struct {
+    char production_class[CBM_SZ_256];
+    char file_path[CBM_SZ_512];
+    ens_item_t items[MAX_ITEMS];
+    int n_items;
+} ens_prod_def_t;
+
+static void extract_xml_attr(const char *xml, int offset, const char *attr, char *out, int outsz) {
+    char needle[CBM_SZ_64];
+    snprintf(needle, sizeof(needle), "%s=\"", attr);
+    const char *p = strstr(xml + offset, needle);
+    out[0] = '\0';
+    if (!p)
+        return;
+    p += strlen(needle);
+    const char *e = strchr(p, '"');
+    if (!e)
+        return;
+    int len = (int)(e - p);
+    if (len >= outsz)
+        len = outsz - 1;
+    memcpy(out, p, (size_t)len);
+    out[len] = '\0';
+}
+
+static bool is_topology_setting(const char *name) {
+    for (int i = 0; TOPOLOGY_SETTINGS[i]; i++)
+        if (strcmp(name, TOPOLOGY_SETTINGS[i]) == 0)
+            return true;
+    return false;
+}
+
+static ens_prod_def_t *parse_production_xml(const char *xml, const char *class_qn,
+                                            const char *file_path) {
+    ens_prod_def_t *def = calloc(1, sizeof(ens_prod_def_t));
+    if (!def)
+        return NULL;
+    snprintf(def->production_class, CBM_SZ_256, "%s", class_qn);
+    snprintf(def->file_path, sizeof(def->file_path), "%s", file_path ? file_path : "");
+
+    const char *p = xml;
+    while (*p && def->n_items < MAX_ITEMS) {
+        const char *item_start = strstr(p, "<Item ");
+        if (!item_start)
+            break;
+
+        ens_item_t *item = &def->items[def->n_items];
+        memset(item, 0, sizeof(*item));
+        item->enabled = true;
+
+        int off = (int)(item_start - xml);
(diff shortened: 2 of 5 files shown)


DeusData:
Thanks for splitting the Ensemble topology pass from the base ObjectScript support. This is a 0.9.2-rc language enhancement. Review will focus on bounded parsing of embedded production XML, deterministic EnsembleItem identity, conservative ROUTES_TO matching, malformed-input handling, and representative IRIS fixtures without requiring a live server.

DeusData:
Reviewed in full. **This is a marked step up from #1064**, and I want to say that first because the contrast is the most useful thing I can tell you.

**It demonstrably works end to end.** Your test runs the full pipeline on two real `.cls` fixtures and asserts at least two `EnsembleItem` nodes and at least one `ROUTES_TO` edge **in the dumped store** — return codes checked, post-dump, not at the extraction boundary. Every CI leg is green. That is exactly what #1064 was missing: there, the queries could not parse and the return codes were dropped, so it always returned empty while looking healthy. Here I could confirm the inputs genuinely exist (`XData` nodes are emitted at `extract_defs.c:5798`, `parent_class` at `pass_definitions.c:288`) and the outputs land in the store.

The review feedback from #467 was clearly absorbed too: the language gate, truncation warnings, the segment-anchored `class_name_matches` so `Ens` cannot match `Ens.BusinessService`, and `cbm_fopen` for Windows paths. The confidence scoring (0.95 literal, 0.85 property-resolved, −0.10 on the entry-point fallback) fits our existing cross-service linking model well.

**The direction question is the maintainer's, and it is about schema rather than quality.** `EnsembleItem` is a **new node label** and `ROUTES_TO` a **new edge type** — permanent core-schema surface, for one vendor's language. Whether vendor-specific labels belong in the core schema is his call, not mine, and I have put it to him.

Worth knowing while that is decided: `EnsembleItem` is excluded by all four hardcoded `label IN ('Function','Method','Class')` lists in `store.c` (arch boundaries, clustering, hotspots, vector search). For topology nodes that is arguably right — but it means `get_architecture` would show none of a production's topology, which cuts against the feature's stated purpose. Three separate contributors have now hit that same allowlist trap in this review round, so it is a systemic thing rather than a criticism of your PR.

**One defect I would fix before merge, because it is a graph-accuracy problem.**

`scan_source_for_send_targets` takes a `method_name` parameter and then discards it (`(void)method_name`, line 327). It scans the **whole file** and returns only the **first** `SendRequestSync` target. `resolve_method_routes` then attributes that single target to **every** production-member method in the file. So a class with N methods and M distinct targets produces N identical edges to the first target — wrong-method attribution and undercounting at the same time. Your single-method fixture cannot catch it; a two-method, two-target fixture would.

**And one thing to correct in the description:** the third bullet advertises WorkMgr dispatch — detecting `.Queue("##class(X).method")` and emitting CALLS edges. That code is not in the diff; there is no `Queue` or `##class` handling anywhere. Either it belongs to a different branch or the body is ahead of the code. Worth fixing so the next reviewer is not looking for it.

**Three smaller notes:**

- **The "zero overhead for non-IRIS projects" claim is not quite right.** `has_objectscript_nodes` walks **every node** in the graph buffer with a strlen and up to three suffix comparisons, and the visitor API has no early break — so it is O(N) per index, in both FAST and FULL modes. On an 8.5M-node graph that is tens of milliseconds; small, but not zero, and the discover layer already knows per-language file counts if you want a cheaper gate. Also `.cls` is shared with Apex, so Salesforce repos pass the extension gate too.
- `extract_xml_attr` searches unbounded from the tag offset, so an `Item` or `Setting` missing an attribute silently inherits it from a **later** tag — a wrong `Enabled` or a mispaired `ClassName`. Bounding the search to the tag's closing `>` fixes it.
- `properties_json` is assembled by `snprintf` without JSON-escaping the XML-derived strings, so a quote in an item or class name produces malformed JSON in the graph.

One genuine question rather than a defect: you match only `TargetConfigName` (singular), but real Ensemble routers commonly use `TargetConfigNames` (plural, comma-separated). Is the singular-only set deliberate?

Static extraction of production topology without a live IRIS instance is a real capability for IRIS shops. The open question on our side is the schema surface, not the work.

DeusData:
Maintainer's answer on the schema question, and it's a **yes to the capability with a change to how it's expressed**: please model the topology in the schema vocabulary we already have, rather than adding `EnsembleItem` and `ROUTES_TO`.

The reasoning is the one you'd expect — node labels and edge types are permanent, agent-visible surface, and a vendor-specific pair sets a precedent we'd owe every other platform. But that's not actually the strongest argument here. The strongest argument is that **the model you need already exists**, and adopting it fixes the problem flagged in my last review instead of leaving it open.

**The existing vocabulary**

`pass_route_nodes.c` already has a general channel-topology model that isn't HTTP-specific:

- **`Route`** — a named channel, deduplicated by a synthetic qualified name of the form `__route__<kind>__<name>`, with the transport recorded in properties (`{"broker":"..."}`). It's already used for pubsub, cloud tasks and generic broker channels, not just URLs.
- **`HANDLES`** — code → channel: "this method services this channel."
- **`ASYNC_CALLS`** — code → channel for non-HTTP transports. `pass_calls.c` picks this over `HTTP_CALLS` precisely when the target isn't an HTTP service.

An Ensemble production maps onto that almost exactly:

| yours | existing |
|---|---|
| `EnsembleItem` node | `Route`, qn `__route__ensemble__<Production>.<Item>`, props `{"broker":"ensemble","class":"...","enabled":...}` |
| `ROUTES_TO` edge | `ASYNC_CALLS` from the sending **method** to the target item's `Route` |
| (implicit today) | `HANDLES` from the item's implementing class/method to its own `Route` |

Note the second row also fixes the defect below for free: `ASYNC_CALLS` originates at the *sending method*, so per-method attribution becomes the natural shape rather than something you have to be careful about.

**What you get immediately by doing it this way**

This is the part I'd have led with if I'd seen it earlier. Last time I told you `EnsembleItem` is excluded by the four hardcoded label lists in `store.c`, so `get_architecture` would show none of a production's topology — cutting against the feature's whole purpose. `Route` doesn't have that problem:

- **`get_architecture` already renders `Route` nodes** in their own section (`arch_routes()` in `store.c`). Your topology shows up in the architecture view with no allowlist edits at all.
- **`search_graph` ranks `Route` above everything else** — the label boost for `Route` is higher than the one for `Function`/`Method` (`compute_search_score()` in `mcp.c`). Production items would surface at the top of results instead of being unfindable.
- Package/grouping queries already collect by `Route`.
- The cross-service matching phase in `pass_route_nodes.c` already bridges producer channels to handler channels by name — the same problem your pass solves, so there may be machinery to reuse rather than rewrite.

So this isn't "rename your labels to please the schema police." Going through `Route` is the difference between a topology the tools can see and one they can't.

If the mapping turns out to lose something real — a property with nowhere to live, or a relationship `ASYNC_CALLS` genuinely can't express — say so concretely and we'll look at it again. A gap that forces new vocabulary is a legitimate answer; I just want to be sure we've established there is one before minting labels.

**Still needs fixing regardless of the above**

The multi-target defect from my last review is a graph-accuracy bug and blocks merge either way: `scan_source_for_send_targets` discards its `method_name` parameter, scans the whole file, and returns only the first target — so every production-member method in a class gets an identical edge to that first target. Wrong-method attribution and undercounting in one. A two-method, two-target fixture catches it; your single-method fixture can't.

And the PR description still advertises WorkMgr dispatch (`.Queue("##class(X).method")` → CALLS edges) that isn't in the diff.

Your `TargetConfigNames` question from last time is still open and still yours to answer — I'd expect the plural form to matter in real routers.

Thanks for your patience on the direction call. The static-extraction-without-a-live-instance capability is genuinely worth having; I'd just like it speaking the vocabulary the rest of the graph already speaks.

isc-tdyar:
Rebased and reworked per your review:

- `EnsembleItem` nodes replaced with `Route` nodes (`__route__ensemble__<prod>.<item>`) so Ensemble topology is visible to `search_graph`, `trace_path`, and `get_architecture` without special-casing
- `ROUTES_TO` edges replaced with `ASYNC_CALLS` (method → Route) and `HANDLES` (entry-point → Route) using the existing vocabulary
- `scan_source_for_send_targets` now scopes its scan to the calling method's brace-delimited body — fixes the cross-method contamination where the first `SendRequestSync` in any method would win regardless of which method was being resolved
- `TargetConfigNames` (plural) added to `TOPOLOGY_SETTINGS` and split on comma so multi-target routers emit a separate edge per target
- Tests updated: `pipeline_ensemble_routing_edges` checks Route/ASYNC_CALLS/HANDLES; new `pipeline_ensemble_routing_method_scoping` test verifies no cross-contamination between two methods routing to different operations

DeusData:
Thank you for the rework, and I'm sorry it sat for three days — the delay was entirely on my side, not yours. You honoured the split out of #467, absorbed two full review rounds, and the 2026-08-24 rebase delivers what was asked for. I checked the parts you claimed and they are real: the vocabulary migration is complete (zero remaining references to `EnsembleItem` / `ROUTES_TO`; everything now rides on the existing `Route` / `ASYNC_CALLS` / `HANDLES` types), and the method-body-scoped scan genuinely fixes the multi-target mis-attribution — the old whole-file scan fails both `MethodTwo` assertions in your new regression test.

Because you had already waited on us and the remaining defects were mechanical rather than design work, I've **pushed the fixups directly to your branch** (6c2f82c8) rather than send you round again. You keep authorship of the pass; if you disagree with any of it, revert it and say so — you are the domain expert on IRIS/Ensemble and I am not.

**Three defects, each with a test that fails without the fix:**

1. **`TargetConfigNames` was dead code.** `emit_async_call` and `emit_handles` key Route nodes as `__route__ensemble__<prod>.<item>`, but the settings loop looked the *source* item up by its bare `<prod>.<item>` qn. `cbm_gbuf_find_by_qn` is an exact hashtable lookup (`graph_buffer.c:769`), so `item_node` was always `NULL` and every declarative routing edge was dropped by `continue`. It reads like an incomplete rename from the `EnsembleItem` version. Because no fixture contained a `<Setting>` element, nothing caught it — the new test builds a production whose *only* routing information is a `TargetConfigNames` setting, so any `ASYNC_CALLS` edge must have come from that path. Against the previous code it reports `async_calls == 0, expected 2`.

2. **Heap-buffer-overflow on malformed XML.** For an `<Item` with no closing `</Item>`, `item_end` became the buffer's NUL terminator, and `p = item_end + 7` then pointed seven bytes past the allocation — which the loop dereferenced before `strstr` scanned unbounded heap from there. ASan confirms it: `heap-buffer-overflow ... pass_ensemble_routing.c:544`. Worth noting how easy this was to hide: my first attempt at a fixture was truncated at the *class* level too, so the pass never reached the XData region and the test passed for the wrong reason. It only reproduces with a well-formed class and XData block containing a single unterminated item.

3. **`extract_xml_attr` was unbounded** — this is the one from the 2026-07-31 review. It scanned forward from the tag with no stop at the tag's closing `>`, so an item omitting an attribute silently inherited it from a *later* item. An `<Item>` with no `ClassName` therefore picked up the next item's class and emitted a bogus Route paired with the wrong implementation. Against the previous code the new test reports `ens_route_count == 2, expected 1`.

Also folded in: `read_file` ignored the `fread` result, which left an uninitialised tail for the string scans to walk on a short read.

Full local suite after the fixups: **7576 passed, 0 failed, 8 skipped** (139 suites).

**Two things still outstanding, both yours:**

- **The PR description is stale.** It still describes "Pass A — EnsembleItem nodes", "Pass B — ROUTES_TO edges", and advertises WorkMgr `.Queue("##class…")` dispatch. None of that is in the diff — the commit message is accurate, the body never got updated. This is the third time it's come up, and since it's what a reader sees first, please bring it in line with what the pass actually does.
- **One design question I answered by inference, so please confirm.** I made the settings-derived edge `Route → Route` (source item's Route node to target item's Route node), because that is what `emit_async_call` already does for the target and it keeps a configured routing relationship between config items rather than mixing a code node in. If you intended the source to be the *implementing class* node instead, say so and I'll change it — the same class can back multiple config items, which is why I went the other way.

Minor, non-blocking, entirely up to you: the language gate is still an O(N) full-node walk with no early break and runs in FAST mode too, so the body's "zero overhead" claim is a little generous; and the commit footer says `Fixes #467`, which is an already-merged PR — the correct reference is `Closes #1061`, which the body already has.

Thanks again for your patience on this one, and for the quality of the rework.

DeusData:
Merged in 20327cd05d48. Thank you, Thomas — this was a genuinely well-run contribution: you took the split out of #467 without complaint, absorbed two review rounds, and the topology pass now gives IRIS/Ensemble productions real structure in the graph rather than opaque `.cls` files.

Full 34-check CI green on the final SHA, merged with `--merge` so your authorship and sign-off are preserved intact.

Two things I raised in my previous comment are still worth closing out, and I did **not** want to hold the merge for them:

1. **The `Route → Route` edge source for settings-derived routing was my inference, not your design.** If a `TargetConfigNames` edge should originate from the implementing class node instead, say so and I'll open the follow-up myself — you know Ensemble semantics and I don't. My reasoning was that the same class can back multiple config items, so the configured item is the more faithful source, and it matches how `emit_async_call` already resolves the target.
2. **The PR description never got updated** and now lives in the merged record describing "Pass A — EnsembleItem nodes" and WorkMgr dispatch that the pass does not implement. Your commit message is accurate, so the history is fine; it is only the PR body. Not worth reopening anything over — flagging it because it will confuse anyone who finds this PR later.

Also worth saying plainly: the `TargetConfigNames` defect was the kind that survives review precisely because it looks healthy — no crash, no error, just an exact-match lookup that silently never matches and a `continue` that drops every edge. It only became visible once there was a fixture with a `<Setting>` in it. That is a good argument for the reproduce-first habit on any path that can fail by returning nothing.

#1061 closes with this. If you want to take the WorkMgr `.Queue` dispatch you originally described as a follow-up, I would review it gladly.

isc-tdyar:
Route → Route is correct for settings-derived edges. In Ensemble, a `TargetConfigNames` setting belongs to the config item, not to a specific method in its implementing class — the same class can appear under different config item names in the same production, and the routing relationship is between config items, not between code and config. Your inference matched the semantics.

On WorkMgr dispatch: yes, I'd like to take that as a follow-up. I'll open an issue first per CONTRIBUTING.md.

PR #1064, opened 2026-07-12
feat(pipeline): version tag on indexed nodes + diff_versions MCP tool

### LINKED ISSUE #1062: Version tags on indexed nodes + diff_versions tool for cross-version change detection
Feature proposal: stamp a version tag on every node at index time, then compare two indexed versions of the same project to find added/removed nodes.

## Motivation

When indexing multiple releases of a large codebase (e.g. version 28.0 and 30.0 of a product with 1,000+ classes), it's useful to query *what changed* between versions without diffing source files directly. The graph representation makes cross-version analysis composable with the rest of the Cypher query surface (`DEFINES_METHOD`, `CALLS`, `IMPORTS`, etc.).

## Proposed API

**`cbm_pipeline_set_version(cbm_pipeline_t *p, const char *version_tag)`**

Sets a version label on a pipeline instance. When set, the tag is written as a `"version"` property on every node. Also bypasses the incremental-index path so versioned indexing always does a full pass — the graph accumulates both versions side-by-side rather than replacing the previous snapshot.

**`index_repository` tool: new `version` parameter**

```json
{ "repo_path": "/repos/myproject-28.0", "version": "28.0" }
```

When omitted, a version-like segment (digits and dots) is auto-derived from the repo path.

**`diff_versions` MCP tool**

```json
{
  "project":      "myproject",
  "from_version": "28.0",
  "to_version":   "30.0",
  "label":        "Class"
}
```

Returns `{"added": [...], "removed": [...], "from_version": "28.0", "to_version": "30.0"}`. Default label `Class`, limit 500 per side.

## Implementation scope

- `pipeline.h` — new public `cbm_pipeline_set_version()` 
- `pipeline_internal.h` — `version_tag` field on ctx
- `pipeline.c` — setter, free, skip-incremental guard, ctx wiring
- `pass_definitions.c` / `pass_parallel.c` — `version` property on all nodes; Storage nodes with structured docstring have fields merged inline as raw JSON
- `mcp.c` — `cbm_derive_version_from_path()` helper, version param on `index_repository`, `diff_versions` handler

## Questions before PR

1. Is there appetite for this feature? It's general-purpose (not IRIS-specific) but the primary motivation is multi-version IRIS codebase analysis.
2. Any preference on how versioned graphs are stored — single DB with version property on nodes (proposed), or separate DBs per version?
3. The `diff_versions` tool only compares node existence (added/removed by name). Should it also surface structural changes (e.g. method count delta, signature changes)?

isc-tdyar:
Adds version tagging on all indexed nodes and a `diff_versions` MCP tool for cross-version change detection.

Closes #1062. Depends on #467 (ObjectScript language support) merging first — the Storage node inline-JSON change targets ObjectScript Storage nodes.

## New API

**`cbm_pipeline_set_version(cbm_pipeline_t *p, const char *version_tag)`**

Sets a version label (e.g. `"28.0"`) on a pipeline instance. When set, the tag is written as a `"version"` property on every node. Also bypasses the incremental-index path so versioned indexing always does a full pass — the graph accumulates both versions side-by-side.

**`index_repository` tool: new `version` parameter**

```json
{ "repo_path": "/repos/myproject-28.0", "version": "28.0" }
```

When omitted, a version-like segment (digits and dots) is auto-derived from the repo path. No behavior change for projects that don't use versioning.

**`diff_versions` MCP tool**

```json
{
  "project":      "myproject",
  "from_version": "28.0",
  "to_version":   "30.0",
  "label":        "Class"
}
```

Returns `{"added": [...], "removed": [...], "from_version": "28.0", "to_version": "30.0"}`. Default label `Class`, limit 500 per side.

## Files changed

- `src/pipeline/pipeline.h` — `cbm_pipeline_set_version()` declaration
- `src/pipeline/pipeline_internal.h` — `version_tag` field on ctx
- `src/pipeline/pipeline.c` — setter, free, skip-incremental guard, ctx wiring
- `src/pipeline/pass_definitions.c` — `version` property on all nodes; Storage nodes with structured docstring have fields merged inline as raw JSON rather than escaped
- `src/pipeline/pass_parallel.c` — same changes mirrored; `version_tag` threaded through parallel worker
- `src/mcp/mcp.c` — `cbm_derive_version_from_path()` helper, `version` param on `index_repository`, `diff_versions` tool registered and handler implemented

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -701,6 +701,7 @@ static const tool_annotation_def_t TOOL_ANNOTATIONS[] = {
     {"detect_changes", false, true, true, false},
     {"manage_adr", false, true, false, false},
     {"ingest_traces", false, false, false, false},
+    {"diff_versions", true, false, true, false},
 };
 
 static const tool_annotation_def_t *mcp_tool_annotations(const char *name) {
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -90,6 +90,7 @@ enum {
 #include <time.h>
 #include <errno.h>
 #include <stdatomic.h>
+#include <ctype.h>
 
 /* ── Constants ────────────────────────────────────────────────── */
 
@@ -375,8 +376,25 @@ static const tool_def_t TOOLS[] = {
      "are normalized.\"},"
      "\"persistence\":{\"type\":\"boolean\",\"default\":false,\"description\":"
      "\"Write compressed artifact to .codebase-memory/graph.db.zst for team sharing. "
-     "Teammates can bootstrap from the artifact instead of full re-indexing.\"}"
+     "Teammates can bootstrap from the artifact instead of full re-indexing.\"},"
+     "\"version\":{\"type\":\"string\",\"description\":"
+     "\"Version tag stamped on every extracted node (e.g. '28.0', '30.0'). "
+     "Enables cross-version diff queries: MATCH (n:Class {version:'30.0'}) ... "
+     "Auto-detected from trailing path segment when the segment is numeric (e.g. "
+     "/exports/MyApp/28.0 → version='28.0'). Versioned runs always do a full index.\"}"
      "},\"required\":[\"repo_path\"]}"},
+
+    {"diff_versions", "Diff versions",
+     "Compare two indexed versions of the same project to find added, removed, and "
+     "changed nodes. Requires version tags set via index_repository version= parameter.",
+     "{\"type\":\"object\",\"properties\":{"
+     "\"project\":{\"type\":\"string\",\"description\":\"Project name\"},"
+     "\"from_version\":{\"type\":\"string\",\"description\":\"Baseline version tag (e.g. "
+     "'28.0')\"},"
+     "\"to_version\":{\"type\":\"string\",\"description\":\"Comparison version tag (e.g. "
+     "'30.0')\"},"
+     "\"label\":{\"type\":\"string\",\"description\":\"Node label to compare (default: Class)\"}"
+     "},\"required\":[\"project\",\"from_version\",\"to_version\"]}"},
 
     {"search_graph", "Search graph",
      "Search the code knowledge graph for functions, classes, routes, and variables. Use INSTEAD "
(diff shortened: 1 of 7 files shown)


DeusData:
The current PR cannot be reviewed against its stated purpose. Its title and body describe version tags and a diff_versions tool, but the branch currently contains the same body-token extractor change as PR #1060 and no versioning or MCP-tool implementation. Please restore the intended branch contents or close this PR if it was superseded; we should not review or merge a title/diff mismatch.

isc-tdyar:
You were right, the branch did contain the body-token change from #1060 instead of the versioning work. That was a mistake in how I built the branch, not a stale description. It has since been rebuilt from scratch off `main` with the intended contents, and I've just rebased it again onto current `main` to clear the merge conflict.

What's in the diff now:

- `version_tag` plumbed through the pipeline (`cbm_pipeline_set_version()`, stamped onto nodes at insert time, including the ObjectScript parallel extraction path)
- `diff_versions` MCP tool: added/removed node detection between two version tags via `OPTIONAL MATCH` anti-joins
- optional `version` argument on `index_repository`, plus derivation from a trailing version-like path segment (`/repos/MyApp/28.0` -> `28.0`)
- `diff_versions` registered in `TOOL_ANNOTATIONS` as read-only, non-destructive, idempotent, and in the annotation-coverage test

Incremental indexing is skipped when a version tag is set, since a version stamp needs a full pass to be coherent.

Sorry for the wasted review cycle on the mismatch. Ready for review when you get to it, and no rush, I know #1063 is also in your queue.

DeusData:
Reviewed in depth. Thank you for the care in how you brought this — and I owe you an apology up front: you asked "is there appetite for this feature?" in #1062 and never got an answer before you built it. That silence is on us, and it is the reason you spent effort on something that now needs a direction decision it should have had first.

**Contributor hygiene here was model.** Issue first, clean DCO'd commits, the parallel extraction path threaded correctly via `extract_ctx_t.version_tag` in `pass_parallel.c` (easy to miss, and people usually do), and honest tool annotations — read-only, idempotent, non-destructive — mirrored in the coverage test.

**Two findings you will want regardless of which way the direction call goes, because both mean the feature cannot work today.**

**1. The diff queries cannot parse.** They use a variable property reference as a map value:

```
{name: n.name, version: '…'}
```

Our Cypher subset does not support that. `parse_props()` in `src/cypher/cypher.c` does `expect(p, TOK_STRING)` — property maps accept **string literals only**, so `n.name` lexes as IDENT DOT IDENT and the parse fails. Both queries error on every invocation.

**2. Two versions can never coexist in one store.** Setting a version forces the full-index path, and that path runs `try_incremental_or_delete_db()` in `src/pipeline/pipeline.c`, which logs "deleting old db" and `cbm_unlink`s the database before the full pass. So indexing 30.0 **erases** 28.0. Using different project names does not help either — those are separate DB files and `resolve_store(project)` opens exactly one. The "both versions accumulate side-by-side" premise is contradicted by the code the PR modifies. (Even within a single run, `cbm_gbuf_upsert_node` dedups by qualified name.)

**Why this shipped looking fine:** `handle_diff_versions` ignores both `cbm_cypher_execute` return codes, so a parse failure is indistinguishable from "nothing changed" and the tool confidently returns `{"added":[],"removed":[]}`. A single functional test calling the tool against a seeded store would have caught both defects immediately — right now the only test line registers annotations, which passes whether or not the tool does anything.

**One thing I would ask you to drop in any revision, independent of everything else.** `cbm_derive_version_from_path` treats any repository whose basename is digits and dots as versioned — `/builds/20260731`, `…/1.2.3`, `…/2024.1`. For those users, incremental indexing is silently disabled forever: every index becomes a database delete plus a full reindex, which at our scale is minutes and tens of gigabytes of RSS, plus version-stamped props they never asked for. Version should stay an explicit opt-in parameter.

**On cost, since it matters for the direction call:** the tag is a per-node string copy rather than an interned id — `,"version":"28.0"` is 17 bytes per node, roughly **145 MB at 8.5M nodes**, paid in graph-buffer RAM, in the store, and in the persistence artifact. No schema change and no forced migration, though, so legacy indexes are unaffected — that part is clean.

**What is actually being decided.** The interesting question is not the tool; it is whether we want a **multi-version snapshot concept in the graph data model at all**. With N versions co-resident, `search_graph` and `trace_path` would return N nodes per symbol, which is direct findability pollution and cuts against our first priority. For what it is worth, `diff_versions` is *not* a duplicate of `detect_changes` — different axis, indexed-snapshot comparison versus git-diff blast radius — but it is a canned two-query wrapper over the same Cypher engine `query_graph` already exposes, so the tool is the cheap part. If the answer is yes, the shape is probably store-level snapshots, or per-version projects with a cross-store diff, rather than a property on every node.

That is with the maintainer now, along with the two defects above so the decision is made on accurate information. I will come back to you with a real answer this time.

Two smaller notes for whenever it resumes: the diff identity uses short `name` rather than `qualified_name`, so same-named symbols across packages collapse and renames double-count; and the tool description promises "changed" nodes while only added and removed exist.

DeusData:
Declining this one — and the first thing I owe you is an apology, not a review.

**You asked whether we wanted this in #1062 and never got an answer.** You then went and built it. That's our failure, not yours, and it's the worst way for a contributor's time to be spent: not on a rejected implementation, but on one that was never green-lit because nobody replied. I'm sorry.

**The direction answer, belatedly: no.** A multi-version graph model cuts against this project's central priority. Findability comes first here — and indexing several versions side by side means every symbol returns duplicate hits across versions, so `search_graph` and `trace_path` get *worse* for everyone in order to serve a comparison case that `git diff` and a second indexed project already cover. That's a bad trade at the level of the whole product, and it's the reason rather than anything about your code.

Since you'd want to know what a review found regardless, there are two defects that make the feature non-functional as submitted, and I mention them because they'd bite in any future version of this idea:

1. **The Cypher queries can't run.** They reference a variable inside a map literal, which our parser rejects — it accepts string literals only there. Every `diff_versions` call fails. It's invisible because the tool handler discards the Cypher return code, so the failure surfaces as an empty result rather than an error.
2. **Versions erase rather than accumulate.** Setting a version forces the full-reindex path, which deletes the existing database first. So indexing v2 destroys v1, and the compare-two-versions premise doesn't hold. Auto-detection from a path segment can trigger that delete-and-full-reindex on any repo with a digits-or-dots directory name, without the user opting in at all.

The second one is worth carrying forward as a general lesson about this codebase: the pipeline publishes by renaming a staging database over the live one, so *anything* that wants data to persist across indexes needs explicit machinery. ADRs only survive because someone built capture-and-restore for them.

Closing this. If you have an appetite for the underlying need — understanding what changed structurally between two states of a codebase — I'd be glad to talk about it as an issue first, and this time you'll get an answer before you write code.

Issue #1017, opened 2026-07-11
docstring truncation splits multi-byte UTF-8 characters, producing invalid UTF-8 in nodes.properties (non-ASCII comments, e.g. Japanese)


jimeng754-eng:
## Summary

Long docstrings/comments containing non-ASCII multi-byte UTF-8 text (observed with Japanese) can get truncated mid-character during extraction, leaving an incomplete UTF-8 byte sequence embedded in the `properties` JSON stored in the SQLite `nodes` table. This produces a `properties` value that is syntactically valid JSON but semantically invalid UTF-8 — any consumer that decodes the SQLite TEXT column strictly as UTF-8 (e.g. Python's `sqlite3` module, which is the default for anyone building tooling on top of the exported database) throws on read.

Because this only manifests with multi-byte UTF-8 characters, it never reproduces on English-only codebases — a single-byte-per-character comment can be truncated anywhere without producing invalid UTF-8. It only shows up when indexing repos with non-ASCII comments (Japanese/Chinese/etc.).

## Environment

- `codebase-memory-mcp` version: 0.9.0 (Windows amd64, pre-built release binary)
- Repro repo: [EC-CUBE/ec-cube](https://github.com/EC-CUBE/ec-cube) (public, MIT-licensed Japanese e-commerce platform — real-world Japanese comments throughout)
- Indexed with `mode=full`

## Repro steps

```
git clone --depth 1 https://github.com/EC-CUBE/ec-cube.git
codebase-memory-mcp cli index_repository --repo-path <path>/ec-cube --mode full --name ec-cube
```

Then read the `nodes.properties` column with any strict-UTF-8 SQLite client, e.g.:

```python
import sqlite3
conn = sqlite3.connect("ec-cube.db")
conn.execute("SELECT id, name, file_path, properties FROM nodes WHERE label IN ('Function','Method')").fetchall()
# sqlite3.OperationalError: Could not decode to UTF-8 column 'properties' with text '...'
```

This one bad row aborts the *entire* `fetchall()` — not just that row — since sqlite3 decodes the whole result set as UTF-8 by default.

## Observed corrupted data

Node `id=5570`, `qualified_name="ec-cube.src.Eccube.Controller.ShoppingController.ShoppingController.index"`, `file_path="src/Eccube/Controller/ShoppingController.php:72-151"`.

Raw bytes of the `properties` value around the break (hex):
```
[...]

PR #1065, opened 2026-07-12
fix(pipeline): escape dynamic edge properties as JSON

tmonestudio:
## What does this PR do?

Prevents malformed JSON in edge properties when parser-derived values contain JSON-significant characters.

Two edge builders serialized dynamic strings directly:
- sequential IMPORTS edges wrote `local_name` without escaping (defence-in-depth; import aliases are identifier-shaped);
- parallel GRAPHQL_CALLS edges escaped the callee but not the operation text.

The reachable GraphQL reproduction is a single-line, single-quoted Python string whose operation contains a bare double quote.

## Fix

Use the existing `cbm_json_escape` helper for both dynamic values before formatting the properties object.

The regression fixture keeps the whole-table `json_valid(properties)` invariant and now passes an embedded quote through extraction to `emit_graphql_edge`.

## Validation

- Focused GREEN proof with the production fix: https://github.com/tmonestudio/codebase-memory-mcp/actions/runs/30671817289
- Focused RED proof with both production escaping hunks reverted: https://github.com/tmonestudio/codebase-memory-mcp/actions/runs/30671818737
- RED failure: `contract_edge_parallel_service_edges` reports `invalid_props == 1`.
- Full upstream CI runs on the final PR head.

## Checklist

- [x] Regression fixture is demonstrably RED without the fix and GREEN with it
- [x] Whole-table JSON-validity invariant retained

### DIFF
--- a/src/pipeline/pass_definitions.c
+++ b/src/pipeline/pass_definitions.c
@@ -24,6 +24,7 @@ enum { PD_JSON_FIELD_OVERHEAD = 6 };
 #include "foundation/compat.h"
 #include "foundation/compat_fs.h"
 #include "foundation/limits.h"
+#include "foundation/str_util.h"
 #include "cbm.h"
 #include "arena.h"
 #include "iris_export_xml.h"
@@ -470,9 +471,10 @@ static int create_import_edges_for_file(cbm_pipeline_ctx_t *ctx, const CBMFileRe
         const cbm_gbuf_node_t *target =
             cbm_pipeline_resolve_import_node(ctx, rel, file_qn, imp, namespace_map);
         if (target && target->id != source_node->id) {
+            char esc_ln[CBM_SZ_128];
+            cbm_json_escape(esc_ln, sizeof(esc_ln), imp->local_name ? imp->local_name : "");
             char imp_props[CBM_SZ_256];
-            snprintf(imp_props, sizeof(imp_props), "{\"local_name\":\"%s\"}",
-                     imp->local_name ? imp->local_name : "");
+            snprintf(imp_props, sizeof(imp_props), "{\"local_name\":\"%s\"}", esc_ln);
             cbm_gbuf_insert_edge(ctx->gbuf, source_node->id, target->id, "IMPORTS", imp_props);
             count++;
         }
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -1921,10 +1921,12 @@ static void emit_graphql_edge(cbm_gbuf_t *gbuf, const cbm_gbuf_node_t *source, c
         cbm_gbuf_upsert_node(gbuf, "Route", p, route_qn, "", 0, 0, "{\"source\":\"graphql\"}");
 
     char esc_c[CBM_SZ_256];
+    char esc_op[CBM_SZ_512];
     cbm_json_escape(esc_c, sizeof(esc_c), call->callee_name);
+    cbm_json_escape(esc_op, sizeof(esc_op), p);
     char props[CBM_SZ_1K];
     snprintf(props, sizeof(props), "{\"callee\":\"%s\",\"operation\":\"%s\",\"confidence\":%.2f}",
-             esc_c, p, res->confidence);
+             esc_c, esc_op, res->confidence);
     cbm_gbuf_insert_edge(gbuf, source->id, route_id, "GRAPHQL_CALLS", props);
 }
 
--- a/tests/test_lang_contract.c
+++ b/tests/test_lang_contract.c
@@ -24,6 +24,7 @@
 #include <store/store.h>
 #include <pipeline/pipeline.h>
 #include <foundation/log.h>
+#include <sqlite3.h>
 
 #include <string.h>
 #include <stdlib.h>
@@ -1518,7 +1519,9 @@ TEST(contract_edge_parallel_service_edges) {
         {"gql.py", "def gql(query_string):\n    return query_string\n"},
         {"client.py",
          "from gql import gql\n\n\ndef fetch_user():\n"
-         "    return gql(\"query GetUser { user { id name } }\")\n\n\n"
+         /* The embedded double quote survives extraction from this single-quoted
+          * Python string and breaks GRAPHQL_CALLS JSON unless operation is escaped. */
+         "    return gql('query GetUser { user(name: \"quoted\") { id name } }')\n\n\n"
          "def create_user():\n"
          "    return gql(\"mutation CreateUser { addUser(name: \\\"x\\\") { id } }\")\n"},
         /* TRPC_CALLS: local createTRPCProxyClient (same-module resolution). */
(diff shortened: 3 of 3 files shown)


DeusData:
Thanks for the focused JSON escaping fix. This is a high-priority 0.9.1-rc graph-correctness PR. Review will verify that dynamic edge properties are escaped identically in sequential and parallel paths and that quotes, backslashes, control characters, and non-ASCII values remain valid round-trip JSON.

DeusData:
The fix is correct and I want to land it — escaping parser-derived strings through `cbm_json_escape` before embedding them in properties JSON is exactly right (same class as the service-edge JSON-validity work in #898 and the UTF-8 fix in #1017). One thing to fix first, though:

**The test as written is a vacuous guard.** I verified it: reverting both production changes (the `esc_ln`/`esc_op` escaping in pass_definitions.c and pass_parallel.c) and re-running `contract_edge_parallel_service_edges` still passes — the `AND json_valid(properties)=0 → invalid_props==0` assertion never fires because the fixture contains no string with a JSON-breaking character. So the test passes with **and** without your fix, which means it can't catch a regression.

To make it a genuine reproduce-first guard, the fixture needs a value that actually breaks JSON when unescaped — e.g. an import whose `local_name` or a service-edge URL/topic string contains a double-quote, backslash, or a CR/LF. With that, `json_valid(properties)` would be 0 on unfixed code (test RED) and 1 with your escaping (GREEN). A URL argument with an embedded quote is probably the easiest to construct in the fixture.

Could you add that? Once the test is RED-without-the-fix I'll merge immediately — the production change is already good. Thanks for closing a real JSON-corruption gap.

tmonestudio:
Thanks — addressed. The fixture now includes a GraphQL operation with a literal double quote (`user(name: "quoted")`), so reverting the `esc_op` escaping makes `json_valid(properties)` fail, while the patched code keeps the properties valid JSON.

I also cleaned the branch onto current `main` as a single commit (`f809e98`). The previous Ubuntu smoke failure was the unrelated EOF shutdown timeout (`binary did not exit within 5 seconds after EOF`); the fresh `pr-smoke (ubuntu-latest)` run is now green, as is macOS. The remaining long-running checks are still in progress. Ready for re-review — thank you!

DeusData:
Thanks for addressing the vacuous regression and for cleaning the branch. Current head `4811126908d77c274850956a8cc9c19e099a41e6` has a green check set. The exact-head re-review will verify that the literal-quote fixture is genuinely RED when the production escaping is reverted and GREEN with the fix. No additional contributor action is requested in the meantime.

DeusData:
Thank you for this — **the production fix is exactly right and we want it.** One problem with the regression test, and then it merges.

**What is correct.** Both sites interpolated parser-derived text straight into JSON `properties` without escaping: `local_name` on IMPORTS (`pass_definitions.c:474`) and `operation` on GRAPHQL_CALLS (`pass_parallel.c:1997`). Nice catch that `callee` was already escaped on that second one and `operation` was not — that asymmetry is easy to read past.

You reused `cbm_json_escape` rather than hand-rolling a second escaper, which is the right instinct. I checked the helper's truncation behaviour too: each escape tests remaining space *before* writing, so output can never end mid-sequence with a dangling backslash. And I did the sizing math on the wrapper — worst case for the graphql props is about 812 bytes against a 1 KB buffer, imports 144 against 256, so it never truncates the closing `"}`. Incidentally your fix also closes a pre-existing malformation vector there: a raw `local_name` over ~238 bytes previously *could* truncate that closing brace.

I tried to break it and could not. Bare quote, trailing backslash, control characters, embedded NUL, and a crafted-length truncation attack are all contained post-fix. The round trip is clean too: encoded once, bound verbatim through the parameterised `bind_text`, decoded once by readers — no double-escape anywhere.

**The problem: the regression test passes with and without your fix.**

The fixture uses a *multiline* triple-quoted Python string. That operation never reaches `emit_graphql_edge` at all — `strip_and_validate_string_arg` (`extract_calls.c:1450`) rejects any string argument containing `\n`, since only `\t` is exempt below `MIN_PRINTABLE`. So `first_string_arg` comes back NULL and the emitter falls back to the callee name `"gql"`, which is perfectly valid JSON either way. Your `ASSERT_EQ(invalid_props, 0)` therefore holds on unpatched main.

The same applies to the mechanism described in the PR body: a multiline operation producing malformed JSON is end-to-end unreachable. **The genuinely reachable vector is a bare double quote on a single line** — which is a better bug than the one you described, and your fix already handles it.

**The change I would like**, and then this goes in:

```python
return gql('query GetUser { user(name: "quoted") { id name } }')
```

A single-quoted Python string, so the inner double quote survives extraction, and every character is ≥ 0x20 so validation passes it through. Pre-fix that emits `"operation":"GetUser { user(name: "quoted")…"` → `json_valid` returns 0 → your assertion goes red.

Please confirm it fails with the production hunks reverted before pushing. That revert-check is the whole point of the exercise here — a green test that would also be green without the fix tells us nothing later.

**Two smaller notes, neither blocking:**

- The IMPORTS `local_name` path has no discriminating fixture (the alias in the test is the benign `"gql"`), and the "multiline Rust import alias" scenario in the description does not really exist since Rust aliases are identifiers. That is completely fine as defence-in-depth — just worth labelling it as such rather than as a reproduction.
- The whole-table `json_valid(properties) = 0` sweep is a genuinely good invariant-style assertion. Please keep it; it becomes valuable the moment the fixture actually exercises the escape path.

One thing I noticed while reviewing that is **not** yours: `emit_grpc_edge` interpolates `service` and `method` raw into properties using the same pattern. Both derive from identifier-shaped text so the risk is low, but we will track it on our side.

tmonestudio:
Updated as requested. The fixture now uses the reachable single-line, single-quoted Python form, so the embedded double quote reaches `emit_graphql_edge`.

I also ran the explicit sensitivity check before pushing the final branch:

- RED, with both production escaping hunks reverted: https://github.com/tmonestudio/codebase-memory-mcp/actions/runs/30671818737 — `contract_edge_parallel_service_edges` fails with `invalid_props == 1`.
- GREEN, with the production fix present: https://github.com/tmonestudio/codebase-memory-mcp/actions/runs/30671817289

The PR is now rebased on current `main` as one signed-off commit (`2d6766f6`). I kept the whole-table `json_valid(properties)` invariant and corrected the PR description to distinguish the reachable GraphQL reproduction from the defensive IMPORTS hardening. Full CI is running on the final head.

DeusData:
Thank you for addressing the earlier review points, adding the RED/GREEN sensitivity evidence, and rebasing the branch. I have recorded the update and returned the current diff to the maintainer review queue. A complete review may still take a little time while the community queue is full.

tmonestudio:
Rebased this PR onto current upstream `main` (`7f23a66a`, v0.10.2+) and force-updated the head with lease protection.

- new head: `283321f0`
- scope remains the two JSON-escaping production sites plus the reachable quoted GraphQL regression
- `git diff --check`: clean
- Windows production GCC build: GREEN (`A5608EA18D4D5D2E579A90D7410ED3A0E05FC24BE9EF000E587A4E1258B715E5` for the combined updated-base candidate)
- focused native suites on the updated base (`mcp lang_contract`): **224 passed, 7 platform skips**
- discriminating fixture `contract_edge_parallel_service_edges`: PASS

Local GCC 15 required suppressing two unrelated warnings-as-errors in existing daemon tests (`free-nonheap-object`, `stringop-overflow`) to reach the focused suites; no source suppression was committed. Fresh CI on this exact head is the authoritative pinned-toolchain result.

tmonestudio:
CI update: all 34 checks on `283321f0` have passed. The rebased PR remains conflict-free and limited to the JSON edge-property escaping fix plus its regression coverage.

DeusData:
Thank you @tmonestudio for rebasing and reporting the pinned verification. The current head `283321f0` is mergeable, all checks are green, and the scope remains the two JSON edge-property serialization sites plus their regression coverage.

This is recorded back in the maintainer review queue. The queue is still full, so the final code review may take some time, but no additional triage information is needed from you right now.

DeusData:
Merged as your commit — thank you, and thank you especially for the two rounds of test iteration.

That part is worth calling out, because it's the difference between a fix and a *proven* fix: your first two fixtures were vacuous — the values never reached the vulnerable interpolation — and rather than argue the point you rebuilt the fixture until it actually exercised the path, then demonstrated RED-with-the-fix-reverted and GREEN-with-it via linked CI runs. A lot of contributors would have stopped at "the test passes".

The bug itself was a good catch precisely because it was asymmetric: `callee` on the same edge was already escaped, so `local_name` and `operation` sitting unescaped next to it read as an oversight rather than a design choice — which is exactly what it was. Malformed JSON in stored edge properties is the kind of thing that surfaces much later as an unexplained query failure.

The whole-table `json_valid(properties)` invariant you added is the right shape too. It doesn't just pin these two fields; it catches the next unescaped interpolation anyone adds anywhere in edge-property construction.

Landed in commit 283321f0 (merge e830293e). Note this doesn't rewrite properties already stored by earlier versions — only newly-written edges are escaped — so anyone with an existing index picks it up on their next reindex.

PR #1067, opened 2026-07-12
fix(tests): isolate integration caches from user state

tmonestudio:
## What does this PR do?

Prevents the C test runner from reading or writing a user's real codebase-memory cache.

Several integration fixtures exercise the production `index_repository` flow. The production cache resolver prefers `CBM_CACHE_DIR`, while older fixtures derive the expected database path from `HOME/.cache/codebase-memory-mcp`. When the runner inherited a real cache, early assertion returns could leave fixture databases behind. Setting only `CBM_CACHE_DIR` was also insufficient because those legacy paths diverged.

## Fix

- create a private temporary `HOME` for every normal test-runner process;
- clear inherited `CBM_CACHE_DIR`, so production code and legacy fixtures converge on the private HOME cache;
- delete the sentinel tree with `atexit`, including after early test returns;
- make the lang-contract helper honor an explicit `CBM_CACHE_DIR` when a test intentionally sets one.

Worker/probe subprocess modes still exit before the sentinel setup.

## Validation

Windows GCC build after rebasing onto current `main`:

- full C runner: **5831 passed, 22 skipped**;
- production-cache sentinel: **501 DB files before, 501 after**;
- added files: **0**;
- removed files: **0**.

The same run previously left `cbm_seq898_*`, `cbm_lc_*`, and `cbm_fcw_*` fixture databases in the inherited cache.

## Checklist

- [x] Full C test runner passes locally
- [ ] Full lint (left to CI)
- [x] Behavior validated with a before/after cache sentinel

### DIFF
--- a/tests/test_lang_contract.c
+++ b/tests/test_lang_contract.c
@@ -71,13 +71,18 @@ static cbm_store_t *lang_open_indexed(LangProj *lp) {
     if (!lp->project) {
         return NULL;
     }
-    const char *home = getenv("HOME");
-    if (!home) {
-        home = "/tmp";
-    }
     char cache_dir[512];
-    snprintf(cache_dir, sizeof(cache_dir), "%s/.cache/codebase-memory-mcp", home);
-    cbm_mkdir(cache_dir);
+    const char *configured_cache = getenv("CBM_CACHE_DIR");
+    if (configured_cache && configured_cache[0]) {
+        snprintf(cache_dir, sizeof(cache_dir), "%s", configured_cache);
+    } else {
+        const char *home = getenv("HOME");
+        if (!home) {
+            home = "/tmp";
+        }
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for isolating integration-test caches from real user state. This is a high-priority 0.9.1-rc safety fix. Review will verify that every relevant test redirects both cache and configuration roots, restores environment state on all exits, and cannot read or mutate a developer installation.

DeusData:
Merged (bb9677f). Verified the isolation works end-to-end: ran the integration + lang_contract suites against a build with your change and confirmed the developer's real `~/.cache/codebase-memory-mcp` was byte-for-byte untouched (80 entries before and after) — the tests now redirect both the cache (`CBM_CACHE_DIR`) and HOME to private temp roots, with state restored on exit. Genuinely valuable hygiene fix; tests exercising the production `index_repository` flow should never be able to read or mutate a developer's real installation. Thank you.

PR #1075, opened 2026-07-13
fix(extract,cypher): decorators dropped after a comment; composite properties truncated at the first comma

JhohanBustamante:
Two independent bugs found while indexing a NestJS + Angular monorepo. Both make
decorator/route/authz queries unusable on frameworks whose decorators take
arguments — and one of them makes a well-documented endpoint disappear from the
graph entirely. Each commit is self-contained and carries a regression test.

## 1. A comment between decorators drops every decorator above it

`extract_decorators()` walks prev-siblings and breaks on the first NAMED node.
Comments **are** named nodes in tree-sitter, so a comment interleaved in a
decorator run ends the walk:

```ts
@Post('login')                     // <-- dropped
@HttpCode(HttpStatus.OK)           // <-- dropped
// throttled per IP and per account
@Throttle({ default: { ttl: 900_000, limit: 5 } })   // <-- kept
async login(dto: LoginDto) { ... }
```

The route then has no `@Post`, so it vanishes from any decorator-based query:
documenting a decorator removes the endpoint from the graph. On the backend I was
indexing, exactly 1 of 95 HTTP endpoints was missing — the one whose throttle
policy carried an explanatory comment.

Comments are now transparent to the walk, the same way the anonymous tokens
(e.g. TS `export`) already were. Reuses the existing `is_comment_node()` helper.

## 2. Composite property values are truncated at the first internal comma

`json_extract_prop()` scanned a non-string value up to the first `,` and a string
value up to the first `"`, ignoring nesting and backslash escapes. Any array or
object property was therefore cut at its first **internal** comma:

```
decorators: ["@Roles('OWNER', 'ADMIN')","@Get()"]
projected as: ["@Roles('OWNER'
```

So `MATCH (m:Method) RETURN m.decorators` returns garbage for NestJS, Angular or
Spring handlers (any decorator with more than one argument), and strings holding
an escaped quote are cut short too. Values are now copied as balanced constructs,
honoring string state and escape pairs; the scalar and plain-string paths are
untouched.

Combined effect on the repo that surfaced this: HTTP endpoints reachable from the
graph went from **3/95 to 95/95**, with full `@Roles(...)` + verb + path.

## Tests

- `tests/test_cypher.c`: `cypher_exec_prop_array_with_internal_commas`,
  `cypher_exec_prop_string_with_escaped_quote`
- `tests/test_extraction.c`: `extract_ts_decorators_survive_interleaved_comment`

`test-runner cypher` → 152 passed · `test-runner extraction` → 214 passed
(ASan + UBSan build). The full `make -f Makefile.cbm test` run got 4635 tests in
with no failures before it was OOM-killed on my 6 GB machine at the
parallel-index test (sanitizer memory, environmental — not a test failure), so CI
coverage there would be appreciated.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -1797,6 +1797,18 @@ static int collect_modifier_decorators(CBMArena *a, TSNode modifiers, const char
     return idx;
 }
 
+/* Comments are NAMED nodes in tree-sitter, so a comment interleaved in a
+ * decorator run would end the walk and silently drop every decorator above it:
+ *
+ *   @Post('login')                    <-- lost
+ *   @HttpCode(HttpStatus.OK)          <-- lost
+ *   // why this route is throttled    <-- walk stopped here
+ *   @Throttle({ ... })                <-- kept
+ *   async login(...)
+ *
+ * Documenting a decorator must not make it disappear from the graph, so treat
+ * comments as transparent — like the anonymous tokens already skipped below.
+ * (is_comment_node() is defined above, near the docstring helpers.) */
 static const char **extract_decorators(CBMArena *a, TSNode node, const char *source,
                                        CBMLanguage lang, const CBMLangSpec *spec) {
     if (!spec->decorator_node_types || !spec->decorator_node_types[0]) {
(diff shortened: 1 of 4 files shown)


DeusData:
Thanks for both fixes. The decorator extraction change and the Cypher composite-property parser change are independent behavior changes with different regression surfaces. Please split them into two focused PRs, each linked to its issue and retaining its reproduce-first test. Both are valid 0.9.1-rc bug candidates once separated.

JhohanBustamante:
Thanks for the review — agreed, these are two independent behavior changes with different regression surfaces. Split into two focused PRs, each linked to its issue and keeping its reproduce-first test:

- Decorator extraction (comment interleaved in a decorator run) → #1097, fixes #1095
- Cypher composite/escaped property truncation → #1098, fixes #1096

Closing this one in favor of those. Each carries the same test as here.

PR #1087, opened 2026-07-14
Harden install.sh (exact-match checksum, --dir validation, wget fallback, Windows .exe)

selfish666420:
## What does this PR do?

This PR hardens the one-line installer by fixing edge cases that can cause incorrect validation or platform-specific install failures.

It updates `install.sh` to:
- validate `--dir <path>` properly and fail fast when the path argument is missing,
- match checksum entries by exact filename instead of substring matching,
- add `wget` fallback for checksum file download,
- handle Windows `.exe` artifact detection and target naming consistently.

Scope is intentionally installer-only.

## Checklist


DeusData:
Thanks for the installer fixes. The checksum lookup, Windows executable handling, and missing directory-argument validation are useful, but this PR mixes installer behavior, a Copilot instruction file, and a test-container documentation change. Please split the installer hardening into a focused bug PR with regression coverage and move the instruction/documentation work into a separate contribution. Installer and agent-instruction changes receive separate security review.

selfish666420:
> Thanks for the installer fixes. The checksum lookup, Windows executable handling, and missing directory-argument validation are useful, but this PR mixes installer behavior, a Copilot instruction file, and a test-container documentation change. Please split the installer hardening into a focused bug PR with regression coverage and move the instruction/documentation work into a separate contribution. Installer and agent-instruction changes receive separate security review.

Done. I split this PR to installer hardening only and removed the Copilot-instructions and test-container comment changes from this branch. The remaining diff is focused on install.sh reliability and cross-platform handling.

DeusData:
Reviewed and **approved to merge** — the installer hardening is a net security improvement. Specifically the checksum match going from a substring `grep` to an exact-filename match is a real win (a substring match could match a sibling/crafted filename in `checksums.txt`), and it's verified correct against the actual `release.yml` format; the `--dir` arg validation and the wget checksum-fetch fallback are clean, with no new download URL/domain/source and the post-install hook + codesign + `chmod` all unchanged.

Two housekeeping items so we can merge:

1. **This PR is still a *draft*, so CI never ran** (and it can't be merged as a draft). Could you click **Ready for review**? That'll trigger the required checks.
2. I've **corrected the title** on our side (it still said "add Copilot instructions", but commit `ff776faf` reverted that file — the net diff is `install.sh` only, which is the right, atomic scope). If you'd tweak the PR body to match, great.

If the required checks don't trigger after marking ready (an `install.sh`-only change can be skipped by path filters), a quick rebase onto current `main` will kick them off. Once it's green I'll merge with a merge commit (keeps your DCO sign-off).

Separately, I filed #1134 to make checksum verification *fail-closed* on a missing/malformed `checksums.txt` — that's a pre-existing gap, not something your PR introduced (your PR actually strengthens verification). Thanks for hardening this path @selfish666420!

selfish666420:
> Two housekeeping items so we can merge:
> 1. This PR is still a draft, so CI never ran (and it can't be merged as a draft). Could you click Ready for review?
> 2. ... If you'd tweak the PR body to match, great.

Done on both: I marked the PR ready for review and updated the PR body so it reflects the installer-only scope.

DeusData:
Closing this one, and I owe you a clear explanation because the reason is mostly bad timing rather than anything wrong with your work.

**First, the thing you could not have known: this PR targets `feat/agent-installations`, not `main`.** That branch is now **0 ahead / 375 behind** main — fully merged and retired. So merging this today would not change the installer any user actually downloads. It also explains the "no checks" you may have noticed: our PR workflow is filtered to `branches: [main]`, so a PR targeting any other branch runs no tests, no lint and no security gates at all. Not your fault, but worth knowing for next time — **please target `main`**.

**Your checksum fix was the right idea, and we arrived at the same one independently.** Replacing the substring `grep "$ARCHIVE"` with an exact awk field match — including stripping the `*` binary-mode marker — is exactly correct: a crafted filename containing the target name as a substring can no longer satisfy the lookup. Main's current verifier uses the same anchoring idea.

**What overtook it:** two days after you opened this, our installer rewrite landed on main (`0e00ef5`, then `afd67c5`). That version made checksum verification **mandatory and fail-closed** — a failed checksums download, a missing entry, or a missing `sha256sum`/`shasum` now all abort rather than silently skipping. It also added digest-format validation, duplicate-digest conflict detection, a size cap on the checksums file, and a unified curl/wget download helper with HTTPS pinning.

That last point is the one worth being straight about: your version keeps the three fail-open paths from its base, so verification is still *optional* — an attacker who can make `checksums.txt` 404 bypasses it entirely. That is precisely the hole the hardening set out to close, and it is why main's version had to go further. Your wget fallback is clean, incidentally — TLS verification is on by default and there is no `--no-check-certificate` anywhere.

**Two things you found are still real on main today, and I would genuinely like one of them as a fresh PR:**

1. **The `--dir <path>` parse is still naive** (`install.sh` around line 100). A trailing `--dir` with no value is silently ignored and the default directory is used, instead of erroring. Your fail-fast is still the right fix — please send it against `main` as a small standalone PR and I will review it quickly.
2. **The Windows branch still expects an un-suffixed binary after unzip** (`install.sh:244`, "binary not found after extraction"). If the Windows archive contains `codebase-memory-mcp.exe`, that path fails. I have flagged it separately — either the branch needs the `.exe` fallback you wrote, or we should stop advertising Windows in `detect_os` and point people at `install.ps1`.

Thank you also for narrowing the PR in your second commit — dropping the unrelated `.github` and Dockerfile changes made this much easier to review, and that kind of self-editing is appreciated.

PR #1094, opened 2026-07-15
fix(extract): snap docstring truncation to UTF-8 codepoint boundary (#1017)

### LINKED ISSUE #1017: docstring truncation splits multi-byte UTF-8 characters, producing invalid UTF-8 in nodes.properties (non-ASCII comments, e.g. Japanese)

harshitaajoshi:
Fixes #1017

## What does this PR do?

Docstring/comment truncation at `MAX_COMMENT_LEN` (500 bytes) in `extract_comment_text()` cuts at a raw byte offset, which can split a multi-byte UTF-8 character and leave an incomplete byte sequence in the stored `nodes.properties` JSON. This breaks any strict UTF-8 consumer reading the SQLite database (e.g. Python's `sqlite3` module).

This PR snaps the truncation point back to the last complete UTF-8 codepoint boundary in `extract_comment_text()`.

A regression test (`docstring_utf8_truncation_boundary_issue1017`) crafts a Go comment with 3-byte CJK characters straddling the 500-byte boundary and validates every byte of the extracted docstring is well-formed UTF-8.

## Checklist

- [x] Every commit is signed off (`git commit -s`) -- required, CI rejects unsigned commits ([DCO](https://developercertificate.org/), see [CONTRIBUTING.md](CONTRIBUTING.md))
  _ASan hangs on macOS Tahoe 26.5 (Apple Clang 17) for any binary, including a trivial `int main(){return 0;}`. All 6294 tests pass when built without sanitizers (`SANITIZE=""`)._

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -1168,10 +1168,14 @@ static bool is_comment_node(const char *kind) {
 }
 
 // Extract comment text, truncating to MAX_COMMENT_LEN.
+// #1017: snap the cut point back to a complete UTF-8 codepoint boundary.
 static char *extract_comment_text(CBMArena *a, TSNode node, const char *source) {
     char *text = cbm_node_text(a, node, source);
     if (text && strlen(text) > MAX_COMMENT_LEN) {
-        text[MAX_COMMENT_LEN] = '\0';
+        size_t cut = MAX_COMMENT_LEN;
+        while (cut > 0 && ((unsigned char)text[cut] & 0xC0) == 0x80)
+            cut--;
+        text[cut] = '\0';
     }
     return text;
 }
--- a/tests/test_extraction.c
+++ b/tests/test_extraction.c
@@ -3517,6 +3517,81 @@ TEST(extract_rust_test_attr_marks_is_test_issue855) {
     PASS();
 }
 
+/* #1017: docstring truncation at MAX_COMMENT_LEN (500 bytes) can split a
+ * multi-byte UTF-8 character, leaving an incomplete byte sequence.
+ * Craft a Go comment whose 498th-500th bytes are a 3-byte CJK character
+ * (U+6210 = 成 = e6 88 90).  The raw byte truncation at offset 500 lands
+ * one byte past the character start, splitting it.  After the fix the
+ * truncated string must end on a complete codepoint boundary. */
+TEST(docstring_utf8_truncation_boundary_issue1017) {
+    /* Build a comment: "// " (3 bytes) + 495 ASCII 'A' + "成成成" (9 bytes)
+     * Total comment text = 3 + 495 + 9 = 507 bytes.
+     * MAX_COMMENT_LEN = 500.  The first kanji (成 = e6 88 90) occupies
+     * offsets 498-500, so text[500] = '\0' keeps bytes 0-499: the lead
+     * byte 0xe6 plus one continuation 0x88 — an incomplete 2-of-3 sequence.
+     * Before fix: the truncated string ended with that broken pair. */
+    char comment[600];
+    int off = 0;
+    comment[off++] = '/';
+    comment[off++] = '/';
+    comment[off++] = ' ';
+    for (int i = 0; i < 495; i++)
+        comment[off++] = 'A';
+    /* U+6210 (成) = 0xe6 0x88 0x90 — 3-byte UTF-8 */
+    const char *kanji = "\xe6\x88\x90";
+    for (int k = 0; k < 3; k++) {
+        memcpy(comment + off, kanji, 3);
+        off += 3;
+    }
+    comment[off] = '\0';
+
+    /* Wrap in a Go function so the comment becomes the docstring. */
+    char src[800];
+    snprintf(src, sizeof(src), "package main\n\n%s\nfunc Compute() {}\n", comment);
+
+    CBMFileResult *r = extract(src, CBM_LANG_GO, "test", "main.go");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    ASSERT(has_def(r, "Function", "Compute"));
+
+    const char *doc = NULL;
+    for (int i = 0; i < r->defs.count; i++) {
+        if (strcmp(r->defs.items[i].name, "Compute") == 0) {
+            doc = r->defs.items[i].docstring;
+            break;
+        }
+    }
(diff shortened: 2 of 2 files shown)


DeusData:
Thanks for the reproduce-first fix. The extract_comment_text boundary change is narrow and memory-safe, and the new test is red on the old raw-byte truncation path. I have triaged this with #1017 as a high-priority 0.9.1-rc parsing bug. The diff adds no dependency, network, installer, or workflow surface.

Two review changes are needed:

1. The test's byte-offset explanation is inaccurate. After "// " plus 495 ASCII bytes, the first CJK codepoint starts at offset 498, so offset 500 is its final continuation byte. The pre-fix string retains the lead plus one continuation byte; it does not end with a lone lead byte from the third codepoint. Please correct the comments so the guard documents the actual red failure.
2. The regression exercises extract_comment_text only, but the PR also changes two independent fixed-buffer JSON serializers. Either add red-on-main boundary coverage for both the sequential and parallel serializer paths, including 2-, 3-, and 4-byte final codepoints, or remove those defense-in-depth hunks from this focused PR.

The existing full CI is green; these asks are about keeping every behavior change proved and reviewable.

harshitaajoshi:
Thanks for the review and for catching the byte-offset math. You're right that offset 500 is the final byte of the first kanji, so the pre-fix string ends with the lead byte plus one continuation (an incomplete 2-of-3 sequence), not a lone lead byte from the third codepoint. I've corrected the test comment to document the actual red failure accurately.

For the second point, I opted to keep this PR tightly focused on `extract_comment_text()` only, so I removed the `append_json_string` defense-in-depth hunks from `pass_definitions.c` and `pass_parallel.c`. Those can go in a follow-up PR with their own red-on-main boundary coverage for 2-, 3-, and 4-byte final codepoints across both serializer paths.

I force-pushed the amended commit and updated the PR description to reflect the narrower scope. Ready for another look when you get a chance.

Issue #1096, opened 2026-07-15
Cypher: composite/escaped property values truncated at the first internal comma/quote


JhohanBustamante:
## Summary

`json_extract_prop()` scans a non-string property value up to the first `,` and a
string value up to the first `"`, ignoring nesting and backslash escapes. Any
array/object property is therefore truncated at its first **internal** comma, and
any string containing an escaped quote is cut short.

## Reproduce

Given a method whose `decorators` property holds a composite value:

```
decorators: ["@Roles('OWNER', 'ADMIN')","@Get()"]
```

`MATCH (m:Method) RETURN m.decorators` projects:

```
["@Roles('OWNER'
```

i.e. the value is cut at the first comma inside `@Roles(...)`. The same happens
to any decorator taking more than one argument (NestJS, Angular, Spring
handlers), and to strings holding an escaped quote.

## Impact

Makes decorator/route/authz queries unusable on frameworks whose decorators take
multiple arguments. On the repo that surfaced this, HTTP endpoints reachable from
the graph went from **3/95 to 95/95** once values were returned whole (with full
`@Roles(...)` + verb + path).

## Expected

Composite values should be copied as balanced constructs, honoring string state
and backslash-escape pairs. The scalar and plain-string paths should be
unaffected.

## Environment

Found while indexing a NestJS + Angular monorepo.

Reproduce-first regression tests
(`cypher_exec_prop_array_with_internal_commas`,
`cypher_exec_prop_string_with_escaped_quote`) and the fix are ready; PR to
follow, linked to this issue.

DeusData:
Thanks for isolating both failure modes. This is triaged as a high-priority 0.9.1-rc Cypher correctness bug because silently truncated composite values can invalidate route and authorization queries. PR #1098 is the linked fix candidate; review will require red-on-main coverage for nested arrays/objects, escaped strings, and unchanged scalar behavior.

DeusData:
Fixed on main in 84837591 (PR #1098): Cypher property projection now copies the whole value for composite (array/object) and escaped-quote properties instead of truncating at the first internal comma or quote — a balanced, escape-aware copy in `json_extract_prop`. Thanks @JhohanBustamante!

PR #1097, opened 2026-07-15
fix(extract): keep decorators separated by a comment

### LINKED ISSUE #1095: TS/JS extraction: a comment interleaved in a decorator run drops every decorator above it
## Summary

`extract_decorators()` walks prev-siblings and breaks on the first **named**
tree-sitter node. Comments **are** named nodes, so a comment interleaved in a
decorator run ends the walk early and silently drops every decorator **above**
the comment.

## Reproduce

Index a TypeScript file with a comment between decorators:

```ts
@Post('login')                                     // dropped
@HttpCode(HttpStatus.OK)                           // dropped
// throttled per IP and per account
@Throttle({ default: { ttl: 900_000, limit: 5 } }) // kept
async login(dto: LoginDto) { ... }
```

The method ends up with only `@Throttle`. `@Post`/`@HttpCode` are gone, so the
route disappears from any decorator/route/authz query — documenting a decorator
with an explanatory comment removes the endpoint from the graph.

## Impact

On a real NestJS backend, exactly **1 of 95 HTTP endpoints** was missing from the
graph — the one whose throttle policy carried an explanatory comment.

## Expected

Comments should be transparent to the prev-sibling walk, the same way the
anonymous tokens (e.g. the TS `export` keyword) already are. The existing
`is_comment_node()` helper can be reused.

## Environment

Found while indexing a NestJS + Angular monorepo. Affects any framework whose
decorators/annotations can be interleaved with comments (NestJS, Angular, etc.).

A reproduce-first regression test
(`extract_ts_decorators_survive_interleaved_comment`) and the fix are ready; PR
to follow, linked to this issue.

JhohanBustamante:
Fixes #1095.

## A comment between decorators drops every decorator above it

`extract_decorators()` walks prev-siblings and breaks on the first NAMED node.
Comments **are** named nodes in tree-sitter, so a comment interleaved in a
decorator run ends the walk:

```ts
@Post('login')                     // <-- dropped
@HttpCode(HttpStatus.OK)           // <-- dropped
// throttled per IP and per account
@Throttle({ default: { ttl: 900_000, limit: 5 } })   // <-- kept
async login(dto: LoginDto) { ... }
```

The route then has no `@Post`, so it vanishes from any decorator-based query:
documenting a decorator removes the endpoint from the graph. On the backend I was
indexing, exactly 1 of 95 HTTP endpoints was missing — the one whose throttle
policy carried an explanatory comment.

Comments are now transparent to the walk, the same way the anonymous tokens
(e.g. TS `export`) already were. Reuses the existing `is_comment_node()` helper.

## Test

- `tests/test_extraction.c`: `extract_ts_decorators_survive_interleaved_comment`

`test-runner extraction` → 214 passed (ASan + UBSan build).

---

Split out of #1075 per review, as one focused PR linked to its issue. The Cypher
composite-property fix is now #1096 / its own PR.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -1886,6 +1886,18 @@ static int collect_modifier_decorators(CBMArena *a, TSNode modifiers, const char
     return idx;
 }
 
+/* Comments are NAMED nodes in tree-sitter, so a comment interleaved in a
+ * decorator run would end the walk and silently drop every decorator above it:
+ *
+ *   @Post('login')                    <-- lost
+ *   @HttpCode(HttpStatus.OK)          <-- lost
+ *   // why this route is throttled    <-- walk stopped here
+ *   @Throttle({ ... })                <-- kept
+ *   async login(...)
+ *
+ * Documenting a decorator must not make it disappear from the graph, so treat
+ * comments as transparent — like the anonymous tokens already skipped below.
+ * (is_comment_node() is defined above, near the docstring helpers.) */
 static const char **extract_decorators(CBMArena *a, TSNode node, const char *source,
                                        CBMLanguage lang, const CBMLangSpec *spec) {
     if (!spec->decorator_node_types || !spec->decorator_node_types[0]) {
@@ -1897,10 +1909,11 @@ static const char **extract_decorators(CBMArena *a, TSNode node, const char *sou
     while (!ts_node_is_null(prev)) {
         if (cbm_kind_in_set(prev, spec->decorator_node_types)) {
             count++;
-        } else if (ts_node_is_named(prev)) {
+        } else if (ts_node_is_named(prev) && !is_comment_node(ts_node_type(prev))) {
             /* A real preceding construct ends the decorator run. Anonymous
              * tokens (e.g. TS `export` between `@Decorator` and the
-             * `class_declaration`) are skipped so the decorator is still seen. */
+             * `class_declaration`) and comments are skipped so the decorator
+             * is still seen. */
             break;
         }
         prev = ts_node_prev_sibling(prev);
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for splitting this into a focused PR. The current head fb7a8082 is atomic, the regression is red on the previous named-comment break behavior, and both counting and collection walks apply the same boundary rule. Non-comment named siblings still terminate the run, so this does not broaden decorator association across declarations.

The full CI and DCO sets are green. The security review found no dependency, workflow, network, credential, prompt-injection, or parser-execution risk. I have triaged it as a high-priority 0.9.1-rc parsing fix. No contributor change is requested at this stage; the next step is final maintainer diff review.

DeusData:
Reviewed — this is a clean, well-scoped fix and the `extract_ts_decorators_survive_interleaved_comment` test is a proper reproduce-first guard (it genuinely fails on `main` without the symmetric `!is_comment_node` guard). The `extract_defs.c` change is exactly right and merges cleanly.

One thing before it can land: the branch base is ~29 commits behind `main`, and there's a mechanical conflict in `tests/test_extraction.c` — `main` has since appended more `RUN_TEST(...)` lines to the tail of `SUITE(extraction)` at the same spot your `RUN_TEST(extract_ts_decorators_survive_interleaved_comment)` lands. Could you rebase (or merge `main` into your branch) and re-anchor that one registration line? The code change itself is conflict-free. Once it's green I'll merge (with a merge commit, to keep your DCO sign-off intact). Thanks @JhohanBustamante!

DeusData:
Merged — thank you, and sorry it sat this long.

This is a small diff that fixes a real recall hole, and the diagnosis is exactly right. `extract_decorators` walks previous siblings and stops at the first *named* node — and in tree-sitter a comment **is** a named node, so a single comment sitting between decorators silently truncated the run. Everything above the comment simply vanished from the graph. For a decorated route or handler that means the documented endpoint stops being findable, which is the kind of loss that is hard to notice because nothing errors — you just get fewer results than the code deserves.

Two things in particular that made this easy to accept:

- You added the `is_comment_node()` guard to **both** walks — the counting pass and the collecting pass — symmetrically. That matters more than it looks: if only one side had skipped comments, the two passes would have disagreed on the count and the arrays would have misaligned. Getting that right is the difference between a fix and a new bug.
- You split it out of #1075 when asked, and kept it to one claim. The test is genuinely binding — the decorators above the comment are absent without the production change, so it fails on unpatched `main` for the right reason.

Verified against current `main` before merging: the truncation still reproduces there, so this was not overtaken by anything.

Thanks for the care on this one.

PR #1098, opened 2026-07-15
fix(cypher): return whole value for composite and escaped properties

### LINKED ISSUE #1096: Cypher: composite/escaped property values truncated at the first internal comma/quote

JhohanBustamante:
Fixes #1096.

## Composite property values are truncated at the first internal comma

`json_extract_prop()` scanned a non-string value up to the first `,` and a string
value up to the first `"`, ignoring nesting and backslash escapes. Any array or
object property was therefore cut at its first **internal** comma:

```
decorators: ["@Roles('OWNER', 'ADMIN')","@Get()"]
projected as: ["@Roles('OWNER'
```

So `MATCH (m:Method) RETURN m.decorators` returns garbage for NestJS, Angular or
Spring handlers (any decorator with more than one argument), and strings holding
an escaped quote are cut short too. Values are now copied as balanced constructs,
honoring string state and escape pairs; the scalar and plain-string paths are
untouched.

Combined effect on the repo that surfaced this: HTTP endpoints reachable from the
graph went from **3/95 to 95/95**, with full `@Roles(...)` + verb + path.

## Tests

- `tests/test_cypher.c`: `cypher_exec_prop_array_with_internal_commas`,
  `cypher_exec_prop_string_with_escaped_quote`

`test-runner cypher` → 152 passed (ASan + UBSan build).

---

Split out of #1075 per review, as one focused PR linked to its issue. The
decorator-extraction fix is now #1095 / its own PR.

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -2230,15 +2230,49 @@ static const char *json_extract_prop(const char *json, const char *key, char *bu
         p++;
     }
     if (*p == '"') {
-        /* String value */
+        /* String value — honor backslash escapes: without this, an embedded \"
+         * cuts the value short at the first escaped quote. */
         p++;
         size_t i = 0;
         while (*p && *p != '"' && i < buf_sz - SKIP_ONE) {
+            if (*p == '\\' && p[SKIP_ONE] && i + SKIP_ONE < buf_sz - SKIP_ONE) {
+                buf[i++] = *p++; /* keep the escape pair intact */
+            }
             buf[i++] = *p++;
         }
         buf[i] = '\0';
+    } else if (*p == '[' || *p == '{') {
+        /* Array/object value — copy the whole balanced construct. A scan-to-comma
+         * truncates at the first comma INSIDE the value: e.g. a decorators array
+         * ["@Roles('OWNER', 'ADMIN')","@Get()"] came back as ["@Roles('OWNER'. */
+        char open = *p;
+        char close = (open == '[') ? ']' : '}';
+        int depth = 0;
+        int in_str = 0;
+        size_t i = 0;
+        while (*p && i < buf_sz - SKIP_ONE) {
+            char c = *p;
+            if (in_str) {
+                if (c == '\\' && p[SKIP_ONE] && i + SKIP_ONE < buf_sz - SKIP_ONE) {
+                    buf[i++] = *p++; /* escape pair stays intact */
+                } else if (c == '"') {
+                    in_str = 0;
+                }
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for splitting this into a focused Cypher fix. The escaped-string path is bounded, and the composite scan is iterative rather than recursive, so this adds no stack-growth or unbounded-input risk. The scalar sibling assertion also guards the unchanged scalar path. I have triaged it as a high-priority 0.9.1-rc bug; full CI and DCO are green, and the security review found no malicious, prompt-injection, dependency, workflow, network, or credential surface.

Two coverage changes are needed before final review:

1. Add a red-on-main object case that contains mixed nested arrays/objects and an escaped quote inside the composite value. The current array test does not execute the new outer-object branch or prove mixed nesting/string state.
2. Add a boundary regression around the existing 511-byte projection limit to prove termination and deterministic truncation without reading past a trailing escape. Keep the current fixed-size contract in this PR; any proposal to return unlimited property values is a separate design change.

Please also narrow "whole value" in the PR description to balanced values within the existing projection cap. Malformed-JSON validation does not need to grow this focused fix, but malformed input must remain bounded and crash-free.

Issue #1126, opened 2026-07-16
Detect ABA edits during watcher indexing


lzl51230:
Related to #1113.

An SVN observation can stage working-copy state A, while the indexing pipeline reads and persists intermediate state B. If the working copy returns to A before acknowledgement, the watcher can acknowledge baseline A even though the graph contains B; the next poll then sees A as unchanged and leaves the graph stale.

Tie acknowledgement to the exact indexed input digest, or perform a post-publication observation check and schedule another update when it differs. Add a deterministic A-to-B-to-A concurrency test.

DeusData:
Tracked as a 0.9.2-rc graph-correctness blocker for #1113/#1127. The acknowledgement must correspond to the content the pipeline actually persisted; otherwise an A-to-B-to-A sequence can permanently bless a stale graph. The deterministic test should control the observation and publication barriers and prove a follow-up update is scheduled whenever the post-publication state differs.

Issue #1125, opened 2026-07-16
Avoid recursive stack exhaustion in SVN candidate traversal


lzl51230:
Related to #1113.

The SVN candidate walker recursively descends directories while retaining several path buffers per stack frame. A deeply nested working copy can exhaust the watcher thread stack even when the total path remains below platform limits.

Replace recursive traversal with a heap-backed worklist or otherwise bound stack usage. Add a deep-directory fixture that demonstrates traversal remains stable.

DeusData:
Tracked as a 0.9.2-rc stability blocker for #1113/#1127. Recursive traversal of an attacker- or repository-controlled directory tree must not consume unbounded watcher-thread stack. Prefer a bounded heap worklist and test a deep tree without depending on a platform stack crash as the assertion.

Issue #1124, opened 2026-07-16
Make SVN fingerprint opens race-safe


lzl51230:
Related to #1113.

The SVN fingerprint path checks run before ordinary `fopen`/`opendir` calls. A concurrent path swap can therefore replace a previously validated regular file or directory with a symlink, reparse point, special file, or path outside the working-copy root before it is opened.

Use no-follow, handle-relative opens where supported, or validate the opened handle identity and containment before reading. Add a regression test that swaps the candidate between validation and open.

DeusData:
Tracked as a 0.9.2-rc path-safety blocker for #1113/#1127. A pre-open path check is not sufficient when the subsequent open can race with a symlink, reparse point, special-file, or containment swap. The regression should exercise the swap at a deterministic test seam and validate the opened handle, not repeat the same pathname check.

Issue #1123, opened 2026-07-16
Enforce SVN fingerprint byte cap during reads


lzl51230:
Related to #1113.

SVN fingerprinting checks the file size before opening, but then reads until EOF. A file that grows concurrently can exceed the configured fingerprint byte cap and consume unbounded watcher-cycle time and I/O.

Track cumulative bytes while reading, stop at the existing cap, and validate size from the opened handle where possible. Add coverage for a file that grows after the initial metadata check.

DeusData:
Tracked as a 0.9.2-rc stability blocker for #1113/#1127. The byte cap must be enforced against bytes actually read from the opened handle, not only pre-open metadata. The RED test should grow the file after the initial check and prove both bounded I/O and deterministic fingerprint behavior at the cap.

PR #1127, opened 2026-07-16
feat(watcher): sync SVN working copies automatically

lzl51230:
## What does this PR do?

SVN working copies can now keep their graph synchronized after local working-copy changes on the watcher's next eligible poll, without requiring a manual full regeneration. The same adaptive cadence and indexing pipeline used by Git remain authoritative: ordinary changes use incremental indexing, while the existing safety threshold can still select a full rebuild.

The watcher observes recursive SVN XML status plus content fingerprints for indexable candidates, pins a trusted system `svn` executable, and acknowledges a new baseline only after explicit indexing success. Failed or busy updates remain pending for retry, Git keeps precedence in mixed roots, and successful background publication invalidates cached MCP stores on the request thread before the next query. Windows probes use isolated direct-argv stdout capture so MSYS2 paths and filename metacharacters never pass through `cmd.exe`, while retaining the existing restricted handle-inheritance model.

| Decision | Behavior |
| --- | --- |
| Source of truth | Current on-disk working-copy state, including eligible versioned and unversioned files |
| Scheduling | Existing completion-relative adaptive watcher interval; no SVN-specific setting |
| Pipeline routing | Existing incremental path with the existing full-rebuild safety fallback |
| Failure handling | Preserve the last acknowledged observation and retry later |

Session-settled decisions carried from planning: on-disk working-copy authority, SVN-aware polling, existing adaptive timing and pipeline routing, the system SVN CLI prerequisite, and Git precedence (all user-directed); using the fork's `main` and proceeding before maintainer feedback were carried out under flagged contribution-process concerns.

## Validation

- `make -f Makefile.cbm test-focused TEST_SUITES='security svn_state watcher integration'` (`135 passed`).
- Production build with `-Wall -Wextra -Werror` completed successfully.
- `make -j2 -f Makefile.cbm lint-ci` passed with CI-matched LLVM 20 `clang-format` plus cppcheck and the NOLINT whitelist.
- `scripts/security-audit.sh` and the no-skips policy passed.
- Real local SVN fixtures cover repeated tracked edits, unversioned additions, watcher retry behavior, Git/SVN precedence, incremental publication, and the existing full-route fallback.
- Windows-only coverage checks direct argv stdout capture, literal `%`/`&` arguments, nonzero exit propagation, MSYS2 path conversion, and an SVN working-copy path containing metacharacters.
- Browser testing was skipped because this PR changes no UI or browser route.

## Related and follow-up tracking

Related: #1113

- Related: #1116 - prevent a rewatch from indexing a retired root.
- Related: #1117 - make watcher publication refresh cached stores on Windows.
- Related: #1118 - enforce the supervisor gate through the public wrapper.
- Related: #1119 - keep primary updates flowing when an SVN external is broken.
- Related: #1120 - add explicit `svn update` and `svn commit` acceptance coverage.
- Related: #1121 - expand the SVN parser and path-safety matrix.
- Related: #1122 - cover SVN watcher recovery and lifecycle transitions.
- Related: #1123 - enforce the fingerprint byte cap during reads.
- Related: #1124 - make SVN fingerprint file opens race-safe.
- Related: #1125 - avoid recursive stack exhaustion in candidate traversal.
- Related: #1126 - detect ABA edits during watcher indexing.

## Checklist

- [x] Tests pass locally.
- [x] CI-matched lint passes locally.
- [x] New behavior is covered by focused and integration tests.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -81,7 +81,11 @@ jobs:
 
       - name: Install deps (Ubuntu)
         if: startsWith(matrix.os, 'ubuntu')
-        run: sudo apt-get update && sudo apt-get install -y zlib1g-dev
+        run: sudo apt-get update && sudo apt-get install -y subversion zlib1g-dev
+
+      - name: Install deps (macOS)
+        if: startsWith(matrix.os, 'macos')
+        run: brew install subversion
 
       - name: Test
         run: scripts/test.sh CC=${{ matrix.cc }} CXX=${{ matrix.cxx }}
@@ -95,7 +99,7 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Install deps (Ubuntu)
-        run: sudo apt-get update && sudo apt-get install -y clang zlib1g-dev
+        run: sudo apt-get update && sudo apt-get install -y clang subversion zlib1g-dev
 
       - name: ThreadSanitizer tests
         run: make -f Makefile.cbm test-tsan CC=clang CXX=clang++
@@ -121,6 +125,7 @@ jobs:
             mingw-w64-clang-${{ matrix.pkg }}-zlib
             make
             git
+            subversion
 
       - name: Test
         shell: msys2 {0}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ src/
     pass_*.c        Individual pipeline passes (definitions, calls, usages, etc.)
     httplink.c      HTTP route extraction (Go/Express/Laravel/Ktor/Python)
   discover/         File discovery with gitignore support
-  watcher/          Git-based background auto-sync
+  watcher/          Git/SVN working-copy background auto-sync
   cli/              CLI subcommands (install, update, uninstall, config)
   ui/               Graph visualization HTTP server (first-party httpd)
 internal/cbm/       Tree-sitter AST extraction (64 languages, vendored C grammars)
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -65,7 +65,8 @@ CXXFLAGS_PROD = $(CXXFLAGS_COMMON) -O2
 SANITIZE = -fsanitize=address,undefined -fno-omit-frame-pointer
 EDITOR_TEST_DEFINES = -DCBM_JSON_LIKE_ENABLE_TEST_API=1 \
 	-DCBM_TOML_EDIT_ENABLE_TEST_API=1 -DCBM_YAML_ENABLE_TEST_API=1 \
-	-DCBM_TEXT_EDIT_ENABLE_TEST_API=1 -DCBM_CLI_ENABLE_TEST_API=1
+	-DCBM_TEXT_EDIT_ENABLE_TEST_API=1 -DCBM_CLI_ENABLE_TEST_API=1 \
+	-DCBM_SVN_STATE_ENABLE_TEST_API=1
 CFLAGS_TEST = $(CFLAGS_COMMON) $(EDITOR_TEST_DEFINES) -g -O1 $(SANITIZE)
 CXXFLAGS_TEST = $(CXXFLAGS_COMMON) -g -O1 $(SANITIZE)
 
@@ -236,7 +237,7 @@ UNIXCODER_BLOB_SRC = vendored/nomic/code_vectors_blob.S
 TRACES_SRCS = src/traces/traces.c
 
 # Watcher module (new)
-WATCHER_SRCS = src/watcher/watcher.c
+WATCHER_SRCS = src/watcher/watcher.c src/watcher/svn_state.c
 
 # Git context module (new)
 GIT_SRCS = src/git/git_context.c
@@ -359,7 +360,7 @@ TEST_GRAPH_BUFFER_SRCS = tests/test_graph_buffer.c
 
 TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_index_resilience.c
 
-TEST_WATCHER_SRCS = tests/test_watcher.c
+TEST_WATCHER_SRCS = tests/test_watcher.c tests/test_svn_state.c
 
 TEST_LZ4_SRCS = tests/test_lz4.c
 TEST_ZSTD_SRCS = tests/test_zstd.c
--- a/README.md
+++ b/README.md
@@ -127,9 +127,11 @@ Enable automatic indexing on MCP session start:
 codebase-memory-mcp config set auto_index true
 ```
 
-When enabled, new projects are indexed automatically on first connection. Previously-indexed projects are registered with the background watcher for ongoing git-based change detection. Configurable file limit: `config set auto_index_limit 50000`.
+When enabled, new projects are indexed automatically on first connection. Previously-indexed projects are registered with the background watcher for ongoing Git or SVN working-copy change detection. Configurable file limit: `config set auto_index_limit 50000`.
 
 Watcher registration is controlled separately by `auto_watch` (default `true`). Set `config set auto_watch false` to keep a session from registering its project with the background watcher — useful when working across many projects and you want each session contained to explicit indexing.
+
+SVN auto-watch requires a trusted system `svn` executable available on `PATH`; project-local clients are not used. Observation runs locally against the working copy and does not query SVN history or add an SVN-specific interval. Git and SVN projects share the watcher's existing adaptive polling cadence and the same indexing pipeline: ordinary changes take the incremental route, while the pipeline may perform a full rebuild when its existing safety threshold requires one. If SVN observation is unavailable or temporarily fails, the current graph remains available and the watcher retries on a later poll.
 
 ### Keeping Up to Date
 
@@ -191,7 +193,7 @@ Removes owned agent config entries, skills, hooks, instructions, and the install
 
 ### Distribution & operation
 - **Single static binary, zero infrastructure**: SQLite-backed, persists to `~/.cache/codebase-memory-mcp/`
-- **Auto-sync**: Background watcher detects file changes and re-indexes automatically
+- **Auto-sync**: Background watcher detects Git and SVN working-copy changes and re-indexes automatically
 - **Route nodes**: REST endpoints are first-class graph entities
 - **CLI mode**: `codebase-memory-mcp cli search_graph '{"project": "my-project", "name_pattern": ".*Handler.*"}'`
 - **Available on**: npm, PyPI, Homebrew, Scoop, Winget, Chocolatey, AUR, `go install`
@@ -598,7 +600,7 @@ See [docs/cbmignore.md](docs/cbmignore.md) for the full `.cbmignore` how-to: syn
 codebase-memory-mcp config list                          # show all settings
 codebase-memory-mcp config set auto_index true           # auto-index on session start
 codebase-memory-mcp config set auto_index_limit 50000    # max files for auto-index
-codebase-memory-mcp config set auto_watch false          # don't register background git watcher (default: true)
+codebase-memory-mcp config set auto_watch false          # don't register background Git/SVN watcher (default: true)
 codebase-memory-mcp config reset auto_index              # reset to default
 ```
 
@@ -712,7 +714,7 @@ src/
   pipeline/           Multi-pass indexing (structure → definitions → calls → HTTP links → config → tests)
   cypher/             Cypher query lexer, parser, planner, executor
   discover/           File discovery (.gitignore, .cbmignore, symlink handling)
-  watcher/            Background auto-sync (git polling, adaptive intervals)
+  watcher/            Background auto-sync (Git/SVN polling, adaptive intervals)
   traces/             Runtime trace ingestion
   ui/                 Embedded HTTP server + 3D graph visualization
   foundation/         Platform abstractions (threads, filesystem, logging, memory)
--- a/docs/plans/2026-07-16-001-feat-svn-auto-watch-graph-sync-plan.md
+++ b/docs/plans/2026-07-16-001-feat-svn-auto-watch-graph-sync-plan.md
@@ -1,0 +1,384 @@
+---
+title: SVN Auto-Watch Graph Sync - Plan
+type: feat
+date: 2026-07-16
+topic: svn-auto-watch-graph-sync
+artifact_contract: ce-unified-plan/v1
+artifact_readiness: implementation-ready
+product_contract_source: ce-brainstorm
+execution: code
+deepened: 2026-07-16
+---
+
+# SVN Auto-Watch Graph Sync - Plan
+
+## Goal Capsule
+
+- **Objective:** Keep an indexed graph synchronized with an SVN working copy without requiring a manual full regeneration after local edits, `svn update`, or `svn commit`.
+- **Product authority:** This Product Contract defines SVN watcher behavior. Existing discovery rules and pipeline routing remain authoritative for file eligibility and incremental-versus-full indexing decisions.
+- **Execution profile:** Deep cross-platform code change, implemented test-first in dependency order U1 through U4.
+- **Stop conditions:** Stop if SVN polling cannot preserve the acknowledged-versus-pending baseline contract, if Git behavior regresses, or if upstream maintainers reject the design in the required tracking issue.
+- **Tail ownership:** LFG owns implementation, review fixes, DCO-compliant English commits, the fork feature branch, the upstream PR, and CI stabilization. The fork's `main` synchronizes only after upstream merge.
+- **Open blocker:** The upstream project requires a tracking issue and maintainer feedback before feature implementation begins. No product-scope question remains open.
+- **Tracking issue:** `DeusData/codebase-memory-mcp#1113` requests the required design feedback before implementation.
+
+---
+
+## Product Contract
+
+### Summary
+
+Extend background auto-watch to pure SVN working copies. Any distinct change to indexable working-copy files triggers the existing graph update pipeline on the next eligible watcher poll so the graph follows the current on-disk development state.
+
+### Problem Frame
+
+SVN users currently receive no background change detection because the watcher only recognizes Git repositories. After every SVN update or commit, they must request a full graph regeneration to restore synchronization, even when only a small set of files changed.
+
+### Actors
+
+- A1. **SVN developer:** Edits, adds, removes, renames, resolves conflicts in, commits, or updates a local SVN working copy.
+- A2. **Background watcher and indexing pipeline:** Detects a new working-copy state, schedules reindexing, and updates the graph through the existing routing rules.
+
+### Key Decisions
+
+- **The on-disk working copy is authoritative** (session-settled: user-directed — chosen over operation-boundary-only or tracked-file-only synchronization: the graph must follow local development state before and after SVN operations). This includes versioned and unversioned source files, deletions, renames, and conflicted contents that remain eligible under existing discovery rules.
+- **Scope is limited to auto-watch graph synchronization** (session-settled: user-directed — chosen over broader SVN feature parity: automatic graph freshness solves the observed manual-regeneration cost without adding unrelated history or analysis work).
+- **Existing timing and pipeline routing remain authoritative** (session-settled: user-directed — chosen over a fixed five-second target or forced incremental routing: SVN should inherit the watcher's adaptive poll timing and the pipeline's safety fallback to full indexing).
+- **SVN-aware polling is the change-detection mechanism** (session-settled: user-directed — chosen over filesystem snapshots or dual-signal detection: it extends the current VCS-aware watcher with the smallest ongoing maintenance surface).
+- **The system SVN CLI is a runtime prerequisite on all supported platforms** (session-settled: user-directed — chosen over a bundled client or CLI-free implementation: this matches the current external VCS command pattern and preserves cross-platform scope).
+- **Git takes precedence in mixed roots** (session-settled: user-directed — chosen over SVN precedence or simultaneous dual watching: one watcher strategy per project avoids duplicate reindex events).
+- **Contribution work uses a fork feature branch** (session-settled: user-directed — chosen over pushing unreviewed feature commits to the fork's `main`: the fork main branch should synchronize with upstream only after the upstream PR merges).
+
+### Requirements
+
+**Registration and state detection**
+
+- R1. With `auto_watch` enabled, an indexed pure SVN working copy with a callable system `svn` command must register for background change detection.
+- R2. A project root containing both `.git` and `.svn` must retain the existing Git watcher strategy and must not run a second SVN strategy.
+- R3. SVN state detection must maintain a working-copy baseline that represents repository revision state and local working-copy changes without repeatedly triggering on an unchanged state.
+- R4. The SVN state must distinguish distinct changes involving versioned files, unversioned source files, deletions, renames, and conflict states.
+
+**Graph synchronization**
+
+- R5. A distinct SVN working-copy state must trigger the existing indexing pipeline on the next eligible adaptive watcher poll.
+- R6. After a successful update, the graph must reflect the current set and contents of on-disk files allowed by existing discovery rules.
+- R7. Changes confined to `.svn` metadata or other non-indexable files must not add those files to the graph.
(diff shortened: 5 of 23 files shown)


DeusData:
Thanks for the substantial implementation work and for documenting the unresolved cases. The exact head cannot be accepted or expanded in place.

**Product and supply-chain boundary**

This introduces a conditional runtime dependency on a system Apache Subversion executable while the project currently promises a zero-runtime-dependency binary. It also changes `.github/workflows/_test.yml` so apt, Homebrew, and MSYS2 install Subversion across existing native and TSan jobs. That product/CI decision has not been approved; please do not add further commits around it until maintainers decide whether this dependency belongs at all.

**Current correctness and security blockers**

1. The validation-then-open fingerprint path has a TOCTOU containment race (#1124): a symlink, reparse point, special file, or outside-root target can replace the validated path before open.
2. Fingerprint reads do not enforce the byte cap against bytes actually read from a growing file (#1123).
3. Recursive candidate traversal can exhaust the watcher thread stack (#1125).
4. The SVN subprocess has no bounded timeout/cancellation path, so a stuck client can block polling and shutdown.
5. Observation acknowledgement can bless stale graph state after an A-to-B-to-A edit race (#1126).
6. A snapshotted retired watch can still invoke the callback after root replacement (#1116).
7. Windows publication fails with a pre-warmed cached SQLite handle (#1117); the current Windows full-suite failure is therefore a real product failure, not a flake.
8. One broken external can block primary-working-copy updates (#1119), contrary to the PR's stated behavior.
9. The recursive status/fingerprint work has no aggregate work budget or large-tree characterization.

**Scope and repository hygiene**

The PR combines general incremental dump-error propagation, watcher lifetime changes, MCP store invalidation, a new Windows process API, build/CI changes, and the SVN feature. Those need focused prerequisite PRs with their own RED guards. Replace the raw `fopen()` added in `tests/test_integration.c` with `cbm_fopen()`. Remove `docs/plans/2026-07-16-001-feat-svn-auto-watch-graph-sync-plan.md`; session execution plans and authority assertions are not repository documentation.

The current security review found no credential, hosted-service, telemetry, vendored-code, or data-exfiltration addition. The blocker is the new local executable/CI dependency plus the unresolved containment and lifecycle defects. Maintainers will decide the product direction separately; this head is not mergeable even if the remaining CI job turns green.

lzl51230:
> please do not add further commits around it until maintainers decide whether this dependency belongs at all.

Acknowledged. I have stopped pushing commits to this head and will leave it unchanged while maintainers decide whether a system SVN dependency fits the product and supply-chain boundary.

If that direction is accepted, I will first split the general prerequisites into focused PRs with RED regression guards, then address the full correctness, lifecycle, timeout, work-budget, and repository-hygiene list before returning to the SVN feature. I will not treat a green CI run as making the current head mergeable.

DeusData:
Reviewed in full — diff read end to end, every `svn` invocation located. Two things to say, and I want to lead with the first.

**The security engineering here is genuinely excellent.** Specifically: the executable is pinned once at registration by scanning `PATH` for absolute entries only, canonicalizing, and then **rejecting any candidate that resolves under the watched repo root or the CWD** — which blocks a malicious repo shipping its own `svn`. On Windows you added `cbm_popen_argv()` and spawn a direct argv rather than going through `cmd.exe`, with literal `%VAR%` and `&` metacharacter tests to prove it. On POSIX both executable and root pass `cbm_validate_shell_arg` before a double-quoted string with a `--` separator and an `@` peg-escape suffix. That is precisely the level of paranoia this codebase wants, and it is not common.

I also want to confirm something readers of the title might assume otherwise: **this never mutates the working copy.** The only production invocation is `svn status --xml --verbose --no-ignore --depth infinity --non-interactive`; `checkout` and `commit` appear solely in `tests/svn_test_helpers.h`. "Sync" here means syncing the *graph* to the working copy, not the other way round. I checked because the title reads both ways.

And the honesty in the PR body about proceeding before maintainer feedback made this review easier, not harder. Thank you for that.

**Why it is not a review-level decision.** Two things in here are the maintainer's call, so I have escalated rather than deciding:

1. **Do we want SVN as a second watcher-owned VCS at all?** That is a direction question about the project's scope, not about your implementation quality.
2. **The bundled MCP store-update gate.** `cbm_mcp_server_begin/end_store_update` changes MCP-core locking for *all* VCSs and *all* platforms — it is the fix for the separate staleness issue riding along inside an SVN feature, which hides its blast radius. Concretely, the gate is taken at the top of `watcher_index_fn` and released at the end, so it is **held across the entire background reindex**; on Windows it also invalidates the cached store at begin and keeps new query handles from opening until publication completes. On a large repo that can gate MCP queries for the full minutes-long sync. That needs measuring and its own review, independently of the SVN question.

**Three things to do regardless of how the direction call goes:**

- **Split the store-update gate into its own PR.** It stands alone, it affects git projects, and it deserves its own before/after numbers on Windows.
- **Drop the committed planning document.** It creates a new `docs/plans/` directory that does not exist on main; session artifacts should not land in the repo.
- **Rebase.** The PR is `CONFLICTING` against current main — sharding, the single-binary change and daemon work have all landed since 16 July, and your green CI is against that older base.

One more thing the maintainer will want to weigh consciously: the PR installs `subversion` on the Ubuntu, macOS and Windows test legs, which is a package on every test job (the `brew install` step in particular is new cost and new flake surface), and `scripts/security-allowlist.txt` gains two entries for your own `popen` usage. Both are reasonable given the code — they simply need to be accepted deliberately rather than inherited.

I will come back to you as soon as I have the direction answer.

DeusData:
**Direction answer, as promised: yes, we will take SVN support** — with two conditions that change the shape of the feature, plus the items already on the list.

I want to be straight about the reasoning, because "yes" here was not automatic. The watcher is 1,440 lines whose entire change-detection strategy is git — `watcher_git_run`, `rev-parse --git-dir`, `rev-parse HEAD`, `status --porcelain`, `ls-files`, submodule handling. A second VCS is not a feature we add once; it is a surface we maintain forever, in a project that deliberately stays lean. What tipped it is that your implementation is contained and unusually careful, and that SVN is still real in exactly the enterprise environments where this tool is most useful.

**Condition 1 — it must be opt-in, defaulting off.**

Today the SVN strategy is auto-selected whenever a watched root lacks git metadata and contains a `.svn` directory, with no config key anywhere. That means the presence of a `.svn` directory somewhere in a watched tree causes us to start spawning the system `svn` binary against it — a behaviour change nobody asked for, on machines where SVN may be installed for unrelated reasons. Please put it behind an explicit config key (the `CONFIG_KEYS` table in `src/cli/cli.c` is where it belongs so `config list`/`config set` can see it), defaulting to **off**, and document it in `docs/CONFIGURATION.md`.

I realise auto-detection is friendlier for SVN users and that opting in is a small tax on them. I would still rather charge that tax than change what the watcher does on machines that never asked for it.

**Condition 2 — the subprocess must be bounded.**

There is no timeout and no cancellation path around the `svn` child. `svn status --depth infinity` on a large or pathological working copy — or one on a stale network mount — can block indefinitely, and the watcher thread has no way out. Please add a timeout with cancellation, and treat expiry the same way you already treat a parse failure: `CBM_SVN_PROBE_UNCERTAIN` rather than a hard error. Your existing uncertain-state handling makes this a small change, which is exactly why it should not be left out.

**One correction to the shell-argument validation.** Both the executable and the status root go through `cbm_validate_shell_arg`, but `src/foundation/str_util.h` documents `cbm_validate_shell_path_arg` as the one to use — its comment says, in as many words, to use it and not `cbm_validate_shell_arg` for every path that crosses into a shell command, so the shell-out sites cannot drift apart again. On POSIX the two are behaviourally identical today and the Windows branch uses no shell, so this is not exploitable at this head. It is drift from a convention that exists precisely to stop these sites diverging, and the fix is one identifier.

**And a small one:** `tests/test_integration.c` adds a raw `fopen(path, "w")`. Repo rule is `cbm_fopen()` — your production code in `svn_state.c` already gets this right with `cbm_fopen(absolute, "rb")`, so it is just the test that slipped.

**Still on the list from last time, unchanged:** split the `cbm_mcp_server_begin/end_store_update` gate into its own PR with before/after numbers on Windows (it changes MCP-core locking for every VCS and every platform, and held across a full background reindex it can gate queries for minutes); drop the committed `docs/plans/` document; and rebase, since the branch is `CONFLICTING` and your green CI is against a mid-July base.

**Two costs I am accepting deliberately rather than inheriting**, so they are on the record: installing `subversion` on the Ubuntu, macOS and Windows test legs is real time and new flake surface on every test job, and `scripts/security-allowlist.txt` gains two entries for your `popen` site. Both are justified by the code. With the feature opt-in and defaulting off, I would also like the SVN tests to keep running unconditionally — the config gate should govern the product's behaviour, not our coverage of it.

Thank you for your patience through a long direction call, and for the security work in here. Pinning the executable at registration and rejecting any candidate resolving under the watched repo root is the kind of thing most contributors would not think to do, and it is the reason this got a "yes" rather than a "not worth the surface."

DeusData:
@lzl51230 — no new asks, and no pressure on timing. This is a heads-up that the ground moved under your rework since my 2026-08-20 answer, so you aim at the right place rather than discover it mid-rebase.

**The watcher integration point you patch no longer exists.** Your diff wires `cbm_watcher_new` and the index callback in `src/main.c`. On today's `main` (`b0a3ca28`) `main.c` does not construct the watcher at all — it moved to `src/daemon/host.c:598`, inside `host_state_prepare`, with a different callback (`host_watcher_index`) and different user_data (the host, not a standalone context). So that part is a redesign against the daemon architecture, not a conflict to resolve. I would rather you knew that before spending an evening on a rebase.

**And Condition 1 now has a neighbour.** #1105 merged today (`b0a3ca28`) and added a `watcher_enabled` config key, default **true**, which gates whether the watcher is built at all — `host->watcher` can now legitimately be NULL. Your SVN key should compose with it as a second-level gate: watcher off means no SVN activity at all, watcher on plus SVN key off means git-only. It also gives you a worked example of exactly the shape Condition 1 asks for — `CONFIG_KEYS` row, `cbm_config_get_bool` with an explicit default, docs entry, and a process-level test that asserts the subsystem does **not** start when the key is off, with a positive control so gate inversion fails too. That test is worth copying; it is the part that makes a config key more than a claim.

Good news on the rest: I re-verified your security work against the current tree and it stands. The executable pinning, the rejection of an `svn` binary under the watched root or CWD, and the Windows direct-argv path with no `cmd.exe` are all still correct, and the read-only nature of `svn status --xml --verbose` is exactly as you described. Nothing in the direction answer has changed — SVN support is still a yes.

Conditions 1 and 2 (opt-in key defaulting off; bounded subprocess with cancellation → `CBM_SVN_PROBE_UNCERTAIN`) stand as written, along with the split of the MCP store gate into its own PR, dropping the committed plan document, and the July-16 correctness list.

Take the time you need — you stopped pushing the moment I asked you to back in July, which was the right call and is why this thread has stayed easy to follow.

DeusData:
@lzl51230 — thank you for this, and for doing exactly what I asked: you stopped pushing the day I requested it and then waited while I took five weeks to answer. That patience deserves a concrete update rather than another silence.

**First, something is landing now, with credit to you.** Buried in this PR is a ~17-line change to `pipeline_incremental.c` that makes `dump_and_persist` return its status, so a failed SQLite dump stops reporting success. That is a **silent data-loss bug affecting every git user today**, it has nothing to do with SVN, and it stands entirely alone. I'm shipping it as its own small PR with `Co-authored-by:` credit to you rather than leaving it behind a large feature review. Thank you for catching it.

**Second, the direction answer from 2026-08-20 stands — SVN is still a yes.** But the target moved underneath you again, and you should hear that from me rather than discover it in a rebase:

- The watcher is no longer built in `src/main.c`; it moved to `src/daemon/host.c:598`.
- `defer_state_free`, which this PR adds, **already exists** on main at `watcher.c:829` and is already called at `:1059`.
- Since your branch point, `mcp.c` has taken ~90 commits and `main.c` ~36.

**Third, one correction I owe you.** My 08-20 comment said a `.svn` directory "somewhere in a watched tree" triggers the SVN path. That was wrong — `vcs_metadata_exists` checks `<root>/.svn` only, and a root `.git` wins first. The blast radius is smaller than I described. Condition 1 (opt-in, default off) still stands on its own merits, since it does still spawn `svn` on machines that never asked for it.

**What the remaining work actually is**, so it's concrete:

1. **Condition 2 is a dropped contract, not a missing feature.** `watcher_git_run` (`src/watcher/watcher.c:322-361`) already runs git through `cbm_subprocess_spawn` with a wall deadline, `cancel_grace_ms` and an `output_limit`. The SVN path uses bare `cbm_popen` with none of the four and blocks in `fgetc` until EOF. Use `watcher_git_run` as the template.
2. **The unbounded recursion at `svn_state.c:702-745` is a merge blocker** and wants a RED guard. It's reachable from ordinary repo content, not just hostile input: `--no-ignore` reports unversioned and ignored *directories*, and `fingerprint_tree` descends them itself. ~8.3 KB frames against a 4096-byte path budget permits ~2040 levels on an 8 MB stack.
3. **The MCP store gate needs redesign, not rebase.** It gates a single `cbm_mcp_server_t`, but the daemon keeps a per-session list (`application.c:2249`), so one gate cannot invalidate the others. It also avoids self-deadlock only because Windows `cbm_mutex_t` is a re-entrant `CRITICAL_SECTION` while POSIX is not — a quiet platform dependency. And as written it makes every MCP query on Windows block for the full duration of a background reindex, **including for git projects**. I'd rather this were split out entirely.
4. **A design answer on per-poll cost.** Unlike git's cheap metadata check, this content-hashes the full bytes of every candidate on every poll, uncapped, with `--no-ignore` pulling ignored subtrees into that set. On the multi-GB enterprise working copies that are the whole point of the feature, that's continuous full-tree I/O every 5-60s. Dropping `--no-ignore`, capping depth and capping bytes read may be enough — but I'd like your view.
5. **The local ladder needs subversion too.** `th_svn_fixture_init` returns -1 when `svn` is absent and callers `FAIL()` rather than skip — so the three-OS local ladder, which is the primary gate here, would go hard red. No `test-infrastructure/` image installs subversion and the Windows VM provisioning is untouched; this PR only edits `.github/workflows/_test.yml`. Either fixtures must **skip**, or provisioning has to come with it.

**Credit where it's due:** the security engineering is genuinely good and I want that on the record. One production `svn` invocation, pre-validated on POSIX and direct-argv on Windows so `cmd.exe` is never involved; `--` separator present; the trailing `@` correctly neutralising peg-revision syntax; and a PATH-hijack defence that rejects any `svn` resolving under the watched repo or CWD. The parser is bounds-checked throughout, and there is not a single `sleep` in the test suite — every test uses the `cbm_watcher_touch()` seam and asserts exact call counts. That is the shape I want tests to have.

Given points 3 and 5, I think this is cleanest as a fresh PR against the daemon with the MCP gate left out. **Do you want to carry that?** A straight "no" or silence is a completely fine answer after the wait I've put you through — say so and I'll take it from here with you credited. Either way the `dump_and_persist` fix ships this week.

DeusData:
Correction to my previous comment, @lzl51230 — I was wrong about the `dump_and_persist` fix, and you should have the accurate version rather than the flattering one.

I said it was "a silent data-loss bug affecting every git user today" and that I'd ship it this week. **Both halves are wrong**, and I checked properly only after saying it.

**It was already fixed, independently, on 2026-07-16** — commit `0e00ef57` ("feat: coordinate concurrent CBM sessions") made the identical three edits: the signature became `static int`, the dump failure returns early, and the caller captures and returns `persist_rc`. That landed in parallel with your PR, so your hunk is a no-op against current main (`pipeline_incremental.c:1463`, `:1487`, `:2851`, `:2862`).

**And the severity I claimed was overstated regardless.** `dump_and_persist` is now test-only — its own comment says "Publish the test-only legacy partial result", and reaching it requires `force_legacy_partial`, which is set only by a test-API function behind `#if defined(CBM_INCREMENTAL_TEST_API)`. The production incremental route is `run_closure_delta`, which returns directly at `:2543` and never falls through. So even before `0e00ef57` this was not on the path a normal user takes.

**What stands: you identified a real bug by reading the code.** Discarding a persistence status is exactly the class of defect worth catching, you spotted it in a file you were only passing through, and you were right that it needed fixing. It simply got fixed by someone else first, which is luck of timing and no reflection on the finding.

Practical consequence: when the SVN work is re-briefed against the daemon, **drop this hunk** rather than carrying it — it would now be an empty diff.

Sorry for the noise. I'd rather correct it the same day than let you discover it when the "shipping this week" PR never appeared.

DeusData:
Apologies first: this has been open since **16 July with no comment from us at all**. Seven weeks of silence on 2,723 lines of work is not acceptable and I am sorry. This note is not a review verdict — it is to tell you where it actually stands.

**This one needs a more careful review than we have given it, and that review has not started.** It is not queued behind a small backlog; nobody has looked at it properly yet. I would rather say that plainly than leave you guessing.

Four things make it heavier than its line count suggests, and are why it cannot be waved through:

**1. It adds its own entries to `scripts/security-allowlist.txt`.** This is the one I need to flag now, because it is policy rather than opinion and it is actionable for you:

```
src/watcher/svn_state.c:cbm_popen:recursive local SVN status XML (executable pinned; args validated)
src/watcher/svn_state.c:popen:via cbm_popen wrapper call
```

**A contribution may not supply its own exemption from a security gate**, however accurate the wording — and yours reads accurately. The allowlist is maintainer-only precisely because an honest entry and a dishonest one look identical in a diff, and every entry on `main` today was added by a maintainer. If this lands, we add those lines ourselves after reviewing the `popen` sites, not as part of your diff. Please drop them from the PR.

**2. It edits `.github/workflows/_test.yml`** to `apt-get install subversion` and `brew install subversion` across three legs. CI workflow changes are the same maintainer-only class, and this one also makes `svn` a build-time requirement on every runner.

**3. It introduces a new external binary dependency** — `svn` — for end users as well as CI. That is a project-direction question, not a code-review question, and it belongs to the maintainer.

**4. The `popen` surface itself deserves real scrutiny.** Your description says the executable is pinned and arguments validated, and the Windows path uses direct-argv capture so nothing passes through `cmd.exe`. That is the right instinct and it is exactly the sort of claim that has to be verified line by line rather than accepted, which is a chunk of careful work rather than a skim.

It is also currently **conflicting with `main`**, so it needs a rebase regardless.

I am not asking you to do that rebase yet — it would be unfair to ask for work before we have committed to reviewing it. What I would suggest, if you want to move it forward: drop the `security-allowlist.txt` and `_test.yml` changes, since neither can be accepted from a contribution in any case. That makes the remaining diff a straight feature review rather than a policy question tangled with one.

Thank you for your patience, and again, sorry for the silence.

DeusData:
@lzl51230 — first I have to clean up our side of this thread, because the last thing you heard from us was wrong.

## The 2 September comment contradicted everything before it. Please disregard most of it.

It said this PR had had "no comment from us at all" and that the review "has not started". Neither is true: it was read end to end on 31 July, you got a direction answer on 20 August, and two follow-ups on 27 and 28 August. It also called the `svn` dependency an open project-direction question. **It is not open. The answer was yes on 20 August and it is still yes.** That comment was written during a backlog sweep without reading the thread it was posted into. I am sorry — you did exactly what was asked, waited, and were then told your work had not been looked at.

What stands, so you have one place to read it:

- **Direction: yes, SVN support is wanted.**
- **Condition 1** (20 Aug): opt-in config key, default off. **Condition 2**: the `svn` child must be bounded — deadline, output cap, cancellation — and expiry maps to `CBM_SVN_PROBE_UNCERTAIN`.
- From 27/28 Aug: target the daemon, not `src/main.c`; leave the MCP store gate out; drop the `dump_and_persist` hunk (already on `main`) and the `docs/plans/` file.
- From 2 Sep, the one part that *was* right: the `scripts/security-allowlist.txt` lines and the `.github/workflows/_test.yml` edits are maintainer-only and must leave the diff.

## Where the branch stands against today's `main`

I re-read head `2222bcc2` against `main` (`d615088a`) today. It is 1,502 commits behind, and a trial merge conflicts in 11 of the 23 files — ten hunks in `watcher.c` alone, where `init_baseline`, `check_changes` and `poll_project` have all changed signature (they take the watcher, return a tri-state, and check a `registered` tombstone before admitting the index callback). Since today (#2075) there is also a per-project index-failure backoff that owns `next_poll_ns`. **This head should not be rebased; it would be a rewrite done the hard way.**

The good news is that the integration point named on 27 August has not moved again: `host->watcher = cbm_watcher_new(host->watch_store, host_watcher_index, host)` is at `src/daemon/host.c:611` in `host_state_prepare`, behind the `watcher_enabled` gate. And `watcher_git_run` (`src/watcher/watcher.c:362`) is still exactly the template Condition 2 needs — it takes a literal argv, runs it through `cbm_subprocess_spawn` with a wall deadline and an output limit into a capture file, and publishes the active child so `cbm_watcher_stop` can cancel it. Generalise its executable resolution to take your pinned `svn` path, hand the capture file to your existing `parse_status_stream(FILE *)`, and the shell string, both allowlist lines and the whole new `cbm_popen_argv` API disappear. Your parser does not need to change at all.

## A way to land something soon: split it, core first

**PR-A — the SVN observation core, inert.** `svn_state.c/.h`, `test_svn_state.c`, `svn_test_helpers.h`, plus the Makefile and `test_main.c` registration lines. Nothing calls it yet, so there is no behaviour change, and every file is new or a one-line registration — **zero conflicts with `main`**. For it to pass today's gates:

1. **Traversal without recursion.** `fingerprint_tree` (`svn_state.c:702-745`) holds two 4 KB paths per frame and descends as deep as the path budget allows; the watcher thread has the default stack, and `--no-ignore` makes unversioned trees reachable from ordinary content. An explicit heap work stack, or a depth cap that returns UNCERTAIN, with a test that is red before it.
2. **Memory core.** Since 19 September `scripts/lint-memory-core.py` gates `lint`: a *new* file may contain zero raw `malloc`/`realloc`/`free`/`strdup` sites. `svn_state.c` has seven (`:204 :206 :221 :223 :618 :638 :914`). Use `cbm_alloc` / `cbm_realloc` / `cbm_mem_strdup` / `cbm_free` with a class from `src/foundation/mem_core.h`. This rule did not exist when you wrote the file.
3. **The byte cap on bytes actually read** (`:662-693` checks `st_size` once, then reads to EOF), and **open-then-`fstat`** containment instead of per-component `lstat` followed by a separate open.
4. **Tests that need no real `svn`**: parser, path-safety and PATH-pinning tests against synthetic XML and a fake executable. Today the fixtures hard-fail where `svn` is absent, and none of our local test images has it.
5. No process spawning in this file at all — expose `cbm_svn_client_init` and a parse-from-`FILE *` entry point.

**PR-B — watcher integration behind the opt-in key.** Strategy dispatch inside `main`'s current `init_baseline` / `check_changes` / `poll_project`, the probe through the generalised `watcher_git_run`, the config key modelled on `watcher_enabled` (`CONFIG_KEYS` row, getter with explicit default, docs, and the process-level negative test with a positive control), and folding SVN rescheduling *into* the new backoff rather than replacing it.

**PR-C is ours:** Subversion in the test images, the Windows VM provisioning and the CI workflow, plus the real-`svn` fixture tests. You should not have to touch maintainer-only files to get your feature tested.

Two design questions I would like your view on before PR-B, because you know SVN working copies far better than I do: the per-poll cost (`--verbose --no-ignore --depth infinity` plus content-hashing every candidate — would plain status plus size and mtime, as the git signature does, be enough?), and broken externals (would `--ignore-externals` be acceptable, so one dead external does not stall the primary working copy?).

## What I checked and found right — this is why the answer stayed yes

One read-only invocation, no `-u`, so no network access; the `--` separator and the trailing-`@` peg neutralisation on both platforms. Executable pinning that is *stricter than our own git lookup*: absolute PATH entries only, `svn`/`svn.exe` only, canonicalised and refused under the watched root or the CWD, resolved once per project. Direct argv through `CreateProcessW` on Windows with no `cmd.exe`, and the MSYS2 path translation tested with `%` and `&` in directory names. A hand-rolled XML parser that is bounds-checked everywhere I looked — 16 KB tag buffer, 64-level depth cap, `<!…>` refused outright so there is no DOCTYPE or entity surface, surrogates and out-of-range code points rejected, every `snprintf` checked, any failure yielding UNCERTAIN. Entry paths contained to the root. No credentials, URLs or content in any log line. A strategy enum with the git logic renamed rather than duplicated, and a baseline that commits only when indexing succeeded. That is careful work, and it carries into PR-A almost unchanged.

## The question you were asked on 28 August and then told to wait on

**Would you like to carry this?** PR-A is a contained piece and I will review it the week it arrives. If you would rather not — after ten weeks and a contradictory thread, that would be entirely understandable — say so and we will build the series ourselves from your code, with `Co-authored-by:` credit on every commit. Either answer is fine. This PR stays open as the thread for the work until PR-A exists.

Thank you for the quality of this, and for your patience with a process that has not deserved it.

PR #1131, opened 2026-07-16
Mention un-skip option in docs

gandie:
## What does this PR do?

Short update of `.cbmignore` docs to make unskip option more discoverable.

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) -> skipped, only docs
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) -> skipped, only docs
- [ ] New behavior is covered by a test (reproduce-first for bug fixes) -> skipped, only docs

### DIFF
--- a/docs/cbmignore.md
+++ b/docs/cbmignore.md
@@ -68,15 +68,18 @@ a path wins. For directories:
 
 1. **Built-in skip list** — `.git`, `node_modules`, `dist`, `target`,
    `vendor`, tool caches, etc. (60+ names; the fast/moderate index modes add
-   more, e.g. `docs`, `examples`, `testdata`). Not overridable from any
-   ignore file today.
+  more, e.g. `docs`, `examples`, `testdata`). A `.cbmignore` negation
+  (for example `!target/`) can un-skip these directories, except the
+  non-negatable safety core: `.git`, `node_modules`, `.worktrees`, and
+  `.claude-worktrees`.
 2. **Repo `.gitignore`** — `<repo>/.gitignore` merged with
    `<git-common-dir>/info/exclude` (worktree-aware); later patterns win on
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks for correcting the stale `.cbmignore` documentation and making the existing un-skip behavior discoverable. The exact head is security-clean, DCO-compliant, and CI is green.

One wording contradiction remains: the precedence introduction says “the first layer that rejects a path wins,” while the new section correctly explains that `.cbmignore` can override an ordinary built-in skip. Please qualify the opening rule with the documented `.cbmignore` exception. The implementation description and protected safety-core list otherwise match current behavior.

DeusData:
Merged — thank you, and apologies for the delay on something this clear-cut.

Documentation PRs get the same fact-check as code here, because a confidently wrong doc is worse than a stale one. Every claim in this diff was verified against production before merging:

- The un-skip behaviour you document is real and lives in `discover.c` — built-in skipped directories are un-skipped on a last-match `!` negation.
- The safety-core list in your text matches the implementation verbatim: `.git`, `node_modules`, `.worktrees`, `.claude-worktrees`. Those genuinely cannot be un-skipped, and there is a test named for exactly that guarantee.
- Repo `.gitignore` and nested gitignores still apply *after* an un-skip, via fall-through — which your wording gets right and is the part most likely to be described wrongly.

The "Planned (not yet implemented)" section you deleted was indeed stale: the behaviour shipped and the doc had simply never caught up. Removing it is the more valuable half of this change, because a "planned" marker on a feature that already works actively discourages people from using it.

Tiny cosmetic note, not worth a round trip: a few list-continuation lines moved from three-space to two-space indent. Renders identically.

Thanks for taking the time on docs — it is the kind of contribution that rarely gets opened and always gets read.

Issue #1083, opened 2026-07-14
WAL grows unbounded (115 GB in 4.5 h, fills system drive): checkpoint starvation from accumulated per-conversation servers + 11 concurrent index workers on one repo


9981H:
### Environment
- codebase-memory-mcp 0.9.0 (Windows 11, installed via install.ps1)
- Multiple concurrent MCP clients against the same repo: Claude Code sessions (spawned by Cursor) + Codex sessions (ChatGPT desktop app). Each conversation spawns its own `codebase-memory-mcp.exe`; over one work day 30+ server processes accumulated (clients keep running after their conversations are closed, so servers are never reaped).

### What happened
The main DB `~/.cache/codebase-memory-mcp/<repo>.db` is only **60 MB**, but its `-wal` file grew from 0 to **115 GB in ~4.5 hours** (created 10:47, 115 GB by 15:30 local) and completely filled the system drive (free space went 11.3 GB -> 7.8 GB while we watched, ~0.5 GB per 2 minutes).

Process snapshot at 15:32 showed, besides 20+ long-lived server processes, **11 concurrent `cli --index-worker index_repository` workers all indexing the same repo path** at the same time.

### Diagnosis
Classic SQLite WAL checkpoint starvation:
- dozens of long-lived connections (one per abandoned conversation) hold read marks on the WAL, so `wal_checkpoint` can never truncate;
- each full re-index (clients run `index_repository` at the end of every task) appends the whole dataset to the WAL again;
- log only grows, never shrinks.

Confirmation: at 15:39 a checkpoint finally succeeded and the WAL collapsed from 115 GB to a few MB — so the 115 GB was pure un-checkpointed log, not data volume.

Probably related: #914 (orphan cbm process), #937 (watcher re-index write amplification).

### Suggestions
1. Set `PRAGMA journal_size_limit` and force `wal_checkpoint(TRUNCATE)` after every index run (and/or periodically), logging a warning when the WAL exceeds a threshold.
2. Serialize index workers per DB with a file lock: if an `index_repository` is already running for the same repo, queue or refuse instead of running 11 full indexes concurrently.
3. Defense in depth: the server should exit when its client stdio disconnects (EOF), so abandoned conversations do not accumulate connection holders that block checkpoints.
4. Consider one shared server per DB (with a supervisor/broker) instead of one server per conversation.

DeusData:
[...]

Issue #914, opened 2026-07-06
Windows v0.8.1: stale graph + Permission denied on delete_project after orphan cbm process (reconfirm #277)


ibrahimqureshae:
## Summary

Just confirming that the symptoms from **#277** ("New files not indexed on Windows") are still present in **v0.8.1** on Windows 11. The same workaround from that issue (kill orphan `codebase-memory-mcp.exe` processes, delete the `.db*` files, reindex) is still the only thing that reliably gets the graph back in sync. New files added after the first index simply don't show up until a full reset.

Related: **#520** ("watcher misses file creation") is also still open.

## Environment

- **Version:** `0.8.1`
- **OS:** Windows 11 (10.0.22631), x86_64
- **Install:** official `install.ps1 --ui`, single static binary at `%LOCALAPPDATA%\Programs\codebase-memory-mcp\`
- **Agents:** Claude Code (`~/.claude/.mcp.json`, `~/.claude.json`) and opencode (`~/.config/opencode/opencode.jsonc`), both with `--ui=true --port=9749`
- **Config:** `auto_index=true`, `auto_watch=true` to start with (I ended up turning `auto_watch` off, see below)
- **Cache:** `~/.cache/codebase-memory-mcp/`

## What happens

I indexed a Rust project, then added some new files (a new module plus a few fixture files) on disk. When I tried to reindex:

- `index_repository` comes back with `incremental.noop reason=no_changes` and the **exact same node/edge counts as before** (626/1298), even though the new files are right there on disk and git-tracked (`git ls-files --error-unmatch` finds them).
- `delete_project` returns `status:"delete_failed", error:"Permission denied"` because the SQLite `.db` is still held open by a leftover cbm process.
- `search_graph` for the new symbols (`DomainSubdomains`, `run_scan`, etc.) returns `total: 0`.
- After killing all cbm processes and wiping the DB, a fresh `index_repository` reported `pipeline.discover files=54` while the actual directory tree (excluding `.git`, `.vscode`, `target`, `node_modules`, `dist`, `src-tauri/gen`) had **97 files**. The reindex did pick them up in the end (+206 nodes / +735 edges), so the parser is fine; it's the incremental path that keeps seeing "no changes" when there clearly are changes.

So on Windows this seems to have two faces:

1. **Stale hash table + orphan lock** (per #277): the incremental reindex thinks nothing changed because it can't checkpoint new writes into the main `.db`.
[...]

PR #1137, opened 2026-07-16
fix(store): bound the WAL (journal_size_limit) + warn on checkpoint starvation (#1083 part 1)

DeusData:
## Problem (#1083)

On-disk write connections set **no** `journal_size_limit`, so the `-wal` file is never physically shrunk. All our checkpoints are PASSIVE (TRUNCATE is deliberately avoided — its `ftruncate(fd,0)` can raise SIGBUS in a sibling process that has the DB mmap'd on macOS, see `cbm_store_checkpoint`), and PASSIVE never ftruncates. Under checkpoint starvation the WAL grows without bound (reporter: **115 GB**, filled the system drive), and even a later successful checkpoint leaves the file at its peak size.

## Fix (the reclaim half)

Set `PRAGMA journal_size_limit = 256 MiB` on write connections. SQLite then truncates the `-wal` back to that bound whenever a checkpoint can finally reset the log — reclaiming disk **opportunistically during a session** (whenever readers transiently clear), not only on clean shutdown. 256 MiB is far above the healthy WAL (~4 MiB under the default 1000-page autocheckpoint), so normal indexing never triggers truncate/regrow churn; the limit only fires after abnormal growth. **No TRUNCATE checkpoint is introduced** (that would reintroduce the macOS SIGBUS risk).

## Scope — this is deliberately partial

Per review, this is the **reclaim** half of #1083. It does **not** prevent the 115 GB peak under *continuous* reader contention (SQLite can't reset the log while any reader holds a mark). The actual write-amplification driver — 11 concurrent full re-indexes of one repo — is serialized by the cross-process pipeline lock in **#892**. Server-reaping overlaps **#914**. A shared-server broker is a separate post-0.9.1 item. #1083 stays open, tracking those.

## Test

`journal_size_limit_bounds_wal_issue1083`: a file-backed write store now reports `journal_size_limit == 256 MiB` (RED-verified: on `main` it's -1 / unlimited). The pragma is per-connection and not persisted, so it's read back on the same connection via a small `cbm_store_journal_size_limit()` accessor.

Value note: 256 MiB is a conservative reclaim floor (well above the healthy WAL to avoid churn); worth validating index throughput on a very large repo if we want to lower it.

Store suites + lint-ci pass. Refs #1083.

### DIFF
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -432,6 +432,20 @@ static int configure_pragmas(cbm_store_t *s, bool in_memory, bool read_only) {
         if (rc != CBM_STORE_OK) {
             return rc;
         }
+        /* #1083: bound the WAL file so a checkpoint-starved log is physically
+         * reclaimed the next time a checkpoint can reset it. Our checkpoints are
+         * all PASSIVE (they never ftruncate — see cbm_store_checkpoint's SIGBUS
+         * note), so without a size limit the -wal file only ever grows; a
+         * journal_size_limit truncates it back to N bytes on the next successful
+         * reset. N is far above the healthy WAL (~4 MiB under the default
+         * 1000-page autocheckpoint), so normal indexing never triggers
+         * truncate/regrow churn — it only fires after abnormal growth. We do NOT
+         * use a TRUNCATE checkpoint: its ftruncate(fd,0) can raise SIGBUS in a
+         * sibling process that has the DB mmap'd on macOS. */
+        rc = exec_sql(s, "PRAGMA journal_size_limit = 268435456;"); /* 256 MiB */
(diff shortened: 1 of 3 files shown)


DeusData:
Added a second commit: a **checkpoint-starvation warning**. `cbm_store_checkpoint` now reads `wal_checkpoint_v2`'s out-params and logs `store.wal.starving` when the WAL is large (≥ ~1 GiB of frames) but a PASSIVE checkpoint can't reset it (readers holding marks) — the operator visibility the reclaim limit can't provide on its own, firing before the -wal file fills the disk. Portable (no file stat), only under genuine starvation.

PR #1139, opened 2026-07-16
feat: coordinate concurrent CBM sessions

DeusData:
## Summary

Adds one mandatory per-account coordination daemon for long-lived MCP and hook sessions.

- The first daemon-backed session starts the daemon; the final session shuts it down.
- Watchers, indexing jobs, resource state, and the optional UI are shared across Claude Code, Codex, OpenCode, and other clients.
- Work remains session-owned: disconnecting one session cancels its exclusive work while shared work continues.
- One-shot CLI commands stay local and leave no daemon behind. They use coordinated admission and per-project locks, and can be cancelled safely during maintenance.
- Interactive CLI commands show progress on stderr; stdout remains machine-safe.

## Coordination and activation

All concurrent CBM processes must agree on the exact version, executable build, coordination ABI, and canonical `CBM_CACHE_DIR`. Conflicting processes fail before doing work, report the conflict to the launching session, and record it in `daemon-conflicts.ndjson`.

Native `install`, `update`, and `uninstall` commands:

1. download, verify, and stage the candidate;
2. publish account-wide maintenance intent;
3. cancel daemon sessions and temporary CLI operations;
4. wait for coordinated processes to exit;
5. activate atomically while holding exclusive admission barriers;
6. log the outcome and tell the user to restart open coding-agent sessions.

Windows uses a permanent launcher with an atomically replaceable payload, allowing the same lifecycle guarantees despite executable-file locking. npm, PyPI, and Go wrappers verify and publish their private cached binaries safely.

## Security and observability

- Local IPC authenticates the connecting OS account using Unix peer credentials or Windows access-token checks.
- Coordination state, locks, logs, and activation files use private permissions and fail closed when ownership, ACLs, path identity, or namespace validation is unsafe.
- MCP stdout remains reserved for JSON-RPC; immediate session failures use stderr.
- Durable daemon, conflict, and activation logs are written beneath the canonical cache root.
- No third-party runtime dependency or external coordination service is introduced.

## Verification

Recent local verification on the merged checkpoint line:

- Sanitizer test-runner build passed at the pushed source state.
- Subprocess **27/27**, index supervisor **6/6**, daemon bootstrap **21/21**, daemon frontend **12/12**, daemon runtime **39/39**, daemon IPC **46/46**, cross-repo **7/7**, and MCP **178/178** passed.
- CLI **256/256** passed before the final macOS fixture-only adjustment; the adjusted fixture compiles in the current sanitizer build.
- Windows launcher/bundle contract passed.
- Portable CI lint (cppcheck, formatting, suppression policy) passed.
- Vendored integrity audit passed after updating the pinned SQLite checksum.

Current-head GitHub CI is intentionally left running on this draft PR.

## Handoff / where work stopped (2026-07-18)

Pushed checkpoint: `7696b1486a10f19e231f5b8285abe6eea71b9a3c`.

The latest checkpoint additionally:

- releases the bootstrap handoff when a spawned generation becomes observable and orders daemon ownership as participant → claim → lifetime, with teardown in the reverse-safe order;
- makes subprocess log delivery bounded per poll, lossless at terminal drain, cancellation-linearized, and UTF-8-path-safe on Windows;
- replaces the cross-repo cancellation polling race with a synchronous internal test seam;
- hardens the Windows guard fixture with a protected current-user DACL without weakening product ACL policy;
- fixes the MCP null-plus-zero UBSan finding and stabilizes daemon frontend timeout diagnostics;
- raises the existing vendored Unix SQLite VFS path ceiling from 512 to 4096 bytes, matching CBM's existing 4 KiB path support and fixing the Linux long-database-path regression;
- uses the running test runner as the macOS signed-update candidate instead of re-signing Apple's platform `/usr/bin/true`, which is policy-dependent on older macOS runners;
- merges current `main` and its isolated parallel-suite runner.

Known next task:

- The previous head's full Windows test job compiled successfully but its runtime suite cascaded after strict private-filesystem checks rejected the GitHub/MSYS shared temporary ancestry. Keep production checks strict. Harden a per-user profile test root with a protected current-SID DACL, then route `TEMP`, `TMP`, `TMPDIR`, runtime/cache paths, and the MSYS `/tmp` mapping through it for the full Windows harness. Re-run exact-head Windows smoke/guards/native tests afterward.

No merge was performed; the PR remains draft for the follow-up session.


DeusData:
## Windows daemon/install state — findings & remaining work

Handoff notes after a deep real-Windows debugging pass. A UTM Windows‑11‑ARM64 VM was used as a local leg (driven over ssh) to see stderr/daemon logs that CI structurally hides. Commits pushed on this branch; the one open item is called out at the end.

### Fixed & validated

- **fix(daemon): transactional request completion (the wandering Phase 5 session drop).** Root cause: `runtime_application_worker` stored its `application_thread_done` flag *after* sending the response, so a client that pipelines request N+1 the instant it sees response N raced the previous worker's teardown/join and admission rejected it `BUSY`. On EOF that `BUSY` was then classified as an expected shutdown and the response was silently dropped with exit 0 (mid‑session it exited non‑zero → the CI "timed out waiting for MCP response id=2"). Timing‑sensitive heisenbug — any added stderr shifted the window and it passed, which is why it evaded instrumented runs. Fix: publish the done flag **before** sending the response (admission joins after observing it, absorbing the send tail); `frontend_should_stop` now also requires no in‑flight request so a failure of the final post‑EOF item is loud; the EOF drain is progress‑aware (resets per completed item). Also added a monotonic `admitted_total` latch for the never‑connected window (the live count can read zero between two short sessions) and structured lifecycle logs (`daemon.runtime_stopping reason=`, `lifetime_end reason=`, `connection_end`). **Verified: real‑Windows loop 12/12 (was ~50% loss), macOS daemon suites 143/143.**
- **fix(cli): per‑user install staging + msys‑form path + directory‑chain creation.** The activation transaction and the release‑descriptor probe validate the full ancestor chain of everything they stage (trusted owner, no foreign mutation grant). The process temp root fails that walk by environment luck — under msys `TMP` is `C:\msys64\tmp`, on runners it's the work dir, both carrying broad inherited grants (e.g. `Authenticated Users:(M)` on an upper component). Staging now goes under `%LOCALAPPDATA%\codebase-memory-mcp` (user‑owned end to end); msys `/c/...` form is normalized to `C:/...`; the whole ancestor chain is created. Combined with the earlier launcher trusted‑owner acceptance, the activation owner stamp, and the private `mkdtemp` descriptor, install now clears the directory/staging gates on real Windows.
- **test(daemon): backpressure announce deadline.** The fixture waited 10s for the forked daemon to announce, but the daemon's own cohort‑claim budget is 12s (`HOST_DAEMON_CLAIM_TIMEOUT_MS`); on 4‑core oversubscription a legitimate startup was SIGKILLed before announcing. Raised to 30s. Verified the backpressure failures no longer reproduce under a local CI‑fidelity 4‑core run.

### Open item — mapped‑image capability probe (Phase 8)

With staging fixed, the VM smoke now reaches the managed‑install E2E and fails at the launcher capability probe:

```
DBG cap posix_rename failed os=5   (ERROR_ACCESS_DENIED)
error: managed launcher capability probe failed ...: mapped-image POSIX replace/delete capability probe failed
```

It fails at `windows_posix_rename_handle(source, probe_a)` — a `FILE_RENAME_INFO_EX` with `POSIX_SEMANTICS | REPLACE_IF_EXISTS` replacing `probe_a` **while a spawned child holds that image mapped/running**. **Open question: is this an ARM64/filesystem semantics difference vs the x86_64 CI runners, or a real gap?** CI never reached this probe before (it failed earlier, at staging, which is now fixed), so we don't yet know whether x64 passes it. Suggested next step: run the smoke on x86_64 (CI arch) — either let CI exercise Phase 8 now that staging is fixed, or build the x64 binary locally. (Local x64 build under emulation is set up via the CLANG64 msys env; note ASan does **not** run under x64‑on‑ARM emulation, but a plain build does.)

### Other findings worth a look

- **`cbm_mcp_server_run` (src/mcp/mcp.c) waits on the stdin handle with `WaitForSingleObject`.** That is documented‑undefined for pipe handles and intermittently blocks with data pending. Not on the relay path (which is where the Phase‑5 fix lives), so left unchanged here, but worth a `PeekNamedPipe`‑based wait (+ unbuffered stdin, since a buffered `FILE*` can pull pipelined messages the handle wait then can't see).
- **Makefile has no header‑dependency tracking** (see the comment near the `-MMD` note). A `.h`‑only edit does not rebuild dependents, so incremental builds can silently link stale objects. This caused repeated phantom results during this pass; `touch`/clean discipline is required until dependency generation is added.

### Local Windows harness

`test-infrastructure/vm/` now has a scripted, reproducible real‑Windows leg: `provision-windows.sh` (one‑command zero→built recovery), `win.sh` (build/test/guards/smoke‑install/ubsan/pageheap drivers), `vm-smoke.sh` (runs the CI windows smoke identically), plus `windows-bootstrap.ps1` and a README with the full lifecycle. This is what surfaced the two root causes above and is available for continuing the Phase‑8 work.

DeusData:
## Handoff to Claude — daemon v1 local worktree (2026-07-19)

Work was stopped immediately at the user's request. All workers are stopped. Nothing in this handoff was committed, pushed, or merged.

### Critical state: preserve this worktree

- Repository: `/Users/martinvogel/Documents/Codex/2026-07-13/right-now-i-think-we-have/codebase-memory-mcp`
- Branch: `feat/shared-coordination-daemon`
- Local and remote PR head: `08b7ff54d23b8b94074bf009aeebe85be80c7b4a`
- PR: #1139, still open and draft.
- The latest fixes are **only in the dirty local worktree on this Mac**. Do not pull/reset/checkout/clean over it. Start with `git status --short`, `git diff`, and `git diff --cached`.
- Snapshot at handoff: 45 tracked files differ from HEAD (`6404 insertions, 1066 deletions`, excluding untracked files); `git diff --check` passes.
- There is a mixed index/worktree state: `test-infrastructure/vm/provision-windows.sh` and `test-infrastructure/vm/win.sh` are `MM`; `test-infrastructure/vm/vm-smoke.sh` is staged. Final staging must use the complete worktree, not assume the current index is complete.
- Required untracked files currently present:
  - `scripts/path-safety.sh`
  - `test-infrastructure/vm/ssh-common.sh`
  - `tests/test_build_dir_safety.sh`
  - `tests/test_diagnostics.c`
  - `tests/test_soak_daemon_recovery_contract.sh`
  - `tests/test_ui_dev_proxy_security.sh`
  - `tests/test_vm_worktree_manifest.sh`
- `graph-ui/tsconfig.tsbuildinfo` is a generated refresh and should be removed from the final diff after verification, without disturbing unrelated user changes.

### Product scope remains narrow

Finish the mandatory per-account coordination daemon v1 and make this PR green. Do not add downstream features. CLI one-shot commands remain separate/local; long-lived MCP/hook servers use the daemon. No merge is authorized in this handoff.

### Infrastructure is operational

This is not blocked by the Codex sandbox. Host socket/VM commands merely need the escalated execution path in Codex.

- Colima: running, aarch64, Virtualization.Framework, Docker runtime, 18 CPUs.
- Docker: 29.5.2, aarch64, 18 CPUs.
- Real Windows 11 ARM VM: online and reachable over SSH; repository binaries are present.
- VM entry points: `test-infrastructure/vm/win.sh`, `provision-windows.sh`, and the README in that directory.
- Host VM config: `~/.claude/cbm-vm/config`; key: `~/.claude/cbm-vm/id_ed25519`.
- `win.sh status` was green immediately before this handoff.
- Use the maintained scripts, not ad-hoc SSH command chains. `provision-windows.sh` is disk-loss recovery.
- The Makefile still lacks header dependency tracking. Clean/rebuild when headers change; stale incremental objects previously produced phantom results.
- External non-repo follow-up notes are in `/Users/martinvogel/Documents/Codex/2026-07-13/right-now-i-think-we-have/CBM_FOLLOW_UPS.md`; never commit that file.

### Implemented locally and already proven at least once

#### 1. Windows `cmd.exe /C` subprocess construction

Root cause: the generic CRT argv encoder was incorrectly used for the command-language payload passed to `cmd.exe /C`. This broke the fake-Git/runtime fixtures and could alter shell-command semantics.

Local implementation:

- Added `windows_cmd_payload` to `cbm_proc_opts_t` and the pure `cbm_build_win_cmd_payload()` helper.
- It requires an absolute path ending in `cmd.exe` and emits `"<absolute cmd.exe>" /D /S /V:OFF /C <payload-verbatim>` while leaving generic CRT argv quoting unchanged.
- MCP resolves the trusted absolute system `cmd.exe` via `GetSystemDirectoryW`; it does not trust `PATH` or `COMSPEC`.
- Windows-interpolated base branch/project-root values reject `%`, `!`, and `^`.
- Capacity underflow, exact-fit/one-short, invalid path, and shell metacharacter tests were added.

Verified on the real Windows VM before the final runtime edit:

- `subprocess`: 14 passed, 15 platform skips.
- `mcp`: 173 passed, 7 skips.
- The Windows MCP symlink fixture was corrected to translate the MSYS spelling to native backslashes only for `cmd /c mklink`.

#### 2. Git SHA-256 watcher support

Root cause: watcher buffers were `CBM_SZ_64`, which cannot hold a 64-hex SHA-256 object ID plus newline/NUL. Baseline reads therefore retried forever and auto-refresh never worked for SHA-256 repositories.

- Widened watcher `last_head`, `pending_head`, and local HEAD buffers to `CBM_SZ_128`.
- Added a real SHA-256 Git repository regression test.
- Red-first result reproduced `observed=0 expected=1`; after the fix macOS watcher passed 70/70.

#### 3. Windows watcher stop/unwatch cancellation coverage

`tests/test_watcher.c` now has a native Windows version of the blocked-Git cancellation test:

- copies the current test runner to a private `bin/git.exe`;
- prepends native Windows `PATH` with `;`;
- retains a process handle and verifies exact image path plus volume/file identity with `QueryFullProcessImageNameW`;
- uses bounded waits and permits the `TerminateProcess` cleanup backstop only after immediate exact-image revalidation;
- requires both watcher stop and unwatch to terminate the verified child before the backstop.

Static checks passed (`git diff --check`, clang-format, no-skips, Windows/POSIX cppcheck). **This new Windows branch has not yet been compiled or run on the VM.**

#### 4. Soak harness path/recovery fixes

- Keeps host/MSYS project paths separate from the Windows-native MCP value.
- JSON-escapes repository paths once and uses the escaped value for every request.
- Captures the authoritative project name from `index_repository` instead of deriving it from the path.
- Default soak queries the generated `handle_1`, not nonexistent `compute`.
- Tracks randomized diagnostics generations across frontend/daemon recovery.

`bash -n scripts/soak-test.sh` and `tests/test_soak_daemon_recovery_contract.sh` passed. A full soak has not been rerun after all current changes.

#### 5. Windows permanent-launcher context ownership

`src/cli/windows_launcher_state.c` now has a dedicated context-file opener that accepts trusted owners via `windows_owner_secure(file, false)`, while the generic transaction opener remains exact-current-owner. It retains path-tree, no-reparse, ACL, PID/creation-time, link-count, and same-file checks. Python/static/format checks passed. The final real-VM launcher guard/smoke run is still required.

#### 6. HTTP/UI shutdown lifecycle

Local changes include Windows 50 ms select slices, interrupt checks, nonblocking read/send with a response-wide one-second deadline, explicit connection lifecycle states, and refusal to free an HTTP server while a listener-owned connection remains active. macOS `httpd` passed 59 tests with one expected Windows-only skip. Windows and full-suite verification remain.

### Stop-point: Windows daemon runtime is not green yet

The last real-Windows `daemon_runtime` diagnostic run before the final edit was 33 passed / 3 failed:

1. `daemon_runtime_disconnect_cancels_blocked_non_index_child_and_preserves_other_session`
   - Marker/image/close-begin conditions were all reached.
   - After two seconds the daemon still reported two clients/two connections.
   - The copied child remained alive; the verified cleanup backstop was required.
   - The request eventually returned OK after cleanup, while the second session remained usable.
2. `daemon_runtime_conflict_log_failure_uses_operation_log_fallback`
   - The owner connected and the rejected client was NULL, but the result remained zero/uninitialized (`status=0`, `hello=0`, empty message), indicating the explicit conflict response was not decoded before EOF/timeout.
3. The capacity test failed once but had passed immediately before; treat it as an unresolved flake until repeated, not as fixed.

#### Cancellation root cause and partially applied final edit

Root cause is confirmed: on POSIX, client close interrupts the socket with `shutdown()`, so the server sees EOF and cancels session-owned work. On Windows, `cbm_daemon_ipc_connection_interrupt()` on a connected-client named pipe only calls `CancelIoEx()` on local I/O; it cannot call server-side `DisconnectNamedPipe()` and does not notify the daemon. `close_finish` then waits behind the in-flight exchange, so server session cancellation never fires.

Immediately before work was stopped, `src/daemon/runtime.c` was edited to:

- serialize close-begin with request publication using the existing send→state lock order;
- recheck `closing` under the send lock before publishing a request;
- best-effort send the exact active token's `APPLICATION_CANCEL` frame before `CancelIoEx()`;
- then interrupt the local receive so the exchange can unwind and `close_finish` can close the pipe.

The server already keeps the connection loop available while the application worker runs, and `APPLICATION_CANCEL` reaches `application_request_cancel` → MCP cancellation → subprocess cancellation. This is the correct narrow direction and preserves other sessions.

**Important: that last runtime edit was made just before the worker was interrupted. It has not been compiled or tested anywhere. Audit the lock ordering/state transitions first, then clean-build and rerun macOS/Windows runtime and TSan. Do not assume it is complete.**

The test currently still includes temporary conflict diagnostics. Remove debug-only printing after the failure is understood.

#### Conflict-response failure remains open

Do not fix this by raising the timeout or weakening the assertion. Evidence currently points to a Windows send/close ordering problem: the server constructs and sends the hello conflict response, then immediately finishes/closes the rejected connection; the client returns NULL without decoding it. Instrument the server send result/pipe completion and client receive path. A `FlushFileBuffers` or explicit acknowledgement may be relevant, but no final design was proven before stop; avoid introducing an unbounded pipe flush during rejection.

### Diagnostics discovery/security issue found, not implemented

The diagnostic writer itself is private, but its discovery contract is currently unreliable:

- `cbm_diag_start()` announces exact randomized paths through `cbm_log_info("diagnostics.start", ...)`.
- `CBM_LOG_LEVEL=warn`, `error`, or `none` suppresses that info event entirely.
- Text logging's `append_text_atom()` maps whitespace/control bytes to `_`, so a valid temp/cache path containing spaces is corrupted and cannot be opened.
- The soak parser assumes a text `msg=diagnostics.start` record and cannot recover such paths.
- POSIX `cbm_tmpdir()` globally hardcodes `/tmp`, while the user docs promise `$TMPDIR` or `/tmp`.

Recommended narrow direction:

- Treat diagnostics discovery as an always-delivered internal control/discovery record when `CBM_DIAGNOSTICS=1`, through the existing daemon log sink—not by relabeling it warning/error and not by bypassing the sink with raw stderr.
- Give the record an unambiguous escaped representation (JSON is the natural existing encoder) so exact paths survive spaces/control characters; update soak parsing and tests accordingly.
- Document that enabling diagnostics emits this one discovery record even if the ordinary log threshold is `none`.
- Do **not** globally change `cbm_tmpdir()`: it is used broadly by store, daemon application, pipeline, UI, MCP, watcher, install/update, and tests. Add a diagnostics-local POSIX temp-base helper that honors nonempty `$TMPDIR` with `/tmp` fallback.
- Add red-first coverage for a temp path containing spaces and suppressed log levels.

Windows security detail already established:

- `cbm_mkdtemp()` attempts explicit exact-current-user ownership, a single `FILE_ALL_ACCESS` ACE, and a protected DACL.
- Diagnostics opens the directory with `OPEN_REPARSE_POINT`, then `cbm_private_lock_directory_adopt_windows()` validates owner-only protected DACL, local DOS path, no reparse ancestry/directory, noninheritable handle, and path/handle identity. Adoption validates; it does not weaken ACLs.
- Output files are separately validated as owner-only regular, non-reparse, single-link files.
- Existing `diagnostics_outputs_are_owner_private_regular_files` is POSIX-only. Add a native Windows integration assertion rather than weakening the production validation.

### Old GitHub CI is stale relative to this worktree

PR checks on remote head `08b7ff54` currently show:

- Green: lint, security/static, license gate, CodeQL, DCO, TSan, both macOS jobs, Ubuntu/macOS PR smoke.
- Red: Windows guards, Windows full test, Windows PR smoke, Ubuntu x86_64, Ubuntu ARM64, and aggregate `ci-ok`.

Those runs predate every uncommitted fix above. Do not rerun GitHub yet: first make the complete local tree green on all three platforms, then commit/push one verified SHA and let GitHub verify that exact SHA.

### Required next sequence

1. Preserve/snapshot the dirty tree. Inspect the final untested `runtime.c` edit and complete the Windows cancellation test loop.
2. Resolve the explicit conflict-response failure with evidence; repeat the capacity test.
3. Clean-sync/rebuild the Windows VM and run focused suites: `subprocess`, `mcp`, `daemon_runtime`, `daemon_application`, `watcher`, `diagnostics`, `httpd`, `windows_launcher_state`, then `guards` and `smoke-install`.
4. Implement the diagnostics discovery fix red-first and run its native Windows ACL/security test.
5. Rebuild and run focused macOS suites after all source changes: at minimum subprocess, MCP, daemon runtime/application, watcher, diagnostics, HTTPD, launcher/CLI.
6. Run the mandatory pre-push local gate on the final tree:
   - `./test-infrastructure/run.sh lint`
   - macOS `scripts/test.sh`
   - Linux ARM64 `./test-infrastructure/run.sh test`
   - Linux AMD64 via the maintained `run.sh` AMD64 entry point
   - the CI-like four-CPU fidelity pass for timing/subprocess changes
   - real Windows full parallel tests, guards, and smoke-install through `win.sh`
7. Remove generated-only noise such as `graph-ui/tsconfig.tsbuildinfo`, run final diff/security review, and ensure all seven required untracked files are intentionally staged.
8. Only after local all-green: DCO commit with `martin.vogel.tech@gmail.com`, push, update this PR description/checkpoint, and monitor the full GitHub matrix. Do not merge.

No graph/index tooling was used in this session; the user explicitly requested grep/direct-source discovery only.

PR #1141, opened 2026-07-16
fix(mcp): reject misleading ADR sections writes

### LINKED ISSUE #1140: manage_adr sections argument silently permits whole-document replacement
## Problem

The manage_adr tool schema advertises a sections array beside mode=update, but the handler never reads that argument. An update request that supplies sections still calls the whole-document store primitive and replaces the complete ADR.

A caller can therefore believe it is updating selected sections while silently deleting every omitted decision.

## Reproduction

1. Store an ADR containing two or more sections.
2. Call manage_adr with mode=update, a sections array naming one section, and content containing only that section.
3. Call manage_adr with mode=get.
4. Observe that the returned ADR contains only the replacement content; all omitted sections are gone.

## Expected behavior

Until an append or section-scoped CAS API exists, the tool must fail closed:

- do not advertise sections as a write argument
- explicitly reject legacy requests that still provide sections before opening or writing the store
- document and label mode=update as whole-document replacement

## Scope

This issue is limited to closing the misleading write-contract footgun. It does not add append, CAS, or Markdown-authoritative storage.

Related durability and architecture work: #516 and #507.

M-Marbouh:
## Summary

- remove the dead sections input from the manage_adr schema
- explicitly reject legacy requests that still pass sections before opening a store
- label successful update responses as whole-document replacement
- add a regression proving rejected legacy requests leave the ADR byte-identical

## Why

The schema advertised sections as an update argument, but the handler never consumed it. A caller could therefore believe it was performing a scoped update while replacing the complete ADR document.

Closes #1140.

Related context only: #516 and #507.

## Verification

- ASan/UBSan focused MCP suite: 166 passed
- production code compiled with warnings as errors as part of the test-runner build
- git diff --check clean

LeakSanitizer was disabled for the focused run because it is incompatible with the sandbox ptrace wrapper; AddressSanitizer and UndefinedBehaviorSanitizer remained enabled.

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -643,8 +643,9 @@ static const tool_def_t TOOLS[] = {
     {"manage_adr", "Manage ADR", "Create or update Architecture Decision Records",
      "{\"type\":\"object\",\"properties\":{\"project\":{\"type\":\"string\"},\"mode\":{\"type\":"
      "\"string\",\"enum\":[\"get\",\"update\",\"sections\"],\"description\":\"update replaces "
-     "the entire ADR document; sections only lists existing headings\"},\"content\":{\"type\":\"string\","
-     "\"description\":\"Complete replacement document required by update\"}},\"additionalProperties\":false,"
+     "the entire ADR document; sections only lists existing "
+     "headings\"},\"content\":{\"type\":\"string\",\"description\":\"Complete replacement document "
+     "required by update\"}},\"additionalProperties\":false,"
      "\"required\":[\"project\"]}"},
 
     {"ingest_traces", "Ingest traces", "Ingest runtime traces to enhance the knowledge graph",
@@ -10167,8 +10168,8 @@ static char *handle_manage_adr(cbm_mcp_server_t *srv, const char *args) {
     yyjson_doc *args_doc = yyjson_read(args, strlen(args), 0);
     if (args_doc) {
         yyjson_val *args_root = yyjson_doc_get_root(args_doc);
-        has_sections_arg = args_root && yyjson_is_obj(args_root) &&
-                           yyjson_obj_get(args_root, "sections") != NULL;
+        has_sections_arg =
+            args_root && yyjson_is_obj(args_root) && yyjson_obj_get(args_root, "sections") != NULL;
         yyjson_doc_free(args_doc);
     }
     if (has_sections_arg) {
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -642,9 +642,10 @@ static const tool_def_t TOOLS[] = {
 
     {"manage_adr", "Manage ADR", "Create or update Architecture Decision Records",
      "{\"type\":\"object\",\"properties\":{\"project\":{\"type\":\"string\"},\"mode\":{\"type\":"
-     "\"string\",\"enum\":[\"get\",\"update\",\"sections\"]},\"content\":{\"type\":\"string\"},"
-     "\"sections\":{\"type\":\"array\",\"items\":{\"type\":\"string\"}}},\"required\":[\"project\"]"
-     "}"},
+     "\"string\",\"enum\":[\"get\",\"update\",\"sections\"],\"description\":\"update replaces "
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for pairing the report with a focused fail-closed regression. I’ve classified this as a high-priority 0.9.1-rc bug. Exact-head review will verify that stale `sections` requests are rejected before store access, the schema no longer advertises scoped writes, and existing whole-document callers retain an explicit, unambiguous contract.

DeusData:
Thanks for this, and sorry for the delay in acknowledging it. Queued for review.

Currently CONFLICTING against `main` — CI is green across all 20 checks, so a rebase is the only mechanical thing standing in the way. Worth doing before I read it so I am reviewing the intended diff.

Rejecting misleading ADR section writes is a correctness-of-contract fix rather than a feature, which is the kind I like receiving.

M-Marbouh:
No worries. Thanks for the clear review criteria.

I’ve rebased the PR onto current main, and it’s conflict-free now. The resolution keeps the newer manage_adr lifecycle intact while ensuring legacy sections requests are rejected before mutation coordination or store access. I also tightened the regression to confirm the existing ADR remains byte-identical and that the mutation guard is never entered.

Local verification passed with 179 MCP tests and 2 platform-only skips. I also removed the guard temporarily as a negative control, and the new regression failed as expected, so the test is proven to catch the old behavior.

One heads-up, GitHub cancelled two CI attempts across all active runners with a “runner received a shutdown signal” message. The checks that finished beforehand were green, but I don’t have permission to rerun the upstream workflow from the fork. When convenient, could you please rerun the cancelled checks?

The diff remains limited to the original contract fix and its regression, nothing else was bundled in.

DeusData:
Merged — and this was a genuinely good catch on a data-destroying footgun.

**I verified the premise end to end on current `main` rather than taking the description's word for it**, because the claim is serious:

- The `manage_adr` input schema advertises `sections` as a property (`src/mcp/mcp.c:653-654`).
- `handle_manage_adr` never reads it — it extracts only `project`, `mode` and `content`. A full-file search confirms the only other uses of `sections` are the read-only heading lister and its `mode="sections"` call site.
- And `mode=update` is an unconditional **whole-document replacement** (`mcp.c:10479-10481`).

So `mode=update, sections=["PURPOSE"], content="## PURPOSE…"` returned `status:"updated"` while **silently deleting every other section** of the user's ADR. A success status on a destructive no-op is about the worst shape a tool contract can have, and the dead property has been there since the original C pipeline commit without ever being consumed.

**The fix is textbook fail-closed.** Rejecting *before* `mcp_project_mutation_begin` and before any store is opened means there is no guard to imbalance, no leak on the early return, and zero interaction with a concurrent reindex. I checked each of those: `project`, `mode_str` and `content` are all freed on the early path, `args_doc` is freed on both branches, and malformed argument JSON falls through to the previous behaviour rather than erroring — consistent with the other argument helpers.

**Your test is properly binding, in two independent ways.** On unpatched main it fails both on the status assertion *and* on the byte-identical-ADR assertion, and the mutation-guard probe pins begin/end at 0/0. I also confirmed you are not overcorrecting: the existing round-trip and guard-balance tests still cover honest, sections-free updates, and they are green.

**And the scoping call was right.** Deliberately deferring append and compare-and-set to the existing issues, and keeping this PR to just closing the contract footgun, made the diff atomic and the review fast. That is worth saying because the temptation in this area is always to fix everything at once.

Three small notes, none of which held up the merge:

- The rejection fires on *any* mode carrying a `sections` key, including `get` and `sections` where nothing destructive would have happened. Fail-closed and defensible, and the error explains itself — just slightly broader than the destructive case, and the message reads a little update-centric for a read.
- `invalid_arguments` is a new status string with no exact precedent in the existing vocabulary. Stylistically consistent, and now part of it.
- `additionalProperties:false` is advertised while only `sections` is actually enforced server-side. Harmless, and precedented by `ingest_traces`.

For the record: the red checks this PR carried earlier were **cancellation noise, not your code**. A plain re-run went green with nothing changed.

Thank you — this is exactly the kind of contract bug that is invisible until it eats someone's ADR.

PR #1148, opened 2026-07-17
fix(mcp): preserve coverage scope labels

aaiyer:
## What does this PR do?

`check_index_coverage` stored each normalized scope with yyjson's non-copying
string API even though the source was a loop-local buffer. With multiple
scopes, later iterations could overwrite earlier labels; AddressSanitizer also
reported a stack-use-after-scope when the response was serialized.

Use `yyjson_mut_obj_add_strcpy` so the mutable document owns each scope label.
Add a regression that requests three distinct, equal-length scopes and asserts
that every returned label and its order are preserved.

This PR is intentionally limited to one issue and two files:

- `src/mcp/mcp.c`
- `tests/test_mcp.c`

It does not change coverage semantics, tool inputs, schemas, pagination, or
status handling.

### Verification

- Reproduce-first: the focused MCP sanitizer suite aborted before the fix with
  `stack-use-after-scope` in `handle_check_index_coverage` during yyjson
  serialization.
- `make -j4 -f Makefile.cbm test-focused TEST_SUITES=mcp`: 166/166 passed.
- `scripts/lint.sh --ci` with CI's clang-format 20.1.8 and cppcheck 2.20.0:
  passed.
- `make -j"$(nproc)" -f Makefile.cbm security`: all seven security layers
  passed, including MCP robustness 23/23.
- Focused review: 0 High, 0 Medium, 0 Low findings.

The broad `scripts/test.sh` run reached the end with all test cases reporting
`PASS`, then LeakSanitizer exited nonzero for a 4,976-byte allocation originating
from the unchanged `cli_install_plan_receipt_no_mutation_issue388` test path
(`tests/test_cli.c:2398` / `src/cli/cli.c:7539`). This PR does not touch that
path; it is being kept out of this one-issue change and investigated separately.

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) — all cases passed, but
      the separate unchanged CLI LeakSanitizer finding above made the process
      exit nonzero

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -3511,7 +3511,7 @@ static char *handle_check_index_coverage(cbm_mcp_server_t *srv, const char *args
                 yyjson_mut_arr_add_val(scope_results, item);
                 continue;
             }
-            yyjson_mut_obj_add_str(doc, item, "scope", scope[0] ? scope : ".");
+            yyjson_mut_obj_add_strcpy(doc, item, "scope", scope[0] ? scope : ".");
             cbm_coverage_row_t *rows = NULL;
             int row_count = 0;
             int cov_rc = cbm_store_coverage_get_scope(store, project, scope, &rows, &row_count);
--- a/tests/test_mcp.c
+++ b/tests/test_mcp.c
@@ -1458,6 +1458,40 @@ TEST(tool_check_index_coverage_reports_paths_scopes_and_ranges) {
     PASS();
 }
 
+TEST(tool_check_index_coverage_preserves_multiple_scope_labels) {
+    char tmp[256];
+    cbm_mcp_server_t *srv = setup_snippet_server(tmp, sizeof(tmp));
+    ASSERT_NOT_NULL(srv);
+
+    char *coverage = cbm_mcp_handle_tool(srv, "check_index_coverage",
+                                         "{\"project\":\"test-project\","
+                                         "\"scopes\":[\"alpha/one\",\"bravo/two\",\"charl/tri\"]}");
+    ASSERT_NOT_NULL(coverage);
+    char *inner = extract_text_content(coverage);
+    ASSERT_NOT_NULL(inner);
+    yyjson_doc *doc = yyjson_read(inner, strlen(inner), 0);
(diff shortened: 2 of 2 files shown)


DeusData:
Thanks for isolating the yyjson ownership defect and retaining a multi-scope sanitizer regression. Because this is a stack-use-after-scope in an MCP response path, I’ve classified it as a high-priority 0.9.1-rc memory-safety fix. Review will stay focused on the copying API, response lifetime, and whether equivalent loop-local yyjson strings exist in adjacent coverage output.

DeusData:
Thanks for this, and apologies for the slow first response. Queued for review.

Mechanically clean: MERGEABLE against `main` with all 20 checks green, and a tight +36/-1 across two files. Nothing for you to do — it is waiting on my read, not on you.

Label precision on coverage scope is something this project cares about specifically, so a fix in that area is welcome.

DeusData:
Merged — and this one deserved to land much faster than it did. Thank you.

You found a live correctness bug in a tool we ship and our own agents call. In `handle_check_index_coverage` the `scope` buffer is declared *inside* the `yyjson_arr_foreach` body, and it was being handed to `yyjson_mut_obj_add_str`, which stores the pointer rather than copying it. By the time the document was serialized, every scope entry pointed at the same dead — and reused — stack slot. So multi-scope responses reported the **last** scope's label for every entry, on top of being stack-use-after-scope undefined behaviour.

What makes the fix right rather than merely plausible:

- `add_strcpy` is the established pattern in this very function — `requested_scope` eleven lines above already uses it. You matched the local convention instead of inventing a new one.
- The test is cleverly constructed: three *equal-length* scope names, so the buffer reuse produces a deterministically wrong label rather than something that might accidentally look correct. And it fails under ASan regardless, because the use-after-scope is real. That is a test that proves the bug, not one that merely accompanies the fix.

We also swept the rest of the function while reviewing — every other `add_str` in that loop passes a string literal, so there is no second instance of this hiding nearby.

This is the same class as a pointer-vs-copy bug we have been chasing elsewhere in the MCP layer, and having it independently caught and fixed here is genuinely useful. Thanks.

PR #1149, opened 2026-07-17
test(cli): isolate install plan receipt fixture

aaiyer:
## What does this PR do?

Fixes the cli_install_plan_receipt_no_mutation_issue388 test harness so it cannot inherit a developer or CI runner CODEX_HOME while using a temporary HOME fixture.

The observed 4,976-byte LeakSanitizer report was a harness-cleanup symptom, not a production allocation defect:

- With inherited CODEX_HOME, production correctly resolved the Codex path outside the temporary fixture.
- The receipt assertion then failed before free(json).
- The same pre-fix sanitizer binary passed all 222 CLI tests when run with CODEX_HOME unset.

The fixture now saves, unsets, and restores CODEX_HOME. It also records the receipt and no-mutation results, then frees the JSON, restores the environment, and removes the temporary directory before reporting an assertion failure.

## Reproduce-first evidence

- RED: with inherited CODEX_HOME, ASan/UBSan CLI execution failed at the install-plan receipt assertion and LeakSanitizer reported 4,976 bytes allocated through yyjson in cbm_build_install_plan_json.
- CONTROL: the same pre-fix binary with CODEX_HOME unset passed 222/222 CLI tests without a leak.
- GREEN: the patched focused CLI ASan/UBSan suite passes 222/222 under the inherited CODEX_HOME environment, without a sanitizer report.

## Checks

- [x] One signed-off commit; scripts/check-dco.sh upstream/main..HEAD passes
- [x] Focused CLI ASan/UBSan suite: 222/222
- [x] scripts/lint.sh --ci
- [x] make -f Makefile.cbm security
- [x] scripts/check-no-test-skips.sh
- [x] clang-format 20.1.8 dry-run
- [x] git diff --check
- [x] Independent focused review: 0 High / 0 Medium / 0 Low
- [ ] Full local scripts/test.sh: started but stopped after the relevant CLI phase while unrelated deep/matrix suites continued pathologically long; it did not reach a final summary. GitHub CI is authoritative for the exhaustive suite.

## Scope

Only tests/test_cli.c changes. Production behavior is unchanged.

### DIFF
--- a/tests/test_cli.c
+++ b/tests/test_cli.c
@@ -2388,6 +2388,11 @@ TEST(cli_install_plan_receipt_no_mutation_issue388) {
     if (!cbm_mkdtemp(tmpdir))
         FAIL("cbm_mkdtemp failed");
 
+    /* Production correctly honors CODEX_HOME, but this fixture must not inherit
+     * the developer or CI runner's real Codex home. */
+    char *saved_codex_home = save_test_env("CODEX_HOME");
+    cbm_unsetenv("CODEX_HOME");
+
     /* Make Cursor + Codex "detected". */
     char dir[512];
     snprintf(dir, sizeof(dir), "%s/.cursor", tmpdir);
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks for separating the fixture cleanup symptom from production behavior. This is classified as a normal-priority 0.9.1-rc test-reliability fix. We will review it alongside #1150 because both touch the same CODEX_HOME fallback fixture; the combined result must preserve environment restoration and cleanup on every assertion path without masking production path precedence.

DeusData:

Mechanically clean — MERGEABLE, checks green, +16/-9 in a single file. Nothing needed from you; it is waiting on review time rather than on any blocker.

Fixture isolation is unglamorous and pays for itself the first time it prevents a flaky CI run.

DeusData:
Thank you — reviewed, and the analysis behind this is correct on both counts.

The `CODEX_HOME` save/unset/restore is right, and your reading of the 4,976-byte "leak" is right too: it was an assertion aborting before the frees, not a production leak. Diagnosing that correctly rather than papering over it with a suppression is the harder and more useful thing to have done.

The second half — restructuring so cleanup happens *before* `FAIL` (free the JSON, restore the env, remove the temp dir, then fail) — is genuinely valuable and is **not** present in any other PR. An assertion failure previously left both a leak and a stray `/tmp` directory behind, which makes a red test harder to debug than it needs to be.

**The coordination problem, and what we would like to do about it.** This PR and your #1150 touch the same lines, and they conflict textually. #1150 is the superset for isolation: it puts a global `CODEX_HOME` sentinel in `test_main.c`, adds a regression test, and unsets per-test in both `detect_agents` and issue388. Landing #1149 on its own would fix the isolation for one fixture while the rest of the suite kept reading a developer's real `CODEX_HOME`.

So the sequence we would like, if you agree:

1. **#1150 first** — it needs a rebase on current `main` (it is conflicting; the tree moved tonight). We have asked over there.
2. **Then this PR, rebased down to just the cleanup-before-FAIL restructure.** The env-isolation hunks here become redundant once #1150 is in, but the cleanup ordering is still worth having on its own.

If you would rather collapse the whole thing — fold the cleanup restructure into #1150 and close this one — that is equally fine and saves you a round trip. Entirely your preference; you keep the credit either way.

One small, non-blocking observation while we were in there: folding the checks into a single bool loses which `strstr` actually failed, so a future failure reports less than it could. Worth keeping them separate if it costs nothing.

Thanks for the care on test hygiene — it is unglamorous work and it is exactly what keeps CI from becoming a lottery.

DeusData:
Carried forward with credit — **#1775**, commit `f3b356af`, `Co-authored-by: Anand Aiyer <anand@aiyer.com>`.

The reason I have not simply asked you to rebase is that **the plan I gave you no longer exists**. My last comment set out a sequence: land #1150 first, then rebase this down to just the cleanup restructure. #1150 was subsequently **closed unmerged**. So you have been sitting on advice whose precondition evaporated, with no way to know that from this thread. That is on me.

**What survived and what did not.** The `CODEX_HOME` isolation you added is now redundant, but not because your reading was wrong — main grew a global sentinel (`tf_client_home_overrides[]` in `tests/test_main.c`) that unsets `CODEX_HOME` along with every other client home at suite setup. Your diagnosis was right; the fix arrived from a different direction.

**The cleanup-before-`FAIL` restructure is entirely yours and it has landed.** That half was never redundant. The test asserted inline, so any failure returned before `free(json)` and `test_rmdir_r(tmpdir)`, meaning every red run leaked the receipt and left a stray `/tmp/cli-plan-*` behind. A failing test that also litters is strictly harder to debug than one that doesn't, and you were the one who noticed.

I took your other diagnosis seriously too: reading the 4,976-byte "leak" as an assertion aborting before the frees, rather than a production leak, is the harder call and the correct one. Papering over that with a suppression would have hidden a real reporting defect.

One change on top of your version, from the non-blocking note I left last time: instead of folding the marker checks into a single boolean, the loop records **which** marker was missing, so a failure says `install plan receipt is missing next_safe_command` rather than reporting that something was absent. Same cleanup guarantee, without losing the diagnostic.

Verified at `build/c/test-runner cli` — 281 passed, 0 failed.

Closing this in favour of #1775. Both of your PRs were correct and both waited far too long; thank you for the care on test hygiene, and I hope the delay does not put you off.

PR #1150, opened 2026-07-17
fix(tests): isolate Codex home from user state

aaiyer:
## What does this PR do?

Completes the test-state isolation added in #1067. The runner replaced HOME and cleared CBM_CACHE_DIR, but it still inherited the invoking process's CODEX_HOME. When that directory existed, CLI fixtures detected the developer's real Codex installation and repeatedly installed and uninstalled MCP config, AGENTS.md content, skills, and Scout/Verify/Auditor profiles there.

A sentinel run on exact upstream/main e678b2b6acb02bc1ab84a854f2df0e1d092f2cc0 reproduced thousands of accesses and explicit write syscalls under the inherited CODEX_HOME, left an extra skills directory, and failed the install-plan no-mutation fixture because the inherited override shadowed its temporary .codex directory.

## Fix

- set CODEX_HOME to the private test HOME's .codex path before normal suites run;
- fail test-home setup instead of continuing if the environment isolation cannot be established;
- add a regression that binds CODEX_HOME to the private test HOME;
- explicitly unset and restore CODEX_HOME in the two tests that intentionally exercise the home-directory fallback.

Worker and socket-probe modes still exit before normal suite setup, as before.

## Validation

- ASan + UBSan test runner build with warnings as errors: passed;
- make -f Makefile.cbm test-focused TEST_SUITES=cli: 223 passed;
- inherited fresh CODEX_HOME sentinel under strace: 0 sentinel-path syscalls after the fix, identical file hashes and directory inventory;
- bash scripts/check-no-test-skips.sh: passed;
- bash scripts/security-audit.sh: passed with existing review notices only;
- bash scripts/security-vendored.sh: passed (75 files);
- git diff --check: passed;
- DCO check: passed.

Local CI lint was attempted but cppcheck was unavailable; clang-format and clang-tidy are also not installed in this environment. GitHub CI remains the lint authority.

## Coordination

#1149 overlaps the same install-plan fallback test for a separate fixture-lifetime issue. Whichever PR lands second will need to rebase and preserve both intents: free the receipt JSON on every validation-failure path, and explicitly isolate CODEX_HOME while exercising the fallback path.

## Scope

Only tests/test_main.c and tests/test_cli.c change. There is no production behavior or dependency change.

### DIFF
--- a/tests/test_cli.c
+++ b/tests/test_cli.c
@@ -226,6 +226,19 @@ static void restore_test_env(const char *name, char *saved) {
     }
 }
 
+TEST(cli_runner_uses_private_codex_home) {
+    const char *home = getenv("HOME");
+    const char *codex_home = getenv("CODEX_HOME");
+    ASSERT_NOT_NULL(home);
+    ASSERT_NOT_NULL(codex_home);
+
+    char expected[1024];
+    int expected_len = snprintf(expected, sizeof(expected), "%s/.codex", home);
+    ASSERT(expected_len >= 0 && (size_t)expected_len < sizeof(expected));
+    ASSERT_STR_EQ(codex_home, expected);
+    PASS();
+}
+
 /* Helper: mkdirp */
 static int test_mkdirp(const char *path) {
     char tmp[1024];
@@ -2354,10 +2367,13 @@ TEST(cli_detect_agents_finds_codex) {
     snprintf(dir, sizeof(dir), "%s/.codex", tmpdir);
     test_mkdirp(dir);
 
+    char *saved_codex = save_test_env("CODEX_HOME");
+    cbm_unsetenv("CODEX_HOME");
     cbm_detected_agents_t agents = cbm_detect_agents(tmpdir);
-    ASSERT_TRUE(agents.codex);
+    restore_test_env("CODEX_HOME", saved_codex);
 
     test_rmdir_r(tmpdir);
+    ASSERT_TRUE(agents.codex);
     PASS();
 }
 
@@ -2395,7 +2411,10 @@ TEST(cli_install_plan_receipt_no_mutation_issue388) {
     snprintf(dir, sizeof(dir), "%s/.codex", tmpdir);
     test_mkdirp(dir);
 
+    char *saved_codex = save_test_env("CODEX_HOME");
+    cbm_unsetenv("CODEX_HOME");
     char *json = cbm_build_install_plan_json(tmpdir, "/usr/local/bin/codebase-memory-mcp");
+    restore_test_env("CODEX_HOME", saved_codex);
     ASSERT_NOT_NULL(json);
     ASSERT(strstr(json, "agent.install.plan.v1") != NULL);
     ASSERT(strstr(json, "writes_started") != NULL);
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the concrete sentinel evidence. Tests writing into an inherited real CODEX_HOME are a high-priority 0.9.1-rc safety defect even though production behavior is unchanged. Review will verify that all normal suites receive the private test home, intentional fallback tests isolate and restore the override, and worker/probe modes still exit before suite-level environment mutation.

DeusData:

Reports as CONFLICTING against `main`. Checks are otherwise green, so a rebase should be all that is needed mechanically.

Isolating tests from user state is exactly the sort of thing that prevents mystery CI failures later.

DeusData:
Thank you — this is a good catch and a genuinely uncomfortable one.

The gap is real: production honours `CODEX_HOME` with a `~/.codex` fallback, but the test runner only replaced `HOME`, so CLI fixtures could reach into a developer's actual Codex install. A test suite that can mutate the machine it runs on is a bug regardless of whether anyone has been bitten yet.

Verified before merging: the buffer math is sound, the failure path cleans up the temp directory before returning, and the two fallback tests correctly save, unset and restore `CODEX_HOME` rather than leaking state into their neighbours. The new assertion binds — it fails if the setenv is reverted.

Merging. Note your own flag about overlapping with #1149 on the install-plan fallback test is right; whichever lands second will want a rebase that preserves both intents.

DeusData:
Correction to my previous comment: I went to merge this and GitHub reports the branch as conflicting with current `main`, so it did not go in. My apologies for saying "merging" before confirming that — the review stands, the mechanics did not.

This is almost certainly the overlap you flagged yourself with #1149 on the install-plan fallback test.

Could you rebase onto current `main`? Nothing about the change needs rethinking — required checks (`dco`, `ci-ok`) are green and the review above is complete, so once it applies cleanly it goes straight in.

Thanks again, and sorry for the noise.

DeusData:
Thank you for this, and for flagging the overlap with #1149 yourself in the description — that note is the reason these two got reviewed together instead of one quietly clobbering the other.

Reviewed and the approach is right. Of the two, **this is the one that matters**, and it should land first.

The reason: #1149 isolates `CODEX_HOME` for the issue388 fixture specifically. This PR does the general thing — a global `CODEX_HOME` sentinel in `test_main.c`, a regression test, and per-test unset in both `detect_agents` and issue388. Merging #1149 alone would fix one test while the rest of the suite kept reading and writing a developer's real `CODEX_HOME`. That is exactly the property we care about: a test's verdict has to be a function of the code and the platform, not of what happens to be in the person's home directory.

**What we need from you:** this branch is currently conflicting with `main`. The tree has moved a fair amount tonight — several fixes landed, including some in the test files. Could you rebase on current `main` and push? Once it is mergeable, this goes in.

We do not force-push or rebase contributors' branches on their behalf, so this one genuinely needs your hand on it.

After it lands we will ask @aaiyer — you — to reduce #1149 down to just its unique piece, the cleanup-before-FAIL restructure (free the JSON, restore the env, remove the temp dir, *then* fail), which is a real improvement this PR does not contain and which stops an assertion failure from leaking and littering `/tmp`. Alternatively, if you would rather fold that restructure into this PR and close #1149, that is completely fine by us — your call, and either way you get the credit for both.

Thanks for the patience; the delay here was ours.

DeusData:
Thank you — and an apology with the close: your strace report was the FIRST identification of the test-runner leaking into the developer's real ~/.codex, a month before we fixed it in-house. Main now handles it more broadly via 6d09c19e: a 22-entry tf_client_home_overrides table (CODEX_HOME included, tests/test_main.c:177) unset for every client home in tf_setup_cache_sentinel, plus per-test save/unset/restore in test_cli.c — which is also why this branch conflicts and why its regression test (asserting CODEX_HOME points into $HOME) now contradicts the unset design. Closing as superseded by that broader fix, with clear credit to your report for finding the class first.

PR #1163, opened 2026-07-18
fix: create the first persistent artifact on the incremental path

### LINKED ISSUE #434: cursor | vscode : persistence= true is silently ignored on first artifact creation
# `persistence: true` is silently ignored on first artifact creation

## Summary

`index_repository(persistence=true)` only writes `.codebase-memory/graph.db.zst` when an artifact **already exists** beside the repo. After the local cache `~/.cache/codebase-memory-mcp/<project>.db` is populated by an earlier call, every subsequent `index_repository` runs through the **incremental** pipeline, which silently ignores the `persistence` flag. As a result, a repo that has never been persisted before never gets its first artifact — even though the user explicitly asked for one.

The MCP response also gives no hint that the flag was a no-op:

- Returns `"status":"indexed"` (no error)
- Returns `"artifact_present":false` (truthful, but easily misread as "indexing went fine, just no artifact yet")
- Omits the usual `artifact_hint: "Persistent artifact written to .codebase-memory/graph.db.zst..."` field

## Reproduction

```bash
# Clean slate
rm -rf ~/.cache/codebase-memory-mcp
rm -rf /path/to/repo/.codebase-memory

# Call 1 — full pipeline runs (persistence:false here for clarity)
index_repository(repo_path="/path/to/repo", mode="full", persistence=false)
# → ~/.cache/codebase-memory-mcp/<project>.db is created
# → no artifact (correct, persistence wasn't requested)

# Call 2 — incremental pipeline runs (BUG: persistence:true ignored)
index_repository(repo_path="/path/to/repo", mode="full", persistence=true)
# Response:
#   "status":"indexed", "artifact_present":false   ← no artifact_hint
# Disk:
#   /path/to/repo/.codebase-memory/  ← still does not exist
```

Expected: `.codebase-memory/graph.db.zst` is written because `persistence=true` was passed.
Actual: the flag is dropped on the floor; no artifact is created.

## Real-world impact

I hit this while indexing 6 sibling repos in one chat session. First call per repo had `persistence=false` (default). On the follow-up "let's persist them all" pass with `persistence=true`, only the **one** repo that already had an artifact from a previous session got refreshed; the other 5 silently produced nothing. The only workaround I found was `delete_project` followed by `index_repository(persistence=true)` per repo — which forces the full pipeline and works correctly.

## Root cause

`src/pipeline/pipeline_incremental.c::dump_and_persist()`:

```c
/* Auto-update artifact if one already exists (persistence was enabled previously) */
if (repo_path && cbm_artifact_exists(repo_path)) {
    cbm_artifact_export(db_path, repo_path, project, CBM_ARTIFACT_FAST);
}
```

The function does not take a `persistence` parameter and the caller (`cbm_pipeline_run_incremental`, line ~657) doesn't pass one. So the incremental path can only **refresh** an artifact that already exists — it can never create the first one.

The full-pipeline path in `src/pipeline/pipeline.c::dump_and_persist_hashes()` does honor it:

```c
/* Export persistent artifact if enabled */
if (p->persistence) {
    cbm_artifact_export(db_path, p->repo_path, p->project_name, CBM_ARTIFACT_BEST);
}
```

This asymmetry between the full and incremental paths is the bug.

mvanhorn:
## What does this PR do?

Fixes #434: enabling `persistence=true` did not create the first persistent artifact on the incremental indexing path. `dump_and_persist` in `pipeline_incremental.c` only refreshed an artifact that already existed and never created one on first run, so a `persistence=true` request was silently a no-op until an artifact happened to exist. `git ai stats`-style consumers saw the index reported as successful with no persistent artifact written.

This threads a new `cbm_pipeline_persistence()` accessor into the incremental path (mirroring the existing `cbm_pipeline_repo_path` / `cbm_pipeline_project_name` accessors) so it exports the first artifact with `CBM_ARTIFACT_BEST` when persistence is requested and none exists, matching the full pipeline. Existing artifacts keep refreshing with `CBM_ARTIFACT_FAST`. It covers changed-file, deleted-file, and no-op incremental runs, and it now propagates a failed `cbm_artifact_export` as a run error instead of reporting a successful index when the artifact could not be written.

## Checklist

### DIFF
--- a/src/pipeline/pipeline_incremental.c
+++ b/src/pipeline/pipeline_incremental.c
@@ -1493,10 +1493,6 @@ static int dump_and_persist(cbm_gbuf_t *gbuf, const char *db_path, const char *p
         return rc;
     }
 
-    /* Auto-update artifact if one already exists (persistence was enabled previously) */
-    if (repo_path && cbm_artifact_exists(repo_path)) {
-        cbm_artifact_export(db_path, repo_path, project, CBM_ARTIFACT_FAST);
-    }
     return 0;
 }
 
@@ -2488,7 +2484,9 @@ int cbm_pipeline_run_incremental(cbm_pipeline_t *p, const char *db_path, cbm_fil
         free_mode_skipped(mode_skipped, mode_skipped_count);
         cbm_store_free_file_hashes(stored, stored_count);
         cbm_store_close(store);
-        return 0;
+        /* Same contract as the semantic_manifest_equal no-op above: a run that
+         * changed nothing still owes the first artifact when persistence is on. */
+        return cbm_pipeline_refresh_artifact(p, db_path);
     }
 
 #if defined(CBM_INCREMENTAL_TEST_API) && CBM_INCREMENTAL_TEST_API
@@ -2859,5 +2857,11 @@ int cbm_pipeline_run_incremental(cbm_pipeline_t *p, const char *db_path, cbm_fil
     cbm_gbuf_free(existing);
 
     cbm_log_info("incremental.done", "elapsed_ms", itoa_buf((int)elapsed_ms(t0)));
-    return persist_rc;
+    if (persist_rc != 0) {
+        return persist_rc;
+    }
+    /* Mirrors the full path: publication succeeded, so refresh the artifact.
+     * The helper creates the FIRST one when persistence is on and propagates
+     * an export failure instead of leaving the run looking successful. */
+    return cbm_pipeline_refresh_artifact(p, db_path);
 }


DeusData:
Thanks for covering both changed-file and no-op incremental paths and for propagating export failure rather than reporting a false success. This is classified with #434 as a 0.9.1-rc bug fix. Review will verify explicit persistence versus automatic refresh semantics, first-export quality selection, artifact failure behavior, and compatibility with the existing incremental artifact stack.

DeusData:
Thanks for this, and apologies for the slow acknowledgement. Queued for review.

Currently CONFLICTING against `main` with CI green (20 checks) — a rebase is the only mechanical blocker. The incremental-indexing area has moved recently, so rebasing will also tell you quickly whether the change still lands the way you intended.

mvanhorn:
The Windows job failure looks like a timing flake in an unrelated test rather than anything from this PR.

The run is 3669 passed / 1 failed, and the one failure is:

```
daemon_ipc_windows_rendezvous_bridges_concurrent_lifetime_owner
  FAIL tests/test_daemon_ipc.c:1163: ASSERT(startup_observed)
```

`startup_observed` is set by a bounded poll -- 200 iterations of `cbm_usleep(1000)`, so a fixed ~200ms budget -- waiting for a background thread to take `cbm-startup-v2.lock`. If the runner is loaded enough that the thread doesn't acquire it inside that window, the flag stays false and the assert trips. That Windows job took 17 minutes on this run, so a 200ms race losing is plausible. `ci-ok` then fails only because it aggregates `test`.

This PR touches `src/pipeline/pipeline.c`, `pipeline.h`, `pipeline_incremental.c` and `tests/test_pipeline.c` -- nothing in the daemon or IPC paths, and `test_pipeline` passes. A re-run should clear it; I can't trigger one.

If it's worth hardening, raising that poll budget or waiting on the lock rather than sampling it would make the test deterministic -- happy to send that separately.

DeusData:

The failure is `daemon_ipc_windows_rendezvous_bridges_concurrent_lifetime_owner` at `tests/test_daemon_ipc.c:1163` — a Windows daemon-startup interleaving flake. It was fixed on `main` by #1320, which landed after your run.

One thing worth knowing, because it will save you a wasted click: **"Re-run failed jobs" will not clear it.** A re-run re-tests the same recorded merge commit, so it will fail identically. Only a fresh push refreshes the merge ref — so a rebase on current `main` (or a `git merge origin/main`) and a push is what picks up the fix.

Apologies for the delay in getting to this. We are working through a large review backlog oldest-first, and the queue is real rather than an indication your PR was skipped.

mvanhorn:
Rebased onto current main and the diff got much smaller, so here is why rather than leaving it looking like work went missing.

Most of what this PR added now exists upstream as cbm_pipeline_refresh_artifact: it exports when persistence is on even with no artifact present, picks BEST over FAST for that first export, and returns the export failure instead of swallowing it. The full path and the semantic_manifest_equal no-op already call it.

What was still missing is the other two incremental exits. The no_changes no-op returned 0 without refreshing, so a run that changed nothing never created the first artifact, and the dump path hand-rolled an exists-only FAST export and discarded its result. Both now go through the helper.

5ed19a2, one file, +10/-6. Builds clean under -Werror and the focused artifact and pipeline suites pass, 265 tests, including the existing pipeline_persistence_export_failure_returns_error.

DeusData:
Thank you @mvanhorn for rebasing and explaining exactly what remained after upstream consolidation. The current diff is now one file, +10/-6, routing the two missed incremental exits through the existing refresh helper.

The only red is the macOS PR smoke at its repeated-`--semantic-query` output assertion; the emitted semantic table is present in the log, and the failure is outside the pipeline files changed here. I have recorded the narrowed head and returned it to the review queue without attributing that smoke failure to your patch. The queue is full, so final review may still take some time.

DeusData:
Closing in favor of #1263 (merged as 415a64a2) — with genuine thanks and credit: you identified the #434 incremental-persistence gap first, and your per-route analysis (the mode-skipped noop at :2493 and the real-run tail at :2864, beyond the one site 87a0e3f7 covered) is what made the architecture comparison possible at all. We ultimately chose the centralized publish-boundary export because its export-exactly-once property is enforced at one place instead of three, but that decision was informed by your mapping of the routes. If you're up for it, the #1263 design could still use eyes on the artifact-refresh edge you knew best.

PR #1175, opened 2026-07-19
fix(store): clamp ADR render cursor to prevent stack overflow

SEPURI-SAI-KRISHNA:
## What does this PR do?

Fixes a stack out-of-bounds write (CWE-787) in `adr_render_section`
(`src/store/store.c`), the helper `cbm_adr_render` uses to concatenate ADR
sections into a fixed 16 KB stack buffer (`buf[ST_MAX_DEGREE * ST_GROWTH]`).

The cursor was advanced by the return value of `snprintf`:

```c
pos += snprintf(buf + pos, buf_sz - pos, "\n\n");
pos += snprintf(buf + pos, buf_sz - pos, "## %s\n%s", key, value);
```

`snprintf` returns the number of bytes it *would* have written, so once the
accumulated sections exceed `buf_sz`, `pos` moves past the end. On the next
section `buf + pos` points past the buffer and `buf_sz - pos` (an `int`
subtraction) goes negative, converting to a huge `size_t` — so `snprintf`
writes out of bounds.

The fix clamps `pos` into `[0, buf_sz - 1]` after every write, so each
`snprintf` receives a positive size and the returned cursor never escapes the
buffer. This mirrors the clamping already done by `http_appendf` in
`src/ui/http_server.c` and the sibling accumulators in `store.c`
(`search_build_exclude_labels`, `bfs_build_types_clause`).

**Reproduced first (ASan, before the fix):**

```
==ERROR: AddressSanitizer: stack-buffer-overflow ... WRITE of size 3
    #3 adr_render_section src/store/store.c:7008
    #4 cbm_adr_render     src/store/store.c:7057
    #5 test_adr_render_oversized_sections_no_overflow tests/test_store_arch.c
SUMMARY: AddressSanitizer: stack-buffer-overflow ... in vsnprintf
```

After the fix the full `store_arch` suite (53 tests) passes with no sanitizer
errors.

Adds a regression test `adr_render_oversized_sections_no_overflow` that sizes
three sections so the cursor lands exactly `buf_sz + 8` after the second
section — the third section then performs the first out-of-bounds write right
at the buffer tail (ASan redzone), making the failure deterministic.

### Note on scope

`cbm_adr_render` / `cbm_store_adr_update_sections` are exported in `store.h`
and exercised by the test suite, but currently have no production caller — the
`manage_adr` MCP tool and the UI `/api/adr` endpoint both persist raw content
via `cbm_store_adr_store`. So this is a latent defect in exported API rather
than a live remote overflow; the fix hardens the renderer before anything
wires up the section-merge path.

## Checklist

### DIFF
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -7002,12 +7002,33 @@ cbm_adr_sections_t cbm_adr_parse_sections(const char *content) {
     return result;
 }
 
-/* Append a section to the render buffer. */
+/* Append a section to the render buffer.
+ *
+ * snprintf returns the length it WOULD have written, so a section larger than
+ * the space left would otherwise push pos past buf_sz; the next call would then
+ * compute a wrapped (huge) remaining size from (buf_sz - pos) and write out of
+ * bounds. Clamp pos into [0, buf_sz-1] after every write so each snprintf gets
+ * a positive size and the returned cursor never escapes the buffer. */
 static int adr_render_section(char *buf, int buf_sz, int pos, const char *key, const char *value) {
-    if (pos > 0) {
-        pos += snprintf(buf + pos, buf_sz - pos, "\n\n");
+    if (buf_sz <= 0) {
+        return 0;
     }
-    pos += snprintf(buf + pos, buf_sz - pos, "## %s\n%s", key, value);
+    if (pos < 0) {
+        pos = 0;
+    }
+    if (pos >= buf_sz) {
+        return buf_sz - SKIP_ONE;
+    }
+    if (pos > 0) {
+        pos += snprintf(buf + pos, (size_t)(buf_sz - pos), "\n\n");
+        if (pos >= buf_sz) {
+            return buf_sz - SKIP_ONE;
+        }
+    }
+    pos += snprintf(buf + pos, (size_t)(buf_sz - pos), "## %s\n%s", key, value);
+    if (pos >= buf_sz) {
+        pos = buf_sz - SKIP_ONE;
+    }
     return pos;
 }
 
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks for the reproduce-first hardening fix. Triage is complete: this is a high-priority `0.9.1-rc` stability/security item. Exact-head review at `13d523de484381cdbaff6fa52549836983cc88f0` found the change scoped to the exported ADR renderer and its deterministic ASan regression, with no new public surface or supply-chain changes. It remains in the maintainer review queue; this triage pass does not merge PRs.

DeusData:

MERGEABLE, all 20 checks green, +71/-3 across two files — nothing blocking on your side.

Clamping a render cursor to prevent stack overflow is exactly the kind of bounded, provable fix that is easy to review and easy to trust.

DeusData:
Thank you — merged.

I verified the overflow rather than trusting the writeup: in `adr_render_section` an unclamped cursor makes `buf_sz - pos` negative, which becomes an enormous `size_t` at the snprintf call. Your clamp keeps the returned cursor at `buf_sz - 1` and the NUL-termination invariant holds through truncation. I also re-derived your test's arithmetic by hand — the third append genuinely lands past the buffer into the ASan redzone pre-fix, so the test binds deterministically rather than probabilistically.

One thing I want to record accurately, because it affects how urgently anyone should backport this: `cbm_adr_render` and `cbm_store_adr_update_sections` currently have **no production callers** — I grepped `src/` to confirm. So this hardens a latent API rather than closing a live remote overflow. That makes it no less worth merging: the next caller would have inherited the bug, and now they will not.

Appreciated.

Issue #962, opened 2026-07-08
Option to index new filetypes


kenlyle2:
### What problem does this solve?

Feature Request: Configurable filetype/path indexing for codebase-memory-mcp

  Current behavior: index_repository (and therefore search_graph/search_code) only covers a fixed default source tree — in our case app/, lib/, components/, hooks/, and supabase/. Root-level one-off scripts (scripts/*.cjs, *.mjs) and research/documentation directories (research/) are silently excluded from indexing entirely.

  Impact: This isn't just a coverage gap — it's a silent one. A search_graph/search_code query for a symbol that genuinely exists in an unindexed script (confirmed case: scoreImageOnce() in scripts/model-bench-template.cjs) returns zero hits with no distinction from "this symbol doesn't exist anywhere in the repo." The tool has no way to say "not indexed" vs. "not found," so a user has to already know a file lives outside the graph's coverage before they know to fall back to grep — defeating the "use the graph first" workflow entirely for anything in these directories.

  This has become a real, elevated concern for us because our Playwright e2e test suite (a root-level .cjs script) has grown into what we consider critical infrastructure — the executable specification of an entire multi-agent content pipeline, not isposable test scaffolding. Bugs in it currently go uncaught by any indexed-code tooling, including this one.

### Proposed solution

  Requested feature: Expose configuration on index_repository (or a project-level config file) to specify additional include paths/glob patterns beyond the default source-tree assumption — e.g., --include
  "scripts/**/*.cjs,scripts/**/*.mjs,research/**/*.md" — so one-off scripts, benchmark harnesses, and research write-ups can opt into full graph coverage (search_graph, search_code, get_code_snippet, trace_path) on demand, without requiring them to live under the conventional app//lib install layout.

  Secondary ask: At minimum, have search_graph/search_code surface an explicit "N files matched but are outside indexed paths" signal when a query's zero-result set coincides with known-unindexed directories, so the silent-miss failure mode is at least visible rather than indistinguishable from a true negative.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
Thanks for the clear feature request. Triage: this is a real indexing-coverage/UX gap rather than a parser bug. The current silent-miss behavior is the important part: users cannot tell whether a symbol does not exist or lives under paths the indexer skipped. Putting this into 0.9.2-rc for configurable include paths and/or an explicit not-indexed signal in search results.

Issue #929, opened 2026-07-07
Windows: Claude hook installer creates extensionless hook files causing OpenWith dialog in Cursor


bsahafian:
### Version

codebase-memory-mcp 0.8.1

### Platform

Windows 10 x64

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### Summary

On Windows, installing `codebase-memory-mcp` creates extensionless hook files under:

```
%USERPROFILE%\.claude\hooks
```

Example:

```
cbm-code-discovery-gate
cbm-session-reminder
```

These files are Bash scripts (`#!/bin/bash`) without any executable extension.

When Cursor starts a new Agent, it scans the Claude hooks directory.

When Cursor starts a new Agent, the Windows "How do you want to open this file?" dialog appears.

Selecting Cursor opens one of the hook files (for example `cbm-session-reminder`) in the editor.

Removing or renaming `%USERPROFILE%\.claude\hooks` immediately eliminates the issue.

---

### Environment

- OS: Windows 10 x64
- Cursor: 3.10.17
- codebase-memory-mcp: 0.8.1
- Claude Code: **Not installed / Not used**

---

### Steps to reproduce

1. Install `codebase-memory-mcp` on Windows using the installer.
2. Verify that the installer creates:

```
%USERPROFILE%\.claude\hooks\
```

containing extensionless Bash scripts.

3. Open Cursor.
4. Create a new Agent.

---

### Expected behavior

If Claude Code is not installed or the platform is Windows:

- do not install Bash hook files
- or install Windows-compatible hooks (`.cmd` / `.ps1`)
- or skip Claude hook installation entirely.

No OpenWith dialog should appear.

---

### Actual behavior

Windows opens:

> How do you want to open this file?

If Cursor is selected, the hook source file itself is opened in the editor.

For example:

- cbm-code-discovery-gate
- cbm-session-reminder

---

### Suspected cause

The installer creates extensionless Bash hook files under `%USERPROFILE%\.claude\hooks`.

These files appear to trigger the OpenWith dialog during Cursor Agent startup on Windows.

I cannot say whether the underlying bug is in Cursor or in the installer, but the problem disappears as soon as the hook directory is removed or renamed.

---

### Workaround

Renaming either of these fixes the issue immediately:

```
%USERPROFILE%\.claude
```

or

```
%USERPROFILE%\.claude\hooks
```

Since I do not use Claude Code, disabling the directory completely resolves the problem.

---

[...]

PR #487, opened 2026-06-17
docs(config): document extra_extensions values and skip behavior

JOhnsonKC201:
Closes #359.

The README's **Custom File Extensions** section previously showed a single example key and stated unknown language values are "silently skipped". #359 asks which other config-file options exist and where they're documented.

This clarifies the config-file surface and fixes an inaccuracy:

- States that `extra_extensions` is the **only** key the JSON config files (`.codebase-memory.json` / `config.json`) accept, and points to the Environment Variables table and `config` subcommand for other tunables.
- Lists the **accepted, case-insensitive language names** (and aliases) from `LANG_NAME_TABLE` in `src/discover/userconfig.c`, so users know which values are valid.
- Documents the "extension must start with `.`" rule.
- Corrects the behavior note: invalid entries are **skipped with a warning logged to stderr** (`parse_extra_extensions` calls `cbm_log_warn("userconfig.unknown_lang", …)`), not silently.

Docs only — no code changes. Verified against `main` (`e599df1`).

PR #1203, opened 2026-07-21
fix(discover): detect extensionless scripts by shebang

### LINKED ISSUE #1199: bug(discover): extensionless scripts with valid shebangs are silently skipped
## Version

`v0.9.0` and current `main` at `7d6cdb23ef5ca2fd51f5d5b7509b33e112ef15f3`

## Platform

Linux (x64)

## Install channel

Built from source

## Binary variant

standard

An extensionless executable script is silently excluded from discovery even when its first line unambiguously declares a supported language:

```python
#!/usr/bin/env python3

def extensionless_probe():
    return 42
```

`cbm_language_for_filename("tool")` returns `CBM_LANG_COUNT`, and `detect_file_language()` returns immediately without inspecting the shebang. The file therefore produces no `File` or `Function` nodes, and graph/code searches are indistinguishable from a true negative.

Expected behavior: preserve existing special-filename and extension detection, but when they return unknown, inspect a bounded first line and infer a supported scripting language from a valid shebang.

## Reproduction

1. Create a dummy repository:

```bash
mkdir -p /tmp/cbm-shebang-repro/bin
cat >/tmp/cbm-shebang-repro/bin/tool <<'PY'
#!/usr/bin/env python3

def extensionless_probe():
    return 42
PY
chmod +x /tmp/cbm-shebang-repro/bin/tool
```

2. Index it:

```bash
codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/cbm-shebang-repro","mode":"full"}'
```

3. Search for `extensionless_probe` with `search_graph` or `search_code`.

Actual: zero results; the file is absent from discovery. Expected: Python `File`/`Function` nodes and a matching search result.

The executable bit is not essential to parsing; the shebang is the language signal and should also work on platforms without POSIX mode bits.

## Proposed fix

- Keep filename/extension mappings authoritative.
- Only when they return unknown, read at most the first line.
- Recognize direct interpreter paths plus `/usr/bin/env` forms, including `env -S`.
- Match interpreter basenames exactly (with version suffixes such as `python3.12` where applicable).
- Reject malformed/NUL-containing lines and read failures as unknown.
- Add reproduce-first discovery tests for direct Python, `env python3`, `env -S python3 -u`, precedence, and negative inputs.

## Related work checked

- #962 covers configurable include paths and a not-indexed signal, not language detection for a file reached by the walker.
- #306 fixed a missing extension mapping before content disambiguation, but still requires an extension.
- `extra_extensions` (#73/#487) maps extensions only and cannot represent an extensionless basename.
- Existing shebang PRs #674/#812 change project script portability; they do not add shebang-based source discovery.

## Confirmations

- [x] I searched existing open and closed issues/PRs and found no equivalent fix.
- [x] The reproduction uses a shareable dummy snippet, not proprietary code.

ibaldr89:
## Summary

- fall back to a bounded first-line shebang probe when filename/special-filename/extension detection returns unknown
- recognize direct interpreter paths plus `/usr/bin/env` and `env -S` for common supported scripting languages
- preserve filename/extension precedence and fail closed on malformed, NUL-containing, truncated, or unreadable first lines
- add production-discovery regressions for valid, conflicting, malformed, binary, env-assignment, and boundary cases

Fixes #1199

## Why

Executable source scripts commonly omit a filename extension. Before this change, files such as:

```python
#!/usr/bin/env python3

def extensionless_probe():
    return 42
```

were silently absent from the graph even though Python is supported. Full indexing did not change that because language detection was filename-only.

## Behavior

Existing detection remains authoritative. Shebang detection runs only when `cbm_language_for_filename()` returns `CBM_LANG_COUNT`.

Supported interpreter basenames:

- Python: `python`, `python2`, `python3`, dotted numeric versions such as `python3.12`
- shell: `sh`, `bash`, `dash`, `ksh`, `zsh`
- JavaScript: `node`, `nodejs`
- `ruby`, `perl`, `php`, `lua`

The probe reads at most 255 bytes plus one EOF/truncation probe byte, rejects embedded NUL, read errors, overlong first lines, unsupported `env` option/assignment shapes, and arbitrary prefix matches such as `python-wrapper`.

## Related work checked

- #73 and its documentation follow-up #487 add user-defined extension mappings, but those mappings require a dotted extension and cannot represent extensionless files.
- #962 concerns path/glob coverage and reporting for skipped files, not interpreter-based language detection.
- #929 / #1057 concern extensionless Windows installer hook scripts, not source discovery.

No existing issue or PR found in the checked open/closed history implements shebang-based language fallback for extensionless source files.

## Testing

- [x] New tests added
- [x] Focused discovery suite: **107 passed** under ASan/UBSan
- [x] CI lint stack: clang-format 20 + cppcheck 2.20 + NOLINT policy passed
- [x] GitHub Actions full matrix: **20/20 checks passed**, including Linux x86/ARM, macOS, Windows, TSan, CodeQL, DCO, and lint
- [x] Real-repository regression: **7/7** extensionless entrypoints indexed versus **0/7** before; expected function and call graph became queryable
- [x] Independent exact-tree review: approved for commit

Full suite comparison in the same isolated Docker environment:

- candidate: **6198 passed, 0 failed**
- clean `main`: **6180 passed, 0 failed**
- both runs exit non-zero on the same pre-existing 61-byte LeakSanitizer finding in `tests/test_httpd.c:358`; no candidate-only failure was introduced

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -651,7 +651,11 @@ static const char *file_skip_reason(const char *entry_name, const char *rel_path
 static CBMLanguage detect_file_language(const char *entry_name, const char *abs_path) {
     CBMLanguage lang = cbm_language_for_filename(entry_name);
     if (lang == CBM_LANG_COUNT) {
-        return CBM_LANG_COUNT;
+        /* Filename/extension detection failed: fall back to a conservative
+         * shebang probe so extensionless scripts get indexed (#1199). Filename
+         * detection stays authoritative — this runs only when it returns
+         * unknown. */
+        return cbm_language_from_shebang(abs_path);
     }
     /* Special: .m files need content-based disambiguation */
     const char *dot = strrchr(entry_name, '.');
--- a/src/discover/discover.h
+++ b/src/discover/discover.h
@@ -50,6 +50,20 @@ CBMLanguage cbm_disambiguate_cls(const char *path);
  * On read failure, defaults to CBM_LANG_BITBAKE. */
 CBMLanguage cbm_disambiguate_inc(const char *path);
 
+/* Detect a supported script language from a file's shebang (#!...) first line.
+ * Conservative fallback used only when filename/extension detection is unknown
+ * (see detect_file_language); it never overrides extension or special-filename
+ * matches. Opens the file read-only and reads only a bounded first line.
+ * Recognizes the interpreter *basename* (not the parent path):
+ *   python / python2 / python3 / dotted versions (python3.12) -> CBM_LANG_PYTHON
+ *   sh / bash / dash / ksh / zsh                              -> CBM_LANG_BASH
+ *   node / nodejs                                             -> CBM_LANG_JAVASCRIPT
+ *   ruby -> RUBY, perl -> PERL, php -> PHP, lua -> LUA
+ * Handles direct paths, "env <interp>", "env -S <interp> <args>", and CRLF.
+ * Fails closed (returns CBM_LANG_COUNT) on read error, missing/malformed
+ * shebang, an embedded NUL in the first line, or an unknown interpreter. */
+CBMLanguage cbm_language_from_shebang(const char *path);
+
 /* ── Gitignore pattern matching ──────────────────────────────────── */
 
 typedef struct cbm_gitignore cbm_gitignore_t;
--- a/src/discover/language.c
+++ b/src/discover/language.c
@@ -945,6 +945,172 @@ const char *cbm_language_name(CBMLanguage lang) {
     return LANG_NAMES[lang] ? LANG_NAMES[lang] : "Unknown";
 }
 
+/* ── Shebang interpreter detection (extensionless scripts) ────────── */
+
+/* Basename of an interpreter path: the segment after the last '/'.  Shebangs
+ * are a POSIX convention, so only '/' is treated as a separator. */
+static const char *interp_basename(const char *path) {
+    const char *slash = strrchr(path, '/');
+    return slash ? slash + SKIP_ONE : path;
+}
+
+/* "python" optionally followed by an explicit numeric version (digits and dots
+ * only, e.g. "python3", "python3.12").  Bounded and explicit so arbitrary
+ * suffixes like "python-wrapper" are rejected. */
+static bool is_python_interp(const char *base) {
+    if (strncmp(base, "python", SLEN("python")) != 0) {
+        return false;
+    }
+    const char *version = base + SLEN("python");
+    if (*version == '\0') {
+        return true;
+    }
+
+    /* Each numeric component must contain at least one digit. */
+    bool need_digit = true;
+    for (const char *v = version; *v; v++) {
+        if (isdigit((unsigned char)*v)) {
+            need_digit = false;
+        } else if (*v == '.' && !need_digit) {
+            need_digit = true;
+        } else {
+            return false;
+        }
+    }
+    return !need_digit;
+}
+
+/* Map an interpreter basename to a language, or CBM_LANG_COUNT if unrecognized.
+ * Non-python interpreters are matched exactly (no prefix/suffix logic). */
+static CBMLanguage lang_for_interpreter(const char *base) {
+    if (is_python_interp(base)) {
+        return CBM_LANG_PYTHON;
+    }
+    static const struct {
+        const char *name;
+        CBMLanguage lang;
+    } INTERP_TABLE[] = {
+        {"sh", CBM_LANG_BASH},           {"bash", CBM_LANG_BASH}, {"dash", CBM_LANG_BASH},
+        {"ksh", CBM_LANG_BASH},          {"zsh", CBM_LANG_BASH},  {"node", CBM_LANG_JAVASCRIPT},
+        {"nodejs", CBM_LANG_JAVASCRIPT}, {"ruby", CBM_LANG_RUBY}, {"perl", CBM_LANG_PERL},
+        {"php", CBM_LANG_PHP},           {"lua", CBM_LANG_LUA},
+    };
+    for (size_t i = 0; i < sizeof(INTERP_TABLE) / sizeof(INTERP_TABLE[0]); i++) {
+        if (strcmp(base, INTERP_TABLE[i].name) == 0) {
+            return INTERP_TABLE[i].lang;
+        }
+    }
+    return CBM_LANG_COUNT;
+}
+
+/* Advance *cursor past leading blanks and return the next whitespace-delimited
+ * token (NUL-terminated in place), or NULL when the line is exhausted. */
+static char *shebang_next_token(char **cursor) {
+    char *p = *cursor;
+    while (*p == ' ' || *p == '\t') {
+        p++;
+    }
+    if (*p == '\0') {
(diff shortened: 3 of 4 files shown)


DeusData:
Thanks — acknowledged here too, and I have replied in detail on #1199 confirming the underlying bug on current `main`.

Short version of that confirmation: `detect_file_language()` returns at `src/discover/discover.c:653-655` before any content inspection, and every content-based path in that function keys off an extension, so an extensionless file can never reach one. Your diagnosis was exactly right.

Queued for review. The branch is CONFLICTING against `main` — a rebase before I read it would be ideal.

DeusData:
Thank you — the implementation here is careful, and I want to say what I checked before raising anything: the probe is bounded and read-only, it uses `cbm_fopen` correctly, it fails closed on error, NUL bytes and truncation, the interpreter list is a fixed allowlist, nothing is executed, and CRLF is handled. The `env`/`-S`/assignment/option rejections are conservative in the right direction — false negatives rather than false positives. The 18 tests include negative and vacuity-guarded cases and they bind.

Two things before it can land.

**1. It needs a rebase** — it currently conflicts with `main`.

**2. One behavioural question I would like your view on.** The description says extension detection stays authoritative, but the fallback fires whenever `cbm_language_for_filename()` returns unknown — which includes files with an **unknown extension**, not only extensionless names. So `data.conf` beginning with a bash shebang would now index as Bash. That may well be what you intended, but it is broader than the stated contract.

There is a cost attached to that breadth: in FULL mode every unknown file gets an open and a read. `ALWAYS_IGNORED_SUFFIXES` keeps binaries out, but README, LICENSE, .txt, .zip and lockfiles would all be probed, and this project routinely indexes repositories in the millions-of-files range where that is measurable.

Either resolution is fine by me: gate the probe to genuinely extensionless basenames, or keep the breadth and say explicitly in the description why unknown extensions should participate and why the discovery I/O is acceptable at scale. I would just rather that be a decision than a side effect.

Worth noting for transparency: #1199 was filed by you a day before this PR, so the direction has not been through maintainer triage — but the fit is good. Finding code the graph currently misses is exactly the kind of coverage gap we want closed.

DeusData:
This is the best-engineered contribution I have reviewed in this backlog sweep, and I want to be specific about why rather than just saying so. **It needs one rebase and then I would like to merge it.**

**I measured the cost concern empirically rather than estimating it**, because "open and read every candidate file" is exactly the kind of change that quietly costs real time at scale. On the actual 94,851-file Linux kernel checkout:

- **~5,066 files get probed — 5.3% of the repo.** Everything else is filtered first: `Makefile` (×3,193) and `Kconfig` (×1,828) are filename-mapped and never touched, and gitignore, cbmignore, skip-lists and the size cap all run *before* the probe.
- **~15,200 extra syscalls**, three per candidate.
- **Well under half a second warm-cache against a ~260 s index — under 0.2%.** Even the pathological case (every file unknown) lands at a second or two.

And the over-inclusion question came out equally clean: **301 candidates have a shebang, ~294 newly indexed** — 196 `sh`, 67 `bash`, 18 `python3`, 13 `perl`, all genuine scripts in `scripts/` and `tools/` that were previously invisible to the graph. No junk. LICENSE and README have no shebang so they are never touched; `vendor`/`vendored` are always-skip so blobs are never probed; compiled artefacts do not start with `#!`. That is +0.31% files, all of them real source.

**The fail-closed discipline throughout is exemplary.** Every ambiguous input declines rather than guessing:

- a bounded 255-byte read plus a **one-byte EOF probe** to distinguish "short file" from "truncated first line" — a genuinely clever detail;
- embedded NUL before the newline → treated as binary, declined;
- `env NAME=value prog` and option tokens rejected, since `env` would set a variable rather than run that token;
- strict per-component digit matching on Python version suffixes, so `python3.` and `python3..12` are rejected and `python-wrapper` does not match by prefix;
- UTF-8 BOM before `#!` declines — which is *correct*, because the kernel refuses BOM'd shebangs too. Matching exec semantics rather than being liberal is the right instinct.

**You used `cbm_fopen` throughout — including in the test helper — without being told.** That rule exists because raw `fopen` breaks non-ASCII paths on Windows, and it is one people usually have to be reminded of. `"rb"` is also the right mode, with `\r` handled explicitly.

**The tests are what this repo actually asks for.** Eighteen of them, running end-to-end through the production `cbm_discover` walk rather than the unit function, red without the fix. Ten negative cases assert the file stays *unindexed*. Precedence is pinned both ways (`weird.py` with a bash shebang stays Python; `Makefile` with a python shebang stays Makefile). And the `shebang_probe` helper asserts setup success separately so a negative expectation cannot pass vacuously — that is precisely the discipline we care about, and most contributions miss it.

**What is needed: a rebase.** The branch conflicts with main, but the conflict is mechanical — one call-site line plus a self-contained function, and the early-return branch it hooks into is unchanged. The `test_discover.c` suite-list churn is mechanical too. Your PR body's "20/20 checks passed" predates the conflict, so CI needs re-establishing on the rebased head.

**One thing to look at after rebasing**, which is new since you branched: main gained `cbm_discover_count_bounded`, a deadline-bounded count path that also calls `detect_file_language` — so the probe cost now lands inside the coverage-count budget too. Counts staying consistent with real discovery is correct behaviour, and at these numbers the impact should be small, but it is worth one check on a very large repo.

**One optional follow-up, not for this PR:** the interpreter table is conservative and misses several languages we already support — `awk`/`gawk` (four real Linux files decline today), `Rscript`, `tclsh`, `fish`, `julia`, `elixir`, `groovy`, `pwsh`, `deno`/`bun`. Under-inclusion only, so it is safe as a v1, and a natural second PR.

Rebase and ping me; I will re-diff and merge.

ibaldr89:
Rebased onto current `main` (`d698db8e`) and force-updated the branch as requested. I also added a focused regression for the new `cbm_discover_count_bounded` path: full discovery and bounded counting now prove the same extensionless-shebang classification at the exact limit and one past it.

Local verification on the pushed tree `5fe1f4f8`:
- `clang-format-20 --dry-run --Werror ...`: pass
- ASan/UBSan focused discover suite: **112/112 passed**
- `git diff --check`: pass

GitHub matrix is running on the rebased commit. Ready for re-review.

ibaldr89:
Follow-up: the rebased head `5fe1f4f8` remains cleanly mergeable and all 28 GitHub checks are green. Is anything else needed from me before merge?

DeusData:
@ibaldr89, nothing further is needed from you before maintainer re-review. I rechecked head `5fe1f4f8`: the branch is cleanly mergeable and all 28 GitHub checks are green, including the platform matrix, DCO, CodeQL, and `ci-ok`.

The next action is ours. Unless `main` moves in a way that makes the branch conflict again, please do not spend more time rebasing or adding validation. Thank you for following through on the requested rebase and the bounded-count regression, and sorry you had to ask again.

DeusData:
Merged — thank you! The fail-closed probe design (bounded 256 bytes, NUL/truncation refusal, cbm_fopen per house rule, extension detection staying authoritative) plus eighteen tests including the adversarial negatives made this easy to trust. Findability-first is exactly this project's bar and shebang-only scripts were a real blind spot. #1199 closed.

PR #1204, opened 2026-07-22
fix(mcp): survive atomic binary replacement

DavidMusk93:
## Summary

- prefer the saved executable launch path when it still points to an executable
- reject a deleted `/proc/self/exe` path before spawning index workers
- add a Linux integration regression that forks, atomically replaces the running test binary, and verifies subprocess resolution

## Tests

- `build/c/test-runner httpd`: 44 passed, 1 Windows-only skip
- `scripts/run-tests-parallel.sh build/c/test-runner 16`: 6386 passed, 1 unrelated baseline failure, 1 Windows-only skip

The baseline failure is `tests/test_cli.c:6781` (`installer must not follow symlinked agent roots outside the selected home`). It is unrelated to the three files in this PR and is fixed by a separate local change, intentionally excluded here.

## Runtime verification

A static build carrying this patch was loaded by both host and container MCP instances. A real `index_repository` call completed with 17,237 nodes, 99,775 edges, `skipped_count=0`, and `parse_partial_count=0`.

### DIFF
--- a/src/ui/http_server.c
+++ b/src/ui/http_server.c
@@ -962,9 +962,18 @@ bool cbm_http_server_resolve_binary_path(const char *argv0, char *out, size_t ou
     }
 #endif
 
+#ifndef _WIN32
+    if (g_binary_path[0] && is_executable_file(g_binary_path)) {
+        return copy_path(out, outsz, g_binary_path);
+    }
+    if (resolve_self_executable(out, outsz) && is_executable_file(out)) {
+        return true;
+    }
+#else
     if (resolve_self_executable(out, outsz)) {
         return true;
     }
+#endif
     return copy_path(out, outsz, argv0);
 }
 
--- a/src/ui/http_server.c
+++ b/src/ui/http_server.c
@@ -1053,11 +1053,29 @@ bool cbm_http_server_resolve_binary_path(const char *argv0, char *out, size_t ou
 #endif
 
 #ifndef _WIN32
-    if (g_binary_path[0] && is_executable_file(g_binary_path)) {
-        return copy_path(out, outsz, g_binary_path);
-    }
-    if (resolve_self_executable(out, outsz) && is_executable_file(out)) {
-        return true;
+    /* #1204: never hand back a self-resolved path we have not confirmed is
+     * executable. After an installer's atomic rename-over, the resolved image
+     * path no longer exists (Linux readlink reads "<path> (deleted)"), and an
+     * unvalidated return turns into a doomed worker spawn (ENOENT). */
+    if (resolve_self_executable(out, outsz)) {
+        if (is_executable_file(out)) {
+            return true;
+        }
+#if defined(__linux__)
+        /* Deleted image: the /proc/self/exe magic link still executes the
+         * in-memory OLD build — the only spawn that satisfies the worker's
+         * build-fingerprint gate (a rename-over leaves the on-disk path
+         * holding a DIFFERENT build). Hand back the link, not the stale
+         * path. Deliberately NOT the saved launch path: preferring it swaps
+         * an ENOENT for a fingerprint refusal. */
+        return copy_path(out, outsz, "/proc/self/exe");
+#else
+        /* macOS has no magic link. Fail closed: the supervisor logs
+         * index.supervisor.no_self_path and indexing resumes after a daemon
+         * restart, instead of spawning a missing or mismatched binary. */
+        out[0] = '\0';
+        return false;
+#endif
     }
 #else
     if (resolve_self_executable(out, outsz)) {
@@ -1069,7 +1087,15 @@ bool cbm_http_server_resolve_binary_path(const char *argv0, char *out, size_t ou
 
 void cbm_http_server_set_binary_path(const char *path) {
     if (path) {
-        if (!cbm_http_server_resolve_binary_path(path, g_binary_path, sizeof(g_binary_path))) {
+        /* Resolve into a local buffer first: `path` may alias g_binary_path
+         * on a repeated call. Resolving in place first hits resolve's leading
+         * out[0]='\0', which zeroes the shared buffer and silently discards
+         * the very path we were asked to re-resolve — and the in-place write
+         * is one reset-reorder away from an overlapping snprintf. */
+        char resolved[sizeof(g_binary_path)];
+        if (cbm_http_server_resolve_binary_path(path, resolved, sizeof(resolved))) {
+            snprintf(g_binary_path, sizeof(g_binary_path), "%s", resolved);
+        } else {
             g_binary_path[0] = '\0';
         }
     }
--- a/src/ui/http_server.c
+++ b/src/ui/http_server.c
@@ -1052,9 +1052,18 @@ bool cbm_http_server_resolve_binary_path(const char *argv0, char *out, size_t ou
     }
 #endif
 
+#ifndef _WIN32
+    if (g_binary_path[0] && is_executable_file(g_binary_path)) {
+        return copy_path(out, outsz, g_binary_path);
+    }
+    if (resolve_self_executable(out, outsz) && is_executable_file(out)) {
+        return true;
+    }
+#else
     if (resolve_self_executable(out, outsz)) {
         return true;
     }
+#endif
     return copy_path(out, outsz, argv0);
 }
 
--- a/tests/test_httpd.c
+++ b/tests/test_httpd.c
@@ -373,6 +373,69 @@ TEST(httpd_resolves_bare_binary_path_from_path) {
 #endif
 }
 
+TEST(httpd_resolves_deleted_self_to_replaced_launch_path) {
+#if defined(__linux__)
+    char tmpdir[256];
+    snprintf(tmpdir, sizeof(tmpdir), "/tmp/cbm_httpd_deleted_self_XXXXXX");
+    ASSERT_NOT_NULL(cbm_mkdtemp(tmpdir));
+
+    char current[1024];
+    ssize_t current_len = readlink("/proc/self/exe", current, sizeof(current) - 1);
+    ASSERT_GT(current_len, 0);
+    ASSERT_LT(current_len, (ssize_t)(sizeof(current) - 1));
+    current[current_len] = '\0';
+
+    char launch_path[512];
+    char replacement_path[512];
+    snprintf(launch_path, sizeof(launch_path), "%s/test-runner", tmpdir);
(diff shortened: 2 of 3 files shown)


DeusData:
Thanks for this, and sorry for the slow acknowledgement. Queued for review.

Two things to clear first: the branch is CONFLICTING against `main`, and CI shows 2 failing checks. A rebase is worth doing before diagnosing the failures, since some may be drift rather than the change.

Surviving atomic binary replacement is a real scenario — self-update and package-manager upgrades both hit it — so this is a welcome area to harden.

DeusData:
Reviewed in depth — and I want to be upfront that the news is awkward: the work is good, the bug is real, and a rebase alone will **not** restore what your fix does. That needs a maintainer decision, which I have raised.

**Your diagnosis is correct and I verified it in source.** `cbm_http_server_resolve_binary_path` skips the argv0 checks when called with NULL from the index supervisor, goes straight to `resolve_self_executable`, and on Linux that is `readlink("/proc/self/exe")`. After the installer's atomic rename-over, that returns `"/path/to/bin (deleted)"` — a path that does not exist — and `http_server.c:963` hands it back **unvalidated**, so the worker spawn fails with ENOENT and `index_repository` breaks from a long-running server. Corroborating detail I liked: `g_binary_path` is currently **write-only** on main — set in three places, read nowhere. Your change makes it meaningful again.

**The integration test is exactly the right shape.** Copy the running binary to a launch path, fork and exec it as a probe, pipe-handshake, atomically `rename()` a replacement over it, then assert resolution. That is genuine reproduce-first work, and it is binding on Linux: without the source change the probe resolves the dead `(deleted)` string and exits 44.

**What changed underneath you.** Your base is `7d6cdb2` — **299 commits** behind current main. The daemon rework since then added a **build-fingerprint gate**: the supervisor passes a content hash of its *own running image* (`src/mcp/index_supervisor.c:603`), and the spawned worker refuses on mismatch (`index_supervisor.c:216-219`, fail-closed by design).

Your fix prefers the saved launch path — which, after replacement, holds the **new** build. So the old daemon would spawn the new binary, and that gate would reject it. The net user-visible outcome on current main would still be a failure, just a different one: a deliberate fingerprint refusal instead of an ENOENT. That is why this is a semantic rebase rather than a mechanical one.

**A fingerprint-compatible variant exists on Linux:** spawn the *literal* `/proc/self/exe` rather than resolving it to a string. The magic link execs the original inode even after deletion, so parent and worker stay the same build and the gate is satisfied. macOS has no equivalent once the old inode is unlinked, which pushes that platform toward daemon-restart-on-replacement instead.

Which of those we want is a maintainer call — same-build via the magic link, fail-closed until restart, or new-build workers as you have it — so I have put the question upstream rather than deciding it in review. **Please hold off reworking until there is an answer**; a 299-commit rebase is a lot to spend on semantics that might change.

**One thing is worth keeping regardless of that decision:** the `is_executable_file` guard on the self-resolved path. Returning a non-existent path unvalidated is wrong under any of the three designs.

Two smaller notes. There is a latent aliasing bug: `cbm_http_server_set_binary_path` resolves *into* `g_binary_path` while your new code reads it, so a second call with a non-directly-executable path would `snprintf` a buffer into itself — overlapping copy, undefined behaviour. Resolving into a local, or guarding `out != g_binary_path`, avoids it. And the test is `#if defined(__linux__)` with a silent PASS elsewhere, so the title's "survive atomic binary replacement" is Linux-only in substance.

Thank you for this — the reproduction is the hard part and you did it properly.

DeusData:
The design here is right and still needed — main's resolve path still never re-reads the saved launch path (g_binary_path is write-only at http_server.c:987) and has no deleted-image handling, so worker spawns after an atomic binary replacement remain broken. The argv-guarded execl self-probe mirroring the #798 socket-probe pattern is a legitimate test idiom. What it needs is a rebase: http_server.c moved under you (d58afe56/8018561c) and the daemon_runtime test suite grew conventions your new Linux test should adopt. The CodeQL red on your run matches the known 45-minute gate-timeout pattern, not a finding. Rebase and we'll re-review promptly — this fixes a real operational hole.

DeusData:
Re-verified against current `main`, and I have a proposal that unblocks half of this without waiting on the design call I still owe you.

**The hole is still open, with the same shape you diagnosed.** `g_binary_path` is declared at `http_server.c:957`, written at `:1063-1064`, and read nowhere in `src/` — still write-only. `cbm_http_server_resolve_binary_path` is called with `argv0 == NULL` from `index_supervisor.c:638`, so the argv0 branch is skipped, `resolve_from_path` returns false on a NULL name, and it hands back `resolve_self_executable()`'s output **unvalidated**. That function is a bare `readlink("/proc/self/exe")` with no existence or executability check, and there is no `(deleted)` handling anywhere in the tree. Nothing has superseded this.

Your test is genuinely reproduce-first: without the change the probe's `strcmp(resolved, argv[4])` fails against the `" (deleted)"`-suffixed string, it exits 44, and the parent's `ASSERT_EQ(WEXITSTATUS(status), 0)` goes red.

**The proposal: split it, because the two halves have very different risk profiles.**

*Half A — the `is_executable_file` guard on the self-resolved path.* Uncontroversial, not superseded, and it stands on its own merit: returning a path we have not confirmed exists is wrong regardless of which spawn strategy we settle on. It turns a doomed spawn into an honest fail-closed with `index.supervisor.no_self_path`. I would take this now.

*Half B — preferring the saved launch path.* This is the part that collides with the build-fingerprint gate added after your base. The supervisor requires a fingerprint before spawning (`index_supervisor.c:631-635`) and the worker refuses on mismatch (`:247-248`, `CBM_INDEX_WORKER_ARGV_BUILD_MISMATCH`). After an atomic replacement the saved path holds the *new* build, so preferring it swaps an ENOENT for a deliberate fingerprint refusal. Better error, still broken.

Splitting means Half A lands while I resolve the strategy question, instead of both sitting behind it.

**Two things to fix in Half A regardless of the split.**

1. **Aliasing UB.** `cbm_http_server_set_binary_path` (`:1061-1067`) calls resolve with `out == g_binary_path`. If your new first branch fires you get `snprintf(g_binary_path, sz, "%s", g_binary_path)` — overlapping source and destination, which is undefined. Reachable on a second `set_binary_path` call with a path that is neither directly executable nor PATH-resolvable.
2. **Coverage asymmetry.** The test is `#if defined(__linux__)` with a silent `PASS()` elsewhere, but the production edit is `#ifndef _WIN32` — so macOS takes the behaviour change with zero coverage. Either widen the guard to Linux-only or add a macOS path; a silent PASS on a platform that received the change is the shape that hides regressions.

**The strategy decision is still mine and still owed to you**, and I am sorry it has been a month. The three options remain: spawn the literal `/proc/self/exe` magic link (same build, satisfies the fingerprint, Linux-only), fail closed until daemon restart, or accept new-build workers. You were right to stop and ask rather than guess, and you should not have had to wait this long for the answer.

One note on the Copilot review comment: it argued your `readlink` assertions permit a truncated return. I think it is wrong — you assert both `ASSERT_GT(current_len, 0)` and `ASSERT_LT(current_len, (ssize_t)(sizeof(current) - 1))` and called `readlink` with `sizeof(current) - 1`, so truncation is rejected. No action needed.

Would you like to split it, or would you rather wait for the strategy call and do it once?

DeusData:
The strategy decision I have owed you since 31 July, finally — and with it, the work done rather than another request.

**The ruling: on Linux, spawn the literal `/proc/self/exe` magic link; on macOS, fail closed until daemon restart.** The saved-launch-path preference from your original head is not taken — after a rename-over that path holds a *different* build, so preferring it converts the ENOENT into a build-fingerprint refusal: a better error, but not survival. The magic link is the one spawn that still executes the in-memory old build, which is exactly what the fingerprint gate exists to demand. One more reason the accept-new-build option was never viable on today's main: since #1104, the index carries a format version, so a new-build worker could rebuild the store to a format the still-running old daemon cannot serve.

**Because you were told to hold off — and did, to the letter — I have implemented it rather than hand you a month-late answer plus a work order.** Your branch now carries two commits:

1. **Your commit, rebased, authorship and sign-off intact.** Your diagnosis was right, your `is_executable_file` guard survives as written, and your fork/exec/rename-over test remains the backbone — it was, as I said in review, exactly the right shape.
2. A maintainer commit on top implementing the ruling, with you as co-author. The probe now proves the `(deleted)` state via `readlink` before asserting (so it cannot pass vacuously on an intact image), asserts the magic-link contract on Linux, and asserts fail-closed on macOS — which also closes the coverage asymmetry from the 21 August list: the test now runs on every platform the production change touches.

One correction to my own 21 August review, in fairness to your original code: I claimed the `set_binary_path` aliasing was reachable UB. It is not — the resolver's leading `out[0]='\0'` zeroes the shared buffer before the argv0 branch can read it. The real defect is quieter: that same reset silently discards the path the caller asked to re-resolve. The fix (resolve into a local) is unchanged; the characterisation in the record is now accurate.

Verified before pushing: macOS 63/63 with a revert-check (main's resolver restored → the driver fails at exit 44), and the full Linux arm64 leg green — 7,461 tests, 0 failures — at this exact tip. CI runs now; I will merge when the matrix confirms.

Thank you for a textbook contribution: a real operational bug, a correct diagnosis, a binding reproduce-first test, and the discipline to stop and ask when the design was genuinely not yours to guess. I am sorry the answer took a month.

DeusData:
Merged in 2cc80f20. Thank you, @DavidMusk93 — this closes a real operational hole that every self-update and package upgrade walked past: a daemon that outlived its own binary could no longer spawn index workers, and the failure surfaced as a bare ENOENT.

Both commits are in: yours exactly as authored, and the strategy commit on top implementing the ruling you were owed since July — on Linux the resolver now hands back the `/proc/self/exe` magic link (the in-memory old build, the only spawn the fingerprint gate accepts), and on macOS it fails closed with `index.supervisor.no_self_path` until the daemon restarts, rather than spawning a missing or mismatched binary.

Your reproduce-first test is the backbone of the verification: it now proves the `(deleted)` state via `readlink` before asserting, so it cannot pass vacuously, and it ran green on macOS (63/63 with a revert-check failing at exit 44 against the unguarded resolver), the full Linux arm64 leg (7,461/0), and the complete CI matrix including TSan. The one red on the way in was a known watcher flake unrelated to this change.

Diagnosing this correctly, writing a binding test for it, and then stopping to ask when the design genuinely wasn't yours to guess — that is exactly how a hard contribution should run. Sorry again that the answer took a month.

PR #1232, opened 2026-07-23
fix(parser): use alias-safe Haskell arrays

### LINKED ISSUE #1231: Parallel indexing can corrupt the heap in the Haskell scanner
### Version

codebase-memory-mcp dev at 5e7d3eb77e355179d07ff5236bd5bcda4448b81c

### Platform

Linux x64, built from source, UI variant, GCC 16.1.1

### What happened

Indexing a mixed-language repository with two or more workers aborts with glibc heap-metadata corruption. One worker completes. Parallel indexing should complete without changing worker count or disabling concurrency.

### Reproduction

1. Check out oraios/serena at dd7eb6d72ae179aa940e50cd6276ec5646f306f8.
2. Build the production binary with scripts/build.sh --with-ui.
3. Run index_repository against that checkout with CBM_WORKERS=2.
4. The index worker aborts during extraction. The same corpus also reproduces with 4 and 20 workers.

Valgrind with the production O2 build reports the first invalid write in advance -> take_line -> newline_lookahead -> tree_sitter_haskell_external_scanner_scan. The write uses the old contents pointer immediately after realloc moved the Lookahead buffer. The pinned Haskell tree_sitter/array.h passes differently typed Array structs through Array(void) pointers, violating strict aliasing; Tree-sitter fixed this class upstream in tree-sitter/tree-sitter#5242.

### Expected

The existing parallel worker count remains enabled and the repository indexes successfully.

### Scale

10,138 nodes and roughly 50,800 edges in the isolated reproduction corpus.

### Confirmations

- Searched existing issues; no duplicate found.
- Reproduction uses the public oraios/serena repository.

WarGloom:
## What does this PR do?

Fixes the optimized-build heap corruption in the pinned Haskell external scanner without reducing indexing parallelism.

The scanner vendors an older Tree-sitter array helper that type-puns concrete Array structs through Array(void) pointers. At -O2, the Haskell lookahead growth path can retain the old contents pointer across realloc and write into freed memory. Valgrind identified the first invalid write in advance -> take_line -> newline_lookahead -> tree_sitter_haskell_external_scanner_scan.

The Haskell grammar wrapper now preloads the repository existing strict-aliasing-safe Tree-sitter array helpers, using the shared include guard to skip the unsafe pinned copy. A C11 static assertion locks the helper signature so the unsafe implementation cannot silently return.

Fixes #1231

## Validation

- Reproduced before the fix on oraios/serena at dd7eb6d72ae179aa940e50cd6276ec5646f306f8 with 2, 4, and 20 workers.
- Production -O2 indexing passed after the fix on fresh copies with 2, 4, and 20 workers: 10,138 nodes each.
- Sanitized suite: 6,780 passed, 0 failed, 4 skipped across 121 suites.
- Production parent and worker watchdog tests passed.
- make -f Makefile.cbm lint-ci passed.
- scripts/build.sh --with-ui passed.
- Full local scripts/lint.sh was also attempted; its clang-tidy leg reports existing findings in untouched files under the local toolchain, while the CI lint target above passes.

## Checklist

- [x] Every commit is signed off (git commit -s) — required, CI rejects unsigned commits ([DCO](../DCO), see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] Tests pass locally (make -f Makefile.cbm test)
- [x] Lint passes (make -f Makefile.cbm lint-ci)

### DIFF
--- a/internal/cbm/grammar_haskell.c
+++ b/internal/cbm/grammar_haskell.c
@@ -1,4 +1,8 @@
 // Vendored tree-sitter grammar: haskell
 // Each grammar compiled as separate unit (conflicting static symbols).
+#include "vendored/common/tree_sitter/array.h" // strict-aliasing-safe helpers (tree-sitter#5242)
+typedef void *(*cbm_haskell_array_grow_fn)(void *, uint32_t, uint32_t *, uint32_t, size_t);
+_Static_assert(_Generic(&_array__grow, cbm_haskell_array_grow_fn: 1, default: 0),
+               "Haskell scanner requires strict-aliasing-safe array helpers");
 #include "vendored/grammars/haskell/parser.c"
 #include "vendored/grammars/haskell/scanner.c"


DeusData:
Thanks for the very small and plausible fix. Reusing the already-vendored strict-aliasing-safe Tree-sitter array helper is the right shape, introduces no new vendor or dependency, and the head is fully green. The remaining blocker is reproduce-first coverage: no test changed, and the `_Static_assert` only protects the selected helper signature, not the reported optimized multi-worker heap corruption. Please add a minimized Haskell fixture/runtime test that fails on current main under the relevant optimized strict-aliasing build and passes with this patch.

DeusData:
Thanks, and sorry for the slow acknowledgement. Queued for review.

A four-line change with 28 green checks is about as reviewable as a contribution gets — nothing blocking on your side.

Alias-safe Haskell arrays is a precise fix to a real parsing hazard.

DeusData:
Thank you — merged, and this one was genuinely satisfying to verify.

The mechanism checks out end to end. Haskell's pinned `tree_sitter/array.h` carries the old type-punning `_array__grow` — the known upstream strict-aliasing bug (tree-sitter#5242) where -O2 can keep a stale `contents` pointer across a realloc. Because the common copy uses the **same include guard**, preloading it makes the scanner's own include a no-op, so both the macros and the helper come from the alias-safe version. Four lines, no third-party code added, and the only include is an in-repo header we already vendored with its LICENSE.

The `_Static_assert` over `_Generic(&_array__grow, …)` is the part I liked most: it matches the safe signature exactly, so if a future grammar refresh ever restores the unsafe copy the build breaks instead of silently regressing to a heisenbug. Your checklist says the change is "covered by a test", which slightly overstates it — there is no runtime repro, and realistically there cannot be a practical one in-suite since it needs -O2 plus a specific corpus. The static assert is the real binding guard, and it is the right one.

I am adding a line to `internal/cbm/vendored/grammars/MANIFEST.md` recording this override, so that a future `refresh-grammars` pass does not quietly drop the preload. The longer-term cleanup is to refresh the Haskell grammar from upstream, which has since fixed this — your shim is the correct minimal fix in the meantime.

Issue #1197, opened 2026-07-21
UI: Left sidebar directory selection does not highlight 3D points or update detail panel


Maystern:
### What problem does this solve?

In the current UI, selecting an item from the left sidebar directory fails to:
  1. Highlight/select the corresponding 3D point in the visualization.
  2. Update the right-side detail panel (it still shows data from the previously selected point)
The user must instead click the point directly in the 3D viewport to see its call relationships on the right. This
  breaks the expected navigation flow and makes it hard to explore the code graph and spot issues.

### Proposed solution

When the user selects an entry in the left sidebar directory, the UI should:

  1. Automatically select/highlight the corresponding 3D point in the central viewport.
  2. Automatically update the right-side panel to display that point's call relationships.

  This would restore the expected navigation behavior — left sidebar → 3D viewport → right detail panel — and make the code graph much easier to explore.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
Thanks for the clear interaction sequence. I have classified this as a `0.9.1-rc` UI behavior bug under #396 rather than a new feature: selecting an existing graph item in the directory tree should update the shared selection state, 3D highlight, and detail panel consistently.

DeusData:
Fixed on `main` in 4149c504 (#1235) by @Maystern.

All three shapes of the report are covered and regression-tested: sidebar leaf clicks now open the detail panel through the same canonical path as canvas clicks (selection, neighbour highlight, camera), directory clicks clear stale details instead of showing the previously selected node, and search results without a `file_path` no longer silently do nothing. Verified locally: the new tests fail 4/5 against the pre-fix components and pass 5/5 with the fix.

Thank you for the precise report — the leaf-vs-directory distinction in it is what shaped the final design.

PR #1235, opened 2026-07-23
fix(sidebar): enhance node selection to show details for single nodes

Maystern:
### What

Fixes sidebar node selection in the graph UI:

- **Directory clicks now clear the details panel** instead of leaving stale details from a previously selected node (previously `handleSelectPath` never reset `selectedNode`, so the panel kept showing the last selected file)
- **Search results without a `file_path` can now open the panel** — previously they were dropped by an early `if (!path) return`, so clicking them did nothing
- **Unified selection through one code path**: `handleSelectPath` now takes an optional node and delegates to `handleNodeClick`, so selecting from the sidebar, search results, and the graph all share the same selection/highlighting behaviour (and leaf clicks now highlight direct neighbours, matching graph clicks)

### Tests

Adds `GraphTab.selection.test.tsx` (168 lines, 5 cases) covering leaf clicks, directory clicks, file_path-less search results, clearing selection, and neighbour highlighting — the leaf-click test fails against the previous implementation.

### Misc

Drops the accidentally committed build artifact `src/ui/embedded_assets.c` (generated by `scripts/embed-frontend.sh`; contains machine-specific asset hashes).

## Checklist


DeusData:
Thanks for addressing #1197. Two selection cases remain incorrect in this head. `nodeIds.size === 1` does not distinguish a leaf from a directory containing one node, so a one-node directory opens details even though Sidebar passes no node for directories. Search results without `file_path` still pass an empty path and clear selection, while the new `GraphNode` callback argument is ignored. The patch also duplicates the existing `handleNodeClick` neighbor and camera behavior. Please route explicit leaf/search-node selection through one canonical node-selection handler and add RED tests for leaf click, one-node directory, search result without `file_path`, clear selection, and neighbor highlighting.

DeusData:
Thank you for this — the UI fix itself is good and I would like to merge it. One file needs to come out first, and it is not your logic.

**What is right.** `handleSelectPath` taking an optional node and delegating to `handleNodeClick` is the correct shape: it sets the selected node and the neighbour highlight through the existing path rather than duplicating that behaviour. Two real bugs fall out of it — directory clicks now clear the panel instead of leaving stale details from a previously selected node, and search results with no `file_path` can finally open the panel at all. The 168-line vitest suite is solid, and the leaf-click test genuinely fails against the old code.

Security-wise it is clean: everything renders through React text nodes, no `innerHTML`, so none of the graph-derived strings (which come from indexed repositories and are therefore attacker-influenced) reach the DOM as markup.

**What needs to change: please drop `src/ui/embedded_assets.c` from the diff.**

That file is a build artifact — it carries a "DO NOT EDIT — regenerated by `scripts/embed-frontend.sh`" header, and the copy in this PR contains asset hashes from your machine (`index-D27RSPSK.css`). It is harmless to the build, because the default build uses `embedded_stub.c` and the with-UI build regenerates the file first, which is why the smoke legs are green. But if it becomes tracked, every with-UI build will dirty the working tree with rotating hashes, forever, for everyone.

That is our gap, not yours — the file should have been in `.gitignore` and was not. We will fix that on our side. Removing it here is a one-commit change:

```
git rm --cached src/ui/embedded_assets.c
git commit -s -m "chore: drop generated embedded_assets.c from the diff"
git push
```

**One small ask:** the PR description is currently the empty template. A sentence or two on the behaviour you were fixing would help the next person reading the history — and would have made this review faster.

With the artifact removed, this merges. Thanks for the clean test coverage; it is the reason the review was straightforward.

DeusData:
Merged in 4149c504. Thank you, @Maystern — a first contribution that went through two full review rounds and came out the other side better each time is exactly how this is supposed to work.

Worth naming what you did right, because each one is rarer than it should be: when round 1 said the `nodeIds.size === 1` heuristic couldn't distinguish a leaf from a one-node directory, you didn't patch the heuristic — you removed it, making the distinction explicit at the call sites and delegating to the one canonical `handleNodeClick` path. When an embedded artifact slipped into the branch, you executed the removal exactly as prescribed and moved on. And the five regression tests are the maintainer's demanded cases one-for-one.

Before merging I ran what CI currently cannot: the frontend vitest suite doesn't run in any CI leg (a gap on our side, now tracked), so I executed your tests locally in both directions — **5/5 pass with your fix, and 4/5 fail against main's unfixed components** with only the already-working clear-selection case passing. That is a genuinely binding regression suite, and the full 10-file frontend suite is 43/43 with your change in.

#1197 closes with this SHA. Welcome aboard.

PR #1242, opened 2026-07-24
docs: add mcpindex screened badge

gautamgb:
Hey, Codebase Memory's registered MCP server came back clean on mcpindex's semantic screen (injection/manipulation check on the tool description, nothing deeper). Added the badge to your README in case it's useful. Advisory signal, not a security certification. Feel free to close if not a fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![mcpindex](https://mcpindex.ai/api/v1/badge/io-github-deusdata-codebase-memory-mcp)](https://mcpindex.ai/server/io-github-deusdata-codebase-memory-mcp)
+
 # codebase-memory-mcp
 
 [![GitHub Release](https://img.shields.io/github/v/release/DeusData/codebase-memory-mcp?style=flat&color=blue)](https://github.com/DeusData/codebase-memory-mcp/releases/latest)


DeusData:
Thanks for the contribution, and sorry for the slow first response.

Mechanically trivial — MERGEABLE, checks green, two added lines.

Being upfront so it does not sit in limbo: adding a third-party badge to the README is a decision about the projects public presentation rather than a code change, so it is the maintainers call rather than something triage should merge. I am queuing it as a decision item rather than a review item. That is not a no, just an honest account of what it is waiting on.

Thanks for thinking of the project.

DeusData:
Thank you for thinking of us, and for making it a two-line change rather than a redesign — but I'm going to decline this one.

The reason isn't the badge's placement (though above the `# codebase-memory-mcp` heading isn't where it would go). It's that this is a third-party service publishing a quality judgement about this project, on this project's own front page, where the meaning of the judgement is entirely outside our control. "Screened" means whatever mcpindex decides it means, and it can change — in either direction — without anyone here being consulted or even notified. The badges already in the README are all things we can verify ourselves: release version, license, build status, star count. That's a line I'd rather keep.

None of which is a comment on mcpindex, which may well be useful; it's about what a README asserts on our behalf. If they list us, that's genuinely appreciated and I'm glad the project is indexed there.

Closing with thanks — drive-by improvements from people who don't otherwise have a stake are how a lot of small things get fixed, and I'd rather say no clearly to this one than leave it sitting unanswered.

PR #1245, opened 2026-07-24
True incremental indexing: clean-rebuild parity, 6x post-edit indexing speedup, explicit freshness, and graph-derived search ranking

ahundt:
## 1. What does this PR do?

Hey, thanks again for the cool project! This PR continues my earlier PR #151. This Aug 2 addresses your requests: the `api-consolidation` branch is now merged with upstream main. Two true merge commits bring in every upstream commit through `f0c65fb2a997` (2026-07-29): `d0f00de6bd6d` combines upstream `560ad40d19b2` with the previously verified head `b42801f3f765` and resolves all 14 textual conflicts, and `41fd25cdd86b` integrates the `f0c65fb2a997` release-hardening wave. No history was rewritten; both parents are preserved in each merge. At that merge the branch was 0 commits behind upstream main, and every upstream commit through `f0c65fb2a997` remains contained here, so CI can run. A third merge, `a0f4380b6f18` (2026-08-01), integrates every upstream commit through `d6be58ef9d43` (2026-07-31). At the time of writing (2026-08-02 16:39 UTC, `d6be58ef9d43`) that is still the upstream main head, so the branch has caught main and is 0 commits behind. Every defect claim in section 4 was re-verified at that revision. Beyond the three merges, 37 follow-up commits fix defects found while validating the merged results (listed with their verification in each commit message). The merge also removed the 86 committed soak/memlab run-log files (1.3 MiB) reported separately as issue #1330; upstream has since removed them on main in `66326c2c`.

The core claim is unchanged: upstream main's post-edit indexing is labeled incremental but returns a graph that differs from an independent clean rebuild in 3 of 3 repetitions on the measured build, that is, the graph it serves after an edit is incorrect, and upstream's own source comment documents that `FILE_CHANGES_WITH` and `DATA_FLOWS` edges are not reproduced incrementally (`pipeline_incremental.c:351-360` at `d6be58ef9d43`). This PR makes incremental source-structure indexing exact, and in its default configuration indexes the same large-codebase edit 3.2× faster than the measured upstream build (569 vs 1,828 ms). After an edit the server names every deferred derived view (rank, routes, semantic edges, architecture), omits stale rank values where returning them would mislead, and with `incremental_derived_results_refresh=at_publish` produces a graph identical to the clean rebuild in 3 of 3 repetitions while still beating its own full rebuild (8,568 vs 9,768 ms).

The PR adds optional indexing of installed dependency sources with package identity, optional graph-derived search ranking, and a shorter default MCP tool list, and it extends the read-only Cypher subset, Git-snapshot change detection, and semantic analysis. One attribution correction to my previous description: the shared daemon is upstream's subsystem (`src/daemon/`, merged here on 2026-07-23); this PR retains it and contributes fixes and hardening on it, listed in section 2. Since findings like these are easy to misread from descriptions alone, I would appreciate review against the code, tests, and published run reports rather than summaries.

The full classic API, request schemas, and upstream 3D visualization are retained. Every major capability has a one-command switch and defaults can be trivially revised if preferred.

## 2. Capabilities and controls

Every setting is declared in one registry with its default, accepted values, and environment override: `CBM_CONFIG_REGISTRY` in `src/cli/cli.c:12997`. Two commands control everything:

```bash
codebase-memory-mcp config set <switch> <value>   # flip one switch, e.g. config set rank_enabled false
codebase-memory-mcp config preset apply <name>    # apply a validated multi-setting bundle in one atomic transaction
```

The six preset bundles pair a tool surface (`streamlined-` or `classic-`) with `automatic-dependency-source-indexing-disabled` or `-enabled`, plus the `rank-disabled` and `minimal-indexing` benchmark ablations (`codebase-memory-mcp config preset list` shows each name with its description; the apply transaction is `cbm_config_apply_preset()` in `cli.c:12925`). When an environment override pins one of a preset's keys, apply returns status 1 rather than claiming an ineffective change.

File names below are unique in the tree and live under `src/`; each switch is the exact argument list for `codebase-memory-mcp`, with the default in parentheses. Line numbers are at the branch head `7433fee6bed3`.

| Capability | What it does and why it is better | Key code | Switch (default) |
|---|---|---|---|
| Exact incremental indexing | After an edit, the non-stale incremental graph equals an independent clean rebuild instead of silently diverging; changed files are scoped by per-file Git status | `cbm_pipeline_run_incremental()`<br>`pipeline_incremental.c:2826` | `config set incremental_reindex always`<br>(default `always`; `full_rebuild` rebuilds atomically from scratch and `fast_mode_indexes_only` restricts incremental reindexing to fast-mode indexes) |
| Explicit post-edit freshness | The default policy quickly refreshes source structure and names each deferred derived view (rank, routes, semantic edges, architecture) as stale with an actionable refresh instruction; `at_publish` recomputes everything immediately and matches the clean rebuild exactly | `cbm_pipeline_run_incremental()`<br>`pipeline_incremental.c:2826` | `config set incremental_derived_results_refresh at_publish`<br>`config set rank_refresh at_publish`<br>(default: `defer_all_incremental_reindexes` for both) |
| Shared daemon fixes | The upstream shared daemon is retained with its session-managed lifecycle; this PR's fixes on it: honor the `auto_index` default at session start, suppress release notices for dev builds, initialize the active-image fingerprint before comparison, validate fingerprint-cache epochs and admit cross-volume caches, and initialize capped memory once in the daemon instead of in every CLI frontend | `main_run_daemon_ctl()`<br>`main.c:1728` | `daemon start` / `daemon stop` / `daemon status`<br>(default: session-managed, started automatically) |
| Class-to-method connection | Upstream links a class to its methods only when extraction supplies the parent; this PR adds the `MEMBER_OF` reverse edge and a cross-language normalization pass that repairs extraction-miss cases, so classes participate in traversals and inherit rank from their members instead of staying disconnected | `process_def()`<br>`pass_definitions.c:279`<br>repair: `pass_normalize.c` | always on |
| Graph-derived search ranking | Orders search results by PageRank/LinkRank computed over the code graph; in the measured search task the known correct answer, `zz_order_core`, moved from ninth to first | `cbm_pagerank_compute()`<br>`pagerank.c:407` | `config set rank_enabled false`<br>(default `true`) |
| Dependency source indexing | Indexes installed package source so search and tracing return external APIs with package identity; upstream resolves manifest imports but does not index installed source. Automatic indexing is off by default so the dependency cost stays out of the default first index; `index_dependencies` indexes named packages on demand, and when installed packages exceed `auto_dep_limit` (default 20, 0 = unlimited, max 10,000) the most-imported packages are selected, ranked by project import references | `cbm_dep_auto_index_effective()`<br>`depindex.c:984`<br>ranking: `depindex.c:583-616` | `config set auto_index_deps true`<br>(default `false`) |
| Streamlined tool list | New MCP sessions see the common tools plus `_hidden_tools`; calling `_hidden_tools` reveals the full classic tool list live, no restart (upstream advertises all tools and curates only statically via startup `--profile` flags) | `cbm_mcp_tools_list_range()`<br>`mcp.c:3429` | `config set tool_mode classic`<br>(default `streamlined`) |
| Response encoding | Default reply encoding is configurable between TOON (Token-Oriented Object Notation, a compact text format) and JSON, with a per-call `format` override and an actionable invalid-format error; the merged head keeps upstream's UTF-8 output guard so quoting or truncation cannot split a multibyte character | `cbm_mcp_response_format()`<br>`mcp.c:2685` | `config set default_response_format json`<br>(default `toon`) |
| Automatic first-use indexing | Indexes a project when it is first queried, so first-use calls do not fail on an absent index; while startup indexing is still publishing, tool calls receive structured retry metadata instead of an error or an indefinite wait | registry entry `auto_index`<br>in `CBM_CONFIG_REGISTRY` | `config set auto_index false`<br>(default `true`; upstream defaults to `false` and skips with a log hint) |
| Read-only Cypher subset | Fixes aggregate, post-`WITH`, and optional-match semantics; adds multi-key ordering, output caps, and schema-aware rejection of unsupported syntax; the tool description, parser, and executor share one definition, and the query schema is stable against incidental storage changes | `cbm_cypher_execute_impl()`<br>`cypher.c:9091` | always on |
| Semantic and similarity edges | Retains and extends upstream semantic analysis; each is now independently selectable instead of tied to index modes | `cbm_pipeline_pass_semantic_edges()`<br>`pass_semantic_edges.c:1783` | `config set semantic_edges_enabled false`<br>`config set similarity_enabled false`<br>(default `true` for both) |
| Git-history analysis | Extends upstream snapshot change detection into a shared module and scopes incremental reindexing by changed Git-status paths; the Git scan now runs concurrently with graph population | `cbm_git_snapshot_read()`<br>`git_snapshot.c:319` | `config set githistory_enabled false`<br>(default `true`) |
| HTTP route linking | Keeps upstream's Route-node bridging (`pass_route_nodes.c`, run in both the full and incremental pipelines) and adds a dedicated pass that links each HTTP call site to its handler, fixing missing or misclassified links | `cbm_pipeline_pass_httplinks()`<br>`pass_httplinks.c:1375` | `config set httplinks_enabled false`<br>(default `true`) |
| Startup fingerprint cost policy | The exact executable SHA-256 is reused while kernel-bound file identity and strong change metadata are unchanged, removing a full-image hash from every process start; any ambiguity falls back to an exact full hash, and installer/update candidates always use full exact hashes | registry entry `build_fingerprint_mode`<br>in `CBM_CONFIG_REGISTRY` | `config set build_fingerprint_mode always_rehash`<br>(default `cached_exact`) |

## 3. Measured results

The headline numbers come from a 39-cell benchmark experiment: 13 configurations, each run 3 times, paired and interleaved across configurations to reduce timing drift, on production builds in isolated Linux Docker containers (clang-18, `-O2`, 16 build jobs, 16 container CPUs, container `--init`, fresh case-owned repositories and caches, arm64). Every cell applies the same real C source edit to a large codebase, runs incremental indexing, and compares the resulting graph with a separate clean rebuild after the same edit; every configuration also passes 12 retrieval-evidence checks (4 post-edit retrieval probes across the 3 repetitions: the changed file, its source text, its graph node, and its architecture evidence). All 39 planned cells completed with zero missing, corrupt, duplicate, or unplanned cells. Docker Desktop exposed 7.75 GiB despite the requested 16 GiB limit; the largest observed candidate peak was about 1.6 GiB, so no run was memory-capacity limited.

Two scope statements, up front. First, the measured upstream build is `aedb979f5570` (upstream main as of 2026-06-28), pinned when the benchmark cohort was frozen; the upstream main now merged into this branch is `d6be58ef9d43` (2026-07-31). The graph-correctness defect reproduced on the measured comparator, and at current upstream main `d6be58ef9d43` upstream's source still documents non-reproduced incremental edges (`pipeline_incremental.c:351-360`, the file is unchanged since `f0c65fb2a997`) with no parity verification mechanism. Second, this matrix ran at PR revision `2c2de761a24e`, so its timing ratios apply to the named revisions, edit, and machine; a second full 39-cell matrix has since re-measured the merged branch head `7433fee6bed3` against an upstream build at `d6be58ef9d43` (below). What the merged head does run is the full test suite, including the incremental-parity tests, at the final verified head listed in section 7.

All timings are medians (p50) over the 3 repetitions. Repeated query p50 is the median latency of the repeated read-only graph queries the retrieval probes issue within each candidate's persistent MCP session, so it measures a warm service, not process startup. Response payload p50 is the exact default tool-response size (tokens are the recorded deterministic bytes/4 estimate). Peak RSS is the highest peak resident memory observed across the 3 repetitions. "Deferred views named" means the server explicitly names each optional derived view awaiting recomputation after the fast source-structure refresh, and the freshness-scoped comparison still checks every remaining canonical node, edge, property, and file hash against the independent rebuild. A clean rebuild of the same post-edit tree is the ground truth: a post-edit graph that differs from it is incorrect, and queries against it can return wrong answers.

| Configuration | Post-edit graph vs independent clean rebuild (3 repetitions) | Post-edit reindex p50 | Full index p50 | Repeated query p50 | Response payload p50 | Peak RSS |
|---|---|---:|---:|---:|---:|---:|
| Upstream main `aedb979f5570`, default | **Incorrect after the edit: differs from the ground-truth rebuild, 3 of 3** | 1,828 ms | 7.055 s | 6.2 ms | 604 B (~151 tokens) | not recorded |
| This PR, upstream-equivalent capabilities (rank, route linking, and dependency indexing off to match upstream's feature set) | Correct on every non-stale row, verified against the ground-truth rebuild, 3 of 3; deferred views named | 651 ms | 10.265 s | 19.5 ms | 429 B (~108 tokens) | 1,469 MiB |
| This PR, default | Correct on every non-stale row, verified against the ground-truth rebuild, 3 of 3; deferred views named | 569 ms | 10.029 s | 23.2 ms | 429 B (~108 tokens) | 1,434 MiB |
| This PR, automatic dependency indexing enabled | Correct on every non-stale row, verified against the ground-truth rebuild, 3 of 3; deferred views named | 774 ms | 36.148 s | 28.8 ms | 429 B (~108 tokens) | 1,614 MiB |
| This PR, derived results refreshed at publish | **Fully correct: identical to the ground-truth rebuild, 3 of 3** | 8,568 ms | 9.768 s | 22.7 ms | 429 B (~108 tokens) | 1,450 MiB |
| This PR, `minimal-indexing` preset | **Fully correct: identical to the ground-truth rebuild, 3 of 3** | **508 ms** | **5.136 s** | **16.7 ms** | 429 B (~108 tokens) | 1,459 MiB |

Upstream main's post-edit reindex is labeled incremental but returns an incorrect graph, one that differs from the ground-truth rebuild (speed bought with wrong answers), while this PR's default is both 3.2× faster on the edit (569 vs 1,828 ms) and exact on non-stale source structure. The minimal preset outputs a graph identical to the clean rebuild while indexing the edit 3.6× faster (508 vs 1,828 ms) and full-indexing 1.37× faster (5.136 vs 7.055 s) than the measured upstream build. Two metrics go the other way and I state them plainly: upstream's repeated read-only query latency is lower (6.2 vs 16.7 to 23.2 ms; this PR's replies carry first-response freshness and provenance context and are 29% smaller, 429 vs 604 bytes), and upstream's full index is faster than this PR's default configuration (7.055 vs 10.029 s) because the default enables optional capabilities upstream does not have: disabling semantic edges alone brings the full index to 6.659 s, and the rank, similarity, git-history, and route ablations measure 9.269 to 10.140 s. Upstream peak RSS was not recorded in this run, so no memory comparison against upstream is claimed.

Refresh-at-publish recomputes every global derived result (rank, semantic edges, architecture, routes) immediately after the edit; at 8,568 ms that is still faster than its own 9,768 ms full rebuild, which is the only other way to get them all fresh. It is opt-in; the default defers exactly that graph-wide work and names the deferred views until refreshed.

Automatic dependency indexing stays off by default so its cost stays out of the default first index: enabling it indexed this workload's six installed packages in about 27 s of dependency work at first index, which is work upstream cannot do, and after an edit, checking unchanged dependencies cost 13 ms.

Two earlier development checkpoints (`c9726c7fd9b7`, `0fcb7e109518`) ran in the same matrix and quantify what the post-merge work fixed: their native configurations auto-indexed dependencies (about 20.5 to 20.8 s of the 43.8 s full index) and carried a since-removed startup regression that pushed repeated queries to 187 to 206 ms; the current default full-indexes in 10.0 s with 23.2 ms repeated queries, and both checkpoints also failed the canonical-graph gate that the current head passes.

A second 39-cell matrix re-measured the merged result: same container design (13 configurations, 3 repetitions each, paired-interleaved, isolated Linux Docker containers, arm64, 16 container CPUs), run at branch head `7433fee6bed3` (2026-08-02) with the upstream comparator freshly built at `d6be58ef9d43`, the upstream main head this branch merges. All 39 planned cells completed with zero missing, corrupt, duplicate, or unplanned cells. The mutation differs from the first matrix (a new isolated C source file is added instead of editing an existing one), so timings are not comparable between the two matrices. Upstream's default again served an incorrect post-mutation graph in 3 of 3 repetitions: it differs from the ground-truth rebuild of the same tree, so queries after the edit can return wrong answers, and upstream ships no mechanism that detects this. Every configuration of this PR passed its graph gate.

| Configuration | Post-mutation graph vs independent clean rebuild (3 repetitions) | Post-mutation reindex p50 | Clean rebuild p50 | Repeated query p50 | Response payload p50 | Peak RSS |
|---|---|---:|---:|---:|---:|---:|
| Upstream main `d6be58ef9d43`, default | **Incorrect after the edit: differs from the ground-truth rebuild, 3 of 3** | 3,208 ms | 7.508 s | 17.3 ms | 306 B (~77 tokens) | 1,536 MB |
| This PR, upstream-equivalent capabilities | Correct on every non-stale row, verified against the ground-truth rebuild, 3 of 3; deferred views named | 473 ms | 8.025 s | 17.6 ms | 407 B (~102 tokens) | 1,532 MB |
| This PR, default | Correct on every non-stale row, verified against the ground-truth rebuild, 3 of 3; deferred views named | 523 ms | 8.674 s | 20.0 ms | 407 B (~102 tokens) | 1,566 MB |
| This PR, derived results refreshed at publish | **Fully correct: identical to the ground-truth rebuild, 3 of 3** | 7,805 ms | 8.437 s | 22.0 ms | 407 B (~102 tokens) | 1,522 MB |
| This PR, `minimal-indexing` preset | **Fully correct: identical to the ground-truth rebuild, 3 of 3** | **437 ms** | **4.141 s** | **16.6 ms** | 407 B (~102 tokens) | **1,515 MB** |

`minimal-indexing` is the measured Pareto frontier of this matrix: no passing, fully measured candidate in the same workload is at least as good on overall quality and every cost axis while strictly better on any. All speed comparisons below are how much faster this PR is than current main (`d6be58ef9d43`) in this matrix; a negative value means slower, and every negative is bought by a named capability gain. Post-mutation indexing: +578% (473 vs 3,208 ms) at upstream-equivalent capabilities and +513% (523 ms) at the default, roughly 6x either way, and correct where upstream's result is incorrect. Clean rebuild: -6% (8.025 vs 7.508 s) at upstream-equivalent capabilities (parity verification runs in the test and benchmark harness, so it adds nothing here; the difference is graph-building work this branch always does, including the `MEMBER_OF` class-connectivity repair and exact statistics publication, with no single component isolated by an ablation, and at this price the post-mutation graph is verified correct where upstream's is incorrect); -13% (8.674 s) at the default, which adds ranking, similarity, git history, and route linking with the measured benefits in the table below; +81% (4.141 s) with `minimal-indexing`, fully correct. Repeated query: -2% (17.6 vs 17.3 ms) at upstream-equivalent capabilities, closing the first matrix's 3 to 4x gap, -14% (20.0 ms) at the default, +4% (16.6 ms) with `minimal-indexing`, at effectively equal peak memory (1,532 vs 1,536 MB). In this workload upstream's default reply is smaller (306 vs 407 bytes); the PR's replies carry the first-response freshness and provenance context described above. Refresh-at-publish again produced the fully identical graph while beating its own clean rebuild (7,805 ms vs 8,437 ms).

The default's remaining costs are itemized capability prices, not overhead, and each is refundable with one switch. Overhead, cost that returns no value, is what this head (`7433fee6bed3`) removed: the first matrix's repeated-query deficit does not reproduce (17.6 ms with upstream-equivalent capabilities vs upstream's 17.3 ms), and the like-for-like clean rebuild is -6% (0.52 s, 8.025 vs 7.508 s) at equal peak memory. Parity verification runs in the test and benchmark harness, not in production indexing, so that 0.52 s is not a verification tax: it is graph-building work this branch always does, including the `MEMBER_OF` class-connectivity repair and exact statistics publication, with no single component isolated by an ablation. What the build delivers at that price is the guarantee upstream does not offer at any price: the post-mutation result is verified correct. Every cost above that baseline is priced below against the same build with exactly that capability toggled (all other defaults unchanged); the default configuration's own rebuild spread is [8.437, 9.599 s] over 3 repetitions, so prices under 0.6 s are smaller than that spread and should be read as approximate:

| Capability (default state) | Clean-rebuild price; repeated-query price | Measured benefit on the same build (section 3 quality experiment; matrix cells) | Refund switch |
|---|---|---|---|
| Semantic edges (on) | +3.25 s (8.674 vs 5.421 s); +0.8 ms | Every judged `SEMANTICALLY_RELATED` pair matches the clean rebuild, none missing, none spurious | `config set semantic_edges_enabled false` |
| Graph-derived ranking (on) | +0.56 s, below the run spread; +1.8 ms | The known correct answer returned 1st of 9 instead of 9th of 9; MRR 1.000 vs 0.111 | `config set rank_enabled false` |
| Similarity edges (on) | +0.34 s, below the run spread; +1.0 ms | `SIMILAR_TO` edges for near-duplicate discovery | `config set similarity_enabled false` |
| HTTP route linking (on) | +0.31 s, below the run spread; +1.0 ms | Each HTTP call site linked to its handler | `config set httplinks_enabled false` |
| Git-history analysis (on) | +0.18 s, below the run spread; +0.7 ms | `FILE_CHANGES_WITH` coupling edges and Git-scoped incremental change detection | `config set githistory_enabled false` |
| Dependency source indexing (off) | +23.25 s and +5.4 ms only when enabled | An installed dependency's API returned 1st with package identity, vs absent | off by default; `config set auto_index_deps true` |

Summed, the default's entire repeated-query premium over upstream at the merged head (`7433fee6bed3`) is 2.7 ms (20.0 vs 17.3 ms). The one axis where upstream is faster like-for-like is the clean rebuild, where this PR is -6%; what stands on the other side of that price is the post-mutation result: +578% (473 vs 3,208 ms) with a verified-correct graph, where upstream's faster-looking incremental serves an incorrect graph, one that differs from its own ground-truth rebuild in 3 of 3 repetitions, speed bought with wrong answers and no switch that refunds it.

Result-quality contribution of the two new optional capabilities, isolated on one PR build (`13a4679b27f6`) and test project by repeating the same index-then-query measurement 5 times per state, alternating the capability between enabled and disabled. MRR (Mean Reciprocal Rank) rewards placing the known correct answer near the top, Hit@1 and Hit@5 report whether it appears first or within the first five, nDCG@5 (normalized Discounted Cumulative Gain) rewards correct ordering within the first five, and task success means the task's named expected answer was returned:

| Capability | Enabled | Disabled (same build) | Cost of enabling |
|---|---|---|---|
| Graph-derived ranking. Task: rank the test project's structurally central order workflow (`zz_order_core`) ahead of eight lexical-only decoys | `zz_order_core` returned **1st of 9**; MRR, Hit@1, Hit@5, nDCG@5, task success all 1.000 | `zz_order_core` **9th of 9**; MRR 0.111, the other four metrics 0 | +0.2 ms median query, +12 ms test-project index |
| Dependency source indexing. Task: retrieve an installed dependency's imported API (`canonicalDependencyAPI`) with its dependency, package, and read-only provenance on the same result | `canonicalDependencyAPI` returned **1st**, with package identity | `canonicalDependencyAPI` **absent** | +0.6 ms query, +13 MiB RSS, test-project full index 81 vs 51 ms |

Semantic refresh quality was re-verified in the same matrix: with refresh at publish, every judged `SEMANTICALLY_RELATED` pair matches the clean rebuild, with none missing and none spurious.

## 4. Upstream main defects this PR corrects

Every claim in this table was re-verified by direct code reading at upstream main `d6be58ef9d43` (2026-07-31, the revision merged into this branch) unless a different revision is named; upstream file:line references are at that revision.

| Improvement in this PR | Key code | Issue at upstream main `d6be58ef9d43` | Consequence |
|---|---|---|---|
| The non-stale graph produced after an edit equals an independent clean rebuild, and the fully refreshed graph is identical (measured above); the PR's core correction | `cbm_pipeline_run_incremental()`<br>`pipeline_incremental.c` | Upstream's own comment documents that `FILE_CHANGES_WITH` and `DATA_FLOWS` edges are not reproduced incrementally (`pipeline_incremental.c:351-360`); no parity guarantee or verification mechanism exists; measured divergence 3 of 3 on `aedb979f5570` | **Any answer could be wrong after an edit** |
| A query detects that another process replaced the database file and reopens it, with no global lock (ThreadSanitizer clean) | `resolve_store()`<br>`mcp.c` | The cached fast path returns the held store with no file-identity check (`resolve_store_internal`, `mcp.c:2081-2083`); upstream's daemon single-ownership makes cross-process replacement rarer but the check is still absent | **Queries silently read a replaced or deleted database** |
| Cypher aggregates, post-`WITH` stages, optional matches, and multi-key `ORDER BY` compute correctly | `cbm_cypher_execute_impl()`<br>`cypher.c` | `parse_order_by_clause` parses exactly one ordering key, so secondary `ORDER BY` keys are silently dropped (`cypher.c:1699-1709`); the aggregate, post-`WITH`, and optional-match failures were verified on the measured comparator `aedb979f5570`; upstream has since fixed aggregate-column classification and two optional-match defects (PRs #1221, #1177, #1385, contained here through the `d6be58ef9d43` merge), while multi-key `ORDER BY` remains absent upstream | **Wrong values, no error** |
| A Cypher query considers every eligible row before output limits apply; `LIMIT` lowers but cannot bypass the configured row cap, and truncation is reported with the limiting condition instead of being silent | `scan_pattern_nodes()`<br>`cypher.c` | Unlabeled scans cap candidates at `max_rows` times a fixed growth factor before inline property filters and `WHERE` run (`cypher.c:3027-3052`, cap applied at `cypher.c:3037`), so matching rows past the prefix are discarded | **Results silently missing, whatever the query needed** |
| The first query against a project indexes it automatically; when indexing is blocked or still publishing, the reply names the required action or returns structured retry metadata | registry entry `auto_index`<br>in `CBM_CONFIG_REGISTRY` | Auto-index defaults to `false` and an unset value skips indexing with only a log hint (`mcp.c:11156-11169`) | **First use fails outright, recovery requires reading server logs** |
| Classes get a `MEMBER_OF` reverse edge and a cross-language repair pass for extraction-miss cases | `process_def()`<br>`pass_definitions.c:279`<br>repair: `pass_normalize.c` | Upstream mints `DEFINES_METHOD` only when extraction supplies the parent; there is no reverse edge and no repair pass (`pass_normalize.c` does not exist upstream), so extraction-miss classes stay disconnected | **Disconnected classes invisible to traversals and rank** |
| `index_status` reports whether the working tree is dirty and whether the index matches it | `add_git_context_json()`<br>`mcp.c` | `add_git_context_json` (`mcp.c:2242`) emits branch and commit identity but no working-tree-state field; upstream's watcher tracks a dirty signature internally but does not report it as provenance | **Change tracking can trust stale provenance** |
| Git-history coupling edges are selected best-first under a configurable budget, and a partial result is reported with exact written, eligible, and omitted counts; per-file temporal data grows exactly with observations | `cbm_pipeline_githistory_compute_with_limits()`<br>`pass_githistory.c:646` | The initial index caps `FILE_CHANGES_WITH` candidates at a hard-coded 8,192 (`MAX_COUPLINGS`, `pass_githistory.c:346`) collected in hash-table iteration order with no sort, no configuration, and no log (`pass_githistory.c:215-217, 331`), so on a history with more qualifying pairs the emitted coupling edges are an arbitrary subset rather than the strongest couplings; per-file temporal facts stop at the first 16,384 files encountered (`MAX_FILE_TEMPORAL`, `pass_githistory.c:406`) | **Coupling and hotspot answers silently arbitrary on large histories** |

Three defect claims from my previous description are retired because re-auditing upstream showed them already fixed there: the half-opened SQLite handle on failed open (closed upstream in `e6e9c58e`, before the measured comparator), the config-linking pass's large stack arrays (the pass was rewritten upstream and now holds only small static tables), and the dropped route-discovery pass (rebuilt upstream as `pass_route_nodes.c`; the merged branch keeps upstream's pass and adds its own, section 2). Shutdown lifecycle also converged: upstream's daemon now joins worker threads and plumbs cancellation, and the merged branch carries both sides' shutdown work; this PR's remaining distinct contribution there is the verification discipline (leak gate, cancellation and reaping tests) rather than a defect claim.

One further upstream hard-limit defect is fixed here but deliberately kept out of the table, because neither tree currently wires the affected scanner into a production pass: upstream's env-URL walker holds a fixed 512-byte-per-frame path stack and silently skips directories nested deeper than 256 levels (`pass_envscan.c:514, 532` at `d6be58ef9d43`). This branch's walker uses dynamically grown path storage (`walk_path.h`), logs unrepresentable bindings instead of dropping them, and pins both behaviors with the regression tests `envscan_walks_more_than_256_pending_directories` and `envscan_accepts_root_path_longer_than_512_bytes`.

Every remaining row is regression-tested. Defects found while merging upstream and validating the result were fixed before this update; each is documented in its commit message. Two behaviors are boundaries rather than defects: the Cypher subset does not claim full openCypher conformance, and optional derived results refresh on the documented deferred schedule. Production `src/` contains no `system()` calls, and the heap-leak gate reports 0 leaked bytes at the final verified head.

## 5. Compatibility, breaking changes, and migration

1. No classic API removal: classic request schemas, handlers, and the full tool list remain, and upstream's CLI access to every read-only MCP operation is retained; `config set tool_mode classic` restores the classic list on a running installation. Streamlined and classic surfaces answer from the same graph and Cypher contract.
2. New installations change only the defaults in the capability table; each row names the switch that restores upstream-like behavior.
3. The shared daemon is upstream's subsystem and keeps its upstream lifecycle: session-managed start, stop with the last session, `daemon stop` refusal while committed clients remain, and orderly SIGINT/SIGTERM shutdown (cancel and reap work, join background threads, close transports, release claims and leases, remove the endpoint). This PR's daemon changes are the fixes listed in section 2.
4. Upstream's packaging decisions are carried through the merge unchanged: one binary per platform (the Windows launcher stub upstream removed after antivirus false positives stays removed), and the binary carries no self-update capability or download URL; install scripts own updating.
5. Upgrades migrate installer-owned state, preserve user edits, and keep existing indexes unless `--reset-indexes` is passed; the system-wide upgrade verified at an earlier head preserved all 462 existing indexes across 17 detected clients. The installer recognizes configuration written by earlier development builds and replaces it canonically instead of rejecting it, keeps user-owned TOML child tables when replacing a managed parent table, and refuses to overwrite genuinely user-edited entries.
6. This local validation ran on macOS, with the benchmark matrix in Linux Docker containers. Windows-specific code paths are covered by automated tests, and because the branch contains every upstream commit through `d6be58ef9d43`, the repository's CI has run its full matrix on this pull request at `7433fee6bed3`: the test shards (Linux gcc, macOS 14 and 15-intel, Windows CLANG64), ThreadSanitizer on three platforms, lint, smoke on Linux, macOS, and Windows, CodeQL, and the security gates all pass; the DCO sign-off check is the only failing check.

## 6. How to run the tests

```bash
make -f Makefile.cbm build/c/codebase-memory-mcp   # production binary
make -f Makefile.cbm test          # full C suite under AddressSanitizer/UndefinedBehaviorSanitizer
make -f Makefile.cbm test-leak     # heap-leak check (Apple leaks on macOS, LeakSanitizer on Linux); report in build/c/leak-report.txt
make -f Makefile.cbm lint-ci       # the CI-pinned cppcheck 2.20 and LLVM 20 format gates
make -f Makefile.cbm test-analyze  # Clang static analyzer (requires clang)
uv run python benchmarks/run_experiments.py        # one-repetition benchmark smoke; --full reruns the 39-cell matrix in Docker
```

The full suite passed with zero failures at the final verified head (7,823 passed, 2 platform-specific skips), and the CI lint gates (`lint-ci`) exit 0 there. The suite includes generalized extraction and incremental tests across C, C++, Go, Python, JavaScript, TypeScript, Java, Kotlin, and Rust, mixed-language call resolution, the daemon lifecycle, bootstrap, and IPC suites, and the tests that arrived with the merged upstream commits. The benchmark runner records exact candidate commits, binary SHA-256 values, compiler and optimization flags, CPU and memory allocation, environment overrides, case cleanup, and invalid-run status in an immutable audit manifest.

## 7. Evidence and revisions

| Name | Revision | Role |
|---|---|---|
| Upstream main merged | `f0c65fb2a997` | Upstream main head at the 2026-07-29 merge; every upstream commit through this revision is contained in the branch |
| Upstream main merged, current | `d6be58ef9d43` | Upstream main head (2026-07-31), integrated by the full-sync merge below so the branch is again 0 commits behind; every section 4 defect claim re-verified at this revision; also the second benchmark matrix's freshly built upstream comparator |
| Upstream sync merge | `d0f00de6bd6d` | True two-parent merge of upstream `560ad40d19b2` and the previous verified head `b42801f3f765`; all 14 textual conflicts resolved, no history rewrite |
| Release-hardening merge | `41fd25cdd86b` | True two-parent merge integrating upstream `f0c65fb2a997` |
| Full-sync merge | `a0f4380b6f18` | True two-parent merge (2026-08-01) integrating every upstream commit through `d6be58ef9d43`; no history rewrite |
| Final verified HEAD | `24aa8a415f5c` | Full C suite, ThreadSanitizer suite, heap-leak gate, and CI lint gates ran at this commit |
| Branch head | `7433fee6bed3` | Current head (2026-08-02); the repository CI matrix passes here (section 5) and the second 39-cell matrix ran here |
| Benchmark production build, first matrix | `2c2de761a24e` | The first 39-cell matrix and the installed dogfood ran at this commit |
| Measured upstream comparator, first matrix | `aedb979f5570` | Upstream main 2026-06-28, pinned when the first benchmark cohort was frozen (section 3) |
| Development checkpoints | `c9726c7fd9b7`, `0fcb7e109518` | Regression boundaries measured in both matrices |
| Quality-experiment build | `13a4679b27f6` | Enabled/disabled ranking and dependency measurements |

All experiments ran on production builds. The first 39-cell run set's immutable audit manifest has SHA-256 `1becd57c00edbe965073f4d7c9a79cca8812947f207ed51851c856369aa2c002`, and the second run set's (branch head `7433fee6bed3`, 2026-08-02) has SHA-256 `1e8f0d96128bd352107a0fb1633bcb0bcbd696a5ea8cb8226569d62e580a65e0`; each experiment root retains the input specification, per-run manifests, container environment records, logs, build identities, and cleanup status, so the numbers are independently reproducible with the command above. The [generated reports are published in a gist](https://gist.github.com/ahundt/3723eced4ad025b2c796a0c8fb9cce53); its current tree contains the 2026-07-28 run set and the earlier 39-cell run `be2097bdbf14`, and the focused ranking, dependency, and semantic reports remain in the gist revision history.

Ranking follows [PageRank (Brin and Page, 1998)](https://doi.org/10.1016/S0169-7552%2898%2900110-X) and [LinkRank (Kim, Son, and Jeong, 2010)](https://doi.org/10.1103/PhysRevE.81.016103); [nDCG (Järvelin and Kekäläinen, 2002)](https://doi.org/10.1145/582415.582418) defines the ranking metric; [RepoGraph (ICLR 2025)](https://proceedings.iclr.cc/paper_files/paper/2025/hash/4a4a3c197deac042461c677219efd36c-Abstract-Conference.html) motivates repository graphs. The repository experiments, not the citations, are the evidence that this implementation works.

## 8. Checklist

- [x] Tests pass locally (`make -f Makefile.cbm test` at the final verified head: 7,823 passed, 2 platform-specific skips)
- [x] Lint passes with the CI-pinned toolchain (`make -f Makefile.cbm lint-ci` exit 0 at the final verified head: cppcheck 2.20, LLVM 20 clang-format, suppression-policy and source-safety checks; the previous body's 49 open findings and 20 formatting violations are cleared)


DeusData:
Thanks for this — there is clearly a lot of substantial work here, and I want to be straight with you about the review problem it creates rather than let it sit silently.

As it stands the PR spans 17 top-level directories: `src` (28 files), `internal` (24), `benchmarks` (17), `docs` (8), `scripts` (6), plus `cmd`, `graph-ui`, `pkg`, and single-file changes to `README.md`, `CONTRIBUTING.md`, `CLAUDE.md`, `Makefile.cbm`, `install.ps1`, `server.json`, `.mcp.json`, `.clangd` and `.gitignore`. The core diff alone touches `cypher.c`, `cli.c`, `depindex.c` and `lang_specs.c`.

The title promises incremental-indexing parity and a 3x edit-indexing win, and that is a change I am genuinely interested in. But it is currently bundled with a benchmark harness, documentation, packaging and tooling changes, and this project reviews atomically — one PR, one claim. A change of this size touching this many subsystems cannot be reviewed responsibly as a unit, and the incremental-indexing work deserves better than being approved or rejected on the strength of everything around it.

Could you split it? The natural seam looks like: (1) the incremental-indexing correctness and performance change in `src`/`internal`, (2) the benchmark harness under `benchmarks/`, and (3) the docs/packaging/tooling changes. Landing them in that order also means your performance claim can be verified against the harness rather than asserted alongside it.

Two mechanical notes: the branch is CONFLICTING against `main`, and no CI checks appear to have run at all, so nothing has been validated yet.

I realise splitting completed work is annoying, and I would rather ask than quietly let a 124k-line PR go stale. The incremental-indexing piece is the part I most want to see land.

ahundt:
Thanks for the positive feedback about incremental indexing, and I totally understand the review friction. A diff touching 17 directories looks heavy, but I want to explain why splitting this work breaks functional validation and why trying it live is the lowest-friction path forward.

First, I am in the middle of merging latest upstream main right now and hope to land that update later today. Because upstream main moves so fast, full merging and validation will likely trail slightly behind HEAD again. Merging non-trivial upstream changes and running full validation takes about two full days of compute and testing. That cycle includes the actual merging process and going through all the changes to repair previously repaired and new bugs, production builds, AddressSanitizer memory safety checks, and the 39-cell benchmark suite across 9 languages to ensure exact clean-rebuild graph parity. If this work is broken into three separate PRs, managing three parallel merge and 2-day validation loops against a fast-moving main becomes unmanageable. Landing the change cleanly may ultimately require a brief two-day window of minor merges upstream so a conflict-free, fully validated commit can land without main immediately jumping ahead.

Second, the benchmark harness under benchmarks/ and the engine logic cannot be separated. The runner is not extra bloat; it directly toggles the configuration switches and preset flags added in this PR to run ablations and verify that incremental indexing does not silently drop graph edges. Likewise, the incremental engine depends on underlying fixes to storage locking, class-method edge linking, and Cypher query bounds. Separating the harness or sub-fixes from the engine removes the exact mechanism that proves correctness. I did very deep manual and automated validation to guarantee that net bugs drop significantly once this lands.

Additionally, as noted in the original description, any default parameters, tool lists, or feature flags can be trivially tuned to match whatever preferences you want for upstream. The flag registry was built specifically so default behaviors are easy to adjust without touching core engine logic.

Because reading a 124k-line diff line by line is hard, the lowest-cost option on your end is simply to check out the branch and try running it. You can execute `make -f Makefile.cbm test` and `uv run python benchmarks/run_experiments.py` directly. In under an hour, you can verify the 3x indexing speedup, zero memory leaks under AddressSanitizer, and exact graph parity live on your machine.

Once my current merge and validation pass finishes landing, hopefully later today (though it might trail a bit again), please give the build and benchmark suite a quick run and try using it live for a bit. If you test it and still feel the scope cannot land as a single unit due to review rules, I understand completely, and I will simply continue maintaining these performance and quality standards, harness tests, and graph-parity guarantees on my existing fork. But please give it a quick test run before making a final call.

ahundt:
Update: it is taking me a bit longer than I hoped but I'm now doing benchmarks on the merged version. I expect to be done today. 

Also it might be worth having benchmarks of reasonable duration be part of CI, to be able to compare and prevent performance regressions on an ongoing basis. Though GitHub CI might have some variations in overhead so that might be noisier than desired.

ahundt:
I've updated the code and also the code predates the DCO requirements so I would appreciate if that could be waived. Please take a look quickly as it takes too many resources to re-merge and there are major perf and correctness improvements.

DeusData:
Reviewed as far as a PR this size can be reviewed, and I want to be straight with you about both halves: **the evidence quality here is well above what we normally see, and the PR cannot merge in this shape.** Those are separate facts and I do not want the second to erase the first.

**What is genuinely impressive.** A 39-cell benchmark matrix with an immutable audit manifest and a published SHA. **Honest counter-metrics** — you report that your own query latency is 3–4× worse than upstream's and that your full index is slower by default. And real parity testing: `test_graph_diff.h` plus ~47 parity references means clean-rebuild parity is **tested, not asserted**. Publishing numbers that make your own work look worse is the behaviour that makes a benchmark trustworthy, and it is rare.

**Why it cannot merge as-is.** 896 commits, 395 files, and roughly nine distinct features — incremental rework, freshness surfacing, PageRank ranking, dependency indexing, a streamlined tool surface, Cypher extensions, git snapshot/history, HTTP-call linking, and a containerised Python/uv/Docker benchmark framework — plus flipped defaults and four shipped agent skills. Against a main that is moving fast, this is in a permanent merge-chase by construction: the last CI run was on a SHA that is no longer the head, the benchmark ran on a third SHA, and your "final verified head" is a fourth.

Three things need saying plainly:

- **CodeQL found a HIGH-severity issue in code this PR adds**: `cpp/toctou-race-condition` at `src/cli/config_yaml_edit.c:519` — a check-then-use race on the config path in the new config writer.
- **Nine CI failures are genuine**, in this PR's own subsystems and consistent across platforms — `test_subprocess.c:756 ASSERT(result.tree_quiesced)` on all three TSan legs and both Linux shards, plus `test_httplink.c`, `test_pipeline.c` and `test_watcher.c` on Windows. Incidentally, that quiesce wait looks like the transient-window shape our determinism doctrine forbids.
- **DCO is red** somewhere in those 896 commits.

**And one number needs a correction.** The headline 3.2× is measured against upstream `aedb979` from 28 June — **before** main's kernel perf work landed (−61% indexing) and before the daemon search gains. You disclose this, which I appreciate, but it means the figure is not valid against current main and should not be quoted as such.

**The direction problem is the real blocker, though.** This PR decides several one-way doors unilaterally: a 73-key config registry with presets and persistent config writing, a new always-on `MEMBER_OF` edge type, a new `_hidden_tools` MCP tool with streamlined-by-default tool surface, and PageRank ranking with 13 edge-weight knobs. Most consequentially, **it collides head-on with #879 and #880**, which are already with the maintainer as "do we want an importance/ranking concept in the graph at all?" Your PageRank design is a heavier competing answer to that same question. Merging any of the three pre-empts the others, so the decision has to come first.

**What I would like to salvage, as atomic reproduce-first PRs.** These are worth having regardless of how the direction call goes:

1. **The parity oracle** (`test_graph_diff.h` and its harness). Main's own comment documents that `FILE_CHANGES_WITH` and `DATA_FLOWS` are not incrementally reproduced — the gap you target is real, and the oracle is valuable as a test asset even if the incremental rework itself is declined.
2. **Multi-key `ORDER BY`.** I verified this on main: `parse_order_by_clause` (`cypher.c:1699`) parses exactly one order expression with no loop for comma-separated keys. `ORDER BY a, b` is not supported. Small, self-contained, clearly a bug.
3. **The `resolve_store` file-identity check** — detecting a replaced or deleted DB under a held handle. Small and testable.
4. **The daemon micro-fixes** — fingerprint-cache epoch validation, capped-memory init once in the daemon, `auto_index` honoured at session start.
5. **The `index_status` dirty-working-tree provenance field** — a small findability and trust win.

Each of those would land quickly on its own. Sent as one PR, they cannot land at all.

I would rather have five merged PRs from you than one that stays open indefinitely — and if you would prefer, we can distill any of them with `Co-Authored-By` credit rather than you carrying the rebases. Your call.

ahundt:
Thanks for the careful review. My preference is still for this PR to be merged, and the current head addresses the concrete blockers you listed. If distillation is the only option, I request and would appreciate Co-Authored-By credit.

**Status: the blockers from your review are closed, at one SHA.** CI, the benchmark, and the head now run on the same commit, which also ends the merge-chase you called out: the `api-consolidation` branch has caught main, with every upstream commit merged and 0 commits behind (at the time of writing, 2026-08-02 16:39 UTC, branch `7433fee6bed3`, main `d6be58ef9d43`).

- All nine test failures are root-cause fixed and green across the full matrix. The TSan quiesce read unrelated /proc churn as owned-group state, fixed in `df2f9406`, which also removes the transient-window shape; the Windows legs were caused by UTF-8, CRLF, and long-path defects.
- The config writer validates through the opened descriptor (O_NOFOLLOW, fstat, flock; no pathname check/use pair) and the CodeQL and security/codeql-gate checks both pass at head.
- Only DCO is red: those commits predate the requirement, more recent ones carry sign-offs.

**On the numbers: 3.2x was correct against the main of its day, and the gap has now improved to 6x against the main of today!** A fresh 39-cell matrix at this branch head measures against a build of current main directly (at the time of writing, 2026-08-02, `7433fee6bed3` vs `d6be58ef9d43`). Each figure below is how much faster this PR is than that build of main, so a negative number means this PR is slower. You are right the 3.2x should not be quoted against current main: it was measured against the main of 2026-06-28 (`aedb979f5570`), and the updated description replaces it. The two matrices apply different mutations, so each stands against the main of its own day rather than against each other.

- Post-mutation indexing (the vast majority of work during ongoing use): +578% (473 vs 3,208 ms, roughly 6x). This PR's post-edit graph is also correct where main's is incorrect: main's graph differs from a clean rebuild of the same tree in 3 of 3 repetitions, so queries after an edit can return wrong answers, and main ships no mechanism that detects the divergence.
- Repeated query: -2% (17.6 vs 17.3 ms; equal within noise bounds). The 3 to 4x query gap you flagged does not reproduce at this head; it was overhead and it is gone.
- Clean rebuild: -6% (8.025 vs 7.508 s). That 0.52 s difference is graph-building work this branch always does, including class-connectivity repair (the `MEMBER_OF` reverse edges and the cross-language pass that repairs extraction misses) and exact project-statistics publication; no ablation isolates a single component of the 0.52 s.
- Peak RSS: parity (1,532 vs 1,536 MB).

**None of the remaining deltas is overhead.** They are the itemized costs of capabilities and of fixing upstream bugs correctly, each named with what it buys: the semantic-edges pass (`SEMANTICALLY_RELATED`) costs +3.25 s rebuild and buys judged semantic pairs that match a clean rebuild exactly; PageRank ranking costs +0.56 s and buys the correct answer 1st instead of 9th (MRR 1.000 vs 0.111); similarity (`SIMILAR_TO`), HTTP route linking, and git-history coupling (`FILE_CHANGES_WITH`) each cost under 0.35 s, smaller than the run-to-run spread; dependency source indexing costs +23 s only when explicitly enabled and buys installed-API retrieval that has no upstream price at all. The updated description carries this as a table, so your preferred defaults are a registry tweak, not a design change.

**On #879/#880, the direction question: this PR already contains #879's capability in stronger form.** #879 proposes weighted degree as a persisted per-node importance score. This branch already computes, persists, indexes, and serves weighted degree, and two graph-derived scores besides. [`node_degree`](https://github.com/ahundt/codebase-memory-mcp/blob/7433fee6bed3da5b5118983c3ee41e6c3f6a9b69/src/store/store.c#L604-L615) carries `weighted_in` and `weighted_out` per node; [`pagerank`](https://github.com/ahundt/codebase-memory-mcp/blob/7433fee6bed3da5b5118983c3ee41e6c3f6a9b69/src/store/store.c#L592-L597) carries the PageRank score per node and [`linkrank`](https://github.com/ahundt/codebase-memory-mcp/blob/7433fee6bed3da5b5118983c3ee41e6c3f6a9b69/src/store/store.c#L598-L603) the LinkRank score per edge.

All three tables are written in one publication by [`cbm_pagerank_compute()`](https://github.com/ahundt/codebase-memory-mcp/blob/7433fee6bed3da5b5118983c3ee41e6c3f6a9b69/src/pagerank/pagerank.c#L407), indexed for ranked reads (`idx_pagerank_rank ON pagerank(project, rank DESC)`), and consumed by the search path, which prefers the precomputed `node_degree` join over an O(edges) subquery. #880's `repo_map` can read `pagerank` or `node_degree` today rather than waiting for a new table.

**Which score to rank by is the question your review actually poses, and it has now been measured.** #879 has some strong evidence: its real-corpus top-40 audit caught the test-penalty under-firing, 33 of 40 top-ranked symbols were test scaffolding and 0 after the fix. That audit validates scaffolding suppression, a guard against one failure mode; it does not measure whether #879's score surfaces the right answer, and #879's own text defers exactly that question as "a measured decision" on PageRank. This PR contains the deferred measurement, currently the only retrieval-quality measurement published for either ranking design: with ranking on, the known correct answer (`zz_order_core`, a structurally central workflow ranked against eight lexical-only decoys) returns 1st of 9 instead of 9th of 9, MRR 1.000 vs 0.111, on the same build (`13a4679b27f6`) with only that capability toggled. PageRank's price at this head is +0.56 s at clean rebuild, +1.8 ms per query, 0.70 s at initial index, and 35 ms on the post-edit path where rank refresh defers by default; `config set rank_enabled false` is the zero-cost option.

**Tuning the constants cannot close that gap, because the formula's inputs are lexical.** #879's writer multiplies degree by name-shape constants, declared at [`pass_importance.c` lines 69 to 72](https://github.com/petercoxphoto/codebase-memory-mcp/blob/cce0f680dc18040ba7c419ae0ab4c40cd267cb2d/src/pipeline/pass_importance.c#L69-L72) and applied at [lines 191 to 201](https://github.com/petercoxphoto/codebase-memory-mcp/blob/cce0f680dc18040ba7c419ae0ab4c40cd267cb2d/src/pipeline/pass_importance.c#L191-L201): leading underscore x0.1, and snake_case or camelCase with length >= 8 gets x10. Those two multipliers assume one naming convention. Across the 150+ languages this server indexes (158 vendored tree-sitter grammars), whole families never match: Lisp-family kebab-case and Go's short exported names never receive the x10 factor, and leading-underscore conventions invert meaning between languages. The failure is systematic and needs no unlucky input: across languages the formula will fail in general, collapsing to raw degree with arbitrary penalties exactly where the graph is most multilingual. The top-40 audit fixed a real defect, and it was scoped to one corpus, so it does not show how those two constants behave on a codebase with different naming conventions. PageRank reads graph structure only, which is why it transfers across languages.

Since you raised the direction question, my answer is that #879's writer should carry the same retrieval-quality measurement before it becomes the ranking anyone depends on. If either PR's approach measures better than what is here, I would readily adopt that approach, and I hold this PR to the same standard. #879's writer is also not required for #880: `repo_map` reading this branch's `pagerank` table gets a stronger score than the weighted-degree one it was written against, and I would support it landing that way. This PR and #880 do not otherwise conflict. Thanks again for the time this review took, and for naming the direction question directly, which is what made it concrete enough to answer with a measurement.

**Executive summary.** What I am asking for, and what I would suggest on the ranking question:

1. My ask is that you merge this PR at head `7433fee6bed3`, which now contains every commit of current main (`d6be58ef9d43`) and passes every technical check, with post-edit indexing 6x faster than main and a post-edit graph verified correct where main's is incorrect. The one thing I cannot resolve on my end is DCO, so a waiver for the commits that predate the requirement would unblock it.
2. If distillation turns out to be the only option, please have the work ported with Co-Authored-By credit.
3. On ranking, the direction question you raised with #879 and #880: this branch already persists weighted degree in `node_degree`, PageRank in `pagerank`, and LinkRank in `linkrank`, and serves them from the search path, so #879's capability is available here in stronger form, and I would want #879's name-shape formula to meet the "measured decision" bar #879 set for itself before that formula becomes the ranking anyone depends on. If either PR's approach measures better than what is here, I would readily adopt that approach. #880's `repo_map` looks ready to land against `pagerank` or `node_degree` whenever you want it.

DeusData:
Thank you for the substantial follow-up and for working through the earlier blockers. We have recorded the updated benchmark, scope changes, and current DCO state.

This still needs a fresh end-to-end correctness and security review of the current diff before we can make any merge commitment. Please keep contributor-authored commits DCO-compliant. If maintainers ultimately distill part of the work into a separate implementation, we will preserve clear contributor credit through appropriate co-authoring or explicit attribution. The review queue remains full, but this is queued for that fresh review.

ahundt:
Ok thanks! 

Just a brief hand typed constructive quick update. 

Performance has continued improving on this branch, I believe cold start is now ~10x faster than upstream main at the time of writing, and similarly the performance and correctness gaps have continued widening further in favor of this branch according to my metrics, so I'd really appreciate if it can be merged soon!

I've noticed a few performance improvements from upstream! At the same time, it takes a lot of work and resources to keep this branch updated because there are frequent upstream regressions. For that reason, I suggest merging this whole rather than piece-wise as the critical regression fixes this branch contains like preventing silent algorithm failures and difficult allocation and threading bugs that are present in upstream main (at the latest shared base with upstream main at the time of writing) would likely be lost and risk continued incorrect results.

I can also share skills that could help keep the perf up and address the runtime allocation and threading bugs on am ongoing basis if you'd like, they could go in a local .skills folder in this repo, or be shared in a gist, or in one of my other projects, however you'd prefer, but that can easily wait until a future PR too. 

I've also been able to do empirical testing on this branch vs the other one you were considering that handles graph info, and the data indicates that the ranking results are much better across multiple languages and large repos on this branch, and could continue getting better with tuning in addition to better overall perf on this branch. Hopefully I can post the data this week!

Best wishes and thanks again for putting some of your time into considering such a complex branch!

DeusData:
Andrew — thank you. Genuinely. There is a lot in here that we want, and I want to lead with that
rather than bury it under the process answer.

Specifically, these stood out as unusually good work:

- **You publish numbers that make your own work look worse.** Reporting that your query latency and
  clean-rebuild are slower than ours, and itemising each cost against the capability that buys it, is
  the behaviour that makes a benchmark trustworthy. That is rare and we noticed it.
- **The parity harness is real.** `tests/test_graph_diff.h` compares graph facts by stable keys rather
  than transient row IDs, and it is wired into `test_incremental.c`, `test_pipeline.c` and
  `test_store_nodes.c` with concrete cases like `incremental_fast_new_folder_exact_delta_parity`.
  Clean-rebuild parity is *tested* here, not asserted.
- **You fixed four languages' cross-file inheritance.** Your rework of base-class extraction in
  `internal/cbm/extract_defs.c` promotes `es_inherits_crossfile_{python,typescript,php,kotlin}` from
  known-RED reproductions to green. Our own test file invites exactly that
  ("*extraction bug may have been fixed — promote to GREEN*"). That is graph-quality-first work and we
  want it.
- You cleaned the local `notes/` files off the branch tip without being asked, and every commit now
  carries a DCO sign-off — the `dco` check is green.

So this is not a rejection of the work. It is a rejection of the *shape* of the delivery.

## The honest problem

As it stands the PR is **+180,230 / −32,226 across 391 files and 966 commits**. We cannot review that
responsibly as one unit, and merging it would mean approving nine or ten unrelated changes on the
strength of the one or two we actually managed to verify. Your incremental-indexing work deserves
better than that, and so do the users who would inherit whatever we failed to catch.

Our [`CONTRIBUTING.md`](https://github.com/DeusData/codebase-memory-mcp/blob/main/CONTRIBUTING.md) is explicit about this, and it is what we hold every
contributor to:

> **One issue per PR.** Each PR must address exactly one bug, one feature, or one refactor. Do not
> bundle multiple fixes or feature additions into a single PR. Kitchen-sink PRs will be closed with a
> request to split.
>
> **Keep PRs small.** A good PR is under 500 lines. If your change is larger, split it into reviewable
> increments that each stand on their own.

The same document lists what needs a design discussion *before* implementation — and this PR touches
every single line of that list:

> - **API surface changes** — adding, removing, renaming, or changing defaults of MCP tools
> - **New pipeline passes or indexing algorithms**
> - **Build system / Makefile changes** — beyond trivial fixes
> - **Project configuration** — CLAUDE.md, skill files, .mcp.json, CI workflows
> - **New dependencies** — vendored or otherwise

We are a large, widely-installed project. Those rules are not bureaucracy — they are the only reason
we can move as fast as we do without shipping regressions to a very large installed base.

## Why this has sat in review for so long

I want to answer this directly, because you have asked more than once for this to be merged sooner and
you deserve a straight answer rather than silence.

**The delay is not disinterest, and it is not a queue problem. It is the shape of the PR itself.**

A 391-file, 180,000-line change spanning nine or ten subsystems needs a single reviewer to hold all of
it in their head at once, verify each claim independently, and then accept joint responsibility for
every part of it in one irreversible decision. There is no safe way to do that in small increments of
time, so it needs a large contiguous block that keeps not arriving — and each time it *nearly* arrives,
the branch has moved and the verification has to start over. That is the loop we have been stuck in
since July.

Submitted as small PRs, most of this work would be close to a no-brainer. A 250-line test-only harness
gets reviewed and merged the same day. A self-contained extraction fix with a test that goes from red
to green needs one reading, not five. The bounded-limit fixes ("retain past N entries") are individually
obvious and each would take minutes.

**The thing you want — your work merged — is on the other side of splitting it up.** Not because we are
withholding it, but because small PRs are the only form in which this volume of work can actually move
through review at the speed you are asking for. You would very likely have had a large fraction of this
merged already.

## Why splitting also protects the work itself

This is not just policy. Your own branch demonstrates the problem:

- On **2026-08-10** (`8fa5df2b`) this PR was **31/31 green**.
- On **2026-08-13** (`4ac8c4a5`) it is **5 failing / 26 passing**.

The only delta is one upstream sync merge plus a handful of commits. Three days and one sync turned a
fully verified branch red. The current failures are:

- `FAIL 9b-9: update accepted retired --standard flag or returned no exact error` — on **all three**
  platforms. This is a semantic merge artifact: your branch kept a strict smoke assertion that main
  made permissive in #1538. A textual merge succeeded while the behaviour diverged.
- `pipeline rc=124 ... exceeded 900s wall clock (killed as hung)` on Windows CLANG64. Main's Aug-10
  baseline has every Windows shard green, so this needs attribution — and it is in the subsystem this
  PR reworks most heavily.
- One merge conflict remains, in `src/cypher/cypher.c`, where your rewrite meets a smaller fix from
  another contributor (#1378, `expand_from_bound_terminal` OPTIONAL fallback).

A 300-line PR does not have this failure mode. It rebases in minutes, and it stays green. Every day
this branch waits for review capacity, it decays — that is a bad deal for you, and it is the single
strongest argument for splitting.

## What we will not merge in any form

To save you effort, these are out of scope for this repository regardless of how they are packaged:

1. **`CONTRIBUTING.md`, and our other governance files.** This one I want to be direct about, because
   it matters more than the others: a contributor does not set contribution policy for this repository.
   Rewriting our contribution guide as part of your own feature PR is not something we will accept, and
   it is the change in here that gave us the most pause. If you think a rule is wrong or a documented
   command is inaccurate, open an issue and make the case — we do change our rules, and several of them
   exist because a contributor argued for them. But that has to be a separate, explicit conversation,
   not a diff inside a 391-file PR.
2. **The benchmark harness** (`benchmarks/`, ~36,600 added lines — about 20% of the whole PR) and its
   generated reports. We appreciate that it produced your evidence, and the evidence was useful to us,
   but we are not adopting a Python/uv/Docker benchmark framework as a maintained part of a lean C
   project. Please keep it in your own repository or a gist and link to it — that is where it does its
   job best.
3. **Documentation rewrites**, including `README.md`, `docs/*`, and `CLAUDE.md`.
4. **Comment, help-text, error-string and installer-description rewording** that is a matter of style
   rather than a fix. We will take a description change that corrects something *wrong*; we will not
   take a large-scale re-voicing of existing text.
5. **`.mcp.json`, `.clangd`, and the four `assets/skills/*/SKILL.md` files.** These are project
   configuration and need a design discussion first. On the skills specifically: we already ship skills
   of those exact names generated from `src/cli/cli.c`, and nothing on your branch appears to reference
   `assets/skills` — so if we have misread that, tell us, but as far as we can see they are unreferenced
   duplicates.

## What we would like you to do

**1. Squash your history.** 966 commits would grow this repository's commit count by **43%** (2,248 →
3,214). Please squash each split-out PR down to a small number of meaningful commits, each with a DCO
sign-off. Note this is about *your* branch history — it does not conflict with our no-squash-merge
policy, which only governs how we merge on our side. Squashing will also drop the 1,004 lines of local
`notes/` files that still sit in the branch's history even though you removed them from the tip.

**2. Split by layer, one concern per PR.** You know your commits better than we do, but the natural
seams we can see are below. **Ten PRs is fine. Thirty is fine. Fifty is fine** — we would genuinely
rather review fifty small PRs than one large one, and small ones get merged much faster.

On the 500-line guideline: we will be pragmatic. **Up to roughly 1,500 lines is acceptable** when the
change genuinely does not decompose further — a fix plus the connected test suite it repairs is the
usual legitimate case. What we ask in return is that the PR description states *explicitly* why it
cannot live under the 500-line rule, and what the single concern holding it together is. An honest
paragraph explaining the coupling is enough; an unexplained large PR is not.

Please start with these two, in this order. Both are small, neither depends on anything contentious,
and we expect both to land quickly:

| Order | PR | Scope | Rough size |
|---|---|---|---|
| 1 | **Pin the `mcp-publisher` download** | `.github/workflows/release.yml` — version + SHA-256 pin, verify before extract | ~10 lines |
| 2 | **Socket `CLOEXEC` hardening** | `src/ui/httpd.c` — `SOCK_CLOEXEC` + `ensure_socket_close_on_exec` | ~30 lines |
| 3 | **Clean-rebuild parity harness** | `tests/test_graph_diff.h` + its parity test cases, *tests only* | ~250 lines + tests |
| 4 | **Cross-file INHERITS extraction fix** | `internal/cbm/extract_defs.c` + `tests/test_extraction_inheritance.c`, promoting the four `_red` cases | ~1.2k lines |

On PRs 1 and 2 specifically: **thank you, these are real security fixes and we want them now.** Our
`release.yml` currently pipes an unpinned `curl … /releases/latest/download/… | tar xz` straight into
the job that publishes to the MCP Registry; your version pins the release, verifies a SHA-256, and only
then extracts. That closes a genuine supply-chain hole on our side, and it is a ten-line PR that we can
merge essentially on sight. The `httpd.c` change likewise stops the listening socket leaking across
`exec` into spawned children. Please send those two first — they are pure wins with no design question
attached, and they will give us both something merged while the larger pieces are still being split.

One small cleanup to fold in wherever it fits: `scripts/security-allowlist.txt` on your branch adds
`src/git/git_command.c:cbm_popen`, but that file uses `cbm_git_run_argv` and contains no `cbm_popen`.
Allowlisting a call that no longer exists weakens the gate's meaning over time, so please drop entries
that no longer correspond to real call sites. Relatedly, the comment in `src/watcher/watcher.c` stating
that "`src/git/git_command.c` still builds quoted shell command strings" no longer matches that file.

PR 1 is the most valuable thing you can give us first: it lets *our* CI measure clean-rebuild parity
directly. If our incremental path really does diverge, our own suite goes red and proves it — which
settles the central question of this PR on our infrastructure rather than on a benchmark either of us
has to take on trust. That is the fastest possible route to getting your core claim accepted.

Then, each as its own PR:

- Incremental indexing correctness (`pipeline_incremental.c`) — *after* PR 1 is in, so the parity tests
  demonstrate the improvement
- Cypher fixes — ideally split further: multi-key `ORDER BY`, the scan/cap behaviour, and aggregates as
  separate PRs
- Git snapshot / `pass_githistory` coupling budget
- `MEMBER_OF` class connectivity + `pass_normalize.c`
- HTTP route linking (`pass_httplinks.c`)
- Store changes
- Daemon / foundation / subprocess fixes
- Smoke layer (`scripts/smoke-*.sh`) — on its own
- Soak layer (`scripts/soak-test.sh`, `memlab.sh`) — on its own
- Test infra / parallel runners — on its own
- The individual bounded-limit fixes (the "retain past N entries" family) — these are small and
  self-contained and should each be quick merges

**3. Open an issue first for the feature-level items.** PageRank/LinkRank ranking, dependency-source
indexing, the streamlined tool surface, the config-registry preset system, and every default flip are
direction questions, not review questions. Per `CONTRIBUTING.md` they need agreement before
implementation. Please open one issue each describing the capability and the proposed default, and we
will give you a straight yes or no on the concept before you spend more effort. Some of these we may
well want — but the decision has to come first, and changing a default for a large installed base is a
one-way door we have to open deliberately.

**4. Reference a tracking issue** from each PR, as `CONTRIBUTING.md` asks.

## What we can and cannot promise

I would rather set expectations honestly now than have you feel misled later:

- **We cannot guarantee that everything will be merged, even after splitting.** Each PR gets a real
  review on its own merits, and some will come back with questions, change requests, or a "no". That is
  the point of splitting — each piece finally gets judged on what it actually does, instead of being
  carried or sunk by everything around it. What we *can* promise is that each one gets a genuine,
  reasoned answer rather than silence.
- **Feature-level items may be declined on direction alone**, independently of code quality. If we
  decide we do not want dependency-source indexing as a concept, that is not a judgement on your
  implementation. This is why we are asking for issues first on those — so you get the answer before
  you invest the effort, not after.

## On your benchmarks

Please put the relevant benchmark summary **in the description of the PR it belongs to**. That is
genuinely useful and we want it — a performance claim attached to the specific change it describes is
exactly right. Link out to your gist for the full run sets and manifests.

What we will then do is **run our own benchmarks to validate**, using our harness and our machines, or
come back to you with questions where our numbers disagree with yours. That is not distrust; it is that
cross-machine numbers are not comparable, and we record our baselines with host, arch, binary SHA and
dataset for exactly this reason. Where our measurements confirm yours, we will say so in the PR.

## Status of this PR

We will leave #1245 open for now so nothing is lost and the discussion stays in one place, though we
may move it to **draft** to reflect reality — it is not a merge candidate in its current form and we do
not want to leave a misleading signal in the review queue.

To be unambiguous: **this PR, as it stands, will not be merged.** The smaller PRs will be, once each is
verified and discussed on its own terms. When the split is done and the individual pieces have landed
or been answered, we will close this one out with a summary pointing at everything that came from it.

## On credit

You asked about attribution, and that is a fair ask given the volume of work. Our commitment: anything
of yours that lands, in any form — merged directly, or re-implemented by us — carries your authorship
or an explicit `Co-Authored-By: Andrew Hundt <ATHundt@gmail.com>`, and we will say plainly in the
commit and release notes where the work came from. That holds whether or not the split goes smoothly.

We would rather merge your work than reimplement it. Split it up and we will review the pieces
promptly and individually.

## Last thing — and the good news 🙂

I know this reads like a long list of asks, so I want to end on what it actually amounts to: **the hard
part is already behind you.**

The engineering is done. You have written the code, you have tested it, you have measured it, and in
several places you have fixed things we knew were broken and had not got to yet. None of that has to be
redone. What is left is not really engineering at all — it is packaging: some `git` surgery to slice the
branch along its natural seams, a squash, and a short honest description on each piece. That is an
evening or two of unglamorous work, not another two-day validation marathon. And unlike the current
loop, every hour you put into it converts directly into merged commits instead of evaporating on the
next upstream sync.

Your ambition here is genuinely appreciated — very few people would take on a rework at this scale, and
fewer still would document it as carefully or report their own numbers as honestly as you have. That is
exactly the kind of contributor we want around, which is why we would rather spend this effort getting
your work in than politely letting it stall.

So: send us those first two small PRs whenever you like, and let's get some of this landed. We are
looking forward to it, and thank you again — really — for the care you have put into this project. 🙂

DeusData:
Quick follow-up, and a concrete one 🙂

We did not want to sit on the `release.yml` finding while the split is being worked out, because it affects our publishing path today. So we have extracted it as **#1592**, with you credited as co-author:

```
Co-Authored-By: Andrew Hundt <ATHundt@gmail.com>
```

You were right that this mattered. `publish-mcp-registry` was pulling `mcp-publisher` from the `latest` release and piping `curl` straight into `tar`, inside the one job that holds our MCP Registry publish credential — so whatever upstream happened to publish at that moment executed before anything could verify it. Good catch.

We made one change to your version. Yours pinned the `linux/amd64` asset hash directly, which verifies correctly but hard-codes the platform, since it replaces the `uname`-based selection with a fixed filename. Instead we pinned the SHA-256 of the release's **own checksums file** and verify whichever asset the runner selects against it. Upstream ships that file for the whole release, so one pinned value now covers linux, darwin and windows on both amd64 and arm64, the `uname` selection keeps working, and bumping the version is still a one-line change.

We also verified it end to end rather than trusting the checksum: your value matches upstream's `registry_1.8.1_checksums.txt` exactly, the asset extracts to a valid statically linked ELF, and appending a single byte to the download makes the step **fail** — so the gate actually binds instead of passing vacuously.

**This is exactly what the split looks like in practice.** One finding, one small PR, verified on its own merits, merged on its own timeline, credit attached — start to finish in well under an hour. Multiply that by the rest of your branch and you can see why we are asking for it: the same work, landing steadily, instead of waiting on a single review that never gets large enough to happen.

The `httpd.c` `SOCK_CLOEXEC` fix is the other one we would take immediately — that one we would rather have from you directly, so please send it when you get a chance. Thanks again 🙂

Issue #1209, opened 2026-07-22
[v0.9.0][Windows] search_code content pipeline transcodes UTF-8 through the ANSI codepage — CJK pattern search always returns 0 matches, raw_matches show mojibake


raypower178:
## Summary

On Windows (system locale zh-TW, ACP=CP950), `search_code` never matches CJK patterns even when the pattern verifiably exists in UTF-8 files, and `raw_matches[].content` returns mojibake with `?` replacement characters. The tree-sitter/snippet pipeline (`get_code_snippet`) handles the same files' CJK content perfectly — only the grep/content pipeline is affected.

## Environment

- codebase-memory-mcp 0.9.0 (standard variant, Windows amd64)
- Windows 11 Home 10.0.26200, system locale zh-TW (ANSI codepage 950)
- Repo: Python project, all source files verified UTF-8 (no BOM), indexed via a junction path (ASCII) to work around #903

## Repro

1. A UTF-8 Python file contains the string `未來函數` (verified: present in 28 files; `file` reports "UTF-8 text" for all of them).
2. Call `search_code` with `{"pattern": "未來函數", "project": "..."}`.
3. Result: `total_grep_matches: 0, total_results: 0`.

Control A (ASCII pattern, same index): `search_code` for `DualTrackRouter` → 55 grep matches, works fine.

Control B (same file, snippet pipeline): `get_code_snippet` on a function whose docstring contains CJK → CJK returned byte-perfect.

## Evidence of ANSI-codepage transcoding

For an ASCII pattern that lands on a line that *also* contains CJK, `raw_matches[].content` comes back like:

```
"content": "                    # ?w?w STEP 2d?G???J???T???J?w?y?{?]DualTrackRouter?^?A"
```

The original UTF-8 line is `── STEP 2d：載入資訊…（DualTrackRouter）…`. This `?`-folding pattern is the signature of routing UTF-8 bytes through `WideCharToMultiByte`/`MultiByteToWideChar` with CP_ACP (CP950 here): characters unmappable in the ANSI codepage become `?`. Since the grep index/content path stores (or compares) the transcoded text, a UTF-8 CJK pattern can never match.

## Expected behavior (industry baseline)

[...]

PR #1253, opened 2026-07-24
fix: preserve UTF-8 in search_code source

anfedoro:
## Summary

- Preserve valid UTF-8 in `search_code` full results and raw match content.
- Replace the ASCII-only sanitizer that converted every byte >= 0x80 to `?` with the existing UTF-8-safe lossy sanitizer.
- Add a regression test using Russian Markdown content.

## Root cause

`search_code` constructed result fields through `sanitize_ascii()`. Because UTF-8 characters are encoded as bytes >= 0x80, the sanitizer replaced their individual bytes with `?` after the file and line match had already been found correctly.

## Scope

This PR addresses result serialization after a successful match. It intentionally does not address the separate Windows CP_ACP pattern-matching issue discussed in #1209.

Related to #1209

## Validation

- `make -f Makefile.cbm test-focused TEST_SUITES=mcp` — 179 passed, 2 skipped.
- Production build via `scripts/build.sh` on macOS arm64.
- Manual `search_code --mode full` smoke test confirmed Cyrillic is preserved in `source`.


DeusData:
Thanks for opening the focused split requested in #1209. The exact patch stays in post-match serialization, reuses the existing `sanitize_utf8_lossy` implementation, and does not claim to fix Windows CP_ACP matching. No dependency, workflow, network, prompt, or raw-file-I/O surface is added. The Russian full-source fixture is a useful RED guard. Before exact-head approval, please add coverage for the other behavior claimed by the PR: raw match `content`, attached `context`, and one invalid UTF-8 sequence proving lossy replacement still yields valid JSON. Both commits currently lack visible DCO sign-offs, so the branch also needs a signed-off commit history and the complete CI result.

DeusData:
A quick note so this does not sit here looking like your problem: **your red CI is not caused by your change — and a rebase will not fix it either.**

The failure is `tests/test_daemon_version.c:604 ASSERT(all_succeeded)`, a Windows concurrent conflict-log append test. It is a **live flake on `main` itself**, still failing as of yesterday — so it is ours to fix, not something you can clear by rebasing.

I have flagged it internally as a production-race candidate rather than a test to tune: under our own testing doctrine a test whose verdict depends on scheduler timing is a defect, and the right response is a deterministic repro plus a fix, not a wider budget.

So: **nothing is needed from you on the CI front.** Your PR is queued for a proper content review and I will come back to you on the change itself. Apologies that a failure of ours has been sitting on your PR looking like yours.

DeusData:
Merged — thank you! Swapping the four search_code sites onto the existing UTF-8-safe lossy sanitizer fixes real pain for every non-Latin codebase, and we took the context-column addition with it: a parameter that silently did nothing was its own small bug, and your tests cover both formats. The Windows CP_ACP matching gap you scoped out cleanly remains tracked in #1209.

PR #1255, opened 2026-07-24
fix(installer): stop PS 5.1 stderr wrapping from aborting version probes

MasterFede5:
## What does this PR do?

Fixes a Windows PowerShell 5.1 failure mode in `install.ps1`: the script sets `$ErrorActionPreference = "Stop"` globally, and PS 5.1 (the default `powershell.exe` the README instructs users to run) wraps redirected native stderr in ErrorRecords. The two `--version 2>&1` probes (candidate at ~L265, installed binary at ~L288) therefore abort the whole install for any healthy binary that prints a single warning line to stderr while exiting 0 — reporting `error: downloaded binary failed to run`.

The fix relaxes EAP to `Continue` only around the two native probe calls, restoring it in a `finally` block. Failure detection is unchanged: it stays on `$LASTEXITCODE`.

**Reproduction** (PS 5.1.26100.8875, `powershell.exe -File`): stub binary printing a version to stdout + one warning to stderr, exit 0 → previous pattern aborts with the stderr text as the exception; new pattern captures the version output and still detects a nonzero exit (tested with exit 3). EAP is restored to `Stop` afterwards.

The literal invariants checked by `tests/test_windows_bundle_contract.sh` (`& $DownloadedLauncher --version`, launcher-only invocation) are preserved — that contract test passes locally.

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) — not run: no C toolchain on this machine; the change touches only `install.ps1`. `tests/test_windows_bundle_contract.sh` (the suite that asserts installer invariants) passes locally.
- [x] Lint passes (`make -f Makefile.cbm lint-ci`) — C linters not applicable to this script-only change; `install.ps1` parses clean (PSParser, 0 errors).
- [x] New behavior is covered by a test (reproduce-first for bug fixes) — reproduced first with a stderr-writing stub as described above; the existing bundle contract test still guards the probe invariants. Happy to add a harness test if you point me at the preferred place for installer regression tests.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -262,7 +262,17 @@ if (($launcherItem.Attributes -band [System.IO.FileAttributes]::ReparsePoint) -o
 # Verify the complete launcher/payload path before it asks all coordinated CBM
 # processes to stop.
 try {
-    $candidateVersion = & $DownloadedLauncher --version 2>&1
+    # Windows PowerShell 5.1 wraps redirected native stderr in ErrorRecords, so
+    # under $ErrorActionPreference = "Stop" a healthy candidate that prints any
+    # warning would abort here. Probe with EAP relaxed; failure detection stays
+    # on $LASTEXITCODE.
+    $PrevErrorActionPreference = $ErrorActionPreference
+    try {
+        $ErrorActionPreference = "Continue"
+        $candidateVersion = & $DownloadedLauncher --version 2>&1
+    } finally {
+        $ErrorActionPreference = $PrevErrorActionPreference
+    }
     if ($LASTEXITCODE -ne 0) { throw "candidate exited with $LASTEXITCODE" }
     Write-Host "Verified candidate: $candidateVersion"
 } catch {
@@ -285,7 +295,14 @@ if ($LASTEXITCODE -ne 0) {
 
 # Verify
 try {
-    $ver = & $Dest --version 2>&1
+    # Same PS 5.1 stderr-wrapping hazard as the candidate probe above.
+    $PrevErrorActionPreference = $ErrorActionPreference
+    try {
+        $ErrorActionPreference = "Continue"
+        $ver = & $Dest --version 2>&1
+    } finally {
+        $ErrorActionPreference = $PrevErrorActionPreference
+    }
     if ($LASTEXITCODE -ne 0) { throw "installed binary exited with $LASTEXITCODE" }
     Write-Host "Installed: $ver"
 } catch {


DeusData:

A +19/-2 single-file change is easy to review — nothing blocking from a size standpoint. CI shows 1-2 failing checks worth checking while it waits.

PowerShell 5.1 wrapping stderr and thereby aborting verification is a nasty class of bug: it makes a working install look broken for reasons that have nothing to do with the install. Good catch.

DeusData:
Great catch, and a textbook writeup. **The hazard is real and still live on current main** — I verified it rather than taking your word for it: `install.ps1:8` sets `$ErrorActionPreference = "Stop"` globally, and both probes run bare underneath it (`:259` and `:327`). PowerShell 5.1 wraps redirected native stderr in ErrorRecords, so under `Stop` the first stderr line becomes terminating — meaning a **healthy** binary that prints any warning to stderr while exiting 0 aborts the install with "downloaded binary failed to run".

That is exactly the kind of latent installer bomb that only detonates the day the binary prints its first warning, and with the daemon work adding warning paths it is a plausible one. Your scoped `try`/`finally` with detection kept on `$LASTEXITCODE` is the right conservative shape — the failure path is untouched, and the restore is exception-safe.

**Thank you for the pure-ASCII discipline.** I byte-scanned the patch and it is clean, comments deliberately avoiding em-dashes. We have been burned by a single non-ASCII character in this exact file before — PowerShell 5.1 decodes a BOM-less `.ps1` as ANSI and one em-dash injected a stray quote that made `install.ps1` unparseable on main. Not everyone knows that; you clearly did.

**Three things needed, all small:**

**1. A mechanical rebase.** Main removed the launcher stub (`a54ea95`), so `$DownloadedLauncher` is now `$DownloadedBinary` and the probes sit at `install.ps1:259` and `:327`. Your fix survives the rename intact. Worth knowing: the literal invariant pinned at `tests/test_windows_bundle_contract.sh:206` requires the substring `& $DownloadedBinary --version`, and the rebased line still contains it since `2>&1` comes after — so that contract stays green.

**2. One hardening ask — a narrow false-pass window.** With `EAP=Continue`, a binary that fails to *launch* (corrupt or wrong-architecture image, as opposed to exiting nonzero) may surface as a non-terminating error, in which case `$LASTEXITCODE` is not updated. For probe 1 that fails safe: no native command has run yet, so `$LASTEXITCODE` is `$null` and the `-ne 0` check throws. But probe 2 has prior native calls before it (probe 1 exiting 0, and `& $Dest daemon stop`), so a **stale 0 could false-pass a corrupt install** and print "Installed: …" alongside an error record. Pre-fix, `EAP=Stop` promoted that to terminating and aborted correctly.

Pre-seeding `$global:LASTEXITCODE = 1` inside each relaxed block before the call closes it, and costs nothing. If you get the chance on your 5.1 box, it would also be genuinely useful to confirm how a start-failure actually surfaces there — I could not settle that from source alone.

**3. Yes please to the test you offered.** You mentioned you would add a harness test if pointed at the right place: a stderr-writing stub under the Windows VM `win.sh smoke-install`/guards family is the right home. Worth saying why it matters — our real-Windows smoke leg runs `install.ps1` but with a stderr-quiet binary, so it would not have caught the original hazard either. At minimum, a contract pin on the EAP save/restore block would stop the fix being silently removed later.

With the rebase and the pre-seed this merges. Thanks for a precise, well-scoped fix.

DeusData:
Your diagnosis is exactly right and the bug is still alive on main: install.ps1 sets $ErrorActionPreference='Stop' at line 8 and probes with bare '& $binary --version 2>&1' at lines 285 and 353, so PS 5.1 wraps redirected native stderr into ErrorRecords and a healthy binary printing any warning aborts the install. The branch conflicts only because it patches the removed launcher-era script — the launcher was eliminated in favor of a single binary per platform. Same two probe sites, same EAP save/relax/restore pattern you wrote, just on the current file. Rebase and this merges; if you'd prefer, we'll distill the ~10-line change with credit — your call.

DeusData:
@MasterFede5 — your fix is on `main` in 638b1164 (via #1869), with `Co-authored-by:` credit to you in the commit.

The 31 July review approved your shape twice and offered a maintainer-side distill if you preferred; after five quiet weeks I took that path rather than let a priority/high installer bug rot further. What shipped is your diagnosis and your conservative structure — probe-scoped `$ErrorActionPreference = "Continue"` with an exception-safe restore, failure detection still on `$LASTEXITCODE` — rebased onto the single-binary installer and completed with the review's two conditions: the `$global:LASTEXITCODE = 1` pre-seed (closing the start-failure false-pass window on the installed-binary probe), and a bundle-contract pin on both load-bearing parts, verified failing when either is stripped.

Your PS 5.1 reproduction with the stderr-writing stub and the exit-3 negative test is what made this reviewable in one pass — thank you for a precise piece of Windows detective work. Closing this PR in favour of the distilled commit.

Issue #1211, opened 2026-07-22
index_repository by project name fails with 'repo_path is required' but reports 'crashed on a file'


asemraza:
**Version:** 0.9.0, macOS (Darwin 25.5, arm64)

**Repro:**
1. Have an already-indexed project (indexed originally by path).
2. Call the MCP tool index_repository with only the project name: `{"project": "<name>"}` (or equivalently `cli index_repository '{"project": "<name>"}'`).
3. Result: `{"status":"error","outcome":"exit_nonzero","hint":"Indexing worker crashed on a file..."}`

**Actual cause** (from the worker log referenced by index.supervisor.worker_failed):
```
repo_path is required
level=info msg=index.worker.fast_exit action=_Exit
```
The worker exits because it never receives repo_path when the call is made by project name - no file crash is involved.

**Two issues:**
1. Re-indexing an existing project by name does not resolve the stored repo path before spawning the worker (list_projects clearly knows the root_path, so resolution data exists).
2. The supervisor's generic hint ("crashed on a file... a future release isolates the culprit file") is misleading for this failure class - surfacing the worker's actual stderr line would have saved the debugging detour.

**Workaround:** call with the explicit path: `cli index_repository --repo-path /path/to/repo` - works perfectly (and the built-in exclusion list in 0.9.0 is excellent, thanks for .claude/.codex).

Great tool otherwise - we run it across a multi-repo Laravel/TS setup shared between Claude Code and Codex.

DeusData:
Thanks for including the actual worker error. This is a `0.9.1-rc` integration/diagnostic bug under #395. Re-indexing an existing named project should resolve its stored root before the worker is spawned; invalid invocation must return the specific argument error rather than the generic file-crash hint.

xantorres:
Another instance of the same misleading-message family, different root cause.

When `CBM_ALLOWED_ROOT` refuses a path, the CLI reports:

```json
{"status":"error","outcome":"exit_nonzero",
 "hint":"Indexing worker crashed on a file. The crash was contained (the server survived). Re-run to retry; a future release isolates the culprit file.",
 "repo_path":"/etc"}
```

Nothing crashed. The worker log has the real reason:

```
repo_path is outside the allowed root
level=info msg=index.worker.fast_exit action=_Exit
```

[...]

PR #1258, opened 2026-07-25
fix(mcp): resolve index_repository's repo_path from a known project name

AmirF194:
## What does this PR do?

`handle_index_repository` requires `repo_path` and errors with `"repo_path is required"` when it is absent. `list_projects` only ever advertises the project NAME, never `repo_path`, and every read tool accepts that name back through `get_project_arg`'s `project`/`project_name`/`project_id`/`projectName` aliases (mcp.c:1385), but `index_repository` did not, so re-indexing an already-indexed project by the name `list_projects` gave you had no resolution path and failed immediately, even though the project's root was already on file.

Fix: added `resolved_repo_path_from_project_arg()`, a query-only lookup of the project's own stored `root_path`, tried when `repo_path` is absent before giving up. It never touches `srv->store` or `srv->current_project`, so it cannot disturb whatever project the server already has cached, and an unresolvable name still falls through to the original, correct `"repo_path is required"` error.

Verified in a clean `ubuntu:24.04` container (`TEST_SUITES=mcp make -f Makefile.cbm test-focused`):
- New test `tool_index_repository_resolves_root_path_from_project_name_issue1211` fails on unpatched main (`strstr(resp, "repo_path is required")` is not NULL) and passes with the fix.
- Companion test `tool_index_repository_unknown_project_name_still_requires_repo_path` guards the fallback: a name with no stored project still gets the exact original error, not a resolver crash or silent no-op.
- Full `mcp` suite: 179 passed on both sides, one pre-existing, unrelated failure (`tool_detect_changes_contained_commands_clean_up_error_and_success`, test_mcp.c:5814) reproducing identically on unmodified `origin/main`, so not introduced by this change.
- `scripts/lint.sh --ci CLANG_FORMAT=clang-format-20` and `scripts/security-audit.sh` both clean.

Refs #1211 (this PR fixes only the repo_path-required/name-resolution bug the issue was filed for; a separate, unrelated defect was reported in the thread afterward: CBM_ALLOWED_ROOT refusals get mis-reported as a worker crash. Leaving the issue open rather than auto-closing it on merge.)

## Checklist

- [x] Tests pass locally (`make -f Makefile.cbm test`, focused on the `mcp` suite)

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -7566,11 +7566,46 @@ static char *index_args_with_repo_path(const char *args, const char *canonical_r
     return rewritten;
 }
 
+/* #1211: index_repository requires repo_path, but list_projects only ever
+ * advertises the project NAME, and every read tool accepts that name back via
+ * get_project_arg's "project"/"project_name"/"project_id"/"projectName"
+ * aliases (mcp.c:1385). Re-indexing an already-indexed project by that same
+ * name had no resolution path and fell straight to "repo_path is required".
+ * Look up the project's own stored root_path (list_projects proves it's on
+ * file) before giving up. Query-only open, always closed here: this never
+ * creates a store or touches srv->store/srv->current_project, so it cannot
+ * disturb whatever project the server has cached. */
+static char *resolved_repo_path_from_project_arg(const char *args) {
+    char *project = get_project_arg(args);
+    if (!project) {
+        return NULL;
+    }
+    char db_path[CBM_SZ_1K];
+    project_db_path(project, db_path, sizeof(db_path));
+    cbm_store_t *store = db_path[0] ? cbm_store_open_path_query(db_path) : NULL;
+    char *root_path = NULL;
+    if (store) {
+        cbm_project_t proj = {0};
+        if (cbm_store_get_project(store, project, &proj) == CBM_STORE_OK) {
+            root_path = proj.root_path ? heap_strdup(proj.root_path) : NULL;
+            cbm_project_free_fields(&proj);
+        }
+        cbm_store_close(store);
+    }
+    free(project);
+    return root_path;
+}
+
 static char *handle_index_repository(cbm_mcp_server_t *srv, const char *args) {
     char *repo_path = cbm_mcp_get_string_arg(args, "repo_path");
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for the contribution and for isolating the `list_projects` and `index_repository` mismatch with regression coverage. This is now triaged as a `0.9.1-rc` integration bug at normal priority. Our community PR queue is currently quite full, so it may take a little time before we can complete the review and, if approved, merge it. We are doing our best to support community contributions and will return with code-grounded feedback as capacity opens.

AmirF194:
Thanks for the update, no rush. Happy to adjust the fix if the 0.9.1-rc review turns up anything.

DeusData:
Merged — thank you. This removes a genuinely annoying papercut: every read tool accepts a project name, and `index_repository` was the one that demanded a path.

Because this resolves a path from stored state, I checked the authorization question specifically rather than assuming: the project name is validated first (`cbm_validate_project_name` rejects `..`, path separators, leading dots and anything outside `[a-zA-Z0-9._-]`), the store is opened query-only so a lookup cannot create anything, and — the part that actually matters — the fallback runs **before** `resolve_session_repo_path`, canonicalisation and the `CBM_ALLOWED_ROOT` boundary check. So a resolved `root_path` is re-validated exactly like a caller-supplied one, and a project indexed under an older, looser policy cannot smuggle a path past the current allowed root. That was the failure mode worth ruling out.

Also verified: no leaks on any path, `cbm_normalize_path_sep(NULL)` is safe, and an unknown name still falls through to the original error rather than silently doing something surprising. The negative test pins that.

Two follow-ups, neither blocking. The tool's input schema still declares `"required": ["repo_path"]`, so a strict MCP client will reject a project-only call before it ever reaches this code — worth a schema tweak, yours or ours. And the fallback covers exact and tail-resolved names but not the internal-name-drift case from #704, so a drifted DB still errors; a reasonable gap to leave open.

Good call leaving #1211 open for the separate `CBM_ALLOWED_ROOT`-misreported-as-crash defect rather than folding it in here.

AmirF194:
Appreciate the thorough check on the authorization path, that's exactly the failure mode worth ruling out. Will follow up on the schema tweak separately.

PR #1259, opened 2026-07-25
fix(install): accept flat Windows release archives without payload wrapper (#1257)

### LINKED ISSUE #1257: install.ps1 rejects v0.9.0 windows-arm64 UI archive: 'archive must contain exactly one codebase-memory-mcp.payload.exe'
﻿## Title

`install.ps1` rejects v0.9.0 `codebase-memory-mcp-ui-windows-arm64.zip` — "archive must contain exactly one codebase-memory-mcp.payload.exe"

## Environment

- OS: Windows 11 on ARM64
- Installer: `install.ps1` from `main` branch (raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.ps1)
- Release: v0.9.0 (latest)
- Command:
  ```powershell
  Invoke-WebRequest -Uri https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.ps1 -OutFile install.ps1
  Unblock-File .\install.ps1
  .\install.ps1 --ui --skip-config
  ```

## Output

```
codebase-memory-mcp installer (Windows)
  variant: ui
  arch:    arm64
  target:  C:\Users\srini\AppData\Local\Programs\codebase-memory-mcp\codebase-memory-mcp.exe

Downloading codebase-memory-mcp-ui-windows-arm64.zip...
Checksum verified.
error: unsafe or incomplete release archive: archive must contain exactly one codebase-memory-mcp.payload.exe
```

## Expected

Archive passes layout validation, installer stages the binary and the bundled `install.ps1` to the target directory, and exits 0.

## Actual

Installer exits non-zero after the layout check. The downloaded archive contents are:

```
codebase-memory-mcp.exe      274,818,560 bytes
install.ps1                       6,450 bytes
LICENSE                           1,128 bytes
THIRD_PARTY_NOTICES.md          570,451 bytes
```

I confirmed the `windows-amd64` UI archive has the same flat layout (`codebase-memory-mcp.exe` at the root, not `*.payload.exe`), so this is consistent across Windows builds, not specific to ARM64.

## Root cause (observed)

The installer archive-layout check expects a `*.payload.exe` wrapper, but the actual release zips for v0.9.0 Windows builds do not contain that wrapper — only a flat `codebase-memory-mcp.exe`. The checksum itself passes; the failure is purely the layout assertion.

## Workaround

`--skip-config` is already requested, so manually extracting the archive into `%LOCALAPPDATA%\Programs\codebase-memory-mcp\` and launching `codebase-memory-mcp.exe --ui=true --port=9749` works fine. The binary itself is healthy (0/72 on VirusTotal, UI responds 200 on http://localhost:9749).

## Suggested fix

Either:
1. Update the layout check in `install.ps1` to accept `codebase-memory-mcp.exe` (or `codebase-memory-mcp-ui.exe`) at the archive root as a valid Windows layout, **or**
2. Rebuild the Windows release zips to include the `*.payload.exe` wrapper that the current check requires.

Option 1 is preferable — the wrapper name appears to be a leftover from a self-extracting-installer design and adds no value when the binary is already static.

## SHA-256

- windows-arm64 binary from release page: `a299150de19bb4da9a880a6139757076bd4139b2f6a7e7cf31bdd431d53aed9c`
- installer-printed SHA matched the downloaded archive before the layout check rejected it.

skbs-eng:
Fixes #1257

### Problem
install.ps1's archive namespace check strictly required codebase-memory-mcp.payload.exe to exist as a separate entry in the release zip. In release builds where codebase-memory-mcp.exe is packaged as a flat standalone binary without a separate *.payload.exe entry (such as the v0.9.0 windows-arm64 and windows-amd64 UI archives), install.ps1 aborted with:
\rror: unsafe or incomplete release archive: archive must contain exactly one codebase-memory-mcp.payload.exe\

### Fix
- Updated install.ps1 archive layout validation to require codebase-memory-mcp.exe, LICENSE, install.ps1, and THIRD_PARTY_NOTICES.md, while allowing codebase-memory-mcp.payload.exe to be optional (0 or 1 instance).
- If codebase-memory-mcp.payload.exe is missing from the extracted archive, install.ps1 stage-copies codebase-memory-mcp.exe to codebase-memory-mcp.payload.exe in the temporary directory prior to executing the launcher installation.
- Updated 	ests/test_windows_bundle_contract.sh static contract checks to reflect the updated archive bounds.

### Validation
- Verified with python contract test script (\	est_windows_bundle_contract.sh\).
- Tested \install.ps1\ archive extraction against both 4-file (flat layout) and 5-file (pair layout) zip archives.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -220,12 +220,21 @@ try {
             }
             $archiveCounts[$entryName] = $archiveCounts[$entryName] + 1
         }
-        foreach ($archiveName in $WindowsArchiveNames) {
+        $RequiredArchiveNames = @(
+            $BinName,
+            "LICENSE",
+            "install.ps1",
+            "THIRD_PARTY_NOTICES.md"
+        )
+        foreach ($archiveName in $RequiredArchiveNames) {
             if ($archiveCounts[$archiveName] -ne 1) {
                 throw "archive must contain exactly one $archiveName"
             }
         }
-        if ($seen.Count -ne $WindowsArchiveNames.Count) {
+        if ($archiveCounts[$PayloadName] -gt 1) {
+            throw "archive contains duplicate $PayloadName"
+        }
+        if ($seen.Count -lt $RequiredArchiveNames.Count -or $seen.Count -gt $WindowsArchiveNames.Count) {
             throw "archive does not match the exact Windows release allowlist"
         }
     } finally {
@@ -244,6 +253,10 @@ Expand-Archive -Path "$TmpDir\$Archive" -DestinationPath $TmpDir -Force
 
 $DownloadedLauncher = Join-Path $TmpDir $BinName
 $DownloadedPayload = Join-Path $TmpDir $PayloadName
+if (-not (Test-Path -LiteralPath $DownloadedPayload -PathType Leaf) -and
+    (Test-Path -LiteralPath $DownloadedLauncher -PathType Leaf)) {
+    Copy-Item -LiteralPath $DownloadedLauncher -Destination $DownloadedPayload
+}
 if (-not (Test-Path -LiteralPath $DownloadedLauncher -PathType Leaf) -or
     -not (Test-Path -LiteralPath $DownloadedPayload -PathType Leaf)) {
     Write-Host "error: launcher or payload not found after extraction" -ForegroundColor Red
--- a/tests/test_windows_bundle_contract.sh
+++ b/tests/test_windows_bundle_contract.sh
@@ -367,7 +367,7 @@ require(
 # attacker-controlled sixth member.
 exact_archive_guards = {
     "install.ps1": (
-        "$seen.Count -ne $WindowsArchiveNames.Count",
+        "$seen.Count -gt $WindowsArchiveNames.Count",
         '"LICENSE"',
         '"install.ps1"',
         "THIRD_PARTY_NOTICES.md",


DeusData:
Thank you for the contribution and for validating both Windows archive layouts. This is now triaged as a high-priority Windows installation bug for `0.9.1-rc`. Current main already contains an overlapping fix for #1257, so review will determine whether this branch still contains any distinct work. Our community PR queue is currently quite full, so it may take a little time before we can complete that review. We are doing our best to support community contributions and will return with code-grounded feedback as capacity opens.

DeusData:
Thank you — your diagnosis of #1257 was correct, and I am sorry this sat long enough for the ground to move under it.

The reason I am closing rather than asking for a rebase is that a rebase would leave nothing to apply. #1257 was fixed on `main` by a more radical route: the launcher/payload pair is gone entirely. Windows now ships **one binary**, `install.ps1` has no payload concept left in it (the allowlist is an exact four-entry set), and `package-release.sh` zips precisely those four members. The commit that did it is a54ea95, with b0eae97 and afd67c5 following on. Your PR's hunks anchor to code that no longer exists, which is why GitHub shows it as conflicting, and #1257 is already closed.

One thing I want to flag honestly, because it is useful feedback rather than criticism: the shim relaxed the archive guard from an exact allowlist to a range check, and softened the contract test's assertion from an equality to a bound. That guard exists specifically to stop an attacker-controlled extra member riding along in a release archive, so even as a compatibility measure it would have traded away a little more than it gained. Nothing about it was careless — it is simply the kind of trade worth naming out loud when the thing being loosened is a security guard.

Genuinely appreciated: you found a real packaging break and reported it precisely. Closing as already fixed.

PR #1263, opened 2026-07-25
fix: rebuild index when mode adds capabilities

### LINKED ISSUE #1273: Mode upgrade silently keeps weaker graph: incremental routing ignores stored index capability level
### Version

main @ 97ce23f9 (0.9.x line)

### Platform

macOS (Apple Silicon) — the routing logic is platform-independent C, so all platforms are affected; verified on macOS.

### Install channel

Built from source

### Binary variant

standard

Indexing a repository with a weaker mode and then re-running with a stronger mode on an unchanged repo silently keeps the weaker graph. The incremental router compares stored file hashes and the discovered-file threshold, but never the capability level of the stored index, so a `fast`-built graph is treated as valid for `moderate`/`full` requests.

Expected: when the requested mode requires capabilities missing from the stored index (e.g. `full` builds SIMILAR_TO / SEMANTICALLY_RELATED edges that `fast` never produces), the run forces a full rebuild. Conversely, re-running with a weaker requested mode (as the failure hint on large repos suggests, cf. #563) should not tear down stronger capabilities already present in the graph.

### Reproduction

1. Clone any small public repo, e.g. `colinhacks/zod`, to `/tmp/zod`.
2. `codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/zod","mode":"fast"}'` → completes with `pipeline.route path=full`; graph contains no full-mode-only edges.
3. Re-run unchanged: `codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/zod","mode":"moderate"}'`.

Actual: the log shows the incremental no-op path; the graph never gains the capabilities `moderate`/`full` are supposed to add, and nothing signals that the stored index is weaker than requested.

Expected: the router detects the capability gap and forces a full rebuild.

Symmetric downgrade case: after a `full` index, re-run with `"mode":"fast"` on a repo with changed files. Before the fix this rebuilt from scratch at the weaker mode, permanently discarding the stronger capabilities; expected is a run that keeps the full graph and only narrows the requested view.

### Logs

```text
level=info msg=pipeline.route path=incremental
# ^ after fast → moderate on an unchanged repo; should have been a forced full rebuild
```

Reproduces on repos of any size (observed in the wild on a ~124k-file repo, see #563).

- [x] My reproduction uses shareable code (a public OSS repository), not proprietary code.

astandrik:
## Summary

Fixes #1273
Related: #563

- keep the caller-requested discovery/reporting mode separate from the stronger effective coverage already stored in `index_coverage_meta`
- compare downgrade runs against a fresh semantic manifest at the stored effective scope
- rebuild at that effective scope when a stored-only input changes or disappears, without widening the caller-visible excluded/ignored report
- export explicit and existing persistent artifacts exactly once, after the replacement database is atomically published

## What this PR completes

Current main already stores the canonical mode in `index_coverage_meta` and forces a rebuild when a stronger request needs capabilities missing from the stored generation. This PR does not reimplement that upgrade route. It completes the remaining downgrade half of #1273 and fixes artifact publication ordering in the same pipeline path.

Previously, preserving stronger stored coverage promoted the pipeline's single mutable mode before discovery. A `full -> fast` request therefore widened back to full discovery/reporting. Conversely, comparing only the requested subset could leave changed or deleted full-only inputs stale.

The pipeline now keeps `requested_mode` and effective `mode` separate. It reports exclusions from requested discovery, but when stored coverage is stronger it compares the stored generation with a fresh manifest built at that effective scope. An exact match remains an incremental no-op. Any stored-only change or deletion forces a complete effective-scope rebuild, after which the requested exclusion/ignored report is restored.

Artifact export is centralized at the outer publication boundary. No route exports from the staging database; explicit persistence and refresh of an existing artifact happen only after the final database rename.

## Mode contract

| Stored coverage | Requested mode | Result |
| --- | --- | --- |
| weaker | stronger | existing current-main route rebuilds and records the stronger coverage |
| same | same | existing exact no-op / repair / rebuild routing |
| stronger | weaker, effective manifest unchanged | requested-scope reporting, stronger capabilities retained, incremental no-op |
| stronger | weaker, effective-only input changed or deleted | effective-scope full rebuild, requested-scope reporting retained |

Cancellation or a staging/publication failure before the final rename preserves the previous database and artifact. An explicitly requested artifact-export failure is returned after the database has been published; automatic refresh of an existing artifact remains best-effort.

The public MCP API, SQLite schema, and artifact format are unchanged. `index_coverage_meta` remains the sole mode representation. This PR intentionally does not add duplicate Project-node metadata, special normalization of externally modified/corrupt mode values, or generic SQLite read-error hardening; those are separate artifact-integrity/recovery concerns rather than normal #1273 mode transitions.

## Regression coverage

- an unchanged `fast -> moderate` run forces a full rebuild, creates `SIMILAR_TO`, records `index_mode=moderate`, and refreshes the artifact; a later `fast` request retains those capabilities
- `full -> fast` reports `tools/` as excluded while retaining FULL stored coverage
- changed and deleted full-only inputs force an effective-FULL rebuild
- an unchanged downgrade takes the incremental no-op route
- cancellation preserves both the previous database and existing artifact; retry publishes the new generation
- explicit full, implicit forced-full, explicit no-op, and implicit no-op paths each export exactly once after the real rename
- explicit artifact-export failures still propagate after database publication

## Verification

Exact published head: `3a4ad3a2` on base `43531195`.

- exact current-main affected matrix: `343 passed` (`pipeline`, `pipeline_semantic_manifest_repro`, `artifact`, and upstream `mem`)
- focused UBSan matrix on the byte-identical candidate: `42 passed`
- post-rewrite compatibility audit: seven representation-independent scenarios from the previous implementation passed on the current product code (`32 passed` in a disposable focused fixture)
- fresh isolated `fast -> moderate -> fast` capability contract: `1 passed`
- production build, formatting, NOLINT policy, `git diff --check`, diff-only clang-format, static security audit, cppcheck, and CI lint: passed
- local macOS ASan remained blocked inside sanitizer runtime initialization before any test; no product failure was observed

Exact-head GitHub CI is still running; clean-runner sanitizer, platform, and analyzer jobs remain authoritative.

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -406,7 +406,8 @@ static const tool_def_t TOOLS[] = {
      "are normalized.\"},"
      "\"persistence\":{\"type\":\"boolean\",\"default\":false,\"description\":"
      "\"Write compressed artifact to .codebase-memory/graph.db.zst for team sharing. "
-     "Teammates can bootstrap from the artifact instead of full re-indexing.\"}"
+     "Teammates can bootstrap from the artifact instead of full re-indexing. "
+     "An existing artifact is refreshed after any reindex even when persistence is false.\"}"
      "},\"required\":[\"repo_path\"]}"},
 
     {"search_graph", "Search graph",
--- a/src/pipeline/pipeline.c
+++ b/src/pipeline/pipeline.c
@@ -36,6 +36,7 @@ enum { CBM_DIR_PERMS = 0755, PL_RING = 4, PL_RING_MASK = 3, PL_SEQ_PASSES = 6 };
 #include "foundation/compat_thread.h"
 #include "foundation/profile.h"
 #include "foundation/mem.h"
+#include "yyjson/yyjson.h"
 
 #include <errno.h>
 #include <stdint.h>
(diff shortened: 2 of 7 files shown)


DeusData:
Thank you for the contribution and for covering both index mode upgrades and downgrades, including malformed metadata. This is now triaged as a high-priority graph correctness bug for `0.9.1-rc`. Our community PR queue is currently quite full, so it may take a little time before we can complete the review and, if approved, merge it. We are doing our best to support community contributions and will return with code-grounded feedback as capacity opens.

DeusData:
Merged (415a64a2) — thank you for what was the technically strongest external PR in this batch. The export-exactly-once-after-rename fault-injection tests are the reason this could land as one unit: they prove the publish-boundary design under cancel/retry/delete, which is precisely the evidence a ~400-line rework of the pipeline's most critical routing needs. The requested/effective mode split completes #1273's downgrade half cleanly. We chose your centralized export over our own per-route refresh design — #1163 is being closed in its favor with credit.

PR #1268, opened 2026-07-25
docs(npm): fix stale MCP tool count in package README

nhtkylc91:
## Summary
- \`pkg/npm/README.md\` claimed **14 MCP tools**, but the actual \`TOOLS[]\` registry in \`src/mcp/mcp.c\` defines 15 (verified by counting entries in the array), matching the count already stated in the root \`README.md\`.
- Fixes both occurrences in the npm package README so it's consistent with the source of truth and the other docs.

## Test plan
- [x] Counted tool definitions in \`src/mcp/mcp.c\`'s \`TOOLS[]\` array: 15.
- [x] Confirmed root \`README.md\` and \`pkg/npm/README.md\` now agree (15).
- [x] Docs-only change, no code/behavior affected.

### DIFF
--- a/pkg/npm/README.md
+++ b/pkg/npm/README.md
@@ -7,7 +7,7 @@
 
 **The fastest and most efficient code intelligence engine for AI coding agents.** Full-indexes an average repository in milliseconds, the Linux kernel (28M LOC, 75K files) in 3 minutes. Answers structural queries in under 1ms. Ships as a single static binary — this package downloads and runs it automatically.
 
-High-quality parsing through [tree-sitter](https://tree-sitter.github.io/tree-sitter/) AST analysis across 159 languages — producing a persistent knowledge graph of functions, classes, call chains, HTTP routes, and cross-service links. 14 MCP tools. Zero dependencies. Plug and play across 43 automatic/conditional client surfaces.
+High-quality parsing through [tree-sitter](https://tree-sitter.github.io/tree-sitter/) AST analysis across 159 languages — producing a persistent knowledge graph of functions, classes, call chains, HTTP routes, and cross-service links. 15 MCP tools. Zero dependencies. Plug and play across 43 automatic/conditional client surfaces.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you for the contribution and for checking the package README against the tool registry. This is now triaged as a normal-priority documentation correction for `0.9.1-rc`. The tool-count point overlaps the newer #1297 and #1315 report and will need to be reconciled against the actual `tools/list` response during review. Our community PR queue is currently quite full, so it may take a little time before we can complete that review. We are doing our best to support community contributions and will return with code-grounded feedback as capacity opens.

DeusData:
Thank you — this one is simply correct. The count is 15: `check_index_coverage` shipped as the 15th tool in v0.9.1, and the npm README still said 14.

(Amusing companion note: another PR is currently proposing the *opposite* change to the main README, because it was matching `--help` — which was itself omitting that same 15th tool. #1370 fixes the help text to derive from the tool registry so the two cannot drift again. Yours is the version that matches reality.)

I cannot merge it yet, and the reason is on our side rather than yours. `main` requires two status checks, `dco` and `ci-ok`, and **neither ran on this PR** — the only file it touches, `pkg/npm/README.md`, sits outside the path filters that trigger our workflows, so the required checks never report and GitHub holds the merge as `BLOCKED`.

If you would not mind: **rebase onto current `main` and push the branch again**. That normally re-triggers the workflow set so the checks can report, after which this goes straight in.

If it still shows no checks afterwards, please say so and leave it with me — at that point it is our path-filter configuration that needs fixing rather than anything about your patch, and I will not ask you to keep poking at it. Thanks again for catching this.

DeusData:
Thank you — **you are right, and I verified it independently rather than taking the number on trust.**

`TOOLS[]` in `src/mcp/mcp.c` defines exactly **15** tools today: `index_repository`, `search_graph`, `query_graph`, `trace_path`, `get_code_snippet`, `get_graph_schema`, `get_architecture`, `search_code`, `list_projects`, `delete_project`, `index_status`, `check_index_coverage`, `detect_changes`, `manage_adr`, `ingest_traces`. I counted it on current `main`, including after a commit that touched `mcp.c` more recently than your PR — still 15. The root `README.md` already says 15, so the npm package README had simply drifted. Your change is correct on both lines.

**One thing is needed from you before it can land: a DCO sign-off.**

Every commit here needs a `Signed-off-by:` trailer matching the author, and commit `881eec73` does not have one. That is enforced by CI, so it would fail the moment the checks run. The fix:

```
git commit --amend -s
git push --force-with-lease
```

The sign-off email needs to match your commit author email.

**And one thing that is on us, not you.** The CI checks have not run on this PR at all yet — they are showing `action_required`, which means they are waiting on a maintainer to approve workflow runs for a first-time contributor. That is a GitHub default, not anything you did wrong, and it is why this PR has looked stuck with no checks rather than with failing ones. We will trigger that once the sign-off is in, so you get a clean run rather than a red DCO followed by a second round.

**A small heads-up, explicitly not something to fix here:** the same file also says "159 languages" while the root README currently says 158, and our own grammar manifest says 159. That inconsistency predates you and is ours to sort out — please leave it alone in this PR so the diff stays exactly what it claims to be.

Thanks for catching the drift. Small documentation corrections like this are genuinely useful, because the npm README is what most people actually read first.

DeusData:
Approved the workflow runs — CI is executing now. Apologies that this sat: your checks were never failing, they were waiting on a first-time-contributor approval, which is on us rather than on you.

**And you are right on the facts.** I counted the registry on current `main`: `TOOLS[]` in `src/mcp/mcp.c` has exactly 15 entries, and the one the README was missing is `check_index_coverage`. So both edits are correct.

**One thing will still be red, and it needs you:** the `dco` check. Your commit has no `Signed-off-by:` line, and that check is required. For a single commit, `git commit --amend --signoff` followed by a lease-guarded force push to your branch will do it. The sign-off has to use the same name and email as the commit author, which `--signoff` handles for you.

Once that lands green it can go in. Thank you for the fix — a stale number in a package README is exactly the kind of thing that quietly erodes trust in the rest of the document, and it is the sort of contribution that often goes unmade because it feels too small to bother with. It isn't.

DeusData:
Thank you — you were right, and the fix has shipped as #1765 with `Co-authored-by: nhtkylc91` on the commit.

I verified your number independently before taking it: the `TOOLS[]` table in `src/mcp/mcp.c` has exactly 15 entries, so `14 MCP tools` was genuinely stale in both places you changed.

Closing this rather than merging because it was half the problem. The same file also claimed `159 languages` where the root README says 158 — @ketpatil77 had independently caught that half in #967, and each stale figure appears **twice** (line 10 carries both). Since both PRs touched the same file, merging them in sequence would have needed a rebase and still left line 10 wrong, so they're combined into one complete fix crediting both of you.

Worth noting for anyone who looks at this next: there are 159 grammar directories in the tree, which doesn't reconcile trivially with 158 either — some languages share a grammar and some enums have no grammar. I aligned the package README to the root README rather than re-deriving the number, and flagged the discrepancy on the new PR rather than quietly picking a figure.

Thanks for spotting it. Stale counts are the kind of thing everyone reads past.

PR #1305, opened 2026-07-27
fix(extract): model Nix attrpaths (leaf name, scoped QN) and mint Variables for module-level bindings

sini:
## What does this PR do?

Models Nix attrpaths, and mints Variable nodes for Nix module-level bindings.

**Built on top of #1304.** That PR makes definitions reachable in function-headed files at all; this
one fixes how they are *named* once reachable, and adds the Variable half. The two are separate
concerns and the four commits here are extracted cleanly for independent review — but the branch
carries #1304's two commits as its base, so please merge that first. The one hard dependency is
`pipeline_nix_scoped_binding_calls_resolve`, whose fixture is function-headed and mints nothing
without #1304.

Filing under the CONTRIBUTING carve-out for focused bug fixes, as with #1304. Commits 3–4 (the
Variable half) add a node class rather than repair one, so if you would rather see that split out or
discussed in an issue first, say the word and I will pull it.

### 1. Attrpath names — a merge bug and an encoding bug

A Nix binding's name is a PATH. The resolver took only its first segment
(`child_by_field_name(attrpath, "attr")`), and three defects followed from that one modelling gap:

| source | before | after |
| --- | --- | --- |
| `setA = { dup = …; }` | name `dup`, QN `proj.f.dup` | QN `proj.f.setA.dup` |
| `setB = { dup = …; }` | **same QN — collides, def dropped** | QN `proj.f.setB.dup` |
| `a.b.fn = …` | name `a` | name `fn`, QN `proj.f.a.b.fn` |
| `"kebab-case" = …` | name `"kebab-case"` (quotes included) | name `kebab-case` |
| `"${dyn}" = …` | name `"${dyn}"` | not minted |

The collision is the serious one: two bindings sharing a leaf name in one file collapsed onto a
single node, and the second definition **and every CALLS edge sourced from it** were silently
discarded at write. On a 320-file Nix repository a conservative scan found ≥54 files affected and
≥275 bindings collapsed.

The convention adopted is the one already used for C++ namespaces — `ns::serialize` is name
`serialize`, QN `proj.file.ns.serialize`. For Nix: name is the leaf segment, qualified_name is
enclosing scope plus leaf. A binding whose value is an attribute set becomes a scope, mirroring
`is_namespace_scope_kind`'s treatment of a namespace. `let` bindings deliberately do **not** scope —
they are lexical, and C++ does not qualify by block scope either.

The correctness argument is that Nix has two spellings of the same thing:

```nix
a.b.fn = z: …;                 # sugar for
a = { b = { fn = z: …; }; };   # this
```

Before this they produced different, both-wrong answers. `nix_dotted_attrpath_qualifies_like_nested`
pins that they now agree.

### 2. Variables

`nix_var_types` has declared `binding` since the language was added, but `extract_var_names` had no
Nix case and `extract_variables` walks only the file root's direct children — which for a
function-headed file is the lambda. So the Nix Variable count was unconditionally zero.

Scope follows the existing rule rather than a new one: `extract_variables` mints **file scope and
never locals**, exactly as a C++ `declaration` inside a function body is not a Variable. For Nix,
file scope is the `let` bindings and the returned attrset; a binding in a deeper attrset is not.

That bound is load-bearing. Every Nix binding's parent is a `binding_set` at any depth, so admitting
them all would mint a node per `enable = true` in a NixOS module's settings tree — the per-leaf flood
the Helm `values.yaml` special case in this same function already exists to avoid. Measured on a
499-file NixOS configuration: **226 Nix Variables, at most 9 in any one file.** No flood.

Lambda-valued and attrset-valued bindings are skipped — the first is already a Function, the second
is a scope — so nothing is double-counted.

### The constraint that shaped the implementation

Definition QNs and call-scope QNs are computed by **two different functions**: `compute_class_qn` in
extract_defs.c and a separate one in extract_unified.c. If they disagree by one segment, the CALLS
edge names a source node that was never minted and is dropped at write — no error, and every
extraction-level assertion still passes.

They are also asymmetric today: the def-side enclosing-QN gate is language-gated, the call-side is
not. So widening one without the other would have produced exactly that silent mismatch.

Rather than implement the rule twice, the Nix computation lives in `helpers.c` and both sides call
it, which makes that class of drift impossible rather than merely unlikely.
`pipeline_nix_scoped_binding_calls_resolve` guards it end to end through the store, because no
extraction-level test can see a dropped edge.

### Impact

Cumulative with #1304, same binary and `mode=full` throughout, `Function` / `Variable` / `CALLS`:

| repo | `main` | with #1304 | with this PR |
| --- | --- | --- | --- |
| library (76 .nix) | 36 / 0 / 24 | 390 / 0 / 287 | 463 / 132 / 292 |
| framework (320 .nix) | 48 / 0 / 56 | 1900 / 0 / 1443 | 2213 / 59 / 1461 |
| system config (499 .nix) | 293 / 0 / 204 | 654 / 0 / 325 | 684 / 226 / 327 |

Variable counts are Nix-sourced only; the system config also carries ~7.4k pre-existing JSON
Variables, unchanged by this.

The framework repository's Function count rising 1900 → 2213 with no new files is the collision fix
alone — 313 definitions that previously shared a qualified name with a sibling and were discarded.
That independently corroborates the ≥275 estimate from the static scan.

### Tests

Seven new cases. Extraction: leaf-name disambiguation across two attrsets; dotted-vs-nested QN
equality; quote stripping (asserting the quoted form is *absent*, not merely that the bare form is
present); interpolated attrpath minting nothing; module-level Variables including the attrpath QN;
nested bindings minting nothing; and the Function/Variable split in both directions.

Every absence assertion is paired with a positive one on the same predicate in the same result, so
none can pass by extracting nothing at all.

Pipeline: `pipeline_nix_scoped_binding_calls_resolve` covers both routes into a qualified name — an
enclosing attrset and a dotted attrpath — asserting each call still reaches its target through the
store.

Extraction and pipeline suites: **494 passed, 0 failed.**

## Checklist

- [x] Tests pass locally (`make -f Makefile.cbm test`) — same pre-existing LeakSanitizer caveat as
      #1304: `3679 byte(s) leaked in 5 allocation(s)` from the test harness, identical at `d587deab`
      untouched

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -31,6 +31,10 @@ enum {
 
     EXPORT_ANCESTOR_DEPTH = 4,
     FUNC_PARENT_CLIMB_LIMIT = 4, /* fun_expr -> term -> uni_term -> let_binding (Nickel) */
+    /* Nix header lambdas to descend before the file's body: `{ pkgs, ... }:` is one,
+     * the nixpkgs overlay `final: prev:` is two. Bounded so a pathological chain
+     * cannot spin. */
+    NIX_HEADER_HOP_MAX = 8,
     DECORATOR_SCAN_LIMIT = 3,
     C_RETURN_WALK_DEPTH = 5,
     VAR_RECURSION_LIMIT = 8,
@@ -188,6 +192,7 @@ static void extract_func_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec 
 static void extract_class_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec);
 static void walk_defs(CBMExtractCtx *ctx, TSNode root, const CBMLangSpec *spec, int depth_unused);
 static void extract_variables(CBMExtractCtx *ctx, TSNode root, const CBMLangSpec *spec);
+static void extract_var_names(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec);
 static void extract_class_variables(CBMExtractCtx *ctx, TSNode class_node, const CBMLangSpec *spec);
 static void extract_rust_impl(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec);
 static void extract_class_methods(CBMExtractCtx *ctx, TSNode class_node, const char *class_qn,
@@ -4751,7 +4756,12 @@ static void extract_elixir_call(CBMExtractCtx *ctx, TSNode node, const CBMLangSp
 // --- Variable extraction ---
 
 // Helper to push a Variable definition
-static void push_var_def(CBMExtractCtx *ctx, const char *name, TSNode node) {
+/* `qn_name` is the name as it should appear in the qualified name, which differs
+ * from `name` only where a language scopes a variable below the module — Nix,
+ * whose binding names are attrpaths (`a.b.c = …` is name `c`, QN suffix `a.b.c`).
+ * Pass NULL to use `name` for both. */
+static void push_var_def_qn(CBMExtractCtx *ctx, const char *name, const char *qn_name,
+                            TSNode node) {
     if (!name || !name[0] || strcmp(name, "_") == 0) {
         return;
     }
@@ -4761,14 +4771,18 @@ static void push_var_def(CBMExtractCtx *ctx, const char *name, TSNode node) {
     def.name = name;
     /* Java/Go: directory-based module (package), so a Go package-level var in
      * myapp/db/conn.go is proj.myapp.db.Var, matching its siblings. */
-    def.qualified_name =
-        cbm_fqn_compute_source_lang(a, ctx->project, ctx->rel_path, name, ctx->language);
+    def.qualified_name = cbm_fqn_compute_source_lang(a, ctx->project, ctx->rel_path,
+                                                     qn_name ? qn_name : name, ctx->language);
     def.label = "Variable";
     def.file_path = ctx->rel_path;
     def.start_line = ts_node_start_point(node).row + TS_LINE_OFFSET;
     def.end_line = ts_node_end_point(node).row + TS_LINE_OFFSET;
     def.is_exported = cbm_is_exported(name, ctx->language);
     cbm_defs_push(&ctx->result->defs, a, def);
+}
+
+static void push_var_def(CBMExtractCtx *ctx, const char *name, TSNode node) {
+    push_var_def_qn(ctx, name, NULL, node);
 }
 
 // Helper: extract name from a declarator chain (C/C++/ObjC)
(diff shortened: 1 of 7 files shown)


DeusData:
Thank you for the contribution and for separating the Nix attrpath and Variable work while clearly documenting its dependency on #1304. This is now triaged as a high-priority parsing-quality bug for `0.9.1-rc`. Our community PR queue is currently quite full, so it may take a little time before we can complete the review and, if approved, merge it. We are doing our best to support community contributions and will return with code-grounded feedback as capacity opens.

DeusData:
I merged #1304 a moment ago, which means this branch now conflicts — it carried those two commits, and they have since landed on `main` through a merge commit. A rebase onto current `main` should drop them cleanly and leave just the attrpath and Variable work.

The review itself is done and positive. What I checked, since the whole design rests on it: definition QNs and call-scope QNs both route through the same shared helpers (`cbm_nix_binding_scope_qn`, `cbm_nix_qn_name`, `cbm_nix_attrpath_scope`, with the quote-strip inside `cbm_func_name_node_text`), and I traced both the enclosing-attrset route and the dotted-attrpath route end to end — they compose to identical QNs. That matters because before this PR no Nix scope was pushed on the call side at all, so the definition and call gates genuinely move together here rather than drifting apart later.

A few things I appreciated specifically: the quote test asserts the quoted form is **absent** rather than merely checking the bare form is present; every absence assertion is paired with a positive control, so none of them can pass vacuously; and `pipeline_nix_scoped_binding_calls_resolve` is the only test class that could catch a silently dropped CALLS edge, which is exactly the failure this design risks.

On the Variable half you offered to split: I am keeping it. `nix_var_types` has declared `binding` since the language landed, so this completes intended-but-dead functionality rather than adding a foreign concept, and the file-scope-only rule plus your measurements (226 variables across a 499-file config, at most 9 per file) show it will not flood the graph.

Two asymmetries worth recording for whoever reads this later, neither blocking: `deep.nested.x = 2` at top level mints a Variable while the nested-attrset spelling does not — defensible under a lexical file-scope rule, but surprising; and a `let … in let … in { … }` chain skips the second `let`. Both are fine as follow-ups.

Rebase and it goes in. Thank you for the care in both of these — self-reporting your own limitations is rarer than it should be.

sini:
Rebased onto `63a244a8`. Both #1304 commits (`30622fe1`, `ecde3188`) were dropped automatically as already-applied — no conflicts, nothing to resolve by hand. What remains is the five commits of attrpath and Variable work, and the diff is byte-identical to what you reviewed: 7 files, +636/−22.

On the two asymmetries you recorded: leaving both as-is for follow-up, as you called them. For the record on the first one — `deep.nested.x = 2` at top level minting a Variable while the nested-attrset spelling does not — that falls out of the file-scope rule being lexical: the dotted spelling is one binding whose leaf sits at file scope, whereas the nested spelling puts the leaf inside an attrset expression. Same value, different binder depth. I agree it is surprising, and it is the kind of thing that wants a stated rule rather than a fix.

One disclosure on my local run, since none of it is in the paths this PR touches. `extraction` (270/270), `pipeline` (227/227), and `grammar_labels` (2/2) are green. Two unrelated suites are red on my machine: `daemon_runtime` at 41 pass / 1 fail, and `cli` aborting under LeakSanitizer on a 26-byte `strdup` leak from `cli_activation_save_env` at `tests/test_cli.c:509` — a test-local helper, not production code. I have not run these against `main` to confirm they are pre-existing, so I am flagging rather than claiming.

DeusData:
Merged as `8f06867d7`. The rebase came through exactly as you described — 5 commits, 7 files, +636/−22, byte-identical to what was reviewed, and 28/28 green. Thank you for the care in both of these.

**On the two red suites you flagged: neither is yours, and you were right to flag rather than claim.**

I checked `tests/test_cli.c` against `origin/main` at the commit you rebased onto — it is **byte-identical**, so nothing in this PR can have caused that leak. The mechanism is worth knowing: `cli_activation_save_env` (`tests/test_cli.c:506`) `strdup`s `HOME` and `CBM_CACHE_DIR`, and the matching `free` lives at the *end* of `cli_activation_restore_env` (`:526`). Every one of those tests runs a long stretch of assertions between the two calls, so **if any assertion fires early, the test aborts before the restore and the `strdup` leaks**. In other words the 26-byte leak is a *symptom of a failing assertion*, not the failure itself — LSan is reporting the consequence and the actual defect is whatever assert fired first. If you still have the output, the assert line above the leak report is the interesting one.

`daemon_runtime` at 41/1 is likewise nowhere near the Nix extraction paths. Both are ours to look at.

**And your explanation of the first asymmetry is better than my description of it.** "The dotted spelling is one binding whose leaf sits at file scope, whereas the nested spelling puts the leaf inside an attrset expression — same value, different binder depth" is exactly right, and I agree with your conclusion: this wants a **stated rule** rather than a fix, because the alternative is a scope model that is no longer lexical. I have recorded it that way.

Thank you again — self-reporting your own red suites, unprompted and without hedging, is rarer than it should be.

PR #1308, opened 2026-07-27
fix(extract): converge Function/Method is_test with the tests/ path filter

### LINKED ISSUE #1294: Function.is_test is false for functions under tests/ — query_graph cannot exclude tests
### Summary

The stored `is_test` node property disagrees with the tool-level test filter. The property reports `false` for a function that lives in `tests/`, while `trace_path` classifies the same node as a test. Anyone filtering with `WHERE f.is_test = false` in `query_graph` gets tests back.

### Repro

```
query_graph --query "MATCH (f:Function {name:'pipeline_create_free'}) RETURN f.file_path, f.is_test"
-> ["tests/test_pipeline.c", "false"]
```

The same node is treated as a test by `trace_path`: inbound callers of `cbm_pipeline_new` go from 83 with `include_tests=true` to 3 with the default `false`, and this node is among those excluded.

So the tool-level filter is correct and the stored property is wrong.

### Secondary: the path filter misses `tests/repro/`

With the default `include_tests=false`, `rpd_index_and_fingerprint` (in `tests/repro/`) still appears in results. The filter appears to match `tests/test_*` rather than the whole `tests/` tree.

### Impact

- `query_graph` cannot exclude tests at all — the documented property is unusable as a filter.
- The `search_graph(max_degree=0)` dead-code candidate list is polluted with test files; `GraphTab.deadcode.test.tsx` and `GraphTab.filters.test.tsx` appear at the top.

---

**Environment**

- Version: codebase-memory-mcp 0.9.0 (release binary, darwin-arm64)
- OS: macOS 26.5.2, Apple Silicon
- Repo under test: DeusData/codebase-memory-mcp @ af3ffbd2 (21,143 nodes / 123,103 edges)
- Verified identically through both the MCP server and `cli` mode (outputs byte-identical)

Yyunozor:
Closes #1294.

## Summary
- `Function`/`Method.is_test` only ever became `true` via the Rust `#[test]` attribute check; the per-file test-file flag (`cbm_is_test_file`) was already computed and stored on the `Module` node, but never propagated to the definitions inside it. A C function in `tests/helpers/fixtures.c` stayed `is_test=false` and leaked past `store.c`'s `is_test != 1` filters, even though `trace_path`'s own independent path check already treated the file as a test.
- `extract_func_def` and `push_method_def` now OR `ctx->result->is_test_file` into `def.is_test`, additive to the existing Rust attribute check (`#[test]` functions embedded in an otherwise regular `.rs` file are unaffected).
- `cbm_is_test_file` (extraction-time) gained the same `tests/`/`test/`/`spec/`/`__tests__/` directory-prefix convention the project's own `cbm_is_test_path` (`pass_tests.c`) already uses, so a file under `tests/` with no `test_`/`_test` naming is caught too — applying an existing convention where it was missing, not inventing a new one.
- Secondary bug: `trace_call_path`'s own `include_tests` filter (`mcp.c`'s separate, query-time `is_test_file()`) matched a nested `.../tests/...` path but not a project-root-relative one, so `tests/repro/rpd_index_and_fingerprint` leaked into results under the default `include_tests=false`. Added the same missing directory-prefix check there.

## Verification
- `make -f Makefile.cbm test -j12`: parent 6786 passed / 1 failed / 4 skipped; fix branch 6789 passed / 1 failed / 4 skipped. Delta is exactly the 3 new tests; the 1 pre-existing failure (`test_cli.c:8526`, Copilot CLI/VS Code detection) and 4 skips (Windows-only) are identical on both sides.
- New tests: `extract_c_test_dir_marks_is_test_issue1294`, `extract_python_method_test_dir_marks_is_test_issue1294` (positive/negative for both extraction call sites) and `tool_trace_totals_respect_test_filter_tests_root_subtree_issue1294` (end-to-end `trace_call_path` `include_tests` toggle).
- `clang-format`/`clang-tidy`/`cppcheck` are not available in my local environment (`clang-format` errors on a `libz3` version mismatch in my Homebrew LLVM), so CI is the authority for lint, as in #857. The build itself uses `-Wall -Wextra -Werror` and compiled clean on both branches.

## Known limitations
- Path matching (here and in the pre-existing `cbm_is_test_path`) is forward-slash-only and case-sensitive; unchanged, to keep this diff minimal.
- `store.c`'s `cbm_is_test_file_path()` (a third, broader `strstr(fp,"test")` implementation feeding `search_graph`/dead-code) is untouched — outside this issue's reported scope.
- A natural follow-up: unify the four divergent is-test-path implementations (`pass_tests.c`, `helpers.c`, `mcp.c`, `store.c`) behind one shared function instead of a third copy of the same directory list — left for a separate PR.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -3461,6 +3461,14 @@ static void extract_func_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec 
         def.is_test = true;
     }
 
+    // A function/method defined in a file cbm treats as a test file is itself
+    // a test, regardless of its own name (tests/helpers/fixtures.c has none
+    // of the test_/_test naming conventions, but every function in it is
+    // still test code) (#1294). OR'd so the Rust attribute check above, which
+    // additionally catches #[test] functions embedded in an otherwise regular
+    // .rs file, is never downgraded by this.
+    def.is_test = def.is_test || ctx->result->is_test_file;
+
     // Docstring
     def.docstring = extract_docstring(a, node, ctx->source, ctx->language);
 
@@ -4424,6 +4432,10 @@ static void push_method_def(CBMExtractCtx *ctx, TSNode child, TSNode class_node,
 
     // MinHash fingerprint
     compute_fingerprint(ctx, &def, child);
+
+    // A method on a class defined in a test file (e.g. a JUnit/pytest
+    // TestFoo.test_bar) is itself a test, same as free functions (#1294).
+    def.is_test = def.is_test || ctx->result->is_test_file;
 
     cbm_defs_push(&ctx->result->defs, a, def);
 }
--- a/internal/cbm/helpers.c
+++ b/internal/cbm/helpers.c
@@ -295,6 +295,19 @@ bool cbm_is_test_file(const char *rel_path, CBMLanguage lang) {
     }
     const char *base = path_basename(rel_path);
 
+    /* Directory-based, language-agnostic: a file under a conventional test
+     * directory is a test file regardless of its own basename (e.g.
+     * tests/helpers/fixtures.c, which no per-language suffix/prefix rule
+     * below would otherwise catch). Mirrors the directory set cbm_is_test_path
+     * (src/pipeline/pass_tests.c) already uses for TESTS-edge detection, so
+     * the extraction-time is_test_file flag agrees with it (#1294). */
+    if (strstr(rel_path, "__tests__/") || strstr(rel_path, "/tests/") ||
+        strstr(rel_path, "/test/") || strstr(rel_path, "/spec/") ||
+        has_prefix(rel_path, "tests/") || has_prefix(rel_path, "test/") ||
+        has_prefix(rel_path, "spec/") || has_prefix(rel_path, "__tests__/")) {
+        return true;
+    }
+
     switch (lang) {
     case CBM_LANG_GO:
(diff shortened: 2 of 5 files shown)


DeusData:
Thank you for the contribution and for tracing the inconsistent test classification through extraction and MCP filtering. This is now triaged as a high-priority parsing-quality bug for `0.9.1-rc`. Our community PR queue is currently quite full, so it may take a little time before we can complete the review and, if approved, merge it. We are doing our best to support community contributions and will return with code-grounded feedback as capacity opens.

DeusData:

The failure is `FAIL: invalid Windows PATH smoke seam fell back to the live registry` — a main-side smoke seam issue, fixed on `main` by #1346 after your run. Your diff is not implicated.

One thing worth knowing, because it will save you a wasted click: **"Re-run failed jobs" will not clear it.** A re-run re-tests the same recorded merge commit and fails identically. Only a fresh push refreshes the merge ref — so a rebase on current `main` and a push is what picks up the fix.

Apologies for the delay. We are working through a large review backlog oldest-first; the queue is real rather than a judgement on this PR.

DeusData:
Merged (33a3f032) — thank you! Mirroring cbm_is_test_path instead of inventing a new convention, and shipping both positive and negative reproduce-first tests, made this a textbook review. The 4-way is-test-path duplication you flagged is a real follow-up; noted. Heads-up in the notes: graphs reindexed on the next release will intentionally reclassify defs under tests/ as tests.

PR #1310, opened 2026-07-28
fix(hooks): scope pre-commit clang-tidy to staged changes (#1264)

### LINKED ISSUE #1264: Local pre-commit hook (scripts/hooks/pre-commit -> make lint) blocks every commit: ~5,100 pre-existing clang-tidy findings across 113 files
## What happened

Running `git commit -s` on a fresh `main` checkout (no changes of my own in
the affected files) fails at the `scripts/hooks/pre-commit` hook, which
runs `make -j3 -f Makefile.cbm lint` unconditionally before every commit.
`lint-tidy` (clang-tidy, `WarningsAsErrors: '*'`) fails with **5,116
`error:`-level findings across 113 files** in `src/`, `internal/cbm/`, and
one vendored header pulled in via `src/foundation/`. None of the affected
files were touched by my change (PR #1222, `pass_pkgmap.c` /
`tests/test_pipeline.c` only) -- this reproduces on a clean checkout.

This means the pre-commit hook, as currently configured, blocks **every**
commit for **every** contributor who has clang-tidy on `PATH` (it fails
silent-success only when clang-tidy is absent, since `CLANG_TIDY ?=
$(shell [ -x "$(LLVM_BIN)/clang-tidy" ] && ... || echo clang-tidy)` falls
back to a bare `clang-tidy` that then errors "command not found" -- which
`make -j3 lint` may or may not propagate depending on shell, so this can
silently pass for contributors without it installed and silently block
everyone else).

PR #1222's own verification section already flagged the milder version of
this: "Full local `make lint` (clang-tidy included) surfaces pre-existing
findings in unrelated files (`internal/cbm/cbm.c`, `internal/cbm/extract_calls.c`,
etc.) that are present on a clean `main` checkout too... This matches
`_lint.yml`'s own comment that clang-tidy is 'enforced locally,' not in
CI." This issue is the follow-up: the "local-only, not CI-enforced" gap
between `scripts/lint.sh --ci` (cppcheck + clang-format + no-suppress,
CI-enforced, passes cleanly) and the full `make lint` (clang-tidy included,
hook-enforced, does not pass) is large enough that the hook itself is not
usable as shipped.

## Repro

```
git config core.hooksPath scripts/hooks   # per CONTRIBUTING.md
make -j3 -f Makefile.cbm lint-tidy
```

## Environment

- `clang-tidy --version`: Homebrew LLVM 22.1.8 (`/opt/homebrew/opt/llvm/bin/clang-tidy`)
- macOS arm64, Xcode Command Line Tools SDK (MacOSX26.sdk)
- `.clang-tidy`: `Checks: -*, bugprone-*, cert-*, ... readability-*, clang-analyzer-*` with a short disabled-list, `WarningsAsErrors: '*'`, `HeaderFilterRegex: '^(src|tests)/.*\.h$'`

## Evidence

Total: 5,116 `error:` lines across 113 distinct project files. Breakdown by check (top entries):

| Check | Count |
|---|---|
| `readability-magic-numbers` | 3,932 |
| `misc-include-cleaner` | 380 |
| `readability-function-cognitive-complexity` | 270 |
| `readability-braces-around-statements` | 179 |
| `bugprone-implicit-widening-of-multiplication-result` | 58 |
| `readability-math-missing-parentheses` | 54 |
| `readability-avoid-nested-conditional-operator` | 32 |
| `readability-uppercase-literal-suffix` | 23 |
| `cert-err33-c` | 19 |
| `misc-no-recursion` | 15 |
| `readability-function-size` | 13 |
| `bugprone-branch-clone` | 12 |
| `misc-use-internal-linkage` | 11 |
| `clang-analyzer-optin.core.EnumCastOutOfRange` | 11 |

Representative examples (all pre-existing, none touched by PR #1222):

```
internal/cbm/cbm.c:22:1: error: included header mimalloc.h is not used directly [misc-include-cleaner,-warnings-as-errors]
internal/cbm/cbm.c:479:12: error: function 'count_params_from_signature' has cognitive complexity of 29 (threshold 25) [readability-function-cognitive-complexity,-warnings-as-errors]
internal/cbm/cbm.c:604:12: error: Potential leak of memory pointed to by 'key' [clang-analyzer-unix.Malloc,-warnings-as-errors]
src/ui/httpd.h:54:17: error: 16 is a magic number; consider replacing it with a named constant [readability-magic-numbers,-warnings-as-errors]
src/pipeline/pass_lsp_cross.h:154:6: error: function 'cbm_pxc_run_one' has a definition with different parameter names [readability-inconsistent-declaration-parameter-name,-warnings-as-errors]
```

`readability-magic-numbers` alone (3,932 of 5,116) suggests either a
`.clang-tidy` config drift (this repo's own `SKIP_ONE`/`PAIR_LEN`-style
named-constant convention, visible throughout `pass_pkgmap.c`, was clearly
meant to satisfy this check -- but it's still firing thousands of times
elsewhere) or a version-driven behavior change in how the check treats
struct field array sizes (e.g. `char method[16];` in `httpd.h`) versus
whatever clang-tidy version this hook was last verified against.

## Suggested remediation (any of)

1. Pin/document a known-good clang-tidy version the config was authored
   against, if this is a version-drift issue -- possibly resolves most of
   the 3,932 magic-number findings at once.
2. Scope the pre-commit hook's `lint-tidy` invocation to staged/changed
   files only (a `git diff --cached --name-only` filter), matching how a
   PR's own diff is what's actually being reviewed -- rather than
   re-linting the entire tree on every commit.
3. If the intent really is "every commit must leave the whole tree
   clang-tidy-clean," that's a large, separate cleanup effort (113 files) 
   that should probably be tracked and paid down deliberately rather than
   discovered by contributors as a blocked `git commit`.
4. At minimum, note in `CONTRIBUTING.md` that the hook is currently
   non-functional so contributors know `--no-verify` is expected until
   this is resolved, rather than each rediscovering 5,000+ unrelated
   errors per commit.

## Context

Found while addressing review feedback on PR #1222 (SwiftPM
`Package.swift` manifest resolution, item 1 of #551) -- my actual changed
file (`src/pipeline/pass_pkgmap.c`) has zero clang-tidy findings on the
lines I touched, confirmed by running clang-tidy scoped to just that file;
the hook only blocks because of these ~5,100 pre-existing findings
elsewhere.

Yyunozor:
Fixes #1264

## Problem

`scripts/hooks/pre-commit` runs `make -j3 -f Makefile.cbm lint`
unconditionally on every commit. `lint` includes `lint-tidy`, which runs
clang-tidy over the whole `LINT_SRCS` tree. On a clean `main` checkout
that surfaces ~5,100 pre-existing `error:`-level findings across 113
files (mostly `readability-magic-numbers`), unrelated to whatever the
commit touches. The hook blocks every commit for every contributor with
clang-tidy on `PATH` — reproduced on a clean checkout, one-line non-C
change.

## Fix

Add `lint-tidy-diff` (`scripts/lint-tidy-diff.sh`), a Makefile target
running the same clang-tidy binary/config, but passing clang-tidy's own
`-line-filter` so a diagnostic can only fire on lines the commit actually
added or modified (parsed from `git diff --cached -U0` hunk headers).
Pre-existing findings on untouched lines are never reported, whether in
an untouched file or an untouched line of a touched file — baseline-aware
at the line level, not just the file level.

The hook now runs `lint-ci` (unchanged: cppcheck + clang-format +
no-suppress, already CI's contract) plus `lint-tidy-diff` instead of the
full `lint` target. `make lint`, `make lint-tidy`, and `scripts/lint.sh`
are untouched, still the full-tree audit; CI (`_lint.yml` →
`scripts/lint.sh --ci`) never ran clang-tidy and is unaffected. This
follows the scoping the issue itself suggested (remediation option 2), at
line rather than file granularity.

## Verification

- Reproduced: clean checkout, `git commit -s` on a one-line
  `CONTRIBUTING.md` change, clang-tidy on `PATH` → blocked, 5,190
  pre-existing findings, commit did not happen.
- After the fix, same clang-tidy binary/config:
  - No staged `.c` changes → skipped, exit 0.
  - Clean addition to `internal/cbm/cbm.c` (one of the dirtiest files, 57
    pre-existing findings on a full-file run of the same edit) → exit 0,
    nothing reported.
  - A deliberate new finding (`return 424242;`) added to the same file →
    exit 2, reports only that one finding.
  - Header-only change (no paired `.c` change) → skipped, exit 0 (see
    Known limitation).
- `scripts/lint.sh` (cppcheck + clang-format + no-suppress) unaffected,
  still clean.

## Known limitation

clang-tidy reports a header's diagnostics under whatever path it was
`#include`-d with, not reliably its working-tree path (verified:
absolute vs. search-path-relative, depending on include style). So
`lint-tidy-diff` scopes `.c` files only; a header-only commit isn't
covered and needs `make lint-tidy`. A proper diff-to-header mapping would
be a natural follow-up, left for the maintainer to size.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -651,7 +651,7 @@ PP_OBJ_TSAN = $(BUILD_DIR)/tsan_preprocessor.o
 
 # ── Targets ──────────────────────────────────────────────────────
 
-.PHONY: test test-par test-repro test-foundation test-tsan test-daemon-smoke cbm cbm-with-ui frontend embed clean-c lint lint-tidy lint-cppcheck lint-format security
+.PHONY: test test-par test-repro test-foundation test-tsan test-daemon-smoke cbm cbm-with-ui frontend embed clean-c lint lint-tidy lint-tidy-diff lint-cppcheck lint-format security
 
 $(BUILD_DIR):
 	mkdir -p $(BUILD_DIR)
@@ -980,6 +980,12 @@ LINT_TEST_SRCS = $(ALL_TEST_SRCS)
 lint-tidy:
 	@echo "=== clang-tidy ==="
 	@$(CLANG_TIDY) --quiet $(LINT_SRCS) -- $(CFLAGS_COMMON) $(SYSROOT_FLAG)
+
+# clang-tidy, scoped to this commit's staged changes only (scripts/lint-tidy-diff.sh).
+# The pre-commit hook runs this instead of the full lint-tidy sweep above, so a
+# clean commit is not blocked by pre-existing findings elsewhere in the tree.
+lint-tidy-diff:
+	@scripts/lint-tidy-diff.sh "$(CLANG_TIDY)" $(LINT_SRCS) -- $(CFLAGS_COMMON) $(SYSROOT_FLAG)
 
 # cppcheck: complementary analysis (config in .cppcheck)
 lint-cppcheck:
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -7,7 +7,11 @@ set -euo pipefail
 #   git config core.hooksPath scripts/hooks
 
 echo "pre-commit: running all linters in parallel..."
-make -j3 -f Makefile.cbm lint
+# clang-tidy runs diff-scoped (lint-tidy-diff, see scripts/lint-tidy-diff.sh):
+# the full lint-tidy sweep surfaces thousands of pre-existing findings that
+# have nothing to do with this commit and would block every contributor who
+# has clang-tidy on PATH. `make lint-tidy` remains available for a full audit.
+make -j3 -f Makefile.cbm lint-ci lint-tidy-diff
 
 echo "pre-commit: security audit (source-level)..."
 scripts/security-audit.sh
--- a/scripts/lint-tidy-diff.sh
+++ b/scripts/lint-tidy-diff.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+# lint-tidy-diff.sh — clang-tidy scoped to the lines this commit changes.
+#
+# `make lint-tidy` (the full sweep) checks the entire LINT_SRCS tree, which
+# currently surfaces ~5,100 pre-existing findings unrelated to any given
+# commit (see Makefile.cbm's `lint-tidy` target). That is fine for a manual
+# deep audit but makes the pre-commit hook unusable: it blocks every commit,
+# including ones that touch files with zero findings on the touched lines.
+#
+# This script runs the same clang-tidy binary/config against the SAME staged
+# .c files, but passes clang-tidy's own `-line-filter` so only lines the
+# commit actually added/modified can produce a diagnostic. Pre-existing
+# findings on untouched lines — in the same file or elsewhere — are not
+# reported.
+#
+# Invoked by `make lint-tidy-diff` (scripts/hooks/pre-commit), which supplies
+# the resolved clang-tidy binary, then LINT_SRCS, then `--`, then the
+# compiler flags — the same argv shape as the `lint-tidy` target itself.
+#
+# Known limitation: only staged *.c files that are members of LINT_SRCS are
+# checked (matching the full sweep's own scope: vendored code and tests/ are
+# never clang-tidy'd). clang-tidy reports a header's diagnostics under
+# whatever path it was #include-d with, which is not reliably the same
+# string as its path in the working tree, so a commit that touches only a
+# header is not covered here even if a .c file including it also changed;
+# run `make lint-tidy` for a full pass in that case.
+
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+cd "$ROOT"
+
+if [ "$#" -lt 1 ]; then
+    echo "usage: lint-tidy-diff.sh <clang-tidy-bin> [LINT_SRCS...] -- [CFLAGS...]" >&2
+    exit 2
+fi
+
+CLANG_TIDY="$1"
+shift
+
+LINT_SRCS=()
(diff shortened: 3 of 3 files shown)


DeusData:
Thank you for the contribution and for reproducing how repository-wide clang-tidy findings block unrelated staged changes. This is now triaged as a normal-priority contributor-workflow bug for `0.9.1-rc`. Our community PR queue is currently quite full, so it may take a little time before we can complete the review and, if approved, merge it. We are doing our best to support community contributions and will return with code-grounded feedback as capacity opens.

DeusData:
A quick note so this does not sit here looking like your problem: **neither of your red checks is caused by your change.**

Your diff touches `Makefile.cbm` and two lint-hook scripts. The failures are:

1. the Windows PATH smoke seam (`FAIL: invalid Windows PATH smoke seam fell back to the live registry`) — a main-side issue fixed by #1346 after your run; and
2. an AddressSanitizer access-violation in `cbm_daemon_ipc_connection_interrupt` (`src/daemon/ipc.c:5608`) during `test_daemon_ipc_wait_forever_is_interruptible`, which crashed the suite at `pass=0`.

A lint-hook and Makefile change cannot cause a daemon-IPC ASan crash. That second one is a **production crash in our own code** and I have flagged it internally for its own investigation — thank you for surfacing it, entirely by accident.

So nothing is needed from you on the CI front; a rebase would pick up the smoke fix, but the ASan crash is ours. Your PR is queued for a proper content review.

Apologies for the delay — we are working through a large backlog oldest-first, and I did not want a crash of ours sitting on your PR looking like your bug.

DeusData:
Merged (0273250c) — thank you! Scoping the hook to the staged diff with -line-filter while keeping lint-ci's full gates untouched is exactly the right split, and documenting the header-only-commit blind spot honestly rather than papering over it is appreciated. #1264's daily paper-cut is gone.

PR #1312, opened 2026-07-28
test(daemon-ipc): 10s onset budget for the startup-lock observation waits

DeusData:
Release run 30322785509 failed on windows-latest x64 1/2: `test_daemon_ipc.c:1075 ASSERT(startup_observed)` — first occurrence in 6 runs. Attribution: the awaited state (startup thread holding `cbm-startup-v2.lock` while the test's reader lock blocks the rendezvous write) is **stable once reached**; only the onset is timing-dependent, and the 200ms poll budget missed thread-spawn + IPC preamble on a starved runner. Per the flaky-test ladder this is the legitimate budget-bump case (stable state, budget sole flake source): both sibling onset waits go 200ms → 10s, healthy runs still exit on first observation. 46/46 locally; 2× ARM64 VM verification in flight. Unblocks v0.9.1-rc.1.

### DIFF
--- a/tests/test_daemon_ipc.c
+++ b/tests/test_daemon_ipc.c
@@ -1048,7 +1048,13 @@ TEST(daemon_ipc_windows_startup_retries_transient_rendezvous_reader) {
     if (record_status == CBM_PRIVATE_FILE_LOCK_OK) {
         thread_started = cbm_thread_create(&thread, 0, ipc_test_win_startup_call, &call) == 0;
     }
-    for (size_t attempt = 0; thread_started && attempt < 200U; attempt++) {
+    /* Onset wait: once the startup thread reaches its startup lock it HOLDS it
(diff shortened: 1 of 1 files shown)


DeusData:
Closing unmerged: falsified by local verification before merge. With the 10s budget in place, 2× ARM64 VM runs still failed the sibling test (`daemon_ipc_windows_rendezvous_bridges_concurrent_lifetime_owner`, ASSERT(startup_observed)) on an idle 18-core machine — so the observed state is not stable-after-onset and no onset budget fixes it. The startup path (`acquire → prepare_handoff → release`) can collapse the busy window to microseconds when the rendezvous-write retry resolves fast, making the lock-busy probe another transient-window sample. Correct fix needs the deterministic route (production seam à la the application_job_thread held-gate, or observing a persistent side-effect instead of live lock state) — tracked for the follow-up, branch kept.

PR #1313, opened 2026-07-28
fix(graph-ui): rank Accept-Language by q instead of substring match

lukiod:
## What does this PR do?

`detectLanguage` tested whether `"zh"` appeared anywhere in the header, so any visitor whose `Accept-Language` mentions Chinese got the Chinese UI regardless of preference.

| Header | Before | Expected |
|---|---|---|
| `en-US,en;q=0.9,zh;q=0.5` | `zh` | `en` |
| `zh;q=0, en` | `zh` | `en` |

The first is the common case: a bilingual visitor who prefers English and lists Chinese as a fallback. That is a normal Chrome language configuration, and it currently produces a Chinese interface for someone who asked for English.

The second sets `q=0`, which RFC 7231 defines as *not acceptable* — the header explicitly excludes Chinese and it was served anyway.

Now parses `q`, drops `q=0`, sorts descending, and matches the primary subtag rather than a substring. Chinese still wins whenever it is actually preferred, and the two existing assertions are unchanged and still pass.

Reproduce-first: both rows above are in the new tests and fail against the previous implementation.

## Checklist

The two unticked items are the C toolchain, which this change does not touch and which I have not set up. For `graph-ui` I ran its own suite:

```
npx vitest run   9 files, 38 tests passed
npx tsc -b       clean
```

Happy to run the C targets too if you would like that before merging.

### DIFF
--- a/graph-ui/src/lib/i18n.test.ts
+++ b/graph-ui/src/lib/i18n.test.ts
@@ -7,6 +7,31 @@ describe("i18n", () => {
     expect(detectLanguage("de-DE,de;q=0.9")).toBe("en");
   });
 
+  it("ranks Accept-Language by q rather than by presence", () => {
+    // A bilingual visitor who prefers English but lists Chinese as a fallback.
+    expect(detectLanguage("en-US,en;q=0.9,zh;q=0.5")).toBe("en");
+    expect(detectLanguage("en,zh;q=0.1")).toBe("en");
+    // Chinese still wins when it is actually preferred.
+    expect(detectLanguage("zh;q=0.9,en;q=0.5")).toBe("zh");
+  });
+
+  it("treats q=0 as unacceptable", () => {
+    expect(detectLanguage("zh;q=0, en")).toBe("en");
+    expect(detectLanguage("zh;q=0")).toBe("en");
+  });
+
+  it("matches the language subtag, not a substring", () => {
+    expect(detectLanguage("en-GB")).toBe("en");
+    expect(detectLanguage("zh-Hant-TW")).toBe("zh");
+  });
+
+  it("honours the override and handles empty input", () => {
+    expect(detectLanguage("zh-CN", "en")).toBe("en");
+    expect(detectLanguage("en-US", "zh")).toBe("zh");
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for the contribution and for covering both preference ordering and `q=0` in the graph UI language negotiation. This is now triaged as a normal-priority UX bug for `0.9.1-rc`. Our community PR queue is currently quite full, so it may take a little time before we can complete the review and, if approved, merge it. We are doing our best to support community contributions and will return with code-grounded feedback as capacity opens.

DeusData:
Merged — thank you, this is a textbook contribution.

The old code substring-matched the language tag anywhere in the header, which meant `Accept-Language` was being read as "does this string contain `zh`" rather than as a ranked preference list. So a browser sending `en-US,en;q=0.9,zh;q=0.5` — English strongly preferred, Chinese as a distant fallback — could land on Chinese. Your version parses the q-values, sorts by them, and matches on the primary subtag, which is what RFC 7231 actually specifies.

The detail I appreciated most is that you handled **`q=0`**. That is the half people almost always miss: `q=0` does not mean "lowest priority", it means "not acceptable", and a language marked that way must be excluded outright rather than merely ranked last. You got that right and wrote a test named for it.

Because this parses an attacker-controllable request header, it was reviewed on that footing too: the parse is linear with no backtracking blowup, there is no unbounded accumulation on a hostile header, and nothing from the header reaches the DOM.

One thing worth telling you, since it is our gap and not yours: **CI does not currently execute the graph-ui vitest suite** — only the npm package tests run. So your tests were not actually being exercised by the green checks on this PR. I ran them locally before merging: 38 passed across 9 files, and I then reverted just `i18n.ts` and re-ran to confirm the tests bind — exactly your two new cases fail without the production change, and nothing else does. That is precisely the property a regression test should have.

You were also upfront that you ran vitest and tsc but not the C suite, which was the right call since you touched no C. Being explicit about what you did and did not verify makes a reviewer's job much easier — thank you for that.

Wiring vitest into CI is on us as a follow-up.

PR #1323, opened 2026-07-28
fix(cypher): scan all unlabeled query candidates

### LINKED ISSUE #1196: query_graph: patterns with no label on the source node silently return empty — same DB and binary previously returned correct rows (v0.9.0)
**Environment:** v0.9.0, Windows amd64 (ui variant), same index DB as #1110 / #1132 / #1133 (10 projects; main graph 31k nodes / 176k edges, 43,851 CALLS edges per `get_graph_schema`). The CLI and the MCP server return identical results. First mentioned in https://github.com/DeusData/codebase-memory-mcp/issues/1110#issuecomment-5034538279 — filing it on its own as it's unrelated to the (invalid) `direction` flag discussed there.

**Summary:** On 2026-07-15 this query returned 2 correct rows (documented verbatim in #1110's body):

```
MATCH (a)-[r:CALLS]->(b) WHERE b.name = 'auto_match' RETURN a.qualified_name   → 2 rows (correct)
```

Today, same binary (SHA-pinned) and same project, it returns `{"rows":[],"total":0}` with the generic no-results hint. Probe matrix:

| Query shape | Result |
|---|---|
| `MATCH (n) RETURN n.name` | ✅ rows |
| `MATCH (n) WHERE n.name = '<existing>' RETURN ...` | ❌ silent empty |
| `MATCH (n) WHERE n.lines > 100 RETURN ...` | ❌ silent empty |
| `MATCH (n:Method) WHERE n.name = '<same>' RETURN ...` | ✅ 1 row |
| `MATCH (a)-[r]->(b) RETURN a.name` (untyped edge) | ✅ rows |
| `MATCH (a)-[r:CALLS]->(b) RETURN a.name` | ❌ silent empty |
| `MATCH (a:Function)-[r:CALLS]->(b) RETURN a.name, b.name` | ✅ rows |
| `MATCH (a:Function)-[r:CALLS]->(b) WHERE b.name = 'X' RETURN ...` | ✅ correct callers (verified vs grep) |
| `MATCH (a)-[r:CALLS]->(b:Method) WHERE b.name = 'X' RETURN ...` | ❌ silent empty (label on the TARGET doesn't help) |

Reproduced identically on a second project in the same DB (13.5k nodes), for string and numeric predicates, via both the CLI and the MCP server.

**What changed between "worked" (2026-07-15) and "broken" (2026-07-21):** not the binary. The DB went through several `delete_project` + fresh-index cycles (projects added and removed on 2026-07-16) plus incremental auto-watch updates. It looks like whatever powers unlabeled scans / typed-edge anchor selection consults per-label state that goes stale after those cycles, while label-anchored plans still read the live data.

**Impact:** the most natural Cypher an LLM agent emits (`MATCH (a)-[:CALLS]->(b) WHERE b.name = ...`) now silently reads as "no callers" — the same confidently-wrong-empty failure family as #1111 / #1133. Workaround: always put a label on the pattern's source node.

**Asks:**
1. Make unlabeled matching consistent with label-anchored matching after delete/reindex cycles (or rebuild the relevant index on write).
2. Until then, fail explicit: if the planner resolves a zero-cardinality anchor while the schema reports non-empty labels/edge types, return an error or a diagnostic instead of a clean empty set.

Happy to run a diagnostic build against this DB — it's private, but I can run anything read-only and share the output.

Enferlain:
## What does this PR do?

Fixes #1196.

Unlabeled Cypher queries previously capped candidate enumeration at the requested result limit before applying `WHERE`. A valid match beyond that early candidate window could therefore disappear.

This change materializes the complete relevant candidate set with an unbounded store query, applies filtering, and only then enforces the final query limit. It includes a reproduce-first regression test with the matching row beyond the result-limit boundary.

Issue #1364 independently exercises the same early unlabeled-scan cap with `RETURN DISTINCT type(r)`. Its separately observed cap above roughly 100k raw projected rows occurs later, before `DISTINCT`, and is not claimed as part of this fix.

## Verification

- Rebasing over current `main` preserves the checked `cross_join_nodes` failure path introduced by the newer bounds work.
- `nix develop --command make -j8 -f Makefile.cbm test-focused TEST_SUITES=cypher`
- Result: 170 passed, including `cypher_exec_unlabeled_where_beyond_result_limit_issue1196` and `cypher_cross_join_alloc_rejects_overflow`.

## Checklist

- [ ] Full test suite passes locally (`make -f Makefile.cbm test`) — focused sanitized Cypher suite passed
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) — not rerun for this rebase-only update
- [x] New behavior is covered by a regression test that would fail without this fix

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -3024,31 +3024,14 @@ static void scan_alternation_labels(cbm_store_t *store, const char *project, con
     free(copy);
 }
 
-static void scan_pattern_nodes(cbm_store_t *store, const char *project, int max_rows,
-                               cbm_node_pattern_t *first, cbm_node_t **out_nodes, int *out_count) {
+static void scan_pattern_nodes(cbm_store_t *store, const char *project, cbm_node_pattern_t *first,
+                               cbm_node_t **out_nodes, int *out_count) {
     if (first->label && strchr(first->label, '|')) {
         scan_alternation_labels(store, project, first->label, out_nodes, out_count);
     } else if (first->label) {
         cbm_store_find_nodes_by_label(store, project, first->label, out_nodes, out_count);
     } else {
-        cbm_search_params_t params = {.project = project,
-                                      .min_degree = CYP_FOUND_NONE,
-                                      .max_degree = CYP_FOUND_NONE,
-                                      .limit = max_rows * CYP_GROWTH_10};
-        cbm_search_output_t sout = {0};
-        cbm_store_search(store, &params, &sout);
-        *out_count = sout.count;
-        *out_nodes = malloc(sout.count * sizeof(cbm_node_t));
-        for (int i = 0; i < sout.count; i++) {
-            (*out_nodes)[i] = sout.results[i].node;
-            sout.results[i].node.name = NULL;
-            sout.results[i].node.project = NULL;
-            sout.results[i].node.label = NULL;
-            sout.results[i].node.qualified_name = NULL;
-            sout.results[i].node.file_path = NULL;
-            sout.results[i].node.properties_json = NULL;
-        }
-        cbm_store_search_free(&sout);
+        cbm_store_find_nodes(store, project, out_nodes, out_count);
     }
     /* Apply inline property filters — free rejected nodes' strings */
     if (first->prop_count > 0) {
@@ -4656,7 +4639,7 @@ static int expand_additional_patterns(cbm_store_t *store, cbm_query_t *q, const 
 
         cbm_node_t *extra_nodes = NULL;
         int extra_count = 0;
-        scan_pattern_nodes(store, project, max_rows, &patn->nodes[0], &extra_nodes, &extra_count);
+        scan_pattern_nodes(store, project, &patn->nodes[0], &extra_nodes, &extra_count);
         int rc = 0;
         if (patn->rel_count == 0) {
             rc = cross_join_nodes(bindings, bind_count, extra_nodes, extra_count, nvar, opt);
@@ -4708,7 +4691,7 @@ static int execute_single(cbm_store_t *store, cbm_query_t *q, const char *projec
     /* Step 1: Scan initial nodes */
     cbm_node_t *scanned = NULL;
     int scan_count = 0;
-    scan_pattern_nodes(store, project, max_rows, &pat0->nodes[0], &scanned, &scan_count);
+    scan_pattern_nodes(store, project, &pat0->nodes[0], &scanned, &scan_count);
 
     /* Build initial bindings with early WHERE */
     int bind_cap = scan_count > max_rows ? scan_count : (max_rows > 0 ? max_rows : SKIP_ONE);
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -115,6 +115,7 @@ struct cbm_store {
     sqlite3_stmt *stmt_find_node_by_id;
     sqlite3_stmt *stmt_find_node_by_qn;
     sqlite3_stmt *stmt_find_node_by_qn_any; /* QN lookup without project filter */
+    sqlite3_stmt *stmt_find_nodes;
     sqlite3_stmt *stmt_find_nodes_by_name;
     sqlite3_stmt *stmt_find_nodes_by_name_any; /* name lookup without project filter */
     sqlite3_stmt *stmt_find_nodes_by_label;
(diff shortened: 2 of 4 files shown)


Enferlain:
These are just how I fixed them in my local build until solved on upstream, and since it worked in tests and were built according to the contributing guidelines, gpt said they could be sent as prs.

They can be used as reference/alternative fixes if it can't be merged to hopefully speed up the process. Applies to the subsequent 3 other prs as well.

DeusData:
Reviewed — **the root cause is right and this is the correct fix layer.** Two things stand between it and a merge, and neither of your failing checks is one of them.

**The bug is exactly as you diagnosed.** Unlabeled Cypher scans fetched at most `max_rows * 10` candidates through `cbm_store_search` **before** WHERE was applied, so a matching row beyond that window silently vanished. That is #1196's "confidently wrong empty" — `MATCH (n) WHERE n.name = 'existing'` returning nothing while the labelled form finds it, and drifting after delete-and-reindex cycles because the search ordering changes. Filtering before limiting, the way labelled scans already do, is the right shape.

I also liked that the replacement is simpler than what it removes: the old shallow-copy-and-NULL-out dance became one allocator with a clear caller-frees contract. In an agent-reachable engine that is a memory-safety improvement in its own right, and the new SQL is a constant string with bound parameters.

Your test is properly binding — 12 nodes, `max_rows=1` so the old window was 10, match inserted last, asserting both the row count and the row value. Red on main.

**Neither failing check is yours.** `security / codeql-gate` polled for its full budget and gave up with `BLOCKED: CodeQL timeout`, while the actual `analyze` job **passed** on the same PR — the gate's timeout races the analyze job nondeterministically, and it passed in 5m22s on your sibling #1326 with no relevant difference. `ci-ok` is just the aggregator. Every real leg is green: all three smoke platforms, all unix/windows/tsan legs including macos-15-intel, lint, security-static and license-gate.

**What is needed from you: a rebase.** Your base predates #1173 and #1176, both merged on 30 July and both touching `src/cypher/cypher.c` and `store.c`. Those are rebase-class conflicts rather than anything you did. One thing to re-verify afterwards: #1176 added a bounds-check with error propagation in `cross_join_nodes`, and your change enlarges the candidate sets feeding it — worth confirming the two compose as expected.

**And one question that is the maintainer's, not yours.** An unlabeled `MATCH (n) WHERE …` now materialises every node in the project. At our top scale (~8.5M nodes) that is multi-GB for a single query, and a multi-pattern unlabeled query like `MATCH (a), (b)` is N×N in the binding allocation.

I want to be fair about this: **the exposure is not new.** `cbm_store_find_nodes_by_label` is already unbounded, so `MATCH (a:Function), (b:Function)` does the same thing on a large repo today, and #1176's freshly-merged bounds check means it now fails cleanly rather than overflowing. Your change makes unlabeled *consistent* with labelled, which is defensible. But full-scan semantics for unlabeled queries is a policy the maintainer should accept explicitly rather than inherit, and #1196 itself asked for a "fail explicit" option. I have raised it — it should not block the correctness fix.

One small thing: the PR body says the scan goes "through the store visitor", but the implementation is a new materialising store query. Reads like drift from an earlier iteration — worth a one-line correction so the next reviewer is not looking for a visitor.

Rebase and I will pick this straight back up.

Enferlain:
Addressed the follow-up:

- rebased onto current `main` (including the newer cross-join bounds work),
- resolved the overlap by retaining the unbounded unlabeled scan while preserving checked `cross_join_nodes` error propagation,
- reran the complete sanitized Cypher suite: **170 passed**, including both `cypher_exec_unlabeled_where_beyond_result_limit_issue1196` and `cypher_cross_join_alloc_rejects_overflow`, and
- corrected the PR description: this implementation materializes an unbounded store query; it does not use a store visitor.

#1364 independently confirms this PR's early unlabeled-candidate cap. I documented its later roughly-100k pre-`DISTINCT` projection cap as separate scope rather than silently broadening this fix.

DeusData:
Merged (22d93dd1) — thank you! Removing the 10×max_rows pre-filter cap and materializing the full candidate set makes unlabeled matches honest again, consistent with the already-unbounded labeled path; memory is bounded by the same store the labeled scan uses and the #601 deadline still governs time. One transparency note: the original reporter's fresh v0.10.6 retest on #1196 shows a SECOND mechanism (aggregates computed over a max_rows-bounded expansion, hitting labeled sources too), so the issue is reopened for that half — your fixture and fix stand as the complete answer to the scan half.

PR #1324, opened 2026-07-28
fix(pipeline): suppress weak Python member calls

### LINKED ISSUE #1276: Python unresolved member calls produce false-positive CALLS edges via suffix_match and field_type_hint
### Version

`codebase-memory-mcp dev`

### Platform

Linux (x64)

### Install channel

Built from source

### Binary variant

standard

When Python receiver/type resolution does not identify a member-call target, the indexing pipeline falls back to name-based heuristics and emits an ordinary `CALLS` edge to an unrelated same-named method.

This produces structurally impossible edges, including production code calling test mocks and external-library calls being connected to unrelated project methods.

Representative results from one production function:

| Actual call | Stored target | Strategy / confidence |
|---|---|---|
| `accelerator.print(...)` | `tests.unit.test_resume_logic.MockAccelerator.print` | `field_type_hint`, `0.85`, 2 candidates |
| `accelerator.backward(...)` | `library.models.sdxl.unet.FlashAttentionFunction.backward` | `suffix_match`, `0.28`, 2 candidates |
| `trainer.lr_scheduler.step()` | `library.pipelines.gradual_latent.EulerAncestralDiscreteSchedulerGL.step` | `suffix_match`, `0.01`, 84 candidates |

The correct target for the first two calls is an external `accelerate.Accelerator` method. For the scheduler call, the concrete runtime scheduler cannot be established from this call site. None of these calls should be represented as a certain call to the selected target above.

I expected receiver-aware resolution when sufficient type evidence exists, and conservative behavior otherwise: either preserve the call as unresolved/ambiguous or omit the ordinary `CALLS` edge. A same-named project method should not become the target solely because the actual receiver type is unknown or external.

This is the Python equivalent of the false-positive method-call class described for PHP in #606 and fixed for Perl in #476. It is separate from #1271: that issue concerns collapsing a genuine in-graph polymorphic target set to one override, while this issue concerns fabricating a target that is not valid at all.

#### Public source

Repository and commit:

- https://github.com/Enferlain/sd-scripts/tree/de2dd0ebae40174e8336157e8928e36cfd013284
- Call sites: https://github.com/Enferlain/sd-scripts/blob/de2dd0ebae40174e8336157e8928e36cfd013284/library/training/phases/training_loop.py#L474-L545
- Unrelated test mock selected for `accelerator.print`: https://github.com/Enferlain/sd-scripts/blob/de2dd0ebae40174e8336157e8928e36cfd013284/tests/unit/test_resume_logic.py#L19-L38
- Unrelated `backward` implementation: https://github.com/Enferlain/sd-scripts/blob/de2dd0ebae40174e8336157e8928e36cfd013284/library/models/sdxl/unet.py#L52-L155
- Arbitrarily selected scheduler implementation: https://github.com/Enferlain/sd-scripts/blob/de2dd0ebae40174e8336157e8928e36cfd013284/library/pipelines/gradual_latent.py#L103-L150

Relevant source:

```python
accelerator.print(...)
# ...
accelerator.backward(modifier_output.loss)
# ...
trainer.optimizer.step()
trainer.lr_scheduler.step()
trainer.optimizer.zero_grad(set_to_none=True)
```

Run a full index, then inspect the direct call edges:

```json
{
  "project": "home-imi-Projects-sd-scripts",
  "query": "MATCH (a:Function)-[r:CALLS]->(b:Method) WHERE a.qualified_name = 'home-imi-Projects-sd-scripts.library.training.phases.training_loop._run_epoch_steps' RETURN b.qualified_name, r.line, r.callee, r.strategy, r.confidence, r.candidates ORDER BY r.line"
}
```

Relevant rows include:

```text
line=482 callee=accelerator.print
  -> tests.unit.test_resume_logic.MockAccelerator.print
  strategy=field_type_hint confidence=0.85 candidates=2

line=532 callee=accelerator.backward
  -> library.models.sdxl.unet.FlashAttentionFunction.backward
  strategy=suffix_match confidence=0.28 candidates=2

line=541 callee=trainer.lr_scheduler.step
  -> library.pipelines.gradual_latent.EulerAncestralDiscreteSchedulerGL.step
  strategy=suffix_match confidence=0.01 candidates=84
```

There is also a compact repository-wide invariant demonstrating the false positives. Count production-library callers connected to test targets:

```cypher
MATCH (caller:Function)-[r:CALLS]->(callee)
WHERE caller.file_path STARTS WITH 'library/'
  AND callee.file_path STARTS WITH 'tests/'
RETURN count(r)
```

Result: `26`.

Repeat with `(caller:Method)`:

```cypher
MATCH (caller:Method)-[r:CALLS]->(callee)
WHERE caller.file_path STARTS WITH 'library/'
  AND callee.file_path STARTS WITH 'tests/'
RETURN count(r)
```

Result: `29`.

The current index therefore contains 55 production-to-test `CALLS` edges. Production modules in this repository do not import or call the test suite, so these are false targets rather than merely low-confidence but possible calls.

The reproduced full index reported 12,131 nodes and 70,850 edges. `check_index_coverage` returned `no_recorded_issue` with matching generation metadata for every source path cited above.

### Confirmed implementation behavior

Inspection of the resolver implementation matching this observed behavior found:

1. Python attribute calls do not set `CBMCall.is_method`; extraction currently sets it only for Perl and JavaScript/TypeScript/TSX member calls.
2. When Python LSP/type resolution misses, the generic registry resolves by the callee's simple name. For 2–256 same-name candidates it selects one candidate and still emits an ordinary `CALLS` edge, with confidence reduced according to candidate count.
3. `try_field_type_hint()` derives a type-like string by capitalizing the receiver name (`accelerator` -> `Accelerator`) and selects the first candidate qualified name containing that string. It does not require exact receiver-type evidence or deprioritize test/mock classes, allowing `MockAccelerator.print` to receive confidence `0.85`.
4. The existing Python unknown-method test only asserts that any resolution has confidence below `0.6`; it does not require the false target to be absent and does not verify the final graph edge. It also does not cover the `field_type_hint` path, which can raise a fabricated edge to high confidence.

Perl and TS/JS already have language-gated safeguards that suppress weak member-call strategies such as `suffix_match`, `unique_name`, and `field_type_hint` when receiver-aware resolution failed. Python currently cannot use an equivalent safeguard because its attribute calls are not marked as member calls.

### Expected acceptance criteria

- An unresolved Python member call must not become an ordinary `CALLS` edge to an unrelated same-named project method through `suffix_match`, `unique_name`, `field_type_hint`, or similar weak textual fallback.
- Production callers must not resolve to methods defined only in test/mock classes without actual dependency/type evidence.
- LSP-, import-, same-module-, and otherwise receiver-aware genuine calls must remain supported.
- `field_type_hint` should require stronger receiver-type evidence and should not select test/mock candidates for production callers based only on a substring match.
- Regression coverage should exercise the complete extraction-to-graph pipeline and assert final `CALLS` edges, rather than only checking intermediate resolver confidence.
- If unresolved calls are retained for diagnostics, their ambiguity should be explicit rather than represented as one concrete target.

### Impact

These false edges directly corrupt caller/callee tracing and impact analysis. `trace_path` may report unrelated implementations as dependencies, and `include_tests=false` can hide a false production-to-test edge entirely, making the original call disappear instead of indicating that it was unresolved.

The practical workaround is to inspect raw edge metadata and distrust Python member-call edges produced by weak strategies, especially `suffix_match` with multiple candidates or `field_type_hint` without exact type evidence. This is difficult to apply through normal traversal output because `trace_path` does not surface those properties.

### Logs

No runtime error is emitted. Indexing completes successfully; the problem is visible in the resulting graph.

### Project scale

12,131 nodes / 70,850 edges / 801 indexed files

### Confirmations

- [x] I searched existing issues and did not find a Python report covering this failure.
- [x] The reproduction uses shareable code from a public OSS repository.

Enferlain:
## What does this PR do?

Fixes #1276.

Python attribute calls were not consistently classified as member calls, allowing weak `suffix_match`, `unique`, and `field_type_hint` resolutions to create false-positive `CALLS` edges when no import-reachable target existed.

This change marks unresolved receiver calls as method calls and applies the existing weak-member suppression policy consistently in sequential and parallel resolution. Direct `self`/`cls`/`super` calls and receivers rooted in an imported local name are exempt, so canonical calls such as `from pkg import helper; helper.compute()` still resolve. Strong LSP, import, same-module, and service resolutions remain available.

Regression coverage exercises extraction plus sequential and parallel pipelines, including positive imported-module receiver calls and negative unrelated member calls.

## Verification

- `nix develop --command make -j8 -f Makefile.cbm test-focused TEST_SUITES='extraction registry pipeline'`
- Result: 568 passed.
- `nix develop --command make -f Makefile.cbm lint-format lint-no-suppress`
- Result: passed.
- `git diff --check` and changed-range `clang-format --dry-run --Werror`: passed.
- Full local smoke was blocked by the active installed MCP version-cohort guard; the updated PR CI is the clean-process smoke verification.
- Full `lint-ci` reached cppcheck but returned its repository-wide `toomanyconfigs` informational diagnostic; no patch-local format or suppression violation remained.

## Checklist

- [x] Patch-local formatting and suppression checks pass
- [x] New behavior is covered by regression tests that would fail without this fix

### DIFF
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -240,8 +240,9 @@ typedef struct {
     int loop_depth;                     // enclosing loop nesting at the call site
     int branch_depth;                   // enclosing branch nesting at the call site
     int start_line;                     // 1-based source line of the call (for def range-match)
-    bool is_method;                     // method/member call with a non-self receiver. Perl:
-                                        // arrow/method call ($obj->m). TS/JS/TSX: member call
+    bool is_method;                     // method/member call with an unresolved receiver. Perl:
+                                        // arrow/method call ($obj->m). Python: x.foo() where x is
+                                        // not self/cls/super or an imported name. TS/JS/TSX:
                                         // x.foo() whose receiver is not this/super. Default false.
 } CBMCall;
 
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -2145,6 +2145,49 @@ static char *resolve_objectscript_instance_call(CBMArena *a, TSNode node, const 
     return NULL;
 }
 
+/* True when a Python attribute call receiver is safe from the weak-member
+ * guard. Direct self/cls and super() receivers retain class-local semantics.
+ * An imported identifier — including the root of an attribute chain — must
+ * remain eligible for import-map resolution (`helper.compute()`). */
+static bool python_receiver_is_exempt(CBMExtractCtx *ctx, TSNode receiver) {
+    if (ts_node_is_null(receiver)) {
+        return false;
+    }
+
+    if (strcmp(ts_node_type(receiver), "call") == 0) {
+        TSNode fn = ts_node_child_by_field_name(receiver, TS_FIELD("function"));
+        if (!ts_node_is_null(fn) && strcmp(ts_node_type(fn), "identifier") == 0) {
+            char *name = cbm_node_text(ctx->arena, fn, ctx->source);
+            return name && strcmp(name, "super") == 0;
+        }
+        return false;
+    }
+
+    bool direct_identifier = strcmp(ts_node_type(receiver), "identifier") == 0;
+    TSNode root = receiver;
+    while (!ts_node_is_null(root) && strcmp(ts_node_type(root), "attribute") == 0) {
+        root = ts_node_child_by_field_name(root, TS_FIELD("object"));
+    }
+    if (ts_node_is_null(root) || strcmp(ts_node_type(root), "identifier") != 0) {
+        return false;
+    }
+
+    char *name = cbm_node_text(ctx->arena, root, ctx->source);
+    if (!name) {
+        return false;
+    }
+    if (direct_identifier && (strcmp(name, "self") == 0 || strcmp(name, "cls") == 0)) {
+        return true;
+    }
+    for (int i = 0; i < ctx->result->imports.count; i++) {
+        const char *local_name = ctx->result->imports.items[i].local_name;
+        if (local_name && strcmp(local_name, name) == 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
 void handle_calls(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec, WalkState *state) {
     if (!spec->call_node_types || !spec->call_node_types[0]) {
         return;
@@ -2220,6 +2263,19 @@ void handle_calls(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec, Walk
             if (ctx->language == CBM_LANG_PERL &&
                 strcmp(ts_node_type(node), "method_call_expression") == 0) {
                 call.is_method = true;
+            }
+            // Python receiver-aware guard (#1276). Imported receivers remain
+            // ordinary calls: module.function() is Python's canonical
+            // cross-file call shape and the import map can resolve it. Unknown
+            // object receivers are flagged so weak short-name matching cannot
+            // fabricate an edge. self/cls/super() keep class-local semantics.
+            if (ctx->language == CBM_LANG_PYTHON && strcmp(ts_node_type(node), "call") == 0) {
+                TSNode fn = ts_node_child_by_field_name(node, TS_FIELD("function"));
+                if (!ts_node_is_null(fn) && strcmp(ts_node_type(fn), "attribute") == 0) {
+                    TSNode obj = ts_node_child_by_field_name(fn, TS_FIELD("object"));
+                    bool receiver_is_exempt = python_receiver_is_exempt(ctx, obj);
+                    call.is_method = !receiver_is_exempt;
+                }
             }
             // TS/JS/TSX receiver-aware guard (#592/#606 direction; same intent
             // as the Perl flag above). Flag a member call x.foo() whose receiver
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -571,8 +571,9 @@ static int resolve_single_call(cbm_pipeline_ctx_t *ctx, CBMCall *call,
         return 0;
     }
 
-    /* TS/JS/TSX weak-method suppression (#592/#606). A member call x.foo() only
-     * reaches the registry when the TS-LSP could not resolve the receiver type
+    /* Dynamic-language weak-method suppression (#592/#606/#1276). A member call
+     * x.foo() only reaches the registry when the language-specific resolver
+     * could not resolve the receiver type
      * (the LSP block above already returned for type-resolved calls, including
      * the "resolved but target out of gbuf" fall-through). Binding such a call
      * by a weak short-name strategy fabricates an edge (`re.test()` -> a project
@@ -581,10 +582,10 @@ static int resolve_single_call(cbm_pipeline_ctx_t *ctx, CBMCall *call,
      * defer to emit_classified_edge and suppress ONLY the plain-CALLS
      * fall-through, so every service edge stays main-identical. res.strategy may
      * be lsp_* here; the helper's explicit drop-list leaves lsp_* untouched. */
-    bool is_tsjs =
-        lang == CBM_LANG_JAVASCRIPT || lang == CBM_LANG_TYPESCRIPT || lang == CBM_LANG_TSX;
-    bool tsjs_drop_plain_call =
-        cbm_tsjs_suppress_weak_method_match(is_tsjs, call->is_method, res.strategy);
+    bool suppress_weak_member = lang == CBM_LANG_PYTHON || lang == CBM_LANG_JAVASCRIPT ||
+                                lang == CBM_LANG_TYPESCRIPT || lang == CBM_LANG_TSX;
+    bool drop_plain_call =
+        cbm_suppress_weak_member_match(suppress_weak_member, call->is_method, res.strategy);
 
     /* Service-pattern HTTP/ASYNC calls to an EXTERNAL client library (e.g.
      * `requests.get("/api/orders/{id}")`) resolve to a QN containing the library
@@ -611,7 +612,7 @@ static int resolve_single_call(cbm_pipeline_ctx_t *ctx, CBMCall *call,
         return 0;
     }
     emit_classified_edge(ctx, call, source_node, target_node, &res, module_qn, imp_keys, imp_vals,
-                         imp_count, tsjs_drop_plain_call);
+                         imp_count, drop_plain_call);
     return SKIP_ONE;
 }
 
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -2321,20 +2321,21 @@ static void resolve_file_calls(resolve_ctx_t *rc, resolve_worker_state_t *ws, CB
             continue;
         }
 
-        /* TS/JS/TSX weak-method suppression (#592/#606). The receiver-aware guard
-         * must NOT drop this call here: doing so would also skip the #523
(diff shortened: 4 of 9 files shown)


DeusData:
Reviewed properly, and this needs changes — but I want to start with what is right, because the direction is correct and the diagnosis behind it is excellent.

**Issue #1276 is one of the best bug reports this repo has received.** Reproducible on public code, a quantified repo-wide invariant (55 production→test edges in a 12k-node repo), and a correct implementation-level diagnosis — unflagged `is_method` plus `field_type_hint` doing little more than capitalising the receiver token. The fabricated edges you documented are genuinely wrong: `accelerator.print()` landing on a **test mock** at 0.85 confidence, `.backward()` on an unrelated `FlashAttentionFunction`, `.step()` picking one of 84 candidates at 0.01. Under our "a wrong edge is worse than a missing one" rule, suppressing those is right, and you landed it on the established Perl and TS/JS precedent including the subtle explicit-drop-list rationale that protects the `lsp_*` strategies in the parallel path.

Your tests are exemplary too — both resolver paths exercised, the parallel one reconstructed with the `field_type_hint` trigger, and assertions on the **final graph** rather than intermediate confidences. That is exactly what #1276's acceptance criteria asked for.

**Now the problem, and unlike most red PRs I have looked at tonight, this CI failure is genuinely yours.**

Our smoke fixture is:

```python
from pkg import helper
result = helper.compute(42)
```

and all three smoke legs fail identically with `FAIL: trace_path found 0 callers for 'compute'`. The suppression removed a **true** `main → compute` edge.

The reason is a real Python-versus-TypeScript semantic difference rather than a coding slip. In Python, `module.function()` is *the* dominant cross-file call shape, and its receiver is an imported module the indexer already knows — not an unresolvable object. But the blanket `is_method` flag cannot distinguish a module receiver from an unknown-object receiver, so whenever such a call resolves through a weak strategy the true edge dies. TS/JS never had this failure mode because ES-module imports are not attribute calls on a namespace in the same way. This is not an edge case; it is the canonical Python pattern.

**The fix shape:** exempt receivers that are **known imports of the file**. Extraction already has the import table, or the resolution side can consult the import map before applying the guard — flag `is_method` only when the receiver identifier is *not* bound by an import. That preserves everything #1276 wants suppressed (`accelerator` and `trainer` are parameters, not imports) while keeping `helper.compute` alive.

Please also add a positive test pinning `module.function()` survival — that coverage gap is precisely what let this through, and your own checklist honestly noted the smoke suite was not run locally.

**One more genuine failure:** clang-format violations in the new `extract_calls.c` block, lines 2224-2235. `make -f Makefile.cbm lint-ci` reproduces it. One caveat that will save you a wasted round trip — our formatter is the **Homebrew LLVM** build; a standalone `clang-format-20` produces spurious whole-file drift that looks nothing like the real lines.

**Two of your six failures are not yours:** `security / codeql-gate` timed out while the actual `analyze` job passed — a nondeterministic gate we have seen race on other PRs tonight — and `ci-ok` is just its aggregator.

One thing worth the maintainer's eyes that I have flagged separately: the `unique_name` drop also costs true edges in small projects or where the LSP misses. That trade was accepted for TS/JS on the strength of an LSP backstop, and Python's hybrid LSP gives a similar one — but Python is the flagship language, so it is worth a deliberate nod rather than inheritance.

With the import-receiver exemption this lands as a flagship graph-quality improvement. The hard part — finding and proving the false-edge factory — is already done.

Enferlain:
Addressed the review findings in `1ec5212`:

- imported module receivers are now exempt from weak-member suppression, including roots of attribute chains,
- added positive regressions for imported-module calls in extraction and both sequential/parallel pipelines,
- retained negative coverage for unrelated member calls,
- fixed the affected formatting, and
- rebased onto current `main`.

Local verification:

- focused sanitized `extraction registry pipeline`: **568 passed**
- `lint-format` and `lint-no-suppress`: passed
- changed-range `clang-format --dry-run --Werror` and `git diff --check`: passed

The full local smoke cannot coexist with the currently active installed MCP version cohort, so I left that process untouched and am relying on the updated PR's clean CI environment for smoke verification. Full `lint-ci` reached cppcheck but ended on its repository-wide `toomanyconfigs` informational diagnostic; there are no remaining patch-local format or suppression failures.

DeusData:
Thank you for updating the Python weak-member-call suppression after the earlier review. It is now routed as a high-priority parsing fix in `0.9.1-rc` and is back in the maintainer review queue. The queue is currently full, so the final re-review may take a little time.

DeusData:
The direction is right — Python needs the same weak-match suppression TS/JS got (#592/#606 precedent), and #1276 is real. But CI is telling us the guard overshoots: the resolution probe fails on all five legs at python/S6/inherited_method (calls=0, expected>=1) — a legitimate inherited-method edge your suppression removes. Per this repo's rules the probe must not be weakened; production has to keep emitting that edge, e.g. by exempting receivers resolvable to class-hierarchy members before the weak-match check. That plus a rebase (cbm.h/extract_calls.c drifted) and we're glad to re-review. Also note the blanket is_method=true for all Python attribute calls touches every other is_method consumer — worth narrowing or justifying explicitly.

DeusData:
Landed. **On main as of `5d96377a`** (PR #1903), with `Co-authored-by:` credit to you. Fixes #1276.

@Enferlain — thank you. The design, the receiver exemption and the tests are yours; the distill exists because the branch had drifted 584 commits and I would rather absorb that than hand it back to you.

**Your scoping was right, and that mattered more than it might look.** A parallel PR in this cluster restricted `cbm_registry_resolve` globally instead of per-language. It was measured against the full corpus set: **−244,489 CALLS edges, −11.4% of the entire call graph, 32.5 edges destroyed per suspicious edge removed** — and 5,571 Go call sites lost a correct same-module target and gained a wrong cross-package one. Yours is gated on file language, wired at both `pass_calls.c` and `pass_parallel.c`, and keeps `same_module` — which is why it lands and that one does not.

**One thing I had to fix in the rebase, worth knowing.** Main's gate grew `CBM_LANG_ARKTS` after ArkTS landed in `0df990aa`. Your line predates it, so taking it verbatim would have silently dropped ArkTS from the weak-member guard and reintroduced the #592/#606 false-edge class — with no test to catch it, since ArkTS did not exist when you wrote this. Both gates now read `PYTHON || JAVASCRIPT || TYPESCRIPT || TSX || ARKTS`, and the lockstep requirement is written into the code at both sites so nobody adds a language to one resolver only.

**On the recall question you raised, now settled with measurement.** The guard costs one real edge class: an inherited method reached through a parameter-held receiver. `python/S6` was measured to be resolving via `strategy=unique_name, cands=1, conf=0.7500` — a short-name guess that happened to be right in a 2-file fixture. TypeScript hit exactly this when the receiver-aware guard landed and the repo **accepted it**, rewriting its S6 probe from `>= 1` to `== 0` with a tripwire. Python now records the same trade, with the measurement in the comment and the flip-back condition noted for when `py_lsp_cross` resolves inheritance. That work is underway.

CI then caught one more: `test_parallel.c`'s LSP-ambiguity probe also depended on a weak match. Measured before changing anything — the suppressed edge was `suffix_match` with **3 candidates at conf 0.55**, a coin flip among `Alpha/Beta/Legacy.render`, while sibling tests on identical source resolve via `lsp_method` at conf 0.90 and are untouched. That contrast is the line between a guess and knowledge, and it is why the guard is doing the right thing rather than over-reaching.

Your aliased-import handling still holds too, pinned by test — `#1371` changed Python aliased-from-import handling since you wrote this, and `CBMImport.local_name` survives it.

Thank you for a well-scoped fix with tests that pinned both directions. It made the distill straightforward, and the queue is better for it.

PR #1325, opened 2026-07-28
fix(search): preserve compound requested fields

### LINKED ISSUE #1284: search_graph returns null for list-valued requested fields despite stored graph properties
### Version

`codebase-memory-mcp dev`

### Platform

Linux (x64)

### What happened?

`search_graph(fields=[...])` returns `null` for list-valued node properties even though the properties exist in the graph and `query_graph` returns their values correctly.

This was reproduced with:

- `Class.base_classes`
- `Method.decorators`
- `Method.decorator_tags`

Scalar requested fields such as `signature` render normally.

### Public reproduction

Repository and commit:

- https://github.com/Enferlain/sd-scripts/tree/de2dd0ebae40174e8336157e8928e36cfd013284
- Example class: https://github.com/Enferlain/sd-scripts/blob/de2dd0ebae40174e8336157e8928e36cfd013284/library/strategies/sd/training.py#L19-L50
- Example decorated method: https://github.com/Enferlain/sd-scripts/blob/de2dd0ebae40174e8336157e8928e36cfd013284/library/strategies/base/contracts.py#L376-L403

A fresh full index reported:

```text
12,131 nodes
70,850 edges
generation_matches=true
```

### Reproduction: class bases

```json
{
  "project": "home-imi-Projects-sd-scripts",
  "label": "Class",
  "name_pattern": "^(SdTrainingStrategy|Sd3TrainingStrategy|SdxlTrainingStrategy)$",
  "fields": ["base_classes"],
  "format": "json",
  "limit": 10,
  "offset": 0
}
```

Actual `search_graph` rows contain:

```json
["SdTrainingStrategy", "Class", "19-50", 15, 13, null]
["Sd3TrainingStrategy", "Class", "17-63", 2, 12, null]
["SdxlTrainingStrategy", "Class", "20-58", 12, 13, null]
```

But the property is present and ordered in `query_graph`:

```cypher
MATCH (c:Class)
WHERE c.name IN ['SdTrainingStrategy', 'Sd3TrainingStrategy', 'SdxlTrainingStrategy']
RETURN c.qualified_name, c.base_classes
ORDER BY c.qualified_name
```

For example, `SdTrainingStrategy.base_classes` returns:

```json
[
  "SdModelLoadingStrategy",
  "SdTokenizeStrategy",
  "SdTextEncodingStrategy",
  "SdConditioningStrategy",
  "SdCachingStrategy",
  "SdSampleGenerationStrategy",
  "SdCheckpointingStrategy",
  "SdValidationStrategy",
  "SdDiffusionTrainingStrategy",
  "SdDenoiserCallingStrategy",
  "SdModelPreparationStrategy",
  "WeightedPromptStrategy",
  "TrainingStrategy"
]
```

### Reproduction: method decorators

```json
{
  "project": "home-imi-Projects-sd-scripts",
  "label": "Method",
  "name_pattern": "^process_batch$",
  "qn_pattern": "^home-imi-Projects-sd-scripts\\.library\\.strategies\\.base\\.contracts\\..*",
  "fields": ["decorators", "decorator_tags"],
  "format": "json",
  "limit": 10,
  "offset": 0
}
```

Actual row:

```json
["process_batch", "Method", "380-403", 0, 10, null, null]
```

But `query_graph` returns:

```text
decorators=["@abstractmethod"]
decorator_tags=["abstractmethod"]
```

The graph schema also lists `base_classes`, `decorators`, and `decorator_tags` as properties of their respective node labels.

All cited source paths returned `no_recorded_issue` with matching coverage metadata.

### Expected behavior

Requested list-valued properties should be serialized in `search_graph` results, preserving order where the property is ordered:

```json
["SdTrainingStrategy", "Class", "19-50", 15, 13, ["SdModelLoadingStrategy", "..."]]
```

If list-valued properties are intentionally unsupported, `search_graph` should return an explicit validation error or document the limitation rather than silently returning `null` for an existing property.

### Impact

This is a smaller query-output issue; the underlying index data is intact and `query_graph` is a workaround. It still affects normal agent workflows because `search_graph` is the preferred discovery tool:

- inheritance order appears absent;
- abstract/decorator metadata appears absent;
- agents may draw false conclusions or issue unnecessary fallback queries;
- inheritance and framework audits become more expensive.

### Confirmations

- [x] Searched current open and closed issues; no exact duplicate found.
- [x] Reproduced across three separate list-valued properties and two node labels.
- [x] Scalar requested fields work.
- [x] Underlying values are confirmed through `query_graph`.

Enferlain:
## What does this PR do?

Fixes #1284.

Requested extra fields in `search_graph` were serialized only for scalar JSON values. Stored arrays and objects fell through to an empty cell in table output and were silently misrepresented.

This change centralizes requested-field serialization: table output emits compact JSON text for compound values, while JSON output copies the original JSON value so its type is preserved. Regression coverage verifies both output formats.

## Verification

- `make -f Makefile.cbm test-focused TEST_SUITES='cypher extraction registry pipeline mcp index_resilience'`
- Result: 884 passed, 2 skipped

## Checklist

- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) — not run locally
- [x] New behavior is covered by regression tests that would fail without this fix

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -3082,6 +3082,27 @@ static int sg_parse_fields(const char *args, const char *out[], int max_out, yyj
     return n;
 }
 
+/* Append a property as one compact-output cell. Compound values stay one
+ * column by using their compact JSON representation; the cell emitter quotes
+ * and escapes that representation as needed. */
+static void sg_toon_property_cell(cbm_sb_t *sb, yyjson_val *v) {
+    if (v && yyjson_is_str(v)) {
+        cbm_tree_cell_str(sb, yyjson_get_str(v), false);
+    } else if (v && yyjson_is_bool(v)) {
+        cbm_tree_cell_bool(sb, yyjson_get_bool(v), false);
+    } else if (v && yyjson_is_int(v)) {
+        cbm_tree_cell_int(sb, yyjson_get_int(v), false);
+    } else if (v && yyjson_is_real(v)) {
+        cbm_tree_cell_real(sb, yyjson_get_real(v), false);
+    } else if (v && !yyjson_is_null(v)) {
+        char *json = yyjson_val_write(v, 0, NULL);
+        cbm_tree_cell_str(sb, json ? json : "", false);
+        free(json);
+    } else {
+        cbm_tree_cell_str(sb, "", false);
+    }
+}
+
 /* Append one row's extra-field cells, pulled from the node's properties. */
 static void sg_toon_extra_cells(cbm_sb_t *sb, const char *props_json, const char *const *fields,
                                 int nfields) {
@@ -3090,17 +3111,7 @@ static void sg_toon_extra_cells(cbm_sb_t *sb, const char *props_json, const char
     yyjson_val *pr = pd ? yyjson_doc_get_root(pd) : NULL;
     for (int f = 0; f < nfields; f++) {
         yyjson_val *v = (pr && yyjson_is_obj(pr)) ? yyjson_obj_get(pr, fields[f]) : NULL;
-        if (v && yyjson_is_str(v)) {
-            cbm_tree_cell_str(sb, yyjson_get_str(v), false);
-        } else if (v && yyjson_is_bool(v)) {
-            cbm_tree_cell_bool(sb, yyjson_get_bool(v), false);
-        } else if (v && yyjson_is_int(v)) {
-            cbm_tree_cell_int(sb, yyjson_get_int(v), false);
-        } else if (v && yyjson_is_real(v)) {
-            cbm_tree_cell_real(sb, yyjson_get_real(v), false);
-        } else {
-            cbm_tree_cell_str(sb, "", false);
-        }
+        sg_toon_property_cell(sb, v);
     }
     if (pd) {
         yyjson_doc_free(pd);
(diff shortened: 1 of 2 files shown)


DeusData:
Reviewed this properly, and the short version is: **it is ready, and we would like to merge it as shown — may we?**

It is still marked draft, so I am asking rather than just merging. If you were planning further changes, say so and we will wait. If draft is simply left over, mark it ready (or reply here) and it goes in.

**What I verified**, since it is worth knowing the review was real:

The bug reproduces on today's `main`. All three emitters type-switch on scalars only, so a node property whose stored value is a JSON array or object — `base_classes`, `decorators` — silently falls through: the text and TOON tables render an empty `"-"` cell, and `format:"json"` puts `null` in the row. That is #1284's exact symptom.

What I liked is that you did not patch one emitter. You found all three duplicated scalar-only switches — table, skeleton tree, and JSON — and centralised the cell logic into `sg_toon_property_cell`, so the two text paths can no longer drift apart. They had already drifted, incidentally: their int/real ordering differed. That is the fix built the way we would have wanted it built.

The type-preservation split is the right call too: compact JSON where the output has to stay inside one table cell, and `yyjson_val_mut_copy` where the consumer is getting real JSON — rather than the stringified-JSON-inside-JSON compromise that would have been the easy way out.

I also checked the text path stays safe: compact JSON contains `"`, which forces `needs_quotes()` → `append_quoted()`, so quotes and backslashes are escaped, control bytes are JSON-escaped, and invalid UTF-8 is sanitised. The cell remains one parseable, valid-UTF-8 column, consistent with the recent tree-format UTF-8 hardening.

Both regression tests bind. Reverting the fix turns the text assert red (`HandlerBase`/`Audited` versus the old `"-"`) and the JSON assert red (`["HandlerBase","Audited"]` versus the old `null`).

No defects found. Two entirely optional notes:

- `ASSERT_NOT_NULL(strstr(inner, "base_classes"))` passes without the fix, since the column header alone satisfies it — harmless next to the value asserts that do the real work.
- Worth a CHANGELOG sentence eventually: `format:"json"` rows can now contain nested arrays/objects where they previously held only scalars or `null`. Anyone consuming rows positionally will want to know.

Thank you for the minimal public repro in #1284 as well — it made confirming both the symptom and the fix scope trivial, and that is a real kindness to whoever reviews.

DeusData:
Thank you for marking the compound-field projection fix ready and for addressing the prior feedback. It is now routed as a high-priority `0.9.1-rc` search correctness fix and is back in the maintainer review queue. The queue is currently full, so the final re-review may take a little time.

DeusData:
Merged (2397d58d) — thank you! Deduplicating the three scalar emitters into sg_toon_property_cell while adding the type-preserving JSON copy fixed the bug and left the code better than it was found — much appreciated. #1284 closed.

PR #1326, opened 2026-07-28
fix(index): preserve persisted coverage summaries

### LINKED ISSUE #1287: index_repository reports parse_partial_count: 0 after incremental/artifact reload despite persisted coverage gaps
### Version

`codebase-memory-mcp dev` on Linux x86_64.

### What happened

`index_repository` reports `parse_partial_count: 0` after an incremental/no-change full request and after importing a persisted team artifact, even though the resulting persisted graph contains known parse-partial files.

The graph data and persisted coverage metadata are otherwise correct:

- `index_status` immediately reports `parse_partial.count: 2`
- `query_graph(graph="missed")` immediately returns the same two `parse_partial` files
- `check_index_coverage` reports `generation_matches: true`
- the same coverage survives daemon restart and reload

This means the response from `index_repository` can falsely make the current index look coverage-clean.

### Reproduction

Using a repository with two stable parse-partial files and an existing persisted artifact:

```bash
codebase-memory-mcp cli --json index_repository \
  '{"repo_path":"/path/to/repo","mode":"full","persistence":true}'

codebase-memory-mcp cli --json index_status \
  '{"project":"project-name","verbose":true}'

codebase-memory-mcp cli --json query_graph \
  '{"project":"project-name","graph":"missed","query":"MATCH (f:File) RETURN f.file_path, f.kind, f.detail ORDER BY f.file_path"}'
```

Observed `index_repository` response:

```json
{
  "nodes": 12131,
  "edges": 70812,
  "skipped_count": 0,
  "parse_partial_count": 0,
  "status": "indexed"
}
```

Immediate `index_status` result:

```json
{
  "parse_partial": {
    "files": [
      {
        "path": "library/logging/live_plotter/web/dashboard_extras.html",
        "error_ranges": "244-244"
      },
      {
        "path": "run_benchmark.ps1",
        "error_ranges": "..."
      }
    ],
    "count": 2,
    "truncated": false
  }
}
```

Immediate missed-graph query:

```text
library/logging/live_plotter/web/dashboard_extras.html  parse_partial  244-244
run_benchmark.ps1                                      parse_partial  ...
total: 2
```

I reproduced the same mismatch with an empty isolated `CBM_CACHE_DIR`, where `index_repository` imported `.codebase-memory/graph.db.zst` and produced the correct persisted graph. Its response still said `parse_partial_count: 0`, while a fresh CLI process loading that cache reported both partial files.

### Expected behavior

After incremental reconciliation or artifact import, `index_repository` should report the current persisted coverage state, including pre-existing/imported parse-partial and skipped files.

If these fields are intentionally run-local rather than current-state totals, they should be named/documented accordingly and the response should also expose the authoritative persisted totals. The current tool description says the response reports files that were not fully indexed, so a zero is unsafe for agents deciding whether graph-only evidence is sufficient.

### Scope

This appears to be a response-assembly/coverage-summary issue, not graph corruption:

- no stale or duplicate symbols were found
- `OVERRIDE` edges survived incremental refresh, deletion, restart, and artifact reload
- node and edge totals remained stable across reloads
- generation metadata remained matched

This follows up the coverage-signal work in #963.

It is separate from the cross-clone artifact identity problem in #885.

Enferlain:
## What does this PR do?

Fixes #1287.

The index response derived `parse_partial_count` only from errors produced by the current pipeline run. Incremental no-op and artifact-backed paths could consequently report zero even when persisted coverage rows still recorded parse gaps.

This change loads the authoritative coverage summary from the stored graph when composing the response, with a run-local fallback if that read fails. By-design exclusions remain reported separately. A regression test covers an unchanged incremental run retaining its persisted partial-parse status.

## Verification

- `make -f Makefile.cbm test-focused TEST_SUITES=index_resilience` — 6 passed
- `make -f Makefile.cbm test-focused TEST_SUITES='cypher extraction registry pipeline mcp index_resilience'` — 884 passed, 2 skipped

## Checklist

- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) — not run locally
- [x] New behavior is covered by a regression test that would fail without this fix


DeusData:
Reviewed — **this is correct and merge-quality, and neither of its two failing checks is your fault.** It is the cleanest of your four PRs.

**The loss is real on main, and I verified it rather than taking the issue's word for it.** `build_index_success_response` derives `skipped` and `parse_partial` from the *current run's* `file_errors` only (`mcp.c:6969-6970`). So an incremental run that never revisits a previously-flagged file reports `parse_partial_count: 0`, while `index_status` — which reads the persisted coverage rows — says 2. That is exactly #1287's repro: two surfaces disagreeing about the same fact, with the more prominent one wrong.

Your fix reads the authoritative persisted coverage, filters out the by-design `not_indexed_dir`/`not_indexed_file` rows (which have their own `not_indexed_files` surface, so counting them here would double-report), maps the rest into the existing `cbm_file_error_t` view, and feeds **the same two emitters**. Reusing the emitters rather than writing a parallel formatter is the right call — the two paths cannot drift.

I checked the memory handling specifically, since it is the kind of change where a borrowed pointer escapes: the borrowed `rows[i]` strings are copied into the JSON document via `yyjson_mut_obj_add_strcpy` **before** `free(failures)` and `cbm_store_free_coverage`, `calloc` failure is checked with correct cleanup, and the const-cast is read-only plumbing. No use-after-free. The fallback to run-local errors when the coverage read fails is the right degradation.

**On the two red checks — both non-genuine:**

- `test / test-unix (macos-15-intel)` failed on `lock_registry_large_queue_parks_non_head_waiters` and `lock_registry_absolute_deadline_survives_repeated_wakes` — daemon lock-registry threading tests that wait on a 500 ms window for waiters to park, on the slowest runner we have. Your diff touches only `src/mcp/mcp.c` response assembly and one test file. The decisive evidence: **the identical leg passed on your sibling PR #1323**, which shares your base and has byte-identical lock-registry code. So it is a scheduler-timing flake, not your change. I have flagged those two tests separately — a bounded-transient-window wait is exactly the flaky class we are supposed to rebuild deterministic.
- `ci-ok` is just the aggregator of that.

Everything else is green, including `codeql-gate` in 5m22s — which incidentally helped prove that the gate timeouts on your #1323 and #1324 are gate nondeterminism rather than anything in those PRs.

**So: may we merge this as shown?** It is still marked draft, which is the only thing holding it. If you were planning further changes, say so and I will wait. If draft is simply left over, mark it ready — or say the word here — and I will re-run CI and merge.

For context on your other three: #1325 also came back MERGE with no defects and I have asked the same question there. #1323 is a sound fix that needs a rebase over #1173/#1176 plus a maintainer decision on full-scan semantics. #1324 has a real recall regression that our smoke suite caught — details on that thread.

DeusData:
Thank you for marking the persisted-coverage fix ready and for responding to the earlier review. It is now routed as a high-priority `0.9.1-rc` correctness fix and is back in the maintainer review queue. The queue is currently full, so the final re-review may take a little time.

PR #1327, opened 2026-07-29
fix(extract): prevent duplicate Function nodes for Java interface/enum methods (#1234)

### LINKED ISSUE #1234: Java: interface methods emitted twice — as both Method and Function nodes (472 spurious Function nodes)
## Environment
- codebase-memory-mcp 0.9.0 (macOS arm64, Darwin 25.0.0)
- Project: Java 21 / Spring Boot 3 monorepo, ~1,358 Java files, indexed in full mode (20,530 nodes / 70,485 edges)

## Summary
Methods declared on Java interfaces are emitted as **two nodes**: a `Method` (correct, with `parent_class`) and a duplicate `Function` node at the same file/line. Java has no free-standing functions, so any `Function` node in a `.java` file is spurious. Our graph has **472** Function nodes pointing into `.java` files.

## Repro
Source (`MarketplaceReservationService.java`, an interface):
```java
ReservationResponseDTO createReservation(Authentication authentication, ReservationRequestDTO request);
```

Cypher:
```cypher
MATCH (m) WHERE m.name = 'createReservation' RETURN m.qualified_name, labels(m)
```

Result contains both:
```
...reservation.service.MarketplaceReservationService.createReservation  ["Method"]
...reservation.service.createReservation                                ["Function"]   <- duplicate, same file/line
```

Aggregate check:
```cypher
MATCH (f:Function) WHERE f.file_path ENDS WITH '.java' RETURN count(f)  // 472
```

## Impact
- Search results show doubled entries for interface methods.
- Node/label counts and any per-label analytics are skewed.
- Structural boosting in BM25 (Functions +10) double-boosts these phantoms.

## Notes
Looks like the interface-body visitor emits the method both under the parent type and at file scope. Happy to test a fix.

harshitaajoshi:
## What does this PR do?

Fixes #1234.

Java interface and enum methods were emitted as both a Method node (correct, via `extract_class_methods`) and a duplicate Function node (incorrect, via the `walk_defs` fallback).

Two changes fix this:

1. **Prevention in `push_class_body_children`**: adds `interface_body` and `enum_body` to the body-container check, gated to `CBM_LANG_JAVA` so grammars that share these node names (TypeScript, Apex, Dart) are unaffected. Follows the same pattern as the existing Groovy `closure` gate. This stops the fallback path from re-walking `method_declaration` children as top-level functions.

2. **`extract_class_methods` descends into `enum_body_declarations`**: Java enums wrap methods inside this node. Without this, enum methods never received proper class-qualified Method definitions. This also fixes a side effect on main where eight other record kinds carrying `enclosing_func_qn` (usages, throws, read/write, type refs, env accesses, type assigns, string refs, channels) were dangling for enum method bodies.

## Verification

- `build/c/test-runner extraction` -- 266 passed
- `build/c/test-runner convergence_probe` -- 48 passed
- Full suite: 6792 passed, 0 failed
- New extraction tests: `java_interface_no_duplicate_function_issue1234` and `java_enum_dedup_preserves_calls_issue1234` assert Method nodes survive and Function count is zero
- New convergence probe: `cp_interface_method_no_dup_function` asserts surviving Method node carries CALLS edges

## Checklist

      *ASan hangs on macOS Tahoe 26.5 (Apple Clang 17). Full suite passes when built without sanitizers (`SANITIZE=""`).*

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -1239,6 +1239,8 @@ CBMFileResult *cbm_extract_file_ex(const char *source, int source_len, CBMLangua
     cbm_extract_imports(&ctx);
     cbm_extract_unified(&ctx);
 
+    cbm_dedup_class_method_functions(result);
+
     // Channel detection (Socket.IO / EventEmitter) — JS/TS only.
     cbm_extract_channels(&ctx);
 
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -1239,8 +1239,6 @@ CBMFileResult *cbm_extract_file_ex(const char *source, int source_len, CBMLangua
     cbm_extract_imports(&ctx);
     cbm_extract_unified(&ctx);
 
-    cbm_dedup_class_method_functions(result);
-
     // Channel detection (Socket.IO / EventEmitter) — JS/TS only.
     cbm_extract_channels(&ctx);
 
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -596,12 +596,6 @@ CBMFileResult *cbm_extract_file_ex(
     const CBMReturnTypeTable *return_type_table // OS return types, or NULL
 );
 
-// Merge-dedup: remove Function defs that duplicate a Method def at the same
-// location, rewriting any call records that referenced the Function QN to use
-// the surviving Method QN. Must run after both cbm_extract_definitions and
-// cbm_extract_unified so that result->calls is populated.
-void cbm_dedup_class_method_functions(CBMFileResult *result);
-
 // Free all memory associated with a result.
 void cbm_free_result(CBMFileResult *result);
 
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -596,6 +596,12 @@ CBMFileResult *cbm_extract_file_ex(
     const CBMReturnTypeTable *return_type_table // OS return types, or NULL
 );
 
+// Merge-dedup: remove Function defs that duplicate a Method def at the same
+// location, rewriting any call records that referenced the Function QN to use
+// the surviving Method QN. Must run after both cbm_extract_definitions and
+// cbm_extract_unified so that result->calls is populated.
+void cbm_dedup_class_method_functions(CBMFileResult *result);
+
 // Free all memory associated with a result.
 void cbm_free_result(CBMFileResult *result);
 
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -6368,6 +6368,10 @@ static void push_class_body_children(TSNode node, const CBMLangSpec *spec, wd_st
         if (strcmp(ck, "field_declaration_list") == 0 || strcmp(ck, "class_body") == 0 ||
             strcmp(ck, "declaration_list") == 0 || strcmp(ck, "body") == 0 ||
             strcmp(ck, "block") == 0 || strcmp(ck, "suite") == 0 ||
+            // Java interface/enum bodies: gated to Java so grammars that
+            // share these node names (TypeScript, Apex, Dart) are unaffected.
+            (strcmp(ck, "interface_body") == 0 && spec->language == CBM_LANG_JAVA) ||
+            (strcmp(ck, "enum_body") == 0 && spec->language == CBM_LANG_JAVA) ||
             // Groovy class bodies are a `closure` node; routing through the
             // nested-class path keeps methods from being re-walked (and thus
             // double-extracted) as top-level functions. Gated to Groovy so other
@@ -6932,54 +6936,4 @@ void cbm_extract_definitions(CBMExtractCtx *ctx) {
 
     // Extract module-level variables
     extract_variables(ctx, ctx->root, spec);
-}
-
-void cbm_dedup_class_method_functions(CBMFileResult *result) {
-    if (!result || result->defs.count < 2) {
-        return;
-    }
-    CBMDefArray *defs = &result->defs;
-    for (int fi = defs->count - 1; fi >= 0; fi--) {
-        CBMDefinition *func = &defs->items[fi];
-        if (!func->label || strcmp(func->label, "Function") != 0) {
-            continue;
-        }
-        if (!func->name || !func->file_path) {
-            continue;
-        }
-        for (int mi = 0; mi < defs->count; mi++) {
-            if (mi == fi) {
-                continue;
-            }
-            CBMDefinition *method = &defs->items[mi];
-            if (!method->label || strcmp(method->label, "Method") != 0) {
-                continue;
-            }
-            if (!method->name || strcmp(method->name, func->name) != 0) {
-                continue;
-            }
-            if (!method->file_path || strcmp(method->file_path, func->file_path) != 0) {
-                continue;
-            }
-            if (method->start_line != func->start_line) {
-                continue;
-            }
-            const char *func_qn = func->qualified_name;
-            const char *method_qn = method->qualified_name;
-            if (func_qn && method_qn) {
-                for (int ci = 0; ci < result->calls.count; ci++) {
-                    CBMCall *call = &result->calls.items[ci];
-                    if (call->enclosing_func_qn && strcmp(call->enclosing_func_qn, func_qn) == 0) {
-                        call->enclosing_func_qn = method_qn;
-                    }
-                }
-            }
-            if (fi < defs->count - 1) {
-                memmove(&defs->items[fi], &defs->items[fi + 1],
-                        (size_t)(defs->count - fi - 1) * sizeof(defs->items[0]));
-            }
-            defs->count--;
-            break;
-        }
-    }
 }
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -4440,6 +4440,25 @@ static void extract_class_methods(CBMExtractCtx *ctx, TSNode class_node, const c
             }
         }
 
+        // Java enums wrap methods inside enum_body_declarations; iterate
+        // through to reach the method_declaration nodes.
+        if (strcmp(ts_node_type(child), "enum_body_declarations") == 0) {
+            uint32_t dc = ts_node_child_count(child);
+            for (uint32_t j = 0; j < dc; j++) {
+                TSNode dchild = ts_node_child(child, j);
+                if (ts_node_is_null(dchild) ||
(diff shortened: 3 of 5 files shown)


DeusData:
Thank you for this — and thank you for the sibling PR #1329, which I merged earlier tonight (the GoogleTest unique-name fix, `4573a51d8`). This one needs a change before it can follow.

**The failing checks here are genuinely yours**, which is worth saying plainly because it is *not* true of most of the red PRs I have been through tonight — several were cancellation noise or main-side breakage. This one is not.

The signature is unambiguous: the same assertion fails **deterministically on all five platforms** —

```
FAIL tests/test_convergence_probe.c:1143: ASSERT(calls >= 1)
```

macOS 14, macOS 15 Intel, Ubuntu latest, Ubuntu arm64 and Windows CLANG64 all fail identically, while other PRs' runs pass that suite on comparable merge refs. One cause, five legs, no flake pattern.

**What is happening.** The dedup in `internal/cbm/extract_defs.c` removes the duplicate Java interface-method `Function` node, which is the right goal — but the node it drops is the one the convergence probe expects to carry at least one `CALLS` edge. So the duplicate goes away and the edge goes away with it.

**The fix shape:** re-attach the edges to the surviving node *before* deduplicating, rather than dropping a node that still owns relationships. Deduplication in this graph has to be a merge, not a delete — otherwise it trades one accuracy bug (a spurious duplicate node) for a worse one (a missing `CALLS` edge), and we care more about losing true edges than about carrying a redundant node.

**To reproduce locally in seconds**, rather than waiting on CI:

```
build/c/test-runner convergence_probe
```

That is the smallest arena that shows the failure, so you can iterate on it directly and only widen once it is green.

Worth adding a test alongside the fix that asserts the surviving node still carries its `CALLS` edge after dedup — that is exactly the property that broke, and nothing currently guards it.

The underlying change is a real improvement and closes a genuine issue (#1234); it is only the edge-preservation half that is missing. Ping me when it is pushed and I will pick it straight back up.

harshitaajoshi:
@DeusData Pushed the fix. You were right, the dedup needed to be a merge, not a delete. Rewrote it as a post-extraction pass that transfers call records to the surviving Method node before removing the duplicate Function. Also added `cp_interface_method_calls_after_dedup` to guard that property. Both `convergence_probe` and `extraction` suites fully green locally. Thanks for the detailed breakdown, it made the fix path clear.

DeusData:
Thanks for the fast turnaround — this round holds up, and I traced the whole path rather than taking the description at its word.

**What actually fixes the edge is not the part you'd expect.** The `enum_body_declarations` reach-through in `extract_class_methods` gives enum methods a proper class-qualified `Method` definition, and *that* is what re-anchors the records already sitting in those bodies. The call-record rewrite is, for Java specifically, a no-op: the unified walk pushes `SCOPE_CLASS` for `interface_declaration`/`enum_declaration` (`lang_specs.c:320–323`) and class-qualifies the method scope QN, while the class-qualify branch in `extract_func_def` is C++/CUDA-only — so no call record ever carried the spurious `Function`'s QN in the first place. The rewrite is harmless defensive code that would cover a hypothetical language whose walk emits the Function-flavoured QN. Worth knowing so you don't rely on it as the mechanism.

**There's a genuinely nice side effect you didn't claim.** Eight other record kinds carry `enclosing_func_qn` — usages, throws, read/write, type refs, env accesses, type assigns, string refs, channels. For enum method bodies those were **already dangling on main**, because no definition with the class-qualified QN existed at all. Your `Method` emission makes them resolve. So this improves `THROWS`, `WRITES` and usage edges out of enum methods too, not just `CALLS`.

**The new probe binds properly.** `fns == 0` fails without the pass, and `calls >= 1` pins survival of the `App.run → greet` edge at store level — the exact class the earlier CI failure exposed. Survivor is the precise `Method` label with `parent_class` set and full metrics, which is the right side of our label-precision rule.

**Two asks before merge:**

1. **The PR description is now stale.** It still documents the round-1 `push_class_body_children` approach, which commit `11ee4905` reverted — cumulative deletions are zero. Please rewrite it to describe the dedup-pass design you actually shipped, so the merge commit doesn't reference code that isn't there.
2. **One assertion proves less than its name suggests.** In `java_enum_dedup_preserves_calls_issue1234`, `ASSERT_TRUE(r->calls.count >= 1)` passes whether or not the dedup runs — raw call records survive regardless. The binding assertions in that test are the `Method`-existence ones and `Function == 0`. Don't lean on the `calls.count` line as the merge-vs-delete guard; the D6d probe is the real one.

**And one thing that is on me, not you.** My round-1 wording — "the dedup must merge, not delete" — plausibly steered you away from your own commit 1, which was the *root-cause* fix: it stopped the spurious `Function` being minted at all. What ships now mints it for every interface and enum method in every Java file and then deletes it again. Your two commits actually compose: prevention *plus* the `Method` emission, with the dedup pass optionally kept as a safety net. There's a real counter-argument — `interface_body`/`enum_body` are shared node-type names across grammars (TypeScript, Apex, Dart), so prevention carries its own cross-language blast radius — which is exactly why I've put the choice to the maintainer rather than asking you to redo it again on my say-so.

So: **please don't rework anything yet.** Do the two small asks above, and I'll come back with the architecture answer. Nothing you shipped is defective — this is a question about which of two correct shapes we want to live with.

harshitaajoshi:
@DeusData Done! Updated the PR description to document the dedup-pass design and dropped the non-binding `calls.count` assertion from the extraction test. The Method-existence and Function == 0 checks remain as the binding assertions, with the D6d probe covering the CALLS edge property end-to-end. Also changed the PR title to reflect the merge-dedup approach, happy to revert that if you prefer the original. Appreciate the thorough review and the callout on the enum method side effect, that was a good catch.

DeusData:
Both asks done, thank you — and yes, keep the new title. It describes what the diff actually does, which is the only thing a title owes anyone.

Dropping the `calls.count` assertion was the right call rather than the tempting one. An assertion that passes either way is worse than no assertion at all, because it reads like coverage in a review and provides none.

**Nothing further is needed from you right now.** The remaining question is the architecture one I flagged — prevention at the source versus the post-hoc dedup pass — and that is genuinely the maintainer's to answer, not a defect in what you shipped. I have put it to him with both sides written out, including the argument *for* your current shape: `interface_body` and `enum_body` are shared node-type names across TypeScript, Apex and Dart, so restoring the prevention line carries a cross-language blast radius that the dedup pass avoids. That is a real point in its favour, not a consolation.

I will come back to you with the answer. If it goes the prevention way, the work is mostly un-reverting your own commit 1, so nothing you have done is wasted either way.

DeusData:
The maintainer has decided, and the answer is **your original instinct**: prevention at the source, not the post-hoc pass. Which means my round-1 wording sent you the wrong way, and I am sorry for the round trip.

**What we would like, concretely:**

1. **Restore your commit `7f1e80ce` line** — adding `interface_body` / `enum_body` to `push_class_body_children` — **but gated to Java**, e.g. `(strcmp(ck, "interface_body") == 0 && spec->language == CBM_LANG_JAVA)`. The precedent is already in that exact function: the Groovy `closure` case is gated the same way, with the comment *"grammars that also name a node 'closure' are unaffected."*
2. **Keep the `enum_body_declarations` reach-through** in `extract_class_methods`. That is doing real work — it is what mints the class-qualified `Method` whose QN the in-body records already reference, and it is why `THROWS`/`WRITES`/usage edges out of enum method bodies started resolving.
3. **Drop `cbm_dedup_class_method_functions` and its call site.** With prevention in place the duplicate is never created, so there is nothing to delete.
4. **Keep both tests.** `cp_interface_method_calls_after_dedup` still binds — it asserts the surviving node keeps its `CALLS` edge, which is exactly the property prevention has to preserve too. You may want to rename it, since "after_dedup" will no longer describe the mechanism.

**Why the maintainer went this way**, since you put real work into the other shape and deserve the reasoning:

The dedup pass was **unconditional across all 159 languages**. Its job is to delete nodes we should not have created — which means that from the moment it exists, any *future* double-extraction bug in any language gets silently swept up instead of failing a test. That is a permanent reduction in what our tests can catch, in exchange for fixing one language's bug.

The objection I originally raised against prevention — that `interface_body` and `enum_body` are shared node names across grammars (6 and 10 respectively) — was a real concern but the wrong conclusion. It argues for **gating**, not for avoiding prevention. The gating seam was already sitting in that function.

**And your original commit was closer to right than my feedback implied.** "Merge, not delete" was the correct instruction for a dedup pass; it was the wrong instruction because a dedup pass was not the right shape. That distinction is on me, not on you.

Thank you for the patience — you have now turned this around twice within the hour each time, and both rounds improved the change.

harshitaajoshi:
@DeusData Pushed. Restored the prevention line in `push_class_body_children` gated to `CBM_LANG_JAVA`, kept the `enum_body_declarations` reach-through, dropped `cbm_dedup_class_method_functions` and its call site, and renamed the probe to `cp_interface_method_no_dup_function`. Full suite 6792 passed, 0 failed locally. PR description updated to match. Thanks for walking through the architecture decision and for being upfront about the round trip, learned a lot from this one.

DeusData:
Thank you for reworking the Java method identity along the requested architecture and adding the interface/enum regression cases. It is now routed as a high-priority parsing fix in `0.9.1-rc` and is back in the maintainer review queue. The queue is currently full, so the final re-review may take a little time.

DeusData:
Thank you so much for this contribution — and for the exemplary shape of it: exact scope, zero drive-by changes, and honest notes about your local ASan situation. Your diagnosis of #1234 was spot on.

While this PR was open, the routing it patches was reworked on main: 87a0e3f (via #1166) made class-body selection language-agnostic (`find_class_body` / `find_class_member_body` now recognize `interface_body`/`enum_body` for all grammars, and `push_class_body_children` routes through the same selection as method extraction). That supersedes the code half of this PR — the spelling-list block it patches no longer exists — and we verified your scenario is fixed on current main.

Your tests, however, covered a genuine gap: nothing in the suite bound that fix. We've landed them (adapted to current main) in #1433 with you credited as co-author — they pass on main and go RED the moment the routing regresses, so this can't silently break again. That's a lasting contribution to the suite; thank you!

Also: your note about ASan hanging on macOS Tahoe 26.5 / Apple Clang 17 is a useful datapoint we're keeping an eye on — appreciated.

harshitaajoshi:
Makes sense, and glad the routing got fixed properly on main. Thanks for distilling the tests into #1433 and for the co-author credit, I appreciate you taking the time to preserve that half rather than just closing this out. Happy to pick up another parsing or extraction issue if there is one that would be useful.

PR #1329, opened 2026-07-29
fix(extract): derive unique names for GoogleTest macros to prevent same-file node collisions (#1266)

### LINKED ISSUE #1266: C++: same-named TEST cases in one file collapse into a single node, dropping their call edges
## Summary

In C++ projects, multiple GoogleTest `TEST(...)` cases defined in the **same file** collapse into a **single** graph node. Only the last one survives, and the call edges of all the others are dropped.

The practical effect is that `trace_path` / caller analysis silently under-reports test callers: in a test-heavy C++ codebase, most of the tests that exercise a symbol become invisible.

## Environment

- `codebase-memory-mcp` 0.9.0
- macOS 15 (Darwin 25.6.0), arm64
- Indexed with `index_repository(mode="full")`

## Minimal reproduction

Four files, no build required (indexing is static; GoogleTest does not need to be installed).

`src/widget.hpp`
```cpp
#pragma once

namespace demo {

int assembleWidget(int size);

} // namespace demo
```

`src/widget.cpp`
```cpp
#include "widget.hpp"

namespace demo {

int assembleWidget(int size) { return size * 2; }

} // namespace demo
```

`tests/direct_test.cpp` — three test cases in one file, each calling the symbol directly
```cpp
#include "widget.hpp"

#include <gtest/gtest.h>

TEST(WidgetSuite, DoublesSmallSize) { EXPECT_EQ(demo::assembleWidget(1), 2); }

TEST(WidgetSuite, DoublesZero) { EXPECT_EQ(demo::assembleWidget(0), 0); }

TEST(WidgetSuite, DoublesLargeSize) {
  EXPECT_EQ(demo::assembleWidget(1000), 2000);
}
```

`tests/indirect_test.cpp` — one test reaching the symbol through a helper
```cpp
#include "widget.hpp"

#include <gtest/gtest.h>

namespace {

int makeFixture() { return demo::assembleWidget(21); }

} // namespace

TEST(FixtureSuite, UsesHelper) { EXPECT_EQ(makeFixture(), 42); }
```

### Steps

```
index_repository(repo_path="<repro>", name="gtest-caller-repro", mode="full")
search_graph(project="gtest-caller-repro", file_pattern="%direct_test%")
trace_path(project="gtest-caller-repro", function_name="assembleWidget",
           direction="inbound", include_tests=true, depth=3)
```

### Expected

`tests/direct_test.cpp` contributes **three** distinct function nodes (one per test case), each with a `CALLS` edge to `assembleWidget`.

### Actual

`search_graph` returns exactly **one** node for the whole file:

```json
{
  "name": "TEST",
  "qualified_name": "gtest-caller-repro.tests.direct_test.TEST",
  "file_path": "tests/direct_test.cpp",
  "signature": "(WidgetSuite, DoublesLargeSize)",
  "lines": 3,
  "out_degree": 1
}
```

- `signature` is the **last** test in the file — the first two are gone.
- `out_degree` is **1**, so two of the three call sites were dropped.

`trace_path` correspondingly reports one caller for that file instead of three:

```json
{"function":"assembleWidget","direction":"inbound","callers":[
  {"name":"TEST","qualified_name":"gtest-caller-repro.tests.direct_test.TEST","hop":1},
  {"name":"makeFixture","qualified_name":"gtest-caller-repro.tests.indirect_test.makeFixture","hop":1},
  {"name":"TEST","qualified_name":"gtest-caller-repro.tests.indirect_test.TEST","hop":2}
]}
```

## Why this happens (guess)

`TEST(Suite, Case)` parses as a function-like construct whose *name* is `TEST`, so every case in a file maps to the same qualified name `<...>.<file>.TEST` and the nodes overwrite each other. The same applies to `TEST_F`, `TEST_P`, `TYPED_TEST`, and to any macro-defined entry point that repeats a name within a file (e.g. Catch2's `TEST_CASE`, which is name-plus-string).

A fix could derive the qualified name from the macro arguments (`...direct_test.TEST.WidgetSuite.DoublesZero`) or disambiguate by start line.

## Impact

Caller and impact analysis under-count test coverage of a symbol. For C++ repositories where GoogleTest is the norm, a file with N test cases reports 1 caller instead of N, and "which tests exercise this function" — a primary reason to consult the graph before editing — becomes unreliable.

## Secondary observation (not minimally reproducible)

On a large C++ codebase (~129k nodes / 383k edges) I hit a stronger version of this: a free function in a nested namespace is called directly by five same-named `TEST` cases in one file, and `trace_path(direction="inbound", depth=3, include_tests=true)` returns **no** caller from that file at all — only callers from a *different* file. The collapsed node for that file does exist, is reachable via `search_graph`, has `out_degree: 12`, and its body tokens include the callee's name — yet no inbound edge is reported.

I could not reduce this to a minimal case: in the repro above the direct edge is found (including when the call goes through a `using`-declaration with an unqualified call). So it may be a separate resolution issue that only appears at scale, or an additional consequence of the collapsing above. Happy to gather more diagnostics if you can suggest what would be useful — I can query the graph but cannot share that codebase.

harshitaajoshi:
## What does this PR do?

Fixes #1266

Multiple GoogleTest `TEST(...)` cases defined in the same file collapse into a single graph node because `cbm_resolve_func_name` returns the bare macro name `TEST` with no argument-derived identity. Every test case in the file maps to the same qualified name (e.g. `proj.file.TEST`), so the graph upsert keeps only the last one and the call edges of all previous test cases are dropped.

This PR adds a GoogleTest macro detection step in `extract_func_def` that composes a unique name from the two macro arguments: `TEST(WidgetSuite, DoublesSmallSize)` becomes `TEST_WidgetSuite_DoublesSmallSize`. This gives each test case its own qualified name, preventing node collisions. The fix covers `TEST`, `TEST_F`, `TEST_P`, `TYPED_TEST`, and `TYPED_TEST_P`. It also sets `is_test = true` for these nodes so test exclusion queries work correctly.

**Changes:**
- `internal/cbm/extract_defs.c`: Add `is_cpp_test_macro()` and `resolve_cpp_test_macro_name()` helpers; call them in `extract_func_def` before QN computation; set `is_test` flag
- `tests/test_extraction.c`: Two regression tests using the exact fixture from the issue (`TEST()` three-case collision and `TEST_F()` two-case collision)

## Checklist

- [x] Every commit is signed off (`git commit -s`) -- required, CI rejects unsigned commits ([DCO](../DCO), see [CONTRIBUTING.md](../CONTRIBUTING.md))
  _ASan test-runner hangs on startup on macOS Tahoe (Apple Clang 17), a pre-existing environment issue unrelated to this change._

Note: the local test checkbox is unchecked because the ASan test-runner hangs on startup for all suites locally (pre-existing environment issue unrelated to this change). The production build compiles cleanly with `-Werror` and clang-format passes on all changed files. CI will run the full test suite.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -3196,6 +3196,54 @@ static char *go_receiver_type_name(CBMArena *a, TSNode recv, const char *source)
     return NULL;
 }
 
+/* C++/CUDA: true when name is a GoogleTest test-definition macro whose
+ * invocations parse as function definitions with bare-identifier parameters.
+ * Multiple such macros in one file all share the extracted name (e.g. "TEST"),
+ * causing qualified-name collisions and node loss (#1266). */
+static bool is_cpp_test_macro(const char *name) {
+    return strcmp(name, "TEST") == 0 || strcmp(name, "TEST_F") == 0 ||
+           strcmp(name, "TEST_P") == 0 || strcmp(name, "TYPED_TEST") == 0 ||
+           strcmp(name, "TYPED_TEST_P") == 0;
+}
+
+/* Compose a unique name from a GoogleTest-style macro invocation by appending
+ * the macro arguments: TEST(Suite, Case) -> "TEST_Suite_Case".
+ * Returns an arena-allocated string, or NULL when the parameters cannot be
+ * resolved (caller keeps the original name in that case). */
+static char *resolve_cpp_test_macro_name(CBMArena *a, const char *macro, TSNode node,
+                                         const char *source) {
+    TSNode params = ts_node_child_by_field_name(node, TS_FIELD("parameters"));
+    if (ts_node_is_null(params)) {
+        params = find_c_params(node);
+    }
+    if (ts_node_is_null(params)) {
+        return NULL;
+    }
+
+    const char *args[2] = {NULL, NULL};
+    int count = 0;
+    uint32_t nc = ts_node_named_child_count(params);
+    for (uint32_t i = 0; i < nc && count < 2; i++) {
+        TSNode child = ts_node_named_child(params, i);
+        if (ts_node_is_null(child)) {
+            continue;
+        }
+        TSNode type_node = ts_node_child_by_field_name(child, TS_FIELD("type"));
+        char *text = cbm_node_text(a, ts_node_is_null(type_node) ? child : type_node, source);
+        if (text && text[0]) {
+            args[count++] = text;
+        }
+    }
+
+    if (count == 2) {
+        return cbm_arena_sprintf(a, "%s_%s_%s", macro, args[0], args[1]);
+    }
+    if (count == 1) {
+        return cbm_arena_sprintf(a, "%s_%s", macro, args[0]);
+    }
+    return NULL;
+}
+
 static void extract_func_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec) {
     CBMArena *a = ctx->arena;
 
@@ -3215,6 +3263,20 @@ static void extract_func_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec 
     // any dot-prefixed Make target.
     if (ctx->language == CBM_LANG_MAKEFILE && name[0] == '.') {
         return;
+    }
+
+    /* C++/CUDA: GoogleTest macros (TEST, TEST_F, TEST_P, ...) parse as
+     * function definitions whose name resolves to the bare macro identifier.
+     * Multiple test cases per file collide on qualified name; derive a unique
+     * name from the macro arguments so each gets its own graph node (#1266). */
+    bool is_gtest = false;
+    if ((ctx->language == CBM_LANG_CPP || ctx->language == CBM_LANG_CUDA) &&
+        is_cpp_test_macro(name)) {
+        char *gtest_name = resolve_cpp_test_macro_name(a, name, node, ctx->source);
+        if (gtest_name) {
+            name = gtest_name;
+            is_gtest = true;
+        }
     }
 
     TSNode func_node = unwrap_template_inner(node, ctx->language);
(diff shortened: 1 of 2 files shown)


DeusData:
Merged — thank you, this closes #1266 properly.

The failure mode you fixed is a nasty one: `TEST(Suite, Case)` parses as a `function_definition` whose name is literally `TEST`, so every GoogleTest case in a file collided on the same qualified name and the upsert kept only the last one. The earlier cases did not merely get mislabelled — they disappeared, and their `CALLS` edges went with them. For a C++ codebase with a real test suite that is a large, silent hole in the graph.

What I liked about the implementation:

- You compose the name from the two macro arguments (`TEST_Suite_Case`), which means the derived name is **stable** — it depends only on the source text, not on iteration order, a counter, or a byte offset. An unstable name would have churned the graph on every reindex and quietly broken incremental diffing, and that trap is easy to fall into here.
- The result stays human-findable. `TEST_WidgetSuite_DoublesSmallSize` is something a person or an agent can actually search for, which matters a lot to us.
- It is NULL-safe: unresolvable parameters fall back to the original name rather than producing something malformed.
- Coverage across `TEST` / `TEST_F` / `TEST_P` / `TYPED_TEST` / `TYPED_TEST_P`, and `is_test` set consistently with the existing Rust convention rather than a new one.
- Both polarities are asserted in the test — the new distinct node must exist *and* the old colliding `TEST` node must not.

One entirely optional follow-up, not a blocker and not something I want you to change now: a genuine C++ function literally named `TEST`/`TEST_F` with one or two typed parameters would also be renamed and flagged `is_test`. It is vanishingly rare and strictly bounded to those five names. If you ever want to close it, requiring the parameter shape to be bare identifiers (a `parameter_declaration` with a type but no declarator) before treating it as a macro would eliminate the surface entirely.

Thanks for the thorough work.

harshitaajoshi:
Thank you for the detailed review. The stability point about incremental diffing is something I will keep in mind for future contributions. I have noted the optional follow-up on tightening the parameter shape check and may pick that up separately.

PR #1332, opened 2026-07-29
refactor(cli): move self-update to the install scripts; strip Mach-O fully

DeusData:
Two changes, both grounded in measurement rather than inference.

## 1. Strip Mach-O fully

The first strip landed as `strip -x` on Mach-O, out of caution that a full strip
could leave an image dyld will not load. Wrong for this binary, and it cost a
release cycle.

`-x` retains external symbols — **4058** of them — so the macOS artifacts kept
the symbol table the ELF legs had just shed. Run 30414312364 then flagged
exactly those two artifacts while every Linux and Windows binary came back
clean. That read as the detection "moving to macOS"; it had simply stayed on the
unstripped set.

| darwin-arm64 | symbols | VirusTotal |
|---|---|---|
| `strip -x` (shipped) | 4058 | 1 malicious / 61 |
| `--strip-all` | 373 | **0 / 61 CLEAN** |

Packaged binary still passes `codesign --verify --strict` (re-signed after
stripping) and still runs. Symbol removal has now cleared the detection on
**three independent artifacts across two platforms**.

## 2. Self-update out of the product, everywhere

Windows already handed updating to `install.ps1` because a running `.exe` cannot
replace itself. The other platforms now do the same for a better reason: an
in-process updater is structurally a downloader — fetch, verify, extract,
`chmod +x`, exec — and that composite shipped in every artifact, with the
release URLs it fetched, to serve a command most users run a handful of times.

`update` now prints the exact command and exits 0. Both install scripts ship
inside their archive, so the printed path resolves next to the binary. Flags are
still parsed, so `update --dry-run --ui` keeps rejecting typos.

The updater is **excluded from release builds** rather than left to dead-code
elimination — `build_update_url`, `download_verify_install`,
`extract_and_install_binary`, checksum fetch, `detect_os/arch` and the
`cli_download_*` helpers compile only under `CBM_CLI_ENABLE_TEST_API`. Verified
against the release binary: `releases/latest/download` is **gone**.

Also removes the MCP background update check, which ran on the first tool call:

```c
cbm_popen("curl -sf --max-time 5 -H 'Accept: application/vnd.github+json' "
          "'https://api.github.com/repos/.../releases/latest' 2>/dev/null")
```

An embedded shell command plus a startup network callout in every agent session,
to report that a newer version exists. It shelled out and depended on `curl`
being present. The install scripts report versions.

## Smoke

Phases 6c and 14 collapse to **one platform-neutral contract** selected by
`UPDATE_SCRIPT`. Two things worth calling out:

- **14a now compares the driver against itself** before and after. Comparing
  against `$BINARY` passed on Windows only because Windows has no signing step —
  the POSIX fixture ad-hoc re-signs its copy, so the two differed before `update`
  ever ran.
- **The Linux glibc guard is relocated, not retired.** The standard linux asset
  links glibc 2.38+ and breaks Debian 11 / RHEL 8 / Ubuntu 20.04, so the
  installer must fetch the static `-portable` build. That constraint moved from
  the binary into `install.sh`; the smoke now guards it where the behaviour lives.

## Verification

| Check | Result |
|---|---|
| `cli` `mcp` `daemon_application` `activation_transaction` | **500 passed, 2 skipped** |
| Full macOS smoke (no-self-update binary) | **ALL PASSED** |
| 10 shell contracts | pass |
| `lint-ci` (cppcheck + clang-format) | clean |
| macOS packaged binary | 368 symbols, `codesign --verify --strict` ok, runs |

Note on the cppcheck fix: the always-true `!update_seam_portable` was resolved by
**restructuring** the handoff, not by suppressing the warning.

### DIFF
--- a/scripts/package-release.sh
+++ b/scripts/package-release.sh
@@ -113,17 +113,21 @@ NAME="codebase-memory-mcp${SUFFIX}-${GOOS}-${GOARCH}"
 strip_release_binary() {
     local binary="$1"
     [ -f "$binary" ] || return 0
+    # --strip-all on every format, Mach-O included.
+    #
+    # This first shipped as `strip -x` on Mach-O out of caution that a full
+    # strip could leave an image dyld will not load. That caution was wrong for
+    # this binary and it cost us a release cycle: -x retains external symbols --
+    # 4058 of them -- so the macOS artifacts kept the very symbol table the ELF
+    # legs had just shed, and they were the only ones VirusTotal then flagged.
+    # Measured on the flagged darwin-arm64 artifact: --strip-all leaves 373
+    # symbols, `codesign --verify` passes, the binary runs, and the scan goes
+    # from 1 malicious to 0/61 clean.
     local stripped=""
     for tool in "${STRIP:-}" llvm-strip strip; do
         [ -n "$tool" ] || continue
         command -v "$tool" >/dev/null 2>&1 || continue
-        if [ "$GOOS" = "darwin" ]; then
-            # -x keeps external symbols: a full strip of a Mach-O can leave an
-            # image dyld will not load.
-            "$tool" -x "$binary" 2>/dev/null && stripped="$tool"
-        else
-            "$tool" --strip-all "$binary" 2>/dev/null && stripped="$tool"
-        fi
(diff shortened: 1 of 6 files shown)


adfjadfj16-a11y:
Esto para que es y de que sirve me puedes explicar porfavor

Issue #1335, opened 2026-07-29
hook-augment: needs a warm daemon to ever emit context; install leaves it timing out on every Grep/Glob (follow-up to #858)


b1rdex:
**Version:** `v0.9.1-rc.1` (built from the tag, `560ad40`); same behaviour with the `v0.9.0` release binary
**Platform:** macOS 15 (Darwin 24.6.0), arm64, 16 GB RAM
**Client:** Claude Code, hooks installed by `codebase-memory-mcp install`
**Graph:** 24146 nodes / 79897 edges, 90 MB SQLite DB (Sylius 2.0 / PHP 8.3 repo, 2984 files)

Follow-up to the closed #858 ("hook-augment additionalContext never appears in real sessions (0/24 observed) — suspect HA_DEADLINE_MS (300ms) too tight for real cold-start conditions"). The deadline is 2000 ms now and the handler does work — but **only if a CBM daemon is already warm**, and the margin is about 20 ms. `install` does not ensure a warm daemon, so out of the box the hook times out on every single `Grep`/`Glob`.

## Measured: no warm daemon → 0 injections, every time

```bash
echo '{"cwd":"/path/to/repo","hook_event_name":"PreToolUse","tool_name":"Grep",
       "tool_input":{"pattern":"PosthogTracker"}}' | codebase-memory-mcp hook-augment
```

8 invocations, 8 × 0 bytes on stdout, exit 0. `~/.cache/codebase-memory-mcp/logs/hook-augment-timeouts.log` gains one line each time:

```
hook-augment: deadline_exceeded ms=2000 pid=32400 (raise via CBM_HOOK_DEADLINE_MS)
hook-augment: deadline_exceeded ms=2000 pid=32409 (raise via CBM_HOOK_DEADLINE_MS)
hook-augment: deadline_exceeded ms=2000 pid=32826 (raise via CBM_HOOK_DEADLINE_MS)
hook-augment: deadline_exceeded ms=2000 pid=32964 (raise via CBM_HOOK_DEADLINE_MS)
```

Curiously, raising the budget did **not** produce output either — with `CBM_HOOK_DEADLINE_MS=20000` the timeout lines stop, but 5 consecutive runs still returned 0 bytes at ~2.26–2.30 s each. So a cold invocation appears to fail for a second reason beyond the deadline (I did not chase it into `ha_process`).

## Measured: warm daemon → works, at ~1.98 s against a 2.00 s deadline

```bash
[...]

PR #1337, opened 2026-07-29
feat(hook-augment): augment Bash tool searches via graph

### LINKED ISSUE #1082: Claude Code Bash searches bypass codebase-memory hook augmentation
### Reproduction

With codebase-memory-mcp 0.9.0, invoke `hook-augment` with:

```json
{"tool_name":"Grep","tool_input":{"pattern":"CreateStripeCheckout"}}
```

It returns four graph symbol matches.

Invoke it with the equivalent Claude Code Bash payload:

```json
{"tool_name":"Bash","tool_input":{"command":"rg -n CreateStripeCheckout ."}}
```

It returns no graph context.

Claude Code commonly runs `rg`, `grep`, and `find` through the `Bash` tool, so code discovery silently bypasses graph augmentation when the PreToolUse matcher only covers `Grep|Glob`.

### Expected behavior

Bash-based code searches should receive the same graph augmentation as native `Grep` searches, or the installer should provide a reliable Claude Code integration that detects these searches.

### Workaround

A Bash PreToolUse adapter that extracts the search pattern, reshapes the payload as `Grep`, and forwards it to `hook-augment` restores the expected context.

This is an integration/installer issue: the MCP graph query itself works, but Claude Code's common Bash search path bypasses it.

abendrothj:
Fixes #1082.

## What

Claude Code frequently runs searches through the `Bash` tool (`rg -n Foo .`, `grep -rn Foo .`, etc.) rather than the native `Grep`/`Glob` tools. Because `hook-augment`'s `PreToolUse` matcher only covered `Grep` and `Glob`, these searches silently bypassed graph augmentation.

This PR adds detection directly in `hook_augment` itself (Q1 from my earlier comment — I went with the former approach since it reuses the existing `search_graph` path without duplicating resolver logic). Pattern extraction happens via a small shell tokenizer and flag parser; the result feeds into the same `ha_extract_token` → `search_graph` path Grep already uses.

## Approach

**`ha_tokenize`** — minimal shell tokenizer: whitespace split with `'…'`/`"…"` quoting and backslash handling. No evaluation, no expansion.

**`ha_parse_bash_search_pattern`** — pure string parser, never touches the command:
- Strips leading `VAR=value` env assignments
- Strips wrappers: `env`, `nice`, `time`, `command`, `rtk`, `tokf run`
- Identifies the binary (`grep`/`egrep`/`fgrep`, `rg`, `ag`, `ack`, `ugrep`/`ug`, `git grep`)
- Walks flags: bails on `-f`/`--file` (pattern file), bails on multiple `-e`, handles embedded (`-ePATTERN`) and space-separated (`-e PATTERN`) forms
- Per-binary value-taking flag tables so e.g. `grep -A 5 PATTERN` correctly skips `5`
- `rtk grep -l <N>`: treats `-l` as value-taking (consumes `N`, arrives at `PATTERN`)

**`ha_tool_event_supported`** — `Bash` added alongside `Grep`/`Glob` for `PreToolUse` in `HA_DIALECT_EVENT`.

**`ha_process`** — Bash branch extracts pattern from `tool_input.command`; everything after is identical to the Grep path.

## Scope (Q2)

grep/egrep/fgrep, rg, ag, ack, ugrep/ug, git grep — plus the wrapper handling dergachoff noted in the thread.

## Testing

`cli_hook_augment_bash_pattern_extractor` in `tests/test_cli.c` covers: common invocation forms, value-taking flag skipping, env-var prefixes, all wrappers, the `rtk grep -l <N>` case, bail-out cases (`-f`, multiple `-e`, non-search binary, empty/NULL), and `--` end-of-flags. Verified clean compile and production binary build on macOS; full test run blocked by an ASan/dyld hang on macOS 26 pre-release — CI should cover it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,7 +441,7 @@ overwrite user-modified agents.
 
 | Agent | Activation | MCP config | Durable context / augmentation |
 |-------|------------|------------|--------------------------------|
-| Claude Code | Detected | `~/.claude.json` | Skill + three exact-tool graph agents; `SessionStart`, `SubagentStart`, non-blocking `PreToolUse` for `Grep`/`Glob`, and post-`Read` coverage |
+| Claude Code | Detected | `~/.claude.json` | Skill + three exact-tool graph agents; `SessionStart`, `SubagentStart`, non-blocking `PreToolUse` for `Grep`/`Glob`/`Bash`, and post-`Read` coverage |
 | Codex CLI | Detected | `$CODEX_HOME/config.toml` | `AGENTS.md`, skill, three read-only agents; `SessionStart` + `SubagentStart` |
 | Gemini CLI | Detected | `.gemini/settings.json` | `GEMINI.md`, three explicit read/graph-tool subagents; `BeforeTool`, `AfterTool` `read_file` coverage, and `SessionStart` |
 | Zed | Detected | platform `settings.json` (JSONC) | `AGENTS.md` + shared skill |
@@ -488,7 +488,7 @@ overwrite user-modified agents.
 ### Sessions, compaction, and subagents
 
 Hooks installed by this project are fail-open and context-only. Claude Code's
-`PreToolUse` observes `Grep`/`Glob` and injects matching graph symbols as
+`PreToolUse` observes `Grep`/`Glob`/`Bash` and injects matching graph symbols as
 `additionalContext`; `PostToolUse` on `Read` adds targeted coverage context when
 the graph could not fully parse or index that file. It never denies or replaces
 the requested tool call.
--- a/scripts/smoke-test.sh
+++ b/scripts/smoke-test.sh
@@ -1415,7 +1415,7 @@ fi
 echo "OK 8c-i: Claude exact-tool graph subagent"
 
 # 8d: Claude Code hooks keep search augmentation and read-coverage reporting
-# separate: PreToolUse matches exactly Grep|Glob, while PostToolUse matches
+# separate: PreToolUse matches exactly Grep|Glob|Bash, while PostToolUse matches
 # exactly Read. Neither hook may grow a Search or catch-all matcher.
 if ! cat "$FAKE_HOME/.claude/settings.json" 2>/dev/null | python3 -c "
 import json, sys
@@ -1423,7 +1423,7 @@ d = json.load(sys.stdin)
 all_hooks = d.get('hooks', {})
 pre = all_hooks.get('PreToolUse', [])
 post = all_hooks.get('PostToolUse', [])
-ok = (any(h.get('matcher') == 'Grep|Glob' for h in pre) and
+ok = (any(h.get('matcher') == 'Grep|Glob|Bash' for h in pre) and
       any(h.get('matcher') == 'Read' for h in post))
 bad = any('Search' in str(h.get('matcher', '')) for h in pre + post)
 sys.exit(0 if (ok and not bad) else 1)
@@ -1431,7 +1431,7 @@ sys.exit(0 if (ok and not bad) else 1)
   echo "FAIL 8d: Claude search/read hook matchers are not exact"
   exit 1
 fi
-echo "OK 8d: Claude Code PreToolUse Grep|Glob + PostToolUse Read"
+echo "OK 8d: Claude Code PreToolUse Grep|Glob|Bash + PostToolUse Read"
 
 # 8e: Claude Code shim script — must be non-blocking augmenter, not a gate.
 # #929: Windows installs a .cmd script (extensionless bash shims triggered the
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -3819,9 +3819,9 @@ static int cbm_remove_vibe_mcp_owned(const char *binary_path, const char *config
 
 /* ── Claude Code pre-tool hooks ───────────────────────────────── */
 
-/* Search augmentation runs before Grep/Glob; exact coverage context runs after
+/* Search augmentation runs before Grep/Glob/Bash; exact coverage context runs after
  * Read. Both adapters are context-only and fail open. */
-#define CMM_HOOK_SEARCH_MATCHER "Grep|Glob"
+#define CMM_HOOK_SEARCH_MATCHER "Grep|Glob|Bash"
 #define CMM_HOOK_READ_MATCHER "Read"
 /* Basename only; the full command path is resolved at install time via
  * cbm_resolve_hook_command so $CLAUDE_CONFIG_DIR is honored. */
@@ -3845,6 +3845,7 @@ static int cbm_remove_vibe_mcp_owned(const char *binary_path, const char *config
 static const char *const cmm_claude_old_matchers[] = {
     "Grep|Glob|Read|Search",
     "Grep|Glob|Read",
+    "Grep|Glob",
     NULL,
 };
 static const char *const cmm_gemini_old_matchers[] = {
@@ -4961,7 +4962,7 @@ static int cbm_remove_owned_hook_script(const char *path, const char *expected_c
 
 /* Install the search-augmenter shim to ~/.claude/hooks/.
  * The shim is a thin wrapper that delegates to `<binary> hook-augment`,
- * which adds graph context to Grep/Glob calls. It NEVER blocks a tool call:
+ * which adds graph context to Grep/Glob/Bash search calls. It NEVER blocks a tool call:
  * a missing/old/hung binary results in a silent exit 0 (issue #362/#288).
  * The legacy filename `cbm-code-discovery-gate` is retained so existing
  * settings.json entries and uninstall keep working with zero migration. */
@@ -7204,7 +7205,7 @@ static void install_claude_code_config(const char *home, const char *binary_path
         }
     }
     if (gate_ok) {
-        printf("  hooks: PreToolUse Grep/Glob search augmentation + PostToolUse Read coverage "
+        printf("  hooks: PreToolUse Grep/Glob/Bash search augmentation + PostToolUse Read coverage "
                "(non-blocking)\n");
     }
     if (session_ok) {
--- a/src/cli/cli.h
+++ b/src/cli/cli.h
@@ -220,6 +220,7 @@ bool cbm_hook_augment_invocation_supported_for_testing(const char *dialect,
 bool cbm_hook_path_contains_for_testing(const char *root, const char *candidate,
                                         bool case_insensitive);
 const char *cbm_hook_no_project_index_guidance_for_testing(const char *event);
+bool cbm_hook_augment_parse_bash_pattern_for_testing(const char *cmd, char *out, size_t out_sz);
 bool cbm_mcp_command_path_probe_safe_for_testing(const char *command, bool windows);
 void cbm_set_mcp_command_path_probe_counter_for_testing(int *counter);
 int cbm_install_editor_mcp_with_previous_for_testing(const char *binary_path,
--- a/src/cli/cli.h
+++ b/src/cli/cli.h
@@ -477,6 +477,10 @@ char *cbm_hook_augment_lifecycle_json_for(const char *input, const char *forced_
  * hard fail-open deadline without constructing a local MCP/store instance. */
 void cbm_hook_augment_arm_deadline(void);
 char *cbm_hook_augment_read_stdin(void);
+/* Pure no-op gate for the hook-client fast path (see hook_augment.c). */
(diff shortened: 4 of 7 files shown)


abendrothj:
The cppcheck failure is pre-existing in src/main.c lines 283 and 2049; introduced by a recent merge to main, not by this PR.

DeusData:
Reviewed in full. The parser is well built and security-clean — but there is one finding that matters more than the rest: **as shipped, this feature never runs.**

**The hook matcher is never updated to include Bash.** The install path's hook registration still matches only the existing tools, so `hook-augment` is never invoked for a Bash call. Everything in this PR — the command parsing, the graph lookup, the augmentation — is unreachable in a real installation. Your tests exercise the parser directly, which is why they pass and why the gap is invisible from CI.

That is worth knowing before you invest more: the code is not wrong, it is simply not wired in.

**And wiring it up is the part that needs a maintainer decision**, which is why I have not just asked you to add the matcher. Registering Bash means the hook fires on **every Bash tool call an agent makes** — not just searches. That is a per-call cost on the hottest path an agent has, and it is a deliberate trade rather than an implementation detail. Our existing hooks are scoped to Grep and Glob precisely because those are search operations; Bash is everything.

**One thing to correct regardless:** the maintainer's recorded requirement was that `rtk grep -l 80` be a **negative** test — i.e. it must *not* be treated as a search — and your implementation treats it as one. Worth re-reading that thread before revising, since it points at where the parser's boundary is meant to sit.

**Your CI failure is not yours.** The `lint` job failed on cppcheck `knownConditionTrueFalse` findings in `src/main.c` — and this PR touches only `src/cli/cli.h`, `src/cli/hook_augment.c` and `tests/test_cli.c`. I checked the file list explicitly. A rebase onto current main should clear it; `test` was auto-skipped as a consequence of lint, and `ci-ok` is just the aggregate.

**What is good here**, and worth saying: the command parser is careful — it is a real tokeniser rather than a regex, it handles quoting, and it fails closed on shapes it does not recognise rather than guessing. Nothing shells out, and there is no unquoted interpolation anywhere. That is the right foundation.

My suggestion: hold off on further work until the maintainer answers whether we want graph augmentation on every Bash call at all. If the answer is yes, the missing matcher registration and the `rtk grep -l 80` case are both small; if it is no, you will have saved yourself a rebase.

abendrothj:
Fixed `rtk grep -l` bail out, test flipped to `ASSERT_FALSE`. That one was right.

The Bash matcher finding is not. `ha_tool_event_supported` has included `Bash` since the first commit on this branch. Please read the diff before requesting changes.

>And wiring it up is the part that needs a maintainer decision, which is why I have not just asked you to add the matcher. Registering Bash means the hook fires on every Bash tool call an agent makes — not just searches. That is a per-call cost on the hottest path an agent has, and it is a deliberate trade rather than an implementation detail. Our existing hooks are scoped to Grep and Glob precisely because those are search operations; Bash is everything.

On the per-call cost question, that is a design decision only you can make. Happy to discuss tradeoffs, but I cannot wait on you to consult yourself.

DeusData:
Thank you for correcting the review. You are right: the PR patch has included `Bash` in `ha_tool_event_supported` from the first commit, and `ha_process` routes Bash commands through the new parser before the normal search path. My statement that the feature was unreachable was wrong. I am sorry for making you spend time correcting a finding that the diff did not support.

The `rtk grep -l` negative is now corrected as well. The remaining question is narrower and genuinely about product cost: registering Bash invokes the hook for every Bash call even though non-search commands then fail closed quickly. With the open hook deadline and silent-output reports in #1335 and #1388, this stays in `0.9.2-rc` while we decide default versus opt-in behavior and measure the real no-op cost. No further rework is requested from you while that direction is pending. Thank you for the careful parser and for pushing back with the exact code reference.

abendrothj (commented):
Acknowledged. The parser implementation and the `rtk grep -l <N>` negative case are corrected.

I understand the remaining decision is whether Bash augmentation should be enabled by default or remain opt-in because the hook would run for every Bash call. I’ll leave the PR unchanged in its current `0.9.2-rc` state while that product decision is made, and won’t do further rework unless requested.

DeusData:
Thank you for this, and I'm sorry — this has been sitting for over two weeks blocked on something that has nothing to do with your code.

The only failing check is `security / license-gate`, and the failure is:

```
vendored/nomic: DIFFERS [apache.org canonical LICENSE-2.0.txt]
PROVENANCE AUDIT FAILED: 1 unexplained verdict(s)
```

Your PR touches `src/cli/cli.h`, `src/cli/hook_augment.c` and `tests/test_cli.c`. It could not have caused that.

I checked it just now rather than guessing: our vendored `vendored/nomic/LICENSE` is **byte-identical** to the canonical Apache-2.0 text — 11358 bytes on both sides, `diff` clean. So the verdict was a transient failure, not a real licence discrepancy.

The cause is our gate, not your branch. `scripts/audit-license-provenance.py` fetches `https://www.apache.org/licenses/LICENSE-2.0.txt` with `curl` **at gate time** and byte-compares. If that fetch fails for any reason, the captured output is empty, the comparison fails, and it reports `DIFFERS` — indistinguishable from an actual licence change. A gating verdict that depends on a live HTTP request is a defect on our side, and I'm filing it: the canonical text should be compared against a vendored copy, and a fetch failure must report as a fetch failure rather than as a content difference.

I've re-run the checks on your branch. Sorry again for the wait — being stuck this long on an unrelated infrastructure flake, with no explanation, is a bad experience and it was avoidable.

abendrothj:
Thanks for fixing and re-running the CI checks—the PR is now green.

The implementation and negative parser cases are complete. The remaining decision is whether Bash augmentation should be:

1. enabled by default by registering `Bash` in the installed `PreToolUse` matcher, accepting the lightweight fail-closed path for non-search commands; or
2. kept opt-in through installer/configuration.

Could you confirm which direction you want for `0.9.2-rc`? Once confirmed, I can make the final wiring change if needed.

DeusData:
**Decision: on by default. Please register `Bash` in the installed matcher and I will merge.**

Before that, a correction I owe you — my retraction on 3 August went further than the evidence supported, and the result was that we both walked away thinking the question was settled when it was not.

You were right: `ha_tool_event_supported` has accepted `Bash` since your first commit, and `ha_process` routes it to the new parser. I was wrong to say the in-binary guard was missing it, and I apologised for that correctly.

But my original concern was about a different layer, and **that part was never actually refuted**. What we write into the user's `settings.json` is:

```c
/* src/cli/cli.c:4069 */
#define CMM_HOOK_SEARCH_MATCHER "Grep|Glob"
```

used at `cli.c:4794` and `:4835`. This PR touches `cli.h`, `hook_augment.c` and `tests/test_cli.c` — not `cli.c`. So on a real installation the hook binary is still never invoked for a `Bash` tool call, and the parser you wrote is unreachable. Two layers, two correct statements, and I collapsed them into one retraction. That is on me, and it cost you three weeks of the PR sitting in a state neither of us had described accurately.

Your own closing comment is consistent with this — you offered to "make the final wiring change if needed", which is exactly the change that is still outstanding.

**So: option 1. Add `Bash` to `CMM_HOOK_SEARCH_MATCHER`.**

The reasoning, since you deserve it rather than just a verdict. Registering `Bash` means a hook process spawn on every Bash call, not only search-shaped ones, and Bash is the agent's hottest tool. That is a real cost. I am taking it anyway because this is the highest-value augmentation point in the product: the moment an agent reaches for `rg` is precisely the moment it is about to do by hand what the graph already knows, and that is the flailing this project exists to prevent. Augmenting *after* the search has run would be too late; augmenting only `Grep`/`Glob` misses every agent that shells out instead. A hook that fires only where it is already unnecessary is not worth having.

I also considered your option 2 and rejected it on grounds worth stating: there is no opt-in configuration surface for hook matchers today, so "keep it opt-in" would mean building a new permanent config key *and* a conditional write in the install path — and then shipping a feature almost nobody would find. We have been bitten by exactly that: a contributor recently rebuilt an existing feature from scratch because the switch that already did the job was undiscoverable. More surface, less use, same maintenance.

**Two things to include with the matcher change:**

1. **A test that the wiring is reachable.** Your parser tests are thorough — the `rtk grep -l 80` negative case is exactly right — but they call the seam directly, which is *why* CI could not see this gap and why we spent three weeks talking past each other. Please add something that pins `ha_tool_event_supported` accepting a `PreToolUse` + `Bash` event, and ideally that the installed matcher string contains `Bash`. A test at the layer where the bug lived.
2. **`ha_search_bin_val_flags` takes an unused `rtk_grep` parameter** (the rtk case is handled inline in the caller) — likely to trip `-Wunused-parameter`. Worth dropping while you are in there.

One non-blocking note: `ha_tokenize` truncates at 32 tokens and 256 bytes per token, so a very long command degrades to a wrong-or-absent pattern rather than an explicit bail-out. It fails toward no augmentation, so it is safe — I mention it only because "silently wrong pattern" and "no pattern" are worth distinguishing if this ever misbehaves in the field.

Thank you for the parser itself, which is genuinely careful work: a real tokeniser rather than a regex, correct handling of value-taking flags, env prefixes and wrapper binaries, and it never shells out or rewrites the command. And thank you for pushing back on 31 July instead of accepting a wrong finding — you were right to, and the fact that I then over-corrected does not change that.

abendrothj:
Implemented the requested wiring: Bash is now registered in the installed `PreToolUse` matcher, with regression coverage for the real `PreToolUse` + Bash graph path. I also removed the unused `rtk_grep` parameter and added migration coverage for existing `Grep|Glob` installations. Focused hook tests and the production build pass locally.

DeusData:
@abendrothj — one maintainer commit added on top of your branch before the merge, closing a cost question from my own 3 August comment that should have been answered before the default-on ruling, not after.

**What I measured.** On a production binary, *every* hook invocation was paying ~1.1–1.2 s of user CPU before any gating — for `ls -la`, for garbage stdin, for anything. The first suspect was `cbm_mcp_server_new`, but reordering inside `cbm_cmd_hook_augment` moved nothing: the cost turned out to live in `main()`'s executable-identity hash for cohort admission, which the hook-client path runs before it ever looks at stdin. Your parsing was never the problem — it costs microseconds. But with Bash in the installed matcher, nearly every command an agent runs would have paid that 1.2 s for nothing.

**What the commit does.** `main()` now reads stdin on the hook-client path and exits before identity hashing when the event can never produce output — an un-forced PreToolUse Bash event whose command yields no search pattern or no queryable token. A no-op touches no shared state, so it needs no identity. The gate is a small exported function built on *your* parser pair (`ha_parse_bash_search_pattern` + `ha_extract_token`), so the gate and the augmenter cannot disagree about what counts as a search; anything uncertain — lifecycle events, other tools, the coverage adapters, oversized stdin, forced events — pays full fare exactly as before. The consumed stdin is handed back through a prefetch so every downstream reader sees it once.

**Measured after:** no-op Bash calls ~10 ms wall / 0.00 s user CPU (from ~1.2 s); real Bash searches and Grep/Glob unchanged. `cli` 259/0, `mcp` 179/0 locally.

Your PR itself was merge-ready against everything asked of it — the matcher registration, the reachability test at the exact layer where the earlier confusion lived, and the `rtk grep -l 80` negative case are all as ruled. The e2e test you wrote is what gives me confidence the full-fare path still works after this reordering. CI runs now; I merge on green, `--merge` as always.

And credit where due: on Claude Code ≥2.1.117, where Grep/Glob no longer exist natively, your work is the only way search augmentation fires at all. That argument — which you made — is what carried default-on.

DeusData:
Merged in d706c33a. Thank you, @abendrothj — Bash search augmentation ships default-on, and on Claude Code ≥2.1.117, where Grep/Glob no longer exist natively, your work is now the only way search augmentation fires at all. That argument was yours, and it carried the decision.

The full arc, for the record: a review that wrongly claimed the feature never ran, your correct refutation, the retraction, the real gap found at the *installed matcher* layer, your same-day fix with a reachability test at exactly the layer where the confusion lived, and the `rtk grep -l 80` negative case from the issue thread — all in. The maintainer perf commit on top closes the cost question that should have been settled before the default-on ruling: no-op Bash calls now bail before executable-identity hashing at ~10 ms instead of ~1.2 s, gated on your own parser pair so the gate and the augmenter cannot disagree.

The one red on the way in was the CodeQL-gate 45-minute timeout with CodeQL itself green — our documented infrastructure pattern, not your code.

Fixes #1082 — closing it with this SHA. A textbook contribution thread, including the part where you were right and the review wasn't.

PR #1368, opened 2026-07-30
fix(similarity): skip empty fingerprint sort

### LINKED ISSUE #1367: pass_similarity calls qsort with a null base for an empty fingerprint set
### What happens

`collect_fp_entries()` starts with `entries == NULL` and calls
`qsort(entries, count, ...)` unconditionally. When a graph buffer has no
Function/Method nodes with fingerprints, `count == 0`, so the call is
`qsort(NULL, 0, ...)`.

The project targets C standards where the sorting API is specified in terms of
`base` pointing at an array. Zero elements suppress comparator calls, but do not
make a null base a portable array pointer. Sorting is also unnecessary for the
one-entry case.

### Proposed fix

Call `qsort` only when `count > 1`. This preserves deterministic ordering for
the only case that needs sorting and avoids passing an empty null buffer into
the C library.

### Scope

- `src/pipeline/pass_similarity.c`
- no behavior change for two or more entries
- no new dependency or API change

### Validation

The focused sanitizer-backed `simhash` suite passes: 24 tests.

WarGloom:
## What changed

`collect_fp_entries()` now calls `qsort` only when at least two fingerprint
entries were collected.

## Why

The empty path leaves `entries == NULL`. Calling `qsort(NULL, 0, ...)` relies on
library behavior outside the portable array-pointer contract, while the
one-entry path needs no sorting at all. Guarding with `count > 1` avoids both
unnecessary cases without changing deterministic ordering.

Fixes #1367.

## Validation

- `scripts/build.sh` — passed
- `scripts/test.sh --suites simhash` — passed, 24 sanitizer-backed tests
- `git diff --check` — passed

## Checklist

- [x] Commit is signed off (DCO)
- [x] Change is focused on one tracked bug
- [x] Existing focused tests pass

### DIFF
--- a/src/pipeline/pass_similarity.c
+++ b/src/pipeline/pass_similarity.c
@@ -148,8 +148,12 @@ static int collect_fp_entries(cbm_gbuf_t *gbuf, fp_entry_t **out_entries) {
             };
         }
     }
-    /* Canonicalize (determinism) — see cmp_fp_entry_by_qn. */
-    qsort(entries, (size_t)count, sizeof(fp_entry_t), cmp_fp_entry_by_qn);
+    /* Canonicalize (determinism) — see cmp_fp_entry_by_qn. Avoid passing the
+     * NULL empty-set buffer to qsort: libc treats zero elements as a no-op,
+     * but the standard function contract still requires a valid base pointer. */
+    if (count > 1) {
+        qsort(entries, (size_t)count, sizeof(fp_entry_t), cmp_fp_entry_by_qn);
+    }
     *out_entries = entries;
     return count;
 }


DeusData:
Thank you for this — and doubly so for reporting the issue *and* bringing the fix.

This is exactly the shape a fix should have: one file, one guard, and a comment that explains the *why* rather than restating the code. The distinction you drew in #1367 is the right one — zero elements suppress comparator calls, but that has never made a null base a valid array pointer, and `count > 1` also skips the one-entry sort that could never have changed anything.

I checked the caller before merging: `entry_count < MIN_FP_ENTRIES` returns early via `free(entries)`, and `free(NULL)` is well-defined, so the empty path stays correct end to end. Determinism at two or more entries is untouched.

Merging. Thanks again for the care you put into both the report and the patch.

DeusData:
One thing before I merge: this is still marked as a **draft**, so I have held off rather than promote it myself.

If that is just left over from opening it, mark it ready (or say the word here) and it goes in as-is — the review above is complete and CI is 28/28 green. If instead you are still polishing something, take the time you need; there is no rush from our side and I would rather merge the version you consider finished.

DeusData:
Thank you for the focused empty-fingerprint guard and for marking the PR ready. I have routed it at high priority in `0.9.1-rc` because it removes undefined behavior from the indexing pipeline. It is back in the maintainer review queue, which is currently full, so the final re-review may take a little time.

DeusData:
Thank you for this one — a textbook fix: exact scope, a clear rationale comment right at the call site, and an honest write-up of why the empty path violates the qsort contract even though every libc happens to tolerate it. The determinism contract for count >= 2 is untouched, so this is behavior-identical where it matters and strictly more correct where it doesn't. Much appreciated, as always!

One note: we verified that no C-level assertion can bind this fix (the existing simhash suite already drives the empty path, but only a trap-on-UBSan build would ever fail without the guard). We're taking a follow-up on our side to bind it in the sanitizer lane so it can never silently regress.

PR #1370, opened 2026-07-30
fix(cli): derive --help tool list from the MCP registry

### LINKED ISSUE #1361: --help omits check_index_coverage from the Tools list (MCP tools/list correctly returns all 15)
### Version

codebase-memory-mcp 0.9.1-rc.1

### Platform

macOS (Apple Silicon)

### Install channel

### Binary variant

standard

`--help` lists 14 tools and omits `check_index_coverage`, the new 15th tool. `tools/list` over MCP correctly returns all 15, so this is only the CLI help text.

Observed:

```
$ codebase-memory-mcp --help
...
Tools: index_repository, search_graph, query_graph, trace_path,
  get_code_snippet, get_graph_schema, get_architecture, search_code,
  list_projects, delete_project, index_status, detect_changes,
  manage_adr, ingest_traces
```

The tool itself works: `codebase-memory-mcp cli check_index_coverage --project <p> --scopes .` returns coverage entries, and `cli check_index_coverage --help` documents its flags. It's just missing from the top-level list, so anyone discovering tools via `--help` won't know the headline feature of this release exists.

Expected: `check_index_coverage` included in the `Tools:` line.

### Reproduction

```bash
codebase-memory-mcp --help | grep -c check_index_coverage   # 0
codebase-memory-mcp cli check_index_coverage --help          # works
```

### Confirmations

- I searched existing issues and this is not a duplicate.
- My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

---

*Posted by Claude Code on behalf of @artaommahe*

DeusData:
## Problem

`codebase-memory-mcp --help` listed 14 tools and omitted `check_index_coverage` — the headline feature of v0.9.1. MCP `tools/list` correctly returned all 15, and the tool itself worked (`cli check_index_coverage` and its `--help` were fine); only the top-level help was wrong.

Root cause: the `Tools:` block in `print_help()` was a hand-maintained copy of the tool registry. When `check_index_coverage` was added to `TOOLS[]` in `src/mcp/mcp.c`, nothing forced the help string to follow, so it drifted silently.

## Fix

Derive the help line from the registry instead of testing a second copy — a list that cannot drift beats a test that catches drift:

- `cbm_mcp_tools_help_list()` renders the `Tools:` block directly from the same `TOOLS[]` array that answers `tools/list`, wrapped for 80-column terminals in the existing style (registry order preserved).
- `print_help()` now prints that render; the hard-coded list is gone.
- `cbm_mcp_tool_count()` / `cbm_mcp_tool_name()` expose the registry, following the existing `cbm_mcp_tool_input_schema()` precedent.
- New test `mcp_tools_help_list_matches_registry` pins the render against the registry: every advertised tool appears, comma cardinality equals registry count (catches truncation/duplication from a formatter bug), and lines stay under 80 columns. A future 16th tool is covered automatically — no test update needed.

## Verification

- `./build/c/codebase-memory-mcp --help | grep -c check_index_coverage` → `1` (was 0); the help block now lists all 15 registry tools in registry order.
- `mcp` suite green (180 passed); bind proven by temporarily re-introducing the omission in the formatter → test fails with `strstr(help, name) is NULL`, restored → green.
- `make -f Makefile.cbm lint-ci` clean.

Fixes #1361

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -888,10 +888,13 @@ static void print_help(void) {
     printf("  Manual/UI MCP boundaries: Qodo, Warp, JetBrains AI/ACP, Replit,\n");
     printf("  Plandex, SWE-agent, BLACKBOX, GitHub cloud agents, Jules,\n");
     printf("  CodeRabbit.\n");
-    printf("\nTools: index_repository, search_graph, query_graph, trace_path,\n");
-    printf("  get_code_snippet, get_graph_schema, get_architecture, search_code,\n");
-    printf("  list_projects, delete_project, index_status, detect_changes,\n");
-    printf("  manage_adr, ingest_traces\n");
+    /* Rendered from the MCP tool registry: a hand-maintained copy here
+     * omitted check_index_coverage (#1361) and could silently drift again. */
+    char *tools_help = cbm_mcp_tools_help_list();
+    if (tools_help) {
+        printf("\n%s", tools_help);
+        free(tools_help);
+    }
 }
 
 /* ── Main ───────────────────────────────────────────────────────── */
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -89,6 +89,7 @@ enum {
 #include <yyjson/yyjson.h>
 #include <ctype.h>
 #include <limits.h>
+#include <stdarg.h> // va_list, for the bounded help-list appender
 #include <stdint.h> // int64_t
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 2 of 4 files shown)


DeusData:
CodeQL found two high-severity alerts in the first commit, and it was right — pushed 644c8fd to fix them.

The help-list builder accumulated `snprintf`'s return value, which is the length it *would* have written rather than what it did. Once `len` passes `cap`, the next `cap - len` underflows to a huge `size_t` and the following write lands outside the buffer.

Worth noting what makes this uncomfortable: it is the same class as #1173, which merged an hour before this — there `blen` walked past a 2 KB buffer by exactly the same mechanism.

The capacity computed here happens to be sufficient for the current registry, so nothing actually overflowed. That makes it the more dangerous shape rather than the safer one: correctness depended on an argument made ten lines away in the cap loop, so renaming a tool or changing the wrap column would have converted it into an out-of-bounds write with nothing in the function to notice. `help_append()` now returns the bytes actually written, clamped to the remaining space, so `len <= cap - 1` is a local invariant that survives whatever the registry becomes.

Verified before pushing: builds clean under `-Werror`, `--help` still renders all 15 tools in registry order with identical wrapping, mcp + cli suites 437 passed / 0 failed, and `lint-ci` clean.

Issue #1000, opened 2026-07-10
Regression (main, post-v0.9.0): #979 import-map direct-hit swallows call suffixes — django loses ~11K CALLS/TESTS edges


DeusData:
Bisected to #979 (merge a6207c6, the #875 aliased direct-symbol import fix). The new direct-hit early return in `resolve_import_map` fires **before suffix handling**, so any `imported_symbol.method()` call whose base symbol is itself an indexed node resolves to the **base node** instead of `base.method` — the exact mis-resolution the comment above the insertion warns about.

**Measured on django (identical checkout, full mode):** CALLS 74,685 → 72,175 (−2,510), TESTS 38,032 → 29,236 (−8,796; test linkage consumes the same resolution path). Concrete example: `auth.login ⇒ Signal.send` degraded to `auth.login ⇒ signals.user_logged_in` (the Variable node, `.send` swallowed). ~9.6K true edges lost, ~7.1K degraded edges gained.

**Fix (ready locally, reproduce-first):** take the direct hit only when the callee has **no suffix** — which is precisely #979's intended case (`from m import f as g; g()`). With the fix: django CALLS 74,708 (= old + 23 genuine #979 wins), TESTS 38,035; #979's own `resolve_import_map_bare_alias` test stays green; new guard `resolve_import_map_alias_with_suffix_hits_method` is RED on current main and GREEN with the fix (registry suite 56/56).

Not in any release — v0.9.0 predates #979; needs to land before 0.9.1. Thanks @LA-10 for #979 itself — the feature is kept intact, only the early-return gets the suffix guard.

PR #1371, opened 2026-07-31
fix(python): resolve aliased from-import CALLS to real def

Joseph-MingEn:
## Summary
- Sequential `build_import_map` now uses `resolve_import_node` (not `fqn_module` on `pkg.symbol` paths) so `from M import f as g` lands in the import map.
- Absolute/relative from-import: when Strategy-1 lands on a Module/File but `module_path` names a member Function/Class, **prefer the member** as the IMPORTS target so bare alias calls can `import_map` to the real def.
- Regression / pins: relative + absolute aliased from-import CALLS contracts; ghost-module negative (`from ghost_module import nope as alias; alias()` → **no CALLS**); registry adversarial pin for def-QN-under-alias-key (behavior already on main via #875/#979 — comment-only `registry.c`).

## Broader graph-shape note (maintainer call)
Plain and aliased from-imports share the same `module_path` (`"M.f"`), so the member-preference block retargets IMPORTS **Module → member def** for **every** Python from-import that resolves that way — not only aliases. That is more precise for `from M import f`, but it changes:
- module-dependency / file→Module IMPORTS expectations
- `detect_changes` impact propagation that walked Module IMPORTS
- any query assuming from-import always ends on Module

This is intentionally in front of the maintainer (same mechanism class as the ~11K-edge #1000 regression). Happy to split or gate if preferred.

## Evidence (not Django-scale)
- Local: `./build/c/test-runner registry lang_contract` → **97 passed** (includes new ghost-module negative).
- Consumer lab (Yui BASIC, mid-size Python monorepo): after patched MCP + full re-index, `trace_path(submit_task)` hop-1 lands on `services.satori_bridge.gate.execute` without changing the `import execute as bridge_execute` alias in source. Graph after ≈ 11549 nodes / 46879 edges. No Django-scale before/after edge-count diff available from this contributor machine.

## Test plan
- [x] `make -f Makefile.cbm lint-format` (Homebrew LLVM `clang-format`)
- [x] `./build/c/test-runner registry lang_contract` — 97 passed
- [ ] CI `lint` + **`test`** (was skipped while lint failed) + `ci-ok`

### DIFF
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -114,20 +114,23 @@ static int build_import_map(cbm_pipeline_ctx_t *ctx, const char *rel_path,
     *out_vals = NULL;
     *out_count = 0;
 
-    /* Fast path: build from cached extraction result (no JSON parsing) */
+    /* Fast path: build from cached extraction via the same resolver used for
+     * IMPORTS edges. Do NOT use cbm_pipeline_fqn_module(module_path) — Python
+     * from-imports store "pkg.symbol" (and aliases) in module_path, which is
+     * not a filesystem rel-path and misses the def node. */
     if (result && result->imports.count > 0) {
         const char **keys = calloc((size_t)result->imports.count, sizeof(const char *));
         const char **vals = calloc((size_t)result->imports.count, sizeof(const char *));
         int count = 0;
+        char *file_qn = cbm_pipeline_fqn_compute(ctx->project_name, rel_path, "__file__");
 
         for (int i = 0; i < result->imports.count; i++) {
             const CBMImport *imp = &result->imports.items[i];
             if (!imp->local_name || !imp->local_name[0] || !imp->module_path) {
                 continue;
             }
-            char *target_qn = cbm_pipeline_fqn_module(ctx->project_name, imp->module_path);
-            const cbm_gbuf_node_t *target = cbm_gbuf_find_by_qn(ctx->gbuf, target_qn);
-            free(target_qn);
+            const cbm_gbuf_node_t *target =
+                cbm_pipeline_resolve_import_node(ctx, rel_path, file_qn, imp, NULL);
             if (!target) {
                 continue;
             }
@@ -135,11 +138,18 @@ static int build_import_map(cbm_pipeline_ctx_t *ctx, const char *rel_path,
             vals[count] = target->qualified_name; /* borrowed from gbuf */
             count++;
         }
+        free(file_qn);
 
-        *out_keys = keys;
-        *out_vals = vals;
-        *out_count = count;
-        return 0;
+        if (count > 0) {
+            *out_keys = keys;
+            *out_vals = vals;
+            *out_count = count;
+            return 0;
+        }
+        free((void *)keys);
+        free((void *)vals);
+        /* Fall through to IMPORTS-edge scan when extraction paths did not
+         * resolve (should be rare once resolve_import_node is used). */
     }
 
     /* Slow path: scan graph buffer IMPORTS edges + parse JSON properties */
--- a/src/pipeline/pass_pkgmap.c
+++ b/src/pipeline/pass_pkgmap.c
@@ -1530,9 +1530,8 @@ const cbm_gbuf_node_t *cbm_pipeline_resolve_import_node(const cbm_pipeline_ctx_t
             char symbuf[256];
             const char *sym = import_candidate_symbol(imp->module_path, symbuf, sizeof(symbuf));
             if (sym && sym[0] && strcmp(sym, "*") != 0) {
-                const char *mod_tail = import_last_segment(target->qualified_name
-                                                               ? target->qualified_name
-                                                               : "");
+                const char *mod_tail =
+                    import_last_segment(target->qualified_name ? target->qualified_name : "");
                 if (!mod_tail || strcmp(mod_tail, sym) != 0) {
                     char member_qn[CBM_SZ_512];
                     snprintf(member_qn, sizeof(member_qn), "%s.%s", target->qualified_name, sym);
--- a/src/pipeline/pass_pkgmap.c
+++ b/src/pipeline/pass_pkgmap.c
@@ -1520,6 +1520,31 @@ const cbm_gbuf_node_t *cbm_pipeline_resolve_import_node(const cbm_pipeline_ctx_t
     const cbm_gbuf_node_t *target = target_qn ? cbm_gbuf_find_by_qn(ctx->gbuf, target_qn) : NULL;
     free(target_qn);
     if (target) {
+        /* Python/TS from-import of a member: module_path is often
+         * "pkg.mod.symbol" while resolve_module lands on the Module node
+         * "pkg.mod". Prefer the member Function/Class when it exists —
+         * required for `from M import f as g` so CALLS can import_map to f
+         * (local_name g ≠ f). */
+        if (target->label &&
+            (strcmp(target->label, "Module") == 0 || strcmp(target->label, "File") == 0)) {
+            char symbuf[256];
+            const char *sym = import_candidate_symbol(imp->module_path, symbuf, sizeof(symbuf));
+            if (sym && sym[0] && strcmp(sym, "*") != 0) {
+                const char *mod_tail = import_last_segment(target->qualified_name
+                                                               ? target->qualified_name
+                                                               : "");
+                if (!mod_tail || strcmp(mod_tail, sym) != 0) {
+                    char member_qn[CBM_SZ_512];
+                    snprintf(member_qn, sizeof(member_qn), "%s.%s", target->qualified_name, sym);
+                    const cbm_gbuf_node_t *member = cbm_gbuf_find_by_qn(ctx->gbuf, member_qn);
+                    if (member && import_targetable_label(member->label) &&
+                        strcmp(member->label, "Module") != 0 &&
+                        strcmp(member->label, "File") != 0) {
+                        return member;
+                    }
+                }
+            }
+        }
         return target;
     }
 
--- a/src/pipeline/registry.c
+++ b/src/pipeline/registry.c
@@ -619,11 +619,12 @@ static cbm_resolution_t resolve_import_map(const cbm_registry_t *r, const char *
      * resolved.requireAdmin — not just resolved, which would point at the
      * module node and miss the function entirely. */
     /* Direct hit ONLY for suffix-less callees (an aliased direct-symbol
-     * import called bare: `from m import f as g; g()` — #875/#979). With a
-     * suffix present (`imported.method()`), returning the bare base here
-     * would swallow the suffix and bind the call to the imported symbol's
-     * own node (a Variable/Class/module) instead of base.method — exactly
-     * the mis-resolution the comment above warns about. That regressed
+     * import called bare: `from m import f as g; g()` — #875/#979; Yui
+     * `import execute as bridge_execute`). With a suffix present
+     * (`imported.method()`), returning the bare base here would swallow
+     * the suffix and bind the call to the imported symbol's own node
+     * (a Variable/Class/module) instead of base.method — exactly the
+     * mis-resolution the comment above warns about. That regressed
      * django-scale graphs by ~11K CALLS/TESTS edges (Signal.send calls
      * degraded to edges onto the signal variables themselves). #1000 */
     if (!suffix || !suffix[0]) {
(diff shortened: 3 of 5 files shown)


Joseph-MingEn:
Rebased onto current `DeusData/main` and resolved conflicts in `registry.c` / `test_registry.c`:

- Kept upstream `#875/#979/#1000` suffix-safe bare-alias direct hit (do not swallow `.method` suffix).
- Kept Yui-shaped `resolve_import_map_aliased_from_import` + absolute from-import member IMPORTS (`pass_pkgmap`) + lang_contract tests.
- Local: `./build/c/test-runner registry lang_contract` → **96 passed**.

Ready for CI when the merge queue picks this up.

Joseph-MingEn:
**Conflict recheck:** rebased/force-pushed head `b81c1e2`. Local merge vs current `main` (`69599b7`) is a clean **fast-forward** (no conflict markers). If the UI still shows conflicts on `registry.c` / `test_registry.c`, please refresh mergeability — those were from the pre-rebase tip.

DeusData:
Thank you for this — and welcome. Reviewed properly rather than skimmed, because the root cause you found is a good one.

**The diagnosis is right and I verified it on main.** `cbm_pipeline_fqn_module` → `fqn_compute(name=NULL)` → `strip_file_extension()` strips the last dot-segment unconditionally, so a Python from-import path `pkg.mod.symbol` collapses to the Module QN `proj.pkg.mod`. The alias then maps to the Module, a bare call builds the candidate `moduleQN.g` — a ghost — and resolves to nothing. Routing the sequential import map through `cbm_pipeline_resolve_import_node`, the same resolver the IMPORTS pass already uses, fixes the sequential and parallel legs at a single choke point. That is the right shape, and it is the right layer: pipeline base, below the LSP passes, so the hybrid-LSP work stays additive on top.

I also went looking for the failure mode that would make this dangerous — resolving to a same-named symbol in a *different* module — and it cannot happen. The member QN is built strictly inside the resolved module's namespace and must hit an existing node by exact QN with a targetable label, with Module and File excluded. Cross-module leakage is not reachable. Import maps are per-file, so the same alias in two files cannot collide either.

There is a nice side benefit too: `from M import C; C.method()` now direct-hits `proj.M.C.method` instead of leaning on the by-name suffix fallback, which could previously pick a *different* class's `.method` in the same module. That is a genuine accuracy win beyond the aliased case.

**Two things need your hand, one trivial and one that needs the maintainer.**

**1. The lint failure is hiding something important.** `pass_pkgmap.c:1533-1534` has clang-format violations — your new `import_last_segment(...)` line break. Trivial in itself, but the consequence is not: because `lint` failed, the **entire `test` job was auto-skipped**, so your new contract tests have *never actually run on CI*, on any platform. The green `pr-smoke` checks make the PR look better-verified than it is.

Run `make -f Makefile.cbm lint-ci` locally to reproduce it. One caveat that will save you a wasted round trip: our clang-format is the **Homebrew LLVM** build — a standalone `clang-format-20` produces spurious whole-file drift that looks nothing like the two real lines. Once that is fixed, CI will exercise your tests for the first time.

**2. The change is wider than the title, and that part is the maintainer's call.** Because plain and aliased from-imports store the *same* `module_path` (`"M.f"`), the member-preference block retargets the IMPORTS edge from the Module node to the member def for **every Python from-import in every indexed repo** — not just aliased ones. It is arguably more precise, and arguably what `from M import f` actually means. But it changes module-dependency queries, `detect_changes` impact propagation, and anything expecting file→Module IMPORTS edges.

That is a graph-shape decision rather than a bug fix, and it lands in the exact mechanism that once regressed about 11,000 edges (#1000), so I have put it in front of the maintainer instead of deciding it in review. Two things would help it along: state the broader effect explicitly in the PR description, and — if you are able — a before/after edge-count diff on a Django-scale repository. That evidence is what would derisk it.

**One test I would like:** a pure negative — `from ghost_module import nope as alias; alias()` must produce **no** CALLS edge at all. Your existing ghost-negative assertions cover the invented-alias case nicely; this would pin the unresolvable-module case too, which is the one where an over-eager resolver would invent an edge.

Smaller notes: the PR body says the change makes `resolve_import_map` prefer an exact registry hit, but that behaviour already exists on main and your `registry.c` diff is comment-only — the registry test is a useful adversarial pin rather than a regression test, worth relabelling. And the empty commit `345a350` is harmless; it disappears in a merge commit.

Fix the formatting and this gets a real CI run; I will come back to you on the IMPORTS-edge question either way.

Joseph-MingEn:
Thanks for the careful review — addressed:

1. **Lint / format:** `pass_pkgmap.c` member-prefer block reformatted with Homebrew LLVM `clang-format` (`make -f Makefile.cbm lint-format` clean). Rebased onto current `main`.
2. **Ghost-module negative:** added `contract_python_ghost_module_aliased_from_import_no_calls` — `from ghost_module import nope as alias; alias()` must emit **no** CALLS. Local `registry` + `lang_contract` → **97 passed**.
3. **PR description:** now states the broader Module→member IMPORTS retarget for all Python from-imports (not just aliases), and that this is a maintainer graph-shape call. Yui mid-size lab cited; no Django-scale edge-count diff from this machine.
4. **Registry test:** relabelled as adversarial pin (exact-hit behavior already on main; `registry.c` remains comment-only).

Please re-check once CI runs `test` (previously skipped behind lint). Happy to iterate on the IMPORTS-edge scope if you want it narrowed or gated.

Joseph-MingEn:
**CI note:** the red `test-unix (macos-15-intel)` / `ci-ok` is unrelated to this PR’s Python alias / IMPORTS changes.

Failure site:
```
lock_registry_absolute_deadline_survives_repeated_wakes
FAIL tests/test_lock_registry.c:1153: ASSERT(tail_queued)
```

On that same job, `lang_contract` was **40/40** (incl. the new ghost-module negative). All other `test-unix` / Windows / tsan / lint / smoke matrices were green.

Re-running failed jobs now. If it flakes again we can wait for maintainer guidance or another rerun — not planning to touch `lock_registry` in this PR.

DeusData:
**Your diagnosis is correct, and you do not need to prove it — it is already on our list.**

`lock_registry_absolute_deadline_survives_repeated_wakes` (`tests/test_lock_registry.c:1153`) is one of two tests in that suite that wait on a **500 ms observation window** for waiters to park, and then assert against it. That is a transient-window assertion, which under our own testing doctrine is a defect in the test rather than a property of your change: a verdict that depends on how a shared runner is scheduled is not a verdict. Its sibling `lock_registry_large_queue_parks_non_head_waiters` (`:897`) has the same shape. Both fail preferentially on the slower `macos-15-intel` runner, which is exactly where yours failed.

So: **not yours, and you were right not to touch `lock_registry` in this PR.** A re-run may clear it, but please read that as a coin landing the right way rather than as evidence — the fix we owe is a seam or a held gate, not a wider budget. It is recorded for us to do.

Your PR itself is still with the maintainer for the graph-shape question raised earlier; that is a direction call, not a quality one, and nothing further is needed from you while it waits. Thank you for the clean triage — reporting the failing site with the surrounding evidence (`lang_contract` 40/40 on the same job, everything else green) is precisely what makes a CI report actionable.

DeusData:
Thank you @Joseph-MingEn — and thank you especially for flagging the graph-shape consequence yourself instead of letting it land quietly. That is the part that decides this PR.

The alias fix is clearly right: `from M import f as g` followed by `g()` should reach the real def, and the ghost-module negative (`from ghost_module import nope as alias` → no CALLS) is exactly the control I would have asked for.

But as you say, the member-preference block retargets IMPORTS **Module → member def** for *every* Python from-import that resolves that way, not only aliased ones. That is more precise in isolation and it changes module-dependency shape, `detect_changes` impact propagation, and any query that assumes a from-import terminates on a Module. Your comparison to the ~11K-edge #1000 regression is apt, which is why I am not merging this on my own judgement — it is a direction call for the maintainer, and I have put it in front of him with your framing intact.

Your offer to split or gate is the likely path: the alias resolution could land now on its own merits, with the broader Module→member retarget as a separate, deliberate decision that gets a before/after edge-count comparison on a large repo. If he prefers that, would you be up for splitting it? If not, I will do the split myself and keep you as author on both halves.

Either way the work is not going to sit — it is in the active batch, just gated on one decision rather than on review.

DeusData:
@Joseph-MingEn — the maintainer came back with a better option than the either/or I put to you, and I think it dissolves the direction question entirely.

**Emit both edges instead of retargeting.** Keep `File → Module M`, and *add* `File → Function M.f`, rather than replacing one with the other.

I checked that the graph can actually hold that: edge identity is `source:target:type` (plus `local_name` for IMPORTS), so the Module edge and the member edge have different keys and coexist naturally — no collision, and genuinely repeated imports still dedup on their own. IMPORTS already carries `local_name`, so the alias itself is modelled today.

What that buys, compared to retargeting:

- the alias call resolution you are after works the same way;
- module-dependency queries and `detect_changes` propagation over Module IMPORTS keep working, so the shape change that made this a maintainer call goes away;
- the risk becomes purely additive rather than semantic — nothing that works today stops working.

The cost is edge growth: one extra IMPORTS edge per from-import that resolves to a member, which on a Python-heavy repo is not nothing. Two things worth doing about that:

1. **Mark the member edge** with a distinguishing property so consumers counting IMPORTS for dependency metrics can filter it, instead of having to infer intent from the target's label.
2. **Measure it.** Your consumer-lab numbers (≈11549 nodes / 46879 edges) are a fine baseline — a before/after on the same repo would tell us the real multiplier, and that is the number that decides whether this needs gating.

Would you like to take that direction? If you would rather not rework it, say so and I will implement it with you as co-author — the analysis and the ghost-module control are the hard parts and they are already done.

Your instinct to flag the shape change rather than bury it is what made this outcome possible; a quieter PR would have shipped the regression.

DeusData:
Correction to my previous comment, because I framed the edge growth wrongly and the maintainer pushed back on it.

I called the extra edges a "cost" to be measured before deciding whether to gate. That was over-cautious. These are **true additions**: `from M import f` really does import the symbol, and the graph has only ever recorded half of that relationship. And since the IMPORTS edge key includes `local_name`, repeated identical imports dedup on their own — so growth is bounded by the number of *distinct* (file, imported symbol) pairs, which is a real property of the codebase rather than an inflation factor.

An edge count that rises because the graph finally represents something true is this project working as intended, not a regression to be defended against. Graph quality comes first here.

So the proposal stands unchanged in shape, but without the gating implication:

- **emit both edges** (Module + member), no retargeting — nothing that works today stops working;
- **mark the member edge** so consumers counting IMPORTS for dependency metrics can distinguish a module dependency from a symbol import. That is a labelling concern, and it is the only real one left;
- **record the before/after numbers** on your consumer lab as the new baseline — useful to have, not a gate to pass.

Same offer as before: yours to take, or I implement it with you as co-author. And apologies for the mixed signal — the second read was better than my first.

DeusData:
Decision: we want BOTH halves — the import-map CALLS fix and the Module→member IMPORTS retarget — but the shape change needs the corpus-scale evidence you rightly flagged you couldn't produce on your machine. So rather than take your split offer or ask more of you, we'll run the before/after ourselves on our bench corpora (Django-scale plus the multi-million-node sets) against your branch as-is, and report the edge deltas here. If the numbers hold (our prior for this mechanism class is caution — a similar retarget once cost ~11K edges), this merges whole; if not, we take the CALLS half and revisit. Nothing needed from you right now except patience — thank you for flagging the risk on your own PR; that honesty is why we're investing the bench time.

DeusData:
Ran the promised before/after on our bench corpora (cold full index, isolated stores, per-edge-type counts straight from the DB; patch = this PR's full diff applied onto current main):

**django** — nodes identical (55,371); edges 343,557 → 343,613 (**+56**), indexing time flat (11.6s → 11.4s). The delta is exactly the quality direction we hoped for: **INHERITS +26** and **OVERRIDE +42** (aliased base-class imports now land on the real class defs, and the override detection cascades from there), one USAGE upgraded to a resolved CALLS, and 12 heuristic SEMANTICALLY_RELATED edges yielding to exact ones. Crucially the #1000-sensitive types are untouched: TESTS 45,944 → 45,944, IMPORTS 16,068 → 16,068, no losses anywhere.

**typescript** — 879,469 edges before and after with **zero** per-type deltas, time flat. So the resolver change is a no-op for TS on a large real corpus — no collateral.

Suite-level: registry + lang_contract 98/98 with the patch on current main; reverting just the src half turns `contract_python_aliased_from_import_calls` red for the right reason (`ASSERT(saw_execute)`). One honest note: the absolute-import variant already passes on unpatched main — the behavioral delta this PR ships is specifically the **relative** aliased from-import path (plus the member-preference improvement the django numbers show).

Numbers hold — merging both halves. Thanks for a carefully scoped fix in a historically treacherous corner (this code region once ate ~11K edges in #1000), and for the adversarial tests that pin the alias-ghost behavior down.

PR #1372, opened 2026-07-31
fix(mcp): scope detect_changes seed detection to changed line ranges

### LINKED ISSUE #1363: detect_changes: seed_symbols scoped to whole file, not the changed line range
## Problem

`detect_changes`'s "seed" set (the symbols it treats as directly changed, before
running the blast-radius traversal) is computed at **whole-file granularity**,
not at the diff-hunk/line level. A one-line, single-statement edit inside one
method reports every definition in that file as a seed symbol, then runs the
blast-radius traversal from all of them — producing an impact report an order
of magnitude larger than what actually changed.

This contradicts the stated intent in #1154 ("It now resolves the git diff
**to those symbols** and runs one multi-source BFS ... to the transitive
impact set") — the BFS traversal is real, but the *input* to it (`seed_symbols`)
is not scoped to the touched lines.

## Reproduction

Tested on `microsoft/qlib` @ `3097dcc9` (any repo with more than one definition
per file should reproduce this; the two cases below were chosen for very
different file sizes to rule out coincidence).

```bash
codebase-memory-mcp cli index_repository --repo-path <qlib-checkout> --mode full --name repro
```

**Case A** — `qlib/log.py` (24 top-level/class-member definitions):

```bash
# single-line in-place edit inside TimeInspector.logt(), no line-count change
sed -i '' 's/            pass/            cls.timer_logger.debug("x")/' qlib/log.py
codebase-memory-mcp cli detect_changes --project repro --format json --depth 1
```
```json
{"seed_symbols": 25, "impacted_total": 141, ...}
```
`25 = 24 definitions + 1 module node.` Only `logt` was edited; the other 23
definitions in the file are untouched.

**Case B** — `qlib/backtest/decision.py` (42 definitions), same methodology,
one line replaced with itself plus a trailing comment (no line-count change,
touches exactly one `return` statement in one method):

```bash
sed -i '' '96s/.*/        return self.amount * self.sign  # bench-marker/' qlib/backtest/decision.py
codebase-memory-mcp cli detect_changes --project repro --format json --depth 1
```
```json
{"seed_symbols": 43, "impacted_total": 68, ...}
```
`43 = 42 definitions + 1 module node.`

Both cases reproduce the same formula — `seed_symbols == (definitions in
changed file) + 1` — regardless of file size or which single line was
touched, which rules out this being an artifact of one file's structure.

## Expected Behavior

`seed_symbols` should be the symbol(s) whose line range actually overlaps the
diff hunk (e.g. just `logt` for Case A, just the one method for Case B), not
every definition in the file.

## Actual Behavior

Every definition in the changed file is treated as a seed, so the blast-radius
BFS runs from all of them. In Case A, the true direct callers of `logt` (6,
confirmed via oracle cross-check) were present in the output but diluted
across 141 total `impacted` entries pulled in by the other 23 unrelated seeds.

## Impact

For any file with more than a handful of definitions, `detect_changes` on a
single-line change reports an impact set dominated by symbols the edit never
touched. This makes the tool unusable as a precision impact-review aid without
manual filtering — the caller has to guess which of the N seeds is the one
they actually meant.

## Environment

- codebase-memory-mcp 0.9.1-rc.1
- macOS (Apple Silicon)
- Reproduced twice on two files of very different size (24 and 42 definitions)
  with identical `seed_symbols = defs_in_file + 1` in both

## Suggested Fix

Scope seed identification to the definition(s) whose `[start_line, end_line]`
range intersects the diff hunk's changed line range, rather than every
definition record associated with the changed file path.

lishixiang0705:
## What does this PR do?

Fixes #1363: `detect_changes` scoped its BFS seed set to *every* definition
in a changed file, regardless of which lines actually changed. A one-line
edit inside a single method seeded every other definition in the file too,
producing an impact report an order of magnitude larger than what the edit
actually touched.

`cbm_parse_hunks` (in `pass_gitdiff.c`) already existed to parse
`git diff --unified=0` hunks but had no caller. This PR fetches those hunks
alongside the existing changed-file list and scopes seeds to definitions
whose line range overlaps a hunk. Any failure fetching hunks (new/untracked
files, a transient git error) falls back to the previous whole-file
behavior, so this is a precision improvement with no new failure mode.

`cbm_changed_hunk_t` / `cbm_parse_hunks` move from `pipeline_internal.h`
(explicitly not a public header) to `pipeline.h`, since `detect_changes`
(in `src/mcp/mcp.c`) now needs them. `cbm_detect_node_in_hunks` is exposed
non-static via `mcp_internal.h`, matching this file's existing white-box
test-hook pattern, so the overlap logic has a direct unit test independent
of the git/subprocess/index plumbing around it.

**Verified against v0.9.1-rc.1** — the RC's tree-format rewrite (this
release's "real `detect_changes` impact reporting") is a different, earlier
part of this same path: it made the BFS traversal itself real instead of a
placeholder, but the seed set feeding that traversal was still whole-file.
I confirmed the bug still reproduces on the `v0.9.1-rc.1` release binary
before opening this PR:

```
$ codebase-memory-mcp cli --json detect_changes '{"project":"...","depth":1}'
...
seed_symbols: 2   # foo() and bar(), only foo() was touched
```

Also verified on `microsoft/qlib` (an unrelated third-party repo): a
same-line-count edit inside one method now seeds 2 symbols (the method +
its containing class) instead of 25 (every definition in the 24-def file),
matching the equivalent output of a different code-graph tool (GitNexus) on
the identical scenario. Reproduced on a second file with a different
definition count (42 defs -> seeds 43 before the fix, 2 after) to rule out
coincidence.

This branch is rebased onto current `main` (on top of v0.9.1-rc.1 plus the
subsequent commits); local `scripts/test.sh SANITIZE=` and
`scripts/lint.sh --ci` both pass clean.

## Checklist

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -9736,19 +9736,54 @@ static int mcp_run_shell_command_cancellable(cbm_mcp_server_t *srv, const char *
     return contained ? 0 : -1;
 }
 
-/* Collect BFS seed ids: every symbol DEFINED in a changed file (everything but
- * the structural container labels — those have no CALLS edges). These anchor
- * the multi-source impact traversal. */
+/* Does `node`'s line range overlap any recorded hunk for `file`? Used to scope
+ * seed detection to the actually-changed lines rather than the whole file.
+ * Non-static (declared in mcp_internal.h) so tests can exercise the overlap
+ * logic directly, matching this file's existing white-box test hooks. */
+bool cbm_detect_node_in_hunks(const cbm_node_t *node, const cbm_changed_hunk_t *hunks,
+                              int hunk_count, const char *file) {
+    for (int h = 0; h < hunk_count; h++) {
+        if (strcmp(hunks[h].path, file) == 0 && node->start_line <= hunks[h].end_line &&
+            node->end_line >= hunks[h].start_line) {
+            return true;
+        }
+    }
+    return false;
+}
+
+/* Collect BFS seed ids for one changed file (everything but the structural
+ * container labels — those have no CALLS edges). These anchor the
+ * multi-source impact traversal.
+ *
+ * When `hunks` has at least one entry for `file`, only definitions whose line
+ * range overlaps a hunk become seeds — a one-line edit inside a single
+ * function no longer seeds every other definition in the file. When no hunk
+ * is recorded for `file` (a brand-new/untracked file has no comparable
+ * "before" state, or the hunk fetch failed/was skipped), every non-container
+ * definition in the file is a seed — the previous, whole-file behavior — so
+ * this is a precision improvement, not a new failure mode. */
 static void detect_collect_seeds(cbm_store_t *store, const char *project, const char *file,
-                                 int64_t **seeds, int *n, int *cap) {
+                                 const cbm_changed_hunk_t *hunks, int hunk_count, int64_t **seeds,
+                                 int *n, int *cap) {
     cbm_node_t *nodes = NULL;
     int ncount = 0;
     cbm_store_find_nodes_by_file(store, project, file, &nodes, &ncount);
+    bool has_hunks_for_file = false;
+    for (int h = 0; h < hunk_count; h++) {
+        if (strcmp(hunks[h].path, file) == 0) {
+            has_hunks_for_file = true;
+            break;
+        }
+    }
     for (int i = 0; i < ncount; i++) {
         const char *lb = nodes[i].label;
         if (lb && strcmp(lb, "File") != 0 && strcmp(lb, "Folder") != 0 &&
             strcmp(lb, "Project") != 0 && strcmp(lb, "Module") != 0 && strcmp(lb, "Package") != 0 &&
             strcmp(lb, "Section") != 0) {
+            if (has_hunks_for_file &&
+                !cbm_detect_node_in_hunks(&nodes[i], hunks, hunk_count, file)) {
+                continue;
+            }
             if (*n >= *cap) {
                 *cap = *cap ? *cap * 2 : 16;
                 *seeds = safe_realloc(*seeds, (size_t)*cap * sizeof(int64_t));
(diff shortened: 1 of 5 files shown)


DeusData:
Reviewed in full, and this is a good find. **The bug is real and I verified it on main:** `cbm_parse_hunks` sits in `src/pipeline/pass_gitdiff.c:154` with **zero production callers** — the only references outside its own declaration are in `tests/test_pipeline.c`. Meanwhile `detect_collect_seeds` seeds every non-container node in a changed file. Dormant machinery plus an over-approximating caller is exactly the kind of archaeology that is worth doing, and your reproduce-first e2e test (`seed_symbols` 2 → 1) plus the v0.9.1-rc.1 and qlib cross-checks made verifying it straightforward. Thank you.

**On the direction:** narrowing to changed line ranges is noise removal rather than a findability cut, which is the important question for us — the seeds you drop are definitions in a changed file whose lines did not change. Deletion safety is handled correctly too: `parse_hunk_line` rejects `start <= 0`, so a deleted file emits no hunks and falls back to whole-file seeding.

**One CI red is genuinely yours, and it is a small fix.** The Windows shard fails at `tests/test_mcp.c:6202: git fixture setup failed` — "The filename, directory name, or volume label syntax is incorrect." The fixture uses `system("cd '%s' && git init …")` with POSIX single quotes, which `cmd.exe` cannot parse. The repo already has the portable pattern: `wt_git()` in `tests/test_watcher.c:34` (used at `:1183`), which shells `git -C "<dir>" …`. Switch to that shape and the shard should go green. Everything else — 27 checks including both TSan trios and the other Windows shard — passes.

**Two recall edge cases I would like closed before merge**, both inside the shape you already have:

1. **A file with hunks but zero overlapping definitions now contributes no seeds.** Import-only or other module-level edits land outside every definition's line range, so that file drops out entirely where main seeded all of its definitions. Suggest: when the overlap count for a file is 0, fall back to whole-file seeding for that file.
2. **`HUNK_CAP = 4096` truncates silently.** On a large refactor the diff can exceed the cap, and files whose hunks fell past it under-seed while `has_hunks_for_file` may still read true for partially-captured files. Suggest: if `cbm_parse_hunks` returns exactly `HUNK_CAP`, disable scoping for the whole request and seed as before. At this repo's scale a >4096-hunk diff is not hypothetical.

**And one worth a comment rather than a fix:** the coordinate systems are mixed — `base…HEAD` hunks are HEAD-side lines, worktree hunks are worktree-side, and node lines come from the indexed snapshot. That is correct for a freshly refreshed index, but a stale index plus earlier-in-file insertions will mis-scope. A note in the code (or preferring the worktree diff when it is non-empty) would cover it.

Security review is clean: no new dependencies, no CI changes, `cbm_fopen`/`cbm_unlink` used correctly, and the new hunk-fetch command interpolates only `root_path` and `base_branch` — both already validated upstream and quoted identically to the existing name-only command at `mcp.c:9830`. No new injection surface.

Ping me when the fixture and the two edge cases are in and I will pick it straight back up.

lishixiang0705:
Thanks — the two edge cases were both real, and the first one was sharper than I
realised. All three items are in `9235a23e`.

**Windows fixture.** Switched to the `wt_git` shape you pointed at: `git -C
"<dir>"` with identity, `init.defaultBranch` and `commit.gpgsign` passed via
`-c`, so the fixture no longer reads the machine's global git config either.

**Zero-overlap fallback — and a bug in my first attempt at it.** Implemented as
suggested: probe the file for an overlapping definition, and keep whole-file
seeding when there is none. The first version still returned `seed_symbols: 0`
for an import-only edit, because the probe walked raw nodes while the seeding
loop applies a label filter. A `Module` node spans lines 1..EOF, so it overlaps
*every* hunk — the probe always reported an overlap and the fallback could never
fire. Extracted `detect_is_seedable_label()` so the probe and the seeding loop
share one definition of "seedable".

Worth recording that the regression test is what caught it: it asserted
`seed_symbols: 2` for a one-line import edit above two functions and failed at 0
until the probe was filtered. Without the test I would have shipped a fallback
that was structurally dead.

**HUNK_CAP truncation.** `hunk_count >= HUNK_CAP` now drops scoping for the
whole request and logs `detect_changes.hunks action=scoping_disabled
reason=hunk_cap_reached`, rather than letting partially-captured files look
scoped and under-seed.

**Coordinate systems.** Documented in the fetch block rather than changing
behavior, per your suggestion — HEAD-side vs worktree-side vs indexed-snapshot
lines, when they agree, and that the failure mode is bounded by the zero-overlap
fallback (a file whose definitions all miss reverts to whole-file seeding rather
than dropping out). Happy to switch to preferring the worktree diff when it is
non-empty if you would rather have the behavioral fix.

Local: 6792 passed / 0 failed, `scripts/lint.sh --ci` clean.

DeusData:
Merged as `27893016b`. All three items verified, 29/29 green.

**The bug you caught in your own fix is the part worth dwelling on.** I confirmed it independently: `extract_defs.c:6817` gives every file a `Module` node spanning `start_line = FIRST_LINE` to the end of the root node — lines 1..EOF. An unfiltered probe therefore overlaps *every* hunk in *every* file, so the zero-overlap fallback would have been structurally dead code and import-only edits would still have seeded nothing. Your description of it was accurate rather than post-hoc narrative, and finding it required noticing that the probe and the seeding loop disagreed about what counts as a node — which is exactly the kind of mismatch that survives review.

The fix is also structurally right, not just correct: the probe and the seeding loop now call the *same* `detect_is_seedable_label()` and the *same* `cbm_detect_node_in_hunks()`. There is no second copy of the condition that can drift apart later. That is what stops this bug recurring, and it is why I am not asking for anything else here.

**The other two hold up too.** The Windows fixture passing identity, `init.defaultBranch` and `commit.gpgsign` via `-c` means it no longer reads the developer's global git config — a fixture that depends on that is a latent flake, so thank you for going past what was asked. Both Windows shards are green, which is the venue that caught the original problem. And the `HUNK_CAP` guard drops scoping for the **whole request** and logs the decision, which is the right side to fail on: legitimate-zero-hunks and cap-truncated must behave identically, so collapsing them to `hunk_count = 0` with a log line for observability is the correct encoding.

Two small things for whenever you are next in this code — neither blocking, neither worth a PR on its own:

**A comment overstates a bound.** `mcp.c:10108-10110` says the mis-scoping failure "is bounded by `detect_collect_seeds`' zero-overlap fallback". That holds only for a *total* miss. If a stale index leaves *some* definitions overlapping the shifted hunks, scoping stays on and the wrong definitions get seeded — the fallback does not bound that case. Worth a clause so a future reader does not over-trust it.

**A rationale that does not match.** `test_mcp.c:6198` says "the assertions below read `base: main`, so pin `init.defaultBranch`". There is no such assertion in either test. Pinning the branch is still right — I am only flagging the stated reason.

One acknowledged gap: the `HUNK_CAP` guard has no binding test, because the cap is a function-local `enum` and nothing can produce 4096 hunks in a fixture. It fails safe in both directions, so I did not hold the PR on it.

Thank you — this was a good find made better by a careful second pass on your own work.

PR #1378, opened 2026-07-31
fix(cypher): make expand_from_bound_terminal OPTIONAL fallback lossless

SEPURI-SAI-KRISHNA:
## What does this PR do?

Makes the OPTIONAL no-match fallback in `expand_from_bound_terminal`
(`src/cypher/cypher.c`) lossless *and* free of false positives — the follow-up
called out in the merge note on #1177.

That function drives a single-relationship `OPTIONAL MATCH` from its already-bound
terminal node — e.g. `MATCH (f:Function) OPTIONAL MATCH (c)-[:CALLS]->(f)`, where `f`
is bound and the start `c` is not. It sized its hop-output buffer `bind_count*10 + 1`
and gated the no-match fallback write on `new_count < max_new`, with `max_new` equal to
the whole buffer. Once one terminal's expansion saturated the buffer, every later
terminal's no-match row was **silently dropped** — exactly the rows
`OPTIONAL MATCH ... WHERE c IS NULL` is meant to surface. Not a memory-safety bug (the
guard kept the write in bounds), but real data loss under saturation, and the same
shape as the fallback issue fixed for the sibling `expand_pattern_rels` in #1177.

**Fix**, in two parts:

1. *Lossless sizing.* Size the buffer `(size_t)*bind_count * CYP_GROWTH_10 +
   (size_t)*bind_count`. The materialised expansion is capped at `max_new =
   *bind_count * 10`; the OPTIONAL fallback emits at most one row per source
   (`<= *bind_count`). A source either matches (feeds the expansion) or takes the
   fallback, never both, so the two counts are additive and bounded by
   `max_new + *bind_count`. Computed in `size_t`; operands `<= INT_MAX`, so no overflow.

2. *Decouple match detection from materialisation* (so the fallback never fabricates a
   false result). The `ti`/`ei` scan loops are no longer gated on the ceiling; **only the
   write** is (`if (new_count < max_new)`). So `match_count` is the true neighbour count,
   and `match_count == 0` at the fallback means the terminal *genuinely* has none — not
   "the buffer filled before we looked." This mirrors the detection/materialisation split
   applied to the sibling `expand_pattern_rels` path.

Behaviour: no OPTIONAL no-match row is dropped; no live terminal is ever mislabelled dead;
a saturated-but-matching terminal is truncated at the per-hop ceiling (bounded success,
unchanged).

**Tests** (reproduce-first, RED→GREEN under ASan+UBSan):

- `cypher_exec_bound_terminal_optional_fallback_survives` — a hub with more incoming
  `CALLS` edges than `max_new` saturates the buffer; a callerless leaf's OPTIONAL no-match
  row must still survive. Fails under the old `+ 1` sizing.
- `cypher_exec_bound_terminal_saturation_no_false_deadcode` — two hubs, each able to
  saturate the buffer alone (so the construction is order-independent; the label-scan query
  has no `ORDER BY`), plus a genuinely callerless leaf. Asserts no hub with callers ever
  appears as an unbound (`c IS NULL`) row while the leaf still does. Fails against
  gated detection, passes only once detection is decoupled from the write.

## Checklist

<sub>Notes: cypher suite 168 passed under ASan+UBSan. The full suite shows 10 failures that
are pre-existing and unrelated to this change (CLI install/uninstall/binary-publish tests
that need a real install target, a subprocess-timeout timing test, and the vendored
integrity check) — the cypher module is green. clang-format/cppcheck were not available in
my local environment; the change mirrors the sibling `expand_pattern_rels` formatting and
stays under 100 columns, so I am relying on CI for the formal lint gate.</sub>

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -4549,21 +4549,40 @@ static void expand_from_bound_terminal(cbm_store_t *store, cbm_pattern_t *patn,
     bool scan_targets =
         !rel_inbound; /* (start)->(term): start = edge source = scan term's inbound */
 
-    size_t alloc_n = (size_t)*bind_count * (size_t)CYP_GROWTH_10 + SKIP_ONE;
+    /* Size this hop's output for BOTH writers without dropping any row: the
+     * materialised expansion is capped at max_new = *bind_count * 10 rows (only
+     * the WRITE is gated on max_new; match DETECTION below is not, so a full
+     * buffer never hides a real neighbour), and the OPTIONAL fallback
+     * emits at most one row per source (<= *bind_count). A source either matches
+     * (feeds the expansion) or takes the fallback, never both, so the two counts
+     * are additive and bounded by max_new + *bind_count. Computed in size_t so
+     * the product cannot overflow.
+     * The previous "+ SKIP_ONE" sizing tied max_new to the whole buffer, so once
+     * the expansion saturated it the fallback guard silently dropped every later
+     * OPTIONAL no-match row (a data-loss bug, not an OOB — the fallback stayed
+     * in-bounds behind that guard) — exactly the rows
+     * `OPTIONAL MATCH ... WHERE <start> IS NULL` is meant to surface. */
+    size_t alloc_n = (size_t)*bind_count * (size_t)CYP_GROWTH_10 + (size_t)*bind_count;
     binding_t *new_bindings = malloc(alloc_n * sizeof(binding_t));
     if (!new_bindings) {
         return;
     }
     int new_count = 0;
-    int max_new = (int)alloc_n;
+    int max_new = *bind_count * CYP_GROWTH_10;
 
-    for (int bi = 0; bi < *bind_count && new_count < max_new; bi++) {
+    for (int bi = 0; bi < *bind_count; bi++) {
         binding_t *b = &(*bindings)[bi];
         cbm_node_t *term = binding_get(b, patn->nodes[1].variable ? patn->nodes[1].variable : "");
         int match_count = 0;
         if (term) {
-            for (int ti = 0;
-                 ti < (rel->type_count > 0 ? rel->type_count : 1) && new_count < max_new; ti++) {
+            /* Detection is decoupled from the write budget: scan every edge and
+             * type unconditionally so match_count reflects the true neighbour
+             * count, and gate only the WRITE on the ceiling (below). If detection
+             * stopped at max_new too, a saturated buffer would leave match_count at
+             * 0 for a terminal that actually has callers, and the fallback below
+             * would fabricate an UNBOUND "dead code" row for live code — worse than
+             * dropping a row. */
+            for (int ti = 0; ti < (rel->type_count > 0 ? rel->type_count : 1); ti++) {
                 cbm_edge_t *edges = NULL;
                 int edge_count = 0;
                 if (rel->type_count > 0) {
@@ -4579,7 +4598,7 @@ static void expand_from_bound_terminal(cbm_store_t *store, cbm_pattern_t *patn,
                 } else {
                     cbm_store_find_edges_by_source(store, term->id, &edges, &edge_count);
                 }
-                for (int ei = 0; ei < edge_count && new_count < max_new; ei++) {
+                for (int ei = 0; ei < edge_count; ei++) {
                     int64_t sid = scan_targets ? edges[ei].source_id : edges[ei].target_id;
                     cbm_node_t found = {0};
                     if (cbm_store_find_node_by_id(store, sid, &found) != CBM_STORE_OK) {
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for picking this up so quickly — and the bug you are fixing is real. I verified it on current main: `cypher.c:4558` sizes `bind_count*10 + 1`, `:4564` ties `max_new` to the whole buffer, and both the outer loop (`:4566`) and the fallback (`:4611`) are gated on `new_count < max_new`, so once one terminal's expansion saturates the buffer every later terminal's no-match row is silently dropped. Your sizing proof is sound, the allocation is safe, the scope is exactly two files, and the security pass is clean.

**But I have to hold it on one finding, because in the regime it targets it now produces a worse answer than main does.**

Removing the two guards makes every binding reachable, which is the goal. The problem is that `new_count < max_new` still gates the `ti` loop (`:4576-4577`) and the `ei` loop (`:4593`). So once `new_count == max_new`, the remaining terminals never fetch their edges at all — `match_count` stays `0` — and the now-ungated fallback at `:4616` emits an unbound row for each of them **regardless of whether they actually have matching neighbours**.

Worked through: three `Function` terminals, `bind_count = 3`, `max_new = 30`. `A` has 40 callers and saturates. `B` has 0 callers and gets a correct fallback row. `C` has 5 callers, its loop is skipped, `match_count` is 0, and it gets a fallback row too. The late `WHERE c IS NULL` filter at `:4749` then keeps exactly those rows — so

```
MATCH (f:Function) OPTIONAL MATCH (c)-[:CALLS]->(f) WHERE c IS NULL
```

reports `C` as dead code when it has five callers. On main, `C` simply did not appear.

That is the trade I cannot take: missing rows become **fabricated** rows, and on the flagship dead-code query a false positive is the one a user acts on destructively. Under-reporting is bad; confidently telling someone live code is dead is worse.

**The fix is three lines, and it keeps everything you have already proved.** Decouple "did this terminal match" from "was there budget to materialise it" — count the match unconditionally, gate only the write:

```c
for (int ei = 0; ei < edge_count; ei++) {
    /* ... find node, label filter (unchanged) ... */
    match_count++;                          /* a real neighbour exists, budget or not */
    if (new_count < max_new) {              /* only the materialisation is capped */
        binding_t nb = {0};
        binding_copy(&nb, b);
        binding_set(&nb, start_var, &found);
        if (rel->variable) { binding_set_edge(&nb, rel->variable, &edges[ei]); }
        new_bindings[new_count++] = nb;
    }
    node_fields_free(&found);
}
```

Drop `new_count < max_new` from the `ti` loop condition too. Your bound proof is unchanged — writes are still gated by `max_new`, fallbacks are still at most one per binding. The cost is that the saturated tail still does its edge lookups, which are the same lookups the unsaturated path already does, and materialises nothing.

**One thing about the test, said carefully because it is a good test.** It genuinely binds and it is genuinely RED on main — I traced it: `hub` is driven first by rowid order, fills the buffer at `new_count = 21`, the outer guard ends the loop, `leaf` is never visited, and your `ASSERT_TRUE(leaf_fallback)` fails. The assertions are specific rather than `row_count > 0`. All good.

The gap is that it is green *through the path the finding condemns*. With your fix, by the time `leaf` is reached `new_count == max_new`, so its edges are never queried and the asserted row comes from the saturation-blind fallback. **The test would pass identically if `leaf` had 500 callers.** It pins "a row with empty `c` appears for leaf", not "leaf's no-match was determined correctly".

A test that discriminates — green on main, RED on this PR as written, green under the fix above — is: insert three Functions in order, `A` with 40 incoming CALLS, `B` with 0, `C` with 5, then

```
MATCH (f:Function) OPTIONAL MATCH (c)-[:CALLS]->(f) RETURN f.name, c.name
```

and assert **no row has `f == "C"` with an empty `c`**. With `max_new = 30`, `A` saturates, `B` falls back correctly, and `C` is the fabricated row.

**Two things that are ours, not yours.** `expand_pattern_rels` has the same fabrication today — its fallback has been ungated since before #1177, and its outer loop never had a budget guard — so merging yours as-is would make the two paths consistently wrong rather than introducing something new. And the underlying question of what the expansion ceiling should *mean* when it is hit is a design call I have taken to the maintainer rather than deciding in review.

This is your fourth fix in this engine after #1173, #1176 and #1177, and the standard has been high in all four. Send the `match_count` change plus the discriminating test and I will pick it straight back up.

SEPURI-SAI-KRISHNA:
Thanks for the precise diagnosis. Updated so match detection is independent of the write ceiling, detection and materialisation are now decoupled.

- The `ti` and `ei` loops no longer stop at `max_new` for OPTIONAL (`... && (new_count < max_new || opt)`), so every real neighbour is counted regardless of buffer state. Only the write is gated (`if (new_count < max_new)`). `match_count == 0` at the fallback therefore means the terminal genuinely has no neighbour, not that the buffer filled first. Non-OPTIONAL never reads `match_count`, so its guard reduces to `new_count < max_new` and keeps the original early-exit.
- A saturated-but-matching terminal is truncated at the ceiling (bounded, as before); a live terminal is never surfaced as dead, and no genuine no-match row is dropped. Buffer sizing is unchanged (`max_new + *bind_count`) and still overflow-free.

New test `cypher_exec_bound_terminal_saturation_no_false_deadcode`: two hubs, each with enough callers to saturate the buffer on its own, so the case holds whichever order the label scan visits them (that query has no `ORDER BY`),  plus a genuinely callerless leaf. It asserts no hub with callers appears as an unbound (`c IS NULL`) row while the leaf still does; it passes with the decoupled detection and fails without it. Full cypher suite green under ASan+UBSan.

SEPURI-SAI-KRISHNA:
Small refinement after seeing #1385: dropped the `|| opt` early-exit and made detection fully unconditional (the `ti`/`ei` loops no longer reference the budget at all, only the write is gated), so this path now reads identically to the `expand_pattern_rels` fix. Same behaviour, just consistent shape across the two. Force-pushed.

DeusData:
Thank you for addressing the earlier OPTIONAL-fallback finding and updating the PR. It is now routed as a high-priority Cypher correctness fix in `0.9.1-rc` and is back in the maintainer review queue. The queue is currently full, so the final re-review may take a little time.

DeusData:
Merged — thank you @SEPURI-SAI-KRISHNA. This is a good catch and a well-argued one.

The distinction you drew is the important part: the old guard kept the write in bounds, so this was never a memory-safety bug — it was **silent data loss**, and specifically of the rows `OPTIONAL MATCH ... WHERE c IS NULL` exists to produce. A query that quietly returns fewer rows under saturation is worse than one that fails, because nothing signals that the answer is incomplete.

Sizing the allocation as `bind_count * 10 + bind_count` so the fallback has its own headroom — rather than letting the expansion budget consume the whole buffer — is the right shape, and computing it in `size_t` closes the overflow question before anyone has to ask it.

Also appreciated that you picked up the follow-up #1177's merge note actually called for, rather than leaving it as a comment nobody returned to.

Shipping in the next release.

PR #662, opened 2026-06-28
fix(build): rebuild lsp_all.o when its included headers change

KerseyFabrications:
Fixes #661

### Problem
On an **incremental** build, `make` can link a **stale `lsp_all.o`**. ASan surfaces it as a stack-buffer-underflow during Rust-LSP tests:

```
ERROR: AddressSanitizer: stack-buffer-underflow ... in __interceptor_memcpy
  #1 cbm_calls_push internal/cbm/cbm.c:128
  #2 rust_inject_syn_call internal/cbm/lsp/rust_lsp.c:3366
  ...
  'call' (line 3363) <== Memory access at offset 0 partially underflows this variable
```

### Root cause
`lsp_all.o` is a unity object that `#include`s every `lsp/*.c` resolver; those include `cbm.h`, and `CBMCall` is copied **by value** across the lsp→pipeline boundary (`cbm_calls_push`). Its Make rule depended only on `lsp_all.c`, **not** on `cbm.h`. After `CBMCall` gained `bool is_method` (#477), incremental builds kept a stale `lsp_all.o` compiled against the old 304-byte layout, while the rest of the tree recompiled against the new 312-byte layout — an ODR mismatch, so the by-value struct copy under-reads the shorter stale struct.

Clean builds (CI) always recompile both sides together, so the layouts match and CI never sees it. It only bites incremental developer builds after the struct changed.

### Fix
Add an `LSP_UNITY_DEPS` prerequisite list (`lsp/*.c`, `lsp/*.h`, `lsp/generated/*.c`, `internal/cbm/*.h`) to both `lsp_all.o` and `prod_lsp_all.o`, so the unity object rebuilds when any included header or unity source changes. Wildcard-based, so new LSP resolvers (e.g. an added `perl_lsp.c`) are covered automatically. Explicit prerequisites rather than `-MMD` auto-deps, to match the Makefile's existing style.

### Verification
- `touch internal/cbm/cbm.h` → `lsp_all.o` now rebuilds (it didn't before); no-op rebuild is a clean no-op.
- Clean build + full ASan test suite green; `rustlsp_dbg_macro_inner_call` no longer aborts.

PR #1380, opened 2026-07-31
fix(store): route label allowlists through the canonical type-like set

DeusData:
Distilled from #568 by @win4r, with credit on the commit.

## The bug

`cbm.h` documents `cbm_label_is_type_like()` as the single source of truth for type-like labels — *"so adding a new type-like label (e.g. `Struct` for Rust/Go/Swift/D structs) updates them all at once instead of scattering `|| strcmp(label,"Struct")==0` across the tree."*

A SQL string literal cannot call it. Five queries carried their own hardcoded label lists and silently opted out of that contract:

| Site | Hardcoded list |
|---|---|
| `store.c:5279` `arch_boundaries` | `('Function','Method','Class')` |
| `store.c:5422` `arch_packages_from_qn` | `('Function','Method','Class')` |
| `store.c:6689` `arch_clusters` | `('Function','Method','Class')` |
| `store.c:7742` `cbm_store_vector_search` | `('Function','Method','Class')` |
| `mcp.c:2759` BM25 ranking | `('Class','Interface','Type','Enum')` |

**On main today:** `get_architecture` (boundaries, packages, clusters) and vector search **drop every `Struct`, `Interface`, `Enum`, `Type` and `Trait` node in the project**, and `search_code` under-ranks structs. For Rust, Go, Swift and D — where `struct` is the primary type declaration — that means the architecture view omits most of the codebase's types.

Note the BM25 list had drifted *differently* again: it includes `Interface` but omits `Struct` and `Trait`. Five hand-maintained lists, three different answers.

## The fix

Not a wider literal — that would just reset the same trap. `CBM_SQL_TYPE_LIKE_LABELS` and `CBM_SQL_CALLABLE_OR_TYPE_LABELS` live in `constants.h` (the one header both `store.c` and `mcp.c` already include) and are **pinned to the C predicate by a test that checks both directions**:

- every label `cbm_label_is_type_like()` accepts appears in the SQL fragment, and
- nothing it rejects is smuggled in, so the SQL cannot silently widen past the C contract.

Adding a type-like label without updating the SQL now fails that test instead of quietly shrinking query results.

**No new labels are introduced.** Every label added here is one the extractors already emit today.

## Verification

- `store_nodes`, `mcp`, `store_search` — **314 passed, 2 skipped**
- `make -f Makefile.cbm lint-ci` — clean
- **Revert-check:** shrinking `CBM_SQL_TYPE_LIKE_LABELS` back to `'Class'` fails the pin at `test_store_nodes.c:36`. Worth noting *how* that was confirmed — the first revert-check passed spuriously because **`constants.h` is not tracked as a build dependency**, so nothing rebuilt. Forcing a rebuild produced the expected RED. That header-dependency gap is a separate issue and is being reported on its own.

## Credit

@win4r's audit in #568 found four of the five sites — the finding is theirs. Their patch widened the literals in place and added `Actor`, a label main does not emit; this routes through the canonical set instead, and also covers `Interface`, `Type` and `Trait`, which their patch still omitted.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -417,6 +417,19 @@ PROD_SRCS = $(FOUNDATION_SRCS) $(STORE_SRCS) $(CYPHER_SRCS) $(MCP_SRCS) $(DAEMON
 EXISTING_C_SRCS = $(EXTRACTION_SRCS) $(LSP_SRCS) $(TS_RUNTIME_SRC) \
                   $(GRAMMAR_SRCS) $(AC_LZ4_SRCS) $(ZSTD_SRCS) $(SQLITE_WRITER_SRC)
 
+# Project headers, as prerequisites for the targets built in one compiler
+# invocation from sources (test-runner, test-repro-runner, test-runner-tsan,
+# codebase-memory-mcp, test-foundation). Those rules list only .c files, so
+# editing a header changed nothing make could see and the target was NOT
+# rebuilt — an incremental build then silently tested the OLD header value.
+# Found while revert-checking a constants.h change: the check passed because
+# nothing had recompiled.
+#
+# Vendored trees are deliberately excluded — they are pinned, they carry their
+# own explicit dep lists (LSP_UNITY_DEPS, TS_RUNTIME_DEPS), and globbing tens of
+# thousands of grammar headers would cost more than it protects.
+PROJECT_HDRS = $(wildcard src/*.h src/*/*.h $(CBM_DIR)/*.h tests/*.h tests/repro/*.h)
+
 # ── Test sources ─────────────────────────────────────────────────
 
 TEST_FOUNDATION_SRCS = \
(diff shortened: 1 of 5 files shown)


DeusData:
Added a second commit: **the build now rebuilds when a project header changes.**

This is the gap the PR description mentioned as "being reported on its own" — it turned out to belong here, because it is what made the first revert-check of *this* change lie.

`test-runner`, `test-repro-runner`, `test-runner-tsan`, `codebase-memory-mcp` and `test-foundation` are each built in a single compiler invocation from their sources, and their rules listed **only `.c` files** as prerequisites. Editing a header changed nothing make could see, so the target was never rebuilt and an incremental build silently kept testing the **old** header value.

That is exactly what happened here: shrinking `CBM_SQL_TYPE_LIKE_LABELS` back to `'Class'` should have turned the new pin test red, and it did not — because nothing had recompiled. Forcing a rebuild produced the expected failure. A revert-check that cannot fail is worse than no revert-check at all, since it reports confidence it never earned.

Same class as the `lsp_all.o` staleness #662 fixed with an explicit dependency list; that fix was per-object, this one covers the whole-program targets.

`PROJECT_HDRS` globs `src/`, `internal/cbm/` and `tests/` headers. Vendored trees are deliberately excluded — they are pinned, they already carry their own explicit dep lists (`LSP_UNITY_DEPS`, `TS_RUNTIME_DEPS`), and globbing tens of thousands of grammar headers would cost more than it protects.

**Verified the fix directly:** editing only `src/foundation/constants.h`, with no source file touched, now rebuilds and turns the pin test red — the exact edit that was invisible to make before.

Re-ran after restoring: **551 passed, 2 skipped** across `store_nodes`, `mcp`, `store_search` and `pipeline`; `lint-ci` clean.

Issue #1355, opened 2026-07-30
Unresolvable calls are bound to any same-named symbol (Jest `describe`, RxJS `take`) and trace_path exposes no confidence/strategy


artaommahe:
### Version

codebase-memory-mcp 0.9.1-rc.1

### Platform

macOS (Apple Silicon)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

When a called name can't be resolved to a definition, the `unique_name` / `qualified_suffix` / `suffix_match` fallbacks bind it to **any same-named symbol anywhere in the project** — including test-runner globals and library operators that have no relationship to the matched symbol. `trace_path` then reports these as callers with **no confidence or strategy column**, so the caller list reads as fact.

Minimal case: a Jest/Jasmine `describe('...', () => {})` block in a `.spec.ts` file becomes a `CALLS` edge to an application method named `describe` in an unrelated file, at confidence 0.75.

At scale on a private Angular monorepo (11,788 files, 63,840 nodes / 120,805 edges), the entire top of the inbound-call ranking is fabricated:

| callee reported by the graph | inbound `CALLS` | what the call sites actually are |
|---|---|---|
| `ScheduleOpenShiftViewComponent.take` | 457 | the RxJS `take(n)` pipe operator (486 `take(` occurrences repo-wide; `take` imported from `rxjs/operators` in 326 files) |
| `ReportService.describe` | 377 | the Jest `describe()` global (794 spec files) |
| `ReportApi.describe` | 275 | same |
| `PersonalScheduleComponent.date` | 264 | see below |
| `PermissionDirective.permission` | 207 | unrelated `permission` references |

Filtering on confidence does not rescue it: with `WHERE r.confidence >= 0.9`, the top callee is still `PersonalScheduleComponent.date` with 250 inbound edges.

Two of those recorded "call sites" are not calls at all. I opened the source and verified both:

- An `@Input() set date(date: string | Date) { ... }` **setter declaration** in file A is recorded as a `CALLS` edge (strategy `import_map`) to a `set date(...)` setter in unrelated file B. The edge's `line` property points at the setter's own declaration line, not at any call.
- `lazySelect(getTimesheet(this.timesheet.id))` — where `this.timesheet` is a **property access** — is recorded as a `CALLS` edge to a method named `timesheet` on an unrelated API class.

[...]

PR #1386, opened 2026-07-31
fix(python): suppress weak same-name CALLS for get/run/execute

Joseph-MingEn:
## Summary
- Suppress weak short-name CALLS resolution for Python attribute calls and bare generic callees (`get` / `run` / `execute`), mirroring the existing TS/JS weak-method guard.
- Flag Python `attribute` calls as `is_method` during extraction so `prior_cp.get(...)` cannot `suffix_match` onto unrelated Methods (e.g. `_SessionRegistry.get`).
- Same guard drops bare `run()` Callable-parameter calls from binding to unrelated Methods (e.g. `SatoriLive.run`).

## Motivation (Yui WP-B C2 / G2)
On a large Python repo, `trace_path` hop-1 showed fabricated CALLS:
- `router.submit_task` → `_SessionRegistry.get` via `strategy=suffix_match` / `callee=prior_cp.get` / conf `0.28`
- `gate._run_with_heavy_slot` → `SatoriLive.run` via `suffix_match` / bare `run()` / conf `0.28`

Strong strategies (`import_map`, `same_module`, `lsp_*`) are kept.

## Test plan
- [x] `./build/c/test-runner registry` (includes new python suppress unit tests)
- [x] `./build/c/test-runner pipeline` (includes `pipeline_python_suppresses_weak_generic_edges`)
- [ ] Maintainer CI green

### DIFF
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -2244,6 +2244,17 @@ void handle_calls(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec, Walk
                     }
                 }
             }
+            /* Python attribute calls (prior_cp.get(...)): flag as method so weak
+             * short-name resolution can be suppressed (G2 / WP-B C2), mirroring
+             * the TS/JS receiver-aware guard above. Bare calls stay false. */
+            if (ctx->language == CBM_LANG_PYTHON &&
+                strcmp(ts_node_type(node), "call") == 0) {
+                TSNode fn = ts_node_child_by_field_name(node, TS_FIELD("function"));
+                if (!ts_node_is_null(fn) &&
+                    (strcmp(ts_node_type(fn), "attribute") == 0)) {
+                    call.is_method = true;
+                }
+            }
 
             TSNode args = ts_node_child_by_field_name(node, TS_FIELD("arguments"));
             // ObjectScript stores args under oref_method/method_args, not the
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -421,7 +421,7 @@ static void emit_classified_edge(cbm_pipeline_ctx_t *ctx, const CBMCall *call,
         return;
     }
     if (suppress_plain_calls) {
-        return; /* weak TS/JS member-call match with an unresolved receiver (#606) */
+        return; /* weak TS/JS or Python same-name match (#606 / G2) */
     }
     char esc_c2[CBM_SZ_256];
     cbm_json_escape(esc_c2, sizeof(esc_c2), call->callee_name);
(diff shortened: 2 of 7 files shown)


DeusData:
Thank you for the focused Python same-name fallback suppression and the tests around generic method names. This is now routed as a high-priority parsing fix in `0.9.1-rc` and linked to the remaining scope in #1355. This is triage rather than approval; our review queue is full, so the detailed review may take a little time, especially because the PR addresses only one part of the broader binding problem.

DeusData:
The false-positive class is real (prior_cp.get resolving to an unrelated _SessionRegistry.get is exactly the #1276 family) and mirroring the TS/JS guard is the right instinct. Three things before this can merge: (1) mechanical — the branch conflicts, DCO and lint are red, and all three pr-smoke legs fail; rebase + sign-off + lint first. (2) The blanket is_method=true for every Python attribute call in extract_calls.c changes semantics for every other is_method consumer — narrow it or make the case. (3) The hardcoded {get,run,execute} blacklist plus blanket weak-member suppression is a recall cut on a priority language: our #1000 lesson is that graph-shape changes need corpus-scale evidence, so please include Django-scale before/after edge counts (we can help run them once the branch is green). Related work just landed you should build on: #1647 now suppresses cross-language suffix_match CALLS.

DeusData:
Closing this one, and I want to be precise about why, because **you identified a real defect class that nothing else in the codebase catches — and half of your diagnosis is being built now with credit to you.**

## What happened to each half

**The member half is now covered.** #1903 landed on main (`5d96377a`) doing the same thing your PR did for `obj.method()` calls: Python receivers that are not `self`/`cls`/`super()` nor import-rooted get `is_method`, enabling the existing weak-member guard against `suffix_match`/`unique_name`/`field_type_hint`/`fuzzy`. Your motivating case — `prior_cp.get(...)` with `prior_cp` a parameter — is covered identically.

The difference is exemptions. Your version flags **every** Python attribute call with none, which also drops `self.inherited()` resolving across files and `helper.compute()` / `pkg.sub.fn()` — Python's canonical cross-file call shape. Those are true edges, and it was the exemption list that made the difference, not the suppression predicate, which is byte-identical in both.

**The bare-call half is genuinely yours, uncovered by anything, and is being implemented.** After #1903:

```python
def _run_with_heavy_slot(run):
    return run()          # still suffix_matches onto SatoriLive.run
```

#1903 structurally cannot catch this — there is no receiver, so the guard never fires. I checked whether anything else does: #1647's cross-language suffix guard only fires when caller and target languages differ, and both are Python here. So this class survives on main today, and you are the only person who found it.

## The one change to your mechanism

The implementation keys on **local binding** rather than the `{get, run, execute}` list: suppress when the callee identifier is a parameter of the enclosing function, because a local binding shadows any project function and short-name resolution is therefore fabricated *by construction*.

The reasoning against the list is not that it is wrong today — it is that it is a claim about spellings rather than about whether the resolver knew anything, and it would age **invisibly**. Nothing fails when the generic-name distribution shifts; the graph just quietly loses different edges. Local binding covers `handler()`, `callback()`, `fn()` and every Callable-parameter shape without one, and it is decidable rather than empirical. That work carries `Co-authored-by:` credit to you.

## Why this PR itself cannot be the vehicle

Practical, not a judgement on the work: it does not compile against current main (#1903 renamed `cbm_tsjs_suppress_weak_method_match`, which this calls), the branch is four weeks behind, and the commit carries no `Signed-off-by` trailer so DCO blocks it regardless. Rebasing across that to land a member half now redundant would be mostly discarded work.

## Credit where it is due

You have two PRs here and both found real things. On #1371 you flagged the recall risk **on your own PR**, unprompted — that is rare and it is why the before/after numbers got run and it merged. Here you found a false-edge class that survives every guard currently on main. The diagnosis is yours even though the mechanism changed.

Thank you, genuinely. If you would rather carry the bare-call implementation yourself, say so and I will hand it straight back.

Issue #598, opened 2026-06-24
[bug] Route extractor over-fits: GraphQL query literals, whole config-file contents, and regex literals become `Route` nodes


ecosuper2025:
**Version:** codebase-memory-mcp v0.8.1 (Windows, compiled .exe)
**Project:** real-world repo, 93,315 nodes, 996 `Route` nodes

## Summary
`Route` extraction classifies many non-route strings as routes. On our repo, 46 of 996 routes are clearly junk, and the empty-`source` bucket mixes real HTTP endpoints with regex/path string literals.

## Repro
`search_graph(label="Route")` then group by `source`:

| `source` | count | what they actually are |
|---|---|---|
| `decorator` | 41 | ✅ real Flask/Django route decorators (correct) |
| `""` (empty) | 909 | real API paths from frontend HTTP calls **+** regex/path literals |
| `graphql` | 34 | ❌ entire GraphQL **query string literals** (`query { categoryList {...} }`) turned into one Route node each |
| `infra` | 12 | ❌ the **entire text of `.codex/agents/*.toml` files** turned into a single Route node |

Empty-source junk samples (`method=ANY`, degree 0):
`/.{2}/g`, `/<table/i` (JS regex literals), `/_VBA_PROJECT_CUR/VBA/dir`, `/::/`, `/.well-known/openid-configuration`.

## Minimal example
```js
// a JS regex literal — NOT a route
const re = /<table/i;
```
```graphql
# a query string passed to a client — NOT a route
query { products(search: "x") { items { sku } } }
```
Both currently produce `Route` nodes.

## Expected
- Route extraction should require an actual route **declaration** (decorator/router registration/framework binding), not "string contains slashes" or "file looks like config".
- GraphQL query literals should not be `Route` (a `GraphQLOperation` label would be fine).
- Config/agent files (`.codex/*.toml`, `*.md`) should not have their whole contents emitted as a single Route.
- At minimum: don't classify regex literals (`/…/[gimsuy]`) as routes.

## Workaround (consumer side)
Filter `WHERE r.source='decorator' OR (r.method IS NOT NULL AND r.method <> 'ANY')` → 996 → 347, junk removed. But this loses real decorator-less routes inconsistently; a clean extractor is preferable.

## Related
Distinct from the exclude-config issues (#500/#510) — this is mis-classification of *content*, not indexing of build artifacts.

DeusData:
[...]

PR #1389, opened 2026-07-31
fix(pipeline): reject non-HTTP slash arguments

Yyunozor:
## Summary

- apply the existing HTTP route-literal guard before creating `arg_url` routes
- reject raw slash expressions such as JS regex literals
- reject slash strings passed to non-HTTP string and regex consumers
- add focused regressions for filesystem paths and a genuine API path

## Root cause

`detect_url_in_args()` normalized every slash-prefixed call argument and
created a `Route` node directly. That path bypassed the shared route-literal
guard already used by the later route-node pass, so local filesystem paths and
regex operands could become `Route` nodes and `HTTP_CALLS` edges.

The change first rejects slash-prefixed raw expressions (string literals and
propagated constants are carried separately), routes the remaining heuristic
through the shared guard, and extends the callee filter to non-HTTP string
consumers such as `replace`, `match`, `search`, `test`, and `exec`.

## Validation

- `scripts/test.sh --suites 'infrascan pipeline'` — 241 passed
- the same tests on the parent fail on `sink(/<table/i)` and
  `template.replace('/html/g', ...)`
- full suite — 6821 passed, 3 failed, 4 skipped
- the parent reproduces the same failures: `cli` (255 passed, 2 failed) and
  `daemon_runtime` (42 passed, 1 failed)

This addresses the `arg_url` regex/path bucket of #598. GraphQL and
whole-config-text classification remain out of scope.

### DIFF
--- a/internal/cbm/service_patterns.c
+++ b/internal/cbm/service_patterns.c
@@ -678,7 +678,11 @@ static bool callee_is_delimiter_or_filesystem_builder(const char *callee_name) {
         method = last_colon + 2;
     }
     if (strcmp(method, "split") == 0 || strcmp(method, "rsplit") == 0 ||
-        strcmp(method, "partition") == 0 || strcmp(method, "join") == 0) {
+        strcmp(method, "partition") == 0 || strcmp(method, "join") == 0 ||
+        strcmp(method, "replace") == 0 || strcmp(method, "replaceAll") == 0 ||
+        strcmp(method, "match") == 0 || strcmp(method, "matchAll") == 0 ||
+        strcmp(method, "search") == 0 || strcmp(method, "test") == 0 ||
+        strcmp(method, "exec") == 0) {
         return true;
     }
     return strstr(callee_name, "os.path.join") != NULL || strstr(callee_name, "path.join") != NULL;
--- a/internal/cbm/service_patterns.h
+++ b/internal/cbm/service_patterns.h
@@ -44,6 +44,10 @@ cbm_svc_kind_t cbm_service_pattern_match(const char *resolved_qn);
  * classified via its real resolved QN instead and never reaches this check. */
 bool cbm_service_pattern_is_global_fetch(const char *callee_name);
 
+/* True when a string literal is a plausible HTTP route for the given callee.
+ * Rejects filesystem paths and non-HTTP string consumers. */
+bool cbm_service_pattern_is_http_route_literal(const char *literal, const char *callee_name);
+
 /* Per-worker TLS cache for cbm_service_pattern_match results. The
  * pattern matcher runs once per resolved CALL edge in emit_service_
  * edge — that's 6 pattern lists × ~30 patterns × strstr per call ≈
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -1792,12 +1792,22 @@ static void detect_url_in_args(cbm_gbuf_t *gbuf, const cbm_gbuf_node_t *source,
                                const CBMCall *call) {
     for (int ai = 0; ai < call->arg_count; ai++) {
         const CBMCallArg *ca = &call->args[ai];
+        /* A slash-prefixed raw expression is not a URL string. In JS/TS this
+         * is notably a regex literal (`/<table/i`); genuine string literals
+         * and propagated constants are carried in `value`, while template
+         * literals keep their leading backtick in `expr`. */
+        if (!ca->value && ca->expr && ca->expr[0] == '/') {
+            continue;
+        }
         const char *url = ca->value ? ca->value : ca->expr;
         if (!url || (url[0] != '/' && url[0] != '`')) {
             continue;
         }
         char norm[CBM_SZ_256];
         if (!normalize_url_arg(url, norm, (int)sizeof(norm))) {
+            continue;
+        }
+        if (!cbm_service_pattern_is_http_route_literal(norm, call->callee_name)) {
             continue;
         }
         char route_qn[CBM_ROUTE_QN_SIZE];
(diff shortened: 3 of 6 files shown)


DeusData:
Thank you for narrowing the route heuristic to HTTP-shaped slash arguments and for adding focused extraction coverage. This is now routed as a high-priority parsing fix in `0.9.1-rc`. This note is an intake acknowledgement rather than a code review; the queue is currently full, so validating false-positive and false-negative behavior across frameworks may take a little time.

DeusData:
The substance here is merge-worthy: detect_url_in_args minting Route nodes from JS regex literals and filesystem paths is the arg_url bucket of #598, and routing it through the shared cbm_service_pattern_is_http_route_literal guard is the right unification. Your documented pre-existing-failure validation was honest work. It needs only a rebase — pass_parallel.c moved under you in August, and GH now reports conflicts; your green checks predate that drift. Two smalls while you're in there: confirm the delimiter-callee extension (replace/match/search/test/exec) tightening pass_route_nodes via the shared guard is intended (we think it's correct — same FP class), and drop the now-redundant local extern at pass_route_nodes.c:40 in favor of the new header declaration. Rebase and we'll fast-track the re-review.

DeusData:
@Yyunozor — a nudge plus some good news that should make the rebase easier than it looked on 17 August.

**The conflict has shrunk to one file.** Re-checked against today's `main` (`2cc80f20`): `pass_parallel.c` now auto-merges cleanly — the only conflict is `tests/test_pipeline.c`, and it is pure insertion-anchor drift (the file grew a few thousand lines under you). Re-anchor your new e2e test and its `RUN_TEST` line and the rebase is done.

The two smalls from that comment still stand, and they are genuinely small:
1. **Confirm the delimiter-callee extension is intended** to also tighten the two guard call sites in `pass_route_nodes.c` (:182, :684) — we believe it is correct, same false-positive class, but it is your extension and the confirmation should be yours.
2. **Drop the now-redundant local extern** at `pass_route_nodes.c:40` — your header declaration in `service_patterns.h` supersedes it.

The substance was ruled merge-worthy on 17 August and the fast-track promise holds: rebase, and the re-review happens promptly. Your pre-existing-failure validation on the parent tree was, as said then, honest work — for what it's worth, the three unrelated failures you documented have since been confirmed as known flakes on our attribution ledger, independently of your PR.

So this does not sit open-ended: **if we have not heard from you by ~11 September**, we will do the rebase and the two smalls ourselves and land it with `Co-authored-by:` credit — the fix is priority/high and #598's junk routes are worth removing promptly either way. Whichever way it lands, the work is yours.

Yyunozor:
Rebased onto `main` (today's tip). `pass_parallel.c` auto-merged cleanly. `tests/test_pipeline.c` had exactly the one conflict you called out — pure insertion-anchor drift from the three tests that landed under our anchor in August. Re-anchored the new test and its `RUN_TEST` line right after them; the diff this commit carries is otherwise identical to before (same 85 insertions, same 1 deletion, verified line-for-line).

Small #2: dropped the redundant local `extern` in `pass_route_nodes.c` (:40), included `service_patterns.h` instead — the same pattern `pass_calls.c` and `pass_parallel.c` already use.

Small #1, confirmed: `route_edge_visitor` (:181) and `http_call_edge_has_valid_route` (:683) call `cbm_service_pattern_is_http_route_literal` with the same two arguments as our new `detect_url_in_args` check, so the extension tightens all three call sites by construction — checked with a standalone before/after build of the predicate: `is_http_route_literal("/html/g", "template.replace")` is `true` on the parent, `false` after. Today it's defense in depth rather than a change visible in a fresh index (`detect_url_in_args` is the only current producer of that callee shape, and it already self-gates at creation), but it matters for graphs built by an older binary and re-checked on an incremental run, so worth keeping.

`scripts/test.sh --suites 'infrascan pipeline'` on the rebased branch: 267 passed, 0 failed. Full `scripts/test.sh`: 7770 passed, 2 failed, 7 skipped (141 suites) — `subprocess_cancel_grace_is_hard_capped` and one `test_cli.c` VS Code/Copilot-detection case, neither touched by this diff. The subprocess one passed on two reruns (timing flake). The `test_cli.c:9491` one fails identically on today's bare `main` in the same environment, so it is environmental, not from this branch. `daemon_runtime` came back clean this run. Thanks for confirming the pre-existing failures on your side — saved me from chasing them.

adfjadfj16-a11y (approved):

DeusData:
Merged as `3627eff5`. The fast-track promise from 17 August finally honoured — two weeks later than it should have been, and that part is ours.

Both confirmations were answered properly, and the first one was answered the right way:

**Small #1.** You did not just assert the extension was intended — you built the predicate standalone and showed it: `is_http_route_literal("/html/g", "template.replace")` is `true` on the parent and `false` after. That converts "we think it's correct, same FP class" from a shared hunch into a measured fact, and it establishes the tightening applies to `route_edge_visitor` (:181) and `http_call_edge_has_valid_route` (:683) *by construction* rather than by inspection. Noting that it is defence in depth today, since `detect_url_in_args` is the only current producer of that callee shape, is exactly the right caveat to attach.

**Small #2** is done and follows the existing convention — `service_patterns.h` included the way `pass_calls.c` and `pass_parallel.c` already do, rather than a second local extern.

**And you verified the rebase rather than trusting it.** "The diff this commit carries is otherwise identical to before — same 85 insertions, same 1 deletion, verified line-for-line" is the claim that let me merge this without re-reviewing the substance: the arithmetic checks out, since the only delta is small #2's `+1/−2`. Re-anchoring a test after a few thousand lines drifted under it is precisely where a silent change sneaks in, and you closed that off.

The underlying fix — `detect_url_in_args` minting Route nodes from JS regex literals and filesystem paths, routed through the shared guard — was ruled merge-worthy on day one. Everything since was drift and our latency, not your work.

Yyunozor:
Thanks for the merge, and for the clear pointers at every step — the one-file conflict note made the rebase trivial. No worries on the timing. #1766 is rebased on current main and green on the targeted suites, whenever you get to it.

DeusData:

PR #1394, opened 2026-08-01
fix(cypher): unify repeated variable-length nodes

jstar0:
## What does this PR do?

Part of #797. This is the repeated-node-variable fix split from #883 as requested in maintainer review.

Variable-length expansion currently overwrites an existing terminal-node binding with each BFS result. As a result, a pattern such as `MATCH (f)-[:CALLS*1..2]->(f)` can return paths ending at a different node even though both `f` occurrences must identify the same node.

The change applies the same binding-unification rule already used by fixed-length relationships: when the terminal variable is bound, variable-length expansion only accepts traversal results with the same node id.

This PR intentionally does not include the trail-enumeration semantics, BFS fork, row cap, or related tests from #883.

## Verification

```sh
make -f Makefile.cbm test
make -f Makefile.cbm lint-ci
make -f Makefile.cbm security
```

The focused Cypher suite passes 171 tests. The repository lint and all eight security layers pass. The full test target reaches and passes the changed Cypher suite; unrelated subprocess and install-staging failures reproduce on the exact unmodified `main` base in the same environment.

## Checklist

- [x] Every commit is signed off (`git commit -s`) — required, CI rejects unsigned commits
- [x] New behavior is covered by a reproduce-first test
- [x] Lint passes
- [x] Security checks pass

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -3151,12 +3151,17 @@ static void expand_var_length(cbm_store_t *store, cbm_rel_pattern_t *rel,
     cbm_traverse_result_t tr = {0};
     const char *dir = rel->direction ? rel->direction : "outbound";
     cbm_store_bfs(store, src->id, dir, rel->types, rel->type_count, max_depth, CBM_PERCENT, &tr);
+    cbm_node_t *bound_to = binding_get(b, to_var);
+    int64_t bound_to_id = bound_to ? bound_to->id : 0;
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for the repeated-variable-length-node fix and its Cypher regression coverage. It is now routed as a high-priority query-correctness PR in `0.9.1-rc`. This is triage rather than approval; our review queue is currently full, so the full semantic and performance review may take a little time.

PR #1400, opened 2026-08-01
fix(extract): resolve Go route paths built via string concatenation

### LINKED ISSUE #1249: Go: mux registrations built by runtime concatenation (prefixVar + "literal") are not resolved — most of a service's routes missing
## What is broken

Route registrations whose path is a **compile-time-obvious concatenation** of a config variable and a string literal are not indexed, so the graph misses most of the service's surface.

## Minimal repro (Go, gorilla/mux)

```go
r := mux.NewRouter()
r.HandleFunc(settings.V.BaseURL+"/login", loginHandler)          // MISSED
r.HandleFunc(settings.V.BaseURL+"/v2/login", loginV2Handler)     // MISSED
r.HandleFunc(settings.V.BaseURL+"/widget/{rest:.*}", widget)     // MISSED
r.HandleFunc("/health", health)                                   // indexed OK
```

## Observed impact

In a real BFF with 47 `HandleFunc` registrations (42 unique paths), all of them using the `BaseURL+"literal"` pattern, the graph contained only 9 Route nodes — and several of those came from logger labels rather than registrations (see #1248). The same limitation hides outbound calls built the same way (`url := "http://" + host + ":" + port + "/log"` produced no HTTP_CALLS edge), so a real cross-service seam was invisible.

## Expected

When the non-literal part is a simple identifier/selector (config prefix) and the rest is a literal, index the literal suffix as the route path (optionally recording the dynamic prefix in a property, e.g. `{"prefix_var": "settings.V.BaseURL"}`). This pattern is idiomatic in Go services that mount everything under a configurable base path.

AmirF194:
## What does this PR do?

`r.HandleFunc(base+"/login", loginHandler)` and outbound calls built the same way (`"http://" + host + ":" + port + "/log"`) were not indexed: `extract_positional_url` (internal/cbm/extract_calls.c) has no branch for `binary_expression`, the AST node Go's `+` produces, so the argument falls through and `first_string_arg` stays NULL. On the reporter's real BFF (47 `HandleFunc` registrations, all `BaseURL+"literal"`), the graph contained only 9 Route nodes.

Added `extract_binary_concat_suffix`: for a `+` concatenation, recover the literal string on the right-hand side (the config-prefix pattern this issue describes). When the right side is not itself a literal, the call is left unresolved rather than guessing a path, matching the reporter's own expected behavior.

## Checklist

## Verification

- New regression test (`extract_go_binary_concat_url_issue1249`) fails on unmodified `main` (`first_string_arg is NULL`) and passes on this branch, covering both the mux registration and the outbound-URL case from the issue.
- A second test confirms a concatenation whose right side is not a literal (`base+suffixVar`) still resolves to no route, so nothing is fabricated.
- Full `extraction` suite (273 tests) passes under ASan+UBSan; `scripts/lint.sh --ci` (cppcheck + clang-format) is clean on the changed files.
- Not run: the full sanitized suite outside `extraction`, and the macOS/Windows CI legs (no such runners here).

Fixes #1249

### DIFF
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -1707,6 +1707,24 @@ static const char *extract_keyword_url(CBMExtractCtx *ctx, TSNode arg) {
     return extract_string_value(ctx, val_node);
 }
 
+// `prefixVar + "/route"` (Go's idiomatic configurable-base-path pattern):
+// recover the literal suffix. A right side that is not itself a literal is
+// left unresolved rather than guessed (issue #1249).
+static const char *extract_binary_concat_suffix(CBMExtractCtx *ctx, TSNode node) {
+    TSNode op_node = ts_node_child_by_field_name(node, TS_FIELD("operator"));
+    if (!ts_node_is_null(op_node)) {
+        char *op = cbm_node_text(ctx->arena, op_node, ctx->source);
+        if (!op || strcmp(op, "+") != 0) {
+            return NULL;
+        }
+    }
+    TSNode rhs = ts_node_child_by_field_name(node, TS_FIELD("right"));
+    if (ts_node_is_null(rhs) || !is_string_like(ts_node_type(rhs))) {
+        return NULL;
+    }
+    return strip_and_validate_string_arg(ctx->arena, cbm_node_text(ctx->arena, rhs, ctx->source));
+}
+
 // Try to extract URL/topic from a positional argument (string or constant).
 static const char *extract_positional_url(CBMExtractCtx *ctx, TSNode arg, const char *ak) {
     /* JS/TS template literals: `/things/${id}` normalizes to "/things/{}" so the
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for the Go route-path contribution and the concatenation-focused regression coverage. It is now triaged as a high-priority parsing fix in `0.9.1-rc`. This is an intake acknowledgement rather than a code review. The review queue is currently full, so it may take us a little time to validate constant evaluation and make sure other route extractors are unaffected.

AmirF194:
`security / codeql-gate` and its downstream `ci-ok` both failed only because the gate job timed out waiting on CodeQL, not because of anything in this diff. Its own wait step ran 06:24:18Z to 07:10:02Z (the 45 minute budget), and the `CodeQL SAST` workflow run on this same commit (478ce25c) did not finish until 07:40:54Z, about 30 minutes after the gate gave up. CodeQL itself passed cleanly on that run.

This diff only touches Go route-path parsing, nothing CI-related. Re-running the `codeql-gate` job should now pick up the completed, passing CodeQL result.

PR #1403, opened 2026-08-01
fix(search): index docstrings and accept non-ASCII query bytes

### LINKED ISSUE #1401: search_graph: non-ASCII queries return noise, and docstrings are never indexed
## Summary

`search_graph(query=...)` returns unranked noise for any query written in a non-Latin script, and no amount of documentation helps, because docstrings are not part of the full-text index either. On a codebase documented in Russian (or Greek, or Japanese) the BM25 mode is effectively unusable.

Two independent causes, both small:

**1. The query tokenizer drops every non-ASCII byte.** `bm25_build_match()` (`src/mcp/mcp.c`) accepts only `a-z`, `A-Z`, `0-9` and `_` as token bytes; everything else is a separator. A Cyrillic query is therefore tokenized to *nothing*, reaches FTS5 as an empty MATCH, and the caller gets whatever the fallback path returns.

**2. Docstrings are not indexed.** `nodes_fts` is declared as `fts5(name, qualified_name, label, file_path, ...)` in `src/store/store.c`. Identifiers in real projects are overwhelmingly English even when the team writes prose in another language, so the only text that *could* match a native-language query is exactly the text that never enters the index.

## Reproduction

Index any project that has a non-English docstring, e.g.

```python
def _validate_args(name: str, arguments: dict) -> str | None:
    """Провалидировать аргументы по схеме инструмента."""
```

```
codebase-memory-mcp cli search_graph --project <p> --query 'провалидировать аргументы схеме'
```

**Expected:** the documented function.
**Actual (v0.9.0):** unrelated nodes — in my case `.env.example`, a `Coordinates` class and a README section; the target function is nowhere in the results.

The English control query behaves as expected, which is what makes the failure easy to miss.

## Suggested fix

1. Treat bytes `>= 0x80` as token bytes in `bm25_build_match()`. No UTF-8 decoding is required — the point is only to keep multi-byte sequences intact and let the FTS5 `unicode61` tokenizer apply the real word-boundary rules, which it already does correctly on both sides.
2. Add a `docstring` column to `nodes_fts` and populate it from `json_extract(properties, '$.docstring')` in both backfill paths (`pipeline.c`, `pipeline_incremental.c`). Existing databases carry the 4-column table, so the schema probe has to drop and recreate it; the table is contentless, so nothing is lost — the next backfill repopulates it from `nodes`.

## Result after the fix

Same query, same project:

```
total: 1
search_mode: bm25
  ...server._validate_args  Function  src/yandex_biz_mcp/server.py  78-96  -18.89
```

The English query improves as well — `validate arguments schema` now ranks `_validate_args` first, where previously it returned noise: docstrings carry intent that identifiers alone do not.

I have a patch with a regression test ready and will open a PR referencing this issue.

## Environment

- v0.9.0 (release build, UI variant), macOS 15.5 arm64
- Reproduced on several projects; the smallest is an 8-tool Python MCP server, 137 nodes

Metatau:
Fixes #1401

BM25 search was unusable for any project documented in a non-Latin script. Two independent causes, both in this diff.

### 1. The query tokenizer dropped non-ASCII bytes

`bm25_build_match()` accepted only `[a-zA-Z0-9_]` as token bytes, so a Cyrillic query was tokenized to *nothing* and reached FTS5 as an empty MATCH — the caller got unranked noise rather than an empty result, which is why this went unnoticed.

Bytes `>= 0x80` are now token bytes. No UTF-8 decoding is introduced: the only requirement here is to keep multi-byte sequences intact, since the FTS5 `unicode61` tokenizer already applies the real word-boundary rules on both the index and the query side.

### 2. Docstrings were never indexed

`nodes_fts` covered `name, qualified_name, label, file_path` — identifiers only. In real projects identifiers are English even when the team documents in another language, so the one field that could match a native-language query was the one field missing from the index.

A `docstring` column is added and populated from `json_extract(properties, '$.docstring')` in both backfill paths (`pipeline.c`, `pipeline_incremental.c`).

**Schema upgrade.** Databases created before this carry the 4-column table, and `CREATE VIRTUAL TABLE IF NOT EXISTS` would leave them that way — every subsequent backfill would fail on the column count. The new column is probed with a cheap `SELECT docstring FROM nodes_fts LIMIT 0` and the table dropped only when it is missing. The table is contentless, so nothing is lost: the next backfill repopulates it from `nodes`.

### Result

Same project, same query, before and after:

```
# before (v0.9.0)
$ ... search_graph --query 'провалидировать аргументы схеме'
  .env.example      File
  Branch            Class
  Coordinates       Class      ← the documented function is nowhere in the results

# after
  ...server._validate_args   Function   src/…/server.py   78-96   -18.89
```

English search improves as well rather than regressing — `validate arguments schema` now ranks `_validate_args` first, because docstrings carry intent that identifiers alone do not. camelCase splitting is unaffected.

### Testing

- New regression test `tool_search_graph_query_matches_non_ascii_docstring` in `tests/test_mcp.c`: a node whose only Cyrillic text is its docstring must be found by a Cyrillic query, and an unrelated node must not. It fails on `main` and passes here.
- Full suite: **6825 passed, 4 skipped, 0 failed** (`build/c/test-runner`, macOS 15.5 arm64).
- One earlier run reported a single failure that did not reproduce on a clean re-run; I could not identify the test because that run's log was truncated on my side. The `mcp` suite specifically is stable across runs (190 passed, 2 skipped).
- `make -f Makefile.cbm lint-format` could not run locally — `clang-format` is not installed on this machine. The new code follows the surrounding style; happy to adjust if CI disagrees.

### Scope

No API surface change, no new dependency, no new pipeline pass. `json_extract` is already available in the vendored SQLite (3.51.3).

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -2769,6 +2769,18 @@ static sqlite3_destructor_type mcp_sqlite_transient(void) {
 }
 #define MCP_SQLITE_TRANSIENT (mcp_sqlite_transient())
 
+/* A byte that may sit inside a query token.  ASCII alphanumerics and '_' as
+ * before, plus every byte >= 0x80: those are UTF-8 lead/continuation bytes, so
+ * treating them as separators (the previous behaviour) silently dropped every
+ * non-Latin word from the query — a Cyrillic, Greek or CJK search reached the
+ * index as an empty match and returned unranked noise.  We do not need to
+ * decode UTF-8 here, only to keep multi-byte sequences intact: the FTS5
+ * unicode61 tokenizer applies the real word-boundary rules on both sides. */
+static int bm25_is_token_byte(unsigned char c) {
+    return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' ||
+           c >= 0x80;
+}
+
 static int bm25_build_match(const char *query, char *out, size_t out_size) {
     if (!query || !out || out_size < BM25_MIN_BUF) {
         return 0;
@@ -2777,16 +2789,14 @@ static int bm25_build_match(const char *query, char *out, size_t out_size) {
     int tokens = 0;
     const char *p = query;
     while (*p) {
-        while (*p && !((*p >= 'a' && *p <= 'z') || (*p >= 'A' && *p <= 'Z') ||
-                       (*p >= '0' && *p <= '9') || *p == '_')) {
+        while (*p && !bm25_is_token_byte((unsigned char)*p)) {
             p++;
         }
         if (!*p) {
             break;
         }
         const char *tok_start = p;
-        while (*p && ((*p >= 'a' && *p <= 'z') || (*p >= 'A' && *p <= 'Z') ||
-                      (*p >= '0' && *p <= '9') || *p == '_')) {
+        while (*p && bm25_is_token_byte((unsigned char)*p)) {
             p++;
         }
         size_t tok_len = (size_t)(p - tok_start);
--- a/src/pipeline/pipeline.c
+++ b/src/pipeline/pipeline.c
@@ -1418,16 +1418,26 @@ static int dump_and_persist_hashes(cbm_pipeline_t *p, const cbm_file_info_t *fil
          * Contentless FTS5 requires the special 'delete-all' command instead of
          * DELETE FROM to wipe prior rows (there's no underlying content table).
          * Falls back to plain names if cbm_camel_split is unavailable (which
-         * shouldn't happen because we always register it, but we stay defensive). */
+         * shouldn't happen because we always register it, but we stay defensive).
+         *
+         * `docstring` is pulled out of the properties JSON so prose in comments
+         * is searchable, not just identifiers. It matters most for non-English
+         * codebases: identifiers are almost always English, so a query written
+         * in the team's own language matched nothing at all before. The
+         * unicode61 tokenizer already handles non-Latin scripts. */
         CBM_PROF_START(t_fts);
         cbm_store_exec(hash_store, "INSERT INTO nodes_fts(nodes_fts) VALUES('delete-all');");
         if (cbm_store_exec(hash_store,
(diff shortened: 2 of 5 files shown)


Metatau:
Force-pushed a fix to a defect in my own patch — worth describing, because the first version would have been worse than the bug it fixes.

**What was wrong.** The schema upgrade dropped the 4-column `nodes_fts` and left repopulation to the pipeline's backfill. But a project that is already indexed and whose files have not changed never reaches that backfill — there is nothing to index. The FTS index therefore stayed *empty* until something happened to modify a file, and search returned nothing at all in the meantime: not just non-ASCII queries, every query. Anyone upgrading would have seen search break across all their existing projects.

I hit this on my own machine right after installing the patched binary: a previously working project stopped matching even plain `validate`.

**Fix.** The upgrade is now self-contained — immediately after recreating the table it repopulates from `nodes`, without waiting for the pipeline. `cbm_camel_split` is registered per connection later in startup, so the recovery insert uses plain names; the next full index run rewrites the table with split tokens.

**Verified on real databases**, reproducing the exact sequence that broke:

| step | English query | non-ASCII query |
|---|---|---|
| index built by released 0.9.0 (4-column table) | works | — |
| first run of the patched binary | works | **works** |
| second run, no files changed | works | works |

The third row is where the previous revision went empty.

One behavioural note I should flag: the upgrade path runs when the schema is created, not on read. Upgrading the binary alone therefore neither breaks nor fixes an existing database — it keeps working under the old schema until the project is indexed again. That seemed preferable to migrating on every open, but say the word if you would rather it happened eagerly.

**Testing after the fix:** full suite `6824 passed, 1 failed, 4 skipped`. The failure is `tool_detect_changes_contained_commands_clean_up_error_and_success` at `ASSERT(merge_base_reported)` — `git merge-base` reporting, untouched by this diff. It passes 3/3 when run in isolation, and the same build alternated between 0 and 1 failures across full runs, so I believe it is flaky rather than a regression here. Happy to be corrected if it is known-stable on your CI.

PR #1412, opened 2026-08-02
feat(lang): CFML tag-dialect .cfc routing + embedded <cfscript> defs

allanoepping:
## Summary

Two extraction fixes for legacy **tag-based CFML / ColdFusion**. The vendored
`cfmleditor/tree-sitter-cfml` grammar is Lucee-tuned and script-first, so on
older codebases (predominantly `<cfcomponent>` tag components with `<cfscript>`
bodies) most functions were never extracted. Both fixes mirror existing
conventions in the codebase.

### 1. `.cfc` dialect disambiguation — `cbm_disambiguate_cfc()`
The extension table maps `.cfc → cfscript` unconditionally, but a `.cfc` may be
script-dialect (`component { … }`) **or** tag-dialect (`<cfcomponent> … </cfcomponent>`).
New content-sniff (same shape as `cbm_disambiguate_cls/inc`):
- `<cfcomponent>` or a bare top-level `<cffunction>` → `CBM_LANG_CFML` (tag grammar)
- a leading `<cfscript>`-wrapped `component {}`, or a plain script component → `CBM_LANG_CFSCRIPT`
- leading `<!--- --->` comments are skipped before deciding

### 2. Embedded `<cfscript>` definitions — `cbm_extract_embedded_defs()`
The HTML-derived `cfml` grammar keeps `<cfscript>` bodies as an opaque
`cf_script_content` token, so script functions inside tag components produced no
`function_declaration` nodes. This re-parses each `cf_script_tag → cf_script_content`
slice with the `cfscript` grammar (reusing the `embedded_imports` machinery) and
walks it for definitions via a new `cbm_extract_definitions_body()` that skips the
Module node. Block-relative line numbers are shifted by the block's start row —
`CBMDefinition` carries only line positions, so a constant offset is a complete
remap. **Definitions only** (calls/usages would need byte remapping), so existing
edges cannot be corrupted. Bonus: functions are recovered even when a `<cfsetting>`
void-tag ERROR cascade damages the surrounding tree.

Adds an `embedded_defs` field to `CBMLangSpec`, wired only to `CBM_LANG_CFML`.

## Tests
- `extract_cfml_embedded_cfscript_defs` — script funcs inside `<cfscript>` in a tag
  component extract with correct (remapped) line numbers, alongside a `<cffunction>`.
- 8 × `lang_cfc_*` — tag / script / `<cfscript>`-wrapped / bare `<cffunction>` /
  leading-comment / read-fail routing.

`scripts/test.sh --suites "language extraction grammar_regression"` → 497 passed, 0 failed.

## Real-world impact
Indexing a ~700-file legacy CFML app (156/176 `.cfc` are tag-based):
- parse timeouts **4 → 0**
- `api.cfc` **0 → 62** functions, `api-a.cfc` **0 → 48**
- **+1,532** function nodes overall, with exact line numbers

## Notes
Not addressed here: `<cfsetting>` (and other void CF tags) still emit tree-sitter
ERROR nodes in the `cfml` grammar — cosmetic now that functions are recovered, but
a genuine grammar issue that belongs in `cfmleditor/tree-sitter-cfml` (regenerating
`parser.c` from `grammar.js`).

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -1249,6 +1249,7 @@ CBMFileResult *cbm_extract_file_ex(const char *source, int source_len, CBMLangua
     // Run extractors: defs + imports use separate walks (unique recursion patterns),
     // then a single unified cursor walk handles the remaining 7 extractors.
     cbm_extract_definitions(&ctx);
+    cbm_extract_embedded_defs(&ctx); // defs inside <cfscript> blocks of CFML tag files
     cbm_extract_imports(&ctx);
     cbm_extract_unified(&ctx);
 
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -1292,6 +1292,7 @@ CBMFileResult *cbm_extract_file_ex(const char *source, int source_len, CBMLangua
     // Run extractors: defs + imports use separate walks (unique recursion patterns),
     // then a single unified cursor walk handles the remaining 7 extractors.
     cbm_extract_definitions(&ctx);
+    cbm_extract_embedded_defs(&ctx); // defs inside <cfscript> blocks of CFML tag files
     cbm_extract_imports(&ctx);
     cbm_extract_unified(&ctx);
 
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -726,6 +726,8 @@ void cbm_channels_push(CBMChannelArray *arr, CBMArena *a, CBMChannel ch);
 // --- Sub-extractor entry points ---
 
 void cbm_extract_definitions(CBMExtractCtx *ctx);
+void cbm_extract_definitions_body(CBMExtractCtx *ctx); // no Module node (for sub-trees)
+void cbm_extract_embedded_defs(CBMExtractCtx *ctx);    // re-parse embedded script blocks for defs
 void cbm_extract_imports(CBMExtractCtx *ctx);
 void cbm_extract_usages(CBMExtractCtx *ctx);
 void cbm_extract_semantic(CBMExtractCtx *ctx);
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -730,6 +730,8 @@ void cbm_extract_definitions(CBMExtractCtx *ctx);
 // per ref()/source() call. No-op unless the file parses as SQL and actually
 // contains a dbt builtin call. Defined in extract_dbt.c.
 void cbm_extract_dbt(CBMExtractCtx *ctx);
+void cbm_extract_definitions_body(CBMExtractCtx *ctx); // no Module node (for sub-trees)
+void cbm_extract_embedded_defs(CBMExtractCtx *ctx);    // re-parse embedded script blocks for defs
 void cbm_extract_imports(CBMExtractCtx *ctx);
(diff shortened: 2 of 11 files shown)


allanoepping:
The Windows red check (`test-windows … 2/2`, plus the downstream `ci-ok` / `shard-completeness` aggregates) is unrelated to this change — it fails in the **Step-0 static *infrastructure* contracts**, not in any test suite or in code this PR touches (CFML `.cfc` disambiguation + embedded `<cfscript>` definition extraction, all in the C extraction path).

Two Windows-runner environmental failures surface there:

- **Step 0a — `path-safety.sh`**: `ERROR: repository build/ ancestor must be absent or a real directory`. `build/` is gitignored, so a clean checkout has it absent; this fires only when the runner leaves `build/` as a symlink/non-dir (a cache/filesystem artifact on the Windows VM).
- **Step 0i — parallel suite scheduler contract**: a file-lock race during cleanup —
  ```
  rm: cannot remove '…/cbm-parallel-harness.…/logs/timeout_exit_race.log': Device or resource busy
  ValueError: invalid literal for int() with base 10: ''
  ```
  i.e. the scheduler's own timeout/hang fixtures (`hang_after_summary`, `stubborn_tree`, `timeout_exit_race`) left a handle open so the `rm` failed and empty output tripped an `int('')` parse.

Everything relevant is green, including the full suite under clang-TSAN on Linux + macOS (which exercises the new `lang_cfc_*` and `extract_cfml_embedded_cfscript_defs` tests): DCO, lint, all `test-unix` gcc shards, all `test-tsan` (ubuntu/arm/macOS clang), `test-unix (macOS)`, and `test-windows-guards`. A re-run of the Windows shard should clear it.

DeusData:
Thank you for the CFML contribution and for covering both tag-dialect routing and embedded `<cfscript>` definitions. This is now triaged as a high-priority parsing fix in `0.9.1-rc` because the existing route can silently produce an incomplete graph. This is an intake acknowledgement rather than a code review. Our community review queue is currently full, so it may take us a little time to validate the grammar surface, fixtures, and cross-language impact, but the PR is in the correct queue.

DeusData:
@allanoepping — your CFML work is on `main` in 31b611a3 (via #1872), with `Co-authored-by:` credit in the commit.

What shipped is your contribution with one architectural substitution, made honestly on our side: your `.cfc` dialect sniffer and all nine tests landed nearly verbatim, but your embedded-defs mechanism was replaced by ~25 lines wiring the shared included-ranges machinery that #1852 merged *after* your last rebase — your ~90-line implementation faithfully mirrored the in-tree shape of its day, and the day moved under you. The supersession was our churn, not your defect, which is why the port was done maintainer-side rather than asked of you.

Two things your PR forced us to get right in the port: the Vue attribute resolver had to be gated to Vue (it would have silently rewritten CFML's embedded language to JavaScript), and CFML had to join the embedded-imports dispatch switch — your extraction test went red until both were correct, which is exactly what a reproduce-first test is for. Your line-number assertions survive unchanged because included ranges give absolute coordinates for free, and calls inside `<cfscript>` blocks are now extractable as a bonus your design documented as out of reach under slicing.

Legacy tag-dialect components extracting zero functions was a real hole; thank you for finding it, proving it, and building the fix twice over.

PR #1423, opened 2026-08-03
fix(cli): add native FreeBSD self-executable path resolution

PR9000:
## What does this PR do?

`cbm_detect_self_path` had no branch for FreeBSD, falling through to the generic `#else` that reads `/proc/self/exe`. In this environment `/proc` is a native FreeBSD procfs mount left empty (only `/compat/linux/proc` is populated, for Linux-ABI processes under the compat layer), so the `readlink` always failed silently and the function fell back to the expected install path instead of the actual running binary's path.

This broke `install`: with no valid self path, the activation transaction could never confirm what to publish and aborted early before attempting any `linkat`/`rename`.

Add `sysctl(CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1)` for FreeBSD in `src/cli/cli.c`, matching the pattern already used in the test suites, and include `<sys/sysctl.h>` guarded by `__FreeBSD__`.

## Checklist

### DIFF
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -107,6 +107,10 @@ static int cbm_powershell_quote_word(const char *value, char *out, size_t out_si
 #include <stdlib.h>
 #include <string.h>   // strtok_r
 #include <sys/stat.h> // mode_t, S_IXUSR
+#ifdef __FreeBSD__
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#endif
 #include <time.h>
 #include <wchar.h>
 #include <zlib.h> // MAX_WBITS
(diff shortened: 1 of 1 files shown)


DeusData:
Merged — thank you @PR9000. `sysctl(KERN_PROC_PATHNAME)` is the right idiom, the `__FreeBSD__` guard keeps it out of every other build, and falling through to the existing `$HOME/.local/bin` path when the sysctl fails preserves the documented contract that an inexact result is only ever used as a copy source, never as ownership evidence.

Worth noting why this one matters more than its size suggests: `cbm_detect_self_path` feeds the install/activation ownership checks, so on FreeBSD every one of those was reasoning about a guessed path. Together with your #1421 guard, FreeBSD goes from not-building to actually-correct.

Shipping in the next release.

PR #1426, opened 2026-08-03
feat(pkl): extract calls, branches, throws and decorators

tthayer:
## Problem

Pkl was already wired end-to-end — grammar vendored, `CBM_LANG_PKL`, extension mapping, import edges — but the spec was shallow. `pkl_call_types` was `empty_types`, so on a real-world 172-file Pkl corpus the graph had **0 CALLS edges** while imports worked fine.

`branching_node_types`, `throw_node_types` and `decorator_node_types` were empty too, so Pkl contributed nothing to complexity metrics.

## Why this needs a custom extractor

Pkl has no dedicated call node. `unqualifiedAccessExpr` and `qualifiedAccessExpr` are the **same node** whether they are a call or a bare property read — the only discriminator is an `argumentList` child:

```
unqualifiedAccessExpr 'helper(a)'     unqualifiedAccessExpr 'host'
  identifier 'helper'                   identifier 'host'
  argumentList '(a)'                  <-- no argumentList
```

`extract_pkl_callee` gates on that child. It is dispatched from `extract_callee_name` with an **unconditional return**, not via `extract_callee_lang_specific`: returning NULL from the latter falls through to the generic "first identifier child" fallback, which would mint a CALLS edge for every property read in every Pkl file.

Callee resolution:

| source | callee |
|---|---|
| `helper(a)` | `helper` |
| `utils.fallback(a)` | `utils.fallback` (module-qualified; `cbm.c` shortens to the last dotted segment) |
| `s.trim().toLowerCase()` | `toLowerCase` (a receiver that is itself a call has parens in its text, so it is not prefixed) |
| `new Server { ... }` | `Server` (links to the class def) |

## Other spec slots filled

All node names confirmed against real parse trees from a probe linking only the Pkl grammar, not guessed:

- **class**: `+ typeAlias`
- **import**: `+ importGlobClause`, `importExpr`
- **branch**: `ifExpr`, `whenGenerator`, `forGenerator`
- **throw**: `throwExpr`
- **decorator**: `annotation`

`forGenerator` is also registered in `cbm_is_loop_node_type` so `for (x in xs)` counts toward loop-nesting depth. The name is Pkl-unique, so no other grammar collides.

## Validation

**Repro battery** — the Pkl test moves from the structural battery (dims 1-5) to the full callable battery (dims 1-8), plus an inline negative assertion that bare property reads do *not* become CALLS edges. `repro_grammar_config_pkl` PASS.

Full suite: **318 passed, 6 failed, 2 skipped** — byte-identical to the pre-change baseline (verified by stashing and re-running). The 6 failures are pre-existing and unrelated (`repro_issue480/514/546/581`, `repro_issue409`, `repro_ts_inherited_method`).

**Real corpus** — 172 Pkl files of production infrastructure config, indexed clean (0 skipped, 0 parse failures):

| | before | after |
|---|---|---|
| CALLS edges | 0 | **367** |

Split: 270 to `Function`, 97 to `Class` (constructors). ~2.1 calls/file — proportionate; a broken gate would have produced tens of thousands.

Quality was spot-checked rather than inferred from the count: sampled `Function` callees are genuine `function` definitions invoked with arguments, and a bare property read that shares its name with one of those functions correctly produced no edge — the exact case the `argumentList` gate exists to catch.

## Note for reviewers

The repro fixture deliberately keeps every call site inside a method body, because dim 7 treats Module-sourced in-body calls as the enclosing-func gap. Real-world Pkl *does* call at module level (many of the 97 constructor edges are such sites), so this is a fixture constraint, not a claim about the language.


DeusData:
Thank you for expanding the Pkl extraction coverage across calls, branches, throws, and decorators. I have routed this language-support contribution to `0.11.0-rc` with normal priority. The community review queue is currently full, so a complete parser review may take a little time, but the contribution is queued.

DeusData:
Thank you for the Pkl language work and the coverage across calls, branches, throws, and decorators. The Pkl changes form one coherent language contribution and do not need to be split further.

One commit is outside that scope: `a700bff1` changes Windows helper-spawn behavior in `scripts/run-test-wave.py` and its harness contract. Please remove that commit from this PR, or restack the Pkl work onto a clean branch without it. CI process-control changes need their own focused review because they affect every language and platform. Once that unrelated commit is separated, we can review this as the Pkl contribution it is. Thank you for helping keep that boundary clear while the maintainer queue is full.

tthayer:
`a700bff1` removed. Pkl work restacked on `main` (`010569fa`) as `9c25f163` + `597985d2`; diff is 8 Pkl files, `scripts/run-test-wave.py` gone.

`tests/test_parallel_harness_contract.sh` left too: that hunk reads `WINDOWS_HELPER_TIMEOUT_SECONDS` from the scheduler, so it was the contract half of `a700bff1` (misfiled into the Pkl commit) and cannot ship without it. Both halves will come back as their own PR.

`scripts/repro.sh`, darwin arm64:

| | clean `main` | branch |
|---|---|---|
| passed | 689 | 690 |
| failed | 14 | 14 |
| skipped | 2 | 2 |

Same 14 failures either way (`repro_issue56/409/480/514/523/546/581/627`, `repro_lsp_go_py`, `repro_ts_inherited_method`); the added pass is `repro_grammar_config_pkl`. Touched suites green: `repro_grammar_config` 16/16, `repro_call_argument_matrix_b` 47/47, `repro_language_registry` 5/5, `repro_call_node_manifest` 3/3.

tthayer:
Split-out CI change is now #1823.

DeusData:
@tthayer — your 24 August restack did everything the 18 August ruling asked, and cleanly; the split-out #1823 is exactly right. One more rebase is needed through no fault of yours: ArkTS and PL/SQL landed on main in the four days after your restack, and both moved the same ledger constants your PR moves. So you don't have to re-derive anything, here is the complete map — including the part that will bite silently.

**The trap first: `tests/repro/repro_call_node_manifest.c` will NOT show a merge conflict, and will still be wrong.** Both main and your branch moved its constants by identical deltas from the old base, so git auto-resolves them "clean" — short by exactly Pkl's contribution. We have been bitten by this exact shape twice this month. Set these by hand:

| constant | your branch | correct after rebase |
|---|---|---|
| `EXPECTED_HISTORICAL_CALL_NODE_TOTAL` | 222 | **225** |
| `EXPECTED_ACTIVE_PRIMARY_TOTAL` / `EXPECTED_PRIMARY_OWNERS` | 192 | **195** |
| `EXPECTED_DIRECT_CALLS` | 157 | **159** |
| `EXPECTED_CONSTRUCTOR_CALLS` | 21 | **22** |

The other two ledger files conflict visibly, so git will make you look; correct values there: `repro_language_registry.c` → REF_VOCAB **88**, WITHOUT_REF_VOCAB **26**, NO_CALL **49**, MATRIX_A **68**, MATRIX_B **48**, CALL_CAPABLE **114**, NON_CALL **51**; `repro_call_argument_matrix_b.c` → ROUTINE_ARGUMENT_LANGUAGE_COUNT **38**, matrix assert **48**. Or simply run the three repro suites after resolving and take the recounted values — but the manifest file needs the hand edit either way, because nothing will flag it.

Two smaller items for the same pass:

1. **Reword `597985d2`'s commit message** — it still describes the removed `WINDOWS_HELPER_TIMEOUT_SECONDS` harness change. The content is verifiably gone from the diff; the message shouldn't describe a change the commit no longer contains.
2. For transparency: your added header prose in the manifest ("the historical-snapshot column is the audited ledger, not an archaeology record") was checked against how ArkTS and PL/SQL actually moved those rows, found to codify existing practice accurately, and is accepted as-is.

The review verified the parts that matter: every extraction edit is behind a `CBM_LANG_PKL` gate with an unconditional return (so the generic fallback can't mint a CALLS edge per property read — the `argumentList`-gated callee design is exactly right for a language where access-exprs double as property reads); `forGenerator` exists only in the pkl grammar, so the shared loop-list addition is collision-free; and all 18 spec-row slots land on the correct struct fields. Rebase with the table above and this is ready for the final pass.

tthayer:
Rebased onto main (`31b611a3`), pushed as `1d6a140f` + `7b72652a`.

- Visible conflicts resolved with your table; silent auto-merges hand-fixed: manifest 225/195/159/22, registry ref-vocab 87→88, matrix B routine count 37→38 (+ static assert 48).
- Tip commit reworded — harness paragraph removed.

`scripts/repro.sh` darwin arm64: 693 passed, 14 failed, 2 skipped — same pre-existing 14. Touched suites green: registry 5/5, matrix B 48/48, manifest 3/3, grammar_config 16/16.

DeusData:
Your 28 August rebase did everything the ruling asked, and the hand-fixed silent auto-merges were exactly right. **Chialisp (`706bb2ce`) has since landed**, so the ledger constants moved once more — through no fault of yours, and this is the third time our churn has cost you a rebase. Sorry about that.

So you do not have to re-derive anything, here is the complete map, computed against current `main` by taking your branch's delta from base `31b611a3` and re-applying it to main's value.

**Silent — both sides independently reached the same number, so git auto-merges clean and wrong.** These are the dangerous ones; set them by hand:

| constant | file | your branch | main | **correct** |
|---|---|---|---|---|
| `EXPECTED_MATRIX_B_ROWS` | `repro_language_registry.c` | 48 | 48 | **49** |
| `EXPECTED_CALL_CAPABLE_LANGUAGES` | `repro_language_registry.c` | 114 | 114 | **115** |
| `ROUTINE_ARGUMENT_LANGUAGE_COUNT` | `repro_call_argument_matrix_b.c` | 38 | 38 | **39** |
| `MATRIX_LANGUAGE_COUNT` static assert | `repro_call_argument_matrix_b.c` | 48 | 48 | **49** |

**Visible — these will conflict loudly; resolve to:**

| constant | file | your branch | main | **correct** |
|---|---|---|---|---|
| `EXPECTED_HISTORICAL_CALL_NODE_TOTAL` | `repro_call_node_manifest.c` | 225 | 223 | **226** |
| `EXPECTED_ACTIVE_PRIMARY_TOTAL` | `repro_call_node_manifest.c` | 195 | 193 | **196** |
| `EXPECTED_PRIMARY_OWNERS` | `repro_call_node_manifest.c` | 195 | 193 | **196** |
| `EXPECTED_DIRECT_CALLS` | `repro_call_node_manifest.c` | 159 | 158 | **160** |

**Already correct, no action** — `EXPECTED_CONSTRUCTOR_CALLS` (22; main did not move it), `counts[CAP_CALL_WITH_REFERENCE_VOCAB]` (88; yours is the only change), and `counts[CAP_CALL_WITHOUT_REFERENCE_VOCAB]` (27; main's change, your delta is zero).

**Everything else auto-merges cleanly.** I checked with `git merge-tree` against current `main`: `extract_calls.c`, `extract_usages.c`, `helpers.c` and `lang_specs.c` all merge without conflict. The only three files that conflict are the three ledger files above.

One other thing worth knowing, since it cost you time: your `security / codeql-gate` red was never your code. The gate waited 45 minutes for a CodeQL run that takes ~2 hours here, so it timed out on a scan that later **passed**. I re-ran the failed jobs and this PR is green on that check now, and #1960 raises the budget so it stops happening.

Splitting `a700bff1` out into #1823 was exactly the right call, and the before/after `repro.sh` table with the same 14 pre-existing failures either way is the kind of evidence that makes a language contribution reviewable. Rebase with the table above and ping me — the re-review will be prompt.

tthayer:
Resolved the merge conflicts with main (c9d68de6). The conflicts were all in the repro ledgers, where this branch and the Chialisp addition (#1920) each added one call-capable language, so the shared totals needed both deltas applied:

- `repro_language_registry.c`: capability partition 88 with-vocab / 27 without-vocab / 49 no-call; `EXPECTED_CALL_CAPABLE_LANGUAGES` 115; `EXPECTED_NON_CALL_LANGUAGES` 51; `EXPECTED_MATRIX_B_ROWS` 49
- `repro_call_node_manifest.c`: historical total 226, active primary 196, direct calls 160, constructor calls 22, primary owners 196
- `repro_call_argument_matrix_b.c`: `ROUTINE_ARGUMENT_LANGUAGE_COUNT` 39, matrix static assert 49 (`RACKET..CHIALISP`)

Note: two constants (`EXPECTED_CALL_CAPABLE_LANGUAGES` and `EXPECTED_MATRIX_B_ROWS`, plus `ROUTINE_ARGUMENT_LANGUAGE_COUNT`) were bumped 113→114 / 47→48 / 37→38 identically on both sides, so git auto-merged them without a conflict marker — those needed a second manual bump to 115 / 49 / 39.

No changes beyond the conflict resolution. Full C test suite passes locally after the merge: 7772 passed, 7 skipped, 0 failed.

DeusData:
Merged as `f0cf84d2`. Pkl is in.

You applied the constants table exactly — I verified all eleven values against `main` rather than trusting the rebase, including all four of the ones that merge *clean and wrong*:

| constant | main | yours |
|---|---|---|
| `EXPECTED_MATRIX_B_ROWS` | 48 | **49** |
| `EXPECTED_CALL_CAPABLE_LANGUAGES` | 114 | **115** |
| `ROUTINE_ARGUMENT_LANGUAGE_COUNT` | 38 | **39** |
| `MATRIX_LANGUAGE_COUNT` | 48 | **49** |

Those are the four git would have merged silently at the wrong value. Getting them right by hand, twice, across two separate rebases neither of which was your fault, is the part of this that took real care.

Three things worth recording:

**Splitting `a700bff1` out into #1823 was exactly right**, and you did it without argument when it would have been easier to argue. CI process-control changes affect every language and platform; keeping them out of a language contribution is what made this one reviewable at all.

**Your before/after `repro.sh` table did the work a reviewer would otherwise have to do** — same 14 pre-existing failures on clean `main` and on the branch, with the one added pass named. That is how you make a language addition checkable instead of trusting it.

**And you were held up twice by our churn, not yours** — ArkTS and PL/SQL, then Chialisp. Sorry for that.

One thing that was never yours either: this PR spent time parked on our diff scanner, which read the pre-existing `uint32_t nc = ts_node_named_child_count(node)` in `extract_calls.c` as an invocation of netcat. `nc` turns out to be a 287-site house idiom in our own C. The rule has since been scoped to script files, so that will not happen to the next language contribution.

Issue #573, opened 2026-06-23
Feature request: worktree overlay indexing for branch-aware agent workflows


pcristin:
## What problem does this solve?

Agent-driven development workflows often use many short-lived Git worktrees for feature branches, PR review, verification, and parallel tasks. The current worktree/branch-root support from #401 makes the graph shape more explicit, but it still leaves a larger design question open: how should Codebase Memory MCP avoid treating mostly identical worktrees as fully independent indexes?

The remaining pain points are:

- duplicated indexing and storage for worktrees that mostly match the base branch
- project-list clutter from short-lived worktree paths
- query ambiguity when an MCP client is running inside a worktree but the canonical checkout is also indexed
- stale project entries after temporary worktrees are removed, related to #286
- missing or stale branch-specific context if only the canonical checkout is considered
- unnecessary churn if every transient agent edit triggers a full or near-full reindex

Related prior discussion and implementation:

- #48 discussed broad Git worktree support and arbitrary worktree locations.
- #37 skipped `.worktrees/` to avoid nested duplicate indexing.
- #351 proposed efficient indexing for worktree-heavy agent workflows.
- #401 added first-class branch/worktree graph roots and metadata.
- #489 / #493 fixed one concrete nested-worktree OOM class via `.git/info/exclude` handling.
- #499 / #542 cover the remaining global Git exclude layer.

## Proposed solution

I would like to discuss a committed-source, file-granularity overlay model on top of the existing branch/worktree awareness:

1. Keep one canonical/base index for the repository's base branch.
2. For each branch/worktree context, compute the merge base against the configured base branch.
3. Index only files changed between `merge-base..HEAD` into an overlay context.
4. At query time, resolve the active Git context and merge results as:
   - overlay nodes/edges first
   - overlay tombstones for deleted files or symbols
   - base index for unchanged files
5. If a file exists in the overlay, hide that file's base graph for that context.
6. Treat worktree path as an attachment to a branch/HEAD context, not necessarily as the graph identity.

In other words, a repository would look conceptually like:

```text
Project: example-repo
  Base context: main @ abc123
    full graph for committed main

  Overlay context: feature/foo @ def456, base abc123
    graph for changed committed files only
    deleted-file tombstones
    metadata: branch, worktree path, base SHA, head SHA
```

For a first implementation, I suggest indexing committed changes only:

```text
overlay = git diff --name-status $(git merge-base HEAD <base-branch>)..HEAD
```

[...]

PR #1430, opened 2026-08-04
feat(config): add ignore_worktrees to skip linked git worktrees

vitalNohj:
## Problem

Agent workflows increasingly use `git worktree add` to run several branches of the same repo side by side. Every worktree is discovered as its own project, so the cache accumulates a near-duplicate graph per worktree and project pickers get noisy. Today there is no way to say "index the main checkout, leave the worktrees alone".

## Change

Adds an **opt-in** `ignore_worktrees` config key. When enabled, a linked worktree is left alone across every indexing entry point:

- **auto-index on connect** skips it (logged as `autoindex.skip` with `reason=linked_worktree`)
- **the daemon's background auto-index** skips it
- **explicit `index_repository`** refuses with an actionable message naming both ways forward
- **hook-augment guidance** stops telling the agent to run `index_repository` in a directory where it would only be refused

The explicit call keeps an escape hatch — `index_worktree=true` indexes a worktree regardless of the setting, so the key never becomes a dead end:

```
repo_path is a linked git worktree and ignore_worktrees is enabled. Pass
index_worktree=true to index it anyway, or run:
codebase-memory-mcp config set ignore_worktrees false
```

**Default is `false`**, so nothing changes for existing users unless they opt in.

## Detection

`cbm_git_is_linked_worktree()` is git plumbing only, **no subprocess** — it runs on every session start, so shelling out the way `cbm_git_context_resolve()` does was not an option.

`<path>/.git` must be a regular file holding a `gitdir:` pointer **and** that gitdir must contain a `commondir` file.

That second condition is the important one: a **submodule**'s `.git` is *also* a gitlink file, so a naive "is `.git` a regular file?" check misclassifies every submodule as a worktree. A submodule's gitdir (`<super>/.git/modules/<name>`) has no `commondir` entry, which separates the two cleanly. The main checkout of a repo that *has* worktrees is never affected.

## Tests

`tests/test_git_context.c` — detection truth table: linked worktree true; the main checkout of that same repo false; a submodule false; a plain non-git dir false.

`tests/test_mcp.c` — the config gate on explicit `index_repository`: worktree refused, main checkout **not** refused, and `index_worktree=true` escaping the gate.

Both follow the existing fixture patterns in those files (`th_mktempdir`/`th_rmtree`, `SKIP_PLATFORM` on Windows, graceful skip when git is unavailable).

## Verification

- `scripts/build.sh` — clean
- `scripts/test.sh` — **6661 passed, 0 failed** (116 suites)
- `make -f Makefile.cbm security` — all 8 layers pass
- clang-format (repo-wide `lint-format`) and cppcheck — clean; clang-tidy reports **0 findings on the added lines**
- manually verified end-to-end against the built binary: refusal on a worktree, normal indexing on the main checkout, `index_worktree=true` override, and unchanged default-off behaviour

## Docs

`docs/CONFIGURATION.md` gets a table row plus a section covering motivation, the enable command, all three behaviours, and the submodule/main-checkout caveat. `README.md` gets a short paragraph next to `auto_watch`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ When enabled, new projects are indexed automatically on first connection. Previo
 
 Watcher registration is controlled separately by `auto_watch` (default `true`). Set `config set auto_watch false` to keep a session from registering its project with the background watcher — useful when working across many projects and you want each session contained to explicit indexing.
 
+Linked git worktrees are indexed as their own projects by default. Set `config set ignore_worktrees true` to skip them on the automatic paths — useful when short-lived `git worktree add` checkouts would otherwise each leave behind a permanent index of the same repository. An explicit `index_repository` call on a worktree is then refused unless you pass `index_worktree=true`. See [docs/CONFIGURATION.md](docs/CONFIGURATION.md#ignore_worktrees).
+
 ### Keeping Up to Date
 
 **Updates run from the install script on every platform, not from inside the running binary.** `codebase-memory-mcp update` validates your flags and then prints the exact command to run:
@@ -663,6 +665,7 @@ codebase-memory-mcp config list                          # show all settings
 codebase-memory-mcp config set auto_index true           # auto-index on session start
 codebase-memory-mcp config set auto_index_limit 50000    # max files for auto-index
 codebase-memory-mcp config set auto_watch false          # don't register background git watcher (default: true)
+codebase-memory-mcp config set ignore_worktrees true     # skip linked git worktrees when auto-indexing (default: false)
 codebase-memory-mcp config reset auto_index              # reset to default
 ```
 
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -86,6 +86,36 @@ Current keys:
 |---|---|---|
 | `auto_index` | `false` | Automatically index new projects when an MCP session starts. |
 | `auto_index_limit` | `50000` | Maximum file count allowed for automatic indexing of a new project. |
+| `ignore_worktrees` | `false` | Skip linked git worktrees (`git worktree add`) when indexing automatically. |
+
+### `ignore_worktrees`
+
+Every indexed project is registered under its own absolute root path, so each
+linked worktree becomes a separate permanent index. On machines that create many
+short-lived worktrees, the automatic paths (`auto_index`, and the session hook's
+"index this project first" guidance) turn every throwaway checkout into another
+stored index of what is largely the same repository.
+
+Enable the key to keep those checkouts out of the index:
+
+```bash
+codebase-memory-mcp config set ignore_worktrees true
+```
+
+With it enabled:
+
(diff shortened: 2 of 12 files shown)


DeusData:
Thank you for the focused `ignore_worktrees` configuration contribution. This adds public configuration and changes repository-discovery behavior, so I have routed it to `0.11.0-rc` with normal priority for a full behavior review. The community queue is currently full, and detailed feedback may take a little time.

DeusData:
@vitalNohj — the behavior review is done, and the answer is **accepted: both surfaces ship** — the `ignore_worktrees` config key and the `index_worktree` per-call override — once the rework below lands. Your detection design carried the decision: the `commondir` marker is the right discriminator (submodule-safe, since submodule gitdirs never carry it), the fail-open posture on any probe failure is exactly correct, and default-off is verifiably bit-for-bit today's behaviour.

Context you deserve on why this needed a maintainer decision at all: there is a parked direction (see #573) where worktrees eventually get a cheap *view* keyed by the git common dir rather than nothing. Your key is opt-in-off, so it forecloses none of that — and your commondir plumbing is genuinely reusable for that family-identity work later. But "skip worktrees on purpose" and "worktrees get an overlay" are opposite product answers to the same pain, so shipping yours ahead of that decision was a call, not a formality. It is now made, in your favour.

**The rework map — one item is invisible and load-bearing:**

1. **`CONFIG_KEYS` row (the invisible one).** Since #1522/#1558, that table in `src/cli/cli.c` (:6837) is the `config set` allowlist — `config_key_lookup` rejects unknown keys (:6997). Your PR predates this, patches the old printf blocks, and adds no row — so on current main **every enable command your own docs and refusal message suggest would error `unknown config key`**. Add `{CBM_CONFIG_IGNORE_WORKTREES, "false", "<help text>"}` with the default mirroring your runtime reader, per the table's keep-in-sync comment. Nothing will fail a test if you miss this — the #1558 pin covers only the UI keys — which is precisely why it leads this list.
2. **Rebase**: conflicts in README, CONFIGURATION.md and the rewritten config region of cli.c; the other nine files auto-merge.
3. **The default-off pin**: a test asserting that with the key *unset*, a linked worktree still indexes exactly as today — the regression class we guard hardest. Your explicit-gate probe with its positive control and override-escape is good; this is the missing direction.
4. If you can: cover the auto-index path (your `_for_testing` wrapper hardcodes `false`, so the new branch is currently unreachable from the test hook), swap the hardcoded `/tmp` for `th_mktempdir` as your body describes, and let fixture failures SKIP rather than PASS.

The one red CI leg on your final SHA (`test_daemon_ipc.c:2248`, Windows) is a timeout-decided test on a loaded runner, mechanism-unrelated to your diff — it is now on our flake ledger, and a rebase re-runs everything anyway.

Take your time; the acceptance does not expire. This is a first contribution built the way we wish more were.

PR #1438, opened 2026-08-04
fix: quarantine files on EXIT_NONZERO instead of aborting chunk

shashwatMishra96:
When a worker process exits with a nonzero code (e.g. internal parse-limit abort on a pathological file), treat it the same as CRASH/HANG: quarantine the offending file via the existing two-consecutive-strikes mechanism and continue indexing the rest of the chunk.

Previously, EXIT_NONZERO caused the supervisor to abort the entire chunk, losing all indexed data for that directory. This is impactful for large-scale repos (Android AOSP: 253GB, 60M+ files) where machine-generated files routinely trigger internal limits.

The two-consecutive-strikes intersection mechanism already guards against false quarantines — a systemic nonzero exit (e.g. bad CLI arg) produces no recurring suspect, so the intersection is empty and the supervisor gives up (same behavior as before).

Tested on Android QSSI17 (253GB, 24 chunks) + QCM6490 vendor (244 chunks). Result: 11.35M nodes indexed, 0 data loss, 2 genuinely pathological files quarantined (checkpatch.pl, a 140K-line generated .ts file).

## Checklist


DeusData:
The design earns its merge: EXIT_NONZERO joining CRASH/HANG in the quarantine path is right, the two-consecutive-strikes intersection already guards innocents for systemic failures, and the AOSP-scale field validation (253 GB, 11.35M nodes, exactly 2 genuine quarantines) is the strongest evidence a PR in this area has brought. One requirement stands between it and merging: this repo's reproduce-first rule needs a regression test — a deterministic nonzero-exit worker on one file proving the chunk continues and the file lands quarantined with phase "error", plus the systemic-nonzero give_up case. One design thought to address in a comment (not necessarily code): a deterministic first-file nonzero exit progressively quarantining wrong files is a property the crash path already accepts — say so explicitly in the WHY comment so the next reader knows it's a considered trade.

shashwatMishra96:
Thanks for the review and advice.

I’ve revised the PR and rebased it onto the latest version of `main`. The following updates have been made:

1. **Regression Tests** ([`[tests/test_mcp.c](https://github.com/DeusData/codebase-memory-mcp/blob/fix/quarantine-exit-nonzero/tests/test_mcp.c)`](https://github.com/DeusData/codebase-memory-mcp/blob/fix/quarantine-exit-nonzero/tests/test_mcp.c))

   * `TEST(index_recovery_quarantines_exit_nonzero)`: Injects a single file with `CBM_TEST_EXIT_ON` to verify that the chunk continues processing, the overall status is `-indexed`, and the failing file is quarantined with phase `-error`.
   * `TEST(index_recovery_systemic_exit_nonzero_gives_up)`: Simulates a systemic worker failure across all files and verifies that the operation cleanly transitions to `give_up` without creating false quarantines.

2. **Test Seam**

   * Added `CBM_TEST_EXIT_ON` to `internal/cbm/cbm.c`, guarded by `#ifdef CBM_ENABLE_TEST_SEAMS`, to support deterministic testing of non-zero worker exits.

3. **WHY Comment**

   * Updated [`[src/mcp/mcp.c](https://github.com/DeusData/codebase-memory-mcp/blob/fix/quarantine-exit-nonzero/src/mcp/mcp.c)`](https://github.com/DeusData/codebase-memory-mcp/blob/fix/quarantine-exit-nonzero/src/mcp/mcp.c) to document the tradeoff behind progressive, first-file quarantine behavior.

All **191 tests in `suite_mcp` are passing**.

DeusData:
You did everything I asked, and the main test is exactly the shape I wanted. Two fixes and this is ready.

**`index_recovery_quarantines_exit_nonzero` is genuinely red without the fix**, and for the right reasons: pre-fix, EXIT_NONZERO aborts the chunk so `status: "indexed"` never appears, and even if it did, `"phase":"error"` is a string only your `mcp.c` change produces. What I want to credit specifically is that you did not stop at "the offender was quarantined" — you query `search_graph` for `idxpar_good_fn` and fail with `IDXPAR_GOOD_MISSING` if it is absent. That is the assertion that proves the chunk *continued* rather than merely that it did not crash, and it is the difference between a reproduction and a test that would pass if quarantine silently ate everything. It also fails loudly rather than vacuously if the test seam is compiled out.

**Two fixes, both in the systemic test.**

1. **A NULL response is scored as a pass.**

   ```c
   if (!resp) {
       return IDXPAR_OK;
   }
   ```

   If the supervised run produces no response at all — the most likely symptom of a real regression in this path — the test reports success. That should be a failure with its own code, the same way you handled the other error paths in the sibling test.

2. **The accept condition is three-way**, taking `"status":"gave_up"` or `"status":"error"` or `"action":"give_up"`. Combined with your own (correct) claim that systemic behaviour is unchanged, this test very likely passes both before and after the fix. That makes it a regression guard rather than a reproduction — which is a perfectly good thing to have, and I am glad it exists, but it should not be counted as the second half of the red-without-fix evidence. Please pin the one fragment the code actually emits.

I am asking for both because this repo holds itself to "green means fixed", and a test that can pass on a NULL is the exact shape that erodes it.

**Three stale comments to sweep while you are in there.** `src/mcp/mcp.c` lines 7735, 7769 and 7818 still document `phase = "crash"|"hang"`. Your change adds `"error"` as a third value of that enumeration, so those three now describe a contract that is one case short — and `phase` is a value consumers can switch on.

**One thing worth knowing rather than fixing.** `INDEX_SKIPPED_FILE_CAP` is 5, so beyond five quarantined files the response still reports an accurate `count` and sets `truncated`, but the individual paths never appear — they are only in the logfile. Your AOSP run quarantined two files, so this was never exercised. It matters more after your change than before it, because a graceful nonzero exit from an internal parse limit is a far more common trigger than a crash, and a repo full of machine-generated files could quarantine dozens. Not a blocker, and not something you introduced; I mention it because you are the person who now has the field data to say whether five is the right number.

Related, and the reason I am comfortable with the trade: a quarantined file leaves the run reporting `status: "indexed"` with a thinner graph. That is pre-existing behaviour for crash and hang, and your change extends it to a much more frequent trigger class. I still think it is clearly the right call — losing an entire directory's work to one bad file is worse than losing the file — and `skipped_count` being emitted unconditionally, even as 0, is what makes it honest rather than silent.

**One timing note:** your "191 tests passing" covers `a7a9f7c1`, but the branch has taken two more merges from `main` since (the head is now `d37334dc`) and thirteen checks are still running. I will re-check against the current head rather than the tree you measured.

The AOSP-scale field validation remains the strongest evidence any PR in this area has brought. Get those two test fixes in and I will merge it.

shashwatMishra96:
Thanks for the review.

I’ve incorporated both requested fixes into the PR:

1. **Tightened Systemic Test**

   * Updated `idxpar_systemic_exit_nonzero_give_up_check` to assert the exact JSON contract returned for a systemic `give_up`.
   * The test now also explicitly fails with `IDXPAR_NULL_RESP` if the response is `NULL`.

2. **Comment Sweep**

   * Updated the three phase-related comments in `src/mcp/mcp.c` to explicitly include `-error` for consistency with the actual behavior.

All **205 tests in `suite_mcp` are passing**.

DeusData:
Merged in 4b3ab38e. Thank you, @shashwatMishra96 — a first contribution that turns a whole-chunk abort into a surgical, visible, self-healing quarantine is exactly the kind of failure-path engineering this codebase needs more of.

What earned the merge, on the record: both 20 August conditions were delivered next-day and verified in the head diff (the NULL response now fails as `IDXPAR_NULL_RESP`, and the systemic test pins the exact `"status":"error"` / `"outcome":"exit_nonzero"` fragments — which are byte-for-byte what production emits). The primary test was verified genuinely red without the fix, and its double assertion — continuation via `search_graph` AND visibility via the quarantine phase in the response — is the shape the data-quality doctrine demands: honest, not silent. The quarantine feeds the persisted coverage rows, so `index_status` and `check_index_coverage` report it after the run, and the per-run reset means a fixed file returns on the next index with no permanent skiplist.

**One follow-up is ours, discovered in the final review and deliberately not put on you:** the daemon's supervisor (`src/daemon/application.c`) is an independent implementation whose attributable-failure predicate still accepts only `CRASH || HANG` (:1188) — so the daemon entry point currently keeps the old abort behaviour on `EXIT_NONZERO` while your path quarantines. Bringing it to parity needs a daemon-side test of its own and lands as maintainer follow-up work; your design is the template it will follow.

Also swept into that follow-up: the three remaining crash/hang-only comment mentions in `cbm.c` (:607, :638, ~:702).

Your AOSP field report — two quarantines across an 11.35M-node index instead of two lost runs — is the whole argument for this change in one sentence.

Issue #1388, opened 2026-07-31
hook-augment emits nothing on 0.9.1-rc.1, including payloads that work on 0.9.0


dergachoff:
## Problem

`hook-augment` on v0.9.1-rc.1 (macOS arm64, release tarball, permanent daemon active) exits 0 with no stdout/stderr for every payload, including ones 0.9.0 answers.

## Evidence

One-file Python repo, indexed by the rc.1 binary:

```
echo '{"session_id":"probe","hook_event_name":"PreToolUse","tool_name":"Grep","tool_input":{"pattern":"create_stripe_checkout","path":"<repo>"},"cwd":"<repo>"}' | codebase-memory-mcp hook-augment
```

|binary|project name|result|
|-|-|-|
|0.9.0|default path slug|emits additionalContext, 1 symbol|
|0.9.1-rc.1|default path slug|silent, exit 0|
|0.9.1-rc.1|custom name|silent, exit 0|

SessionStart and SubagentStart payloads are silent too, so the `cbm-subagent-reminder` hook the rc installer adds is a no-op. `cli search_graph` returns correct results for the same projects, so this is specific to the hook path. Custom-name resolution from #1069 worked when I ran main at d587deab; on rc.1 both custom and default-slug are silent.

## Expected fix

hook-augment emits additionalContext again for indexed projects; a regression test on the release path would catch the next one.

DeusData:
Thank you for testing known-good 0.9.0 payloads against `0.9.1-rc.1`. The hook path has a tight default deadline and silent skip behavior, so a zero-output result can hide whether startup, daemon dispatch, or augmentation consumed the budget. I have routed this as a high-priority `0.9.1-rc` stability regression; the red case should assert both bounded completion and visible diagnostics for each failure stage.

dergachoff:
Retested on v0.10.8: `hook-augment` is still silent here (fresh one-file repo, default-slug and custom-named index, PreToolUse and SessionStart, warm daemon), and `hook-augment-timeouts.log` gets `deadline_exceeded ms=2000` each time. With `CBM_HOOK_DEADLINE_MS=20000` it emits correctly in ~2.4 s, of which ~2.3 s is user CPU before any hook work — the startup fingerprint hash tracked in #1335. So I think this issue is the same mechanism as #1335 and #1767 should close both; happy for this to be deduped into that one.

Issue #1387, opened 2026-07-31
install on 0.9.1-rc.1 removes existing Claude Code hooks instead of installing them


dergachoff:
macOS arm64, manual tarball install of v0.9.1-rc.1 over a working 0.9.0 setup (release-asset binary, checksums OK).

`install -y` prints this for Claude Code:

```
hooks: PreToolUse Grep/Glob search augmentation + PostToolUse Read coverage (non-blocking)
hooks: SessionStart (MCP usage reminder on startup/resume/clear/compact)
hooks: SubagentStart (MCP usage reminder for subagents)
```

What actually happened to `~/.claude/settings.json` (diffed against a pre-install backup):

- my existing `Grep|Glob` PreToolUse entry (→ `cbm-code-discovery-gate`) — **removed**
- all four SessionStart reminder entries (startup/resume/clear/compact) — **removed**
- PostToolUse Read coverage — never added
- SubagentStart → new `cbm-subagent-reminder` — added (the only thing that was)

So the summary claims three hook groups, installs one, and deletes the two I had. Running `install -y` a second time doesn't restore anything, so it's not a remove-then-migrate that got interrupted — the re-add just never happens. Grepped `~/.claude/settings.json` and `~/.claude.json` to make sure they weren't written somewhere else: nothing. The script files themselves (`cbm-code-discovery-gate`, `cbm-session-reminder`) were left untouched, only the settings entries vanished.

`--dry-run` prints the same three-group summary, so you can't catch it before it bites either.

Workaround: re-add the entries by hand from a settings backup.

Happy to share the full settings diff if useful.

DeusData:
Thank you for capturing the before/after hook configuration. The installer is intended to upsert and preserve existing hook entries, so removing them during an RC install is destructive behavior rather than a migration choice. I have routed this as a high-priority `0.9.1-rc` integration regression, with preservation of unrelated and existing hooks as the binding test contract.

PR #1441, opened 2026-08-04
fix(hook): surface daemon build conflicts to the hook caller (#1388)

DeusData:
Part of #1388 / #1335 (the silent-skip half).

## Root cause (reproduced on main with two different-fingerprint builds)

Install a new build while the old daemon stays warm — the reporter's rc.1-over-0.9.0 setup — and every hook invocation dies in the **version-cohort conflict** path, which reported to **stderr only**. stdout is the only hook channel Claude Code surfaces, so in-session this is eternal silence. The connect-failure path additionally misclassified a conflicted daemon as *absent* and suggested `daemon start`, which cannot heal a build conflict. Manual terminal probes were silent too whenever the throttled notice window had been consumed by earlier hook traffic.

## Fix

- Both conflict sites (version-cohort early return + hello-level `CONNECT_CONFLICT`) now emit a **throttled stdout systemMessage** naming the real state and the actionable step: `daemon stop`, then retry.
- stderr keeps/gains the unconditional formatted conflict detail for terminal diagnosis.
- Absent-daemon reporting is unchanged.

## Verification

- New `CBM_ENABLE_TEST_SEAMS`-only env override (`CBM_TEST_HOOK_CLIENT_BUILD`) lets one binary present a foreign fingerprint as a hook client — the conflict becomes reproducible without a second build.
- `tests/test_hook_conflict_notice.sh` (wired as scripts/test.sh **Step 5d**, running on the step-5 TEST_SEAMS binary): asserts fail-open exit 0, the stderr conflict line, and the stdout systemMessage with restart guidance. **RED** before the fix (empty stdout), **GREEN** after, **RED** again on revert of the exercised call site.
- `lint-ci` clean.

## Remaining on #1388

Whether rc.1 had an additional regression beyond this reporting gap needs a retest on the next published build — the reporter's matrix showed main @ d587dea working, and with this change a conflicted daemon is at least loudly diagnosable instead of silent.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -282,6 +282,12 @@ CBM_TEST_BINARY="$WATCHDOG_BINARY" bash "$ROOT/tests/test_worker_watchdog.sh"
 echo "=== Step 5c: worker error-response transport regression ==="
 bash "$ROOT/tests/test_worker_error_response.sh"
 
+# Step 5d: a hook client facing a daemon BUILD CONFLICT must surface a stdout
+# systemMessage (stderr is invisible to the hook caller). Needs the TEST_SEAMS
+# binary from step 5's build (CBM_TEST_HOOK_CLIENT_BUILD forces the conflict).
+echo "=== Step 5d: hook daemon-conflict notice regression (#1388) ==="
+bash "$ROOT/tests/test_hook_conflict_notice.sh"
+
 # Step 6: security-strings URL allow-list regression. The MSYS2 CLANG64 toolchain
 # bakes its package-tracker URL into the static Windows .exe; the binary string
 # audit must allow-list it (Windows-only — Linux smoke never saw it).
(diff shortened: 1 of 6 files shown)


DeusData:
Correction for the record, since commit d5eb22d ("ci: re-trigger PR validation") states a cause that turned out to be wrong.

That commit message claims GitHub dropped the `pull_request` synchronize dispatch. It did not. The actual cause was that this branch had become **conflicted with main** (`mergeStateStatus: DIRTY`) after #1435 landed a test at the same anchor in `tests/test_cli.c`. GitHub cannot compute a merge ref for a conflicted PR, so it does not dispatch `pull_request` workflow runs at all — which is also why close/reopen and a check-suite re-request both failed to help. Resolving the conflict (keeping both the #1387 and #1388 tests) immediately restored dispatch, and the full matrix went 33/33 green.

The empty commit itself is harmless, but the diagnosis in its message is not accurate — anyone debugging a "CI never started" symptom later should check `gh pr view <n> --json mergeStateStatus` first. Leaving history untouched rather than rewriting main.

PR #1442, opened 2026-08-04
fix(mcp): skip direct store open for invalid project names

### LINKED ISSUE #1425: Invalid project name resolves to empty db path — daemon litters .corrupt.<hex> files into its cwd
## Problem

Querying any tool with a project name that fails `cbm_validate_project_name` (space, `#`, unicode — e.g. a repo dir named `#upstream`) makes the daemon drop a 4 KB `.corrupt.<hex>` file into whatever directory it was started from — typically some unrelated user repo. The tool caller just gets a normal "project not found" error, so nothing points at the litter, and it repeats on every such query.

## Evidence

Chain (same on v0.9.1-rc.1 and main @ d6be58e):

- `project_db_path()` (src/mcp/mcp.c:1853) yields `""` on validation failure; `resolve_store_internal()` (mcp.c:2094) passes it straight to `cbm_store_open_path_query("")`.

- SQLite treats `""` as an anonymous temp db, so the open succeeds; `cbm_store_check_integrity()` fails (no `projects` table) and the healthy temp db is declared corrupt.

- `quarantine_corrupt_store()` renders `"%s.corrupt.%016llx"` with the empty prefix (mcp.c:1889) — a relative path, created in the daemon's cwd.

Log signature: `store.auto_clean project=… path= action=corrupt_generation_quarantined backup=.corrupt.0001769cd4982b97` — note empty `path=`.

Live repro: `tools/call search_graph` with `"project":"bad name"` → clean error to caller, new `.corrupt.*` file in daemon cwd.

## Expected fix

Skip the direct open/recovery in `resolve_store_internal()` when the path is empty (one caller already guards the `""` sentinel, mcp.c:7814) and fall through to the existing fallback scan — which as a bonus can resolve legacy dbs whose internal name predates validation. Belt-and-braces: `quarantine_corrupt_store()` refuses an empty/relative path. Happy to PR: two guards + a test asserting error reply, no `.corrupt.*` file, no `auto_clean` log line.

Related but distinct: #1037 / #1206 / #557 / #391 are wrong corruption verdicts on real project dbs; this quarantines a db that never existed.

dergachoff:
**Problem**

A query with an invalid project name (`bad name`, `#upstream`) makes the server write a `.corrupt.<hex>` file into the process's cwd — for the daemon, whatever repo it was started in. One 4096-byte file per query, invisible to the caller, who just gets "project not found". Fixes #1425.

**Root cause**

`project_db_path()` returns `""` on validation failure. `resolve_store_internal()` passed it to `cbm_store_open_path_query("")`, which SQLite opens as an anonymous temporary database. The temp db fails the integrity check (no `projects` table), so the recovery path "quarantined" it: `"%s.corrupt.%016llx"` of an empty prefix is a relative path, created in cwd.

**Fix**

- `resolve_store_internal()`: skip the direct open when the path is empty, fall through to the #704 internal-name fallback scan. A typo still returns not-found; a legacy db with an invalid internal name can still resolve.
- `quarantine_corrupt_store()`: refuse an empty path (defense in depth; unreachable via the resolver now).

No behavior change for valid names: validation passing always yields a non-empty `cache_dir/<project>.db` path.

**Validation**

- New test `tool_query_invalid_project_name_no_corrupt_litter`: query with `project:"bad name"` from a scratch cwd asserts the not-found reply, zero recovery-lease acquisitions, and zero `.corrupt` entries in cwd and cache dir.
- Red without either guard (log shows the reported `path=` signature); green with both. `mcp` + `mcp_mutation_guard`: 214 passed, 2 skipped.
- Live-reproduced the litter on 0.9.1-rc.1 before the fix (details in #1425).

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -2001,6 +2001,14 @@ static bool quarantine_step_allowed(cbm_mcp_server_t *srv, const char *step) {
  * only recoverable generation. */
 static bool quarantine_corrupt_store(cbm_mcp_server_t *srv, const char *project, const char *path,
                                      char *backup_out, size_t backup_out_size) {
+    /* Never quarantine an empty path: SQLite opens "" as an anonymous temp db,
+     * and "%s.corrupt.*" of an empty prefix would land as relative litter in
+     * the process cwd instead of beside the store. */
+    if (!path || path[0] == '\0') {
+        cbm_log_error("store.auto_clean_failed", "project", project, "path", path ? path : "",
+                      "reason", "empty store path");
+        return false;
+    }
     char backup[CBM_SZ_2K];
     char pending[CBM_SZ_2K];
     if (!reserve_unique_corrupt_pending(path, pending, sizeof(pending), backup, sizeof(backup))) {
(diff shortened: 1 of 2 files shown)


dergachoff:
Duplicate of #1437 (same guards, same test) — closing in favor of the maintainer's version. Happy to port anything if it turns out mine covered a case yours doesn't.

PR #1444, opened 2026-08-04
build(windows): do not pass -z noexecstack when cross-compiling to PE

mlandolfi90:
`ELF_HARDENING_FLAGS` is gated on `IS_LINUX`, which comes from `uname` and therefore describes the **host**. Cross-compiling to Windows from a Linux container satisfies that gate while the compiler emits PE, and the link fails:

```
lld: error: unknown argument: -z
clang: error: linker command failed with exit code 1
```

Reproduced on `main` with the repo's own recipe:

```
docker compose -f test-infrastructure/docker-compose.yml run --rm build-windows
```

The fix restores the flag's documented intent — the comment above it already states it is ELF-only and "meaningless for PE, so it is gated". The gate just tested the wrong end of the toolchain. `IS_MINGW` is derived from the compiler's own `_WIN32` define, so adding it names the target rather than the host.

Native Windows CI (MSYS2 CLANG64) is unaffected: `uname` reports `MINGW64_NT` there, so `IS_LINUX` was already `no`. Linux and macOS are unaffected. Only the Linux-host → Windows-target combination changes, and only to stop emitting a flag that cannot apply.

On CONTRIBUTING's "Build system / Makefile changes — beyond trivial fixes": I read this as a trivial fix (one gate condition, no behaviour change on any venue CI builds). Happy to move it to an issue first if you would rather discuss it.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -162,7 +162,11 @@ endif
 # shipped artifact. ELF-only: Apple's ld rejects -z noexecstack outright and it
 # is meaningless for PE, so it is gated rather than made "common".
 ELF_HARDENING_FLAGS :=
-ifeq ($(IS_LINUX),yes)
+# IS_LINUX comes from uname, i.e. the HOST. Cross-compiling to Windows from a
+# Linux container satisfies it while emitting PE, and lld then rejects the
+# unknown -z argument. Gate on the TARGET too: IS_MINGW is derived from the
+# compiler's own _WIN32 define, so it names what is actually being produced.
+ifeq ($(IS_LINUX)$(IS_MINGW),yesno)
 ELF_HARDENING_FLAGS := -Wl,-z,noexecstack
 endif
 


DeusData:
Reviewed and this is one we want — verified it does not weaken ELF hardening on Linux or macOS; it only stops passing a flag that has no meaning for a PE target.

It has picked up a conflict in `Makefile.cbm` since you opened it (that file has changed a fair bit this week — sanitized-build detection and the release-candidate work). Could you rebase onto current `main`? It should be mechanical, and I will merge it once CI is green.

Sorry it sat for a while.

DeusData:
Thank you for this, and sorry it went unanswered.

Superseded, though your reasoning was exactly right: `IS_LINUX` comes from `uname` and therefore describes the *host*, so cross-compiling to PE from a Linux container satisfied it and lld then rejected `-z noexecstack`. `main` now expresses the same gate by nesting the target test inside the host test:

```make
ifeq ($(IS_LINUX),yes)
ifneq ($(IS_MINGW),yes)
ELF_HARDENING_FLAGS := -Wl,-z,noexecstack -Wl,-z,separate-code
endif
endif
```

Semantically identical to your `yesno` form. Same diagnosis, arrived at independently — which is a reason to be sorry we didn't just take yours when you offered it.

DeusData:
Closing as superseded — `main` nests the MinGW test inside the Linux gate, which is semantically identical to your `yesno` form. Your reasoning about `IS_LINUX` describing the *host* was exactly right, and we reached it independently rather than just taking yours. Thank you, and sorry for the wait.

PR #1445, opened 2026-08-04
fix(windows): stat paths through the long-path-aware converter

mlandolfi90:
`cbm_path_info_utf8` converts with `cbm_utf8_to_wide`, the plain UTF-8 converter, while every other filesystem entry point in the same file — `cbm_opendir`, `cbm_fopen`, `cbm_rename_replace`, `cbm_canonical_path` — uses `cbm_path_to_wide`, which canonicalises and applies the `\\?\` prefix past the legacy limit.

The mismatch is observable within a single walk: `cbm_opendir` successfully enumerates a deep directory *because* it prefixes, then the per-entry stat fails *because* it does not. `GetFileAttributesExW` gets a bare path and returns `ERROR_PATH_NOT_FOUND`.

Consequences, both in the semantic-manifest code:

- `semantic_manifest_walk_controls` — one unstattable entry aborts the entire walk (`CBM_NOT_FOUND`).
- `semantic_manifest_hash_file` — returns `CBM_NOT_FOUND`, so the file is silently dropped from the manifest.

Either way the index degrades with no error surfaced. Machines with `LongPathsEnabled=0` are the exposed case, and a mingw-cross PE carries no `longPathAware` manifest, so flipping that registry policy is not a workaround there.

One-word change, consistent with the convention the rest of the file already follows.

### DIFF
--- a/src/foundation/compat_fs.c
+++ b/src/foundation/compat_fs.c
@@ -125,7 +125,7 @@ int cbm_path_info_utf8(const char *path, cbm_path_info_t *out) {
     if (!path || !out) {
         return CBM_NOT_FOUND;
     }
-    wchar_t *wpath = cbm_utf8_to_wide(path);
+    wchar_t *wpath = cbm_path_to_wide(path);
     if (!wpath) {
         return CBM_NOT_FOUND;
     }


mlandolfi90:
On CI, since the acknowledgement asks for green or an explanation:

**`test / test-windows-guards` — I believe this is a flake, not this change.** `test_non_ascii_path.py` went red on exactly one of its four variants:

```
[PASS] non-ascii/latin1_accents nodes=12 edges=21 definitions=5 (baseline 12/21/5)
[FAIL] non-ascii/cyrillic       nodes=None edges=None definitions=None
[PASS] non-ascii/cjk            nodes=12 edges=21 definitions=5
[PASS] non-ascii/greek          nodes=12 edges=21 definitions=5
```

Two reasons I do not think the diff can explain it:

1. **This change is a no-op for the path under test.** `cbm_path_to_wide` returns exactly what `cbm_utf8_to_wide` returns unless the path is drive-absolute *and* `wcslen >= 240`; below the threshold it is the same early return. The guard's path (`C:\Users\runneradmin\cbm-windows-guards-root-<32 hex>\cbm_win_nonascii_<8>\проект_repo\src\math.ts`) is roughly 126 wide chars, so the branch this PR touches is never taken. If it did regress non-ASCII handling, it would also have to explain why the Cyrillic variant differs from the Greek and CJK ones, which take the identical code path.

2. **The failure shape is wrong for this bug.** #636 produced File/Folder nodes with *zero definitions*. Here the counts are `None` — the tool call returned nothing at all, which is a server/startup failure, not an extraction one.

A re-run would settle it. Happy to be shown wrong if it reproduces.

**`security / codeql-gate`** timed out with the analysis still `queued` after 90 polls (`BLOCKED: CodeQL timeout`) — no alerts, just never scheduled. I opened five PRs in quick succession, so this may well be queue pressure I caused; apologies if so.

DeusData:
@mlandolfi90 — thank you for this, and apologies it has sat since 04 August. Here is exactly where it stands, because the failing check is worth seeing rather than guessing at.

`test-windows-guards` fails on the non-ASCII guard:

```
[FAIL] non-ascii/cyrillic   nodes=None edges=None definitions=None (baseline 12/21/5) name=None
```

A Cyrillic-named fixture indexes to **nothing** where the baseline expects 12 nodes / 21 edges / 5 definitions. Given the PR routes `stat` through the long-path-aware converter, that guard is the one most likely to react to the change — so this reads as a real interaction rather than noise, and I would rather show it to you than re-run it and hope.

Two things worth knowing before you dig in:

1. **Your base is from 04 August and main has moved a long way since** — v0.10.2 and v0.10.3 both shipped, including several Windows path and ownership fixes (most recently #1583, which corrected the activation ownership check to compare against the token's *owner* SID rather than its user SID). It is entirely possible this failure is an artefact of the old base rather than your diff. A rebase would tell us in one run.

2. **If you would rather not spend the time**, say the word and we will rebase it for you and take the failure from there. The change itself is one we want — non-ASCII and extended-length paths on Windows are a recurring source of reports, and routing `stat` through the converter is the right shape.

The `codeql-gate` failure alongside it is almost certainly unrelated residue from an Actions incident and can be ignored until the real one is settled.

DeusData:
Thank you for this, and apologies for leaving it unanswered so long.

This is now superseded: `cbm_path_info_utf8` on `main` routes through the long-path-aware converter already (`src/foundation/compat_fs.c`, `wchar_t *wpath = cbm_path_to_wide(path)`), landed in 8018561c on 2026-08-08.

You were right about the underlying rule, and it is one we take seriously enough that it is now a standing review check on every diff: repo and user paths must never reach a raw `fopen`/`stat`, or non-ASCII paths break on Windows. Thanks for pushing on it.

DeusData:
Closing as superseded — `cbm_path_info_utf8` routes through `cbm_path_to_wide` on `main` since 8018561c. The rule you were enforcing is now a standing check on every diff we review. Thank you.

PR #1446, opened 2026-08-04
fix(daemon): print the validation detail when the secure endpoint cannot be created

mlandolfi90:
The MCP-client startup path swallows the reason for a refused endpoint:

```c
cbm_daemon_ipc_endpoint_t *endpoint = cbm_daemon_bootstrap_endpoint_new(NULL);
if (!endpoint) {
    (void)fprintf(stderr, "codebase-memory-mcp: secure daemon endpoint could not be created\n");
    return EXIT_FAILURE;
}
```

Every refusal inside the ancestor DACL walk — owner mismatch, an untrusted grant, a failed repair — arrives as that one sentence. `cbm_daemon_ipc_validation_detail()` has already recorded exactly which component failed and why, and the worker path a few hundred lines up already prints it for the identity check. This just does the same at the endpoint site.

Concretely, this turns

```
codebase-memory-mcp: secure daemon endpoint could not be created
```

into

```
codebase-memory-mcp: secure daemon endpoint could not be created - C:\Users\<user>\AppData: DACL entry 0 grants mutation rights 0x000d0152 to untrusted identity (other S-1-15-3-...)
```

Diagnosing the first form took a purpose-built instrumented binary. The second names the directory, the mask and the SID.

No behaviour change beyond the message; the detail buffer is a static already populated on this path.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2119,7 +2119,10 @@ int main(int argc, char **argv) {
 
     cbm_daemon_ipc_endpoint_t *endpoint = cbm_daemon_bootstrap_endpoint_new(NULL);
     if (!endpoint) {
-        (void)fprintf(stderr, "codebase-memory-mcp: secure daemon endpoint could not be created\n");
+        const char *validation_detail = cbm_daemon_ipc_validation_detail();
+        (void)fprintf(stderr,
+                      "codebase-memory-mcp: secure daemon endpoint could not be created%s%s\n",
+                      validation_detail[0] ? " - " : "", validation_detail);
         return EXIT_FAILURE;
     }
 


mlandolfi90:
CI note: the only red is `security / codeql-gate`, which timed out with the analysis still `queued` after 90 polls (`BLOCKED: CodeQL timeout`) — no alerts were raised, it was never scheduled. I opened five PRs in quick succession, so this may be queue pressure I caused; apologies if so. Everything else is green.

DeusData:
Thank you for this, and sorry for the long silence.

This one is now superseded — the validation detail is printed on `main` today. `src/main.c:2774` calls `cbm_daemon_ipc_validation_detail()` on the daemon-side failure, and the CLI-side twin landed last night (4cbecb62, #1627) so the `secure CLI coordination could not be created (endpoint)` path names the directory and the rule that refused it too. Both ship in v0.10.5.

Your diagnosis was right, and it was the right fix to ask for: four separate reporters hit that bare message, and one of them ended up building an instrumented binary just to discover a single ACE was the cause. That should never have been necessary, and this is exactly the change that stops it.

Nothing here reflects on the patch — it sat unreviewed while the same problem got fixed from another direction, which is our failure of process rather than anything about your work. Your #1447 fared better: it shipped, with you named as co-author. See my comment there.

DeusData:
Closing as superseded — `src/main.c:2774` prints `cbm_daemon_ipc_validation_detail()` on `main` today, and the CLI-side twin landed in #1627. Both ship in v0.10.5. Your diagnosis was right, and reporters had been building instrumented binaries to get this information; thank you for pushing on it.

PR #1447, opened 2026-08-04
fix(windows): accept AppContainer capability grants on runtime ancestors

mlandolfi90:
## Symptom

On an ordinary Windows 11 developer machine, every CBM process died at startup:

```
codebase-memory-mcp: secure daemon endpoint could not be created
```

No CBM process was competing, and the same machine had run a daemon for days on an older build. The cause is environmental and, I suspect, increasingly common: an agent sandbox (Codex) had stamped the user's `AppData` with capability ACEs.

```
C:\Users\<user>\AppData  ->  S-1-15-3-3557520199-...  FullControl  (inherited to AppData\Local)
```

`win_private_directory_tree_secure` walks every ancestor of the runtime directory and refuses any mutation-granting ACE whose SID is not the exact user, SYSTEM, Administrators, TrustedInstaller, CreatorOwner (inherit-only) or OWNER RIGHTS. A capability SID matches none of them, so `cbm_daemon_ipc_endpoint_new` returns NULL and the process exits — including plain `--version`-adjacent paths that construct an endpoint.

## Why I think accepting them on ancestors is sound

A capability SID (`S-1-15-3-…`, `SECURITY_APP_PACKAGE_AUTHORITY`) grants only to AppContainer processes that the same user's tooling provisioned. It cannot be used by another account, and it does not confer rights on a normal process. This is the same shape of argument the OWNER RIGHTS acceptance in 8b4e0fb5 already makes: the grant is reachable only by the party the walk is protecting.

The relaxation is deliberately narrow:

- **Ancestors only.** `win_directory_component_secure` passes `ancestor_capability_ok = true`.
- **The runtime directory itself is unchanged.** `win_runtime_directory_secure` passes `false`, so it still demands the exact-user owner and a protected DACL, and still re-stamps it.
- The SID test checks the identifier authority (`15`) and the capability RID class (`3`) explicitly, so it cannot widen to other `S-1-15-*` classes.

## Note on the trust model

This does move a trust boundary, so I understand if you would rather discuss it in an issue before reviewing code — happy to convert. I am also open to a narrower shape if you prefer one: gating it behind an opt-in, or restricting acceptance to ancestors that lie inside the user's own profile.

What I would push back on is leaving it as-is: the failure gives the user one sentence with no cause, and the only remedies available to them are removing ACEs their sandbox depends on, or not running CBM.

### DIFF
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -3899,6 +3899,13 @@ static bool win_sid_is_trusted_installer(const uint8_t *sid, size_t sid_length) 
     return true;
 }
 
+static bool win_sid_is_app_capability(const uint8_t *sid, size_t sid_length) {
+    /* S-1-15-3-…: SECURITY_APP_PACKAGE_AUTHORITY, capability RID class. */
+    return windows_sid_valid(sid, sid_length) && sid[1] >= 1U && sid[2] == 0U && sid[3] == 0U &&
+           sid[4] == 0U && sid[5] == 0U && sid[6] == 0U && sid[7] == 15U &&
+           win_sid_read_u32_le(sid + 8U) == 3U;
+}
+
 static bool win_sid_trusted(win_security_t *security, PSID sid) {
     if (!security || !sid || !security->is_valid_sid(sid)) {
         return false;
@@ -3911,7 +3918,8 @@ static bool win_sid_trusted(win_security_t *security, PSID sid) {
 }
 
 static bool win_bounded_sid_trusted(win_security_t *security, const uint8_t *sid,
-                                    size_t sid_capacity, bool creator_owner_inherit_only) {
+                                    size_t sid_capacity, bool creator_owner_inherit_only,
+                                    bool ancestor_capability_ok) {
     if (!security || !sid || sid_capacity < 8U || sid[1] > 15U) {
         return false;
     }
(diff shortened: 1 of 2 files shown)


mlandolfi90:
CI update — the platform-wide red was real and mine, and it is worth naming precisely because it is a good guard.

`tests/test_windows_bundle_contract.sh` pins the ancestor validation call **by literal source text**:

```python
"win_file_security_secure(security, directory, false, mutation)",
```

Adding the capability parameter changed that text, so the contract failed on Linux and macOS too — the check reads the source rather than compiling it. That is the guard doing its job: it is exactly the line that decides how much an ancestor is allowed to grant, and it should not move quietly.

I have pushed an update to the needle, and used the occasion to pin the property this PR actually needs to be true rather than just the call shape:

```python
"win_file_security_secure(security, directory, false, mutation, true)",
"win_private_mutation_rights(), false)",
"win_sid_is_app_capability",
```

The middle one is the important addition: it pins that `win_runtime_directory_secure` still passes `false`. If someone later let a capability ACE satisfy the runtime directory the daemon actually publishes into — the thing this relaxation must never reach — that contract now fails.

If you would rather this guard not be touched by a contributor PR at all, say so and I will drop the test commit and let you land the needle change yourself.

The remaining `security / codeql-gate` red is a timeout with the analysis still `queued` (`BLOCKED: CodeQL timeout`), not an alert.

DeusData:
Thank you for this — and I owe you an apology, because the way this played out is not how it should have gone.

Your approach shipped. It is on `main` as 9a53074c (PR #1623), and the commit says so and names you as co-author:

> Approach and the ancestor-only boundary follow @mlandolfi90's PR #1447, extended from capability SIDs to package SIDs.

The one thing we changed is the SID class. You tested `S-1-15-3-*` (capability SIDs); on the machines in #1533 and #1574 the blocking ACE turned out to be `S-1-15-2-*` — a **package** SID, resolving through `HKCR\...\AppContainer\Mappings` to a shipping desktop application that no user could reasonably uninstall. Covering capability SIDs alone would have left the headline case still failing, so the merged change tolerates both, under identifier authority 15.

Your ancestor-only boundary is the part that made it safe to do at all, and it is preserved exactly: the private runtime directory is still validated with `ancestor=false` and still demands the exact current user with a protected DACL. The flag is threaded explicitly through `win_file_security_secure` and `win_file_acl_secure` rather than inferred, so the strict path cannot pick up the tolerance by accident.

It ships in v0.10.5, going out today.

What I'm sorry about: you opened four PRs on 2026-08-04 (#1444, #1445, #1446, #1447), three of which turned out to be fixed independently, and nobody told you — including on this one, where your work went into the product. That is a bad experience for someone who showed up with a correct diagnosis and a working patch, and the silence was ours, not a judgement on the work. I'm going through the other three now.

DeusData:
Closing as shipped: your approach is on `main` in 9a53074c (#1623) with you as co-author, extended from capability SIDs to package SIDs to cover the field case. It ships in v0.10.5 today. Thank you — the ancestor-only boundary you drew is what made this safe to do at all, and it is preserved exactly.

PR #1448, opened 2026-08-04
fix(windows): hide the console windows of spawned child processes

mlandolfi90:
Two `CreateProcessW` call sites in `src/foundation/` spawn children without `CREATE_NO_WINDOW`, so on Windows each one flashes a console window:

- `cbm_subprocess_spawn_win` (`src/foundation/subprocess.c`) — the supervisor → index-worker spawn.
- `cbm_popen_isolated` (`src/foundation/compat_fs.c`).

`src/daemon/bootstrap.c` already passes the flag for the daemon self-spawn, so this brings the remaining two in line with it.

User-visible effect: when the MCP server is driven by an editor/agent over stdio, indexing a repository pops console windows onto the desktop — one per worker spawn. Nothing breaks, but on a machine that reindexes on file changes it is a steady stream of windows stealing focus.

`CREATE_NO_WINDOW` only suppresses console allocation; it does not detach the process, change the process group, or affect the inherited handle list, so the existing `EXTENDED_STARTUPINFO_PRESENT | CREATE_SUSPENDED | CREATE_NEW_PROCESS_GROUP` semantics and the worker-log handle plumbing are untouched.

### DIFF
--- a/src/foundation/compat_fs.c
+++ b/src/foundation/compat_fs.c
@@ -310,8 +310,9 @@ static FILE *cbm_popen_isolated(const char *cmd, const char **stage, DWORD *gle)
         *stage = "cmdline";
         *gle = ERROR_NOT_ENOUGH_MEMORY;
     } else {
-        created = CreateProcessW(app, wcmdline, NULL, NULL, TRUE, EXTENDED_STARTUPINFO_PRESENT,
-                                 NULL, NULL, &si.StartupInfo, &pi);
+        created = CreateProcessW(app, wcmdline, NULL, NULL, TRUE,
+                                 EXTENDED_STARTUPINFO_PRESENT | CREATE_NO_WINDOW, NULL, NULL,
+                                 &si.StartupInfo, &pi);
         if (!created) {
             *stage = "spawn";
             *gle = GetLastError();
--- a/src/foundation/subprocess.c
+++ b/src/foundation/subprocess.c
@@ -730,7 +730,8 @@ static int cbm_subprocess_spawn_win(cbm_subprocess_t *process) {
 
     PROCESS_INFORMATION child;
     ZeroMemory(&child, sizeof(child));
-    DWORD flags = EXTENDED_STARTUPINFO_PRESENT | CREATE_SUSPENDED | CREATE_NEW_PROCESS_GROUP;
+    DWORD flags = EXTENDED_STARTUPINFO_PRESENT | CREATE_SUSPENDED | CREATE_NEW_PROCESS_GROUP |
+                  CREATE_NO_WINDOW;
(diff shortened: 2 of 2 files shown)


rkisedu:
﻿## Windows reproduction (focus-steal) — confirms this PR is needed

**Host:** Windows 11, codebase-memory-mcp **0.9.0**, Cursor + Claude Code MCP stdio.

**Symptom:** With `auto_watch=true` (default), background git probes spawn visible `cmd.exe` / `OpenConsole.exe` ~every 1–2s. Windows steal keyboard focus — typing in the editor is interrupted continuously until the MCP/index workers are killed.

**Mitigation that works today:**
```
codebase-memory-mcp config set auto_watch false
codebase-memory-mcp config set auto_index false
```
Query tools remain usable; only continuous watch/reindex is unsafe until `CREATE_NO_WINDOW` lands on the two CreateProcessW sites this PR patches (`cbm_subprocess_spawn_win`, `cbm_popen_isolated`).

**Ask:** Please prioritize merge/release — this is a usability blocker on Windows for anyone leaving the MCP connected with the watcher on.

noctrex:
Thanks for this, it fixes the focus-steal for me too on Windows 11 (stdio MCP, `auto_watch=true`).

But I found one more, one spawn site looks missed: `cbm_exec_no_shell` in `src/foundation/compat_fs.c` also calls `CreateProcessW` without `CREATE_NO_WINDOW`. The `git / codesign / open` invocations routed through it still pop a console window. Same one-flag fix:

   ```diff
   -    if (!CreateProcessW(NULL, cmdline, NULL, NULL, FALSE, 0, NULL, NULL, &si, &pi)) {
   +    if (!CreateProcessW(NULL, cmdline, NULL, NULL, FALSE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
   ```
Can you add this also, please?

DeusData:
Merged as 98f292ec — thank you, this is a clean, well-targeted fix, and the two independent reproductions on the thread (@rkisedu, @noctrex) made it easy to land with confidence.

@noctrex — good catch on the third site, and you were right that it's a different function: this PR covers `cbm_popen_isolated` (compat_fs.c:313) and the subprocess spawn (subprocess.c:738), while `cbm_exec_no_shell` (compat_fs.c:688) still spawns with `flags = 0`. That's the one routing `git`/`codesign`/`open`, so those console windows would have survived this merge. I'm sending that as a follow-up rather than holding this PR, and you're credited on it.

For the record, I also checked the fourth `CreateProcessW` call site while reviewing — the detached daemon spawn in `daemon/bootstrap.c:756` — and it already sets `CREATE_NO_WINDOW`, so no change was needed there.

PR #1467, opened 2026-08-06
fix(daemon): add native FreeBSD process-image identity and /home log-path alias

PR9000:
## What does this PR do?

Adds native FreeBSD support to the daemon's process-image identity logic in `src/daemon/runtime.c` and scopes a security-sensitive `/home` log-path alias to FreeBSD only in `src/daemon/ipc.c`.

- `runtime_process_image_reference_acquire`/`_matches_process` were guarded only for `_WIN32`/`__APPLE__`/`__linux__`, with FreeBSD falling through to a stub that always returned `false`. This broke the daemon's build-fingerprint capture and made `install` refuse to proceed on FreeBSD ("active CBM sessions and operations could not be stopped safely"). Adds a `sysctl(CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, pid)`-based FreeBSD branch, mirroring the existing macOS/Linux pattern, and renames `runtime_linux_stat_same_image` to `runtime_posix_stat_same_image` now that the comparison logic serves all three platforms.

- `private_log_directory_path_copy` previously shared the `/home` root-owned alias between `__APPLE__` and `__FreeBSD__`. macOS never needed `/home` resolved in this security-sensitive step; the alias list is now split so Darwin keeps `{/tmp, /var}` and only FreeBSD gains `/home -> /usr/home`.

Split out of #1138 per review feedback — this covers the daemon-side platform port; CLI self-path resolution and the `activation_transaction.c` guard fix are in separate PRs.

## Checklist

### DIFF
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -1337,12 +1337,18 @@ static bool private_log_base_name_valid(const char *base_name) {
 }
 
 static char *private_log_directory_path_copy(const char *directory_path) {
-#ifdef __APPLE__
-    /* Darwin exposes the trusted top-level aliases /tmp -> /private/tmp and
-     * /var -> /private/var.  Resolve only those root-owned aliases before the
+#if defined(__APPLE__) || defined(__FreeBSD__)
+    /* Resolve only those root-owned aliases before the
      * component-wise O_NOFOLLOW walk.  Canonicalizing the complete caller path
      * would follow an attacker-controlled cache/log symlink and is forbidden. */
+#if defined(__APPLE__)
+    /* Darwin exposes the trusted top-level aliases /tmp -> /private/tmp and
+     * /var -> /private/var. */
     static const char *const aliases[] = {"/tmp", "/var"};
+#else
+    /* FreeBSD additionally exposes /home -> /usr/home as a root-owned alias. */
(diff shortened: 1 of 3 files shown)


DeusData:
Thank you for splitting the FreeBSD process-identity and log-path work into a focused change. Because this extends platform support and touches process identity and path handling, I have routed it to `0.11.0-rc` for a careful correctness and security-aware review. The community queue is currently full, so detailed feedback may take a little time.

DeusData:
Merged — thank you, and thank you for the discipline: splitting this out of #1138 exactly as asked, using the native KERN_PROC_PATHNAME sysctl with the O_NOFOLLOW/fstat double-check rather than a procfs shortcut, and scoping the /home→/usr/home alias strictly to FreeBSD. With the CLI self-path half already in (305b89bf), FreeBSD support is now coherent end-to-end. Honest-limits note we'll carry in the docs: there is no FreeBSD CI leg, so this path is community-verified — regressions reports welcome.

PR9000:
Quick note on the FreeBSD CI leg you mentioned: #1422 is already open and green (32/32 checks), just waiting in the queue. Once that lands, the 'community-verified' caveat can drop from the docs.

Issue #1085, opened 2026-07-14
Parallel pipeline drops all JSX-via-alias CALLS edges (21% of call graph); repros on this repo's own source


itayost:
## Summary

The parallel pipeline silently drops CALLS edges. On a 122-file Next.js/TypeScript repo it loses **21% of the call graph** (519 -> 408), and the default configuration is the broken one.

You already know the two pipelines disagree — `d9ea73f` says so:

> a new deliberately-RED `repro_seq_parallel_equivalence` tracks the remaining OPEN bug that the sequential and parallel pipelines systematically disagree (~3459 USAGE / ~1666 WRITES lines on xfs) -- **separate code paths, fix deferred**.

I could not find a GitHub issue for it, so here is one — with **the exact failure rule**, a **public corpus** that reproduces it, and the **file-count threshold** below which it is invisible.

`tests/repro/repro_parallel_determinism.c:54` hardcodes `RPD_CORPUS "/Users/martinvogel/perf-bench/linux/fs/xfs"` and `SKIP`s when absent, so CI never runs it. The header notes synthetic corpora failed to reproduce. Both problems are solved below.

Version: v0.9.0, macOS arm64.

## The failure rule

**An edge is dropped if and only if the target is a JSX component imported through a tsconfig `paths` alias.**

I classified all 519 sequential CALLS edges by import style x usage. The confusion matrix is perfect:

| import style | usage | kept | lost | loss |
|---|---|---|---|---|
| `@/` tsconfig alias | **JSX** | 0 | **122** | **100%** |
| `@/` tsconfig alias | call | 115 | 0 | 0% |
| relative `./` | JSX | 37 | 0 | 0% |
| relative `./` | call | 4 | 0 | 0% |
| external (npm) | call | 14 | 0 | 0% |

Alias-imported *function calls* resolve fine. Relative-imported *JSX* resolves fine. Only the intersection fails. The parallel `import_map` heuristic appears to apply alias mapping to call expressions but not to JSX element references.

Concrete case — `src/app/(dashboard)/_components/dashboard-kpis.tsx`:

```tsx
import { KpiCard } from "@/components/ui/kpi-card";   // alias import
...
<KpiCard ... />   // rendered at lines 32, 40, 47
```

```
sequential: DashboardKpis -> KpiCard   (strategy=unique_name)
parallel:   (no edge)
```

[...]

PR #1470, opened 2026-08-06
fix(configlink): order candidates canonically before the collector cap

### LINKED ISSUE #1468: pass_configlink: CONFIGURES edges are nondeterministic when candidates exceed the collector caps
Indexing the same corpus multi-threaded produces a different `CONFIGURES` edge set on every run, and a different set from the single-threaded reference. All observed divergence is `CONFIGURES`; every other edge type is byte-stable.

This is the same defect class as #998 (`SEMANTICALLY_RELATED` churn), in a pass that did not get the canonical-ordering treatment.

## Reproduction

Corpus: 4191-file TypeScript tree, `--mode moderate`, each run into a fresh `CBM_CACHE_DIR`. Fingerprint is the sorted `(source_qn, type, target_qn)` triple set read straight from the project db, compared with `LC_ALL=C`.

| run | edges | vs single-threaded |
|---|---|---|
| ST (reference) | 70809 | — |
| MT #1 | 70812 | 4 missing, 7 extra |
| MT #2 | 70864 | 6 missing, 61 extra |
| MT #3 | 70807 | 5 missing, 3 extra |

No two multi-threaded runs agree. Across the three runs, 86 differing edges — **100% of them `CONFIGURES`**, out of 414 `CONFIGURES` edges total.

## Cause

`strategy_key_symbols` fills two fixed-capacity arrays:

```c
config_entry_t config_entries[CBM_SZ_4K];
int config_count = collect_config_entries(vars, var_count, config_entries, CBM_SZ_4K);

code_entry_t code_entries[CBM_SZ_8K];
int code_count = collect_code_entries(gb, code_entries, CBM_SZ_8K);
```

Both collectors stop at the cap:

```c
for (int li = 0; labels[li] && n < max_out; li++)      /* Function, Variable, Class, Struct */
    for (int i = 0; i < count && n < max_out; i++)     /* silently stops at 8192 */
```

They walk `cbm_gbuf_find_by_label` in gbuf insertion order, which under parallel extraction is worker-merge order and varies run to run. On this corpus there are 12940 candidates (Function 7121, Variable 5524, Class 295) against the 8192 cap, so 4748 are dropped — and *which* ones depends on thread scheduling. Functions consume the first ~7121 slots, leaving roughly 1071 for Variables, so the cut lands mid-group where the ordering is unstable.

## Confirming the mechanism

A corpus with 2871 candidates — below the cap — is byte-identical across the single-threaded reference and three multi-threaded runs. The divergence appears only once the cap binds.

## Suggested fix

The same recipe #998 called for: order candidates by a pure content key before applying the cap, so the surviving set is a function of the inputs rather than of worker scheduling. Tie-breaks need to be content-only — node ids are handed out in merge order, so an id tie-break would reintroduce the same dependence.

Worth surfacing the truncation too; at present the caps drop candidates with no signal at all.

## Notes

- The count delta understates the divergence: losses and gains partly cancel, so a run can differ in dozens of edges while the total moves by three.
- `tests/repro/repro_parallel_determinism.c` covers the general parallel-determinism property but is pinned to an absolute corpus path and skips when absent, so this pass is not currently guarded in CI.

Environment: v0.9.0 (`9b9b6ff`), darwin-arm64, built from source.

Related: #998, #1085, #921, #923.

Studnicky:
Fixes #1468

## What does this PR do?

`collect_config_entries` and `collect_code_entries` fill fixed-capacity arrays and stop at `max_out`, walking gbuf label indexes in insertion order — which under parallel extraction is worker-merge order and varies run to run. On a repository with more candidates than the cap, the surviving set, and therefore the emitted `CONFIGURES` edges, depended on thread scheduling.

This sorts candidates by a pure content key (`qualified_name`, `file_path`, `start_line`, `name`) before the cap applies. Tie-breaks are content-only — node ids are handed out in merge order and belong in no canonical comparator.

It also logs `configlinker.truncated` when a collector fills to capacity, so the cap reports itself rather than dropping candidates silently.

## Why

Indexing the same tree twice produced a different `CONFIGURES` edge set each time, and a different set from the single-threaded reference — which breaks the deterministic-output property established in #921, makes A/B comparisons noisy, and churns incremental artifacts.

This is the recipe #998 called for ("sort candidates by (score, qn) instead of arrival order"), applied to the pass that did not get it.

## Testing

4191-file TypeScript corpus, `--mode moderate`, fresh `CBM_CACHE_DIR` per run, comparing sorted `(source_qn, type, target_qn)` triples:

| | ST vs MT | MT vs MT |
|---|---|---|
| before | 3–61 edges differ every run | all differ |
| after | identical across 4 runs | identical |

- All divergence before the change was `CONFIGURES` — 86 edges across 3 runs, of 414 total.
- Non-`CONFIGURES` edges are byte-identical before and after (70393), so the blast radius is contained to this pass.
- A corpus below the cap (2871 candidates) was already deterministic, which is what identifies the cap as the mechanism.

New test `configlink_candidate_truncation_is_order_independent` builds a corpus exceeding the code-entry cap and runs the pass over it twice, inserting candidates forwards and then backwards; the emitted `CONFIGURES` set must match. It fails on an unpatched tree (7 passed, 1 failed) and passes with the change (8 passed). The `configlinker.truncated` log fires during the test, confirming the cap is genuinely exercised rather than the assertion passing vacuously.

`scripts/test.sh`: 438 passed across `graph_buffer configlink simhash semantic parallel pipeline`. The full suite matches the pre-change baseline exactly — the two failures (`cli`, `py_lsp_scale`) reproduce identically on an unpatched tree.

`scripts/lint.sh`: no new findings. `pass_configlink.c` reports the same 10 pre-existing items as before the change.

## Risks

This changes *which* `CONFIGURES` edges are produced — 416 → 465 on the test corpus — because a canonical cut necessarily differs from an arbitrary one. That is inherent to any deterministic fix here.

The alternative, removing the caps so nothing is truncated, is not a drop-in: `config_entries` and `code_entries` are stack arrays (`CBM_SZ_8K` × 264 bytes plus `CBM_SZ_4K` × 520 bytes, ~4.3 MB of frame already), so lifting the bound would need heap allocation and a look at the O(config × code) match loop. Happy to take that direction instead if you would prefer it.

Scope is limited to `pass_configlink.c`; no other pass, the store format, or the MCP surface is touched.

## Links

- Fixes #1468
- Same defect class, sibling pass: #998 (and PR #1014)
- Determinism property this restores: #921
- Canonical-content precedent on the node path: #923
- Related parallel-pipeline edge defect: #1085

## Checklist

- [x] Tests pass locally
- [x] Lint passes
- [x] New behavior is covered by a test (reproduce-first: the test is red on an unpatched tree)

### DIFF
--- a/src/pipeline/pass_configlink.c
+++ b/src/pipeline/pass_configlink.c
@@ -60,6 +60,67 @@ static bool is_dep_section(const char *s) {
 
 /* ── Strategy 1: Config Key → Code Symbol ───────────────────────── */
 
+/* Canonical candidate order (determinism). Both collectors below fill a
+ * fixed-capacity array and stop at max_out; the label indexes they walk are
+ * gbuf insertion order = parallel-extraction merge order, which varies run to
+ * run. On a repo with more candidates than the cap, sorting by a pure content
+ * key first is what keeps the surviving set — and therefore the emitted
+ * CONFIGURES edges — a function of the inputs rather than of worker
+ * scheduling. Tie-breaks stay content-only: node ids are handed out in merge
+ * order, so an id tie-break belongs in no canonical comparator. */
+enum {
+    CANON_CMP_LESS = -1,   /* qsort: left sorts before right */
+    CANON_CMP_GREATER = 1, /* qsort: left sorts after right */
+    CANON_CAP_BUF = 32     /* decimal rendering of a cap value */
+};
+
+static int cmp_node_ptr_canonical(const void *pa, const void *pb) {
+    const cbm_gbuf_node_t *a = *(const cbm_gbuf_node_t *const *)pa;
+    const cbm_gbuf_node_t *b = *(const cbm_gbuf_node_t *const *)pb;
+    const char *qa = a->qualified_name ? a->qualified_name : "";
+    const char *qb = b->qualified_name ? b->qualified_name : "";
+    int r = strcmp(qa, qb);
+    if (r != 0) {
+        return r;
+    }
+    const char *fa = a->file_path ? a->file_path : "";
+    const char *fb = b->file_path ? b->file_path : "";
+    r = strcmp(fa, fb);
+    if (r != 0) {
+        return r;
+    }
+    if (a->start_line != b->start_line) {
+        return a->start_line < b->start_line ? CANON_CMP_LESS : CANON_CMP_GREATER;
+    }
+    const char *na = a->name ? a->name : "";
+    const char *nb = b->name ? b->name : "";
+    return strcmp(na, nb);
+}
+
+/* A filled-to-capacity collector dropped candidates; say so rather than
+ * truncating silently. */
+static void log_candidate_truncation(const char *side, int cap) {
+    char cap_buf[CANON_CAP_BUF];
+    snprintf(cap_buf, sizeof(cap_buf), "%d", cap);
+    cbm_log_info("configlinker.truncated", "side", side, "cap", cap_buf);
+}
+
+/* Heap copy of `nodes` sorted by cmp_node_ptr_canonical. Returns NULL (and
+ * leaves the caller on the unsorted borrowed array) only on allocation
+ * failure, which degrades determinism but never correctness. */
+static const cbm_gbuf_node_t **canonical_node_copy(const cbm_gbuf_node_t *const *nodes, int count) {
+    if (!nodes || count <= 0) {
+        return NULL;
+    }
+    const cbm_gbuf_node_t **sorted = malloc((size_t)count * sizeof(*sorted));
+    if (!sorted) {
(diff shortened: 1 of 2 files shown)


DeusData:
Merged — thank you @Studnicky. Sorting by a pure content key before applying the cap is the right fix, and the reasoning in your comment is the part worth keeping: **node ids are handed out in merge order, so an id tie-break belongs in no canonical comparator.** That is the subtle trap here — an id tie-break looks deterministic and reintroduces exactly the nondeterminism it appears to remove.

Together with #1471 this closes two concrete contributors to the multi-threaded edge-count jitter tracked since the 0.9.1 RC.

PR #1471, opened 2026-08-06
fix(graph_buffer): merge duplicate-edge properties commutatively

### LINKED ISSUE #1469: graph_buffer: duplicate-edge property merge is last-write-wins, so edge attributes depend on worker merge order
On a duplicate edge key, `cbm_gbuf_insert_edge` overwrites the stored property blob unconditionally. Because per-worker edge buffers merge in worker-slot order, which strategy's attributes survive is a function of thread scheduling. The edge *set* stays stable; its *attributes* flicker run to run.

## Cause

The dedup key excludes the attributes it needs to arbitrate:

```c
snprintf(buf, bufsz, "%lld:%lld:%s", (long long)src, (long long)tgt, type);
```

`confidence`, `strategy`, `via` and `candidates` live only in `properties_json`. On a key hit:

```c
cbm_gbuf_edge_t *existing = cbm_ht_get(gb->edge_by_key, key);
if (existing) {
    /* Merge properties (just replace for now) */
    if (properties_json && strcmp(properties_json, "{}") != 0) {
        free(existing->properties_json);
        existing->properties_json = heap_strdup(properties_json);
    }
    return existing->id;
}
```

Duplicates carrying *different* attributes genuinely occur: `pass_calls` mints `CALLS` edges from both LSP resolution and registry-textual matching, with different confidence and strategy. `pass_parallel.c` states the intended precedence — "LSP-resolved calls take precedence over registry-textual matching" — but last-write-wins only honours it when the LSP edge happens to arrive second.

Since file→worker assignment comes from an atomic work-stealing queue and worker buffers merge by slot index rather than by content, arrival order is scheduler-dependent.

## Impact

An `mcp__*` consumer reading `CALLS` confidence can see a different value for the same commit on two indexes of the same tree. It also means the stated LSP-over-textual precedence is not actually enforced.

This is invisible to `tests/repro/repro_parallel_determinism.c`, which compares `(source_qn, type, target_qn)` triples and so cannot observe attribute churn.

## Suggested fix

Replace the overwrite with a total order over the two candidates, making the merge commutative and associative and therefore independent of arrival order — e.g. higher `confidence` wins, with a deterministic tie-break between equals. That also makes the documented LSP-over-textual precedence hold regardless of which lands first.

The node path already works this way: `cbm_gbuf_upsert_node` picks the survivor of a same-QN collision by a canonical content rule "so the pick is commutative and scheduling-free" (#923). This is the same idea applied to edge attributes.

Environment: v0.9.0 (`9b9b6ff`), darwin-arm64, built from source.

Related: #923, #998.

Studnicky:
Fixes #1469

## What does this PR do?

On a duplicate edge key, `cbm_gbuf_insert_edge` replaced the stored property blob unconditionally ("just replace for now"). `confidence`, `strategy` and `via` are not part of the dedup key, and per-worker edge buffers merge in worker-slot order, so which strategy's attributes survived was a function of thread scheduling.

This replaces the overwrite with a total order over the two candidates, making the merge commutative and associative and therefore independent of arrival order:

1. higher `confidence` wins, which also enforces the intended precedence that an LSP-resolved call outranks a textual match;
2. on equal confidence, the lexicographically greater blob wins, giving a deterministic choice between otherwise equal candidates.

An empty or absent incoming blob still never displaces stored properties.

## Why

Duplicates carrying different attributes genuinely occur — `pass_calls` mints `CALLS` edges from both LSP resolution and registry-textual matching. The edge *set* was stable while its *attributes* flickered run to run, so an MCP consumer reading `CALLS` confidence could see a different value for the same commit across two indexes of the same tree.

It also means the precedence `pass_parallel.c` documents — "LSP-resolved calls take precedence over registry-textual matching" — was only honoured when the LSP edge happened to arrive second. Under this rule it holds regardless of arrival.

The node path already works this way: `cbm_gbuf_upsert_node` picks the survivor of a same-QN collision by a canonical content rule "so the pick is commutative and scheduling-free" (#923). This applies the same idea to edge attributes.

## Testing

Three new tests in `test_graph_buffer.c`:

- `gbuf_edge_props_merge_is_order_independent` — the same pair of blobs inserted in both orders must leave identical stored properties.
- `gbuf_edge_props_merge_prefers_higher_confidence` — the higher-confidence blob wins even when it arrives first.
- `gbuf_edge_props_merge_keeps_existing_on_empty` — guards the `{}` case.

The first two fail on an unpatched tree (51 passed, 2 failed), with the low-confidence registry blob having overwritten the high-confidence LSP one. All three pass with the change.

`scripts/test.sh`: 438 passed across `graph_buffer configlink simhash semantic parallel pipeline`. The full suite matches the pre-change baseline exactly — the two failures (`cli`, `py_lsp_scale`) reproduce identically on an unpatched tree.

`scripts/lint.sh`: no new findings. `graph_buffer.c` reports the same 8 pre-existing items as before the change.

Note this is invisible to `tests/repro/repro_parallel_determinism.c`, which compares `(source_qn, type, target_qn)` triples and so cannot observe attribute churn — hence unit coverage rather than a corpus repro.

## Risks

Edge *identity* and edge *count* are unchanged; only which blob survives on a duplicate key differs. Where duplicates disagreed, the stored attributes may now differ from a given previous run — but previously that value was scheduling-dependent, so there was no stable prior value to preserve.

`edge_props_confidence` parses `"confidence":` with `strtod`; a blob without one reads as `-1.0`, so any edge stating a confidence outranks one that does not. Blobs that carry no confidence at all fall through to the lexicographic tie-break, which is deterministic but arbitrary — deliberately, since there is no semantic basis to prefer either.

Scope is limited to the duplicate-key path in `cbm_gbuf_insert_edge`; the key format, the store schema, and the MCP surface are untouched.

## Links

- Fixes #1469
- Canonical-content precedent on the node path: #923
- Related determinism work: #998, #921

## Checklist

- [x] Tests pass locally
- [x] Lint passes
- [x] New behavior is covered by a test (reproduce-first: two of the three are red on an unpatched tree)

### DIFF
--- a/src/graph_buffer/graph_buffer.c
+++ b/src/graph_buffer/graph_buffer.c
@@ -22,6 +22,10 @@ enum {
     GB_MIN_FOR_DEDUP = 2,    /* need at least 2 vectors to sort+dedup */
     GB_DEDUP_LOOKAHEAD = 1,  /* compare current with next element */
 };
+
+/* Sentinel for an edge property blob carrying no parseable "confidence": any
+ * blob that states one outranks a blob that does not. */
+#define CBM_EDGE_CONF_ABSENT (-1.0)
 #include "graph_buffer/graph_buffer.h"
 #include <yyjson/yyjson.h> // url_path extraction must match json_extract semantics
 #include "store/store.h"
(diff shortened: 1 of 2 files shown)


DeusData:
Merged — thank you @Studnicky. The key insight is the one most fixes in this area miss: making the merge a **total order** (confidence, then lexicographic) rather than picking a "better" arrival makes it commutative and associative by construction, so worker scheduling cannot influence the outcome. And it doubles as the precedence rule we actually want — an LSP-resolved edge outranks a textual match. The both-orders test is exactly the right shape.

This feeds directly into a known issue we have had open since the 0.9.1 RC: node counts are byte-stable across runs but edge totals drift 0.01–1.3% on resolution-heavy languages. Non-key edge attributes being decided by merge order is one concrete source of that, now closed.

Issue #1427, opened 2026-08-03
Windows v0.9.1-rc.1: detached daemon spawns focus-stealing cmd.exe / Windows Terminal windows


0hirume:
## Summary

On Windows, `v0.9.1-rc.1` opens a burst of Windows Terminal windows running `cmd.exe` during normal MCP tool calls. Reproduced consistently with `list_projects`.

Each spawned terminal takes foreground focus, so the desktop is effectively unusable until the burst finishes. Stable `v0.9.0` does not reproduce this behavior.

## Environment

- OS: Windows
- MCP client: Codex
- codebase-memory-mcp: `v0.9.1-rc.1`
- Commit: `d90986b2185badc3f8ce4fa881fff3991b851a92`
- Windows Terminal configured as the default terminal application
- 9 indexed projects in the cache

## Reproduction

1. Install and configure `v0.9.1-rc.1` as a stdio MCP server on Windows.
2. Have multiple Git repositories indexed.
3. Invoke `list_projects`.
4. Observe repeated Windows Terminal windows/sessions running `cmd.exe`. Each activation steals foreground focus.

A single reproduced `list_projects` call took approximately 9.8 seconds and caused repeated focus theft throughout the call.

## Expected behavior

Git discovery subprocesses run entirely in the background. No terminal window is created or activated.

## Actual behavior

Multiple Windows Terminal / `cmd.exe` sessions are created and activated. The user repeatedly loses focus until the MCP call completes.

## Regression evidence

The stable `v0.9.0` release does not exhibit this behavior. The prerelease introduces the permanent detached daemon, which changes the console context from which Git subprocesses are launched.

## Likely root cause

The RC launches its permanent daemon detached and without a console:

- [`src/daemon/bootstrap.c:752`](https://github.com/DeusData/codebase-memory-mcp/blob/v0.9.1-rc.1/src/daemon/bootstrap.c#L752) uses `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW`.

Git capture later explicitly constructs a `cmd.exe /c ...` command:

- [`src/foundation/compat_fs.c:258-263`](https://github.com/DeusData/codebase-memory-mcp/blob/v0.9.1-rc.1/src/foundation/compat_fs.c#L258-L263)

But its child `CreateProcessW` call uses only `EXTENDED_STARTUPINFO_PRESENT`, omitting `CREATE_NO_WINDOW`:

[...]

PR #1485, opened 2026-08-07
fix(windows): suppress the console window on the last spawn site (#1427)

DeusData:
Follow-up to #1448 (thanks @mlandolfi90), closing out #1427.

## The missed site

#1448 set `CREATE_NO_WINDOW` on the subprocess spawn and on `cbm_popen_isolated`, but `cbm_exec_no_shell` (`compat_fs.c`) still passed a bare `0` for `dwCreationFlags`. That's the helper behind `git`, `codesign` and `open` — the most frequently hit path — so those console windows survived the merge. Spotted by @noctrex after verifying the merged fix on Windows 11, and credited on the commit.

## Why a contract test comes with it

The flag must be set at each call site independently, the sites don't resemble each other, and reviewing *"does this PR add the flag"* says nothing about the sites a PR doesn't touch. That's exactly how the third one survived a fix applied twice:

| site | state |
|---|---|
| `daemon/bootstrap.c` | had it from the start (`DETACHED_PROCESS \| … \| CREATE_NO_WINDOW`) |
| `subprocess.c` | added by #1448 (flags variable) |
| `compat_fs.c:313` | added by #1448 (inline literal) |
| `compat_fs.c:689` | **missed by #1448** (bare `0`) |

`tests/test_spawn_no_window_contract.sh` asserts the whole-tree property instead, so a fifth spawn site can't be added without the flag. It's registered as Step 0k in `scripts/test.sh`, which runs the contracts first on every leg.

## One thing worth flagging in review

The contract **strips C comments before matching**, and that's load-bearing rather than tidiness. The first version of it **passed with the fix reverted** — my explanatory comment above the fixed call contains the string `CREATE_NO_WINDOW`, which satisfied the match. A contract that can be satisfied by prose *about* the contract is a false guard, and this repo has been bitten by string-matching contracts before.

After the fix, verified in both directions:

- **green** with the fix — `OK: all 4 CreateProcessW site(s) set CREATE_NO_WINDOW`
- **red** with the fix reverted *and the comment still in place* — `FAIL: … src/foundation/compat_fs.c:696 (cbm_exec_no_shell) spawns without CREATE_NO_WINDOW`, exit 1

It also fails if it ever finds zero call sites, so it can't quietly stop checking anything if the spawn layer moves.

## Verification

`make lint-ci` clean; contract green both directions; exec-bit contract still passes (invoked via `bash`, and the file is committed `100755` to match its siblings).

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -230,6 +230,9 @@ bash "$ROOT/tests/test_parallel_harness_contract.sh"
 echo "=== Step 0j: venue parity contract (one harness, every venue) ==="
 bash "$ROOT/tests/test_venue_parity_contract.sh"
 
+echo "=== Step 0k: spawn console-window contract (#1427) ==="
+bash "$ROOT/tests/test_spawn_no_window_contract.sh"
+
 # Verify compiler supports target arch
 verify_compiler "$CC"
 
--- a/src/foundation/compat_fs.c
+++ b/src/foundation/compat_fs.c
@@ -686,7 +686,14 @@ int cbm_exec_no_shell(const char *const *argv) {
     memset(&pi, 0, sizeof(pi));
     si.cb = sizeof(si);
 
-    if (!CreateProcessW(NULL, cmdline, NULL, NULL, FALSE, 0, NULL, NULL, &si, &pi)) {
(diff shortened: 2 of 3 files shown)


noctrex:
You da man

PR #1512, opened 2026-08-10
fix(mcp): validate search_code limit

### LINKED ISSUE #1511: search_code: negative limit is echoed back as the result count
`search_code` does not validate `limit`. A negative value passes through and is reported as the result count.

```
cli search_code --project <p> --pattern verify --limit -5
→ results: -5  (cols: qn label file lines matches in out)
```

`--limit 0` behaves the same way (`results: 0`). Both then fall through to the `raw:` path, so the call still returns something, which is why it is easy to miss.

Sibling tools do not do this. `search_graph --limit -5` on the same project returns `results: 50` — clamped, not echoed.

Two places differ from the rest of the surface:

- `handle_search_code` (`src/mcp/mcp.c:9579`) reads `limit` with `cbm_mcp_get_int_arg(args, "limit", MCP_DEFAULT_LIMIT)` and never range checks it, while the same function validates `pattern`.
- The declared schema for `search_code`'s `limit` (`src/mcp/mcp.c:603`) has no `minimum`/`maximum`. Other tools declare them, e.g. `"limit":{"type":"integer","default":100,"minimum":1,"maximum":5000}` at line 516, so a schema-validating MCP client would reject `-5` before it reached the handler while the CLI path accepts it.

Adding `"minimum":1` to the schema and clamping in the handler would make it consistent with the sibling tools.

Built from `main` @ 10cb0e0, Linux x86_64, `make -f Makefile.cbm cbm`.

lukiod:
Closes #1511.

`search_code` never range checked `limit`. `--limit -5` was echoed back as the result count (`results: -5`) and then fell through to the raw path, so the call still returned something.

- clamp `limit < 1` to `MCP_DEFAULT_LIMIT` in `handle_search_code`
- declare `"minimum": 1` in the tool schema, so a schema validating client rejects it before the handler

Both already hold for sibling tools. `search_graph --limit -5` on the same project returns a clamped `results: 50`.

## Test plan

Two tests in `tests/test_mcp.c`:

- `tool_search_code_negative_limit_is_not_echoed` — asserts `results: -5` does not appear in the response
- `tool_search_code_limit_declares_a_minimum` — walks `tools/list` with yyjson to `search_code.inputSchema.properties.limit.minimum`, matching how `mcp_response_has_exact_tool` inspects responses

`./build/c/test-runner mcp` under ASan/UBSan on Linux x86_64: both new tests pass, 94 passing overall.

## Correction, and a pre-existing failure

An earlier version of this description said "0 fail". That was wrong, and I would rather flag it than leave it: the suite exits 1 on `main` here, independently of this change.

```
FAIL tests/test_mcp.c:6803: strstr(dc_resp, "seed_symbols: 1\n") is NULL
```

That is `detect_changes_seeds_only_touched_symbol_issue1363`. The `3417 bytes leaked in 3 allocations` ASan also reports is a consequence rather than a separate bug: the assertion returns before the test's `free(dc_resp)`, `free(project)` and `cbm_mcp_server_free(srv)`, which is exactly three allocations.

This PR touches only `handle_search_code` and the `search_code` schema string, so it cannot reach `detect_changes`. Happy to open a separate issue with the full trace if it is not already known.

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -604,7 +604,7 @@ static const tool_def_t TOOLS[] = {
      "\"description\":\"Max enriched results per call. Default 10. Response includes "
      "'total_grep_matches' and 'total_results' so callers can detect truncation. No "
      "offset parameter — raise limit or narrow with file_pattern / path_filter to see more."
-     "\",\"default\":10}},\"required\":[\"pattern\",\"project\"]}"},
+     "\",\"default\":10,\"minimum\":1}},\"required\":[\"pattern\",\"project\"]}"},
 
     {"list_projects", "List projects", "List all indexed projects",
      "{\"type\":\"object\",\"properties\":{}}"},
@@ -9577,6 +9577,9 @@ static char *handle_search_code(cbm_mcp_server_t *srv, const char *args) {
     char *path_filter = cbm_mcp_get_string_arg(args, "path_filter");
     char *mode_str = cbm_mcp_get_string_arg(args, "mode");
     int limit = cbm_mcp_get_int_arg(args, "limit", MCP_DEFAULT_LIMIT);
+    if (limit < 1) {
+        limit = MCP_DEFAULT_LIMIT;
+    }
     int context_lines = cbm_mcp_get_int_arg(args, "context", 0);
     bool use_regex = cbm_mcp_get_bool_arg(args, "regex");
     uint64_t search_t0 = cbm_now_ms();
--- a/tests/test_mcp.c
+++ b/tests/test_mcp.c
@@ -3973,6 +3973,59 @@ TEST(tool_search_code_missing_pattern) {
     PASS();
 }
 
+TEST(tool_search_code_negative_limit_is_not_echoed) {
+    cbm_mcp_server_t *srv = cbm_mcp_server_new(NULL);
+
+    char *resp =
+        cbm_mcp_server_handle(srv, "{\"jsonrpc\":\"2.0\",\"id\":35,\"method\":\"tools/call\","
+                                   "\"params\":{\"name\":\"search_code\","
+                                   "\"arguments\":{\"pattern\":\"func main\","
+                                   "\"project\":\"nonexistent\",\"limit\":-5}}}");
+    ASSERT_NOT_NULL(resp);
+    ASSERT_NULL(strstr(resp, "results: -5"));
+    free(resp);
+
+    cbm_mcp_server_free(srv);
+    PASS();
+}
+
+TEST(tool_search_code_limit_declares_a_minimum) {
+    cbm_mcp_server_t *srv = cbm_mcp_server_new(NULL);
+
+    char *resp = cbm_mcp_server_handle(
+        srv, "{\"jsonrpc\":\"2.0\",\"id\":36,\"method\":\"tools/list\",\"params\":{}}");
+    ASSERT_NOT_NULL(resp);
+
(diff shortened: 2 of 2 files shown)


lukiod:
CI is red on this PR but not from these changes. Recording what the logs show, in case it is useful.

The `mcp` suite, which is where both new tests live, passes on every platform:

```
mcp rc=0 pass=194 fail=0 skip=2
```

The failing shard is `cli`, and it is the same test on ubuntu-latest, ubuntu-24.04-arm and macos-14:

```
cli rc=1 pass=274 fail=1
  cli_concurrent_ui_then_standard_install_leaves_coherent_standard_set  FAIL tests/test_cli.c:1829
```

Same test on three platforms rather than one, so it reads as a real failure rather than a flake.

One line in that log is misleading and worth flagging: `curl: (37) ... error: download failed (exit 37)` appears just after, but it belongs to `cli_update_download_failure_does_not_quiesce_sessions`, which simulates a failed download on purpose. It is expected output, not the cause.

`test-windows-guards` is also red. The eight guards cover non-ASCII paths, daemon lifecycle and stability, hook augment, drive listing and update handoff; none exercise `search_code`, and the only one that touches `tools/list` (`mcp_stdio.py`) reads `result.tools` without validating schemas.

Happy to open a separate issue for the `cli` failure if it is not already tracked.

DeusData:
Thank you for the narrow validation fix and for isolating the unrelated CI failure. #1513 already tracks that lease-acquisition failure, so no additional issue is needed for the same run. I have routed this PR to the immediate bug train with normal priority; detailed review may take a little time while the queue is full.

DeusData:
Reviewed — the change is correct and the tests are exactly right (schema `minimum` **and** a runtime clamp, so a client that ignores the schema still can't poison the count). Thank you @lukiod.

The red CI here is **not your change**: it's `cli_concurrent_ui_then_standard_install_leaves_coherent_standard_set`, a test that no longer exists on main (it went away with the single-composition consolidation). Your branch predates that, so the stale base is what's failing. A rebase on current main should turn it green and we'll merge immediately — or say the word and we'll rebase it for you.

DeusData:
Thank you @lukiod — this was a correct fix, and it shipped.

Your branch's CI had gone red purely from age (the base predated a day of merges), so rather than leave a good change waiting on a rebase, we distilled it: the schema now declares `minimum: 1` and `handle_search_code` clamps a sub-1 limit, exactly as you had it, in commit `4b886904` on the v0.10.3 branch. Both of your tests came across too — the negative-limit response check and the tools/list schema assertion.

The reasoning in your PR is worth repeating for the next reader, and it is now a comment in the code: a schema is a request to the client, never a guarantee to the server, so the handler clamps even though the schema declares the bound. That is why both halves were needed.

Closes #1511 as well. Shipping in **v0.10.3**.

lukiod:
@DeusData  sry about this  lol i though  i might have rebased it mb   just updated my main o uhm hope further prs will be on latest version

DeusData:
No apology needed at all, @lukiod — the base going stale is just what happens when a day has fourteen merges in it, and it cost you nothing: the fix shipped in v0.10.3, your tests came with it, and the reasoning from your PR is now a comment in the code ("a schema is a request to the client, never a guarantee to the server").

You have two contributions in the last week (this and #1521), both correct and both with tests. That is a better hit rate than most. Please do keep them coming — and if a base ever goes stale again, say the word and we will rebase it for you rather than have you redo work.

PR #1514, opened 2026-08-10
fix(installer): accept legacy release archive layout

### LINKED ISSUE #1510: install.sh member-set check rejects official v0.9.0 release assets (both standard and ui variants)
## Summary

`install.sh` on `main` fails on the official `v0.9.0` release assets for **both** the `--standard` and `--ui` variants, with:

```
error: release archive does not match the exact standard member set
```
and
```
error: release archive does not match the exact ui member set
```

Checksum verification passes in both cases — the archive itself is intact and matches `checksums.txt`. The failure is `install.sh`'s post-extraction member-count/name check, not a corrupted download.

## Repro

```bash
curl -fsSL https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh | bash -s -- --standard
curl -fsSL https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh | bash -s -- --ui
```

Both fail after "Checksum verified." on Linux/amd64.

## Root cause (as far as I can tell)

`install.sh` (`main`) expects:

- standard: **5** members
- ui: **6** members, including `cbm-integrations.json` and a `cbm-ui-<sha256>.pack` file

I downloaded and inspected the actual `v0.9.0` assets directly (bypassing the installer):

`codebase-memory-mcp-linux-amd64-portable.tar.gz` (standard) contains only **4** members:
```
codebase-memory-mcp
LICENSE
install.sh
THIRD_PARTY_NOTICES.md
```
— missing `cbm-integrations.json`.

`codebase-memory-mcp-ui-linux-amd64-portable.tar.gz` (ui) also contains only **4** members — identical list, no `cbm-integrations.json` and no `cbm-ui-*.pack` at all. So the currently published `ui` asset doesn't actually differ from `standard`.

This looks like `install.sh` on `main` was updated for a newer release-asset layout (adding `cbm-integrations.json` and a content-addressed UI pack file) before a matching release was cut. The `v0.9.0` assets (published 2026-07-08) still use the older 4-member layout.

## Impact

The one-line installer in the README currently cannot install either variant on Linux/amd64 — it always exits 1 after checksum verification.

## Suggested fix

Either:
1. Cut a new release whose assets match the member set `install.sh` on `main` expects, or
2. Pin `install.sh`'s member-set check to what's actually in the latest published release until the new asset layout ships.

Happy to provide the full `tar tvzf` output for other platform archives if useful.

cyphercodes:
## What does this PR do?

Fixes #1510.

`install.sh` now accepts the exact legacy four-member release archive layout used by the published v0.9.0/v0.9.x assets, while still keeping the canonical current runtime-asset layouts closed:

- standard: legacy 4-member archive or canonical 5-member archive
- UI: legacy 4-member archive or canonical 6-member archive with one content-addressed UI pack

The installer continues to reject unexpected archive members, duplicate/missing required core members, malformed UI pack names, and UI pack digest mismatches. The smoke fixture contract now covers the legacy standard/UI layouts and canonical UI layout through a loopback HTTP installer fixture.

## Checklist

Local verification run:

- `bash -n install.sh`
- `bash -n tests/test_smoke_fixture_contract.sh`
- `bash tests/test_smoke_fixture_contract.sh`
- `CBM_DOWNLOAD_URL=https://github.com/DeusData/codebase-memory-mcp/releases/download/v0.9.0 bash ./install.sh --standard --skip-config` with a temporary `HOME`, then installed binary `--version`
- `CBM_DOWNLOAD_URL=https://github.com/DeusData/codebase-memory-mcp/releases/download/v0.9.0 bash ./install.sh --ui --skip-config` with a temporary `HOME`, then installed binary `--version`
- `git diff --check HEAD~1..HEAD`
- `scripts/check-dco.sh HEAD~1..HEAD`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -220,6 +220,55 @@ if [ "$EXPECTED" != "$ACTUAL" ]; then
 fi
 echo "Checksum verified."
 
+# Validate the complete archive namespace before extraction. Current and legacy
+# release assets use the same four-member root layout; anything outside that
+# closed set is a release-integrity failure, not a sidecar to ignore.
+if [ "$OS" = "windows" ]; then
+    ARCHIVE_BINARY="codebase-memory-mcp.exe"
+    ARCHIVE_INSTALLER="install.ps1"
+else
+    ARCHIVE_BINARY="codebase-memory-mcp"
+    ARCHIVE_INSTALLER="install.sh"
+fi
+ARCHIVE_MEMBERS_FILE="$DLDIR/archive-members.txt"
+if [ "$EXT" = "zip" ]; then
+    if ! unzip -Z1 "$DLDIR/$ARCHIVE" > "$ARCHIVE_MEMBERS_FILE"; then
+        echo "error: could not enumerate release archive" >&2
+        exit 1
+    fi
+else
+    if ! tar -tzf "$DLDIR/$ARCHIVE" > "$ARCHIVE_MEMBERS_FILE"; then
+        echo "error: could not enumerate release archive" >&2
+        exit 1
+    fi
+fi
+
+BINARY_MEMBERS=0
+LICENSE_MEMBERS=0
+INSTALLER_MEMBERS=0
+NOTICE_MEMBERS=0
+ARCHIVE_MEMBER_COUNT=0
+while IFS= read -r member || [ -n "$member" ]; do
+    ARCHIVE_MEMBER_COUNT=$((ARCHIVE_MEMBER_COUNT + 1))
+    case "$member" in
+        "$ARCHIVE_BINARY") BINARY_MEMBERS=$((BINARY_MEMBERS + 1)) ;;
+        LICENSE) LICENSE_MEMBERS=$((LICENSE_MEMBERS + 1)) ;;
+        "$ARCHIVE_INSTALLER") INSTALLER_MEMBERS=$((INSTALLER_MEMBERS + 1)) ;;
+        THIRD_PARTY_NOTICES.md) NOTICE_MEMBERS=$((NOTICE_MEMBERS + 1)) ;;
+        *)
+            echo "error: release archive contains unexpected member: $member" >&2
+            exit 1
+            ;;
+    esac
+done < "$ARCHIVE_MEMBERS_FILE"
+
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for the installer compatibility patch and the clear link to the affected reports. I have routed it as a high-priority installer bug for the immediate train. The branch currently conflicts with `main`, and the `v0.10.0` installer and release layout changed after this PR opened, so the eventual review will need to evaluate the patch against that current contract. The queue is full, but we will come back with detailed feedback.

DeusData:
Reviewed — thank you @cyphercodes. Same note as #1512: the failing check is `cli_concurrent_ui_then_standard_install_leaves_coherent_standard_set`, a test that no longer exists on current main (removed with the single-composition consolidation), so this red is a stale base rather than your change. Please rebase on main and we'll merge — happy to do the rebase for you if you'd rather.

Related: we're also shipping `ui-*` alias assets on releases (#1538) so already-shipped 0.9.x updaters stop 404ing; your legacy-layout acceptance and that alias publishing solve the two halves of the same upgrade problem.

cyphercodes:
Rebased on current `main` and resolved the installer/smoke-contract conflicts around the single-composition release layout. Kept the Unix installer accepting the exact four-member legacy/current archive shape while retaining the current `tar --no-same-owner` extraction behavior.

Verification run locally:
- `bash -n install.sh tests/test_smoke_fixture_contract.sh`
- `bash tests/test_smoke_fixture_contract.sh`
- `git diff --check upstream/main...HEAD`

DeusData:
Merged — thank you @cyphercodes. Verified against the actual shipped v0.10.2 archives before merging: they contain exactly those four members, so the closed-set validation matches reality on every platform.

Two things worth recording for whoever touches this next:

- The **symlink rejection** is the quietly important part. An archive member that extracts as a symlink is a classic path-escape vector, and `[ -f ]` alone accepts a symlink pointing at a regular file — the explicit `[ -L "$DLBIN" ]` check closes that.
- The closed set is deliberately strict, which means **adding a fifth member to release archives will now break installs until `install.sh` is updated in the same change**. That is the right trade (an unexpected member in a signed-checksum artifact is a release-integrity problem, not a sidecar), and the release smoke tests install from a built archive so a divergence surfaces before shipping rather than after.

Shipping in **v0.10.3**.

PR #1521, opened 2026-08-10
test: cover wrong JSON types in the MCP robustness layer

lukiod:
The robustness layer varies argument *values* thoroughly (oversized, injection, negative, ReDoS) but always sends the expected *type*. These nine send a value of the wrong JSON type for the envelope fields parsed before tool dispatch: `name`, `arguments`, `params`, `method`.

All nine pass against the current binary. This is a regression guard for a class the layer does not currently reach, not a bug report — I went looking for a NULL deref and did not find one; `cbm_mcp_parse_request` rejects a non-string `method` before use, and `heap_strdup` returns NULL safely.

Kept to the envelope. I also wrote per-tool field cases (wrong-typed `name_pattern`, `query`, `limit`) and dropped them: they stop at project validation like the existing adversarial cases, so they added runtime without reaching new code.

Layer goes 23/23 in 88s to 32/32 in 120s on this machine.

### DIFF
--- a/scripts/security-fuzz.sh
+++ b/scripts/security-fuzz.sh
@@ -116,6 +116,19 @@ test_payload "array instead of object" '[1,2,3]'
 test_payload "deeply nested json" '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"search_graph","arguments":{"name_pattern":{"a":{"b":{"c":{"d":{"e":{"f":"deep"}}}}}}}}}'
 
 echo ""
+echo "--- Wrong JSON types ---"
+
+test_payload "null tool name" '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":null,"arguments":{}}}'
+test_payload "numeric tool name" '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":42,"arguments":{}}}'
+test_payload "object tool name" '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":{"a":1},"arguments":{}}}'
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you for the focused wrong-type robustness coverage. I have routed this as security-hardening test work for the immediate train with normal priority. The community review queue is currently full, so a complete maintainer review may take a little time, but the contribution is recorded and queued.

DeusData:
Merged. Nine wrong-JSON-type payloads against the MCP envelope is exactly the kind of coverage that pays for itself quietly — the robustness layer is the one place where a malformed input should never reach anything else, and asserting that is cheaper than discovering it from a report.

Also worth closing a loop with you: your earlier #1511 fix could not land as-is, but the bug was real and we implemented it ourselves — `src/mcp/mcp.c:9637` carries the comment citing #1511, and it shipped in v0.10.3. The issue is closed. Apologies that the credit was not clearer at the time; a negative `limit` echoed back as a result count reads to an agent as an answer rather than a rejected argument, which is precisely why it mattered.

lukiod:
> Merged. Nine wrong-JSON-type payloads against the MCP envelope is exactly the kind of coverage that pays for itself quietly — the robustness layer is the one place where a malformed input should never reach anything else, and asserting that is cheaper than discovering it from a report.
> 
> Also worth closing a loop with you: your earlier #1511 fix could not land as-is, but the bug was real and we implemented it ourselves — `src/mcp/mcp.c:9637` carries the comment citing #1511, and it shipped in v0.10.3. The issue is closed. Apologies that the credit was not clearer at the time; a negative `limit` echoed back as a result count reads to an agent as an answer rather than a rejected argument, which is precisely why it mattered.

no worries  fix is what matters at the end so as we can use it for ourselves .

PR #1531, opened 2026-08-11
fix(daemon): make private directory DACL inheritable

### LINKED ISSUE #1351: Windows x64: v0.9.1-rc.1 daemon leaves existing cache files unreadable (empty ACL)
### Version

codebase-memory-mcp 0.9.1-rc.1

### Platform

Windows (x64)

### Install channel

### Binary variant

standard

When I tested v0.9.1-rc.1 against an existing local CBM installation and cache on Windows, the new daemon/ACL path left existing cache files with an empty ACL. The same Windows user could no longer read _config.db, log files, or sampled project database files.

After that, MCP requests such as search_graph and check_index_coverage did not complete within 15–20 seconds. Starting the RC daemon again reproduced the empty ACL on _config.db.

I expected the installer/daemon either to preserve usable file permissions or to fail safely before changing existing permissions. It must not leave the current user's cache unreadable.

### Reproduction

Prerequisite: Windows x64 with an existing CBM installation/cache. The existing installation or one of its parent directories has inherited cross-account ACL entries.

1. Back up the existing CBM executable, configuration, and cache.
2. Verify the official v0.9.1-rc.1 Windows amd64 release archive and run install.ps1 with --skip-config against the existing installation.
3. The installer rejects the existing installation directory with: acl-grants-cross-account-mutation
4. Retry with an owner-private temporary directory. Activation then aborts with: active CBM sessions and operations could not be stopped safely; no activation was committed.
5. Place the checksum-verified RC binary in a user-private directory without changing agent configuration, then start the daemon: codebase-memory-mcp.exe daemon start
6. Observe: daemon.runtime_config_open_failed reason=config_db_unavailable
7. Inspect the existing CBM cache: _config.db, logs, and sampled project .db files have an empty ACL and are unreadable by the current user.
8. Start the RC daemon again; the empty ACL on _config.db is reproduced.
9. On an already-indexed 31-node project, check_index_coverage did not complete within 15 seconds through the CLI. Small MCP search_graph and check_index_coverage calls did not complete within 15–20 seconds.

No repository source code is required for this reproduction; the failure occurs before source analysis.

### Logs

```text
activation transaction I/O failed: acl-grants-cross-account-mutation
active CBM sessions and operations could not be stopped safely; no activation was committed.
daemon.runtime_config_open_failed reason=config_db_unavailable
```

```text
CBM_DIAGNOSTICS=1 was enabled. The daemon created a local trajectory.ndjson file. I have not attached it because it may contain local environment details; I can provide a redacted extract if needed.
```

31 nodes / small existing project / failure occurs before source analysis

### Confirmations

Kiborgik:
Fixes #1351

## Problem

`win_security_init` builds the owner-only ACL with `AddAccessAllowedAce` (`src/daemon/ipc.c:3664`). That API has no `AceFlags` parameter, so the ACE it produces always has flags set to zero. Applied to a directory together with `PROTECTED_DACL_SECURITY_INFORMATION`, the result is:

```
D:PAI(A;;FA;;;<user>)
```

The protection severs the inherited ACEs and the new ACE propagates nothing, so every child created afterwards is born with an empty DACL and is unreadable even by its owner. That is the `RuleCount=0` state on `_config.db`, `logs`, and project databases reported in #1351.

`win_runtime_directory_secure` is also a repair path: line 4051 continues on `ERROR_ALREADY_EXISTS` and line 4093 re-stamps the existing directory, so a tree repaired by hand with `icacls` is destroyed again on the next daemon or client start.

The three other owner-only ACL builders in the tree avoid this by using `SetEntriesInAclW` with an explicit `grfInheritance` (`compat.c:88`, `compat_fs.c:454`, and `activation_transaction.c:368`, the last correctly using `NO_INHERITANCE` because it is applied only to a `CREATE_NEW` file).

## Change

Build a second ACL for containers using `AddAccessAllowedAceEx` with `CONTAINER_INHERIT_ACE | OBJECT_INHERIT_ACE`, and use it at the three directory sites (`ipc.c:4050`, `4097`, `4184`). Files and kernel objects are leaves and keep the existing flagless ACL (`3819`, `3823`, `4236`, `4260`).

The container ACE carries `FILE_ALL_ACCESS` rather than `GENERIC_ALL`. Windows splits an inheritable generic-rights ACE into an effective mapped ACE plus an `INHERIT_ONLY` one holding the generic bits, and `private_win_owner_only_dacl` requires exactly one ACE (`private_file_lock.c:910`). Using generic rights here fails every lock directory with `secure CLI coordination could not be created (project-locks)`. I hit that while developing this patch, which is why the test asserts the ACE count.

## Verification

Built from this branch and compared against the stock 0.10.0 release binary on the same cache path (Windows 11 26200, MinGW gcc, non-elevated standard token):

| Path | stock 0.10.0 | this branch |
|---|---|---|
| create, fresh directory | `<user>:(F)` | `<user>:(OI)(CI)(F)` |
| repair, existing directory | strips to `(F)` | `(OI)(CI)(F)` |
| `cli list_projects` | works | works |

## Test

`daemon_ipc_windows_private_directory_ace_is_inheritable` asserts that the secured directory has exactly one ACE, that it carries both inherit flags, that a child created inside it has a non-empty DACL, and that `cbm_private_lock_directory_adopt_windows` still accepts the directory. The inherit-flag and child-DACL assertions fail on current `main`.

## What I could not verify locally

MinGW ships no `libsanitizer`, so `scripts/test.sh` could not run in its ASan and UBSan configuration; my run was unsanitized. Several pre-existing `-Werror` diagnostics in `tests/test_daemon_ipc.c` and `tests/test_daemon_application.c` (`stringop-overflow`, `free-nonheap-object`) also had to be downgraded before the runner would build under this gcc. None are in code this patch touches, but CI is the authority on sanitizer and warning cleanliness here.

Separately, `scripts/run-tests-parallel.sh` stamps the build directory using a bare username, which misresolves on a machine whose hostname equals the username and locks the harness out of its own log directory. That is unrelated to this change and I will open it separately.

### DIFF
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -3312,6 +3312,7 @@ typedef BOOL(WINAPI *is_well_known_sid_fn)(PSID, WELL_KNOWN_SID_TYPE);
 typedef BOOL(WINAPI *is_valid_acl_fn)(PACL);
 typedef BOOL(WINAPI *initialize_acl_fn)(PACL, DWORD, DWORD);
 typedef BOOL(WINAPI *add_access_allowed_ace_fn)(PACL, DWORD, DWORD, PSID);
+typedef BOOL(WINAPI *add_access_allowed_ace_ex_fn)(PACL, DWORD, DWORD, DWORD, PSID);
 typedef BOOL(WINAPI *initialize_security_descriptor_fn)(PSECURITY_DESCRIPTOR, DWORD);
 typedef BOOL(WINAPI *set_security_descriptor_dacl_fn)(PSECURITY_DESCRIPTOR, BOOL, PACL, BOOL);
 typedef BOOL(WINAPI *set_security_descriptor_owner_fn)(PSECURITY_DESCRIPTOR, PSID, BOOL);
@@ -3339,6 +3340,7 @@ typedef struct {
     is_valid_acl_fn is_valid_acl;
     initialize_acl_fn initialize_acl;
     add_access_allowed_ace_fn add_access_allowed_ace;
+    add_access_allowed_ace_ex_fn add_access_allowed_ace_ex;
     initialize_security_descriptor_fn initialize_security_descriptor;
     set_security_descriptor_dacl_fn set_security_descriptor_dacl;
     set_security_descriptor_owner_fn set_security_descriptor_owner;
@@ -3352,6 +3354,9 @@ typedef struct {
     PACL acl;
     PSECURITY_DESCRIPTOR descriptor;
     SECURITY_ATTRIBUTES attributes;
+    PACL directory_acl;
+    PSECURITY_DESCRIPTOR directory_descriptor;
+    SECURITY_ATTRIBUTES directory_attributes;
 } win_security_t;
 
 typedef struct win_generation_address {
@@ -3560,6 +3565,8 @@ static void win_security_destroy(win_security_t *security) {
     }
     free(security->descriptor);
     free(security->acl);
+    free(security->directory_descriptor);
+    free(security->directory_acl);
     free(security->user_sid);
     if (security->advapi) {
         (void)FreeLibrary(security->advapi);
@@ -3614,6 +3621,8 @@ static bool win_security_init(win_security_t *security) {
     RESOLVE_ADVAPI_MEMBER(security, initialize_acl, initialize_acl_fn, "InitializeAcl");
     RESOLVE_ADVAPI_MEMBER(security, add_access_allowed_ace, add_access_allowed_ace_fn,
                           "AddAccessAllowedAce");
+    RESOLVE_ADVAPI_MEMBER(security, add_access_allowed_ace_ex, add_access_allowed_ace_ex_fn,
+                          "AddAccessAllowedAceEx");
     RESOLVE_ADVAPI_MEMBER(security, initialize_security_descriptor,
                           initialize_security_descriptor_fn, "InitializeSecurityDescriptor");
     RESOLVE_ADVAPI_MEMBER(security, set_security_descriptor_dacl, set_security_descriptor_dacl_fn,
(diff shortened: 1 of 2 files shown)


Kiborgik:
On the `test-windows-guards` red (`section_cold_storm`, client 3): I could not reproduce it, and the ordering in the captured output argues against this branch.

The failing client's last line was `version_cohort.claimed_unheld`, which `cbm_version_cohort_acquire` emits at `src/main.c:2438`. That runs only after the `coordination_failure` gate at `:2418-2429` has passed, so the client had already completed endpoint creation, `cbm_project_lock_manager_new`, the cohort manager, executable resolution and build identity. Those are the stages this branch touches. It changes the DACL applied to the runtime directory and nothing after it.

Local runs, MinGW gcc on Windows 11 26200:

| Build | `TEST_SEAMS` | Runs | `section_cold_storm` |
|---|---|---|---|
| this branch | off | 2 | green |
| this branch | on | 1 | green |
| `main` (control, same toolchain) | on | 1 | green |

All four runs were green across all seven sections. The control and the patched binary differ only in this change.

One thing may be worth fixing in the harness regardless. `section_cold_storm` prints `out_text(result)[:300]`, and the two startup lines (`version_cohort.claimed_unheld` and the `mem.allocator.preloading_completed` warning) already exceed that, so the actual failure reason is truncated away. Every failure path after the cohort claim does print a reason; `main_local_transition_acquire` returning 0 gives "CLI startup coordination remained busy", for example. Raising that slice, or printing the tail instead of the head, would make the next occurrence diagnosable. `section_churn_stability` already prints `stdout[-300:]` and `stderr[-400:]`, which is the shape that would have shown it.

My guess is startup-transition contention among six racing cold clients on a shared runner, which a 28-core machine resolves and a 2-core one may not. That is a guess. I can rebuild under `mingw-w64-clang` to match the runner more closely if another data point would help.

DeusData:
Merged — thank you @Kiborgik. This is precise Windows work and the diagnosis is the valuable part: a flagless ACE combined with `PROTECTED_DACL_SECURITY_INFORMATION` severs inherited ACEs while propagating nothing, so every child object is created with an **empty** DACL. That is exactly the shape of #1351 (cache files left unreadable), and it is the kind of bug that looks like a permissions mystery from the outside.

The `FILE_ALL_ACCESS` vs `GENERIC_ALL` note deserves highlighting for anyone who touches this later: Windows splits an inheritable generic-rights ACE into a mapped ACE plus an INHERIT_ONLY one carrying the generic bits, which would produce two ACEs — and our owner-only DACL validators require exactly one. Getting that wrong would have traded an empty-DACL bug for a validation failure.

Worth noting this does **not** relax the security posture: children now inherit the same owner-only ACE the parent already enforced. Shipping in **v0.10.3**.

Issue #557, opened 2026-06-22
cbm v0.8.1 silently deletes project DBs on "corrupt" detection — data loss with no recovery


jonlimx:
## Summary

`codebase-memory-mcp` v0.8.1 (and earlier) contains a `deleting corrupt db` code
path that **silently unlinks the project database file** when its startup
integrity check raises any anomaly. This is a "delete on first suspicion"
design that causes total data loss with no backup, no log warning, and no
recovery path — even when the "corruption" is a false positive.

## Evidence

The behavior is in the compiled binary and observable via `strings`:

```
deleting corrupt db
ERROR store.corrupt table=projects rows=%d (expected 1)
ERROR store.corrupt table=projects bad_root_path=%s
integrity_check_failed
malformed database schema (%s)
file is not a database
broke stale lock on %s
```

Related binary paths (WAL mode + atomic rename used for *normal* writes):

```
PRAGMA journal_mode = WAL;
%s.tmp
dump: rename failed
rename_to_cache
```

## Trigger conditions (false positives)

In my case the project DB was deleted after I performed a sequence of
`pkill -9` on the cbm process and binary swaps between `standard` and `ui`
variants (the `install.sh --ui` switch). The next cbm startup detected
"corruption" and deleted the 28 MB `<project>.db` file. Other plausible
triggers I can think of:

1. **WAL file leftover** from a previously-SIGKILL'd cbm (most likely in
   my case). When the old process is killed, SQLite's `-wal` and `-shm`
   sidecar files can be left in a state the new process interprets as
   corruption.
2. **Binary version drift** between cbm builds (e.g., switching between
   `codebase-memory-mcp-darwin-arm64.tar.gz` and
   `codebase-memory-mcp-ui-darwin-arm64.tar.gz`). Even minor schema-version
   differences would trip the check.
3. **Lock-file contention** — the `broke stale lock on %s` string shows cbm
   forcibly breaks stale locks; the DB state immediately after lock break
   may look anomalous and trigger the delete.

## Repro

1. `codebase-memory-mcp cli index_repository '{"repo_path":"/path/to/proj","mode":"full"}'`
2. `pkill -9 -f "codebase-memory-mcp"` while index is in progress (or
   between operations)
3. Start cbm again. The `<proj>.db` file disappears.
4. No log line other than `level=info msg=mem.init ...` precedes the loss.

## Impact

- Total data loss of the indexed project (in my case: 8 740 nodes /
[...]

Issue #391, opened 2026-05-30
task: Silent index corruption / incomplete index reported as success (3 issues)


DeusData:
## Scope

Umbrella tracker for cases where indexing reports `status: "indexed"` / `ready` but the resulting graph is catastrophically incomplete (single-digit % of expected node count, or empty). The reports cluster around macOS, large Rust projects, and rapid kill/restart cycles where the previous process didn't release WAL or hash-row state cleanly.

PR #387 ("checkpoint WAL on close and startup to prevent orphan accumulation") resolves the WAL subset; reporters asked to re-test on v0.7.0.

## Sub-issues

- [ ] #260 — Indexing completes but DB is empty (all zeros) — macOS arm64 v0.6.0 — **awaiting v0.7.0 retest** (WAL checkpoint-on-open/close + startup recovery shipped via #387/a6ad401; not deterministically reproducible)
- [ ] #333 — ~500 nodes for 72k LOC Rust — **reclassified: Rust extraction-depth gap, not corruption**. Rust is tree-sitter-only (not in the hybrid-LSP set), so extraction is shallow; asked reporter for v0.7.0 node count + missing symbol kinds + a public repo to tighten the tree-sitter query.
- [ ] #334 — Silent index corruption after rapid kill/restart cycles — **awaiting v0.7.0 retest** (#387 checkpoints WAL on open+close; crash-safety covered by `bulk_crash_recovery` / `store_checkpoint` tests)

## Status (2026-05-31)

The shared WAL/checkpoint root cause (#260, #334) is addressed by #387 in v0.7.0 (checkpoint on open + close, startup WAL recovery), with existing crash-recovery regression tests. Neither is deterministically reproducible in CI (timing/platform-specific), so both await reporter confirmation on v0.7.0. #333 is a distinct Rust extraction-completeness gap (tree-sitter-only language), not corruption — tracked on that basis pending a reproducible Rust repo.

## Acceptance

1. `status: "indexed"` is never returned when the index is materially incomplete — WAL stranding fixed (#387); a plausibility-threshold gate (criterion 2) remains a candidate enhancement.
2. A plausibility threshold gates the success status — open (enhancement).
3. Recovery path is automatic — WAL is now checkpointed on open+close; no manual MCP reconnect needed.
4. Concurrent-writer / rapid-kill scenarios reproducible in a test — `bulk_crash_recovery` covers the committed/uncommitted invariants.

## Why one task

[...]

PR #1545, opened 2026-08-11
fix(store): distinguish transient SQLITE_BUSY from corruption before quarantine (#1206, #1037)

### LINKED ISSUE #1037: Healthy project db silently quarantined to .corrupt and replaced with near-empty graph during concurrent CLI + multi-session MCP access (likely #391 family; correction in comments)
## Environment

- **Version**: codebase-memory-mcp 0.9.0 (confirmed latest via `codebase-memory-mcp update -n`)
- **Platform**: macOS (Apple Silicon) — macOS 26.5.2, Darwin arm64
- **Install channel**: GitHub release archive / install.sh
- **Binary variant**: ui (`ui_enabled: true`, binary ~273 MB)

A routine `index_repository` call on an **already-indexed, healthy project** silently discarded the existing graph and replaced it with a near-empty one — with `status: "indexed"` reported as success both times, no error, no warning.

Sequence:
1. Project was previously indexed: 8,559 nodes / 25,407 edges, `index_status` reported `status: "ready"`.
2. Ran `index_repository` (`mode: "fast"`) with no other activity in between — no kill signal, no binary swap, no relative path, no artifact export, no restart. Output reported success with the *correct* pre-existing counts (8,559 / 25,407) in that call's own log line.
3. Immediately after, the on-disk `<project>.db` was found renamed to `<project>.db.corrupt` (preserved, not deleted — this is an improvement over the `unlink` behavior in #557, so the rename-on-corruption fix appears at least partially shipped). A **new, near-empty** `<project>.db` had been created in its place.
4. Subsequent searches against the "healthy" project now only see the new database. `index_status` on it reported **20 nodes / 19 edges**.
5. Ran `index_repository` again (`mode: "fast"`) on the small db — reported success, grew to a still-catastrophically-small graph.
6. Ran `index_repository` with `mode: "full"` explicitly, hoping for a clean rebuild — reported success, **784 nodes / 1,445 edges**. Still nowhere near the original 8,559 / 25,407, despite the source repository being fully present and unchanged on disk (only two files had actually been edited since the last real index: a markdown doc and a config file — no source code changed).
7. Recovered by manually renaming `<project>.db.corrupt` back over the fresh `<project>.db` — `index_status` immediately reported the correct 8,559 / 25,407 again, confirming the *original* backup was structurally intact and just needed restoring.

**Expected**: `index_repository` on an already-indexed project with a small, non-code-affecting diff should either (a) do an incremental update preserving the existing graph, or (b) if it decides a full rebuild is warranted, actually rebuild the *complete* graph rather than silently producing a 90%+ smaller one while still reporting `status: "indexed"` / success.

## Possible new clue: SQLite writer-version mismatch between the "corrupt" file and the fresh one

```
$ file <project>.db.corrupt
SQLite 3.x database, last written using SQLite version 3046000, page size 1,
file counter 1, database pages 356, cookie 0x1, schema 4, UTF-8, version-valid-for 1

$ file <project>.db   # the fresh, near-empty replacement
SQLite 3.x database, last written using SQLite version 3051003, page size 1,
writer version 2, read version 2, file counter 3, database pages 68, cookie 0x6,
schema 4, UTF-8, version-valid-for 3
```

`3046000` = SQLite 3.46.0, `3051003` = SQLite 3.51.3. The "corrupt" file was last written by a meaningfully older SQLite than the one the current `codebase-memory-mcp` binary links against. My guess: the original index was built by an earlier `cbm` release (before an `update`/reinstall bumped the bundled SQLite), and the current binary's integrity check treats that older on-disk format as corrupt on first touch, quarantines it, and silently starts fresh — landing squarely in the pattern #391 is tracking, but via a **normal `index_repository` call with no crash/kill involved**, which doesn't match any of the currently-listed sub-issues' trigger conditions.

## Reproduction

I was not able to produce this from scratch on a synthetic/public repo — the trigger appears to require an index built by an *older* `cbm` binary version than the one currently installed, which isn't something I can fabricate without actually having an older release on hand to build the initial index. Happy to attempt this if a maintainer can point me at which older release/commit would be useful to install first (e.g. a build linking SQLite 3.46.x), or if `CBM_DIAGNOSTICS=1` output from a re-trigger would help — I'd rather not re-run this against my own project again without guidance since it destroyed the graph the first time and I've already restored from the `.corrupt` backup.

What I can confirm reproduced twice in a row, deterministically, on my actual project:
1. `codebase-memory-mcp cli index_repository --repo-path <path> --mode fast` on an already-indexed project → graph collapses to a fraction of its size, `status: "indexed"` still reported.
2. Repeating the same command against the now-small db → still small, still reports success.
3. `--mode full` on the now-small db → does **not** recover the original scale (784 vs. 8,559 nodes), despite the actual source tree being fully present and almost entirely unchanged.

Project shape (can't share the repository itself — private business codebase): TypeScript monorepo, 5 services (Next.js frontend, 2 Node backend services, a Python service, a shared package), several hundred source files.

## Logs

```
level=info msg=mem.init budget_mb=11468 total_ram_mb=32768
level=info msg=index.supervisor.reap outcome=clean exit_code=0 signal=0
{"project":"<redacted>","excluded":{"dirs":[...],"count":54,"truncated":true},"skipped_count":0,"nodes":8559,"edges":25407,...,"status":"indexed"}
```
No warning, error, or corruption-related log line preceded or accompanied the collapse — only the same `mem.init` / `index.supervisor.reap outcome=clean exit_code=0` lines seen on every normal run. The `.corrupt` rename happened silently between calls.

## Project scale (if relevant)

8,559 nodes / 25,407 edges → 20 nodes / 19 edges → 784 nodes / 1,445 edges (recovered to 8,559 / 25,407 by manually restoring the `.corrupt` backup)

## Confirmations

- [x] I searched existing issues and this is not a duplicate — related to the #391 umbrella and #557 specifically, but the trigger (plain `index_repository` on a healthy already-indexed project, no kill/restart/relative-path/artifact-export involved) doesn't match any currently-listed sub-issue, so filing separately per the existing triage pattern.
- [ ] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code. **Cannot confirm** — see the Reproduction section for why, and what I'd need to produce one.

### LINKED ISSUE #1206: Concurrent MCP server instances quarantine each other's healthy databases
# Concurrent MCP server instances quarantine each other's healthy databases

**Version:** codebase-memory-mcp 0.9.0 (official release, Linux x86_64)

**Impact:** A healthy index database is renamed to `.corrupt` and rebuilt from
scratch. The replacement is often left sparse (a few dozen nodes instead of
thousands) and stays that way, because `index_repository --mode full` cannot
repair it — only `delete_project` + reindex can.

Patch included below; verified against the reproduction.

## Summary

When two or more MCP server instances have the same project as their session
root, a file change makes all of them re-index at once. Whichever instance loses
the race treats a transient SQLite failure as corruption, renames the database to
`<project>.db.corrupt`, and starts a fresh one — even though the database it
discarded was perfectly healthy.

This is not disk corruption. `PRAGMA integrity_check` returns `ok` on the
quarantined files.

## Reproduction

`watcher-repro.sh` (attached) — `bash watcher-repro.sh <N>`

Creates a throwaway git repo with 20 PHP files, indexes it, starts N MCP server
instances with that repo as cwd (each completing an `initialize` handshake so the
watcher binds), then rewrites the 20 files three times.

| N (instances) | Result |
| --- | --- |
| 5 | `.corrupt` appears on the **first** round of file changes |
| 1 | 3 rounds of changes, no corruption |

Reproduced reliably. The only variable is the number of instances.

## Evidence

**1. The quarantined file in the reproduction is an empty database.**

```
$ stat -c%s tmp-wtest.db.corrupt
4096                                   # one SQLite header page
$ sqlite3 tmp-wtest.db.corrupt "SELECT count(*) FROM sqlite_master;"
0                                      # no tables at all
$ sqlite3 tmp-wtest.db.corrupt "PRAGMA integrity_check;"
ok
```

A database another instance was still creating got opened, found to have no
`projects` table, and quarantined.

**2. In production, the quarantined file was a fully populated database.**

A different manifestation on a real project (~2900 nodes, 7 instances):

```
$ sqlite3 var-www-lk.db.corrupt "PRAGMA integrity_check;"
ok
$ sqlite3 var-www-lk.db.corrupt \
    "SELECT (SELECT count(*) FROM nodes), (SELECT count(*) FROM projects), root_path FROM projects;"
2956|1|/var/www/lk
```

Exactly one `projects` row, correct `root_path`, intact data — quarantined
anyway. The replacement rebuilt to 2961 nodes, so nothing was wrong with the
discarded one.

**3. Every file quarantined in production had a valid `projects` table.**

Instrumented the daily maintenance job to dump the table before deleting the
backups. Three quarantines on one machine in a single day:

```
var-www-callhub      projects: 1 row  root_path=/var/www/callhub
var-www-html-salary  projects: 1 row  root_path=/var/www/html/salary
var-www-lk           projects: 1 row  root_path=/var/www/lk
```

One row, correct path, in all three. None of them was corrupt.

**4. Real-world correlation is exact.**

Across 42 indexed projects on this host, mapping each running instance's cwd:

| Project | Instances | Ever quarantined |
| --- | --- | --- |
| /var/www/lk | 7 | yes (3×) |
| /var/www/html/salary | 5 | yes |
| /var/www/callhub | 3 | yes |
| the other 39 projects | 0–1 | never |

No project with a single instance has ever been quarantined; every project with
more than one has.

## Why multiple instances are easy to end up with

Every editor/agent session that mounts the server spawns its own instance, and
each one starts a watcher on its session root:

```
level=info msg=watcher.start interval_ms=multi-sec
level=info msg=watcher.watch project=var-www-lk path=/var/www/lk
```

Two terminals open on the same repository is enough. On this host the count
reached 7 for one project across concurrent sessions.

## Root cause

`cbm_store_check_integrity()` in `src/store/store.c` treats **any** SQLite error
as corruption:

```c
int rc = sqlite3_prepare_v2(s->db, "SELECT count(*) FROM projects;",
                            CBM_NOT_FOUND, &stmt, NULL);
if (rc != SQLITE_OK) {
    return false;          /* caller quarantines the DB */
}
```

`resolve_store()` in `src/mcp/mcp.c` responds to `false` by renaming the file to
`.corrupt` and rebuilding. Two very common failures here are not corruption:

- **`SQLITE_ERROR` — "no such table: projects".** Another instance is still
  creating this database and has not committed the schema yet. Quarantining here
  destroys the database that instance is building. This produces the 4 KB,
  zero-table `.corrupt` in Evidence 1.
- **`SQLITE_BUSY` / `SQLITE_LOCKED`.** Another instance holds a write lock. The
  database is fine; it just could not be read at that moment. This produces the
  healthy 2956-node `.corrupt` in Evidence 2.

Both become likely as soon as more than one instance watches a project, because a
single file change makes all of them re-index simultaneously.

`cbm_store_check_integrity_deep()` has the same problem on its `quick_check` path.

## Fix

Report corruption only for result codes that actually mean the file is
unreadable, and let transient failures pass:

```c
/* Compare the primary result code so extended codes (SQLITE_CORRUPT_VTAB, …)
 * still match. */
static bool cbm_store_rc_is_corrupt(int rc) {
    int primary = rc & 0xff;
    return primary == SQLITE_CORRUPT || primary == SQLITE_NOTADB;
}
```

then use it at every `rc != SQLITE_OK` / non-`SQLITE_ROW` branch in both
integrity functions instead of returning `false` unconditionally.

Patch attached: `fix-transient-not-corrupt.patch` (42 insertions, 6 deletions,
`src/store/store.c` only).

| Build | 5 instances, 3 rounds of file changes |
| --- | --- |
| 0.9.0 as released | `.corrupt` on round 1 |
| 0.9.0 + patch | no `.corrupt`; index updates normally (102 → 122 nodes) |

Running the patched build here since 2026-07-22. The three projects that were
quarantining daily have stopped.

## Two related issues worth considering separately

1. **A quarantined project is silently left degraded.** After the rename, a fresh
   database is created and only *changed* files get indexed into it, so the
   project can sit at a few dozen nodes while still appearing normal in
   `list_projects`. Observed: 13 nodes instead of ~2700 for several hours.
   `index_repository --mode full` does not fix it — change detection sees no
   modified files and does nothing:

   ```
   # after emptying the nodes table
   $ codebase-memory-mcp cli index_repository --repo-path $T --mode full
   "nodes":0                              # unchanged
   $ codebase-memory-mcp cli delete_project --project $P
   $ codebase-memory-mcp cli index_repository --repo-path $T --mode full
   "nodes":14                             # only delete_project first recovers it
   ```

   Forcing a full reindex after a quarantine would close this.

2. **Re-indexing is not serialised across instances.** The patch stops the
   destructive misclassification, but N instances still redo the same work on
   every file change. An advisory lock on the database file would remove the
   duplicated effort as well as the contention.

---

<details>
<summary>fix-transient-not-corrupt.patch</summary>

```diff
diff --git a/src/store/store.c b/src/store/store.c
index 7b96605..ed00f3f 100644
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -861,6 +861,17 @@ cbm_store_t *cbm_store_open_path_query(const char *db_path) {
 
 /* ── Integrity check ───────────────────────────────────────────── */
 
+/* Does this SQLite result code mean the file itself is unreadable?
+ *
+ * Anything else (SQLITE_BUSY, SQLITE_LOCKED, "no such table" while another
+ * process is still creating the schema) is transient and must NOT be reported
+ * as corruption — the caller renames the DB away on a false positive. Compare
+ * the primary result code so extended codes (SQLITE_CORRUPT_VTAB etc.) match. */
+static bool cbm_store_rc_is_corrupt(int rc) {
+    int primary = rc & 0xff;
+    return primary == SQLITE_CORRUPT || primary == SQLITE_NOTADB;
+}
+
 /* Deep integrity: the shallow check above only sanity-checks the projects
  * table, so page-level corruption (torn artifacts, #895) sails through.
  * quick_check(1) walks the btrees and stops at the first error. Used on
@@ -870,17 +881,21 @@ bool cbm_store_check_integrity_deep(cbm_store_t *s) {
         return false;
     }
     sqlite3_stmt *stmt = NULL;
-    if (sqlite3_prepare_v2(s->db, "PRAGMA quick_check(1);", CBM_NOT_FOUND, &stmt, NULL) !=
-        SQLITE_OK) {
-        return false;
+    int rc = sqlite3_prepare_v2(s->db, "PRAGMA quick_check(1);", CBM_NOT_FOUND, &stmt, NULL);
+    if (rc != SQLITE_OK) {
+        /* Same reasoning as the shallow check: a lock we lost is not corruption. */
+        return cbm_store_rc_is_corrupt(rc) ? false : true;
     }
     bool ok = false;
-    if (sqlite3_step(stmt) == SQLITE_ROW) {
+    int step_rc = sqlite3_step(stmt);
+    if (step_rc == SQLITE_ROW) {
         const char *res = (const char *)sqlite3_column_text(stmt, 0);
         ok = res && strcmp(res, "ok") == 0;
         if (!ok) {
             (void)fprintf(stderr, "ERROR store.corrupt quick_check=%s\n", res ? res : "(null)");
         }
+    } else {
+        ok = !cbm_store_rc_is_corrupt(step_rc);
     }
     sqlite3_finalize(stmt);
     return ok;
@@ -898,17 +913,38 @@ bool cbm_store_check_integrity(cbm_store_t *s) {
     int rc =
         sqlite3_prepare_v2(s->db, "SELECT count(*) FROM projects;", CBM_NOT_FOUND, &stmt, NULL);
     if (rc != SQLITE_OK) {
-        return false;
+        /* Only report corruption for errors that actually mean "this file is
+         * unreadable". The caller's response is destructive (rename the DB to
+         * .corrupt and re-index), so a transient failure must not land here.
+         *
+         * Two transient cases show up constantly once more than one server
+         * instance watches the same project — a file change makes them all
+         * re-index at once:
+         *   - "no such table: projects" (SQLITE_ERROR): another instance is
+         *     still creating this DB and hasn't committed the schema yet.
+         *     Quarantining here destroys the DB that instance is building.
+         *   - SQLITE_BUSY / SQLITE_LOCKED: another instance holds a write lock.
+         *     The DB is perfectly healthy; we just could not read it right now.
+         * Both were observed in the wild quarantining good databases. */
+        if (cbm_store_rc_is_corrupt(rc)) {
+            (void)fprintf(stderr, "ERROR store.corrupt table=projects prepare_rc=%d\n", rc);
+            return false;
+        }
+        return true;
     }
 
     bool ok = true;
-    if (sqlite3_step(stmt) == SQLITE_ROW) {
+    int step_rc = sqlite3_step(stmt);
+    if (step_rc == SQLITE_ROW) {
         int row_count = sqlite3_column_int(stmt, 0);
         if (row_count > ST_MAX_ROW_CHECK) {
             (void)fprintf(stderr, "ERROR store.corrupt table=projects rows=%d (expected 1)\n",
                           row_count);
             ok = false;
         }
+    } else if (cbm_store_rc_is_corrupt(step_rc)) {
+        (void)fprintf(stderr, "ERROR store.corrupt table=projects step_rc=%d\n", step_rc);
+        ok = false;
     }
     sqlite3_finalize(stmt);
 
```

</details>

<details>
<summary>watcher-repro.sh</summary>

```bash
#!/usr/bin/env bash
# 復現假設：多個 MCP server 同時監看同一專案，檔案變動時各自觸發重索引 → projects 表出現重複列
set -u
CM=/home/ying/.local/bin/codebase-memory-mcp
T=/tmp/wtest
DB=~/.cache/codebase-memory-mcp/tmp-wtest.db
N=${1:-5}
RUN=/tmp/wrepro; rm -rf $RUN; mkdir -p $RUN

INIT='{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"t","version":"1"}}}'

cd "$T" || exit 1
for i in $(seq 1 "$N"); do
  mkfifo $RUN/in$i
  $CM < $RUN/in$i > $RUN/out$i 2> $RUN/err$i &
  echo $! >> $RUN/pids
  exec {fd}> $RUN/in$i        # 保持寫入端開著，server 才不會看到 EOF 退出
  eval "FD$i=$fd"
  echo "$INIT" >&$fd
done

sleep 6
echo "--- 啟動 $N 個 server，watcher 綁定狀況 ---"
grep -h "watcher.watch\|session.root.cwd" $RUN/err* 2>/dev/null | sort | uniq -c | sed 's/^/  /'

echo "--- 檢查 DB（變更前）---"
sqlite3 "$DB" "SELECT '  projects='||count(*)||' nodes='||(SELECT count(*) FROM nodes) FROM projects;" 2>&1

echo "--- 連續改檔觸發 watcher（3 輪）---"
for round in 1 2 3; do
  for i in $(seq 1 20); do
    echo "<?php class W$i { function a(){return $i;} function b(){return $((i*round));} function r$round(){return 1;} }" > "$T/f$i.php"
  done
  echo "  第 $round 輪已改 20 檔"
  sleep 12
  sqlite3 "$DB" "SELECT '    projects='||count(*)||' root='||coalesce(group_concat(root_path,'|'),'?') FROM projects;" 2>&1
  ls ~/.cache/codebase-memory-mcp/tmp-wtest.db.corrupt 2>/dev/null && echo "    *** 出現 .corrupt ***"
done

echo "--- 收尾：關閉 server ---"
for i in $(seq 1 "$N"); do eval "exec {FD$i}>&-"; done
sleep 2
for p in $(cat $RUN/pids 2>/dev/null); do kill -TERM "$p" 2>/dev/null; done
sleep 2
echo "--- 最終狀態 ---"
sqlite3 "$DB" "SELECT '  projects='||count(*)||' nodes='||(SELECT count(*) FROM nodes) FROM projects;" 2>&1
ls ~/.cache/codebase-memory-mcp/tmp-wtest* 2>/dev/null | sed 's/^/  /'
echo "--- server stderr 中的異常 ---"
grep -hiE "corrupt|error|malformed|quarantin" $RUN/err* 2>/dev/null | sort -u | head -10
echo "  (空=無異常)"
```

</details>

LynxBay:
## Problem

Concurrent MCP server instances quarantine each other's healthy databases (#1206), and page-torn DBs with intact `projects` tables sail through the integrity check (#1037). Both bugs share one root cause: the quarantine decision in `resolve_store_internal()` relies on `cbm_store_check_integrity()`, which

1. **treats any `sqlite3_prepare_v2` failure as corruption** — so a `SQLITE_BUSY`/`SQLITE_LOCKED` from writer-lock contention (the normal outcome when N instances race on the same project) triggers a quarantine of a DB that `PRAGMA integrity_check` later reports as `ok`;
2. **never runs `PRAGMA quick_check`** — so a DB whose `projects` table is fine but whose node/edge btrees are torn passes the check and keeps being served.

Reproduction from #1206: 5 MCP instances on one repo → `.corrupt` appears on the **first** file change; 1 instance → never. Field evidence from #1037: an 8,559-node graph collapses to 20 nodes after a routine `index_repository`, while the quarantined `.corrupt` backup passes `PRAGMA integrity_check`.

## Fix

Introduce `cbm_store_check_integrity_verdict()` (`src/store/store.h` / `store.c`) returning a tri-state:

| Verdict | Meaning | Quarantine? |
|---|---|---|
| `CBM_INTEGRITY_OK` | DB is healthy | no |
| `CBM_INTEGRITY_CORRUPT` | Structural damage confirmed | **yes** |
| `CBM_INTEGRITY_TRANSIENT` | `SQLITE_BUSY` / `LOCKED` / IO-lock — inconclusive | **no** — retry next resolve |

It runs the existing shallow check (projects-row-count + root-path sanity) **plus `PRAGMA quick_check(1)`** for page-level damage, and classifies each `sqlite3_prepare_v2`/`step` failure via a new `st_rc_is_transient()` helper that recognizes `SQLITE_BUSY`, `SQLITE_LOCKED`, `SQLITE_IOERR_LOCK`, `SQLITE_IOERR_BLOCKED`.

The quarantine decision in `src/mcp/mcp.c` (`resolve_store_internal`, the second check performed after the mutation lease is acquired) now:

- quarantines **only** on `CBM_INTEGRITY_CORRUPT`;
- on `CBM_INTEGRITY_TRANSIENT`, closes the store, releases the lease, sets `STORE_RECOVERY_BUSY`, and returns `NULL` — leaving the DB file untouched so the next caller retries.

The existing `cbm_store_check_integrity()` and `cbm_store_check_integrity_deep()` are **unchanged**, so hot opens, artifact import, and maintenance keep their current behavior. The new function is used solely on the quarantine decision path; its cost (one `PRAGMA quick_check`, O(db size)) is paid only when the cheap open-time check has already failed — i.e. exactly the paths that used to destroy healthy DBs.

## Verification

- Builds clean with `-Wall -Wextra -Werror` (the production CFLAGS in `Makefile.cbm`).
- `cli list_projects` and `cli index_status` run correctly against the patched binary on a real 175k-node / 597k-edge project DB.
- No new compiler warnings.

## Scope

3 files changed, +149 / −2. No change to the on-disk DB format, the quarantine file-naming scheme, or any existing public API.

## Related issues

Fixes #1206, fixes #1037. Related: #391 (silent index corruption umbrella), #557 (the older silent-delete variant), #867 (incremental reindex cost — the amplifier that makes spurious quarantines so painful).

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -2156,9 +2156,34 @@ static cbm_store_t *resolve_store_internal(cbm_mcp_server_t *srv, const char *pr
             }
 
             /* The lease may have waited behind a publisher. Re-open and trust
-             * only the current generation, never the stale pre-wait verdict. */
+             * only the current generation, never the stale pre-wait verdict.
+             * Use the verdict API here — this is the point that decides whether
+             * a healthy DB gets quarantined. The plain bool check cannot tell
+             * corruption from a transient SQLITE_BUSY race (#1206: concurrent
+             * instances quarantining each other's DBs) and does not run
+             * quick_check, so page-torn DBs with an intact projects table sail
+             * through (#1037). Only a confirmed CORRUPT verdict is quarantined;
+             * TRANSIENT (lock/IO) falls through and retries on next access. */
             srv->store = cbm_store_open_path_query(path);
-            bool current_valid = srv->store && cbm_store_check_integrity(srv->store);
+            cbm_integrity_verdict_t verdict = srv->store
+                                                  ? cbm_store_check_integrity_verdict(srv->store)
+                                                  : CBM_INTEGRITY_TRANSIENT;
+            bool current_valid = (verdict == CBM_INTEGRITY_OK);
+            if (verdict == CBM_INTEGRITY_TRANSIENT) {
+                /* The DB could not be conclusively evaluated (lock contention,
+                 * busy writer, IO hiccup). Do NOT quarantine — close and let
+                 * the next resolve retry. A spurious quarantine here is exactly
+                 * what destroys healthy DBs under concurrent access. */
+                cbm_store_close(srv->store);
+                srv->store = NULL;
+                if (recovery_status) {
+                    *recovery_status = STORE_RECOVERY_BUSY;
+                }
+                if (!mutation_already_held) {
+                    mcp_project_mutation_end(srv, project);
+                }
+                return NULL;
+            }
             if (!current_valid) {
                 cbm_store_close(srv->store);
                 srv->store = NULL;
--- a/src/store/store.c
+++ b/src/store/store.c
@@ -948,6 +948,107 @@ bool cbm_store_check_integrity_deep(cbm_store_t *s) {
     return ok;
 }
 
+/* Classify a SQLite result code as a transient (retryable) condition vs. a
+ * hard error. SQLITE_BUSY / SQLITE_LOCKED happen when another connection holds
+ * the writer lock — they are NOT evidence of corruption, yet the bare
+ * cbm_store_check_integrity() treats any prepare failure as "corrupt". This
+ * helper backs the verdict API so the quarantine path stops destroying healthy
+ * DBs that merely lost a lock race (#1206, #1037). */
+static bool st_rc_is_transient(int rc) {
+    switch (rc) {
+    case SQLITE_BUSY:
+    case SQLITE_LOCKED:
+    case SQLITE_IOERR_LOCK:
+    case SQLITE_IOERR_BLOCKED:
+        return true;
+    default:
+        return false;
+    }
+}
+
+cbm_integrity_verdict_t cbm_store_check_integrity_verdict(cbm_store_t *s) {
+    if (!s || !s->db) {
+        /* No handle at all — the caller could not even open the file. That is
+         * an open problem, not a corruption verdict, so the DB must not be
+         * quarantined (renaming it would discard a file we never read). */
+        return CBM_INTEGRITY_TRANSIENT;
+    }
+
+    /* ── Shallow check (projects table shape + root_path sanity) ── */
+    sqlite3_stmt *stmt = NULL;
+    int rc =
+        sqlite3_prepare_v2(s->db, "SELECT count(*) FROM projects;", CBM_NOT_FOUND, &stmt, NULL);
+    if (rc != SQLITE_OK) {
+        /* A prepare failure here is the #1206 trigger: under concurrent access
+         * the schema lookup can return BUSY/LOCKED. Treat transient codes as
+         * "do not quarantine"; only a hard error is corruption evidence. */
+        return st_rc_is_transient(rc) ? CBM_INTEGRITY_TRANSIENT : CBM_INTEGRITY_CORRUPT;
+    }
+    cbm_integrity_verdict_t verdict = CBM_INTEGRITY_OK;
+    if (sqlite3_step(stmt) == SQLITE_ROW) {
+        int row_count = sqlite3_column_int(stmt, 0);
+        if (row_count > ST_MAX_ROW_CHECK) {
+            (void)fprintf(stderr, "ERROR store.corrupt table=projects rows=%d (expected 1)\n",
+                          row_count);
+            verdict = CBM_INTEGRITY_CORRUPT;
+        }
+    } else {
+        /* step returned DONE/ERROR — for a SELECT count(*) this means the table
+         * is unreadable. BUSY-family => transient; otherwise corruption. */
+        int step_rc = sqlite3_errcode(s->db);
+        verdict = st_rc_is_transient(step_rc) ? CBM_INTEGRITY_TRANSIENT : CBM_INTEGRITY_CORRUPT;
+    }
+    sqlite3_finalize(stmt);
+    if (verdict != CBM_INTEGRITY_OK) {
+        return verdict;
+    }
+
+    /* root_path sanity — same logic as cbm_store_check_integrity, but classify
+     * prepare failures as transient vs. corrupt rather than a blanket false. */
+    rc = sqlite3_prepare_v2(s->db,
+                            "SELECT root_path FROM projects WHERE root_path != '' "
+                            "AND NOT (substr(root_path, 1, 1) = '/' "
+                            "OR (substr(root_path, 1, 1) BETWEEN 'A' AND 'Z') "
+                            "OR (substr(root_path, 1, 1) BETWEEN 'a' AND 'z')) LIMIT 1;",
+                            CBM_NOT_FOUND, &stmt, NULL);
+    if (rc != SQLITE_OK) {
+        return st_rc_is_transient(rc) ? CBM_INTEGRITY_TRANSIENT : CBM_INTEGRITY_CORRUPT;
+    }
+    if (sqlite3_step(stmt) == SQLITE_ROW) {
+        const char *bad_path = (const char *)sqlite3_column_text(stmt, 0);
+        (void)fprintf(stderr, "ERROR store.corrupt table=projects bad_root_path=%s\n",
+                      bad_path ? bad_path : "(null)");
+        verdict = CBM_INTEGRITY_CORRUPT;
+    }
+    sqlite3_finalize(stmt);
+    if (verdict != CBM_INTEGRITY_OK) {
+        return verdict;
+    }
+
+    /* ── Deep check: walk the btrees for page-level damage (#1037) ──
+     * The shallow check above passes for a DB whose projects table is intact
+     * but whose node/edge btrees are torn (observed: 185k nodes / 6k edges with
+     * PRAGMA integrity_check reporting btreeInitPage errors). Only quick_check
+     * catches that. quick_check is O(db size); this function runs only on the
+     * quarantine decision path, never on hot opens. */
+    rc = sqlite3_prepare_v2(s->db, "PRAGMA quick_check(1);", CBM_NOT_FOUND, &stmt, NULL);
+    if (rc != SQLITE_OK) {
+        return st_rc_is_transient(rc) ? CBM_INTEGRITY_TRANSIENT : CBM_INTEGRITY_CORRUPT;
+    }
+    if (sqlite3_step(stmt) == SQLITE_ROW) {
+        const char *res = (const char *)sqlite3_column_text(stmt, 0);
+        if (!res || strcmp(res, "ok") != 0) {
+            (void)fprintf(stderr, "ERROR store.corrupt quick_check=%s\n", res ? res : "(null)");
+            verdict = CBM_INTEGRITY_CORRUPT;
+        }
+    } else {
+        int step_rc = sqlite3_errcode(s->db);
+        verdict = st_rc_is_transient(step_rc) ? CBM_INTEGRITY_TRANSIENT : CBM_INTEGRITY_CORRUPT;
+    }
+    sqlite3_finalize(stmt);
+    return verdict;
+}
+
 bool cbm_store_check_integrity(cbm_store_t *s) {
     if (!s || !s->db) {
         return false;
--- a/src/store/store.h
+++ b/src/store/store.h
@@ -282,6 +282,28 @@ bool cbm_store_check_integrity(cbm_store_t *s);
  * O(db size); use on rare paths (artifact import), not hot opens. */
 bool cbm_store_check_integrity_deep(cbm_store_t *s);
 
+/* Outcome of a quarantine-grade integrity check. Used to decide whether a DB
+ * that failed the cheap open-time check should be quarantined (renamed to
+ * .corrupt and rebuilt) or left alone. See cbm_store_check_integrity_verdict. */
+typedef enum {
+    CBM_INTEGRITY_OK = 0,        /* DB is healthy */
+    CBM_INTEGRITY_CORRUPT = 1,   /* DB is structurally damaged — safe to quarantine */
+    CBM_INTEGRITY_TRANSIENT = 2, /* SQL/busy/IO error — NOT corruption, do NOT quarantine */
+} cbm_integrity_verdict_t;
+
+/* Full integrity verdict for the quarantine decision path.
+ *
+ * The plain cbm_store_check_integrity() returns a single bool and cannot
+ * distinguish "the projects table has 99 rows" (real corruption) from
+ * "sqlite3_prepare_v2 returned SQLITE_BUSY because another instance held the
+ * writer lock" (a transient lock contention, #1206). Quarantining on the latter
+ * is what makes concurrent MCP instances destroy each other's healthy DBs.
+ *
+ * This function runs the shallow check, then PRAGMA quick_check, and classifies
+ * the failure mode so the caller can quarantine ONLY on confirmed corruption.
+ * O(db size); use only on the recovery/quarantine path, not hot opens. */
+cbm_integrity_verdict_t cbm_store_check_integrity_verdict(cbm_store_t *s);
+
 /* Open database for a named project in the default cache dir. */
 cbm_store_t *cbm_store_open(const char *project);
 


DeusData:
Merged — thank you @LynxBay. This is an unusually good diff: the verdict enum makes the distinction explicit rather than hiding it in a bool, the transient classification is exactly right (BUSY/LOCKED/IOERR_LOCK/IOERR_BLOCKED are lock-contention evidence, not corruption evidence), and reusing the existing `STORE_RECOVERY_BUSY` path means a contended project now surfaces "project is busy; retry after indexing" instead of quarantining a healthy database. Adding `quick_check` on the same path closes the other half (#1037's page-torn DB with an intact `projects` table).

Two long-standing priority/high issues resolved in one change: #1206 and #1037.

One thing we're adding on top, not a criticism of the diff: it ships without tests, and this is a data-loss-class path — so we're binding it in the v0.10.3 batch with cases that assert (a) a BUSY prepare yields TRANSIENT and leaves the file untouched, and (b) a torn btree with a healthy `projects` table yields CORRUPT. Those tests exist because of your fix and will keep it from regressing.

Shipping in **v0.10.3** today.

PR #1549, opened 2026-08-11
chore(homebrew): point the formula at v0.10.2

DeusData:
The formula still pinned **0.8.1** — four releases and two months behind — so `brew install` handed out a June binary while npm, PyPI and the GitHub archives all served the current one. Nothing in the release pipeline touches this file, which is how it drifted that far unnoticed.

Version plus all four SHA-256 digests taken from the published v0.10.2 `checksums.txt`.

Follow-up worth its own change: have the release workflow rewrite the formula from `checksums.txt` after publish, so this channel cannot silently rot again.


DeusData:
Superseded by the v0.10.3 bump — v0.10.2 was overtaken before this could merge. Same change, current digests.

PR #1553, opened 2026-08-12
fix(cli): default-export Pi extension factory

### LINKED ISSUE #1550: Pi generated extension should default-export its factory function
## Summary

The generated Pi extension currently exports a named `register(pi)` function, but Pi expects an extension file to default-export a factory function.

This makes Pi fail on startup whenever the generated extension is present.

## Environment

- codebase-memory-mcp: `0.10.1`
- Pi coding agent: `0.84.1`
- macOS arm64

## Repro

After installing codebase-memory-mcp with Pi integration, the generated file at:

```text
~/.pi/agent/extensions/cbmem.ts
```

contained:

```ts
export function register(pi) {
  pi.registerTool({ name: 'index_repository', run: (args, ctx) => call('index_repository', args, ctx?.signal) });
  // ...
}
```

Running any Pi command with extensions enabled then failed:

```sh
pi doctor
```

with:

```text
Error: Failed to load extension "/Users/musichen/.pi/agent/extensions/cbmem.ts": Extension does not export a valid factory function: /Users/musichen/.pi/agent/extensions/cbmem.ts
Hint: Start without extensions using "pi -ne".
```

## Local fix verified

Changing the generated export to a default factory fixed Pi startup:

```diff
-export function register(pi) {
+export default function (pi) {
```

After this change, both commands worked with extensions enabled:

```sh
pi doctor
pi -p 'reply exactly: pi-ok'
```

## Expected generator output

The Pi extension generator should emit:

```ts
export default function (pi) {
  pi.registerTool({ name: 'index_repository', run: (args, ctx) => call('index_repository', args, ctx?.signal) });
}
```

## Related

I also opened a Pi issue asking for a clearer loader diagnostic when the default export is missing:

https://github.com/earendil-works/pi/issues/7985

musichen:
## Summary
- emit the generated Pi adapter as a default-exported factory, matching Pi's extension contract
- add a regression test that rejects the incompatible named `register(pi)` export

## Testing
- `make -f Makefile.cbm test-focused TEST_SUITES=agent_clients` (31 passed, ASan/UBSan)
- regression test confirmed failing before the generator fix and passing afterward
- `git diff --check`

Fixes #1550

### DIFF
--- a/src/cli/client_adapter.c
+++ b/src/cli/client_adapter.c
@@ -156,7 +156,7 @@ char *cbm_client_adapter_pi(const char *binary_path) {
 
     /* Register every tool the registry advertises — never a hand-picked subset,
      * which is the drift this generator exists to prevent. */
-    sb_append(&sb, "export function register(pi) {\n");
+    sb_append(&sb, "export default function (pi) {\n");
     for (int i = 0; i < count; i++) {
         const char *name = cbm_mcp_tool_name(i);
         if (!name || !name[0]) {
--- a/tests/test_agent_clients.c
+++ b/tests/test_agent_clients.c
@@ -1071,6 +1071,18 @@ TEST(agent_clients_continue_refuses_foreign_same_name_and_nonsequence_section) {
  * of the tool surface. #534 shipped one mirroring 7 of 15 tools; every tool
  * added afterwards would have been silently missing for that client. This pins
  * the property that makes generation worth doing: EVERY registry tool appears. */
+TEST(client_adapter_pi_default_exports_factory) {
+    char *js = cbm_client_adapter_pi("/usr/local/bin/codebase-memory-mcp");
+    ASSERT_NOT_NULL(js);
+
+    /* Pi imports only the module's default export as its extension factory. */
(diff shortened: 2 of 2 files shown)


musichen:
This addresses the Pi extension loading failure reported in #1550.

DeusData:
Thank you @musichen — you reported this **and** fixed it within the hour, which is the best kind of contribution to receive.

Closing only because ours was already in CI when yours arrived (commit `42e320a6` and its parent on `fix/batch2-regressions-and-determinism`, shipping in **v0.10.3**) — not because there was anything wrong with yours. Your diff and ours are the same change, and you are recorded as a co-author on the commit.

Your unblock instructions in the issue were also exactly right, and I pointed other Pi users at them.

musichen:
@DeusData thank you 
Sure!  Main thing is usability of codebase-memory-mcp for other users  with pi. 
          I'll be reporting anything else that i find during sessions with codebase-memory-mcp

DeusData:
Thank you @musichen. We appreciate both the original report and the prompt fix. The superseding implementation retains your co-author credit, and your follow-up does not require reopening this PR.

Issue #1574, opened 2026-08-12
No product-build way to relocate the daemon runtime directory when %LOCALAPPDATA% has a capability-SID ACE


tmonestudio:
## Summary

On Windows the daemon/CLI coordination endpoint is created under
`%LOCALAPPDATA%`, resolved by `SHGetFolderPathW(CSIDL_LOCAL_APPDATA)` with no
override in a product build. If that ancestry carries a capability-SID ACE with
mutation rights — common on real machines once packaged apps are installed — the
binary cannot start at all, and there is no supported way to point it elsewhere.

The only relocation hook, `CBM_TEST_DAEMON_RUNTIME_PARENT`, is compiled out
unless `CBM_ENABLE_TEST_SEAMS` is defined, so the test binary starts fine while
the product binary does not.

## Environment

- Windows 11 Pro 10.0.26200, single user, non-domain.
- Working build: `main` @ `7d6cdb23` + local `cc56d509`.
- Failing build: current `main`, built 2026-08-12.

## Observed

```
codebase-memory-mcp: secure daemon endpoint could not be created
codebase-memory-mcp: secure CLI coordination could not be created (endpoint)
```

Every invocation path fails, including `config list`, so the settings surface is
unreachable. Independent of `CBM_CACHE_DIR`: with the cache on a clean
volume-root path the cache check passes and the endpoint failure remains.

Walking the ancestry with the same predicate as
`win_private_directory_tree_secure`, the only rejected entries under
`%LOCALAPPDATA%` are a single capability SID:

```
C:\Users\<user>\AppData        S-1-15-3-...  mask=0x000D0152  inherit_only=false
C:\Users\<user>\AppData\Local  S-1-15-3-...  mask=0x000D0152  inherit_only=false
```

`0x000D0152` includes `WRITE_DAC`, `WRITE_OWNER` and `DELETE`. We are **not**
asking you to trust that — we agree it is a real tamper vector. Note it is not a
Windows default: `C:\Users\Default\AppData\Local` has no such ACE. It arrived
with an installed application, and it is present on `AppData`, `Local`,
`Roaming` and `LocalLow`.

## Requests, in order of preference

1. **A product-supported override for the runtime parent directory** — a config
   key or environment variable, validated by the same
   `win_private_directory_tree_secure`. The operator picks a path with clean
   ancestry; the check keeps its full strength. This is the one that unblocks us
   without weakening anything.
[...]

Issue #1537, opened 2026-08-11
No such xattr: com.apple.quarantine


listepo:
### Version

0.10

### Platform

macOS (Apple Silicon)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

codebase-memory-mcp installer
  os:      darwin
  arch:    arm64
  target:  /Users/listepo/.local/bin/codebase-memory-mcp

Downloading codebase-memory-mcp-darwin-arm64.tar.gz...
######################################################################## 100.0%
Checksum verified.
Extracting...
Fixing macOS code signing...
Verified candidate: codebase-memory-mcp 0.10.0
codebase-memory-mcp install 0.10.0

xattr: /tmp/cbm-install-zRIanz/codebase-memory-mcp: No such xattr: com.apple.quarantine
/tmp/cbm-install-zRIanz/codebase-memory-mcp: replacing existing signature
codebase-memory-mcp 0.10.0
error: active CBM sessions and operations could not be stopped safely; no activation was committed.

### Reproduction

codebase-memory-mcp uninstall
curl -fsSL https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh | bash

### Logs

```text

```

### Diagnostics trajectory (memory / performance / leak issues)

```text

```

### Confirmations

- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

DeusData:
Thanks for the clear report, @listepo — the full installer transcript makes this easy to read, and it surfaced two things worth fixing.

**The xattr line in the title is actually harmless** — the installer tries to strip macOS's quarantine attribute and prints that message when there's none to strip (fresh download via `curl` doesn't always carry it). It should be silent; we'll make it so. Sorry it pointed you at the wrong suspect.

**The real failure is the last line:** `active CBM sessions and operations could not be stopped safely; no activation was committed`. The installer refuses to swap the binary while something still holds a live cbm session — most likely an MCP client (Claude Code, Codex, an editor integration) that kept a server process running through your uninstall, or a daemon it respawned. The refusal itself is safety-correct, but it has two real defects on macOS which we're treating as the bug here:

1. It doesn't tell you **which** processes are blocking — on Windows this same refusal lists the pids since v0.10.0 (#1416), and the POSIX path should do exactly the same.
[...]

PR #1576, opened 2026-08-12
feat(daemon): let a product build relocate the runtime directory

tmonestudio:
Closes the blocker reported in #1574.

## Problem

The daemon/CLI rendezvous directory is created under `%LOCALAPPDATA%` (Windows)
or the POSIX runtime/home directory, and a product build has no way to point it
elsewhere. When that ancestry carries a mutation-granting ACE for an untrusted
identity, it fails `cbm_daemon_ipc_private_directory_secure()` and the binary
cannot start at all — `config list` included, since it needs the same endpoint,
so the operator cannot even inspect settings to diagnose it.

The only relocation hook, `CBM_TEST_DAEMON_RUNTIME_PARENT`, is compiled out
unless `CBM_ENABLE_TEST_SEAMS` is defined. That is why a test build starts on an
affected machine while the product build does not.

## Approach

`CBM_RUNTIME_DIR` **does not relax the check**. The named directory goes through
exactly the same validation, and an unusable value is refused rather than
silently ignored. The operator only chooses an ancestry that passes.

Two details worth reviewing:

- The test seam keeps precedence over the product override, so existing
  lifecycle tests keep their isolated rendezvous namespace.
- The override is applied at **every** product endpoint call site — daemon,
  local CLI, tool execution and index worker. Wiring only
  `main_daemon_endpoint_new()` would have left the `cli` path on the default and
  looked fixed while it was not.

## Evidence

Measured on the affected machine (Windows 11, single user), same environment,
same cache, only the patch differing:

| binary | `config list` |
|---|---|
| current `main` | `secure CLI coordination could not be created (endpoint)` |
| this branch, `CBM_RUNTIME_DIR` set to a clean-ancestry parent | starts and prints the configuration |

Two non-obvious properties this surfaced, which may be worth documenting
regardless of this PR:

- The directory **must be created by the tool**, not pre-created by the
  operator. A pre-created directory inherits the volume's ACEs and then fails as
  an ancestor; the tool hardens the final directory itself.
- Consequently the override should name a parent whose own ancestry is clean —
  a volume root works.

## Tests

`tests/test_product_runtime_dir_override.py`, a process-level guard in the style
of `tests/test_daemon_open_readiness.py`. It asserts that the rendezvous is
created under the named directory, that leaving the variable unset creates
nothing there, and that an unusable value is refused rather than ignored.

**Honest disclosure about what I could and could not run locally:**

- The behavior above was verified by hand against a build of this branch.
- The new guard **could not go green on my machine**, because its temporary
  directory lives under the very `%LOCALAPPDATA%` whose ACE causes the bug, so
  the parent it creates is itself rejected. On a clean runner it should pass;
  your CI will be the first to actually execute it.
- I could not run the C suite locally at all: `make -f Makefile.cbm test-focused`
  fails to compile `tests/test_daemon_ipc.c` with
  `-Werror=stringop-overflow` under w64devkit GCC 2.8.0. I verified this is
  **pre-existing and unrelated** by building the same target on the unpatched
  base commit — identical failure, same line. Your CI uses LLVM 20, which does
  not raise it.

Happy to adjust the variable name, move the guard into the C suite, or rework
the approach entirely if you would rather solve this differently.

### DIFF
--- a/src/git/git_context.c
+++ b/src/git/git_context.c
@@ -16,6 +16,17 @@ enum {
     GIT_OUTPUT_MAX = 4096,
 };
 
+/* Bounded worktree-status stream constants.
+ * STATUS_FIELD_MAX is the largest single NUL-delimited record field we
+ * buffer; a field at/over the bound is treated as a malformed stream and the
+ * whole snapshot fails closed. PATH components of real repos are far below
+ * this even with long paths. */
+enum {
+    STATUS_HEADER_LEN = 3, /* porcelain-v1 record prefix "XY " */
+    STATUS_CHUNK_MAX = CBM_SZ_4K,
+    STATUS_FIELD_MAX = CBM_SZ_16K,
+};
+
 static char *git_strdup(const char *s) {
     if (!s) {
         s = "";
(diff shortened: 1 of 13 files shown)


DeusData:
Thank you @tmonestudio for turning #1574 into a fail-closed proposal rather than weakening the ancestry check. The issue is grounded on current `main`, and the runtime-parent direction is reviewable.

Before substantive review, this branch needs to be unstacked. Its first four commits (`36d467dc`, `e49f7e57`, `86fca027`, `167783bb`) are the complete #1561 index-freshness series; only `038194e7` is the runtime-directory change. As a result, GitHub currently shows 13 files and +1,482/-26 here, including store, git-context, and MCP changes unrelated to the stated claim.

Please rebase/cherry-pick the runtime commit onto current `main` so this PR contains one change, then let CI rerun on that standalone diff. The current lint and Windows smoke failures should be evaluated on that clean branch. The review queue is full, and this security-sensitive path will need careful review, but the contribution has been seen and is queued.

DeusData:
Thank you for this — your idea shipped, and you are named in the commit.

`CBM_RUNTIME_DIR` is on `main` as 7f3e30e1 (PR #1645), and the message credits you directly:

> Approach and variable name from #1576 by Leonardo trindade miranda, resolved one layer lower so the activation path is covered too.

The one change is where it is resolved. Yours sat in the daemon path; the merged version resolves it in `bootstrap_runtime_parent_override`, the single function every product endpoint goes through — daemon, MCP client, local CLI, index worker and activation — so no call site can silently keep the default. That matters here because the failure it fixes hits `install` and `config list` too, not only the daemon.

It keeps your semantics exactly: it does **not** relax the check. The directory you name goes through the same validation as the default, and a value that fails is refused rather than ignored — the operator chooses an ancestry that passes, they do not get to switch the rule off.

It ships in v0.10.5 today, and it is the escape hatch several people have needed for weeks — #1537, #1574 and #1621 all reduce to "the rendezvous is not relocatable in a shipped build".

Two things I'm sorry about: this PR stayed open and unanswered while its idea went into the product, and your **#1561** is still open and blocked on nothing but `clang-format` (`src/store/store.c:3171-3172`, `src/mcp/mcp.c:4504-4506`). If you run `make lint-ci` and push, that one should go green — it is not superseded by anything and I'd like to get it in.

DeusData:
Closing as shipped: `CBM_RUNTIME_DIR` is on `main` in 7f3e30e1 (#1645), crediting you by name, resolved one layer lower so `install` and `config list` are covered too. It ships in v0.10.5 today. Your #1561 is still open and blocked on nothing but clang-format — I'd like to get that one in. Thank you.

PR #1595, opened 2026-08-13
fix(subprocess+tsan): widen the spawn-retry window where it was silently disabled

DeusData:
Two commits, and the second is why the first appeared not to work.

**1. Wider spawn-retry budget for sanitized builds.** The exponential backoff shipped in v0.10.3 (~0.6s) fixed the ordinary case, but `subprocess_run_spawn_failure` kept failing under ThreadSanitizer, which runs several times slower and holds far more process state. Raising the budget for everyone would be wrong — an unsanitized machine genuinely out of capacity should fail fast — so the extra patience is scoped to sanitized builds (~5s).

**2. …except `CBM_SANITIZED_BUILD` was never defined on the TSan leg**, so commit 1 did nothing there. It failed again on the very PR meant to fix it.

`SANITIZED_DEFINE` keys off `$(SANITIZE)`. TSan does not use that variable — it has its own `TSAN_SANITIZE` — and `CFLAGS_TSAN` never included `SANITIZED_DEFINE`. So the macro was undefined and every sanitized-budget branch compiled to its **native** value while running an instrumented binary.

The comment above `SANITIZED_DEFINE` already describes this precise failure for trap-UBSan: *"the build system is the single source of truth for is this binary instrumented; compiler-specific probes miss clang's feature-check spelling and every non-ASan sanitizer."* The lesson was recorded; the TSan leg was simply never wired up to it. And compiler probes would not have rescued it either — clang spells this `__has_feature(thread_sanitizer)`, not `__SANITIZE_THREAD__`.

**Checked the neighbours:** MSan (`scripts/msan.sh` passes `SANITIZE=`) and the diag lane (same) already get the define. TSan was the only gap — but while it existed, *any* sanitized-budget logic was silently inert there, including the daemon announce backstop.

**How this was verified.** TSan does not run in PR CI — it lives in `_test.yml`, which only the
dry-run and release workflows call — so these checks cannot prove the fix. The build configuration
proves it directly instead:

```
# pre-fix tree (a92b6be9)
$ make -f Makefile.cbm test-tsan -n | grep -c 'DCBM_SANITIZED_BUILD=1'   -> 0
$ make -f Makefile.cbm test-tsan -n | grep -c 'sanitize=thread'          -> 169

# this branch
$ make -f Makefile.cbm test-tsan -n | grep -c 'DCBM_SANITIZED_BUILD=1'   -> present
$ make -f Makefile.cbm test-tsan -n | grep -c 'sanitize=thread'          -> 169
```

169 thread-instrumented compile lines and not one of them carried the define. That is the whole
bug, and it is gone. The end-to-end confirmation is the TSan leg of the dry-run/release run.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -105,9 +105,15 @@ CXXFLAGS_TEST = $(CXXFLAGS_COMMON) $(SANITIZED_DEFINE) -g -O1 $(SANITIZE) $(CXX_
 
 # TSan (can't combine with ASan)
 TSAN_SANITIZE = -fsanitize=thread -fno-omit-frame-pointer
+# CBM_SANITIZED_BUILD is unconditional here: this flag set exists only to build
+# an instrumented binary. SANITIZED_DEFINE keys off $(SANITIZE), which TSan does
+# not use (it has its own TSAN_SANITIZE), so the define was silently absent and
+# every sanitized-budget branch compiled to its NATIVE value on this leg —
+# exactly the failure the comment above SANITIZED_DEFINE describes for
+# trap-UBSan, repeated on a leg that was never wired up.
 CFLAGS_TSAN = $(CFLAGS_COMMON) $(EDITOR_TEST_DEFINES) $(KOTLIN_DEDUP_TEST_DEFINE) \
 	$(CALL_REFERENCE_LOOKUP_TEST_DEFINE) $(INCREMENTAL_TEST_DEFINE) \
-	-g -O1 $(TSAN_SANITIZE)
+	-DCBM_SANITIZED_BUILD=1 -g -O1 $(TSAN_SANITIZE)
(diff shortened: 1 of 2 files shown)


DeusData:
Correction to my own note in the description above: I claimed TSan does not run in PR CI. That is wrong — `pr.yml:33` calls `_test.yml` as a reusable workflow, which is why grepping `pr.yml` for "tsan" found nothing and why I concluded the leg only ran in dry-run/release.

The upshot is better than what I wrote: this PR got the direct end-to-end proof after all. All three TSan jobs passed, including `test / test-tsan (macos-14, cc, c++)` — the exact job that had been failing with `subprocess_run_spawn_failure` on two separate SHAs.

The build-configuration evidence in the description still stands as the mechanism (169 thread-instrumented compile lines, zero carrying `-DCBM_SANITIZED_BUILD=1` before this change), but it was not the only proof available. Leaving the record straight.

PR #1596, opened 2026-08-13
ci: select release binaries before smoke testing

DeusData:
## Summary

- build stripped and unstripped candidates for every release target
- VirusTotal-scan both candidates before smoke/soak and select per target, preferring stripped when verdicts are equivalent
- package and verify the exact selected bytes, then reuse that evidence in release notes without a duplicate final scan

## Validation

- release/VT contract tests pass
- lint passes
- macOS full suite passes: 7,465 passed, 0 failed, 4 skipped
- remaining validation is delegated to PR CI


DeusData:
Reviewed for inclusion in v0.10.4. The engineering here is careful and the diff is clean — no new dependencies, every action SHA-pinned, and the "scan A, ship B" class is ruled out four independent times (content-addressed objects, re-hash on read and on copy, `--expected-sha256` re-checked against the extracted archive member, then `verify-release-selection.py` again on the real published assets). The registry-ordering fix survives intact (`publish-mcp-registry` still `needs: [publish-registries, publish-final]`). The macOS ad-hoc codesign actually gets *stronger*: signing now happens after strip with a hard `--verify --strict`, where the old path signed, stripped, then re-signed best-effort with stderr discarded.

The problem is not the code, it's the evidence. **No dry-run has ever been dispatched on this branch**, and PR CI never executes `release.yml`, `_build.yml` or `_soak.yml` — so every assurance so far is a contract test that greps YAML, or a local shell test. The first real execution of the new `select-package` join job would be the v0.10.4 release itself, with at least five first-time-in-anger unknowns on the critical path: `resolve_strip` under MSYS2 CLANGARM64 (now hard-fails instead of falling back across `STRIP`/`llvm-strip`/`strip`), darwin and windows containers now packaged on `ubuntu-latest` with GNU tar / Linux `zip -X`, a 16-object VirusTotal submission with a 4-hour poll, soak downloading and executing release artifacts across 6–8 jobs, and `select-package` as a brand-new single point of failure for the whole release.

So I have dispatched that dry-run rather than guessing: run 31729540899, with `skip_virustotal=false` so it exercises the full 16-object scan, packaging, smoke and the artifact-consuming soak. If it comes back green this is ship-with-confidence; v0.10.4 waits for it.

Four things in here are decisions rather than defects, and they need a maintainer call:

1. **Row 3 of the truth table ships an *unstripped* binary to users when an opaque ML classifier flags the stripped one.** Bigger asset, exposed symbol names and build paths, chosen by an AV verdict. Documented and defensible — but it changes what users receive.
2. **A `hard` verdict on the unstripped sibling — a binary we never ship — kills all 8 targets with no operator bypass** (`scan_candidates: true` is hardcoded and a contract test forbids bypassing it). `package-release.sh` used to carry a comment recording that this exact thing happened: an unstripped linux-amd64 artifact was the sole Microsoft detection in run 30398064336 while the stripped siblings scanned clean. Blocking only when the *selected* binary is hard is a one-line policy change.
3. **`soak_level: none` is deleted and soak becomes draft-blocking**, now 6–8 jobs including a new portable pair.
4. **VirusTotal scope narrows** from "every extracted member and unpacked UI asset" to "the 16 executables", and README / SECURITY.md / docs are already rewritten to match. The new wording is the honest one, but it is a public-surface change riding along in a CI PR.

One thing I would change regardless: `secrets: inherit` was added at `release.yml:84` and `dry-run.yml:90`, but `_build.yml` declares exactly one optional secret. `inherit` hands *every* repo secret — NPM_TOKEN, PYPI_TOKEN — to all 8 build jobs, including the ones running `npm ci` over the graph-ui dependency tree. Both workflows are `workflow_dispatch`-only so there is no fork-PR exposure; this is blast radius, not an open hole. Passing the one key explicitly is the fix, and `tests/test_release_gate_chain_contract.sh:176` currently pins the loose form, so it has to change with it.

Also worth noting: `scripts/ci/publish-vt-evidence.sh` is now dead code with no caller, and `package-release.sh` gained a breaking CLI change — bare `scripts/package-release.sh linux amd64` now exits 2.

DeusData:
Dry-run result: **[run 31729540899](https://github.com/DeusData/codebase-memory-mcp/actions/runs/31729540899) failed — 45 jobs green, and the one failure is `build / select-package`**, the new job. Had this shipped in v0.10.4, the release would have died here, after every build completed and before smoke.

The cause is a new check this PR adds, not a pre-existing one:

```
BLOCKED: VirusTotal response analysis id does not match the submitted analysis:
objects/70fbbc37994f7a1620b5bc997cd2f022ba9321273c1554110bd2bd3069b1ddd0
```

from the hunk added to `parse_completed`:

```python
if response_id != submission.analysis_id:
    raise GateError(
        f"VirusTotal response analysis id does not match the submitted analysis: "
        f"{submission.expected.scan_path}"
    )
```

The intent is right — pin the verdict to the submission so a response for some other file can't be read as ours. But strict equality on the analysis id is the wrong invariant: VirusTotal is content-addressed, and for content it has already seen it can return an existing analysis rather than the one your submission just created. That is normal API behaviour, not tampering, and this pipeline makes it *more* likely by submitting 16 objects keyed by SHA-256, several of which are byte-identical across targets or unchanged between runs.

The gate also aborted while four objects were still `queued; will retry round-robin`, so this fired before the scan had even finished — the five that had returned were all clean (61–65 decisive engines).

The property actually worth enforcing is that the verdict belongs to *these bytes*, and the response already carries that: pin on `meta.file_info.sha256` (or `data.attributes.results`' subject) matching the submitted object's hash, and treat an analysis-id difference as information rather than a failure. That keeps the guarantee and drops the false positive.

Worth saying clearly: this is one over-strict line in an otherwise careful change, and the dry-run did exactly its job. Three things in this PR are real improvements independent of the strip-selection idea, and I would want all three regardless:

1. **The scan moves before publication.** Today the VT scan lives in `verify`, which `needs: [release-draft, security]` — we learn about a detection only after the draft release exists and its assets are uploaded. Scanning before smoke/soak/draft is strictly better.
2. **Smoke and soak execute the exact bytes we ship**, rather than a separately-produced artifact.
3. **A real macOS signing bug is fixed.** Main signs at `_build.yml:79`, then `package-release.sh:143` strips (invalidating the signature) and re-signs best-effort with `2>/dev/null` and no verification — so a failed re-sign ships silently, and an ad-hoc signature is not something a launch test catches. This PR signs *after* strip with a hard `--verify --strict`.

Once the analysis-id check is relaxed, please re-run this dry-run rather than reasoning about it — that is what caught this.

PR #1600, opened 2026-08-13
test(cli): neutralize ambient client home overrides in the C runner

phyrexia:
## Problem

The CLI resolves each client's config home from that client's own environment variable **before** falling back to `$HOME`. `cbm_codex_config_dir()` (`src/cli/cli.c:2211`) returns `$CODEX_HOME` and ignores the `home_dir` it is handed; the same holds for `CLAUDE_CONFIG_DIR`, `OPENCODE_CONFIG_DIR`, `COPILOT_HOME` and the rest.

A test that redirects only `HOME` therefore does **not** isolate those clients. On any machine with `$CODEX_HOME` exported — every Codex user, and every Orca user, whose runtime home is exported into the shell — the `cli` suite resolves, reads and *writes* the developer's real configuration.

On this machine that is 27 failures out of 269, and none of them is a defect in the behaviour under test:

```
$ build/c/test-runner cli
  241 passed, 28 failed          # 27 unique assertions
```

Each failure is an `rc` assertion. The CLI was reporting real failures — against the ambient config, not the fixture:

- an unbalanced managed marker left in the real `config.toml` → `toml_find_markers` is fail-closed (`src/cli/config_toml_edit.c:857`)
- an `AGENTS.md` that is a symlink → refused by the `O_NOFOLLOW` writer (`src/cli/config_text_edit.c:441`)
- user-modified agent profiles → correctly preserved, and preservation is counted as an install error by design

Every content assertion in those tests passed. One test prints its own diagnostic, which makes the shape obvious:

```
VS Code durable diag install_rc=-1 hook=1 skill=1 agent=1 mcp_absent=1
  instructions_absent=1 uninstall_rc=1 hook_removed=1 skill_removed=1 agent_preserved=1
```

Worse than the red: the run **mutated the real configuration**, installing and removing `skills/codebase-memory/SKILL.md` under the user's Codex home roughly fifteen times.

## Why not another per-test list

The tests that already neutralize `CODEX_HOME` (`tests/test_cli.c:4535`, `:4969`, `:5546`, …) are exactly the ones that pass. The file also carries a comment admitting a suite was *"green only via env leaked from earlier suites"*. Extending those ad-hoc lists a 13th time leaves the next client to be forgotten again.

`tests/test_smoke_fixture_contract.sh` already defines the canonical inventory and *enforces* it — but only for the shell fixtures. This applies the same contract to the C runner.

## Change

- `tests/test_main.c` — neutralize the client home overrides once, in `tf_setup_cache_sentinel()`, next to the existing `HOME` / `CBM_CACHE_DIR` isolation. A test that exercises one of these sets it after setup, so the existing per-test `setenv` calls keep working unchanged.
- `tests/test_smoke_fixture_contract.sh` — hoist the inventory into one constant and require the C runner to neutralize it too, so a client isolated in the smoke scripts cannot be forgotten in the runner.

No change outside `tests/`.

## Verification

macOS arm64, with `$CODEX_HOME` and `$OPENCODE_CONFIG_DIR` exported (i.e. the polluted environment, no `env -u` workaround):

| | before | after |
|---|---|---|
| `test-runner cli` | 241 passed, 28 failed | **268 passed, 0 failed** |
| `test-runner` (full) | 7431 passed, 30 failed | **7461 passed, 0 failed, 4 skipped** |

The guard was checked to be non-vacuous: removing one variable from the runner's list makes the contract fail with `tests/test_main.c must neutralize ambient KIMI_CODE_HOME`.

## Unrelated observation, not addressed here

While confirming the full-suite numbers, one test failed only when a real daemon was running on the machine:

```
daemon_ipc_posix_record_publication_windows_recover_from_crash
  FAIL tests/test_daemon_ipc.c:3465: ASSERT(case_ok[index])
```

It passes in isolation, and the full suite is green once no daemon is live. The daemon rendezvous directory is keyed per-uid and not derived from `HOME` or `CBM_CACHE_DIR` (`src/daemon/ipc.c:867`), so the suite shares `/tmp/cbm-daemon-<uid>` with whatever daemon the developer has installed. `cbm_cli_set_activation_runtime_parent_for_test` already exists as the seam but only one test uses it. Happy to open a separate issue if that is useful — keeping this PR to one thing.

### DIFF
--- a/tests/test_main.c
+++ b/tests/test_main.c
@@ -54,8 +54,7 @@ static bool tf_invoked_as_windows_git_module(void) {
      * Inspect the actual loaded module so the copied git.exe probe cannot fall
      * through into the ordinary test runner when argv[0] is merely "git". */
     wchar_t image[32768];
-    DWORD image_length =
-        GetModuleFileNameW(NULL, image, (DWORD)(sizeof(image) / sizeof(image[0])));
+    DWORD image_length = GetModuleFileNameW(NULL, image, (DWORD)(sizeof(image) / sizeof(image[0])));
     if (image_length == 0 || image_length >= (DWORD)(sizeof(image) / sizeof(image[0]))) {
         return false;
     }
@@ -86,8 +85,7 @@ static bool tf_invoked_as_blocking_git(const char *argv0) {
             base = cursor + 1;
         }
     }
-    return strcmp(base, "git") == 0 || strcmp(base, "git.exe") == 0 ||
-           strcmp(base, "GIT.EXE") == 0;
+    return strcmp(base, "git") == 0 || strcmp(base, "git.exe") == 0 || strcmp(base, "GIT.EXE") == 0;
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


phyrexia:
I found #1150 after opening this — it predates this PR by a month and diagnoses the same leak, in the same two files. Credit where it is due; I should have searched the open PRs first.

The two are not the same fix, so I am leaving this open for the maintainers to choose:

- #1150 **binds** `CODEX_HOME` to the private test home, hard-fails setup when isolation cannot be established, and adds a regression test for that binding. That is strictly better than unsetting for `CODEX_HOME` itself, since it also covers a client whose resolver has no `HOME` fallback.
- This PR **generalizes** to the whole inventory the shell fixtures already neutralize (22 variables — `CLAUDE_CONFIG_DIR`, `OPENCODE_CONFIG_DIR`, `COPILOT_HOME`, `KIRO_HOME`, the `OPENCLAW_*` set, …) and adds a contract check so the C runner cannot drift from that inventory again.

`CODEX_HOME` is the one that bites hardest today, but it is not the only one: `OPENCODE_CONFIG_DIR` is exported by the same tooling here and already causes a false detection in `cli_agent_client_registry_routes_plan_install_and_uninstall`.

Happy to rebase this on top of #1150 and keep only the generalization plus the contract guard, if you prefer to land that one first.

Issue #1565, opened 2026-08-12
Search performance and search cancellation


ertankucukoglu:
### What problem does this solve?

- There are about 184000 files in database and a simple *.pas filter searches among them all. However, there is only 445 *.pas files in database.
- There is a risk of corrupting common temporary files when there are parallel searches.
- Cancelled search continues its own search operation and does not stop.
- Performance counters are not quite detailed.
- A lot of matching search like letter "a" is not stopping the PowerShell search at 500 files even if C part is only using the first 500 accepted result.

### Proposed solution

I have below diff package against to the current master (hopefully it is still matching as master is quite frequently changing recently).
Total 5 files including new tests. Compiled and tested on my system.
[cbm-main-60a2bbf.zip](https://github.com/user-attachments/files/30978208/cbm-main-60a2bbf.zip)

Please feel free to use them as you like, or simply ignore if they break something else or does not feel right.

Thanks.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
@ertankucukoglu, thank you for profiling this and for preparing tests. We cannot review or apply source changes delivered only as an attached ZIP: it removes the normal GitHub diff, provenance, DCO, and per-commit review path, and the attachment is already drifting as `main` moves.

Please open standard pull requests against current `main`, one independently reviewable concern per PR. The natural slices are extension-filter narrowing, per-search temporary-file isolation, cancellation propagation, performance counters, and stopping the PowerShell producer at the accepted-result cap. Include the relevant regression test or measurement with each slice.

There is no need to maintain the ZIP for us. Focused GitHub PRs will let us review and credit the work normally. Thank you again for offering the improvements.

PR #1604, opened 2026-08-14
perf(mcp): prefilter simple suffix globs on Windows

ertankucukoglu:
## What does this PR do?
Prefilters safe simple suffix globs such as `*.pas` and `*.d.ts` before PowerShell `Select-String` on Windows.

Complex patterns continue to use the existing post-filter path, and the final defensive file-pattern filter is preserved.

Adds regression coverage for safe/unsafe pattern boundaries and PowerShell command ordering.

Test results: 7353 passed, 0 failed, 63 skipped.

Part of #1565

## Checklist

- [x] Every commit is signed off (`git commit -s`) — required, CI rejects
      unsigned commits ([DCO](../DCO), see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] Tests pass locally (`make -f Makefile.cbm test`)
- [x] Lint passes (`make -f Makefile.cbm lint-ci`)
- [x] New behavior is covered by a test (reproduce-first for bug fixes)

PR #1605, opened 2026-08-14
perf(mcp): report search phase timings

ertankucukoglu:
## What does this PR do?
Adds phase-level timing counters to `search_code` results:

- `scope_ms`
- `scan_ms`
- `enrich_ms`

The existing `elapsed_ms` field is preserved. All four timing fields are included in both the default tree output and JSON output.

Adds regression coverage for both output formats.

Test results: 7352 passed, 0 failed, 63 skipped.

Part of #1565

## Checklist

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -597,7 +597,9 @@ static const tool_def_t TOOLS[] = {
      "signatures+metadata (default). full: with source. files: just file list.\"},"
      "\"context\":{\"type\":\"integer\",\"description\":\"Lines of context around each match "
      "(like grep -C). Only used in compact mode.\"},"
-     "\"regex\":{\"type\":\"boolean\",\"default\":false},\"limit\":{\"type\":\"integer\","
+     "\"regex\":{\"type\":\"boolean\",\"default\":false},\"debug\":{\"type\":\"boolean\","
+     "\"default\":false,\"description\":\"Include scope_ms, scan_ms, and enrich_ms phase timing "
+     "diagnostics.\"},\"limit\":{\"type\":\"integer\","
      "\"description\":\"Max enriched results per call. Default 10. Response includes "
      "'total_grep_matches' and 'total_results' so callers can detect truncation. No "
      "offset parameter — raise limit or narrow with file_pattern / path_filter to see more."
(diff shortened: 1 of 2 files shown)


DeusData:
Maintainer decision: **yes, but behind a debug flag rather than always on.**

The diagnostic value is real — `scope_ms` / `scan_ms` / `enrich_ms` is exactly what we would want while chasing #1565. The reservation is that `search_code` is the highest-traffic tool in the server, and three extra scalars on every response is a permanent cost paid by every caller for a signal almost none of them use. A response field is also a one-way door: once agents parse it, removing it is a breaking change.

So the ask is to gate it — a `verbose` / debug flag on the request, absent by default. That keeps the numbers available when someone is tracing a slow search, without taxing the common path or committing the surface permanently.

The measurement work itself is good and the split you chose (scope / scan / enrich) is the right one — it separates the three phases we would actually want to distinguish. Happy to take it once it is behind the flag.

ertankucukoglu:
Updated as requested and rebased onto current main. The three phase timings are now omitted by default and exposed only when search_code receives debug:true; the existing elapsed_ms behavior is unchanged. Tests cover the default absence plus debug output in compact tree and JSON formats.

DeusData:
Checked it rather than taking it on trust — `debug` is declared `"default": false` in the schema and the three fields only emit when `include_phase_timings` is set from it, so a caller who does not ask sees exactly what they saw before. `elapsed_ms` untouched. Tests cover both directions in tree and JSON.

That is exactly the shape we wanted, turned around fast. Merging once CI clears.

Worth saying why the flag mattered rather than it being bureaucracy: `search_code` is the highest-traffic tool in the server, so anything unconditional in its response is paid for by every call forever, and a response field is a one-way door once agents start parsing it. Behind a flag you get the numbers when you are actually chasing something — which is precisely the #1565 case — without that cost.

Your #1606 is already merged, and #1604 is next up in your cluster: it is results-preserving by construction, so it wants to land before #1607/#1608 rather than after.

DeusData:
Merged — thank you, and thank you for reworking this behind a flag.

Reviewing it I nearly held it back: the PR description still says the four timing fields are "included in both the default tree output and JSON output", which would have been a public output-surface change and a direction question rather than a pickup. The code does the opposite — `debug` defaults to false, both emitters are gated on `include_phase_timings`, and even the measurement itself is skipped when it is off, so there is no cost for callers who do not ask.

What settled it was your test: it pins **both** directions — fields absent by default with `elapsed_ms` preserved, present with `debug:true`, in tree and JSON alike. That is exactly the shape we wanted.

Worth updating the description on the next one, since the body is what a reviewer reads first and this one argued against its own diff.

Ships in v0.10.5.

PR #1615, opened 2026-08-14
fix(build): give sanitized-build detection one spelling and a backstop

moffermann:
## What does this PR do?

Follow-up to #1590/#1595. Those widened the spawn-retry budget for sanitized builds and then had to
chase the define onto the TSan leg. This removes the class of failure rather than the instance.

**The four spellings.** Four places ask "is this binary instrumented?", each differently:

| site | condition |
|---|---|
| `src/foundation/subprocess.c` | `CBM_SANITIZED_BUILD` + three `__SANITIZE_*__` |
| `tests/test_daemon_frontend.c` | same four |
| `src/foundation/compat_thread.c` | `defined(X) && X` |
| `tests/test_cs_lsp_bench.c` | **ASan only** |

The last one means TSan and MSan were measuring an instrumented parse against the native 200 ms budget.

**Why the probes could not cover for the missing define.** `0a163d4f` concluded that compiler probes
would not have helped, because clang spells thread instrumentation `__has_feature(thread_sanitizer)`
rather than `__SANITIZE_THREAD__`. That is right about the probes we had — and it is exactly the
probe that was missing. No site consulted `__has_feature`.

**What replaces them.** `src/foundation/sanitized.h` answers once, as `CBM_SANITIZED`, from two
sources with distinct jobs:

- `CBM_SANITIZED_BUILD` from the build system stays the source of truth, and is the **only** thing
  that can answer for UBSan and trap-UBSan — UB instrumentation leaves no macro and no
  `__has_feature` bit behind.
- The clang and GCC probes are the backstop for the three sanitizers that do announce themselves, so
  a lane that forgets the define still gets correct budgets instead of native ones.

No `#error` when a probe fires without the define: promoting leaves the binary correct while the lane
gets fixed, and it does not break an out-of-tree `make CFLAGS_EXTRA=-fsanitize=address`.

**Build lanes.** The define now also goes on the instrumented flag sets of our own code that lacked
it: `CXXFLAGS_TSAN` (`preprocessor.cpp` is ours; `CXXFLAGS_TEST` already had it),
`GRAMMAR_CFLAGS_TEST`, `GRAMMAR_CFLAGS_TSAN`. Neither tree can include the header today (no `-Isrc`),
so this is the build system keeping its own promise, not a behaviour change. Vendored flag sets are
untouched.

**Behaviour change worth naming:** `test_cs_lsp_bench` now allows 2000 ms on the TSan and MSan lanes
instead of 200 ms. It loosens a bound that was being applied to an instrumented binary by accident.

## Verification

Resolution matrix, measured rather than assumed (clang 22, `-dM -E`):

| flags | `CBM_SANITIZED` |
|---|---|
| native | 0 |
| `-DCBM_SANITIZED_BUILD=1` | 1 |
| `-fsanitize=address` | 1 |
| `-fsanitize=thread` (linux target) | **1** |
| `-fsanitize=memory` (linux target) | 1 |
| `-fsanitize=undefined` | 0 — define-only, as designed |

The thread row is the point: the failure this header is named after would have self-healed.

**Not verified locally, needs CI:** this machine has no POSIX-target compiler, so the POSIX half of
`subprocess.c` was compiled only on the Windows leg, and the sanitized lanes (`test-tsan`, ASan,
MSan, trap-UBSan) were not run. `make -f Makefile.cbm test-foundation` does not link here for an
unrelated pre-existing reason — `tests/test_main.c` references the store/cypher/... suites
unconditionally, so that target cannot link against `TEST_FOUNDATION_SRCS` alone.

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) — see above, no POSIX toolchain on this box
- [x] Lint passes — clang-format clean on every touched file
- [x] New behavior is covered — the macro matrix above; the change is a compile-time contract with no
      runtime surface of its own

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -111,10 +111,15 @@ TSAN_SANITIZE = -fsanitize=thread -fno-omit-frame-pointer
 # every sanitized-budget branch compiled to its NATIVE value on this leg —
 # exactly the failure the comment above SANITIZED_DEFINE describes for
 # trap-UBSan, repeated on a leg that was never wired up.
+#
+# The same define now goes on EVERY instrumented flag set of our own code, C and
+# C++ alike (CXXFLAGS_TSAN here, GRAMMAR_CFLAGS_TEST/TSAN below). Vendored flag
+# sets are left alone: mimalloc, sqlite3, tre, zstd, lz4 and tree-sitter read no
+# macro of ours.
 CFLAGS_TSAN = $(CFLAGS_COMMON) $(EDITOR_TEST_DEFINES) $(KOTLIN_DEDUP_TEST_DEFINE) \
 	$(CALL_REFERENCE_LOOKUP_TEST_DEFINE) $(INCREMENTAL_TEST_DEFINE) \
 	-DCBM_SANITIZED_BUILD=1 -g -O1 $(TSAN_SANITIZE)
-CXXFLAGS_TSAN = $(CXXFLAGS_COMMON) -g -O1 \
+CXXFLAGS_TSAN = $(CXXFLAGS_COMMON) -DCBM_SANITIZED_BUILD=1 -g -O1 \
                 $(TSAN_SANITIZE)
 
 # Windows needs ws2_32 (Winsock), psapi (GetProcessMemoryInfo), shell32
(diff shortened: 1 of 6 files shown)


moffermann:
Force-pushed one amend: the first push failed the `lint / lint` job on cppcheck, not on a compiler.

```
src/foundation/sanitized.h:48:2: error: failed to evaluate #if condition,
undefined function-like macro invocation: __has_feature( ... ) [syntaxError]
```

The arm was nested under `#elif defined(__has_feature)`, which every compiler handles but cppcheck does not — it walks each configuration and rejects the file rather than skipping the group. Switched to the define-away idiom (`#ifndef __has_feature / #define __has_feature(x) 0`), which is what clang documents and what `tests/test_mem.c` already does. The resolution matrix in the description is unchanged; re-measured after the amend.

Issue #1616, opened 2026-08-14
cbm_nanosleep never resumes after EINTR: every timed wait can be silently truncated


moffermann:
### What

On POSIX `cbm_nanosleep` is raw `nanosleep` (`src/foundation/compat.h:96`), and all four call sites
pass `rem = NULL` and discard the return value:

- `src/foundation/subprocess.c:958` — `cbm_spawn_backoff`, the spawn-retry patience
- `src/daemon/bootstrap.c:274`
- `src/daemon/runtime.c:275`
- `src/foundation/diagnostics.c:464`

`nanosleep` returns `EINTR` on any signal delivered to the calling thread, and unlike most blocking
calls it is **not** restarted by `SA_RESTART` — the remaining time is simply lost unless the caller
loops on `rem`. So every one of these waits can end early by an arbitrary amount, silently.

### Why it matters

`cbm_spawn_backoff` is the clearest case. The comments around
`CBM_SPAWN_RETRY_ATTEMPTS` promise ~630 ms of patience on an ordinary build and ~5 s on a sanitized
one, and that is the whole point of the retry: ride out a burst of process creation instead of
failing a spawn that would have succeeded. A signal arriving mid-backoff cuts that budget to a
fraction of what the code says it is, and nothing reports it.

`src/cli/hook_augment.c:133` installs a `SIGALRM` handler, so this is reachable and not theoretical.

### Why this is an issue and not a PR

- It is **not** the cause of the `subprocess_run_spawn_failure` flakes that #1590/#1595 chased: CI
  runs no alarm, so the truncation does not fire there. Fixing it will not change those lanes.
- The right fix is not uniform. A `cbm_nanosleep_full()` that loops on the remainder is obviously
  right for the spawn backoff, but `bootstrap.c` and `runtime.c` may well *want* to wake early on a
  signal during shutdown. That is a call-site-by-call-site decision, not a mechanical replacement.

### Suggested shape

Add `cbm_nanosleep_full()` (loop while `nanosleep` returns `-1`/`EINTR`, feeding `rem` back in;
Windows `Sleep` already has no interruptible failure mode) and move the call sites that want an
uninterruptible wait — starting with `cbm_spawn_backoff` — over to it. Leave the daemon pauses on
the interruptible primitive if that is the intent, with a comment saying so.

Found while reviewing the sanitized-build wiring in #1615; unrelated to that change.

DeusData:
[...]

PR #1618, opened 2026-08-14
fix(subprocess): bound the spawn backoff by the wait, not by the budget

moffermann:
## What does this PR do?

`cbm_spawn_backoff` clamps its shift with

```c
int shift = attempt < CBM_SPAWN_RETRY_ATTEMPTS ? attempt : CBM_SPAWN_RETRY_ATTEMPTS;
```

and the comment above it says *"the budget is the only bound needed — callers never exceed it, and a
second clamp would be dead code"*. The clamp that stayed is the dead one: both retry loops return
before `attempt` reaches the budget (`cbm_fork_with_retry`, and the `posix_spawn` loop on Apple), so
it can never fire.

What it leaves unbounded is the case it was written for. Nothing stops the doubling, so the schedule
explodes with the budget instead of flattening:

| budget | waits | total |
|---|---|---|
| 6 | 10 20 40 80 160 320 | 630 ms |
| 9 | 10 20 40 80 160 320 640 1280 2560 | 5110 ms |
| 12 | 10 20 40 80 160 320 640 1280 2560 5120 10240 20480 | **40950 ms** |

40 s of waiting is exactly the "hang instead of fail fast" this retry exists to avoid, and it is one
edit to `CBM_SPAWN_RETRY_ATTEMPTS` away.

This bounds a single **wait** instead, at 2560 ms — the largest wait either shipping budget already
produces.

## Verification

Old vs new schedule, computed for both shipping budgets and one hypothetical:

| budget | old total | new total | |
|---|---|---|---|
| 6 | 630 ms | 630 ms | identical |
| 9 | 5110 ms | 5110 ms | identical |
| 12 | 40950 ms | 12790 ms | flattens at the ceiling |

So **nothing changes today**; a future budget increase costs 2560 ms per extra attempt instead of
doubling. The ceiling is spelled as a shift so the value cannot overflow `long` on the way to being
clamped.

`clang-format` clean. Subprocess suite green on Windows (14 passed, 17 skipped) — the POSIX retry
probes skip there, so the loop bounds themselves are CI's to confirm; this box has no POSIX-target
compiler.

Independent of #1615, which touches the `#if` above this function but not the function itself.

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) — subprocess suite green on Windows; no POSIX
      toolchain on this box
- [x] Lint passes — clang-format clean
- [x] New behavior is covered — there is no new behaviour on either shipping budget, which is the
      claim the table above verifies; `subprocess_gives_up_after_the_retry_budget` already exercises
      the exhausted path

### DIFF
--- a/src/foundation/subprocess.c
+++ b/src/foundation/subprocess.c
@@ -912,8 +912,9 @@ enum { CBM_SPAWN_RETRY = 2, CBM_SPAWN_RETRY_ATTEMPTS = 6 };
 /* Exponential backoff, doubling from 10ms: 10, 20, 40, 80, 160, 320 — ~630ms of
  * total patience on an ordinary build, and three further doublings (640, 1280,
  * 2560) to roughly 5s on a sanitized one. The waits follow from
- * CBM_SPAWN_RETRY_ATTEMPTS above rather than being listed separately here, so
- * changing the budget cannot leave this description behind.
+ * CBM_SPAWN_RETRY_ATTEMPTS above and from the per-wait ceiling at
+ * cbm_spawn_backoff, rather than being listed separately here, so changing the
+ * budget cannot leave this description behind.
  *
  * The first version waited a flat 3 x 10ms, which was enough for a momentary
  * dip and NOT enough for the real thing: a CI runner building and testing in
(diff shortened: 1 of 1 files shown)


moffermann:
`test / test-windows (CLANG64, 2/2)` failed on the first run, and it is not this diff.

The job died at 3m20s, before compiling anything:

```
ERROR: repository build/ ancestor must be absent or a real directory
```

That is `cbm_require_safe_build_dir` in `scripts/path-safety.sh` refusing the runner's `build/` state — a filesystem precondition, not a compile or a test.

Independently: the only function this PR touches, `cbm_spawn_backoff`, sits at line 969, inside the `#else /* POSIX */` block that opens at line 883. It is not compiled on Windows at all, so that leg's binary is byte-identical to main's. #1615, which touches strictly more files on the same base, passed both Windows shards.

I have no rerun rights here, so I force-pushed an amend (same content, same sign-off) to trigger a fresh run rather than leaving a red check with no explanation.

DeusData:
You are right, and the comment you are quoting is mine — I wrote both the clamp and the justification for it in `0a163d4f`/`b8c94d82`, and the justification is wrong in both halves.

I checked rather than took your word for it. `cbm_fork_with_retry` tests `attempt >= CBM_SPAWN_RETRY_ATTEMPTS` and returns **before** reaching the backoff call, and the Apple `posix_spawn` loop bounds `attempt < CBM_SPAWN_RETRY_ATTEMPTS` in its `for` condition. So `attempt` is always strictly below the budget at the call site, which makes the clamp I *kept* the dead one — exactly inverted from what my comment claims. And "the budget is the only bound needed" is the part that matters: the budget bounds the number of waits, not their size, so nothing bounded the doubling at all.

Your schedule table reproduces exactly:

```
budget  6: old=   630ms  new(cap 2560)=   630ms  identical
budget  9: old=  5110ms  new(cap 2560)=  5110ms  identical
budget 12: old= 40950ms  new(cap 2560)= 12790ms  flattened
```

Both shipping budgets are byte-identical, so this changes nothing today — and that is the point. 40 seconds of waiting is precisely the "hang instead of failing fast" that this retry exists to avoid, and it was one edit to `CBM_SPAWN_RETRY_ATTEMPTS` away. Bounding a single wait rather than the shift is the right invariant: the budget should decide *how many times* we are patient, not *how patient* any one wait is.

There is some history worth recording here, because it explains how the bad comment got written. The original code had a constant clamp; cppcheck flagged my first attempt at this function for a redundant condition, and in removing what I thought was the dead branch I replaced the live bound with a dead one and then wrote a confident comment explaining why that was correct. A clamp that cannot fire is not obviously distinguishable from one that can, and I did not check the call sites.

This is the third accurate report you have sent on code I shipped in the last two days — with #1616 (`cbm_nanosleep` losing its remaining time on `EINTR`, which truncates this very backoff) and #1615 (the sanitized-build detection having four spellings, including the ASan-only bench budget I missed). #1616 is particularly good: you separated it from the CI flakes that #1590/#1595 chased instead of attaching it to them, which is the harder and more useful call. Thank you — genuinely.

Reviewing for merge now.

DeusData:
Merged. Thank you — this corrected a bug I introduced and a comment I wrote confidently to justify it.

For the record of how it happened, since the shape is worth remembering: cppcheck flagged my first version for a redundant condition, I removed what I assumed was the dead branch, and in doing so I replaced the live bound with a dead one — then wrote a comment explaining why that was correct. A clamp that cannot fire looks identical to one that can unless you check the call sites, and I did not.

Your framing is the part I want to keep: the budget should decide *how many times* we are patient, not *how patient* any one wait is. That is the invariant, and bounding a single wait rather than the shift expresses it directly.

Both shipping budgets are byte-identical (630 ms and 5110 ms), so nothing changes today — which is exactly why it was worth taking now rather than after someone raised the budget and discovered a 41-second hang.

Your #1615 is already in as well. #1616 (`cbm_nanosleep` losing its remaining time on `EINTR`) is queued as the next item in this area — it truncates this very backoff, and your separation of it from the CI flakes that #1590/#1595 chased was the right call.

PR #1622, opened 2026-08-14
Relax ancestor ACL requirement for runtime dir (fixes #1621)

### LINKED ISSUE #1621: CBM_CACHE_DIR ignored; daemon still requires /private/tmp/cbm-daemon-<uid> on macOS
### Version

built-from-source (unknown)

### Platform

macOS (Apple Silicon)

### Install channel

### Binary variant

standard

I set CBM_CACHE_DIR to a per-user cache directory (e.g. $HOME/.cache/codebase-memory-mcp), but the daemon still fails to create a secure endpoint under /private/tmp/cbm-daemon-<uid>. The binary reports that an ancestor directory is not a usable private-directory parent.

### Reproduction

1. export CBM_CACHE_DIR="$HOME/.cache/codebase-memory-mcp"
2. mkdir -p "$CBM_CACHE_DIR" && chmod 700 "$CBM_CACHE_DIR"
3. codebase-memory-mcp

Observed: the process exits with an error about /private/tmp/cbm-daemon-<uid> not being a usable private-directory parent.
Expected: the program should honour CBM_CACHE_DIR and use that path for cache/runtime artifacts, or at least document why /private/tmp is still required and provide a way to override the runtime parent.

### Logs

```text
{"jsonrpc":"2.0","id":null,"error":{"code":-32001,"message":"secure daemon endpoint could not be created: /private/tmp/cbm-daemon-501: ancestor 'cbm-daemon-501' is not a usable private-directory parent (must be owned by you, not world-writable, and carry no allow-ACL)"}}

codebase-memory-mcp: secure daemon endpoint could not be created: /private/tmp/cbm-daemon-501: ancestor 'cbm-daemon-501' is not a usable private-directory parent (must be owned by you, not world-writable, and carry no allow-ACL)
```

```text

```

_No response_

### Confirmations

Carnival-z:
Summary: Allow ancestors (like /private/tmp on macOS) that carry allow-type ACLs so the daemon can create per-user runtime directories (/private/tmp/cbm-daemon-<uid>). The final runtime directory is still hardened: owner == current user, chmod 0700, and extended ACLs cleared.

Security: Only ancestor observation rules were relaxed for compatibility. Final runtime dir hardening (0700 + ACL clear) remains unchanged.

Testing suggestions:

macOS: verify daemon can create /private/tmp/cbm-daemon-<uid> when /private/tmp carries allow-ACLs; verify runtime dir is 0700 and ACLs cleared.
Ensure socket/lock/log files remain 0600 and ACLs cleared.
Fixes: DeusData/codebase-memory-mcp#1621

### DIFF
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -1387,9 +1387,15 @@ static bool posix_directory_parent_secure(int directory_fd) {
     struct stat status;
     if (directory_fd < 0 || fstat(directory_fd, &status) != 0 || !S_ISDIR(status.st_mode) ||
         !posix_directory_owner_trusted(status.st_uid) ||
-        !cbm_macos_extended_acl_fd_is_deny_only(directory_fd)) {
+        ) {
         return false;
     }
+    /* Note: historically this required ancestors to carry only deny-only ACLs,
+     * which caused failures on macOS where /private/tmp carries allow-type ACLs.
+     * Allowing allow-ACLs on ancestors (e.g. /private/tmp) keeps the shared
+     * tmp area usable while the final cbm-daemon-<uid> directory is still
+     * hardened (0700 and ACL-cleared) by private_directory_tree_open.
+     */
     /* #1537: this ANCESTOR check refused any group-write bit, which is the same
      * rule #1535 removed on the activation side — and the sibling that decision
      * covers but that never got changed. A group-writable ~ or ~/.cache is


DeusData:
Thank you for filing this and for taking a run at the fix — and sorry for the delay in replying.

The problem you reported is fixed, from two directions, both shipping in v0.10.5 today:

- **7f3e30e1 (#1645)** adds `CBM_RUNTIME_DIR`, so the rendezvous directory can be relocated in a shipped build. That is the missing piece behind your report: `CBM_CACHE_DIR` moves the cache and never the rendezvous, which is exactly what you found.
- **daac6bd4 (#1646)** fixes the message that made your issue so hard to act on. The ancestor walk validates the directory it is *already in*, but printed the name of the child it was about to enter — so #1621 read `ancestor 'cbm-daemon-501' is not a usable private-directory parent` when `/private/tmp` was the thing refusing. You inspected what we named, found it clean, and said so. **You were right, and we kept pointing you at the wrong directory.**

On this patch specifically, one thing to flag kindly so it isn't a mystery: as it stands it doesn't compile — removing the ACL call left the `||` behind, and cppcheck stops at `src/daemon/ipc.c:1389` with `syntax error: ||)`. That is why the whole check column went red rather than just one job.

The direction it took also isn't one we can merge as-is: relaxing the requirement for the runtime directory itself would weaken a boundary that protects the coordination endpoint. `CBM_RUNTIME_DIR` is the version of your idea we could ship — same outcome for you, without lowering the rule for everyone.

Genuinely, thank you. Your report is a large part of why the escape hatch exists at all, and the wrong-directory message would probably still be misleading people without it.

DeusData:
Closing — the problem you reported is fixed from two directions in v0.10.5: #1645 adds `CBM_RUNTIME_DIR` so the rendezvous can be relocated, and #1646 fixes the message that had been naming the wrong directory and sending you to inspect a path that was fine. You were right about that all along. Thank you for the report and for taking a run at the fix.

Issue #1620, opened 2026-08-14
Windows: index always fails with generic 'Pipeline failed' — protected-DACL cache dir breaks MoveFileExW(REPLACE_EXISTING) on this host


roosteer:
## Summary

On this Windows host, **every** `index_repository` run fails with the generic hint:

```json
{"project":"...","status":"error","hint":"Pipeline failed. Check repo_path exists and contains source files. Try mode='fast' for a quicker diagnostic run."}
```

The pipeline itself runs all passes successfully (including `gbuf.dump`), the supervised worker exits **0** (`index.supervisor.reap outcome=clean`), no error-level log line is emitted anywhere, and no `.db` is ever published. A 1-file Go repo fails identically, so it is host-specific, not repo-specific. `ui.config.write_fail ... reason=atomic_publish` appears repeatedly in the daemon log as well.

## Root cause (verified end-to-end)

The failure is an interaction between the Windows cache-DACL hardening and `MoveFileExW` on this host:

1. Since v0.9.1-rc.1, every CBM process start re-applies a **protected, owner-only DACL** to the cache directory (`win_runtime_directory_secure` in `src/daemon/ipc.c`, `cbm_windows_stamp_dir_owner` in `src/foundation/compat_fs.c`; re-applied unconditionally on each process start, not only at creation).
2. On this machine, `MoveFileExW(src, dst, MOVEFILE_REPLACE_EXISTING)` returns **ERROR_ACCESS_DENIED (5)** inside any directory whose DACL is protected (inheritance disabled), while plain rename, create and delete all work. Reproduced independently of CBM:
   - `cmd /c "move /y r2.txt r1.txt"` → `Access is denied` in the protected dir, works in `%TEMP%` and in sibling dirs with inherited ACLs.
   - Synthetic dirs: `icacls /inheritance:r` + single owner ACE → fails; `icacls /inheritance:e` → works.
3. CBM publishes the index via rename-replace (`publish_writer_output` in `src/sqlite_writer.c` → `cbm_rename_replace` → `MoveFileExW`; also `finalize_staged_generation`). The failure is silent (`CBM_PIPELINE_PERSIST_FAILED` with no log), which produces the misleading generic hint.
4. **Proof:** keeping the cache dir ACL in the healthy (inherited) state during a run (background loop re-applying `icacls /inheritance:e` while the pipeline executes) makes the exact same index succeed: `status:"indexed"`, `.db` published.

[...]

Issue #1601, opened 2026-08-13
0.10.2: cache ACL grants an ephemeral logon-session SID (S-1-5-5-X-Y) instead of the user SID, so the cache becomes unreadable after reboot (likely incomplete fix for #1351)


SQLBImhugh:
# Summary

`codebase-memory-mcp` hardens its cache files with an ACE granting the **current logon-session SID** (`S-1-5-5-X-Y`) instead of the **user account SID**. Logon-session SIDs are ephemeral — Windows mints a new one at every logon and the old one ceases to exist at reboot.

When the cache file's owner is `BUILTIN\Administrators` (which happens if the file was created by an elevated process), the logon-session ACE is the *only* user-side grant. After the next reboot that ACE is dead, leaving **no ACE matching the interactive user** and an owner a UAC-filtered token cannot exercise. The file becomes permanently inaccessible non-elevated, and the daemon can no longer start.

This is not corruption. The affected file is byte-intact and opens fine when elevated.

# Relationship to #1351

This looks like the **same defect as #1351** ("daemon leaves existing cache files unreadable (empty ACL)"), which was closed as completed on **2026-08-12 00:04 UTC**. Two reasons to think it is not fully fixed:

- I am on **0.10.2**, and the `_config.db` that broke was created at **2026-08-12 17:57:55 UTC** — after that issue was closed.
- The identical failure signature appears: `daemon.runtime_config_open_failed reason=config_db_unavailable`, with the current user unable to read `_config.db`.

The one thing I would refine is the diagnosis. **The ACL is not empty** — it is well-formed and non-trivial. The defect is that it names the *wrong principal*: an ephemeral logon-session SID rather than the user account. A fix aimed at "don't write an empty ACL" would not address this.

That distinction also reconciles the two reports. In #1351 access was lost **immediately**; for me it survived until a reboot. Both follow from the same mechanism — it depends on whether the process writing the ACL shares the interactive user's logon session. If it does, the grant appears to work until that session ends; if it does not, the user is locked out right away.

# Environment

| | |
|---|---|
| Version | `codebase-memory-mcp 0.10.2` |
| OS | Windows 11, NTFS |
| Client | GitHub Copilot CLI (stdio MCP) |
| `CBM_CACHE_DIR` | `D:\CBM` (non-default, on a second physical volume) |
| Account | Domain user, member of Administrators, **UAC enabled** (so the normal token is filtered) |

# Symptom

[...]

PR #1634, opened 2026-08-14
fix(windows): re-stamp the runtime DACL only when it is actually wrong

DeusData:
fix(windows): re-stamp the runtime DACL only when it is actually wrong

win_runtime_directory_secure called set_security_info with
PROTECTED_DACL_SECURITY_INFORMATION on every process start, whether or not
anything needed repairing. Line 4126 computes `created`, but it only guards the
ERROR_ALREADY_EXISTS check - the re-stamp itself ran unconditionally.

Two costs, both observed rather than theorised:

#1601 counted ELEVEN "Security change" USN records against a single _config.db
in one day. Windows propagates a directory's security descriptor to its
children, so a rewrite that changes nothing still churns every file underneath.

#1620 loses its atomic publish to this. MoveFileEx needs DELETE on the
destination, and a concurrent re-protect of the parent is a window in which it
can be refused - for a state that was about to be correct anyway. That reporter
proved the interaction by running a background icacls loop during indexing and
watching the identical index succeed.

The repair is what matters, not the ritual. When the owner is already the exact
current user AND the DACL already passes the private-directory check, there is
nothing to fix and the right action is to leave it alone. When it IS wrong, the
repair is byte-for-byte what it was before.

This does not fix the underlying ACL damage on already-broken installs - files
created under the pre-v0.10.3 flagless regime still carry empty DACLs and need
a child repair, which is a separate change. It stops us from making it worse and
from creating a failure window on every start.

Costs one extra GetSecurityInfo to avoid a SetSecurityInfo, which is the cheap
direction.

Builds clean; daemon_ipc 48 passed. The changed region is Windows-only, so the
real verification is the Windows CI leg.

### DIFF
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -4225,7 +4225,7 @@ static bool win_runtime_directory_secure(const wchar_t *runtime_dir) {
     DWORD secure_result = ERROR_ACCESS_DENIED;
     bool already_correct =
         valid_handle && owner_exact &&
-        win_file_security_secure(&security, directory, true, win_private_mutation_rights(), false);
+        win_file_security_secure(&security, directory, true, win_private_mutation_rights());
     if (already_correct) {
         secure_result = ERROR_SUCCESS;
     } else if (valid_handle && owner_ok) {
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -4225,7 +4225,7 @@ static bool win_runtime_directory_secure(const wchar_t *runtime_dir) {
     DWORD secure_result = ERROR_ACCESS_DENIED;
     bool already_correct =
         valid_handle && owner_exact &&
-        win_file_security_secure(&security, directory, true, win_private_mutation_rights());
+        win_file_security_secure(&security, directory, true, win_private_mutation_rights(), false);
     if (already_correct) {
         secure_result = ERROR_SUCCESS;
     } else if (valid_handle && owner_ok) {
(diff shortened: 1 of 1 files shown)


DeusData:
Converting to draft — **this is a real regression on Windows, not a flake.**

`test-windows` 1/2 and 2/2 fail with `ASSERT(locks_ready)` in `test_daemon_application.c` and `ASSERT(frontend…)` in `test_daemon_frontend.c` — the daemon cannot establish its runtime directory, which is exactly the code this PR changes.

The flaw is in my `already_correct` gate. It tests whether the DACL currently carries no untrusted mutation grants. But the unconditional call it replaces passed `PROTECTED_DACL_SECURITY_INFORMATION`, which *disables inheritance*. Those are not the same property: a freshly created directory can look clean at that instant while still being open to inherited ACEs, so skipping the stamp leaves it unprotected and the later strict validation refuses it.

I optimised away a call whose purpose I had only partly understood. The re-stamp is not merely repair — it is also what makes the DACL protected in the first place.

The underlying complaint is still real (#1601 counted eleven `Security change` USN records in a day, and #1620 loses an atomic publish to the rewrite window), so the right version of this checks *protectedness* as well as grant-cleanliness before skipping — or skips only when the descriptor is already both protected and correct. That needs the Windows leg to verify, since none of it compiles on macOS.

#1636 depends on this and goes to draft with it. Neither will be in v0.10.5; shipping a daemon-startup regression to fix a performance nuisance would be a bad trade.

DeusData:
Superseded by #1684 (rebased stack + the owner-only fast-path predicate its Windows failures were pointing at; full analysis there).

PR #1636, opened 2026-08-14
feat(windows): detect and repair cache children left unusable by the old DACL regime

DeusData:
**Stacked on #1634** — review that one first; this branch contains it.

## The gap this closes

Between v0.9.1-rc and v0.10.2 the runtime directory carried a `PROTECTED` DACL whose ACE was **not inheritable**. Windows therefore gave every file created inside it either an empty DACL or the token default (SYSTEM + `TokenOwner` + logon SID). Under an elevated token `TokenOwner` is `BUILTIN\Administrators`, so the interactive user ends up with **no durable grant at all** and the file is unreadable after the next logon.

That is #1601: `takeown` and `icacls` both fail non-elevated, and the daemon can no longer open `_config.db`.

**#1531 fixed the cause forward-only in v0.10.3** — the directory ACE is inheritable now, so newly created children are fine. But nothing repaired the children already damaged, which is why upgrading rescued nobody whose cache was written under the old regime. Their only route back was the rename-the-parent workaround #1601's reporter worked out for themselves.

This is that repair.

## Why empty-DACL detection needs its own test

`win_file_acl_secure` scans ACEs looking for untrusted mutation grants. A DACL with **zero** ACEs trivially has none — so the damage reads as *compliance*. Reusing the validator would have silently skipped exactly the files that need fixing. Hence `win_file_dacl_is_empty`.

## Deliberately narrow

- immediate children only, **no recursion**, capped at 4096
- regular files only — directories, reparse points and symlinks are skipped, not followed
- a child is touched **only** when demonstrably damaged: empty DACL, or an owner that is not the current user. A merely unusual child is left alone
- runs only after the directory itself is confirmed good, so we never repair into an unsecured parent
- failures are counted and logged (`daemon.runtime_child_acl_repaired`), never fatal — this runs inside daemon startup

Scope: it only ever touches cbm's own runtime/cache directory, which we created and own. It does not reach into user directories.

## Verification caveat, stated plainly

This is Windows-only code and **a clean macOS build proves nothing about it** — the whole `win_*` region is guarded out. The previous commit on this branch shipped a five-argument call to a four-argument function and still built green locally; I caught that by reading, not by building. The Windows CI leg is the first venue that can actually reject this, and the honest end-to-end test is a machine with a cache damaged under the old regime.

### DIFF
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -4225,7 +4225,7 @@ static bool win_runtime_directory_secure(const wchar_t *runtime_dir) {
     DWORD secure_result = ERROR_ACCESS_DENIED;
     bool already_correct =
         valid_handle && owner_exact &&
-        win_file_security_secure(&security, directory, true, win_private_mutation_rights(), false);
+        win_file_security_secure(&security, directory, true, win_private_mutation_rights());
     if (already_correct) {
         secure_result = ERROR_SUCCESS;
     } else if (valid_handle && owner_ok) {
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -4225,7 +4225,7 @@ static bool win_runtime_directory_secure(const wchar_t *runtime_dir) {
     DWORD secure_result = ERROR_ACCESS_DENIED;
     bool already_correct =
         valid_handle && owner_exact &&
-        win_file_security_secure(&security, directory, true, win_private_mutation_rights());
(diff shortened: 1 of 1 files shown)


DeusData:
Superseded by #1684, which carries these commits rebased onto current main plus the fix for the red Windows CI: the conditional-re-stamp fast path needed the adoption-level owner-only DACL predicate (full analysis in #1684). Closing in favor of the green stack.

Issue #1572, opened 2026-08-12
Python stdlib import binds to a same-named TypeScript function (cross-language, strategy=unique_name) — poisons hotspots and Louvain clusters


SQLBImhugh:
**Environment:** v0.10.2, Windows 11, daemon mode. Reproduced on a 2-file synthetic repo and on a real Python+TypeScript monorepo (~41.8k nodes / ~198.8k edges, `--mode full`).

## Symptom

A Python `from unittest.mock import patch` binds to a **TypeScript** function named `patch`, in a different language, in a different top-level package. Every `with patch(...)` in the repo is then recorded as a `CALLS` edge into that TSX function.

In my real repo this produced a 4-line React helper with **498 inbound callers**, all of them backend Python tests:

```tsx
// frontend/src/components/gates/reportgate/ReportGatePanel.tsx:151-154
function patch(changes: Partial<ReportGateState>) {
  if (!state) return
  applyEdit({ ...state, ...changes })
}
```

## Minimal reproducer

Two files, no dependencies (24 nodes / 28 edges):

`frontend/Panel.tsx`
```tsx
export function Panel() {
  function patch(changes: Partial<State>) {
    applyEdit({ ...changes })
  }
  return patch
}
```

`backend/test_thing.py`
```python
from unittest.mock import patch


def test_one():
    with patch("os.getcwd", return_value="/tmp"):
        pass


def test_two():
    with patch("os.getcwd", return_value="/tmp"):
        pass
```

```
codebase-memory-mcp cli index_repository --repo-path <repo> --mode fast
codebase-memory-mcp cli trace_path --project <p> --function-name patch \
    --direction inbound --include-evidence true --include-tests true
```

Result:

```
function: patch
direction: inbound
callers_total: 2
callers: 2  (rows: name hop strategy confidence)
  test_one 1 heuristic 0.75
  test_two 1 heuristic 0.75
```

## Root cause: the IMPORTS edge is already wrong, and it is not an ambiguity failure

The bad binding happens at the **import**, not at the call site:

```
MATCH (a)-[r:IMPORTS]->(b) RETURN a.name, a.file_path, b.qualified_name, r.local_name

  test_thing.py | backend/test_thing.py | <proj>.frontend.Panel.patch | patch
```

`from unittest.mock import patch` resolved to a project-local TSX node. The `CALLS` edges then faithfully follow that import binding.

The edge metadata is the important part:

```
[...]

Issue #1555, opened 2026-08-12
Call resolution picks the wrong target (or none) for common method names in a Kotlin monorepo


PsyTech-Rob:
**Environment:** v0.10.2, Windows 11, daemon mode, `--mode full` index. Kotlin (Spring) + TypeScript monorepo, ~20.6k nodes / ~101.7k edges.

## Symptom

`trace_path(mode=calls)` resolves call hops by name, and for method names shared by many symbols it either picks a wrong target or silently returns nothing:

1. **Wrong target across modules.** A use-case method calls `repository.inserir(...)` where the receiver is the domain port `CompromissoRepository` (constructor-injected). The trace lists the callee as **`JooqFeriadoRepository.inserir`** — an implementation of a *different* port in a *different* bounded context that happens to share the method name. The correct implementation is identifiable in the same graph via the `OVERRIDE` edge (`JooqCompromissoRepository.inserir` → `CompromissoRepository.inserir`), so the information needed to disambiguate exists.

2. **Missing callers.** The same use-case method (`executar`, also a very common name in this codebase) shows `callers_total: 0`, although a controller in the same module calls it directly one file away. Unique names resolve perfectly — an extension function with 4 call sites returned exactly those 4 callers.

## Suggestion

When `candidates > 1` for a hop:

- prefer the declared receiver/parameter type (present in source and, for implementations, derivable from `OVERRIDE`/`INHERITS`), and/or
- surface the ambiguity in `trace_path` output — the `CALLS` edge already stores `candidates` and `confidence`, but the trace renders an ambiguous hop with the same look as an exact one (related: #1542, which asks for strategy/confidence in `format=json`).

Agents treat a trace as evidence; a wrong hop that looks identical to a right hop is worse than no hop.

DeusData:
Thank you @PsyTech-Rob. The current resolver confirms the risk: after stronger lookups fail, same-name candidates are selected through import reachability and import-distance scoring. Receiver/parameter type and existing `OVERRIDE` relationships are not consistently part of that final candidate decision, so a plausible but wrong method can receive the `CALLS` edge.

The evidence-output gap is fixed on `main` by #1542, but that does not fix target selection. This remains a priority/high resolution bug under #594 and is cross-linked with the cross-language `unique_name` case in #1572 and the resulting hotspot inflation in #725.

PR #1647, opened 2026-08-14
fix(registry): drop suffix_match CALLS across language boundaries

### LINKED ISSUE #725: get_architecture hotspots report inflated fan_in that contradicts actual graph in-degree (same-name collision across languages)
**Version:** v0.8.1 (linux-amd64)

## Summary
`get_architecture` reports a hotspot `fan_in` that does not match the node's real in-degree in the graph. For short identifiers shared across languages (e.g. `commit`, `parse`, `add`), the reported `fan_in` appears to aggregate by **bare name** rather than by resolved node — producing large over-counts, and attributing the count to a node that actually has in-degree 0.

## Environment
- Mixed Python + TypeScript/JavaScript monorepo, indexed with `mode: full`.
- Two distinct symbols named `commit`: a JS arrow function `commit` (defined in a React page component) and a Python method `Store.commit`.

## Observed
`get_architecture` →
```
hotspots: [{ name: "commit", qualified_name: "…/web/src/pages/Editor.commit", fan_in: 234 }, …]
```
But that JS `commit` has **one** local call-site, and querying the graph directly shows it has **zero** inbound edges:
```
query_graph:  MATCH (n {name:'commit'})<-[r]-(c)
              RETURN n.qualified_name, count(r)
→ only the Python Store.commit, in_degree 20.
  The JS Editor.commit node returns nothing (in_degree 0).
```
So `fan_in: 234` is not reproducible from CALLS/USAGE edges.

Hotspots whose names are class-qualified Python methods matched their real in-degree correctly (a function reported with `fan_in: 46` had graph in_degree 46), so the inflation is specific to **short bare names shared across the Python/JS boundary**.

## Expected
Hotspot `fan_in` should equal the node's resolved inbound CALLS/USAGE in-degree (the value `query_graph` / `search_code` already report per node), not a count aggregated by bare name.

## Impact
`get_architecture` is the first-call "understand this codebase" tool; an inflated headline metric on the wrong node is misleading. The qualitative structure (clusters, layers, routes, languages) is accurate — only the hotspot `fan_in` figures are affected.

Possibly related to #558 (Class `in_degree`/`out_degree` = 0) — both are node-level degree-metric correctness issues.

rudi193-cmd:
## Summary
- Fixes #725: when two same-named symbols exist in different languages, `suffix_match` no longer attaches CALLS (and therefore hotspot `fan_in`) to the wrong language.
- `unique_name` (candidates == 1) is left to #1572. Receiver-type disambiguation among same-language candidates is #1555.
- JS / TS / TSX are treated as one family so `.ts` calling `.tsx` is not dropped.

## Test plan
- [x] `scripts/test.sh --suites registry` — `cross_language_suffix_match_drops_py_vs_js` PASS
- [x] `scripts/test.sh --suites pipeline` — `pipeline_cross_language_same_name_does_not_share_calls_issue725` PASS (251 passed)
- [ ] Maintainer: confirm this does not collide with the #1572 unique_name language-boundary work

### DIFF
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -635,6 +635,12 @@ static int resolve_single_call(cbm_pipeline_ctx_t *ctx, CBMCall *call,
     if (!target_node || source_node->id == target_node->id) {
         return 0;
     }
+    /* #725: suffix_match is language-agnostic and will attach a Python
+     * Store.commit() call to a JS function named commit (or a Bash main
+     * to a Python main). Drop that weak cross-language edge. */
+    if (cbm_suppress_cross_language_suffix_match(lang, target_node->file_path, res.strategy)) {
+        return 0;
+    }
     emit_classified_edge(ctx, call, source_node, target_node, &res, module_qn, imp_keys, imp_vals,
                          imp_count, tsjs_drop_plain_call);
     return SKIP_ONE;
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -2514,6 +2514,12 @@ static void resolve_file_calls(resolve_ctx_t *rc, resolve_worker_state_t *ws, CB
         }
         atomic_fetch_add_explicit(&rc->time_ns_rc_target, extract_now_ns() - _rc_t0,
                                   memory_order_relaxed);
+        if (target_node && source_node->id != target_node->id &&
+            cbm_suppress_cross_language_suffix_match(lang, target_node->file_path, res.strategy)) {
+            /* #725: same guard as pass_calls.c — do not emit a suffix_match
+             * CALLS edge across a language boundary. */
+            continue;
+        }
         if (!target_node || source_node->id == target_node->id) {
             /* HTTP/ASYNC calls to an EXTERNAL client library (`requests.get(url)`)
              * resolve to an unindexed QN (target_node == NULL), but their edge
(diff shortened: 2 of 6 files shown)


DeusData:
Thank you — and the scoping here is the part I want to call out. You fixed the cross-language case and explicitly left `unique_name` to #1572 and receiver-type disambiguation to #1555, rather than sweeping all three into one diff. That is exactly right for a graph-correctness change: each of those has a different failure mode and deserves to fail its own test.

Treating JS/TS/TSX as one family is the detail most people would have missed. Dropping a `.ts` → `.tsx` call to fix a Python-vs-JS collision would have traded one wrong answer for another, and silently — which in this project counts as a correctness regression, not a tradeoff.

Fair warning on timing: our CI is badly backed up tonight (8 runs queued, no runners allocating) because we are pushing a v0.10.5 that fixes a batch of install and startup blockers. Your PR is queued behind that. It is a graph-quality fix rather than an operability blocker, so it will land after the release rather than in it — no reflection on the change, just the order the queue forces.

I will review it properly once the backlog clears.

DeusData:
Merged (89f0cd43) — thank you! Guarding both emit paths (the header comment at lsp_resolve.h:16 confirming those are exactly the two was a nice touch), keeping unknown-extension targets, and explicitly deferring unique_name to #1572 is precisely the conservative shape we want for graph-precision changes. On your question: no collision with the #1572 work from our side — that track targets unique_name resolution and will build on top of, not against, this suffix_match guard. #725 closed.

PR #1657, opened 2026-08-15
fix(mem): honor explicit worker budget override

atirna:
## What does this PR do?

`CBM_MEM_BUDGET_MB` is resolved before a supervised worker starts, but a worker was then silently reduced to the daemon's equal-share cap. On a 376 GB host with the default four job slots, an explicit 300000 MiB override could still become about 48 GB and hit `mem.backpressure.futile`.

This keeps the aggregate cap for the fraction-derived default, while preserving a valid explicit override (which is still clamped to detected RAM).

## Verification

- `scripts/test.sh --suites mem` (51 passed)
- Added a regression guard that fails before the change when a 64 MiB explicit override is reduced to a 16 MiB worker cap.
- `scripts/lint.sh` could not run on this host because `cppcheck`, `clang-tidy`, and `clang-format` are not installed.

Related to #1654 (override path only; does not close the default-slice or crash-vs-degrade parts)

## Checklist

- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) — unavailable locally


DeusData:
Policy decision on what an explicit CBM_MEM_BUDGET_MB should mean — we're taking a middle path that differs from the PR as written, and we'd love you to carry it: the override becomes an AGGREGATE budget divided across job slots at the parent, rather than per-worker absolute. Reasoning: per-worker absolute lets N slots × override oversubscribe the host, which is precisely the #1654 OOM axis we just closed; aggregate-divided makes the override finally DO something (today it's silently eaten by the cap — your diagnosis of that silent reduction was correct and your source-string discriminator is sound) while the user's number remains a true ceiling. That likely means: keep your discriminator, divide the explicit value by the slot count where the equal-share cap is computed today, and restore the renamed policy test to assert the aggregate ceiling. Your CI red was pure GitHub infra (CodeQL init 503) — a rerun clears it. Up for the rework?

atirna:
yes, took the middle path. 
CBM_MEM_BUDGET_MB stays an aggregate, parent still divides it across the job slots, and the worker cap still applies so N workers cant oversubscribe. kept the source string as CBM_MEM_BUDGET_MB when the env discriminator fired, so a clip isnt logged as daemon_worker_cap. 
also restored resolve_budget_worker_cap_preserves_lower_user_override: 8 MiB explicit stays 8 under a 16 MiB slot share, 64 MiB clips to 16, fraction path still daemon_worker_cap.
scripts/test.sh --suites mem 52 passed.

DeusData:
Thank you for carrying the requested middle path. The latest revision now treats `CBM_MEM_BUDGET_MB` as an aggregate ceiling, preserves a lower user override, and clips a higher value to the per-slot share, which matches the maintainer decision without reopening the host-oversubscription axis from #1654.

The next step is review of the final commit and CI on that exact SHA; no further scope expansion is needed. Thank you for turning the policy correction around cleanly and restoring the test that distinguishes the lower explicit override from the clipped case.

DeusData:
Two things, and neither is a new ask.

**Your red is ours, not yours.** `test / test-windows (CLANG64, 1/2)` fails at *Step 0i: parallel suite scheduler contract* — the harness's own negative-control fixture, before a single product test runs:

```
FAIL: parallel scheduler infrastructure error: parallel scheduler cleanup failed:
hang_after_summary: suite 'hang_after_summary' leader exited leaving live descendants
```

Your diff is three files in `src/foundation/mem.*` and `tests/test_mem.c`. It cannot reach the Windows process-tree cleanup path. `shard-completeness` and `ci-ok` are just downstream of that one failure, which is why it presents as three red checks rather than one.

I traced it and filed **#1963**: the cleanup probe is given `timeout=kill_grace`, the contract fixture runs with `--kill-grace 1`, and the probe fails closed on timeout — so a one-second Windows process-enumeration call on a loaded runner is indistinguishable from a genuine surviving descendant. It has hit at least one other unrelated PR. I have re-run the failed jobs on your branch.

**And an apology for the wait.** On 18 August you were told the next step was review of the final commit, with no further scope expansion needed. That was thirteen days ago and the ball has been on our side the whole time. You turned the policy correction around in under a day, which makes the delay worse rather than better.

For the record, since it deserves saying: you were asked to abandon your original per-worker design for an aggregate-divided one, and you did it without argument and without dropping the part that mattered — the source-string discriminator that keeps a user clip from being logged as `daemon_worker_cap`, and the restored test that distinguishes a *lower* explicit override (8 MiB stays 8 under a 16 MiB slot share) from a *clipped* one (64 clips to 16). Preserving the distinguishing test through a redesign is the thing that usually gets lost.

Nothing needed from you. This goes into review on the current SHA once CI is clean.

DeusData:
Merged as `b357d090`.

Two weeks late, and the delay was entirely ours — you turned the policy correction around in under a day on 18 August and then waited. Thank you for staying with it.

The part worth recording: you were asked to abandon your original per-worker design for an aggregate-divided one, and the thing that usually gets lost in a redesign like that — the *distinguishing* test — is the thing you kept. `resolve_budget_worker_cap_preserves_lower_user_override` still separates a lower explicit override (8 MiB stays 8 under a 16 MiB slot share) from a clipped one (64 clips to the share), and the diff now also pins the fraction path separately. Those two assertions are what stop the next person collapsing the two cases back together.

Keeping `CBM_MEM_BUDGET_MB` as the source when the env discriminator fired is the other detail that mattered and was easy to drop: a clip against the user's own aggregate is a different event from a `daemon_worker_cap` rewrite of a fraction-derived default, and the logs now say which one happened.

Your original diagnosis — that an explicit override was being silently eaten by the cap — was correct throughout; the redesign only changed what the number means, not whether the bug was real.

For the record, your CI red was never yours: `test-windows` shard 1 was dying in the harness's own negative-control fixture. I root-caused it as #1963 and re-ran your jobs, which cleared with identical code.

Issue #553, opened 2026-06-22
Feature Request: i18n support for web UI (Chinese localization)


hyu18866:
## Problem

The web UI (launched via `--ui=true`) has all labels, tooltips, and navigation text hardcoded in English. For non-English-speaking developers — especially in China, where MCP tooling is rapidly growing — this creates an unnecessary friction layer.

## Proposal

Add basic i18n support to the embedded HTTP server:

1. **Language detection**: auto-detect via `Accept-Language` header, fallback to `en`
2. **Config override**: allow `codebase-memory-mcp config set ui-lang zh` to pin a language
3. **Scope**: start with UI chrome only — sidebar labels, search placeholder, tooltips, button text. Node labels remain as-is (source-derived metadata)

## Motivation

- China has one of the largest developer populations globally
- MCP tooling adoption is accelerating in the Chinese AI coding community
- The graph data is language-agnostic — the UI shouldn't be the bottleneck

## Willing to contribute

If you're open to this, I can help with the Chinese translation file — just point me to the template.

DeusData:
Thanks for the thoughtful proposal — the UI-chrome-only scope (sidebar / search / tooltips / button text, leaving source-derived node labels untouched) is exactly the right boundary. `Accept-Language` detection with an `en` fallback plus a `ui-lang` config pin is a clean approach.

Labeling this `enhancement` / `ux/behavior`. A Chinese (`zh`) translation file would be very welcome — once the chrome strings are factored into a message catalog I'll point you at the template here.

DeusData:
Thank you again for the proposal and offer to help with translation. The initial UI localization shipped in d5414111b34034910053772a18cc633ce580a2df and is included in v0.11.0: the English/Chinese message catalog is in graph-ui/src/lib/i18n.ts, and /api/ui-config applies the ui-lang override with an English fallback.

I checked the remaining detection behavior during this triage. The HTTP server currently selects Chinese when Accept-Language contains zh, rather than respecting the header preference weights; the frontend parser already handles those weights. I am keeping this open until that remaining language-negotiation behavior is resolved, rather than claiming the entire request is complete. The message catalog is also the place to propose translation corrections. Thank you for helping make the UI usable for more developers.

PR #1671, opened 2026-08-16
feat(ui): add in-UI language switcher with persisted preference

### LINKED ISSUE #1668: feat(ui): add an in-app language switcher to the graph UI
## Summary

The graph UI already ships complete, parallel English and Simplified-Chinese message catalogs in `graph-ui/src/lib/i18n.ts` (`messages.en` / `messages.zh`), and the backend already computes and exposes a `lang` field from `GET /api/ui-config` — derived from the `ui-lang` config key (`CBM_CONFIG_UI_LANG`, default `auto`) or the request's `Accept-Language` (`detect_ui_lang`).

**However, there is currently no in-app control to choose the language.** A user who wants Chinese today must either set `CBM_CONFIG_UI_LANG=zh` out-of-band (CLI / config store) or change their browser language. The UI itself offers no selector.

## Proposal

Add a small, non-breaking language switcher to the UI header with three options:

- 中文 (zh)
- English (en)
- 跟随浏览器 (auto)

Behavior:

- The selection is persisted by a new `POST /api/ui-config` endpoint that writes `CBM_CONFIG_UI_LANG` (`en` / `zh` / `auto`) via the existing `cbm_config_set` path already opened by `handle_ui_config` (`src/ui/http_server.c`).
- `GET /api/ui-config` is unchanged (still returns `lang` + `upstream_issues_url`).
- The frontend `i18n.ts` gains a `setUiLanguage()` that POSTs the choice, then updates `cachedLanguage` and notifies listeners so the whole UI re-renders immediately.
- Default remains `auto` — no behavior change for existing users.

## Scope / constraints

- Pure C for the binary change; React/TS for the frontend.
- Adds tests in `tests/test_httpd.c` (POST path) and `graph-ui/src/lib/i18n.test.ts`.
- No new dependencies, no new network egress, no hardcoded external URLs (the issues URL stays backend-served per the existing UI security audit).
- Kept under the 500-line / single-issue guidance.

## Ask

I'd like to contribute this as a PR. Posting here first per CONTRIBUTING ("open an issue first"). Happy to adjust the endpoint shape or selector placement to your preference before implementing.

zhangweildlh:
## Summary

Add an in-UI language switcher to the graph UI so users can pick **Chinese / English / Follow browser** without editing config files or restarting. The choice persists to `CBM_CONFIG_UI_LANG` and survives reloads.

## Changes

- **C (`src/ui/http_server.c`)**: `POST /api/ui-config` now persists `ui-lang` (`en`/`zh`/`auto`) via `cbm_config_set`; `GET /api/ui-config` echoes `lang_pref` (`en`/`zh`/`auto`).
- **C tests (`tests/test_httpd.c`)**: new `ui_server_ui_config_post_persists_lang` covers POST persistence and GET echo.
- **TS (`graph-ui/src/lib/i18n.ts`)**: add `UiLangPref`, `langOptionLabels`, `getUiLangPref`, `setUiLanguage` (POST + re-render).
- **TS (`graph-ui/src/App.tsx`)**: render a language `<select>` in the header.
- **TS tests (`graph-ui/src/lib/i18n.test.ts`)**: cover `setUiLanguage` and option labels.

## Testing

- Frontend: `tsc -b` passes; `vitest` 9/9 pass (incl. new switcher tests).
- C: `tests/test_httpd.c` extended; compiled and run by upstream CI (this environment has no C toolchain, so CI validates the C side).

## Notes

- **Non-breaking**: default remains `auto` (follow `Accept-Language`). Nothing changes for existing users.
- Single-issue, <500 lines, builds on the existing `CBM_CONFIG_UI_LANG` mechanism (no new config keys).
- Fixes #1668

### DIFF
--- a/graph-ui/src/App.tsx
+++ b/graph-ui/src/App.tsx
@@ -3,7 +3,16 @@ import { GraphTab } from "./components/GraphTab";
 import { StatsTab } from "./components/StatsTab";
 import { ControlTab } from "./components/ControlTab";
 import type { TabId } from "./lib/types";
-import { useUiMessages } from "./lib/i18n";
+import {
+  useUiMessages,
+  useUiLanguage,
+  getCachedLangPref,
+  getUiLangPref,
+  langOptionLabels,
+  setUiLanguage,
+  messages,
+  type UiLangPref,
+} from "./lib/i18n";
 
 const TAB_IDS: TabId[] = ["graph", "stats", "control"];
 
@@ -28,6 +37,42 @@ function routeUrl(tab: TabId, project: string | null): string {
   params.set("tab", tab);
   if (project) params.set("project", project);
   return `${window.location.pathname}?${params.toString()}${window.location.hash}`;
+}
+
+function LanguageSwitcher() {
+  const lang = useUiLanguage();
+  const [pref, setPref] = useState<UiLangPref>(getCachedLangPref());
+  const labels = langOptionLabels(lang);
+
+  useEffect(() => {
+    let cancelled = false;
+    void getUiLangPref().then((p) => {
+      if (!cancelled) setPref(p);
+    });
+    return () => {
+      cancelled = true;
+    };
+  }, []);
+
+  const onChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const next = e.target.value as UiLangPref;
+    setPref(next);
+    void setUiLanguage(next);
+  };
+
+  return (
+    <select
+      value={pref}
+      onChange={onChange}
+      aria-label={messages[lang].language.label}
+      title={messages[lang].language.label}
+      className="bg-white/[0.04] border border-border/30 rounded-md px-2 py-1 text-[12px] text-foreground/90 cursor-pointer focus:outline-none focus:ring-1 focus:ring-primary/40 hover:bg-white/[0.08] transition-colors"
+    >
+      <option value="zh">{labels.zh}</option>
+      <option value="en">{labels.en}</option>
+      <option value="auto">{labels.auto}</option>
+    </select>
+  );
 }
 
 export function App() {
@@ -100,8 +145,11 @@ export function App() {
           </nav>
         </div>
 
-        {selectedProject && (
-          <div className="flex items-center gap-2 px-3 py-1 rounded-lg bg-white/[0.04] border border-border/30">
+        <div className="flex items-center gap-3">
+          <LanguageSwitcher />
+
+          {selectedProject && (
+            <div className="flex items-center gap-2 px-3 py-1 rounded-lg bg-white/[0.04] border border-border/30">
             <span className="text-[10px] text-foreground/30 uppercase tracking-wider">
               {t.graph.selectedLabel}
             </span>
@@ -114,8 +162,9 @@ export function App() {
             >
               ×
             </button>
-          </div>
-        )}
+            </div>
+          )}
+        </div>
       </header>
 
       {/* Content */}
--- a/graph-ui/src/lib/i18n.test.ts
+++ b/graph-ui/src/lib/i18n.test.ts
@@ -1,5 +1,5 @@
-import { describe, expect, it } from "vitest";
-import { detectLanguage, messages } from "./i18n";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { detectLanguage, langOptionLabels, messages, setUiLanguage } from "./i18n";
 
 describe("i18n", () => {
   it("detects Chinese from Accept-Language and falls back to English", () => {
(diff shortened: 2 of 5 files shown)


DeusData:
Thank you for keeping this on the existing `ui-lang` setting and for adding both C and frontend regressions. The PR is focused and dependency-free. Review will concentrate on the new local POST boundary, strict value and request validation, atomic persistence, and preserving the HTTP server's host and origin defenses; changing language must not become a reason to widen general MCP-over-HTTP access. The queue is full, but the contribution is labeled and routed for normal feature review.

DeusData:
**One missing line is failing all 18 test legs.** `tests/test_httpd.c` defines `TEST(ui_server_ui_config_post_persists_lang)` but never registers it, so every platform fails identically:

```
tests/test_httpd.c:1530:1: error: unused function
    'test_ui_server_ui_config_post_persists_lang' [-Werror,-Wunused-function]
```

Add the matching `RUN_TEST(ui_server_ui_config_post_persists_lang);` in `SUITE(httpd)` — the sibling GET test is registered nearby — and the whole matrix should clear. Nothing else is red: lint, lint-mem, all three security jobs, CodeQL, DCO and all three smoke legs pass.

Worth being blunt about what that costs you right now: because the test never runs, it currently proves nothing. Your C assertions are the right ones — `ASSERT_NOT_NULL(strstr(resp, "\"lang_pref\":\"zh\""))` after the POST and again on a follow-up GET is exactly how you show persistence rather than just acceptance — they simply are not executing. The frontend test in `i18n.test.ts` *is* binding, since `setUiLanguage` and `langOptionLabels` do not exist on `main`.

**The design decision I want to credit: you reused `ui-lang` instead of adding a config key.** That is the right call and it is the one most contributors get wrong — `CBM_CONFIG_UI_LANG` already exists with default `auto`, so this becomes a second way to reach one setting rather than a competing setting. Storing to `<cache_dir>/_config.db` keeps it per-user and user-owned, so one person's choice cannot affect another.

I also checked the security boundary you were asked about, and it holds: `request_passes_http_security` runs before dispatch and enforces Host-required, `host_is_this_server` and same-origin; `route_is_protected` covers the whole `/api/` prefix so the new POST inherits the JSON content-type requirement; the 64-byte body cap and the strict `en|zh|auto` whitelist are both right; and `cbm_config_set` is a single `INSERT OR REPLACE`, so persistence is atomic rather than a file rewrite. Nothing widened.

**One unintended change to fix while you are in there.** The dispatch rewrite turns `if (is_get && path_match)` into `if (path_match) { if (is_post) … else … }`, so any *other* method — PUT, DELETE, HEAD, an arbitrary verb — now falls through to the read handler and returns 200 instead of dropping to the 404 tail. It is read-only so it is not a hole, but it is method widening nobody asked for. Please gate the else on `is_get`.

Register the test, tighten the method check, and this is ready.

DeusData:
I took the missing `RUN_TEST` line locally to unblock you rather than just re-ask for it — and it turned out to be the first of **three** layers, so I am handing you the whole chain instead of a one-line nudge. None of this was visible to you, because the test has never executed.

## Layer 1 — the missing registration (as diagnosed on 21 August)

Adding `RUN_TEST(ui_server_ui_config_post_persists_lang);` next to its two siblings in `SUITE(httpd)` clears the `-Werror,-Wunused-function` failure. **Build then succeeds.** So that diagnosis was right — it just was not the end.

## Layer 2 — the test then times out: `Content-Length` is off by one

With the test finally running:

```
FAIL tests/test_httpd.c:1552: th_status(resp) == 408, expected 200
http.request method= path= status=408 duration_ms=5003 request_bytes=0
```

The body is `{"lang":"zh"}` — **13 bytes** — but the request declares `Content-Length: 14`. The server blocks waiting for a fourteenth byte that never arrives, times out after 5 s, and logs an empty `method=`/`path=` because the request line was never completed. Changing `14` → `13` fixes it.

## Layer 3 — and then ASan aborts on a use-after-free in the handler

With the length corrected, the test reaches your new POST path and the sanitizer fires:

```
ERROR: AddressSanitizer: heap-use-after-free ... READ of size 1
SUMMARY: heap-use-after-free in dispatch_request (src/ui/http_server.c:1924)

allocated by: yyjson_read_opts   (vendored/yyjson/yyjson.c:6253)
freed by:     free               (same frame)
```

The line number is the inlined `handle_ui_config_update`. The allocation coming from `yyjson_read_opts` is the tell: **`yyjson_get_str()` returns a pointer *into* the document's own buffer, so any string you keep from the parsed body dangles the moment `yyjson_doc_free()` runs.** Read it or copy it before the free, or hold the doc alive until you are done with the value.

This repo has been bitten by the same lifetime family before — a review on another PR turned on exactly this, `yyjson_mut_obj_add_str` borrowing where `add_strcpy` copies. Worth a look at every string you take off the request body, not just the one ASan happened to reach first.

## What I did and did not do

I made both test-side changes **locally only** and did not push them. Layer 3 is production code and the fix is a design choice about lifetime that should be yours. Pushing the two test fixes alone would have swapped a compile error for a sanitizer abort on all 18 legs, which is not an improvement to hand you unannounced.

Everything else about this PR still stands as reviewed on 18 August: focused, dependency-free, and reusing the existing `ui-lang` setting instead of adding a config key — which remains the right call and the one most contributors get wrong.

Reproduction, if useful: register the test, set `Content-Length: 13`, then `build/c/test-runner httpd` on an ASan build.

DeusData:
A status update, and an apology for the timing: since the last round (the three-layer chain in the previous comment), a direction decision has landed — the graph UI is being reimplemented from the ground up, and the current React app with its `/api/ui-config` plumbing will be replaced wholesale. That makes the remaining fixes on this branch moot; there is no point putting you through them for a surface that is about to disappear.

Rather than close it, we are moving it to draft, because the requirement stays: a persisted, in-app language choice (yours and #553) is on the list for the new version, your `i18n.ts` is self-contained enough to travel, and a draft keeps the work visible so it can be rebased onto the new UI once that lands. If you would prefer to close it now and come back later, that is fine too — just say so. Thank you for the persistence on this one.

DeusData:
Closing this — and taking the 2026-09-04 offer to its conclusion rather than leaving it in draft indefinitely. Thank you for the work, and I want to be precise about why it is not landing, because none of the reasons are about its quality.

**Your central design call was right.** Reusing the existing `CBM_CONFIG_UI_LANG` setting instead of inventing a second config key was the correct instinct, and the security boundary held up under review: the 64-byte body cap, the strict `en|zh|auto` whitelist, and `cbm_config_set` being a single `INSERT OR REPLACE` so persistence is atomic. Adding a write endpoint to a local HTTP surface is exactly the kind of change that deserves suspicion, and yours survived it.

**What changed is the surface underneath you, and I verified it rather than assuming.** Against the head of the replacement work:

- `graph-ui/src/lib/i18n.ts` — **gone**
- `graph-ui/src/lib/i18n.test.ts` — **gone**
- `graph-ui/src/App.tsx` — rewritten wholesale (+4994/−344)

That is 183 of your 297 lines living in files the replacement deletes. I also checked the new frontend for calls to `/api/ui-config`: **there are none.** So the C half — the part that would survive a rebase — would land as a write endpoint with no caller.

**And the feature's premise no longer holds.** The replacement string catalog is English-only by deliberate design; there is no `zh` tree in it, and its header argues the point directly: a machine-generated `zh` tree would be *"an untranslated lie in a structure that claims to be translated."* A switcher there would switch between English and English.

I should also correct something I told you on 09-04: I said your `i18n.ts` was "self-contained enough to travel." Having now compared it against the replacement, that was too generous — its content is bound to the old UI's string inventory, which has almost no overlap with the new chrome. I would rather say so than let you rebase toward a wall on my earlier encouragement.

**#1668 and #553 stay open.** The requirement is not rejected; it is reordered. The blocker for an in-app language switcher is no longer plumbing — it is a human-written translation of the new UI's chrome strings. If you would like to own a `zh` tree for `graph-ui/src/i18n/messages.ts` once the new UI lands, that is genuinely welcome and clearly scoped, and the switcher you already designed is the right shape to sit on top of it. Say the word and I will point you at the file when it is ready.

One thing from reviewing this that outlives the PR: the `yyjson_get_str()` lifetime issue found in the third review round — the returned pointer borrows the parsed document's buffer, so keeping it past `yyjson_doc_free()` dangles — is a repo-wide hazard rather than a bug in your patch. That is ours to guard against properly, and finding it was a useful side effect of your contribution.

Thank you again, and sorry it took this shape.

PR #1680, opened 2026-08-16
fix(yaml): accept empty flow mappings in mapping bodies

### LINKED ISSUE #1673: Goose: op=mcp_install refuses any config goose itself serialized
### Version

codebase-memory-mcp 0.10.5

### Platform

Linux (x64)

### Install channel

### Binary variant

standard

The one-line installer against a working goose setup fails deterministically at the Goose step, while every other agent (Claude Code, JSON) configures successfully in the same run:

```
error: agent_config agent=Goose op=mcp_install path=/home/<user>/.config/goose/config.yaml (target: regular file, 4202 bytes)
error: one or more agent configurations failed; the published/current executable was kept, and PATH/index cleanup was not attempted
```

**Expected:** the installer appends its `codebase-memory-mcp` entry under the existing `extensions:` mapping. Constructs in *sibling* entries that the editor never touches should not veto the write.

## Root cause (verified against `main` today)

Goose serializes every stdio extension with `envs: {}` — e.g. an extension goose itself configured looks like:

```yaml
extensions:
  linux-mcp-server:
    enabled: true
    type: stdio
    name: linux-mcp-server
    cmd: uvx
    args:
    - --python
    - '3.13'
    - linux-mcp-server
    envs: {}
    env_keys: []
    timeout: 300
```

`cbm_upsert_goose_mcp` (src/cli/cli.c:3762) → `cbm_upsert_yaml_stdio_mcp(..., "extensions", ...)` → `cbm_yaml_upsert_owned_mapping_entry` → `yaml_analyze_mapping` → `yaml_validate_mapping_body`, which calls `yaml_range_has_unsupported` on **every body line of the target section** (src/cli/config_yaml_edit.c:1407). That check rejects any unquoted `{` or `}`:

```c
/* src/cli/config_yaml_edit.c:1271, yaml_range_has_unsupported */
if (c == '{' || c == '}') {
    return YAML_MATCH;   /* caller treats as YAML_ERROR */
}
```

So **any goose config containing at least one stdio extension that goose itself wrote is uneditable** by the installer. Flow *sequences* are fine (`args: []` passes — the canonical CBM entry installs cleanly on an otherwise-empty section), which isolates this to flow mappings.

Consequences worth noting:

- Pre-inserting the canonical CBM entry by hand does **not** help — section-body validation runs before the identity check, so the refusal happens regardless.
- The entry shape proposed in #728 (the Goose support request) literally includes `envs: {}`, so a fresh `goose configure`-generated config will hit this immediately.
- #1631 (Hermes `config.yaml` refusal) may be the same editor refusing a real-world file; this report pins one concrete, reproducible trigger for the shared code path.

## Bisect evidence (sandboxed `$HOME`, so the real config was never at risk)

| Config shape | Result |
|---|---|
| Real goose config (4 stdio extensions with `envs: {}`) | FAIL |
| Same, all stdio extensions removed | OK |
| Minimal single-entry config **with** `envs: {}` (repro below) | FAIL |
| Identical minus the one `envs: {}` line | OK |
| Block-style `args:` converted to flow style `args: ["-c", ...]` | still FAIL |
| `cwd: null` / `bundled: null` lines removed | still FAIL |
| Canonical CBM entry pre-inserted alongside the stdio entry | still FAIL |

## Workaround (verified end-to-end on 0.10.5)

Delete the `envs: {}` lines (semantically identical for goose — an absent `envs` key deserializes to an empty env map), then re-run the installer; it completes, writes the entry, and the server passes a stdio MCP handshake. Caveat: goose will re-serialize `envs: {}` the next time it rewrites `config.yaml` (e.g. via `goose configure`), so this recurs on every future `codebase-memory-mcp update` until fixed.

## Suggested fix direction

The editor only upserts its own entry under `extensions:` — it does not need to interpret flow mappings inside sibling entries' values. #1639 already established the precedent for this class of fix (interior `*`/`&` treated as text rather than anchors). Alternatively/additionally, carrying the refusal reason out instead of collapsing to `-1` (as #1630 asks, and as `cbm_daemon_ipc_validation_detail()` already does) would have made this a one-minute diagnosis.

### Reproduction

No indexing or real code involved — this is a pure installer/config-editor bug, and the repro uses only dummy content.

1. Config: a minimal fake goose home containing one stdio extension with the `envs: {}` line goose itself writes:

```bash
CBM="$(command -v codebase-memory-mcp)"          # real binary location
mkdir -p "$HOME/cbm-repro/.config/goose"
cat > "$HOME/cbm-repro/.config/goose/config.yaml" <<'EOF'
GOOSE_TELEMETRY_ENABLED: true
extensions:
  demo:
    enabled: true
    type: stdio
    cmd: /bin/true
    args:
    - --serve
    envs: {}
    env_keys: []
    timeout: 300
GOOSE_THINKING_EFFORT: max
EOF
```

2. Command (Goose detection triggers via the `~/.config/goose` directory; `$HOME` override sandboxes everything else):

```bash
HOME="$HOME/cbm-repro" "$CBM" install -y --force --dir="$(dirname "$CBM")"
```

3. Result: `error: agent_config agent=Goose op=mcp_install ...` and activation aborts (non-zero exit).

4. Expected: install completes. Proof that the single `envs: {}` line is the entire trigger:

```bash
sed -i '/^    envs: {}$/d' "$HOME/cbm-repro/.config/goose/config.yaml"
HOME="$HOME/cbm-repro" "$CBM" install -y --force --dir="$(dirname "$CBM")"
# → "Install complete."
```

Both directions verified today on 0.10.5, Linux x86_64.

### Logs

```text
codebase-memory-mcp installer
  os:      linux
  arch:    amd64
error: agent_config agent=Goose op=mcp_install path=/home/<user>/.config/goose/config.yaml (target: regular file, 4202 bytes)
error: one or more agent configurations failed; the published/current executable was kept, and PATH/index cleanup was not attempted
error: activation stopped after one or more agent configuration or cleanup operations failed; the published/current executable was kept, and configuration changes that completed may remain. Please restart your coding-agent sessions after resolving the errors above.
Detected agents: Claude-Code Goose
```

```text

```

_No response_

### Confirmations

pcristin:
## What does this PR do?

Fixes #1673.

Goose serializes stdio extension siblings with an empty flow mapping such as
`envs: {}`. The YAML editor rejected every unquoted brace in mapping bodies,
so `op=mcp_install` could not update configuration that Goose itself had
written.

This change adds a narrow mapping-body exception for an exact empty mapping
value (`{}`). The key and prefix still pass through the existing fail-closed
scanner, and non-empty flow mappings remain unsupported.

Regression coverage verifies that:

- a realistic Goose sibling containing `envs: {}` is preserved byte-for-byte
  while the owned entry is appended;
- `envs: {FOO: bar}` is still rejected without changing the file; and
- unsupported key syntax such as `<<: {}` is still rejected without changing
  the file.

Validation:

- `ASAN_OPTIONS=detect_leaks=0 scripts/test.sh --suites config_yaml_edit`
  (47 passed)
- `ASAN_OPTIONS=detect_leaks=0 scripts/test.sh --suites cli` (278 passed)
- `ASAN_OPTIONS=detect_leaks=0 scripts/test.sh`
  (7,498 passed, 0 failed, 5 permitted platform skips)
- `clang-format --dry-run --Werror src/cli/config_yaml_edit.c tests/test_config_yaml_edit.c`
- `git diff --check upstream/main..HEAD`

`scripts/lint.sh --ci` still reports pre-existing findings in untouched files:
formatting in `src/mcp/mcp.c` and `src/pipeline/pipeline_incremental.c`, plus a
cppcheck style note in `src/store/store.c`.

## Checklist

### DIFF
--- a/src/cli/config_yaml_edit.c
+++ b/src/cli/config_yaml_edit.c
@@ -1266,8 +1266,8 @@ static int yaml_find_comment(const char *data, size_t start, size_t end, size_t 
  * The test is positional and deliberately conservative: an indicator is only
  * honoured as one when nothing has appeared before it in the value. `key: *a`
  * is still an alias and still refused; `key: 2 * 3` and `key: a*b` are text.
- * `{`/`}` keep their existing treatment — empty flow mappings are a separate
- * change with its own semantics. */
+ * `{`/`}` keep their existing treatment here; mapping-body validation admits
+ * exact empty flow mappings through a context-specific exception. */
 static int yaml_range_has_unsupported(const char *data, size_t start, size_t end) {
     char quote = '\0';
     /* Last significant character seen, so an indicator can be judged by what
@@ -1387,6 +1387,19 @@ static int yaml_tail_is_explicit_empty_mapping(const char *data, size_t colon, s
     return 0;
 }
 
+static int yaml_mapping_body_line_has_unsupported(const yaml_doc_t *doc, const yaml_line_t *line) {
+    size_t start = line->start + line->indent;
+    size_t colon = 0U;
+    bool empty_mapping = false;
+    if (yaml_find_mapping_colon(doc->data, start, line->text_end, &colon) == 0 &&
+        yaml_tail_is_explicit_empty_mapping(doc->data, colon, line->text_end, &empty_mapping) ==
+            0 &&
+        empty_mapping) {
+        return yaml_range_has_unsupported(doc->data, start, colon + YAML_UNIT);
+    }
+    return yaml_range_has_unsupported(doc->data, start, line->text_end);
+}
+
 static int yaml_value_starts_multiline(const char *data, size_t colon, size_t end) {
     size_t pos = yaml_skip_spaces(data, colon + YAML_UNIT, end);
     return pos < end && (data[pos] == '|' || data[pos] == '>');
@@ -1404,7 +1417,7 @@ static int yaml_validate_mapping_body(const yaml_doc_t *doc, size_t first_line, 
             continue;
         }
         if (line->indent < YAML_ENTRY_INDENT || (line->indent & YAML_UNIT) != 0U ||
-            yaml_range_has_unsupported(doc->data, line->start + line->indent, line->text_end)) {
+            yaml_mapping_body_line_has_unsupported(doc, line)) {
             return YAML_ERROR;
         }
         size_t colon = 0U;
--- a/tests/test_config_yaml_edit.c
+++ b/tests/test_config_yaml_edit.c
@@ -812,6 +812,107 @@ TEST(config_yaml_edit_goose_extensions_preserve_siblings) {
     PASS();
 }
 
+TEST(config_yaml_edit_goose_accepts_empty_flow_mapping_in_sibling_issue1673) {
+    const char *initial = "extensions:\n"
+                          "  demo:\n"
+                          "    enabled: true\n"
+                          "    type: stdio\n"
+                          "    cmd: /bin/true\n"
+                          "    args:\n"
+                          "    - --serve\n"
+                          "    envs: {}\n"
+                          "    env_keys: []\n"
+                          "    timeout: 300\n"
+                          "GOOSE_THINKING_EFFORT: max\n";
+    const char *block = "    type: stdio\n"
+                        "    cmd: \"/opt/codebase-memory-mcp\"\n"
+                        "    args: []\n"
+                        "    enabled: true\n";
+    const char *expected = "extensions:\n"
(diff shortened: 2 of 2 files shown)


DeusData:
Merged — thank you for a model contribution! The mapping-body-scoped exception is exactly the right shape for this editor (narrow, positional, fail-closed on the key side), and the byte-identity regression tests for the reject paths made this an easy review: verified end-to-end here against a Goose-serialized config (`envs: {}`) — `op=mcp_install` now writes the entry cleanly where v0.10.4/5 refused it. Much appreciated!

Issue #1669, opened 2026-08-16
Cross-file LSP pass regressed 16.6x (21.5s → 358s on java) and is 78% of the 0.10.x indexing slowdown


DeusData:
### Version

0.10.5 (regression against 0.9.0)

### Platform

macOS arm64 (48 GB, 18 CPU); measured with the shipped release binaries of both versions

### Summary

The cross-file LSP pass regressed **16.6x** between v0.9.0 and v0.10.5 — from 21.5 s to ~358 s on a Java corpus — and it accounts for roughly **78% of the total java indexing regression**. It contributes 25,885 edges, about **0.5%** of the graph, so the cost/benefit is roughly 14 ms per edge.

### Measurement

Corpus: 46,477-file Java tree. Shipped binaries, same host, isolated `HOME`/`CBM_CACHE_DIR` per run, sequential (no concurrent load).

| | v0.9.0 | v0.10.5 |
|---|---|---|
| baseline | 99.8 s | ~530 s |
| `CBM_DISABLE_LSP_CROSS=1` | 78.3 s | ~172 s |
| **cross-LSP pass alone** | **21.5 s** | **~358 s (16.6x)** |
| rest of the pipeline | 78.3 s | ~172 s (2.2x) |

Reproducibility: the 0.10.5 cuts were run twice — `xlsp-off` 175.0 s and 169.1 s; `lsp-off` 522.0 s and 525.2 s. Baselines 477.1 s (first sweep) and 534.5 s (recheck), so ~10% run-to-run variance, well inside the effect size.

Edge counts barely move, which is what makes the cost hard to justify as-is:

| | edges |
|---|---|
| 0.10.5 baseline | 5,646,309 |
| 0.10.5 `CBM_DISABLE_LSP_CROSS=1` | 5,620,424 |

Note also that `CBM_LSP_DISABLED=1` (disabling LSP entirely) barely changes wall time (522 s vs 530 s baseline) — so the cost is **specifically** the cross-file pass, not per-file LSP.

### Reproduction

```sh
for cut in baseline xlsp-off; do
  W=$(mktemp -d)
  ENVV=""; [ "$cut" = "xlsp-off" ] && ENVV="CBM_DISABLE_LSP_CROSS=1"
  time env HOME="$W" CBM_CACHE_DIR="$W/.cache" $ENVV ./codebase-memory-mcp cli index_repository \
    --repo-path /path/to/java-corpus --mode full --name "$cut" --persistence true
done
```

[...]

PR #1681, opened 2026-08-16
perf(lsp): eliminate the cross-LSP O(n²) — shared Java registry, own-file overlay, complexity guard (#1669)

DeusData:
## What does this PR do?

Fixes the dominant 0.9.0 → 0.10.x indexing slowdown (#1669): cross-file LSP was O(files × corpus_defs) for Java/Kotlin, plus ships the diagnostics that found it and a deterministic complexity-guard test suite that prevents the class from returning.

Five commits, each green:

1. `feat(diagnostics)` — scaling probe + `cross_lsp_cost`/`scan_cost` telemetry: normalized per-file numbers that make a superlinear pass visible from ONE run's log instead of an 11-corpus two-binary A/B.
2. `style` — clang-format of those lines.
3. `feat(diagnostics)` — `perfile_registry` (defs_per_file vs defs_total): the counter that proved the filter reduces by constant factor, not constant size.
4. `perf(lsp)` — **the fix**: shared Java+Kotlin cross-registry built once (two-phase: types → finalize → funcs, because pre-finalize type lookups are linear scans — a one-pass build cost ~300 s), resolved per file through an overlay of **own-file** defs (module/namespace scope re-imports the quadratic on large packages — proven by the suite below). Wired into both the full and incremental pipelines.
5. `test(complexity)` — the guard suite: replicated-module + growing-shared-package corpora, gates on deterministic work-counter RATIOS (never wall time), ~2 s runtime, auto-joins CI sharding. RED at ratio 4.00 on the pre-fix tree, GREEN at 2.00 after; it also rejected an earlier module-scoped overlay design before it shipped.

## Measured (elasticsearch corpus, 46,477 files, same host, this session's baselines)

| | v0.10.5 | this PR | v0.9.0 |
|---|---|---|---|
| wall (cold) | 419.9 s | **85.6 s (4.9×)** | 61.5 s |
| wall (warm daemon) | — | **83.7 s** | — |
| cross-LSP CPU | 6,195,780 ms | **52,513 ms (118×)** | 110,344 ms |
| us_per_file_per_kdef | 300 | **2** | — |
| nodes | 693,100 | **693,100** | — |
| edges | 5,646,235 | 5,649,949 (+0.066 %) | — |

Full 11-corpus validation (warm daemon): java 477 → 86 s; every language improves vs v0.10.5 (total 2,370 → 1,774 s); small corpora halve. Node counts stable on clean corpora.

**Edge delta named**: +0.066 % on java — the shared base resolves cross-package targets the old per-file namespace/import filter could not see, plus source-order independence from the two-phase build. Run-to-run jitter ±54 edges on this corpus (known MT jitter class).

**Explicitly not fixed here** (separate root cause — C# already had a shared registry and did not move): the csharp 2.6×-vs-0.9.0 regression, tracked as the next perf target. Also recorded-and-rejected attempts with numbers (tail-index ×2, batched registration) so they aren't re-tried.

## Verification

- Full macOS suite green + full Linux (Colima arm64, ASan+LSan) leg green on this exact tree.
- Complexity suite reproduce-first: RED on baseline (ratio 4.00) → GREEN with fix (2.00) → re-verified after every later edit.
- `mem`/`diagnostics` suites carry the new deterministic tests (probe math is a pure function — no timing assertions anywhere).
- lint-ci clean (cppcheck + pinned clang-format + NOLINT).
- Windows: local UTM VM was infra-blocked at verification time (guest boots, no network — being investigated separately); the change is platform-independent C, and this PR's CI Windows legs are the gate.

## Checklist

- [x] Tests pass locally (full macOS + Linux legs)
- [x] New behavior is covered by a test (reproduce-first; RED→GREEN proven, including against a wrong fix design)

Refs #1669.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -505,6 +505,7 @@ TEST_FOUNDATION_SRCS = \
     tests/test_workspace.c \
     tests/test_platform.c \
     tests/test_diagnostics.c \
+    tests/test_complexity.c \
     tests/test_dump_verify.c \
     tests/test_subprocess.c \
     tests/test_private_file_lock.c \
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -713,9 +713,9 @@ BUILD_DIR = build/c
 # Grammar + tree-sitter runtime: compiled without -Werror (upstream code has warnings)
 GRAMMAR_CFLAGS = -std=c11 -D_DEFAULT_SOURCE -O2 -w -I$(CBM_DIR) -I$(TS_INCLUDE) -I$(TS_SRC)
 GRAMMAR_CFLAGS_TEST = -std=c11 -D_DEFAULT_SOURCE -g -O1 -w -I$(CBM_DIR) -I$(TS_INCLUDE) -I$(TS_SRC) \
-                      $(SANITIZED_DEFINE) $(SANITIZE)
+                      -DCBM_ENABLE_TEST_SEAMS=1 $(SANITIZED_DEFINE) $(SANITIZE)
 GRAMMAR_CFLAGS_TSAN = -std=c11 -D_DEFAULT_SOURCE -g -O1 -w -I$(CBM_DIR) -I$(TS_INCLUDE) -I$(TS_SRC) \
-                      -DCBM_SANITIZED_BUILD=1 $(TSAN_SANITIZE)
+                      -DCBM_ENABLE_TEST_SEAMS=1 -DCBM_SANITIZED_BUILD=1 $(TSAN_SANITIZE)
 
 # Object files for grammars + ts_runtime + lsp_all + preprocessor
 GRAMMAR_OBJS_TEST = $(patsubst $(CBM_DIR)/%.c,$(BUILD_DIR)/%.o,$(GRAMMAR_SRCS))
--- a/internal/cbm/extract_unified.c
+++ b/internal/cbm/extract_unified.c
@@ -1646,6 +1646,27 @@ static bool is_actual_import_boundary(CBMExtractCtx *ctx, TSNode node, const CBM
     char *name = ts_node_is_null(head) ? NULL : cbm_node_text(ctx->arena, head, ctx->source);
 
     switch (ctx->language) {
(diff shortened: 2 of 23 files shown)


DeusData:
## Update: the regression program continued — C# recovered, TS partially

**C# (dotnet/runtime corpus, 58,656 files) — acceptance met decisively** (commit a4c0ffbc):

| | v0.9.0 | v0.10.5 | now |
|---|---|---|---|
| wall | 444.3 s | 1210.8 s | **449.9 s (1.01×)** |
| edges | 3.95 M | 4.29 M (deflated) | **5.87 M (+48.6 %)** |
| per-edge | 112 µs | 282 µs | **77 µs — 32 % better than v0.9.0** |

Four measured changes: two-phase cs registry build (prepare 280→140 s), the **namespace/using_statement import-context bug** (every namespaced C# file ran inside_import → suppressed usages + ancestor-walk tax; one 147 KB JIT file went 490→6.8 s; restores +1.58 M edges), the usages walker's ancestry counters (6→100 ts_node_parent growth since 0.9.0 reverted to O(1) gates), and registry short-name indexes for cs_lookup_extension + cs_resolve_type_name step 9 (builder 140,095→500 ms; resolve cross-LSP CPU 5.5 M→317 k ms).

**TypeScript (commits a4126427, f95fe55b): 44.1 → 37.5 s** (v0.9.0: 17.9). Coverage shadow-graph rebuild deduped (9.16→2.92 s — it upserted every directory per failure row), TS export re-export boundary fixed (+3,954 restored usage edges), eval-budget hardening. The remaining TS gap is attributed (extract +8.6 s incl. a spread-bomb file class with an open budget-arming lead; persist +5.2 s = FTS + coverage + rows; ~6 s startup) and not yet at the acceptance bar — tracked.

New telemetry shipped along the way: `publish.timing.coverage` sub-blocks and `lsp_cross_prepare.builders` per-language split — both existed as single opaque numbers that could not name their culprit.

Full 11-corpus validation will be re-run on the final tree (the usages-walker change touches every language's extract), and the linux/C residual attribution is queued.

PR #1695, opened 2026-08-17
fix(test): isolate harness daemon runtimes

### LINKED ISSUE #1691: Test scripts share the account daemon runtime and can stop active MCP sessions
### Version

Originally observed with `codebase-memory-mcp 0.10.2`. Reproduced from source on current `main` at `c1a5de3bda2f013b93a2beb66905de17e39a3e11` (`codebase-memory-mcp dev`).

### Platform

macOS (Apple Silicon)

### Install channel

Built from source for the reproduction. The original recurring failures used the standard binary from the GitHub release archive.

### Binary variant

standard

Some test and security scripts launch the product binary without assigning a private daemon runtime. They therefore use the same account-level rendezvous as active MCP sessions.

The directly reproduced case is `scripts/security-install.sh`. The script creates a temporary `HOME`, but then runs a real `install -y` while inheriting the caller's daemon runtime. The install activation guard intentionally stops every client in that runtime. If `make security` runs while an editor or coding agent is using Codebase Memory, the test can terminate those MCP sessions.

The test should write only inside its sandbox and must not connect to, stop, or replace an account daemon used by another process.

### Reproduction

This reproduction uses a private sentinel runtime, so it does not touch the normal account daemon.

```bash
git clone https://github.com/DeusData/codebase-memory-mcp.git
cd codebase-memory-mcp
git checkout c1a5de3bda2f013b93a2beb66905de17e39a3e11
make -f Makefile.cbm cbm

binary="$PWD/build/c/codebase-memory-mcp"
repro_root="$(mktemp -d /private/tmp/cbmri.XXXXXX)"
mkdir -m 700 "$repro_root/runtime" "$repro_root/cache"
export CBM_RUNTIME_DIR="$repro_root/runtime"
export CBM_CACHE_DIR="$repro_root/cache"

mkfifo "$repro_root/stdin"
exec 3<>"$repro_root/stdin"
"$binary" <&3 >"$repro_root/client.out" 2>"$repro_root/client.err" &
client_pid=$!

printf '%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"runtime-isolation-repro","version":"1.0"}}}' \
  >&3

for _ in {1..100}; do
  grep -q '"id":1' "$repro_root/client.out" && break
  sleep 0.1
done

"$binary" daemon status
scripts/security-install.sh "$binary"
"$binary" daemon status

exec 3>&-
wait "$client_pid" || true
"$binary" daemon stop >/dev/null 2>&1 || true
rm -rf "$repro_root"
```

Before the install audit:

```text
daemon: active (session-managed)
  committed clients: 1
```

The audit then prints:

```text
=== Install output audit passed ===
```

The MCP frontend exits with code 0 and the final status is:

```text
daemon: not running
```

The daemon log records:

```text
level=info msg=daemon.runtime_stopping reason=activation_shutdown
level=info msg=daemon.activation_shutdown action=install active_clients=1 active_connections=1
```

As a control, I ran the same install audit with a different `CBM_RUNTIME_DIR` and `CBM_CACHE_DIR`. The sentinel remained connected in runtime A with the same daemon PID.

### Other launch sites to audit

Source inspection found the same missing runtime boundary in these product-binary test paths:

- `scripts/security-fuzz.sh` sets `HOME` and `CBM_CACHE_DIR`, but not `CBM_RUNTIME_DIR`.
- `scripts/security-fuzz-random.sh` and `scripts/security-network.sh` launch the binary without a private runtime.
- `tests/test_parent_watchdog.sh` and `tests/test_worker_watchdog.sh` launch a development binary without a private runtime.
- `tests/test_worker_error_response.sh` sets separate cache directories, but its CLI and `daemon stop` commands still share the account rendezvous.
- `scripts/test.sh` invokes the watchdog and worker scripts as part of its normal full run.

Only `security-install.sh` is required for the reproduction above. The other paths are listed because they cross the same runtime boundary and should be covered by the fix.

### Proposed fix

1. Add a small shared test helper that creates a short, owner-only runtime parent and cache directory.
2. Set both `CBM_RUNTIME_DIR` and `CBM_CACHE_DIR` before every test or security script launches the product binary.
3. Isolate the standalone scripts themselves, rather than relying only on `scripts/test.sh` or the Makefile target, because contributors can run each script directly.
4. Keep processes within one test in the same private runtime, but give concurrent test invocations different runtimes.
5. Correct the `security-install.sh` comment that calls this a dry-run. The script currently performs a sandboxed real install, which is useful for auditing writes.

### Suggested regression

Start a sentinel MCP client in runtime A, run each product-binary test entry point in runtime B, and assert that:

- the sentinel remains connected;
- the runtime A daemon PID and generation do not change;
- no `activation_shutdown` is recorded in runtime A;
- two concurrent security runs use separate rendezvous directories;
- existing test exit codes and assertions are unchanged.

### Logs

The relevant sanitized lines are included in the reproduction. No project indexing is required.

Not applicable. This is a deterministic runtime-isolation failure, not a memory or performance issue.

Not applicable.

### Confirmations

- [x] I searched existing issues and did not find a duplicate. Issue #1664 concerns forwarding daemon environment variables into generated Codex configuration; it does not cover test-runtime isolation.
- [x] The reproduction uses only this public repository and temporary test directories.

astandrik:
## What does this PR do?

Fixes #1691

Several test and security scripts inherited the caller's daemon runtime. That is unsafe for `security-install.sh`, which performs a real install inside a temporary home. Its activation step could stop the account daemon and disconnect MCP sessions used by an editor or coding agent.

This PR adds a shared test helper that creates a private runtime and cache before a harness invokes the product binary. It replaces inherited paths, uses protected temporary roots on each supported platform, and limits cleanup to the root created by that harness. The helper is wired into the eight security, watchdog, worker, and hook entry points covered by the issue.

The change does not touch product code, public APIs, Makefiles, workflows, dependencies, or configuration formats.

## Failure before

The original reproduction started a sentinel MCP client in runtime A, then ran `security-install.sh` with the same inherited runtime. The install activation stopped daemon A, the frontend exited, and the daemon log recorded `activation_shutdown`.

## Result after

The same reproduction now gives the install audit a separate runtime B. The sentinel answers a ping after the audit, daemon A keeps the same PID, and its logs contain no `activation_shutdown`. Runtime B is removed during cleanup.

## Reproduction

The runnable fail-before reproduction is in #1691 under "Reproduction". It builds the product binary and starts a sentinel MCP client in a private runtime, so it does not touch the account daemon.

On this PR head, run the tracked pass-after contract:

```bash
bash tests/test_runtime_isolation_contract.sh
```

It verifies that inherited runtime and cache paths are replaced, parallel runs receive different roots, cleanup leaves the caller's runtime untouched, creation failures stop the harness before the binary starts, and `security-install.sh` uses a private HOME, cache, and runtime.

The live-sentinel script from #1691 can also be run against this branch. The sentinel remains connected after the install audit, daemon A keeps the same PID, and runtime A records no `activation_shutdown`.

## Validation

- `bash -n` on the helper, changed harness scripts, and regression tests
- `bash tests/test_runtime_isolation_contract.sh`
- `bash tests/test_security_fuzz_harness.sh`
- seam-enabled build plus parent watchdog, worker watchdog, worker error, and hook conflict tests
- `make -f Makefile.cbm security`
- `scripts/test.sh`: 7,523 passed, 0 failed, 5 skipped on the remote Linux builder
- exact sentinel pass-after reproduction on macOS

The full `scripts/lint.sh` run on the remote builder reaches the existing repository-wide `clang-tidy` baseline in unchanged C and header files. PR CI runs `scripts/lint.sh --ci`, which gates this test-only diff with cppcheck and clang-format.

## Checklist

- [x] Tests pass
- [ ] Lint passes locally. See the `clang-tidy` baseline note above; the PR CI lint subset remains authoritative.
- [x] New behavior is covered by a reproduce-first regression

### DIFF
--- a/scripts/security-fuzz-random.sh
+++ b/scripts/security-fuzz-random.sh
@@ -27,8 +27,11 @@ fi
 
 echo "=== Fuzz Testing ($DURATION seconds) ==="
 
-FUZZ_TMPDIR=$(mktemp -d)
-trap 'rm -rf "$FUZZ_TMPDIR"' EXIT
+# shellcheck source=test-runtime.sh
+source "$(dirname "${BASH_SOURCE[0]}")/test-runtime.sh"
+cbm_test_runtime_init
+FUZZ_TMPDIR="$CBM_TEST_RUNTIME_ROOT"
+trap 'cbm_test_runtime_cleanup "$BINARY"' EXIT
 
 CRASHES=0
 ITERATIONS=0
--- a/scripts/security-fuzz.sh
+++ b/scripts/security-fuzz.sh
@@ -22,11 +22,14 @@ PASS=0
 TOTAL=0
 
 # Temp directory for input files (avoids pipe/stdin issues with timeout)
-FUZZ_TMPDIR=$(mktemp -d)
-trap 'rm -rf "$FUZZ_TMPDIR"' EXIT
+# shellcheck source=test-runtime.sh
+source "$(dirname "${BASH_SOURCE[0]}")/test-runtime.sh"
+cbm_test_runtime_init
+FUZZ_TMPDIR="$CBM_TEST_RUNTIME_ROOT"
+trap 'cbm_test_runtime_cleanup "$BINARY"' EXIT
 FUZZ_HOME="$FUZZ_TMPDIR/home"
-FUZZ_CACHE="$FUZZ_TMPDIR/cache"
-mkdir -p "$FUZZ_HOME" "$FUZZ_CACHE"
+mkdir "$FUZZ_HOME"
+export HOME="$FUZZ_HOME"
 
 # Helper: send a payload to the MCP server and check it doesn't crash.
 # Uses temp file + perl alarm for portable timeout (works on macOS + Linux).
@@ -49,11 +52,9 @@ test_payload() {
     # Run with 10s timeout: GNU timeout → perl alarm fallback
     local ec=0
     if command -v timeout &>/dev/null; then
-        HOME="$FUZZ_HOME" CBM_CACHE_DIR="$FUZZ_CACHE" \
-            timeout 10 "$BINARY" < "$tmpinput" > "$tmpoutput" 2>&1 || ec=$?
+        timeout 10 "$BINARY" < "$tmpinput" > "$tmpoutput" 2>&1 || ec=$?
     else
-        HOME="$FUZZ_HOME" CBM_CACHE_DIR="$FUZZ_CACHE" \
-            perl -e 'alarm(10); exec @ARGV' -- "$BINARY" \
+        perl -e 'alarm(10); exec @ARGV' -- "$BINARY" \
             < "$tmpinput" > "$tmpoutput" 2>&1 || ec=$?
     fi
 
@@ -88,8 +89,7 @@ test_invalid_index_interactive() {
     local tmpoutput="$FUZZ_TMPDIR/output_${TOTAL}.jsonl"
     local ec=0
 
-    HOME="$FUZZ_HOME" CBM_CACHE_DIR="$FUZZ_CACHE" \
-        python3 "$(dirname "$0")/test_mcp_interactive.py" "$BINARY" \
+    python3 "$(dirname "$0")/test_mcp_interactive.py" "$BINARY" \
         --scenario invalid-index --repo-path /nonexistent/path/abc123 \
         --response-timeout 45 --exit-timeout 15 > "$tmpoutput" 2>&1 || ec=$?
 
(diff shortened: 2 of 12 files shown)


DeusData:
Thank you for preserving one clear invariant across the changed harnesses: no test process may inherit or affect the account daemon runtime. The shared helper plus a deterministic sentinel contract is a coherent slice even though it touches several scripts.

Because this changes executable security and test harnesses, it will receive the full script and filesystem-boundary review once the PR leaves draft. The initial patch scan found no new external dependency, outbound network path, or product-code change. Please keep the broader smoke, soak, profiling, and benchmark audit in #1696 rather than expanding this PR. The queue is full, but the draft is labeled and routed as a high-priority isolation fix.

PR #1702, opened 2026-08-18
fix(registry): drop unique_name CALLS across language boundaries

### LINKED ISSUE #1572: Python stdlib import binds to a same-named TypeScript function (cross-language, strategy=unique_name) — poisons hotspots and Louvain clusters

rudi193-cmd:
## What does this PR do?

Follow-on to #1647 / #725. That PR dropped `suffix_match` CALLS when caller language ≠ target file language and left `unique_name` (`candidates == 1`) for this issue.

Python `from unittest.mock import patch` was binding to a unique TSX `function patch` via `unique_name`. The same helper now drops `unique_name` under the same rules: JS/TS/TSX stay one family; `same_module` / `import_map` / `lsp_*` stay.

Not in this PR: treating `unittest.mock` as an external specifier, IMPORTS-edge cleanup, #1555 receiver disambiguation, or #1128.

## Checklist

- [x] Tests pass locally (`scripts/test.sh --suites registry,pipeline`)

Fixes #1572

### DIFF
--- a/src/pipeline/pass_calls.c
+++ b/src/pipeline/pass_calls.c
@@ -635,9 +635,10 @@ static int resolve_single_call(cbm_pipeline_ctx_t *ctx, CBMCall *call,
     if (!target_node || source_node->id == target_node->id) {
         return 0;
     }
-    /* #725: suffix_match is language-agnostic and will attach a Python
-     * Store.commit() call to a JS function named commit (or a Bash main
-     * to a Python main). Drop that weak cross-language edge. */
+    /* #725/#1572: suffix_match and unique_name are language-agnostic and
+     * will attach a Python Store.commit() call to a JS function named
+     * commit, or a Python `with patch(...)` to a unique TSX `patch`. Drop
+     * those weak cross-language edges. */
     if (cbm_suppress_cross_language_suffix_match(lang, target_node->file_path, res.strategy)) {
         return 0;
     }
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -2541,8 +2541,9 @@ static void resolve_file_calls(resolve_ctx_t *rc, resolve_worker_state_t *ws, CB
                                   memory_order_relaxed);
         if (target_node && source_node->id != target_node->id &&
             cbm_suppress_cross_language_suffix_match(lang, target_node->file_path, res.strategy)) {
-            /* #725: same guard as pass_calls.c — do not emit a suffix_match
-             * CALLS edge across a language boundary. */
+            /* #725/#1572: same guard as pass_calls.c — do not emit a
+             * suffix_match or unique_name CALLS edge across a language
+             * boundary. */
             continue;
         }
         if (!target_node || source_node->id == target_node->id) {
--- a/src/pipeline/pipeline.h
+++ b/src/pipeline/pipeline.h
@@ -264,11 +264,13 @@ bool cbm_perl_suppress_generic_match(bool is_perl, bool is_method, const char *c
  * Pure; unit-tested in test_registry.c. */
 bool cbm_tsjs_suppress_weak_method_match(bool is_tsjs, bool is_method, const char *strategy);
 
-/* #725: drop a suffix_match CALLS edge when the caller language and the
- * target file's language disagree. unique_name (candidates == 1) is #1572
- * and is left alone; same_module / import_map / lsp_* are kept. JS/TS/TSX
- * are one family so a .ts helper calling a .tsx function is not dropped.
- * Pure; unit-tested in test_registry.c. */
+/* #725/#1572: drop a suffix_match or unique_name CALLS edge when the caller
+ * language and the target file's language disagree. suffix_match is the
+ * import-distance winner among many same-named symbols (#725);
+ * unique_name is the candidates==1 case of the same class (#1572).
+ * same_module / import_map / lsp_* are kept. JS/TS/TSX are one family so a
+ * .ts helper calling a .tsx function is not dropped. Pure; unit-tested in
+ * test_registry.c. */
 bool cbm_suppress_cross_language_suffix_match(CBMLanguage caller_lang, const char *target_file_path,
                                               const char *strategy);
 
--- a/src/pipeline/registry.c
+++ b/src/pipeline/registry.c
@@ -476,10 +476,12 @@ static const char *path_basename(const char *path) {
 bool cbm_suppress_cross_language_suffix_match(CBMLanguage caller_lang, const char *target_file_path,
                                               const char *strategy) {
     /* Two same-named symbols in different languages: suffix_match picks one
-     * winner by import-distance and attaches every bare-name call to it
-     * (#725, Bash/Python main, JS/Python commit). unique_name is the
-     * candidates==1 case (#1572) and is not this guard. */
-    if (!strategy || strcmp(strategy, "suffix_match") != 0) {
+     * winner by import-distance (#725, Bash/Python main, JS/Python commit).
+     * unique_name is the candidates==1 case of the same class (#1572,
+     * Python `from unittest.mock import patch` binding to a unique TSX
+     * `patch`). */
+    if (!strategy ||
+        (strcmp(strategy, "suffix_match") != 0 && strcmp(strategy, "unique_name") != 0)) {
         return false;
     }
     if (caller_lang == CBM_LANG_COUNT || !target_file_path || !target_file_path[0]) {
--- a/tests/test_pipeline.c
+++ b/tests/test_pipeline.c
@@ -6034,7 +6034,7 @@ TEST(pipeline_python_cross_module_call) {
 /* #725: two same-named symbols across languages must not share CALLS edges.
  * Python Store.commit is the real callee of save(); the JS Editor.commit
  * function is a distinct binding and must have no inbound CALLS from Python.
- * unique_name (candidates==1) is #1572 and is not this claim. */
+ * unique_name (candidates==1) is the #1572 pipeline test. */
 TEST(pipeline_cross_language_same_name_does_not_share_calls_issue725) {
     const char *files[] = {"store.py", "app.py", "web/src/pages/Editor.js"};
     const char *contents[] = {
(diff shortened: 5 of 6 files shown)


rudi193-cmd:
CI red on this PR is not the #1572 change.

Every failing shard dies on the same two tests, both predating this branch:

- `cli_zero_argument_tool_never_reads_stdin_issue1359`
- `cli_stdin_args_gate_tracks_tool_schema_issue1359`

`list_projects` now advertises pagination properties (`offset`, `limit`, `include_details`) on `main`. The #1359 gate treats “has properties” as “read stdin”, so `cli list_projects` is allowed to slurp stdin again and the tests that pinned “no properties / no stdin” go red. Sibling PRs off the same `main` (`#1697`, `#1698`, `#1699`) fail the same way.

Windows shard 2/2 also has an unrelated UTF-8 compare (`raw-??????? content` vs `raw-Русский content` in `test_mcp.c`).

`scripts/test.sh --suites registry,pipeline` was green locally for the unique_name tests. Not folding a #1359 schema-gate fix into this PR.

DeusData:
Thank you for keeping this to the remaining `unique_name` cross-language case and explicitly excluding the adjacent import, receiver, and external-specifier work. That is the atomic boundary we need. The reported CI failures are also correctly left out of this diff rather than being bundled into the resolver fix. It is labeled as a high-priority graph-quality bug; the final review will verify same-language-family behavior and the negative cross-language cases before any merge decision.

DeusData:
Picking this up, and doing the review that was promised on 18 August rather than leaving you waiting further. Sorry it took two weeks.

**Your CI diagnosis was right, and there is now an extra fact you could not have had.** Both of those tests — `cli_zero_argument_tool_never_reads_stdin_issue1359` and `cli_stdin_args_gate_tracks_tool_schema_issue1359` — **pass on the current tree**. I ran the full suite an hour ago: 7781 passed, 7 skipped, 0 failed, with both of those green. So the base defect you identified has since been fixed, and a rebase should clear your red without you touching anything.

**One caveat before you rebase: `main` does not currently compile.** Two changes landed the same type short-name index into `type_registry.h` and git merged them into duplicate struct members, so anything built against `main` right now fails. #1993 is the repair. Rebase after that lands, not before, or you will trade one unexplained red for another.

## The review

The two things the 18 August note said would be verified before any merge decision — same-language-family behaviour, and the negative cross-language cases — both hold.

The production change is two lines: `unique_name` joins `suffix_match` in the strategy allowlist of the existing `cbm_suppress_cross_language_suffix_match`. Nothing else in `registry.c` moves. That is the right shape for this repo — a per-language-gated helper wired at both `pass_calls.c` and `pass_parallel.c`, rather than a restriction inside `cbm_registry_resolve` itself.

`cross_language_unique_name_drops_py_vs_tsx` asserts the cases that matter, including the ones a weaker suite would omit:

- suppressed in **both** directions (Python → `.tsx`, TSX → `.py`), not just the one from the issue
- **not** suppressed same-language
- `same_module`, `import_map` and `lsp_direct` all explicitly **not** suppressed
- JS/TS/TSX kept as one family, with `.ts` and `.js` callers of a `.tsx` target both asserted

That third bullet is the one I care most about, and you tested it. Every existing guard in this file deliberately preserves `same_module`, because a blocked high-confidence match does not disappear — it demotes into `suffix_match`, which picks a winner by import distance project-wide. A guard that gated `same_module` off could lose a correct edge and mint a wrong one in the same step. Yours cannot.

## What actually holds it up

Not this PR's quality. It belongs to the call-resolution cluster — #1128, #1324, #1386, this one, #1766 — and there is a standing decision that they are judged **against one shared census rather than one at a time**. #1128 is the most global member, since it changes strategies inside `cbm_registry_resolve` while the others add suppressors downstream; whichever of them lands first shifts the baseline and makes the rest unjudgeable on their own numbers.

So this will not merge in isolation, and that is a sequencing constraint rather than a verdict on the change. I would rather tell you that plainly than let it sit silently for another fortnight.

Thank you for holding the scope where you did — excluding the import-edge cleanup, the receiver disambiguation and the external-specifier question was the right call, and it is why this one was quick to read.

DeusData:
Census done — this PR and #1766 were measured together on real graphs (ten languages, ~3.1M CALLS edges, every removed edge classified, three runs each; the two guards compose exactly: removed(both) = removed(#1766) ∪ removed(#1702) on nine corpora with zero additions, so the CALLS sets are deterministic). Widening the guard to `unique_name` is sound wherever the family rule is right: go −248, rust −128, kotlin −90, django −52, perl −13, ts −43 are all correct false bindings (sh / Makefile / toml / json / css → code — meilisearch's `.zip()` bound to a Cargo.toml `zip` 78×, koel's PHP `noContent()` bound to a `.vue` file 40×). Three places where the family rule is too coarse and the guard now removes true calls:

1. **C family.** The CALLS guard honours only `js_ts_family`, while the sibling ref guard (`registry.c:580`) also honours `c_cpp_family`; `.h` files are `CBM_LANG_CPP`, so a C → header call counts as cross-language. dotnet/runtime: −18,866 CALLS (−1.82%), 8,613 of them C-family edges, about 7.5k of those true static-inline / prototype calls (5,443 in `src/mono`) — e.g. `mini-runtime.c:3401` `m_type_is_byref(sig->ret)` → `metadata-internals.h:1294`.
2. **JVM family.** Gradle/Groovy build scripts → Java build tools: 419 true calls lost in elasticsearch (54% of its removals) — e.g. `server/build.gradle:307` `t.getComponentName().set("server")` → `GenerateTestBuildInfoTask`.
3. **Vue SFC → TS.** ~35 true calls in koel — e.g. `AiAssistantScreen.vue:90` `playback().queueAndPlay()` → `QueuePlaybackService.ts:336`.

Fix: reuse `c_cpp_family` in the CALLS guard, add a JVM family (java / kt / kts / groovy / gradle), and put vue / svelte / astro into the js_ts family. With those three the `unique_name` widening is a pure win on every corpus we measured. Please also rebase — `tests/test_registry.c` conflicts textually, the `src` side merges clean. Same sign-off, and this merges on green. Thank you for the patience while the numbers were produced; they are what makes this safe to land.

PR #1703, opened 2026-08-18
fix(mcp): emit trace evidence columns in header order

### LINKED ISSUE #1542: trace_path format=json drops include_evidence strategy and confidence
### Version

`codebase-memory-mcp 0.10.0`

### Platform

macOS (Apple Silicon / arm64)

### Install channel

Installed release binary at `~/.local/bin/codebase-memory-mcp`

### Binary variant

Standard

`trace_path(include_evidence=true)` correctly returns the public strategy class and confidence in the default tree format, but silently drops both fields when `format:"json"` is also requested.

The installed tool schema and `cli trace_path --help` promise that `include_evidence` adds two columns per hop in addition to `name` and `hop`. The JSON response instead contains only:

```json
"cols": ["name", "hop"]
```

Expected JSON columns:

```json
"cols": ["name", "hop", "strategy", "confidence"]
```

This affects both the one-shot CLI and MCP because they share the same trace response implementation.

Related to #1355: current provenance support works for default tree output, but the JSON response path still omits it.

### Synthetic reproduction

Create `repro.php`:

```php
<?php

final class EvidenceTarget
{
    public static function answer(): int
    {
        return 42;
    }
}

final class EvidenceCaller
{
    public function run(): int
    {
        return EvidenceTarget::answer();
    }
}
```

Index it:

```bash
codebase-memory-mcp cli --json index_repository \
  --repo-path /path/to/fixture \
  --mode full \
  --name cbm-evidence-json-repro
```

The fresh fixture index contained 8 nodes, 10 edges, 0 skipped files, and 0 partial parses.

#### Control: default tree format works

```bash
codebase-memory-mcp cli --json trace_path \
  --project cbm-evidence-json-repro \
  --function-name cbm-evidence-json-repro.repro.EvidenceCaller.run \
  --direction outbound \
  --depth 1 \
  --include-evidence true \
  --limit 20
```

Actual payload:

```text
callees: 1  (rows: name hop strategy confidence; qn = group prefix + "." + name)
cbm-evidence-json-repro.repro.EvidenceTarget:
  answer 1 language_rule 0.95
```

#### Failing case: JSON format drops evidence

```bash
codebase-memory-mcp cli --json trace_path \
  --project cbm-evidence-json-repro \
  --function-name cbm-evidence-json-repro.repro.EvidenceCaller.run \
  --direction outbound \
  --depth 1 \
  --include-evidence true \
  --format json \
  --limit 20
```

Actual structured payload:

```json
{
  "function": "cbm-evidence-json-repro.repro.EvidenceCaller.run",
  "direction": "outbound",
  "callees_total": 1,
  "callees": {
    "cols": ["name", "hop"],
    "groups": [
      {
        "qn_prefix": "cbm-evidence-json-repro.repro.EvidenceTarget",
        "rows": [["answer", 1]]
      }
    ]
  }
}
```

#### The graph edge contains the evidence

```bash
codebase-memory-mcp cli --json query_graph \
  --project cbm-evidence-json-repro \
  --query 'MATCH (a)-[r:CALLS]->(b)
           WHERE a.qualified_name = "cbm-evidence-json-repro.repro.EvidenceCaller.run"
           RETURN b.qualified_name, r.strategy, r.confidence, r.candidates, r.line
           LIMIT 10'
```

Actual row:

```text
cbm-evidence-json-repro.repro.EvidenceTarget.answer
php_static_resolved  "0.95"  "1"  "15"
```

### Exact code cause in v0.10.0

The handler correctly reads `include_evidence`:

https://github.com/DeusData/codebase-memory-mcp/blob/v0.10.0/src/mcp/mcp.c#L6008-L6013

The default tree branch forwards it to the evidence-aware emitter:

https://github.com/DeusData/codebase-memory-mcp/blob/v0.10.0/src/mcp/mcp.c#L6268-L6294

However, `format:"json"` selects the separate JSON branch, which calls `bfs_to_tree_json(...)` without `include_evidence`:

https://github.com/DeusData/codebase-memory-mcp/blob/v0.10.0/src/mcp/mcp.c#L6311-L6330

The JSON emitter signature has no evidence parameter:

https://github.com/DeusData/codebase-memory-mcp/blob/v0.10.0/src/mcp/mcp.c#L5832-L5836

So the evidence exists in the graph and traversal result, but is lost specifically during JSON response projection.

### Impact

Agents requesting structured JSON receive speculative or heuristic `CALLS` hops without the strategy/confidence fields explicitly requested for trust decisions. The response succeeds normally, so clients cannot distinguish the contract failure from evidence-free data.

### Smallest fix

- Pass `include_evidence` into the JSON trace emitter.
- Add `strategy` and `confidence` columns and values using the existing `bfs_edge_evidence_for_hop()` and public strategy-class mapping.
- Add regression coverage for `format:"json"` plus `include_evidence:true` through both CLI and MCP envelopes.

### Confirmations

- [x] I searched existing issues and did not find this JSON-emitter-specific bug.
- [x] The reproduction is fully synthetic and contains no proprietary source.

rudi193-cmd:
## What does this PR do?

The original #1542 json omission landed in v0.10.3. Two leftovers remained on `main`:

1. `bfs_to_tree_json` declared `strategy, confidence` before `args` but emitted `args` first, so `json + include_evidence + data_flow` mislabeled columns (`len(cols)==len(row)` still held).
2. Tree `flat_trace` (`risk_labels` or `data_flow`) called `bfs_to_toon_table` without `include_evidence`, so evidence was a silent no-op on that branch.

This PR emits cells in header order and forwards the flag into the flat table. Tests assert column identity, not just count.

Not folding in #1597 budget signalling.

## Checklist

- [x] Tests pass locally (`scripts/test.sh --suites mcp`)

Fixes #1542


DeusData:
Thank you for the focused follow-up to #1542. This is the right review shape: one output-contract defect, two production branches, and tests that bind column identity rather than merely row width. The initial patch scan found no new dependency, execution, network, or configuration surface. It is labeled high priority because silently mislabeling trace evidence is a correctness bug. The review queue is full, but this contribution is now routed as a compact bug fix.

DeusData:
Reviewed, and the change is right. Sorry it sat two weeks after the 18 August note.

**First, your four red checks: none of them are your code, and the cause has already been fixed.**

`security / codeql-gate` started at 03:38:15 and completed at 04:23:51 — 45 minutes and 36 seconds. That is the gate's old wait budget expiring, not a finding. It then cascaded: `lint / lint-mem` shows `The operation was canceled` with zero warnings of its own, and `analyze` and `ci-ok` fall out of the same run.

The gate's wait was raised from 45 to 150 minutes in #1960, merged earlier today. So a re-run should clear all four without you changing anything.

**One thing to wait for before re-running:** `main` does not currently compile — two changes landed the same type short-name index into `type_registry.h` and git merged them into duplicate struct members. #1993 is the repair and is nearly green. Re-run after that lands, or you will swap one unexplained red for another. I will trigger it from this side; there is nothing for you to do.

## The review

Both defects are real and the fixes are the right shape.

**Column order.** Emitting cells in header order rather than declaration order is the fix, and the reason this survived is worth naming: `len(cols) == len(row)` still held, so a width assertion passed while every column past the divergence was mislabeled. A test that counts columns cannot see this class of bug at all. Yours assert column *identity*, which is why they would have caught it.

**The dropped flag.** `flat_trace` calling `bfs_to_toon_table` without `include_evidence` made the flag a silent no-op on that branch — the worst kind, because the caller has every reason to believe it worked.

**And you caught the consequence of your own fix.** `const char *cols[5]` was exactly right on `main`: qn, hop, risk, test, args. Adding `strategy` and `confidence` makes the maximum seven, and you resized to `cols[7]` rather than leaving a two-element overflow behind the new branch. That is the part of this change that would have been genuinely dangerous to miss, and it is easy to miss because the array is declared forty lines above the code that grows it.

The `-`/`-` placeholders for the root hop and non-CALLS edges keeping the column count fixed, and matching `bfs_to_tree_table`'s existing convention rather than inventing a second one, is right too.

Holding scope out of #1597's budget signalling was the correct call.

**Status:** approved and cleared. It merges once `main` is green and the checks re-run.

DeusData:
Merged as `5fbab7bb`.

Two things about this one are worth recording rather than just closing.

**The bug survived because the test that should have caught it was counting.** `len(cols) == len(row)` held while every column past the divergence was mislabeled, so a width assertion passed on wrong data. Your tests assert column *identity*, which is the only shape that can see this class of defect. That distinction is worth more than the fix.

**And you caught the consequence of your own change.** `const char *cols[5]` was exactly right on `main` — qn, hop, risk, test, args. Adding `strategy` and `confidence` makes the maximum seven, and you resized to `cols[7]` rather than leaving a two-element overflow sitting behind the new branch. The array is declared about forty lines above the code that grows it, which is precisely the distance at which people miss this.

**On the CI you were told about:** your four red checks were never yours. `security / codeql-gate` ran 45 minutes and 36 seconds — the gate's old wait budget expiring, not a finding — and `lint / lint-mem` recorded "The operation was canceled" with zero warnings of its own, so the gating memory lane was never actually objecting. #1960 raised that budget to 150 minutes, and the re-run after `main` was repaired came back clean across the full matrix.

Sorry it sat two weeks after the 18 August note. Thank you for keeping it to one claim and holding #1597's budget signalling out.

PR #1707, opened 2026-08-18
fix(windows): pin the search_code PowerShell pipe to UTF-8

DeusData:
Root-causes the intermittent windows-leg failure `test_mcp.c:4427 "raw-??????? content" != "raw-Русский content"` (seen on PR #1704's run, absent on sibling runs of the same tree).

**Mechanism**: the Windows raw scan pipes `powershell ... Select-String` output into `collect_grep_matches`. PowerShell 5.1 encodes stdout for a native-process pipe in the **console OEM codepage**, so any character the inherited CP can't carry (Cyrillic under CP437/850) degrades to `?` before reaching the parser — and whether it degrades depends on which console the server inherited (MSYS2 CP65001 → clean; classic consoles → mojibake). That's why the test flickers on CI, and it means real Windows users in a default console get `?` for all non-ASCII `search_code` raw content.

**Fix**: every generated command now pins `[Console]::OutputEncoding` to UTF-8 — codepage-independent by construction. The read side needs no pin (Select-String decodes BOM-less UTF-8 via .NET StreamReader defaults). A Windows-side test asserts all five builder variants carry the prelude.

**Verification note**: mechanism attributed from the CI log + PS 5.1 documented behavior; the observed CI failure is the RED on the manifesting platform (x64 MSYS2 — the arch-faithful venue). The local Windows VM is currently not bootable via automation, so the deterministic `chcp 437` E2E repro wasn't run locally; the builder-shape test is deterministic on every Windows leg.

https://claude.ai/code/session_018RiRAw9RQhvCoshqe7eZHV

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -8848,6 +8848,16 @@ bool cbm_search_code_file_pattern_can_prefilter(const char *file_pattern) {
 /* Build the grep/search command string based on scoped vs recursive mode.
  * On Windows, uses PowerShell Select-String with tab-delimited output.
  * On POSIX, uses grep with colon-delimited output. */
+/* Windows PowerShell 5.1 encodes stdout for a native-process pipe in the
+ * console OEM codepage, so any character the inherited CP cannot carry
+ * (Cyrillic under CP437/850, etc.) degrades to '?' before it ever reaches
+ * collect_grep_matches — and WHETHER it degrades depends on the console the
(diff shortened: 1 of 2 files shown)


DeusData:
Folding this into #1704: main is currently red on complementary legs (this fix is red on the mac legs via the stdin-gate composition that #1704 fixes, and #1704 is red on the windows leg via this bug), so as separate PRs they deadlock against the required ci-ok check. The commit is cherry-picked into #1704 unchanged (4dd09967).

PR #1706, opened 2026-08-18
[ledger] feat(lsp): Hybrid LSP support for Ruby — split into #1709-#1713

benrfairless:
> **This PR is now the ledger for the series, not a merge candidate.**
> The work has been split into five ordered PRs as requested in review. This
> branch is left untouched as the record of where the work came from.
> Please review and merge #1709 – #1713; close this one whenever suits.

## The five slices

| # | PR | Branch | Depends on | Diff |
|---|----|--------|------------|------|
| 1 | #1709 | `fix/ruby-superclass-base-classes` | — | 1 file +7, 1 test file +68 |
| 2 | #1710 | `fix/ruby-namespaced-constructor-callee` | — | 1 file +5, 1 test file +52 |
| 3 | #1711 | `feat/ruby-lsp-per-file` | — | 9 files, +2521 |
| 4 | #1712 | `feat/ruby-lsp-cross-file` | 3 | 5 files, +349/−7 |
| 5 | #1713 | `docs/ruby-hybrid-lsp` | 4 | README only, +4/−3 |

Ordering follows the review exactly. Stacking does not: the review said stack
*where a later slice needs an earlier one*, so only 4→3 and 5→4 are stacked.

**Measured, not assumed.** I built slice 3 on a branch containing neither
extraction fix and ran its rows: 20/20 pass. The per-file resolver reads the
`superclass` field straight off the AST rather than through `base_classes`, so
it does not need slice 1, and no per-file row uses `A::B.new`, so it does not
need slice 2. Slices 1, 2 and 3 are therefore independently mergeable in any
order.

## Deviations from the outline, and why

**1. `cbm_run_ruby_lsp_cross` moved from slice 3 to slice 4.** In the original
commit it shipped with the resolver, where nothing called it until the pipeline
wiring landed — ~154 lines of untested-in-its-own-PR code. Moving it makes each
slice's tests exercise everything that slice adds. The two halves reassemble
**byte-identically** to the original file; I checked with `git diff` against
this branch.

**2. `docs/index.html` dropped entirely.** The original commit corrected a
pre-existing error there (structured data said nine language families and
omitted Perl, both wrong before this work). Those are the same strings the Ruby
edit lands on, so adding Ruby without fixing them would mean authoring a
sentence I know is wrong, and fixing them in slice 5 would mix an unrelated
correction into a feature PR. Left for its own change.

## Corrections found while splitting

Splitting surfaced two things worth stating plainly rather than quietly fixing:

[...]

Issue #1701, opened 2026-08-18
Add Hybrid LSP support for Ruby


benrfairless:
### What problem does this solve?

Most of our applications are ruby apps - providing Hybrid LSP support for ruby applications makes this tool more valuable to us.

### Proposed solution

Provide Hybrid LSP support for the Ruby language.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

DeusData:
Thank you for opening the language request first. The linked implementation in #1706 is now the review route for this work. Ruby Hybrid LSP support is broader than an extension mapping, so the review will check deterministic import, call, usage, inheritance, and fallback behavior before any merge decision. The contribution queue is full, but this is now labeled and tracked rather than waiting without a route.

Issue #1359, opened 2026-07-30
cli hangs forever reading stdin when no argument-bearing token follows the tool name and stdin is an open pipe (default Node spawn deadlocks)


artaommahe:
### Version

codebase-memory-mcp 0.9.1-rc.1

### Platform

macOS (Apple Silicon)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

`codebase-memory-mcp cli <tool>` invoked **with no flags** blocks forever reading stdin whenever stdin is an open pipe that never delivers data. It produces no output and never exits.

This hits tools that legitimately need no arguments — `list_projects` is the obvious one. Passing any flag avoids it (`index_status --project X` and `query_graph --project X --query "..."` both complete normally against the same open-pipe stdin), so the trigger is specifically "no flags given → unconditionally read args from stdin".

That is a deadlock for automation. Any hook, CI step, git hook, or agent harness that runs `cli list_projects` while stdin is an inherited pipe hangs indefinitely rather than failing. I hit it by accident: a shell script doing

```bash
P=$(codebase-memory-mcp cli list_projects | grep -o '...' | head -1)
```

hung, and the process was still stuck **14 minutes later**. A `sample(1)` of the stuck process shows the main thread parked in `fread` on fd 0:

```
2602 Thread_13001004   DispatchQueue_1: com.apple.main-thread  (serial)
  + 2602 start  (in dyld) + 6992
  +   2602 ??? (in codebase-memory-mcp) + 0x5474
  +     2602 ??? (in codebase-memory-mcp) + 0x5f9c
  +       2602 ??? (in codebase-memory-mcp) + 0x8768
  +         2602 fread  (in libsystem_c.dylib) + 136
  +           2602 __fread  (in libsystem_c.dylib) + 488
  +             2602 __srefill1  (in libsystem_c.dylib) + 36
  +               2602 _sread  (in libsystem_c.dylib) + 32
  +                 2602 __sread  (in libsystem_c.dylib) + 24
  +                   2602 __read_nocancel  (in libsystem_kernel.dylib) + 8
```

Expected: only read stdin when it is actually ready/has data (`select`/`poll` with a zero timeout, or check `isatty`/`S_ISFIFO` plus a non-blocking peek), or only when the tool actually requires arguments that weren't supplied. A tool that needs no arguments should never block on stdin. Failing that, a short read timeout with a clear error beats an unbounded hang.

[...]

PR #1709, opened 2026-08-18
fix(extraction): keep Ruby superclass names bare

benrfairless:
## What does this PR do?

Slice 1 of 5 splitting #1706, in the order requested in review. This one stands alone and depends on nothing else in the series.

tree-sitter-ruby wraps `class C < Base` in a `superclass` node whose child is a `constant` (or a `scope_resolution` for `A::B`). `collect_bases_from_field` matched neither, so the superclass fell through to the raw-text fallback and `base_classes` stored `"< Base"` — operator included. A base spelled that way never resolves, so Ruby subclasses produced zero INHERITS edges and the Rails chain (`User < ApplicationRecord < ActiveRecord::Base`) never linked up.

The fix adds both node kinds to the allow-list that already carries Python's `identifier` / `attribute`.

**Reproduce-first.** The new `inherit_ruby` rows in `tests/test_extraction_inheritance.c` fail 5/5 without the one-line change, each on the `"<"` substring that is the bug's signature:

```
FAIL [Dog] expected base "Animal" not found in base_classes
FAIL [Dog] bad string "<" found inside a base_classes entry
... 5 / 5 cases above failed
```

Incidental observation, not changed here: `inherit_python`, `inherit_typescript`, `inherit_tsx`, `inherit_php` and `inherit_kotlin` all pass now, so the `RED:` markers on them in `SUITE(extraction_inheritance)` are stale. Left alone to keep this PR to one thing; happy to file it separately.

| # | PR | Depends on |
|---|----|------------|
| **1** | **this PR** | — |
| 2 | `fix/ruby-namespaced-constructor-callee` | — |
| 3 | `feat/ruby-lsp-per-file` | — |

Slices 1, 2 and 3 are independent: I verified the resolver in slice 3 passes all its rows with neither extraction fix applied, so they are only stacked where a later slice genuinely needs an earlier one, per the review.

### CI baseline

`main` at 4d7d9f11 is red for reasons unrelated to this change: #1181 gave `list_projects` pagination parameters, retiring the last empty-properties schema that #1359's guards used as their zero-argument example. `cli_zero_argument_tool_never_reads_stdin_issue1359` and `cli_stdin_args_gate_tracks_tool_schema_issue1359` fail on every leg. #1704 already fixes it. Until that merges, `ci-ok` will be red here for that reason.

Refs #1701

## Checklist

- [x] Tests pass locally (`scripts/test.sh`) — full canonical leg run on the five slices merged together: 7542 passed, 32 failed, 5 skipped. All 32 failures are in `cli` and reproduce identically on unmodified `origin/main` (2 are the #1704 breakage above; 30 are this machine's real agent configs interfering with the CLI install tests). No new failure is introduced by this series.

---


DeusData:
Approved. Slice 1 does exactly one thing and the test proves it.

Adding `constant` and `scope_resolution` alongside Python's `identifier` / `attribute` is the right fix in the right place, and mirroring the existing comment's structure means the next reader sees one pattern with two instances rather than two special cases.

**The reproduce-first rows are the good part.** Failing 5/5 on the `"<"` substring is not just "the test is red" — it names the bug's signature, so a future regression announces itself as the same fault rather than as an unexplained diff. `bad string "<" found inside a base_classes entry` is a better assertion than an equality check would have been.

**One thing I verified rather than assumed,** because `collect_bases_from_field` is language-generic and `constant` is not a Ruby-only node kind — it is declared by ten of the vendored grammars (ada, crystal, fortran, gleam, lean, protobuf, ruby, solidity, llvm, vhdl). The widening is still safe: the helper is only entered through an explicit inheritance field (`ts_node_child_by_field_name` over the `fields` list) or a `heritage_children` named child, so a grammar with no inheritance field never reaches it, and Crystal — the one Ruby-shaped language in that list — would be correct to treat a `constant` there as a base anyway. `scope_resolution` is Ruby-only.

Worth stating because the diff does not show it, and "we added a node kind to a shared allow-list" is the sort of change that deserves the check.

**Your CI note is now stale in your favour.** The `#1181` × `#1359` breakage you documented — `cli_zero_argument_tool_never_reads_stdin_issue1359` and `cli_stdin_args_gate_tracks_tool_schema_issue1359` — is fixed on `main`; I ran the full suite earlier today and both pass. So `ci-ok` should no longer be red here for that reason.

**And a heads-up on timing:** `main` was briefly broken today (two changes landed the same type short-name index into `type_registry.h` and git merged them into duplicate struct members), repaired an hour ago by #1993. Your green here predates all of that, so I will update this branch against current `main` and re-run before merging — nothing needed from you.

On the stale `RED:` markers for `inherit_python` / `inherit_typescript` / `inherit_tsx` / `inherit_php` / `inherit_kotlin`: yes please, file that separately. You were right to keep this PR to one thing.

**Status:** cleared, reviewed, approved. It merges once the re-run against current `main` is green.

DeusData:
Merged as `8e1de0403edd` — thank you!

This is slice 1 of the five from #1706 and it stands entirely on its own: `tree-sitter-ruby` wraps `class C < Base` in a `superclass` node whose child is a `constant` (or a `scope_resolution` for `A::B`), `collect_bases_from_field` matched neither, and the raw-text fallback stored `"< Base"` *with* the operator — which never resolves, so every Ruby subclass had zero INHERITS edges. Adding both node kinds to the allow-list that already carries Python's `identifier`/`attribute` is exactly the right shape.

Two things I want to call out because they are the reason this went straight through:

- The five reproduce-first rows assert on the `<` **substring** — the bug's actual signature — instead of plain equality, so a regression re-announces itself as the same fault rather than as a vague mismatch.
- I checked the shared-helper blast radius rather than assuming it: `constant` is declared by ten vendored grammars, but `collect_bases_from_field` is only entered through an explicit inheritance field or a heritage child, so a grammar with no inheritance field never reaches it, and Crystal (the one Ruby-shaped language in that list) would be *correct* to treat a constant there as a base. Safe.

Your documented CI baseline (the #1181 × #1359 `cli` failures) has gone stale in your favour — both pass on current `main` now. On to slice 2 (#1710) next.

PR #1711, opened 2026-08-18
feat(lsp): add the Ruby per-file resolver

benrfairless:
## What does this PR do?

Slice 3 of 5 splitting #1706, and the one that carries the clean-room boundary. Independent of slices 1 and 2 — see the verification note below.

In-process, type-aware call resolution for **single Ruby files**, mirroring the `perl_lsp.c` / `php_lsp.c` shape and dispatched from `cbm_extract_file`:

- **PASS 1** collects classes and modules with lexical nesting, superclasses, and `include` / `prepend` / `extend` mixins, plus method tables with instance and singleton methods classified from the AST. Singleton methods key on `"<class_qn>.self"`, which cannot collide with a constant path.
- **PASS 1.5** infers instance-variable types from `@x = Const.new`; conflicting assignments latch to no type rather than guessing.
- **PASS 2** resolves constructors (`Const.new` targets the class node, the way Python's `lsp_constructor` does), implicit and explicit self dispatch, singleton methods, superclass chains, mixin lookup in Ruby's method-resolution order, `super`, and ivar / local / chained receivers.

The stdlib seed covers Ruby core plus a curated Rails surface, including ActiveRecord query typing: `User.find(1)` is a `User`, and relation chains keep the model type. Dynamic dispatch and unresolved receivers emit no edge, and only project-defined targets are emitted.

Cross-file resolution is deliberately **not** here — it is slice 4, so that every line this PR adds is exercised by this PR's own tests.

## Provenance and originality

This resolver was authored **clean-room** from the tree-sitter-ruby grammar and Ruby language semantics. No Ruby language server source was consulted.

Per the review, the originality reference ships with this slice rather than with the docs, so the generated data and the clean-room boundary can be audited together. `scripts/check-lsp-originality.sh` gains Shopify's ruby-lsp (MIT) as the Ruby reference, and the scan is clean:

```
[lsp-orig] extracted 2898 candidate tokens from 37 LSP sources (147 strings, 2775 comments)
[lsp-orig] fetching ruby (https://github.com/Shopify/ruby-lsp :: lib) ...
[lsp-orig] ok — no verbatim overlap with 'ruby'

[lsp-orig] CLEAN — no verbatim string/comment overlap with any reference, and no
[lsp-orig] structural clones (jscpd) between our C and clangd.
```

## Verification that this slice is independent

The 20 rows in `tests/test_ruby_lsp.c` all pass on this branch, which is based directly on `main` and contains **neither** extraction fix from slices 1 and 2. The per-file resolver reads the `superclass` field straight off the AST rather than going through `base_classes`, so it does not need slice 1; and no per-file row uses `A::B.new`, so it does not need slice 2. That is why this is not stacked on them.

I also ran the Ruby-adjacent suites together (`ruby_lsp extraction extraction_inheritance grammar_regression language lang_contract edge_structural lsp_resolution_probe`): 679 passed, 0 failed. Worth checking because the resolver now runs on every Ruby extraction.

## On size

At ~2.5k lines this is well over the 500-line guidance in CONTRIBUTING. It is the atomic unit the review asked for — resolver, standard-library seed and its focused tests — and splitting it further would separate the generated data from the code that consumes it, which is the opposite of what the originality audit needs.

| # | PR | Depends on |
|---|----|------------|
| 1 | #1709 | — |
| 2 | #1710 | — |
| **3** | **this PR** | — |

### CI baseline

Refs #1701

## Checklist

---


DeusData:
Thank you for the Ruby resolver contribution, and sorry that this foundational slice has had only the automated acknowledgement.

This is now routed as a Ruby language and parsing enhancement with normal priority. Per-file resolution is the base of the later cross-file slices, so we need time to review its semantic coverage and generated stdlib data carefully rather than infer acceptance from the later stack discussion.

The contribution queue is quite full, but we will return with grounded feedback. No changes are requested from you at this triage stage.

PR #1712, opened 2026-08-18
feat(pipeline): wire Ruby cross-file LSP resolution

benrfairless:
## What does this PR do?

Slice 4 of 5 splitting #1706. **Stacked on #1711** — see the review note below.

Adds the cross-file tier on top of the per-file resolver.

`cbm_run_ruby_lsp_cross` builds its class table from the project-wide `CBMLSPDef[]`, deriving each constant path by stripping the def's module QN prefix, and registers methods on both the instance and singleton receiver keys because `CBMLSPDef` does not carry the `def self.m` split. Ruby is registered on the fallback `cbm_pxc_run_one` path, like Java and Kotlin — there is no prebuilt tier-2 registry yet.

**The one deliberate design decision here**, and the thing most worth arguing with: Ruby joins Rust in bypassing the own+imports module-def filter, for a different reason. Ruby constants live in one global namespace and Rails/Zeitwerk autoloads them with no `require` statement, so a filter keyed on requires would starve essentially all cross-file resolution in a real Ruby app.

The two new rows drive `cbm_run_ruby_lsp_cross` with a hand-built `CBMLSPDef[]` in the shape `pass_lsp_cross.c` passes it, the second mirroring the Rails-shaped e2e fixture (`User < ApplicationRecord < ActiveRecord::Base`, a service object, and a caller in a third file).

## Reviewing just this slice

A PR from a fork can only target a branch in this repo, so the base here is `main` and GitHub shows #1711's commit as well as this one. **Only the second commit is this slice**:

- `11418b45` — this slice (5 files, +349/−7)
- `e9f87efe` — belongs to #1711

Merge #1711 first and this PR's diff collapses to that single commit.

Note that `internal/cbm/lsp/ruby_lsp.{c,h}` appear in both PRs. That is the split the review asked for: the cross-file entry point moved out of the resolver slice so each slice's tests exercise everything it adds. I verified the two halves reassemble byte-identically to the original single-commit file.

| # | PR | Depends on |
|---|----|------------|
| 1 | #1709 | — |
| 2 | #1710 | — |
| 3 | #1711 | — |
| **4** | **this PR** | 3 |

### CI baseline

Refs #1701

## Checklist

---


DeusData:
Reviewed slice 4. The wiring is right and the Zeitwerk argument is correct. There is one consequence of it I want measured before this lands, and you have already named half of it yourself.

**The filter exemption is justified.** Ruby constants live in one global namespace and Rails/Zeitwerk autoloads them with no `require`, so an own+imports filter really would starve cross-file resolution. Putting Ruby beside Rust in that branch, with a comment saying *why the mechanics differ*, is the right way to record it.

## The consequence: Ruby has the Rust exemption without the Rust mitigation

Read the branch you edited alongside its own comment:

> Rust therefore always resolves against the FULL def universe: **the lazily built shared registry when available, else a full per-file build.**

Rust survives the exemption because the registry is its normal path and the full per-file build is its fallback. Your PR description says the quiet part plainly: *"there is no prebuilt tier-2 registry yet."* So for Ruby the fallback **is** the only path.

Following it through:

- `cbm_pxc_dispatch_file` runs per file; with `lang != CBM_LANG_RUBY` removed from the filter guard, `file_defs = all_defs`.
- `cbm_pxc_run_one` hands that whole array to `cbm_run_ruby_lsp_cross`.
- Which opens with `for (int i = 0; i < def_count; i++)` to build its class table.

So every Ruby file rebuilds a class table over the entire project's defs. That is F × D, and D grows with F — the corpus-proportional shape, and precisely what "a language without a Tier-2 shared registry is automatically quadratic" describes. A Rails app is exactly the kind of repo where F is large.

I want to be honest about the limits of this: **I am reasoning from the call structure, not from a measurement.** It may be perfectly acceptable at real Ruby project sizes — the constant factor is small and the class table is cheap per def. But that is a claim to check rather than assume, and right now nothing checks it.

## What I would like

Either a measurement, or — better, because it is permanent — **a Ruby fixture under `tests/fixtures/complexity/ruby/`**.

The complexity suite auto-discovers fixture directories at that path (`tests/test_complexity.c:167-175`) and gates on counter *ratios* at k versus 2k replicated copies, so a files×corpus coupling lands at ~4 and goes red. It needs no test edit — just fixture files. That is the mechanism this repo built for exactly this case, and a new language taking the filter exemption is the strongest reason to use it.

If the ratio comes back at ~2, this merges and the fixture stops the next person reintroducing the coupling. If it comes back at ~4, better to know now than after a Rails user reports a slow index.

## Smaller notes

**The split is clean and I checked the part that is easy to get wrong** — `ruby_lsp.{c,h}` appearing in both slices is the reviewer-requested move of `cbm_run_ruby_lsp_cross` out of the resolver slice, and you verified the halves reassemble byte-identically.

**Registering methods on both instance and singleton receiver keys** because `CBMLSPDef` carries no `def self.m` split is a reasonable over-registration, and worth the comment it has.

**Clearance:** the two flagged files here (`Makefile.cbm`, `scripts/check-lsp-originality.sh`) are #1711's commit, already reviewed and cleared on that PR — nothing new in this slice.

**Baseline note:** the `#1181` × `#1359` `cli` failures you documented are fixed on `main` now; I confirmed both pass in a full-suite run today.

PR #1713, opened 2026-08-18
docs: advertise Hybrid LSP support for Ruby

### LINKED ISSUE #1701: Add Hybrid LSP support for Ruby

benrfairless:
## What does this PR do?

Slice 5 of 5 splitting #1706. **Stacked on #1712** — see the review note below.

Ruby joins the Hybrid LSP cohort in the README: the badge moves from 10 to 11 language families (C and C++ count as one, as they already did), Ruby is added to both prose language lists, and the capability table gains a Ruby row.

This lands last on purpose. The review asked that documentation ship with the slice that actually establishes the advertised support level, and the row's cross-file claims — ActiveRecord model typing across files, superclass chains that span files — are only true once #1712 is in.

## Every claim in the row is backed by a test

While writing this I audited the row against the suite and found three capabilities named but untested: `prepend`, `method_missing` and `define_method`. All three are genuinely implemented, but advertising untested behaviour is exactly what this split is meant to prevent, so rather than weaken the wording I added coverage to #1711:

- `rubylsp_prepend_mixin` — a prepended module shadows the class's own method, and the shadowed method is *not* also emitted. Both methods exist in the fixture, so a resolver that walked the class first would still emit an edge, just the wrong one.
- `rubylsp_dynamic_dispatch_family_no_edge` — `public_send`, `__send__`, `define_method` and `instance_eval` each emit no edge (`send` already had its own row).

## What this PR deliberately does not touch

`docs/index.html` is left alone. Its structured data carries a pre-existing error — it says nine language families and omits Perl, both wrong before this work started. The Ruby edits would land on those exact strings, so adding Ruby without correcting them would mean authoring a sentence I know to be wrong, and correcting them here would mix an unrelated fix into a feature PR. It wants its own change; happy to file it.

## Reviewing just this slice

A PR from a fork can only target a branch in this repo, so the base is `main` and GitHub shows #1711 and #1712's commits too. **Only `af9e6200` is this slice** — a 4-line README change. Merge the parents and the diff collapses to it.

| # | PR | Depends on |
|---|----|------------|
| 1 | #1709 | — |
| 2 | #1710 | — |
| 3 | #1711 | — |
| 4 | #1712 | 3 |
| **5** | **this PR** | 4 |

### CI baseline

Resolves #1701

## Checklist

---


DeusData:
Approved. Landing the docs last, with the slice that actually establishes the advertised support level, was the right call and the row reads honestly.

**I checked the count rather than trusting it.** Eleven families with C and C++ as one, matching the prose list exactly: Python, TS/JS/JSX/TSX, PHP, C#, Go, C+C++, Java, Kotlin, Rust, Ruby, Perl. Ten before, eleven after. Correct in the badge and in both prose lists.

**And I checked your capability audit.** You said you found `prepend`, `method_missing` and `define_method` advertised but untested, and added coverage rather than softening the wording. All three are backed in the series, along with `public_send`, `__send__` and `instance_eval`. Auditing a doc row against the suite before advertising it is the right instinct, and it is the opposite of what usually happens to capability tables.

## Two things to pass on

**Your `docs/index.html` observation is confirmed.** It does say "9 language families", and it contains zero occurrences of "Perl". Both errors predate this work exactly as you said. Please do file it — after this merges the gap widens from 10-vs-9 to 11-vs-9, so it gets slightly worse for a reason that is not yours.

**More importantly: nothing gates this number.** I grepped `scripts/`, `tests/` and `.github/` and found no check that ties the Hybrid LSP badge to the actual set of LSP languages. The 158-grammar count *is* gated; this one is not. That is why `docs/index.html` could sit at nine while the README moved to ten without anything noticing.

That is not a change for this PR, and I am not asking you to make it. But it is worth recording alongside the `index.html` issue, because a hand-maintained count in three places with a gate on none of them will drift again — and the fix is cheap next to finding the drift later.

**Status:** `ms` is `DIRTY`, as expected for the top of a five-slice stack. It rebases naturally once the parents land.

Thanks for the whole series. Splitting an umbrella PR into five ordered slices, measuring which ones are genuinely independent, and correcting two of your own earlier claims along the way is a lot more work than shipping the original, and the review was far better for it.

Issue #1239, opened 2026-07-23
Elixir: grammar-layer remediation + Hybrid LSP resolver (tracking)


holsee:
## Summary

Bring Elixir to Hybrid LSP parity with the ten resolver-backed languages, per the design document `docs/elixir-lsp/PLAN.md` on branch `feat/elixir-hybrid-lsp`.

Two workstreams:

1. **Grammar-layer remediation** (defects D1–D8 in the plan): guard-clause defs dropped, enclosing-function attribution broken (calls attach to module QN), no name/arity identity, missing `defprotocol`/`defimpl`/`defstruct`/`defguard`/`defdelegate` forms, module-body directives invisible, unjoined nested module QNs, spurious variables from `binary_operator`, weak test contracts.
2. **`internal/cbm/lsp/elixir_lsp.{c,h}`**: per-file resolver (aliases/imports scoping, arity-keyed resolution, pipe/capture handling), then cross-file fallback tier, curated stdlib seed + `use` expansion table.

Empirical before/after evaluation protocol (objective graph metrics + tool-mediated rubric) is defined in the plan §5; baseline is recorded before the first code change lands.

Delivery is staged as PR-sized items (plan §4: PR-0a…PR-3b), each referencing this issue. Design-review feedback welcome on the plan document itself.

Clean-room note: Expert / ElixirLS / Lexical / Next LS will be added to `scripts/check-lsp-originality.sh` REFS and the scan run before any resolver PR.

holsee:
Development is proceeding on the fork (holsee/codebase-memory-mcp) for now; will re-open or reference an upstream issue when the work is ready to propose upstream per CONTRIBUTING.

PR #1721, opened 2026-08-19
fix(elixir): populate first_string_arg for Elixir calls

henry-hz:
## The bug

`tree-sitter-elixir` attaches **no field name** to a call's arguments node. Its entire field set is:

```
key, left, operand, operator, quoted_start, quoted_end, right, target, value
```

There is no `arguments` field. But `handle_calls()` in `internal/cbm/extract_calls.c` reads exactly that:

```c
TSNode args = ts_node_child_by_field_name(node, TS_FIELD("arguments"));
```

For Elixir this is **always null**, so `call.first_string_arg` was never populated for any Elixir call — not once, in any file. Nothing errored; the field was simply always `NULL`.

## Why it matters

Everything downstream that keys off a call's string argument was silently dead for Elixir:

- **Phoenix route paths** — `get "/wallets", WalletController, :index` never yielded a path
- **HTTP / async service URLs** — the `CBM_SVC_HTTP` / `CBM_SVC_ASYNC` classification needs the URL or topic literal
- **Config keys** — `Application.get_env("...")` is matched as `CBM_SVC_CONFIG`, which reads the same field

## The fix

The definition side has always known about this. `extract_defs.c` carries `elixir_call_args()` with a positional second-child fallback for precisely this reason:

```c
static TSNode elixir_call_args(TSNode node) {
    TSNode args = ts_node_child_by_field_name(node, TS_FIELD("arguments"));
    if (ts_node_is_null(args) && ts_node_child_count(node) > SECOND_CHILD_IDX) {
        args = ts_node_child(node, SECOND_CHILD_IDX);
    }
    return args;
}
```

This adds the same fallback on the call side. It is gated to `call` nodes, because Elixir's other call kinds — `dot`, and the `|>` `binary_operator` — carry no arguments node in that position and would otherwise pick up the operator token.

## Test

`elixir_call_string_argument` pins `first_string_arg` **directly** rather than asserting a downstream effect, so it stays meaningful independently of route or service classification.

## Verification

- The CALLS-breadth contract across 53 languages is unaffected (0 failures); the change is gated to `CBM_LANG_ELIXIR`
- `clang-format --dry-run --Werror` clean on the changed source. `tests/` was deliberately left unformatted, per the note that it is not in `LINT_SRCS`
- Exercised end-to-end against `elixir-plug/plug` and an 829-file Phoenix application

## Scope

Submitted as a **bug fix** under the CONTRIBUTING exception for focused bug fixes and test additions, so there is no prior issue. Additions only — 54 lines, no deletions, no reformatting.

This is the first of four independent Elixir fixes found while auditing Elixir extraction against a real Phoenix codebase. The other three are prepared and gated separately, and I will open them one at a time rather than as a bundle:

1. **this PR** — `first_string_arg` never populated
2. Phoenix channel extraction is unreachable in both branches (`arguments` field again, plus a dispatch on a `def` node type the grammar does not have)
3. Phoenix router macros mint no `Route` nodes (every `route_reg_suffixes` entry is prefixed `.` or `::`, but the macros are bare) — depends on this PR for the path argument
4. Elixir functions carry no complexity, fingerprint or line count

Happy to reorder, split further, or drop any of them if you would rather discuss the approach in an issue first.


henry-hz:
`test / test-windows-guards` is red here, and I do not believe it is this change. Recording the evidence rather than just asserting it.

The build step succeeded; the guard script failed:

```
SETUP FAIL: permanent daemon did not start for the crash check
REGRESSION: 1 green guard(s) went red: tests\windows\test_daemon_stability.py
```

`section_crash_recovery` never reached an assertion — it timed out at `daemon start` (60s budget) before the kill/recover sequence began. Four other guards in the same job also failed preconditions in that run:

- `test_hook_augment.py` — "SETUP FAIL: index did not run"
- `test_cli_non_ascii_arg.py` — "SETUP FAIL: ASCII control did not index via CLI"
- `test_ui_drive_listing.py` — "HTTP server did not start on port 53503"

Five independent subprocess/daemon startups failing in one Windows job reads like a sick runner rather than a code defect.

Why it should not be reachable from this diff:

- `tests/windows/test_daemon_stability.py` contains no Elixir
- the change is 54 added lines, all inside `if (ctx->language == CBM_LANG_ELIXIR)`, altering only how a call's arguments node is located during extraction
- it touches no process startup, IPC, socket, or path code

Everything else on the PR is green: DCO, `lint`, `lint-mem`, `codeql-gate`, `security-static`, `license-gate`, all `test-unix` shards (ubuntu ×3, ubuntu-arm ×3, macos-14, macos-15-intel), all three `test-tsan` variants, `test-lsan-macos`, `test-package-wrappers`, and `pr-smoke` on all three platforms.

I do not have rerun rights on this repo — could a maintainer re-run that single job? Happy to push an empty commit to re-trigger instead if you would rather not spend the click, and equally happy to dig further if you have seen this guard fail for a real reason before.

henry-hz:
Context for this change now lives in #1729, which lays out four Elixir extraction defects found while auditing against a real Phoenix application, and how they relate to the #1239 / `docs/elixir-lsp/PLAN.md` workstream. This PR is defect 1 of 4.

Refs #1729

DeusData:
Reran the Windows guard job for you — and you were right that it is not your diff, so let me give you the mechanism rather than just agreeing.

`tests/windows/test_daemon_stability.py` is a known Windows-only flake. The harness kills daemons with `taskkill /F`, and on Windows a hard kill leaves the process's **byte-range locks** held until the OS reclaims them "as resources allow" — not promptly and not deterministically. The next client to start hits `ERROR_LOCK_VIOLATION`, reads it as BUSY, and is refused. That is why you saw five independent daemon/subprocess setup failures in one job, why it is always Windows, and why a three-line README PR has been seen failing the same guard. #1772 is the fix in flight: it waits out a held startup transition instead of giving up at a fixed deadline. Your reading of "sick runner" was the correct diagnosis from the outside.

**The change itself is right, and the test is the good kind.** What I want to credit specifically is `ASSERT_EQ(1, seen)` at the end. Without it, `ASSERT_NOT_NULL(first_string_arg)` inside a loop passes vacuously if the call is never extracted at all — the failure mode where a green test proves nothing. You closed that yourself. This project has been bitten by exactly that shape before, and a contributor who guards against it unprompted saves a reviewer the archaeology.

I verified your premise too: `elixir_call_args()` at `extract_defs.c:5067` really does apply the positional fallback, so the definition side has always done this and the call side simply never did. The gate is tight — `ts_node_is_null(args) && ctx->language == CBM_LANG_ELIXIR && strcmp(ts_node_type(node), "call") == 0` cannot reach another language, and it sits ahead of the ObjectScript fallback without disturbing it.

**Two things before I merge:**

1. **A stale comment.** The new comment says `cbm_elixir_call_args is the shared second-child fallback` — that symbol does not exist anywhere in the tree. Your actual function is `elixir_call_arguments_fallback`, and the pre-existing one is `elixir_call_args`. It looks like a shared-helper design that got inlined per-file; the comment would send the next reader looking for something that was never written.

2. **A question rather than a request, because it spans your three PRs.** With #1730 also in flight there would be three near-identical positional-args helpers: `elixir_call_args` (extract_defs.c), your new `elixir_call_arguments_fallback` (here), and `elixir_channel_call_args` (#1730). The stale comment suggests you originally intended one shared helper. Would you rather land these as-is and consolidate in a follow-up, or hoist a single helper now? I have no strong preference — three copies of six lines is not a crisis — but you clearly thought about it once and I would rather ask than pick for you.

**On your ordering question: yes, #1721 first.** Nothing forces it, but your unopened defect 3 (Phoenix router macros minting no `Route` nodes) depends on the path argument this PR provides, so it is the natural root.

**One heads-up that affects all three.** Your production files are genuinely disjoint — `extract_calls.c`, `extract_channels.c`, `extract_defs.c` — so nothing here can break anything there. But all three insert into `tests/test_extraction.c` at the *same two anchors*: immediately after `TEST(elixir_function)` closes, and immediately after `RUN_TEST(elixir_function);` in the suite. GitHub reports all three MERGEABLE because it compares each against `main` in isolation. I expect the first merge to flip the other two to CONFLICTING at both anchors. It is a trivial rebase, just not a surprise worth having.

Thank you for the tracking issue and for splitting this into four reviewable pieces instead of one large one — that is exactly how I would want a language's defects to arrive.

DeusData:
Closing the loop: the Windows startup-coordination fix has **merged** as `7b6363d0` (#1772). That is the fix for the byte-range-lock flake you hit — the daemon now waits out a held startup transition instead of giving up at a fixed deadline.

Notably it merged with `test / test-windows-guards` green, which is the job that had been flaking. Future runs on your branch should stop showing it once you pick up current `main`. Thank you for reporting it precisely rather than just re-running.

DeusData:
Approved. The diagnosis is exactly right, and I verified it properly rather than by the obvious grep — which would have told me the opposite.

Searching `elixir/parser.c` for the string `"arguments"` finds three hits, so a quick check makes it look like the field exists. It does not. Those are **node type** names (`sym__call_arguments_with_parentheses` and friends, displayed as "arguments"). The `ts_field_names` table is the authority, and it holds exactly the nine you listed:

```
key, left, operand, operator, quoted_end, quoted_start, right, target, value
```

No `arguments`. So `ts_node_child_by_field_name(node, "arguments")` really did return null for every Elixir call, and `first_string_arg` really was never populated — not once, as you say. Listing the complete field set in the description is what made this checkable in a minute instead of an hour.

**Reusing the shape the definition side already had is the right call.** `elixir_call_args()` has been sitting in `extract_defs.c:5237` with precisely this fallback, which means the repo already learned this lesson once and only applied it on one side. Two instances of one named pattern is a much better outcome than a second ad-hoc special case.

**Gating to `call` nodes matters and you said why** — `dot` and the `|>` `binary_operator` carry no arguments node in that position and would otherwise pick up the operator token. That is the sentence that turns a positional `child(1)` from a guess into a rule.

**And the test pins `first_string_arg` directly rather than a downstream route or service assertion.** That is the right choice: a test that asserted a Phoenix route would have passed or failed for reasons that have nothing to do with this fix, and would go stale the moment route classification changes.

**On sequencing the other three:** opening them one at a time rather than as a bundle is exactly right, and item 3 depending on this one is worth the note. No need to reorder — the order you have puts the enabling fix first.

**Status:** cleared, reviewed, approved. `main` moved twice today (broken, repaired by #1993, then #1703 merged), so I will update this branch and re-run before merging. Nothing needed from you.

Exercising it against `elixir-plug/plug` and an 829-file Phoenix application before submitting is what makes "every Elixir call, always" credible rather than a claim.

DeusData:
Merged as `ffb5ebe9e8ad` — thank you, and this one was a pleasure to verify.

The obvious check is actively misleading: grepping `elixir/parser.c` for `"arguments"` finds three hits, which *looks* like the field exists. Those are node **type** display names (`sym__call_arguments_with_parentheses` and friends). The authority is the `ts_field_names` table, which holds exactly the nine fields you listed — `key, left, operand, operator, quoted_end, quoted_start, right, target, value` — and no `arguments`. So `ts_node_child_by_field_name(node, "arguments")` really did return null for every Elixir call, and `first_string_arg` was never populated: Phoenix route paths, HTTP/async service URLs and config keys all silently missing for Elixir.

What I liked most is that the fix reuses a house pattern instead of inventing one — `elixir_call_args()` in `extract_defs.c` has carried exactly this positional second-child fallback for a while; the repo learned the lesson once and applied it on only one side. Gating it to `call` nodes (so `dot` and the `|>` binary operator do not pick up the operator token) is the right restraint, and pinning `first_string_arg` directly rather than a downstream route assertion means the test cannot go stale when classification changes.

#1730 and #1731 each need the small keep-both resolution I described on them — merge on green after that.

PR #1725, opened 2026-08-19
feat(index): enforce opt-in storage resource limits

liuchong:
> **Stacked on #1724.** A fork PR must target a branch in this repository, so the diff below also contains the two commits under it. The new work here is the last commit, `bd6ea29d`. Once #1724 merges, this diff shrinks to that one commit on its own.

Related to #1347.

## Problem

An index that fits in memory and finishes in time can still fill the disk. Publication needs room for the staging artifacts and the final database at the same time, and running out of space during publication is the one failure that can cost a working index.

## What this changes

Two opt-in limits, both defaulting to `off`:

| Key | Meaning |
| --- | --- |
| `index_cache_max_mb` | Maximum total size of the cache directory |
| `index_min_free_disk_mb` | Minimum free space that must remain on the cache volume |

They are measured before staging and again before publication, because the interesting moment is the one where both copies exist.

Three internal ceilings — final database, staging artifacts, task temporary directory — are added alongside them. They deliberately have no public keys: they are only meaningful as part of one composed decision about how much room a publication may take, which no single operator-facing key can express.

Crossing any of them fails the attempt **before** the old database is touched. Atomic publication is unchanged and the previous index keeps serving.

Two details worth review attention:

- Staging cleanup is scoped by a private per-task token, so a worker removes only the artifacts it created. Cleanup that matched on path alone could delete a concurrent run's staging files.
- An old database is treated as replaceable only after an integrity verdict distinguishes genuine corruption from a transient `SQLITE_BUSY`. Treating busy as corrupt would discard a healthy index.

Probes fail closed: an enabled measurement that cannot be completed fails the attempt rather than being assumed to pass.

## Testing

`make -f Makefile.cbm test` and `make -f Makefile.cbm lint-ci` on macOS. New coverage: directory-size and free-space measurement on each platform path, admission at both checkpoints, task-token-scoped cleanup leaving unrelated cache files intact, preservation of the old index on rejection, and the busy-versus-corrupt integrity verdict.

## Stack

1. #1723 — discovery limits
2. #1724 — worker watchdogs
3. **this PR** — storage limits
4. #1726 — resource profiles

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -251,6 +251,7 @@ FOUNDATION_SRCS = \
     src/foundation/profile.c \
     src/foundation/dump_verify.c \
     src/foundation/limits.c \
+    src/foundation/index_policy.c \
     src/foundation/subprocess.c \
     src/foundation/sha256.c \
     src/foundation/secure_random.c \
@@ -532,6 +533,7 @@ TEST_FOUNDATION_SRCS = \
     tests/test_str_intern.c \
     tests/test_log.c \
     tests/test_str_util.c \
+    tests/test_index_policy.c \
     tests/test_workspace.c \
     tests/test_platform.c \
     tests/test_diagnostics.c \
--- a/README.md
+++ b/README.md
@@ -735,10 +735,12 @@ codebase-memory-mcp config set auto_watch false          # don't register backgr
 codebase-memory-mcp config set watcher_enabled false     # stop the watcher thread entirely (default: true)
 codebase-memory-mcp config set index_max_files 250000    # optional per-index source-file limit
 codebase-memory-mcp config set index_max_source_mb 16384 # optional per-index source-size limit
+codebase-memory-mcp config set index_max_rss_mb 8192     # optional worker-tree current RSS limit
+codebase-memory-mcp config set index_max_duration_seconds 3600 # optional total worker duration
 codebase-memory-mcp config reset auto_index              # reset to default
 ```
 
-The two `index_max_*` settings default to `off`. Exceeding one fails the complete
+The four `index_max_*` settings default to `off`. Exceeding one fails the complete
 index attempt rather than publishing a partial graph; an existing serving index
 is preserved. See [Index resource limits](docs/INDEX_RESOURCE_LIMITS.md).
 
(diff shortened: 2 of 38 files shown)


DeusData:
Reviewed slice 3. I said on #1723 that this is where the configuration surface starts growing and that these slices would get a closer look. It holds up, and the restraint is the first thing worth naming.

**You added two public keys and deliberately withheld three.** The final-database, staging-artifact and task-temp ceilings have no operator-facing keys because, as you put it, they are "only meaningful as part of one composed decision about how much room a publication may take, which no single operator-facing key can express." That is the correct instinct — three more knobs would have been easier to write and much worse to live with, because each one becomes a permanent contract the moment someone sets it.

**Both details you flagged for attention check out. I verified them rather than taking them, because both are destructive.**

**The busy-versus-corrupt verdict.** `CBM_INTEGRITY_TRANSIENT` returns *not replaceable*, so a `SQLITE_BUSY` database is never deducted as reclaimable — a healthy index cannot be discarded because it happened to be in use. Two things make this better than it needed to be: you reuse the existing `cbm_store_check_integrity_verdict` rather than inventing a second busy-versus-corrupt notion, and the branch above it refuses a symlinked destination outright. A symlinked "old database" being judged replaceable would have been an unpleasant way to lose someone's data.

The comment on the unopenable case is right too — keeping its bytes in the projection rather than blocking repair of a confirmed-corrupt destination is the conservative direction.

**The task-token cleanup.** The prefix is `basename + marker + token + "."`, so a concurrent run's staging files cannot match. The part that makes that safe rather than merely intended is `bootstrap_worker_stage_token_valid`: a strict allowlist of `[0-9A-Za-z_-]`, length 6 to 63, with no `.`, no path separator and no glob metacharacter. Since the token arrives through `--index-worker-stage-token` on argv, it is attacker-influenced input that becomes part of a filesystem prefix used for deletion — and a permissive validator there would have been a real hole. The overflow guards on the prefix arithmetic before the `snprintf` are the right shape too.

**Measuring at both checkpoints** — before staging and again before publication — is correct for the stated reason: the interesting moment is the one where both copies exist. And failing before the old database is touched, with atomic publication unchanged, means the worst case is "no new index" rather than "no index".

## Status

**Your only red is not yours.** `test / test-windows-guards` fails because that lane's own setup cannot index a plain ASCII fixture repo — it reports `nodes: None, edges: None` — so every case needing an indexed repo skips on a failed precondition. I have now confirmed the identical cluster on three unrelated pull requests across four days. It is ours; please do not chase it.

**Clearance** is `REVIEW(3)` on the same files inherited from the slices below (`Makefile.cbm`, `scripts/test-runtime.sh`, plus the process-spawning watchdog test). Nothing new in this slice; I will get the marker written before merge.

**Rebase when #1724 lands** and this collapses to `bd6ea29d` alone.

PR #1726, opened 2026-08-19
feat(index): add opt-in index resource profiles

liuchong:
> **Stacked on #1725.** A fork PR must target a branch in this repository, so the diff below also contains the three commits under it. The new work here is the last commit, `2a6a6014`. Once #1725 merges, this diff shrinks to that one commit on its own.

Related to #1347.

## Problem

The previous three PRs add six independent keys. That is an honest interface and a poor default: an operator who only wants indexing to stay within reason has to learn all six, choose a value for each, and keep them consistent as the machine changes.

## What this changes

One key, `index_resource_profile`, accepting `off` (default), `balanced` or `strict`.

- **balanced** guards the host. It lets an index use what the machine can spare and stops a runaway worker before the host is exhausted.
- **strict** holds indexing to the daemon's own budget. A repository that needs more fails fast and attributed instead of finishing at the host's expense. On a very large repository this is expected to fail, and the failure names the limit and the observed value so the operator can choose `balanced`, raise the individual key, or leave the profile `off`.

Any individual key, including an explicit `off`, replaces that dimension of the selected profile. A profile is a baseline, not a lock.

**The balanced worker memory ceiling is derived from detected host memory rather than tabled.** This is the decision most worth challenging in review. Large-repository indexing peaks in the tens of gigabytes, so any round number low enough to feel safe would reject repositories that index successfully today — a "safety" feature whose main effect is breaking working setups. The host share is the only balanced answer that scales with the machine. When host memory cannot be read, balanced falls back to a fixed floor.

Profiles also bound directory count, entry count, traversal depth and discovery time. These have no individual keys for the same reason as the storage internals in #1725: they are only meaningful inside a composed decision.

The resolved profile travels through the same trusted worker contract as the individual limits, so the CLI, the daemon and the supervised worker enforce the same decision.

## Testing

`make -f Makefile.cbm test` and `make -f Makefile.cbm lint-ci` on macOS. New coverage: exact profile baselines for `balanced` and `strict`, the host-share derivation including the unreadable-host fallback, per-key override precedence including explicit `off`, the four discovery dimensions with their rejection paths, and profile propagation through the worker contract.

## Stack

1. #1723 — discovery limits
2. #1724 — worker watchdogs
3. #1725 — storage limits
4. **this PR** — resource profiles

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -251,6 +251,7 @@ FOUNDATION_SRCS = \
     src/foundation/profile.c \
     src/foundation/dump_verify.c \
     src/foundation/limits.c \
+    src/foundation/index_policy.c \
     src/foundation/subprocess.c \
     src/foundation/sha256.c \
     src/foundation/secure_random.c \
@@ -532,6 +533,7 @@ TEST_FOUNDATION_SRCS = \
     tests/test_str_intern.c \
     tests/test_log.c \
     tests/test_str_util.c \
+    tests/test_index_policy.c \
     tests/test_workspace.c \
     tests/test_platform.c \
     tests/test_diagnostics.c \
--- a/README.md
+++ b/README.md
@@ -735,10 +735,12 @@ codebase-memory-mcp config set auto_watch false          # don't register backgr
 codebase-memory-mcp config set watcher_enabled false     # stop the watcher thread entirely (default: true)
 codebase-memory-mcp config set index_max_files 250000    # optional per-index source-file limit
 codebase-memory-mcp config set index_max_source_mb 16384 # optional per-index source-size limit
+codebase-memory-mcp config set index_max_rss_mb 8192     # optional worker-tree current RSS limit
+codebase-memory-mcp config set index_max_duration_seconds 3600 # optional total worker duration
 codebase-memory-mcp config reset auto_index              # reset to default
 ```
 
-The two `index_max_*` settings default to `off`. Exceeding one fails the complete
+The four `index_max_*` settings default to `off`. Exceeding one fails the complete
 index attempt rather than publishing a partial graph; an existing serving index
 is preserved. See [Index resource limits](docs/INDEX_RESOURCE_LIMITS.md).
 
--- a/README.md
+++ b/README.md
@@ -733,6 +733,7 @@ codebase-memory-mcp config set auto_index true           # auto-index on session
 codebase-memory-mcp config set auto_index_limit 50000    # max files for auto-index
 codebase-memory-mcp config set auto_watch false          # don't register background git watcher (default: true)
 codebase-memory-mcp config set watcher_enabled false     # stop the watcher thread entirely (default: true)
+codebase-memory-mcp config set index_resource_profile balanced # opt-in composed resource limits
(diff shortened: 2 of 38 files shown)


DeusData:
Reviewed slice 4. You asked for the host-share derivation to be challenged, so I went at that first — and it survives. The finding I do have is somewhere else.

## The derivation is right, and our own numbers back it

> any round number low enough to feel safe would reject repositories that index successfully today

That is not a hypothesis here, it is measured. A recorded large-repository run in this project peaks around **16.5 GB** resident, and a doubled workload OOMed a 36 GB host. Any tabled ceiling comfortable enough to look safe — 4 GB, 8 GB — would refuse an index that works today, and the operator's reward for enabling a safety feature would be a broken setup. The host share is the honest answer, and 80% is a defensible place to put it.

The arithmetic is careful too: `(x / 5) * 4 + ((x % 5) * 4) / 5` computes `floor(4x/5)` without ever forming `4x`, so it cannot overflow. The unreadable-host fallback of `max(soft_budget × 2, 8 GiB)` guards its own doubling. And `strict` capping at 8 GiB regardless of a larger soft budget is consistent with what strict is for.

Withholding individual keys for directory count, entry count, traversal depth and discovery time is the same restraint as #1725's storage internals, and right for the same reason.

## The finding: the description promises something the code deliberately does not do

> Any individual key, including an explicit `off`, replaces that dimension of the selected profile. **A profile is a baseline, not a lock.**

That holds for every dimension except one. The final clamp sits *outside* the override guard:

```c
if (policy->max_rss_bytes.enabled && host_cap > 0 &&
    policy->max_rss_bytes.value > host_cap) {
    policy->max_rss_bytes.value = host_cap;
}
```

So an operator who explicitly sets `index_max_rss_mb` above 80% of host memory has it silently reduced. On a 36 GB machine, an explicit 32 GB becomes 28.8 GB and nothing says so.

I want to be clear I do **not** think this is a bug — `index_policy_host_clamp_only_tightens_explicit_rss` tells me it is deliberate and pinned. As a safety rail it is defensible: an explicit ceiling above what the host can survive is not really a ceiling.

Two small things would settle it:

**Say so in the description.** "A profile is a baseline, not a lock" is the sentence an operator will read and remember, and for `max_rss` it is not true. One clause — that the host share is a hard ceiling no individual key can exceed — makes the contract match the code.

**And say so at runtime when it bites.** Right now the clamp is silent. This repo has just spent four pull requests removing exactly this shape — a value quietly replaced by a different one, with nothing telling the caller to look. A log line naming the requested value and the applied one, emitted only when the clamp actually tightens something, would cost almost nothing and keeps this consistent with that work.

Also worth considering: the clamp applies whenever `max_rss` is enabled and host memory is readable, which includes `profile = off` with an explicit key. That is the case most likely to surprise, because the operator never opted into a profile at all.

## Status

Clearance is the same inherited set as the slices below; I will handle the marker before merge. Rebase when #1725 lands and this collapses to `2a6a6014` alone.

DeusData:
The separator-spelling fix split out of this PR landed on main as 971111c7 (#2070), credited to you as co-author. When you next rebase this branch that hunk will drop out on its own; the feature stays under discussion here as before.

PR #1727, opened 2026-08-19
feat(index): record and report the latest index attempt

liuchong:
> **Stacked on #1726.** A fork PR must target a branch in this repository, so the diff below also contains the four commits under it. The new work here is the last commit, `aa1b130e`. Once #1726 merges, this diff shrinks to that one commit on its own.

Related to #1347.

## Problem

Everything the previous PRs add can fail an index attempt, and a failed attempt is currently a log line. A watcher-triggered rebuild that dies on a resource limit leaves the previously published graph in service and nothing durable behind it. `index_status` reports a healthy generation, and the operator has no way to learn that it is no longer the tree.

That gap is what makes the limits usable or dangerous: a limit whose enforcement is invisible converts a loud failure into a quiet wrong answer.

## What this changes

One attempt record per project under the cache directory, updated under a file lock across `queued`, `running` and terminal states, carrying the origin, timestamps, the effective profile source and the resource failure when there is one. `index_status` returns it as `last_index_attempt`, and `delete_project` removes a matching record when no database was published.

`freshness` is reported alongside it and is deliberately conservative:

- `fresh` only when matching **clean** Git snapshots were observed before and after indexing and the worktree is still clean at that `HEAD`
- `stale` for a different clean `HEAD`, or a failed watcher rebuild after an observed change
- `unknown` for non-Git roots, dirty trees, a tree that changed during indexing, a failed Git probe, or a corrupt record

A dirty tree is never treated as evidence in either direction. Reporting `stale` for every dirty tree would make the field noise, and reporting `fresh` would be a lie.

The Git probe is read-only (`rev-parse` plus `status --porcelain` with `--no-optional-locks`) and never mutates the repository.

Daemon startup fails attempts whose worker no longer exists. Liveness is judged by process identity rather than the pid alone, so a recycled pid cannot keep a dead attempt alive; when a start time cannot be read the code falls back to plain liveness rather than declaring the worker gone.

Projects with no attempt record keep the previous response shape.

## Testing

`make -f Makefile.cbm test` and `make -f Makefile.cbm lint-ci` on macOS. New coverage: record lifecycle and locked transitions, resource-failure metadata in `index_status`, every freshness verdict including the changed-during-index and dirty cases, read-only Git snapshot behaviour, abandoned-attempt recovery at daemon startup, and record removal on `delete_project`.

## Stack

1. #1723 — discovery limits
2. #1724 — worker watchdogs
3. #1725 — storage limits
4. #1726 — resource profiles
5. **this PR** — attempt record and freshness

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -251,6 +251,7 @@ FOUNDATION_SRCS = \
     src/foundation/profile.c \
     src/foundation/dump_verify.c \
     src/foundation/limits.c \
+    src/foundation/index_policy.c \
     src/foundation/subprocess.c \
     src/foundation/sha256.c \
     src/foundation/secure_random.c \
@@ -532,6 +533,7 @@ TEST_FOUNDATION_SRCS = \
     tests/test_str_intern.c \
     tests/test_log.c \
     tests/test_str_util.c \
+    tests/test_index_policy.c \
     tests/test_workspace.c \
     tests/test_platform.c \
     tests/test_diagnostics.c \
--- a/README.md
+++ b/README.md
@@ -735,10 +735,12 @@ codebase-memory-mcp config set auto_watch false          # don't register backgr
 codebase-memory-mcp config set watcher_enabled false     # stop the watcher thread entirely (default: true)
 codebase-memory-mcp config set index_max_files 250000    # optional per-index source-file limit
 codebase-memory-mcp config set index_max_source_mb 16384 # optional per-index source-size limit
+codebase-memory-mcp config set index_max_rss_mb 8192     # optional worker-tree current RSS limit
+codebase-memory-mcp config set index_max_duration_seconds 3600 # optional total worker duration
 codebase-memory-mcp config reset auto_index              # reset to default
 ```
 
-The two `index_max_*` settings default to `off`. Exceeding one fails the complete
+The four `index_max_*` settings default to `off`. Exceeding one fails the complete
 index attempt rather than publishing a partial graph; an existing serving index
 is preserved. See [Index resource limits](docs/INDEX_RESOURCE_LIMITS.md).
 
--- a/README.md
+++ b/README.md
@@ -733,6 +733,7 @@ codebase-memory-mcp config set auto_index true           # auto-index on session
 codebase-memory-mcp config set auto_index_limit 50000    # max files for auto-index
 codebase-memory-mcp config set auto_watch false          # don't register background git watcher (default: true)
 codebase-memory-mcp config set watcher_enabled false     # stop the watcher thread entirely (default: true)
+codebase-memory-mcp config set index_resource_profile balanced # opt-in composed resource limits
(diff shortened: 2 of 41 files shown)


DeusData:
Reviewed slice 5. The attempt record is the right idea and I want it. There is a collision to resolve first, and the decision does not go your way on one field — so let me be specific about which part and why.

## The collision

**#1561 also adds a top-level `freshness` to `index_status`**, by a different contributor, and the two are not merely different — they are different *types*:

| | #1561 | this PR |
|---|---|---|
| shape | object | string |
| values | `verdict` ∈ current / stale / unknown, plus `reason`, `reasons[]`, `recommended_action` | `fresh` / `stale` / `unknown` |
| evidence | `indexed_checkout_sha` recorded in a new column at the staged-generation boundary, vs live HEAD and bounded worktree status | matching clean snapshots before *and* after indexing, worktree still clean at that HEAD |

Whichever merged second would overwrite the other, and a client parsing `freshness` would get an object or a string depending on merge order. Neither of you could have seen this: the PRs are independent and git would merge them without a conflict.

**Maintainer decision: #1561's object stands, and this PR drops its `freshness` field.** The reason is output richness rather than correctness — a caller that learns *why* the graph is stale and what to do about it is better served than one that gets a single word, and #1561 already carries `reason`, `reasons[]` and `recommended_action`. It is also standalone and already reviewed clean, so it can land without waiting on a six-PR stack.

**Keep `last_index_attempt`.** That is the genuinely new contribution here and nothing else provides it. A durable per-project record under a file lock, carrying origin, timestamps, effective profile source and the resource failure, is precisely what turns "the limits fired" from a log line into something an operator can query. Your own framing is the argument for it:

> a limit whose enforcement is invisible converts a loud failure into a quiet wrong answer

That is exactly right, and it is why the record matters more than the verdict word.

## Two things from your design that should survive into the merged shape

**"A dirty tree is never treated as evidence in either direction."** Reporting `stale` for every dirty tree makes the field noise; reporting `fresh` is a lie. #1561 reaches the same conclusion by a different route, and your sentence states it better than anything currently in that PR. Worth carrying into the documentation.

**Read-only Git probing** — `rev-parse` plus `status --porcelain --no-optional-locks` — is the right way to look at a repository you are only reporting on. `--no-optional-locks` in particular is the detail most people miss.

## The rest of the slice

Judging worker liveness by **process identity rather than pid alone**, so a recycled pid cannot keep a dead attempt alive, is the same instinct as #1724's handle-based watchdog and #1874's `SYNCHRONIZE` handle — and falling back to plain liveness when a start time cannot be read is the right direction for that failure. Failing abandoned attempts at daemon startup closes the loop.

Preserving the previous response shape for projects with no attempt record is the correct default.

## Status

Please drop the `freshness` field and rebase once #1726 lands. Your reds are `security / codeql-gate`, `test-windows-guards` and `ci-ok` — the middle one is a lane-wide precondition failure we have now confirmed on four unrelated PRs, and is not yours.

Sorry to ask you to remove working code. The field genuinely can only exist once.

PR #1728, opened 2026-08-19
feat(mcp): warn on answers served from a failed rebuild

liuchong:
> **Stacked on #1727.** A fork PR must target a branch in this repository, so the diff below also contains the five commits under it. The new work here is the last commit, `3d51b655`. Once #1727 merges, this diff shrinks to that one commit on its own.

This implements the review suggestion on #1348 to surface a stale index on the next MCP response rather than only in `index_status`.

Related to #1347.

## Problem

A recorded attempt is only visible to whoever thinks to ask for it. An agent that calls `search_graph` after a background rebuild failed receives a confident answer from a graph that no longer matches the tree, and nothing in that response says so. Agents do not poll `index_status` before every query.

## What this changes

While the recorded attempt is `failed` or `cancelled`, answers served from that project's graph carry a warning naming the origin, the recorded finish time, and the limit that ended the attempt — enough for the caller to decide whether to trust the answer and what to fix.

- JSON payloads gain a `stale_index_warning` field, so a structured reader cannot miss it
- Text and tree payloads gain a trailing content block
- The answer itself keeps its position and content; `content[0]` is never displaced, because `hook_augment` and the smoke tests read it as the payload

Annotated: `search_graph`, `query_graph`, `trace_path`, `trace_call_path`, `get_architecture`, `get_code_snippet`, `search_code`. Not annotated: `index_status` and `check_index_coverage` already report freshness themselves, and `index_repository` is the remedy.

A queued or running rebuild is not yet a failure and stays silent; a completed rebuild clears the warning. Every failure inside the warning path is silent — a warning that cannot be produced must never damage an answer that is otherwise fine.

Cost is one small record read per answered query. It probes neither Git nor the working tree.

## Testing

`make -f Makefile.cbm test` and `make -f Makefile.cbm lint-ci` on macOS. New coverage: the warning on both JSON and text payloads with the answer intact, the attributed cause text, silence for `index_status`, silence while an attempt is queued or running, and clearing after a completed rebuild.

## Stack

1. #1723 — discovery limits
2. #1724 — worker watchdogs
3. #1725 — storage limits
4. #1726 — resource profiles
6. **this PR** — stale-index warning on answers

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -251,6 +251,7 @@ FOUNDATION_SRCS = \
     src/foundation/profile.c \
     src/foundation/dump_verify.c \
     src/foundation/limits.c \
+    src/foundation/index_policy.c \
     src/foundation/subprocess.c \
     src/foundation/sha256.c \
     src/foundation/secure_random.c \
@@ -532,6 +533,7 @@ TEST_FOUNDATION_SRCS = \
     tests/test_str_intern.c \
     tests/test_log.c \
     tests/test_str_util.c \
+    tests/test_index_policy.c \
     tests/test_workspace.c \
     tests/test_platform.c \
     tests/test_diagnostics.c \
--- a/README.md
+++ b/README.md
@@ -735,10 +735,12 @@ codebase-memory-mcp config set auto_watch false          # don't register backgr
 codebase-memory-mcp config set watcher_enabled false     # stop the watcher thread entirely (default: true)
 codebase-memory-mcp config set index_max_files 250000    # optional per-index source-file limit
 codebase-memory-mcp config set index_max_source_mb 16384 # optional per-index source-size limit
+codebase-memory-mcp config set index_max_rss_mb 8192     # optional worker-tree current RSS limit
+codebase-memory-mcp config set index_max_duration_seconds 3600 # optional total worker duration
 codebase-memory-mcp config reset auto_index              # reset to default
 ```
 
-The two `index_max_*` settings default to `off`. Exceeding one fails the complete
+The four `index_max_*` settings default to `off`. Exceeding one fails the complete
 index attempt rather than publishing a partial graph; an existing serving index
 is preserved. See [Index resource limits](docs/INDEX_RESOURCE_LIMITS.md).
 
--- a/README.md
+++ b/README.md
@@ -749,7 +749,9 @@ explicit value, including `off`, replaces that dimension of a selected
 profile. Exceeding an effective limit fails the complete index attempt rather
 than publishing a partial graph, and an existing serving index is preserved.
 `index_status` reports the latest physical indexing attempt and Git-based
-freshness after the first recorded attempt. See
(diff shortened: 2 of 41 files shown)


DeusData:
Reviewed slice 6, the last of the programme. The mechanism is right and carefully bounded. One thing to reconcile before it lands, and it comes from a decision made on a neighbouring PR after you wrote this.

## What is right

**`content[0]` is never displaced, and you say why** — `hook_augment` and the smoke tests read it as the payload. That is the kind of compatibility constraint that is invisible until it breaks someone, and checking existing consumers before appending is the difference between a warning and an outage.

**Every failure inside the warning path is silent.** "A warning that cannot be produced must never damage an answer that is otherwise fine" is the correct fail direction for an advisory, and the opposite of the fail-closed choice in #1724 — which is right, because those two things carry different consequences.

**The non-annotation list is precise.** `index_status` and `check_index_coverage` already report freshness, and `index_repository` is the remedy rather than a victim. Leaving those alone avoids a server that nags about a problem while you are fixing it.

**And it probes neither Git nor the working tree** — one small record read per answered query. A warning that cost a `git status` on every `search_graph` would not survive contact with an agent loop.

## The thing to reconcile: two staleness signals that can disagree

Your warning fires on exactly one condition — the recorded attempt's `state` is `failed` or `cancelled`. No comparison against the tree.

Meanwhile #1561 adds a `freshness` object to `index_status` whose verdict compares the `indexed_checkout_sha` recorded with the graph against live `HEAD` and the worktree state. That PR's design was confirmed as the one we keep, so both will exist on the same server.

Consider a background rebuild that fails on a resource limit and publishes nothing, while the previously published graph still matches `HEAD` exactly:

- `index_status` reports `freshness.verdict = current` — the graph does match the tree
- `search_graph` carries `stale_index_warning` — the last attempt failed

Both are correct about what they measure, and together they tell a caller two different things. The field name is what makes it jarring: **`stale_index_warning` asserts staleness, but the evidence is attempt failure.** A failed rebuild means the graph *may* not reflect recent work; it does not establish that it is stale, and sometimes it demonstrably is not.

I do not think the mechanism should change — warning on a failed rebuild is exactly what an agent needs, and requiring a tree comparison would reintroduce the per-query Git probe you correctly avoided. What I would like is for the name and the text to claim only what the evidence supports. Something closer to `last_index_attempt_failed`, with the message saying the last rebuild failed and the graph may therefore be missing recent changes, would be true in every case the current name overstates.

Your message text already leans that way — it names the origin, the finish time and the limit rather than asserting divergence. It is mainly the field key that overclaims, and a structured reader keys on that.

## Status

Rebase when #1727 lands and this collapses to `3d51b655`. Note that #1727 has been asked to drop its own `freshness` string for the same collision reason — its `last_index_attempt` record, which this PR reads, is unaffected and stays.

This is a good close to the series. The programme is coherent end to end: limits that can fail an attempt, a watchdog that enforces them, a record that survives the failure, and finally a signal that reaches the caller who never thought to ask.

Issue #1729, opened 2026-08-19
Elixir: four extraction defects outside the #1239 D1-D8 list (two recorded as working today)


henry-hz:
## Summary

While auditing Elixir extraction against a real 829-file Phoenix application, I found four defects that are **not** in the D1–D8 list of #1239 / `docs/elixir-lsp/PLAN.md`. Two of them contradict things that plan records as working today, so I am raising them here before they cost @holsee debugging time.

I also have a fork with a broader Elixir grammar-layer rewrite. **I am not proposing that upstream** — see "Overlap" at the bottom. This issue is about the four defects and about making my other work available if it is useful.

### 1. `first_string_arg` is never populated for any Elixir call

`tree-sitter-elixir` attaches no field name to a call's arguments node — its whole field set is `key, left, operand, operator, quoted_start, quoted_end, right, target, value`. `handle_calls()` reads `ts_node_child_by_field_name(node, "arguments")`, which is therefore always null for Elixir.

Consequence: route paths, HTTP/async service URLs and config keys are all dead for Elixir, silently. `extract_defs.c` has always carried a positional fallback (`elixir_call_args()`) for exactly this reason; the call side never got one.

→ PR #1721 (open, 54 lines, additions only).

### 2. Phoenix channel extraction is unreachable in both branches

`PLAN.md` §1.2 lists "Phoenix channels/PubSub **(works today)** — `extract_channels.c:875-970`". It does not work — it has never emitted a single Channel node:

- the emit path (`elixir_process_call`) reads the same nonexistent `arguments` field, so `elixir_emit_second_arg()` returns early every time
- the listener path is dispatched by `strcmp(kind, "def") == 0`, and there is no `def` node type in the grammar — Elixir definitions are `call` nodes. `elixir_process_function_def()` is dead code, and it also reads `name` and `parameters` fields the grammar does not define.

Fixed and verified against `Phoenix.PubSub.broadcast/subscribe`, bare `push`/`broadcast`, and two `handle_in` clauses including a guarded one.

### 3. Phoenix routers mint zero Route nodes

`PLAN.md` §1.2 lists "Phoenix/Oban/Broadway service patterns **(work today)**". Oban and Broadway do. Phoenix routes do not.

Every entry in `route_reg_suffixes` is prefixed `.` or `::` (`.get`, `::get`, `.MapGet`) because it is matched against the tail of a qualified callee. Phoenix registers with bare macros:

```elixir
[...]

PR #1730, opened 2026-08-19
fix(elixir): Phoenix channel extraction is unreachable in both branches

henry-hz:
Refs #1729 (defect 2 of 4).

## The bug

Neither half of `extract_channels_elixir()` can fire, so an Elixir project produces **zero** `Channel` nodes regardless of how much `Phoenix.PubSub` or `Phoenix.Channel` it uses.

`docs/elixir-lsp/PLAN.md` §1.2 currently records this path as "Phoenix channels/PubSub (works today)", which is what prompted me to write it up rather than just patch it.

### Emit side

`elixir_process_call()` reads:

```c
TSNode args = ts_node_child_by_field_name(call, TS_FIELD("arguments"));
```

`tree-sitter-elixir` defines no `arguments` field — its whole field set is `key, left, operand, operator, quoted_start, quoted_end, right, target, value`. `args` is always null, so `elixir_emit_second_arg()` returns early every time.

### Listener side

```c
} else if (strcmp(kind, "def") == 0) {
    elixir_process_function_def(ctx, node);
}
```

There is no `def` node type in the grammar — an Elixir definition is a `call` whose target is the macro name. That branch is unreachable, and `elixir_process_function_def()` compounds it by reading `name` and `parameters` fields that also do not exist.

## The fix

- a positional arguments fallback, mirroring `elixir_call_args()` which `extract_defs.c` has always used for the same reason
- dispatch on `call` + a def-macro target instead of the phantom `def` node type
- `elixir_process_function_def()` now takes the def call, unwraps a `when` guard, and reads the head's name and parameters positionally — so guarded clauses like `def handle_in(e, p, s) when is_map(s)` are caught too
- the callee lookup accepts `target` or `child(0)`, matching what `extract_defs.c` does

## Test

`elixir_channels_pubsub_and_handle_in` covers `Phoenix.PubSub.broadcast`, `Phoenix.PubSub.subscribe`, a bare `push` emit, and two `handle_in` listener clauses, one guarded.

## Verification

- CALLS-breadth contract across 53 languages unaffected; every change is inside the Elixir path
- `clang-format --dry-run --Werror` clean on `extract_channels.c`. `tests/` left unformatted, per its exclusion from `LINT_SRCS`

Independent of #1721 — these can land in either order.

Submitted as a bug fix under the CONTRIBUTING exception for focused bug fixes.


DeusData:
**Confirmed dead, and the evidence is even cleaner than your report.** You wrote that tree-sitter-elixir never produces a `def` node, so the `else if (strcmp(kind, "def") == 0)` dispatch at `extract_channels.c:969` is unreachable. Our own language spec says the same thing outright — `lang_specs.c:499` declares `elixir_func_types[] = {"call", "anonymous_function", NULL}`. So the file that defines what an Elixir function definition *is* has always said `call`, while the channel walker has been checking for `def`. `elixir_process_function_def` has never executed once, and every Phoenix `handle_in` listener has been invisible to the graph for as long as the code has existed.

That is a good find, and it is the kind that hides indefinitely because the dead branch looks like working code in review.

**The test is binding on all five assertions.** The emit side fails on `main` because `args` comes back null; the listener side fails because it is unreachable. And `"ping"` specifically exercises the `when`-guard unwrap — `def handle_in("ping", _p, socket) when is_map(socket)` — which is exactly the case that would have been quietly skipped by a naive first-child read. Covering the guarded clause rather than only the plain one is the difference between a fix and a fix that holds.

Scope is clean: everything is inside `extract_channels_elixir()` or new `elixir_*` statics, and the only caller chain is `extract_channels.c:1080` → `extract_channels_elixir` → `elixir_process_function_def`. Nothing else can reach it, so no other language is affected.

**One thing I want you to confirm rather than assume.** This is not purely "make the dead branch live" — you restructured the dispatch so a `call` whose target is `def`/`defp`/`defmacro`/`defmacrop` routes to `elixir_process_function_def` **instead of** `elixir_process_call`, as an else-branch rather than an addition. So a definition-macro call can no longer also be examined as a PubSub emit.

I worked through it and I believe it is safe: the loop still pushes every child onto the stack, so broadcasts *inside* a `def` body are still visited, and a `def` node's own target could never match a broadcast pattern anyway. But it is a behaviour change beyond the stated fix, it is not called out in the description, and you know this code better than I do — so please confirm that reading, and it would be worth a line in the commit message either way.

**Your CI here is not a real failure.** The run shows `test` skipped plus two cancelled jobs plus a red aggregate, which is the signature of an incomplete run rather than a test result. I have re-run it; we will see a real signal shortly.

**One heads-up spanning your three PRs.** The production files are genuinely disjoint, so nothing here can break #1721 or #1731. But all three insert into `tests/test_extraction.c` at the same two anchors — right after `TEST(elixir_function)` closes, and right after `RUN_TEST(elixir_function);` in the suite. GitHub reports all three MERGEABLE because it compares each against `main` alone; I expect the first merge to flip the other two to CONFLICTING at both points. Trivial to rebase, just not a surprise worth having.

Also note this adds a third near-identical positional-args helper (`elixir_channel_call_args`, alongside `elixir_call_args` in `extract_defs.c` and #1721's `elixir_call_arguments_fallback`). I have asked on #1721 whether you would rather hoist one shared helper now or consolidate afterwards — same question, no strong preference from me.

DeusData:
Approved. Both halves of your diagnosis check out, and I verified them against the grammar rather than taking them.

**Emit side.** Elixir's `ts_field_names` table holds exactly the nine fields you listed and no `arguments`, so `ts_node_child_by_field_name(call, "arguments")` really did return null for every call. (Confirmed while reviewing #1721, which hits the same wall from the other direction — worth noting the string `"arguments"` *does* appear in `parser.c`, but as a node **type** name, `sym__call_arguments_with_parentheses`. Anyone checking with a quick grep gets the opposite answer.)

**Listener side.** `= "def"` appears **zero** times in the Elixir parser's symbol names. There is no `def` node type, so `strcmp(kind, "def") == 0` could never be true and `elixir_process_function_def()` was unreachable — which is presumably why nobody noticed it was also reading `name` and `parameters` fields that do not exist either. Two dead layers stacked.

**The fix follows the house pattern rather than inventing one.** The positional-arguments fallback mirrors `elixir_call_args()` in `extract_defs.c`, and accepting `target` or `child(0)` for the callee matches what that file already does. Two instances of one pattern beats two special cases.

**Unwrapping `when` guards is the detail I would have missed.** `def handle_in(e, p, s) when is_map(s)` is idiomatic Phoenix, and a fix that dispatched on the def macro but then read the head positionally without unwrapping the guard would have silently skipped every guarded clause — a partial fix that looks complete. Covering one guarded and one unguarded `handle_in` in the same test is the right shape.

## One thing I could not verify

> `docs/elixir-lsp/PLAN.md` §1.2 currently records this path as "Phoenix channels/PubSub (works today)"

That file is not in this repository — there is no `docs/elixir-lsp/` directory, and the phrase does not appear anywhere under `docs/`. Where does it live? If a plan document somewhere really does describe this path as working, that is worth correcting alongside the code, and it is the same class of finding as the code itself: something asserting a behaviour that provably cannot happen. I just cannot fix a file I cannot find.

**Status:** cleared and approved. `main` moved three times yesterday (broken by a duplicate-symbol merge, repaired by #1993, then #1703 landed), so I will update this branch and re-run before merging — nothing needed from you.

Your note that this is independent of #1721 is right, and matches what I found: the two touch different files and neither fallback depends on the other.

DeusData:
Your first Elixir fix just landed — #1721 merged as `ffb5ebe9e8ad`, thank you! That merge is also what turned this PR red-conflicting a minute later, so here is the exact resolution so you do not have to dig:

The only conflict is in `tests/test_extraction.c`, and it is a pure "both sides appended a test at the same spot" clash:

- **`origin/main`** appended `TEST(elixir_call_string_argument)` (from #1721) right where this PR appends `has_channel()` + `TEST(elixir_channels_pubsub_and_handle_in)`.
- Resolution: **keep both blocks** — main's test first, then yours (or the other way round; order does not matter). Same for the `RUN_TEST(...)` lines at the bottom: keep `RUN_TEST(elixir_call_string_argument);` and add `RUN_TEST(elixir_channels_pubsub_and_handle_in);` next to it.

Nothing in `internal/cbm/extract_calls.c` conflicts — that part auto-merges.

Once you have pushed the rebase (or a merge of `main`, either is fine — we merge with a merge commit, never squash) the matrix reruns on the new head and I will merge on green.

Heads-up for #1731: it conflicts with `main` in exactly the same way (same append point in `test_extraction.c`, keep-both). Since #1730 and #1731 both append there, whichever lands second will need one more rebase after the first merges — if you would rather avoid the double round-trip, rebasing #1731 onto this branch instead of `main` and mentioning it here also works.

PR #1731, opened 2026-08-19
fix(elixir): functions carry no complexity, fingerprint or line count

henry-hz:
Refs #1729 (defect 4 of 4).

## The bug

Every Elixir `Function` ships with `complexity = 0`, `cognitive = 0`, `lines = 0` and no MinHash fingerprint. Elixir is therefore invisible to complexity and bottleneck queries, and never participates in `SIMILAR_TO` or duplicate detection — while the `query_graph` tool description states that "every Function and Method node carries queryable complexity properties".

Measured before this change on a 829-file Phoenix application: **2 of 4,504** functions had `complexity > 0`, and **7** had a fingerprint. Both figures came entirely from the repo's TypeScript and Python sources.

## Why the generic path cannot be reused

`cbm_compute_complexity()` matches on node **type**. Elixir has no branching node types — `if`, `unless`, `case`, `cond`, `with`, `for`, `try` and `receive` are all ordinary `call` nodes, distinguished only by the identifier in callee position, and every case/cond/fn/rescue arm is a `stab_clause`.

That is also why `elixir_branch_types = {"call"}` in `lang_specs.c` has never been usable: it would score every function call as a decision.

## The fix

`elixir_compute_complexity()` — one walk, same Campbell cognitive weighting as the generic version, counting:

- the branching construct itself: `if`, `unless`, `with`, `for`, `case`, `cond`, `try`, `receive`
- every `stab_clause` — one per case/cond/fn/rescue/catch/else arm
- the short-circuit operators `&&`, `||`, `and`, `or`
- `when` guards

Container **and** arm are both counted, which is the convention the generic tables already use — Go counts `expression_switch_statement` plus each `expression_case`; Python counts `try_statement` plus each `except_clause`. Elixir was otherwise scoring below every other language by construction.

A `for` comprehension is the only real loop node. `Enum.map/reduce` iterate too, but they are plain function calls and counting them would make `loop_depth` meaningless.

A straight-line single-clause function still scores 0, exactly like every other language's branch-free function — measuring a language must not make every function nonzero.

The fingerprint is taken from the `do_block` rather than the whole def call, so the head does not make every clause of a name look alike.

## Verification

- hand-checked against source: `Bank.Clients.apply_client_filters` scores 34 — exactly its 14 `stab_clause` arms + 6 `when` guards + 6 `and` + 1 `if` + 5 `or` + 1 `and` + the inner `case` container
- distribution on `elixir-lang/elixir` (6,346 `.ex` functions) looks like a healthy tail rather than noise: 4,753 above 0, 448 above 10, 108 above 20, 31 above 40
- `clang-format --dry-run --Werror` clean on `extract_defs.c`. `tests/` left unformatted, per its exclusion from `LINT_SRCS`

Independent of #1721 and #1730.

One note for @holsee, since this touches the same function as D3: this is written against upstream's current `extract_elixir_func_def()`, which emits one node per clause. If name/arity identity lands and clauses merge into a single Function, the per-clause metrics should be summed and the clause count itself charged as dispatch complexity — a multi-clause `defp` with no body branches is still a decision tree. Happy to rework this on top of that instead if the ordering suits you better.

Submitted as a bug fix under the CONTRIBUTING exception for focused bug fixes.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -4985,6 +4985,134 @@ static TSNode elixir_call_args(TSNode node) {
 }
 
 // Handle Elixir def/defp/defmacro — extract function definition.
+/* True when node `n` spans exactly the text `lit`. Avoids an arena allocation
+ * per node in the complexity walk, which visits every node of every def. */
+static bool elixir_node_text_is(TSNode n, const char *source, const char *lit) {
+    if (ts_node_is_null(n)) {
+        return false;
+    }
+    uint32_t s = ts_node_start_byte(n);
+    uint32_t e = ts_node_end_byte(n);
+    size_t len = strlen(lit);
+    return (size_t)(e - s) == len && strncmp(source + s, lit, len) == 0;
+}
+
+/* Elixir has no branching NODE TYPES. `if`, `unless`, `case`, `cond`, `with`,
+ * `for`, `try` and `receive` are ordinary `call` nodes told apart only by the
+ * identifier in callee position, and every case/cond/fn/rescue arm is a
+ * `stab_clause`. lang_specs.c cannot express that — the Elixir branch table is
+ * `{"call"}`, which would score every function call as a decision — and
+ * cbm_compute_complexity matches on node type, so the generic path is unusable
+ * and every Elixir function shipped complexity 0.
+ *
+ * Counted here: the branching construct itself, every `stab_clause`, the
+ * short-circuit operators, and `when` guards. Container AND arm are both
+ * counted, which is the convention the generic tables already use — Go counts
+ * `expression_switch_statement` plus each `expression_case`, Python counts
+ * `try_statement` plus each `except_clause`. Cognitive weighting is Campbell's,
+ * identical to the generic walk (+1 plus the current nesting depth). */
+static bool elixir_is_branch_call(TSNode call, const char *source) {
+    if (ts_node_child_count(call) == 0) {
+        return false;
+    }
+    TSNode callee = ts_node_child(call, 0);
+    if (ts_node_is_null(callee) || strcmp(ts_node_type(callee), "identifier") != 0) {
+        return false;
+    }
+    return elixir_node_text_is(callee, source, "if") ||
+           elixir_node_text_is(callee, source, "unless") ||
+           elixir_node_text_is(callee, source, "with") ||
+           elixir_node_text_is(callee, source, "for") ||
+           elixir_node_text_is(callee, source, "case") ||
+           elixir_node_text_is(callee, source, "cond") ||
+           elixir_node_text_is(callee, source, "try") ||
+           elixir_node_text_is(callee, source, "receive");
+}
+
+/* A comprehension (`for x <- xs, do: ...`) is Elixir's only real loop node.
+ * Enum.map/reduce iterate too, but they are plain function calls; counting them
+ * would make loop_depth meaningless, so they are left out. */
+static bool elixir_is_loop_call(TSNode call, const char *source) {
+    if (ts_node_child_count(call) == 0) {
+        return false;
+    }
+    TSNode callee = ts_node_child(call, 0);
+    return !ts_node_is_null(callee) && strcmp(ts_node_type(callee), "identifier") == 0 &&
+           elixir_node_text_is(callee, source, "for");
+}
+
+static bool elixir_is_branch_operator(TSNode node, const char *source) {
+    TSNode op = ts_node_child_by_field_name(node, TS_FIELD("operator"));
+    return elixir_node_text_is(op, source, "&&") || elixir_node_text_is(op, source, "||") ||
+           elixir_node_text_is(op, source, "and") || elixir_node_text_is(op, source, "or") ||
+           elixir_node_text_is(op, source, "when");
+}
+
+static void elixir_compute_complexity(TSNode node, const char *source, cbm_complexity_t *out) {
+    enum { ELIXIR_CX_STACK_CAP = 4096 };
+    struct {
+        TSNode node;
+        int bdepth;
+        int ldepth;
+        int adepth;
+    } stack[ELIXIR_CX_STACK_CAP];
+    memset(out, 0, sizeof(*out));
+    int top = 0;
+    stack[top].node = node;
+    stack[top].bdepth = 0;
+    stack[top].ldepth = 0;
+    stack[top].adepth = 0;
+    top++;
+    while (top > 0) {
+        int cur = --top;
+        TSNode n = stack[cur].node;
+        int bdepth = stack[cur].bdepth;
+        int ldepth = stack[cur].ldepth;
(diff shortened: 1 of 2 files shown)


DeusData:
**Your premise checks out, and it explains a whole class of missing data.** `lang_specs.c:503` declares `elixir_branch_types[] = {"call", NULL}` — so the generic complexity walker, which counts branch nodes by type, sees every Elixir expression as a branch and is useless for this language. That is why every Elixir function has shipped `complexity = 0`, and why none of them participate in similarity or duplicate detection. Not a small gap.

**The test is binding, and the control is the part I want to credit.** `ASSERT_EQ(5, cls->complexity)` fails on `main` where everything is 0. But `ASSERT_EQ(0, plain->complexity)` is the assertion that makes the first one mean something: it pins that measuring a language must not make *every* function non-zero. A fix that just started emitting numbers would pass the first assertion and fail the second. That is the right shape.

**Answering your direction question directly: land this as-is, do not rebuild it on the clause-merging work.** I checked — there is no open PR touching `extract_elixir_func_def`, and nothing from that workstream is in flight. Rebuilding on top of unmerged work would mean blocking a self-contained fix on something that does not exist yet. Whoever lands second does the rebase; that is the cheaper order. Thank you for asking rather than assuming.

**Three things before merge, and the first one is the real blocker.**

1. **The fingerprint is not tested at all.** The test is named `elixir_complexity_and_fingerprint`, the PR title says "no complexity, fingerprint or line count", and the body claims the fingerprint fix — but the assertions are `complexity`, `cognitive`, `lines` and the control. Nothing touches the fingerprint. So `compute_fingerprint(ctx, &def, ts_node_is_null(fp_body) ? node : fp_body)` could be deleted and the suite would stay green.

   That matters here more than usual: this project has been through an audit specifically for tests that pass without proving the fix exists, and "green means fixed" is the standard the suite is held to. Please assert the fingerprint — non-zero on a real function, and ideally equal for two structurally identical functions with different names, since duplicate detection is the thing it exists for.

2. **`elixir_compute_complexity` puts 192 KB on the stack.** `stack[ELIXIR_CX_STACK_CAP]` is 4096 entries of `{TSNode, int, int, int}`; `TSNode` is 32 bytes, so each entry is 48 and the array is exactly 196,608 bytes in one frame. I measured it rather than estimating.

   This is not theoretical for us: three deep-recursion suites already overflow their thread stacks on our arm64 sanitizer container, and worker threads do not get the main thread's 8 MB. A heap allocation, or a much smaller initial array that grows, would take the risk away. The arena is already threaded through `ctx` if you want somewhere to put it.

3. **Hitting the cap silently under-reports.** The push guard is `i >= 0 && top < ELIXIR_CX_STACK_CAP`, so once the stack is full the remaining children are dropped and the function returns a smaller complexity with no signal. A metric that quietly reads low on exactly the largest functions — the ones you most want flagged — is worse than one that refuses to answer. Either grow rather than drop, or record that the value was truncated.

None of those are hard, and the shape of the change underneath them is right. Ping me when they are in.

**One heads-up spanning your three PRs.** Production files are disjoint, so nothing here interacts with #1721 or #1730. But all three insert into `tests/test_extraction.c` at the same two anchors — right after `TEST(elixir_function)` closes and right after `RUN_TEST(elixir_function);`. GitHub shows all three MERGEABLE because it compares each against `main` alone; expect the first merge to flip the other two to CONFLICTING at both points.

Your CI here also was not a real signal — `test` was skipped and `codeql-gate` shows our known gate-timeout pattern rather than a finding. I have re-run it.

DeusData:
Approved. This is the best-evidenced of the four, and every claim it makes about existing code checks out.

**`elixir_branch_types[] = {"call", NULL}` is real** — `lang_specs.c:523`. You are right that it has never been usable: scoring every `call` node as a decision would make complexity a function-call count. A configuration entry that exists, looks plausible and cannot ever be switched on is worse than an absent one, because it reads as coverage.

**The container-plus-arm convention is real too.** Go counts `expression_switch_statement` at `:187` *and* `expression_case` at `:190`. Following the existing convention rather than inventing a scoring rule for Elixir is what makes these numbers comparable across languages — and you are right that counting only one of the two would have left Elixir scoring below every other language by construction.

**And the tool description does make the promise you quote.** `mcp.c:478`: *"COMPLEXITY / BOTTLENECKS: every Function and Method node carries queryable complexity"*. That was false for Elixir, and this makes it true. Fixing the code so the documentation stops lying is the right direction — the same shape as the parse-audit work landing elsewhere in this queue.

## The judgement calls are the good part

**"A straight-line single-clause function still scores 0, exactly like every other language's branch-free function — measuring a language must not make every function nonzero."** That is the sentence that separates a real metric from a busy one. It would have been easy to charge something per clause and produce a distribution that looked alive while meaning nothing.

**Taking the fingerprint from the `do_block` rather than the whole def call** is the equivalent instinct for similarity: including the head would make every clause of a name resemble every other, which is precisely the false positive `SIMILAR_TO` must not produce.

**Excluding `Enum.map/reduce` from loop depth** for the stated reason — they are plain calls, and counting them would make `loop_depth` meaningless — is the same discipline a third time.

**And your verification is the shape I wish more perf and metric changes had.** A hand-checked function whose score decomposes exactly (34 = 14 arms + 6 guards + 6 `and` + 1 `if` + 5 `or` + 1 `and` + the inner `case`) proves the walk, and the distribution across 6,346 real functions — 4,753 above 0, 448 above 10, 31 above 40 — proves it produces a tail rather than noise. Either alone would be weaker than both together.

## On the @holsee coordination

Flagging the interaction with the name/arity identity work, and offering to rework on top of it, is exactly right. For the record so it is not lost: if clauses do merge into one Function, summing the per-clause metrics **and** charging the clause count itself as dispatch complexity is the correct treatment — a multi-clause `defp` with no body branches genuinely is a decision tree, and dropping to a single clause's score would understate it. No need to reorder on my account; whichever lands first, the other rebases.

**Status:** cleared and approved. Additive only, +183/-0. `main` moved three times yesterday, so I will update this branch and re-run before merging — nothing needed from you.

That is four defects found, explained and fixed against a real Phoenix application. Thank you.

DeusData:
Same situation as #1730, and thank you for #1721 (`ffb5ebe9e8ad`) — that merge is what turned this one conflicting.

The conflict is confined to `tests/test_extraction.c` and is a pure both-sides-append at the same spot:

- **`origin/main`** appended `TEST(elixir_call_string_argument)` where this PR appends `TEST(elixir_complexity_and_fingerprint)`.
- Resolution: **keep both** test blocks, and keep both `RUN_TEST(elixir_call_string_argument);` and `RUN_TEST(elixir_complexity_and_fingerprint);` at the bottom.

`internal/cbm/extract_defs.c` auto-merges cleanly (it picked up the Ruby superclass fix from #1709 today; no overlap with your hunk).

Because #1730 appends at the same spot too, the second of the pair to land will need one more small rebase after the first merges — see my note on #1730 for the option of stacking this one on that branch to save a round-trip. Merge on green either way.

PR #1741, opened 2026-08-19
fix(hooks): clear repository git environment for tests

### LINKED ISSUE #1740: pre-commit hook leaks GIT_INDEX_FILE into linked-worktree tests
## Description

With `core.hooksPath=scripts/hooks`, running `git commit` reaches the pre-commit hook's full test step but fails `git_context_linked_worktree`.

## Reproduction

1. Configure the repository hook: `git config core.hooksPath scripts/hooks`.
2. Stage any change and run `git commit`.
3. Observe the sequential test failure:

```
git_context_linked_worktree
FAIL tests/test_pipeline.c:6855: run_cmd(cmd) == 32768, expected 0 == 0
```

The failure can be reproduced without a commit:

```sh
GIT_INDEX_FILE=.git/index build/c/test-runner pipeline
```

This yields 249 passed and 1 failed. The same pipeline suite passes 250/250 when the repository-local Git environment is absent.

## Root cause

Git exports repository-local variables such as `GIT_INDEX_FILE=.git/index` to hooks. The test runner inherits that value. It works while a fixture uses a normal `.git` directory, but `git worktree add` checks out into a linked worktree where `.git` is a file, so the inherited `.git/index` path is invalid.

## Expected behavior

The hook should preserve Git's staged-index environment for diff-scoped linting, then clear repository-local Git variables only for the test subprocess.

## Validation

After isolating the test subprocess environment:

- poisoned-environment pipeline regression: 250 passed
- real pre-commit hook: 7,555 passed, 8 skipped
- canonical `scripts/test.sh`: 7,555 passed, 0 failed, 8 skipped

bmcnaboe:
## What does this PR do?

Clears Git's repository-local environment variables only for the pre-commit hook's test subprocess. This keeps staged-index context available to diff-scoped linting while preventing `GIT_INDEX_FILE=.git/index` from leaking into temporary linked-worktree fixtures.

## Root cause

Git exports repository-local variables to hooks. The inherited relative index path works for a normal `.git` directory but fails when `git worktree add` checks out into a linked worktree where `.git` is a file.

## Checklist

- [x] New behavior is covered by the poisoned-environment reproduction

## Validation

- `GIT_INDEX_FILE=.git/index ... pipeline` before: 249 passed, 1 failed
- isolated environment after: 250 passed
- real pre-commit hook: 7,555 passed, 8 skipped
- `scripts/test.sh`: 7,555 passed, 0 failed, 8 skipped

Fixes #1740


DeusData:
Thank you for the poisoned-environment reproduction and for limiting the change to the hook test subprocess. The contribution queue is quite full, so review may take a little time. We have the PR routed and will come back with grounded feedback as soon as possible.

DeusData:
Approved. Ten lines, and three separate things in them are the right choice rather than the easy one.

**The subshell makes the scoping structural.** Wrapping the unsets in `( ... )` means the hook's own diff-scoped linting keeps its staged-index context and only the test subprocess runs clean. Doing it with a subshell rather than unsetting and restoring is what makes that true by construction instead of by discipline.

**`git rev-parse --local-env-vars` is the authoritative list.** On this machine that is fifteen variables — `GIT_INDEX_FILE`, `GIT_DIR`, `GIT_WORK_TREE`, `GIT_COMMON_DIR`, `GIT_OBJECT_DIRECTORY` and the rest. A hand-written list of the two or three that bite today would have been shorter, passed review, and quietly rotted as Git adds variables. Asking Git what it exports is the version that stays correct.

**And the process substitution matters more than it looks.** The obvious spelling —

```sh
git rev-parse --local-env-vars | while IFS= read -r v; do unset "$v"; done
```

— puts the loop in a further subshell, so every `unset` evaporates and the hook silently keeps the poisoned environment. `< <(...)` keeps the loop in the shell whose environment you are trying to clean. That is a trap people fall into repeatedly, and the version here is right. The shebang is `#!/usr/bin/env bash`, so the bashism is safe.

The incidental `-j"$(...)"` quoting fix is correct too.

**The diagnosis is the good part though.** "The inherited relative index path works for a normal `.git` directory but fails when `git worktree add` checks out into a linked worktree where `.git` is a file" is precisely it — a relative `GIT_INDEX_FILE=.git/index` resolves to something meaningless once `.git` stops being a directory. That is also why this only shows up in worktree-based fixtures, which is exactly how work on this repository is expected to be done, so the blast radius was larger than the issue count suggested.

## Your CI is not your fault

All five reds are one cascade. `security / codeql-gate` ran from 16:08:24 to 16:54:00 — **45 minutes 36 seconds**, which is the gate's old wait budget expiring rather than a finding. `lint / lint` and `lint / lint-mem` both record `The operation was canceled` with zero warnings of their own, and `analyze` and `ci-ok` fall out of the same run.

That budget was raised from 45 to 150 minutes in #1960, which merged yesterday. A re-run should come back clean.

**I will update this branch and re-run before merging** — `main` moved three times yesterday (broken by a duplicate-symbol merge, repaired by #1993, then #1703 landed). Nothing needed from you.

Sorry it sat two weeks for a ten-line fix that unblocks a workflow the project itself recommends.

DeusData:
The CLANG64 2/2 red on this PR was the harness, not your change: the parallel scheduler's Windows descendant probe was timed by --kill-grace (1 s), so a cold PowerShell start read as a leaked process tree. That is fixed on main in 202c8e42b56a (#2055). This branch only needs an update from main to pick it up, which I'll do when the runner queue has room — nothing for you to do here.

DeusData:
Merged — thank you, and an apology first: this was approved on 1 September, I told you on the 5th that it only needed an update from `main`, and then it sat for two and a half more weeks. That was our queue, not anything you did.

What I checked before landing it tonight: no commit on `main` has touched `scripts/hooks/pre-commit` since your base, so the green run you had is still the truth for this file; the script parses; and the subshell does exactly what the comment says — inside it `GIT_DIR` and `GIT_INDEX_FILE` are gone, outside it the hook's own environment is intact, so the test suite's fixture repositories and linked worktrees stop inheriting the committing repository's index and object paths. Small, correct, and the kind of fix that saves someone an evening of "why does the suite fail only from the hook".

PR #1742, opened 2026-08-19
fix(tests): scale daemon runtime sanitizer timeout

### LINKED ISSUE #1738: daemon_runtime blocking-git probe times out before sanitized runner reaches main
### Version

codebase-memory-mcp dev (46ae198f)

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

Running the full ASan/UBSan test gate consistently fails `daemon_runtime_disconnect_cancels_blocked_non_index_child_and_preserves_other_session` at `ASSERT(marker_published)`.

The test copies the sanitized test runner to a private PATH entry named `git`, starts `detect_changes`, and waits for that copied process to publish its PID. It gives the copied process a fixed 5-second readiness budget.

On this machine the sanitized test runner is about 421 MB and the isolated copied probe takes about 7 seconds to reach `main` and publish its marker. The test therefore fails before it initiates or observes the cancellation behavior it exists to verify.

I expected this readiness wait to be a liveness backstop that accommodates normal sanitizer startup overhead, rather than a timing assertion on process startup. The repository already has a shared `CBM_SANITIZED` predicate and uses sanitizer-specific timing budgets elsewhere.

### Reproduction

1. Check out `DeusData/codebase-memory-mcp@46ae198f` on macOS 26.5.2 arm64.
2. Run `scripts/test.sh`.
3. The focused suite also reproduces it with `scripts/test.sh --suites daemon_runtime`.
4. Actual result:
   ```text
   44 passed, 1 failed
   blocking git marker missing: image=/tmp/cbm-runtime-non-index-.../bin/git completed=0 status=0 response=
   FAIL tests/test_daemon_runtime.c:3812: ASSERT(marker_published)
   ```
5. Expected: allow the copied runner to finish normal sanitizer startup, then verify process-tree cancellation.

Suggested fix: retain the native 5/15-second readiness/request budgets, but use 60/90-second liveness budgets under `CBM_SANITIZED`. A passing test still proceeds immediately when the marker appears.

With that change:
```text
daemon_runtime: 45 passed
full scripts/test.sh: 7555 passed, 0 failed, 8 skipped
=== All tests passed ===
```

### Logs

```text
daemon_runtime_disconnect_cancels_blocked_non_index_child_and_preserves_other_session
level=info msg=daemon.runtime_stopping reason=last_committed_client_disconnected
blocking git marker missing: image=/tmp/cbm-runtime-non-index-ypVYuS/bin/git completed=0 status=0 response=
non-index cancellation diagnostic: close_begun=1 clients_after_wait=2 connections_after_wait=2 child_gone=0 backstop=0 request_completed=0 request_status=0 second_after=0
FAIL tests/test_daemon_runtime.c:3812: ASSERT(marker_published)
```

```text

```

_No response_

### Confirmations

bmcnaboe:
## What does this PR do?

Uses sanitizer-aware timing budgets for the blocking-Git daemon cancellation regression while preserving the existing native timing budgets and cancellation/cleanup behavior.

## Root cause

The ASan/UBSan runner is about 421 MB and can take longer than five seconds to reach `main` after being copied. The test therefore failed at its readiness backstop before exercising cancellation.

## Checklist

- [x] New behavior is covered by the existing regression

## Validation

- `scripts/test.sh --suites daemon_runtime`: 45 passed
- real pre-commit hook: 7,555 passed, 8 skipped
- `scripts/test.sh`: 7,555 passed, 0 failed, 8 skipped
- production watchdog and security guards passed

Fixes #1738


DeusData:
Thank you for isolating the sanitizer-specific startup budget while preserving the native timing and cancellation contracts. The contribution queue is quite full, so review may take a little time. We have the PR routed and will come back with grounded feedback as soon as possible.

DeusData:
Approved. Timing-budget changes to tests get read carefully here, because the usual version of this change is tuning a racy assertion until it passes. Yours is not that, and the diff shows why in one glance.

**You scaled the precondition and left the measurement alone.** `CHILD_CANCEL_BOUND_MS` (3000) and `CHILD_CLEANUP_BOUND_MS` (5000) are untouched — those bound the behaviour actually under test, and they stay exactly as strict as they were. Only `CHILD_READY_BOUND_MS` and `REQUEST_TIMEOUT_MS` move, and only under `#if CBM_SANITIZED`. That split is the whole difference between a legitimate budget and a weakened test: if the cancellation bound had moved too, this would be a different conversation.

**And the comment says which is which**, right where the next person will meet it: *"This readiness wait is a liveness backstop, not the behavior under test."* That sentence is what stops someone in six months treating all four constants as interchangeable knobs.

**The root cause is a slow deterministic thing, not a racy one.** A 421 MB instrumented runner being copied and then taking more than five seconds to reach `main` is latency, not a race — the process gets there, it just takes longer under instrumentation and more still while the parallel gate is loaded. Waiting longer for something that will definitely happen is fine; waiting longer for something that might not is what we do not do.

Using the compile-time `CBM_SANITIZED` discriminator rather than sniffing at runtime is right too — the budget follows the build that actually needs it.

One thing worth naming rather than changing: a 60-second readiness backstop means a genuine hang now costs 60 seconds to surface instead of 5. That is the correct trade for a liveness backstop, but it is a real cost if this ever starts failing for a different reason, and worth remembering then.

## Your CI is the same story as #1741

All five reds are one cascade, and none of them is yours. `security / codeql-gate` ran 16:26:30 → 17:12:17 — **45 minutes 47 seconds** — which is the gate's old wait budget expiring rather than a finding. `lint / lint` and `lint / lint-mem` both record `The operation was canceled` with zero warnings of their own, and `analyze` and `ci-ok` fall out of the same run. That is the identical signature to your other PR.

The budget was raised from 45 to 150 minutes in #1960, merged yesterday. I will update this branch and re-run before merging — `main` also moved three times yesterday. Nothing needed from you.

Two small, well-argued infrastructure fixes sitting two weeks behind a gate timeout that had nothing to do with either of them. Sorry about that.

DeusData:
Merged — thank you, and sorry it waited this long after it went green.

Verified on the merge result with the current `main`: `daemon_runtime` 49 passed, including `daemon_runtime_disconnect_cancels_blocked_non_index_child_and_preserves_other_session`, and the memory-core linter is unchanged. One detail worth knowing: our default local build runs ASan and UBSan, so the sanitized branch of your bounds is the one that compiled and passed here, not the native fallback — the change was genuinely exercised, not just carried along. `main` has not touched `tests/test_daemon_runtime.c` since your base, so your green run still describes this file.

PR #1758, opened 2026-08-20
fix(installer): forward selected clients

### LINKED ISSUE #1753: install.sh silently drops --clients and configures every detected client
## Repro

```
curl -fsSL https://.../install.sh | bash -s -- --clients=claude,codex
```

on a machine with many MCP-capable clients installed/detected.

## Expected

Only the two named clients (`claude`, `codex`) get configured.

## Actual

Every detected client on the machine gets configured — 43 surfaces on the machine this was observed on. `--clients=…` is silently ignored by `install.sh`; the flag is accepted (no error) but has no effect on which clients get written to.

## Notes

The binary's own installer (`codebase-memory-mcp install --clients=claude,codex`, the `=` form) honors the flag correctly — this is specific to the `install.sh` shell wrapper's flag parsing/passthrough, not the underlying Go/C installer logic. Also worth tightening while in that code: unrecognized `install.sh`-level flags are currently accepted silently rather than erroring, and `--skip-config` should keep its current behavior.

Low severity/simple fix — happy to send a PR.

ketpatil77:
## What does this PR do?

Fixes #1753 by forwarding `install.sh --clients` selections to the downloaded binary's installer.

- Supports both `--clients=claude,codex` and `--clients claude,codex`.
- Preserves `--skip-config` behavior.
- Rejects unknown options and unexpected positional arguments instead of silently ignoring them.
- Extends the installer smoke test with detected Claude and Codex clients, selects only Claude, and verifies Codex remains untouched.

## Validation

- `bash -n install.sh`
- `bash -n scripts/smoke-test.sh`
- `install.sh --help` exposes `--clients`
- Unknown-option parser check rejects unsupported flags
- `git diff --check`

Full release-artifact smoke test was not run locally because it requires the repository's built release fixture.

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -19,6 +19,7 @@ main() {
 REPO="DeusData/codebase-memory-mcp"
 INSTALL_DIR="$HOME/.local/bin"
 SKIP_CONFIG=false
+CLIENTS=""
 CBM_DOWNLOAD_URL="${CBM_DOWNLOAD_URL:-https://github.com/${REPO}/releases/latest/download}"
 
 # Security: every remote hop must remain HTTPS. Plain HTTP is accepted only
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for the focused shell-installer fix and smoke coverage. I checked current main: install.sh accepts --skip-config but has no --clients parsing, and the downloaded binary is invoked with an INSTALL_ARGS array that never carries a client selector. That confirms why the release installer drops the user selection.

I have labeled this as a high-priority installer and editor-integration bug. Because this changes the remote bootstrap script and forwarded options, it is also routed through the installer-security review. The current CI surface is green, though the release-artifact smoke path still needs verification during review. Our queue is full, so that may take a little time. Thank you for rejecting unknown options instead of silently accepting them.

ketpatil77:
Closing as superseded by upstream commit 902f85b0 (`fix(install): forward clients selector`), which now implements the same installer forwarding on `main` with dedicated fixture-contract coverage. I rebased locally to verify our patch becomes empty against current `main`.

Issue #1692, opened 2026-08-17
C#: log calls inside Kafka-consumer classes misclassified as Route nodes; ASP.NET Core [Route]/[HttpGet] attribute routes not extracted (cross-repo-intelligence returns 0 edges)


shatulsky:
## Summary

On a large C# monorepo (~14.5k files, ASP.NET Core + Confluent.Kafka), `index_repository` (mode `full`, v0.10.5) produces two related but distinct Route-extraction defects that together make `cross-repo-intelligence` unusable: it returns 0 cross-repo edges even where real HTTP call chains between two indexed projects exist.

1. **Ordinary log calls inside Kafka-consumer classes get misclassified as `__route__kafka__` Route nodes.**
2. **Real ASP.NET Core `[Route]`/`[HttpGet]` controller-attribute routes are not extracted at all**, despite the underlying attribute/decorator parsing clearly working (the method has `DECORATES` edges for other things in the same graph).

## Repro 1 — false-positive Kafka routes from ordinary log calls

`Servers/ActivitiesService/Logic/MessageFilters/MessageFilterHelper.cs:306`

```csharp
catch (Exception ex)
{
    m_Logger.LogError("Can't get Account for account ID {0}: {1}", accountId, ex.Message);
}
```

This is a plain `ILogger.LogError` call — nothing Kafka- or route-related. After indexing, `search_graph(label="Route", name_pattern=".*(GET|POST|PUT|DELETE|api/).*")` returns dozens of nodes like:

```
__route__kafka__Can't get Account for account ID {0}: {1} Route  0 0
__route__kafka__Can't get Asset for asset ID {0}: {1} Route  0 0
__route__kafka__Failed to send EndOfDayQuotesSaved, exception: {0} Route  0 0
```

680 total `Route` nodes in this repo, the overwhelming majority of which are log-message strings misclassified this way (also affects `__route__hangfire__`).

### Root cause hypothesis (traced in source)

`cbm_service_pattern_match(resolved_qn)` (`internal/cbm/service_patterns.c:789`) correctly recognizes `Confluent.Kafka` as an async library (`service_patterns.c:191`), and only fires based on the **resolved qualified name of the callee**, not textual/file proximity — this part looks correct by design.

[...]

PR #1759, opened 2026-08-20
fix(extract): scan every attribute_list sibling, not just the first (#1692)

Jumaga2015:
## Summary
- Addresses #1692 (Repro 2 — the ASP.NET route-extraction half, not the Kafka
  misclassification half, which is a separate call-resolution defect left
  for its own atomic fix as requested in the issue thread): a C# method or
  class with 2+ stacked attributes only got a `DECORATES` edge for the
  first bracket group. `find_jvm_modifiers()` used
  `ts_node_child_by_field_name()`, which returns only the first matching
  child, so a route attribute stacked behind another one (e.g.
  `[ServiceFilter(...)]` then `[Route("api/[controller]")]`) was silently
  invisible to `references`/`symbols` and to route extraction.
- `find_jvm_modifiers()` now returns every matching wrapper via a new
  `cbm_find_children_by_kind()` helper; both callers (`extract_decorators`,
  `scan_route_annotations`) iterate all of them.
- C#-only: Java/Kotlin/Swift use the `modifiers` wrapper (never repeats);
  PHP already groups a stack under one `attribute_list`.

## Test plan
- [x] `repro_issue1692` (`Makefile.cbm` `TEST_REPRO_SRCS`) — indexes
      `[Foo][Bar("x")][Baz]` through the real production pipeline and
      asserts `DECORATES` == 3. Proven RED with the pre-fix lookup
      (decorates=1), GREEN with the fix.
- [x] `scripts/test.sh CC=clang CXX=clang++` — full suite, no regressions
- [x] `scripts/lint.sh CC=clang CXX=clang++` — clang-tidy (diff-scoped),
      clang-format, cppcheck all clean

Made with [Claude Code](https://claude.com/claude-code)


DeusData:
Thank you for keeping the ASP.NET route-extraction half separate from the Kafka classification defect. I checked current main: find_jvm_modifiers() obtains the matching wrapper with ts_node_child_by_field_name(), so only the first repeated C# attribute_list can reach decorator and route scanning. That grounds the stacked-attribute failure.

I have labeled this as a high-priority parsing bug and routed it for review. The small Makefile change is only the focused repro registration, not a broader build-system redesign. Our review queue is full, so detailed review may take a little time. Thank you for splitting the two #1692 defects and covering the production pipeline path.

DeusData:
Approved, and this is now the surviving fix for the stacked-attribute bug. #1895 addressed the same defect and has been closed in favour of this one — its author found the bug independently and their diagnosis was right; yours is the more complete repair.

**Fixing the helper rather than one caller is why.** `find_jvm_modifiers()` has two callers on `main`:

```
extract_defs.c:1714   scan_route_annotations   (route extraction)
extract_defs.c:1981   extract_decorators
```

The other PR iterated wrappers inside `extract_decorators` and left the helper returning a single node, so `:1714` still saw only the first `attribute_list`. Since route extraction runs through `scan_route_annotations`, that version would have corrected the `decorators` array while leaving #1692's actual reported symptom — `[ServiceFilter(...)]` then `[Route("api/[controller]")]`, with the route invisible — exactly where it was.

Your version fixes the source, so both callers are correct and the next one is correct without having to know it needed to iterate.

**Scoping to C# with a stated reason is right.** Java, Kotlin and Swift use the `modifiers` wrapper, which does not repeat; PHP groups a stack under a single `attribute_list` with several `attribute_group` children. I checked the PHP grammar on that point specifically — the node names bear you out, and I had previously told the other contributor the opposite, which was my error and not theirs.

**And the repro is the right shape.** `repro_issue1692` indexing `[Foo][Bar("x")][Baz]` through the real production pipeline and asserting `DECORATES == 3`, proven RED at 1 before the fix, tests the behaviour end to end rather than the helper in isolation. A unit test on `cbm_find_children_by_kind()` would have passed against a pipeline that still dropped the attributes.

Splitting the Kafka misclassification out as its own atomic fix, as the issue thread asked, was also the right call.

## Status

**Clearance** is `REVIEW(1)` on `Makefile.cbm` — registering the new repro source, the established shape here. It needs a maintainer marker rather than a change from you.

`main` moved three times yesterday (broken by a duplicate-symbol merge, repaired by #1993, then #1703 landed), so I will update this branch and re-run before merging. Nothing needed from you.

DeusData:
As promised on 09-01, I've updated this branch myself rather than asking you to: 683d4bc3 merges current main into it. The only conflict was internal/cbm/helpers.h, where both sides had added declarations after cbm_find_child_by_kind — both blocks are kept, your cbm_find_children_by_kind prototype first. Your two commits are untouched. Verified on the merged tree before pushing: the #1692 repro is red with main's extract_defs.c and green with yours, and extraction plus the Java/Kotlin/PHP/C# LSP suites and pipeline are green (1,472 cases). CI runs now; I'll merge on green. Nothing needed from you.

DeusData:
Merged as 5b8217d43f60 — thank you, and thank you for the patience through the rebases. Scanning every attribute_list sibling instead of stopping at the first one is the right fix for stacked C# attributes, and the repro_issue1692 case on the bug-repro board keeps it honest. This ships in v0.10.9.

PR #1768, opened 2026-08-20
test(hook-augment): pin worktree payload-cwd resolution

### LINKED ISSUE #1752: hook-augment returns 0 bytes when payload cwd is an indexed linked worktree
## Repro

```
git worktree add ../wt -b test
# index both the main checkout and ../wt separately
```

Pipe a `PreToolUse` payload into `hook-augment`, varying only the payload's `cwd` field (process cwd held constant, deadline raised to 8000ms to rule out #1335):

| process cwd | payload `cwd` | bytes |
|---|---|---|
| main | main | **663** |
| main | worktree | **0** |
| worktree | main | **663** |
| worktree | worktree | **0** |

The payload's `cwd` alone decides the outcome; the process's own cwd is irrelevant.

## Expected

A payload whose `cwd` is an indexed linked worktree should return that worktree's own project context — the worktree has its own full index (confirmed via `index_status`, `is_worktree=true`, resolving `canonical_root` back to the main checkout's `.git`), and `search_graph` against the worktree's project directly returns correct, non-stale results.

## Actual

Empty output (exit 0, nothing on stderr, `hook-augment-timeouts.log` unaffected) at both the default 2s deadline and an 8s deadline — this is not the deadline/fingerprint issue in #1335, it's a clean no-output path with its own project already indexed.

## Notes

`index_status` already resolves worktree paths correctly (`is_worktree`, `git_common_dir`, `canonical_root`) — the hook's own cwd → project resolution appears not to use that same logic, so an indexed linked worktree's payload cwd fails to resolve to its project. Distinct from #1335 (that's a timeout under the SHA-256 fingerprint cost); this is a resolution defect independent of timing.

Happy to share the full measurement writeup if useful.

bmcnaboe:
Regression tests for the hook augment payload-cwd -> project resolution across a linked git worktree, on a real worktree fixture:

- payload cwd inside an indexed linked worktree (and a subdir) resolves to that worktree own project, never a sibling checkout or nothing;
- an unindexed worktree gets a deliberate "no indexed graph project matched" notice, never a silent 0 bytes;
- main-checkout resolution is unchanged.

Current `main` already resolves worktrees correctly; this adds the coverage that was missing so it cannot regress. Opened as a draft.

Fixes #1752

### DIFF
--- a/tests/test_cli.c
+++ b/tests/test_cli.c
@@ -9342,6 +9342,175 @@ TEST(cli_hook_session_exhausts_project_registry_pages) {
     PASS();
 }
 
+#ifndef _WIN32
+/* A linked git worktree is indexed as its OWN project, keyed by its checkout
+ * path. The SessionStart hook must return that project for a payload cwd inside
+ * the worktree (or a subdir of it) — never the main checkout's — and the main
+ * checkout's own resolution must be unchanged. Pins the payload-cwd → project
+ * resolution so a worktree can never regress to resolving a sibling checkout or
+ * to nothing (the historical silent-0-bytes worktree bug). */
+TEST(cli_hook_worktree_cwd_resolves_own_indexed_project) {
+    char tmpdir[256];
+    snprintf(tmpdir, sizeof(tmpdir), "/tmp/cli-hook-worktree-XXXXXX");
+    if (!cbm_mkdtemp(tmpdir))
+        FAIL("cbm_mkdtemp failed");
+    char cache[512];
+    char maindir[512];
+    char wtdir[512];
+    char subdir[600];
+    snprintf(cache, sizeof(cache), "%s/cache", tmpdir);
+    snprintf(maindir, sizeof(maindir), "%s/checkout-main", tmpdir);
+    snprintf(wtdir, sizeof(wtdir), "%s/checkout-side", tmpdir);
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you for adding deterministic linked-worktree coverage while leaving current production behavior unchanged. The positive, unindexed, and main-checkout controls give this a clear maintenance boundary. The contribution queue is quite full, so review may take a little time, but we will come back with feedback as soon as possible.

DeusData:
Picking this back up, with thanks for the patience: the branch now conflicts with main (test_cli.c moved a lot underneath it in the last two weeks), so it cannot be evaluated as-is. Could you rebase onto current main and keep the sign-off? Once it applies cleanly I will take the host-side clearance for the four `system()` sites in the test (they are the same shape as the fourteen existing ones in the suite) and give it a proper review straight away.

bmcnaboe:
Rebased onto current main (b5185d10), sign-off kept. Same two tests, one fewer hunk: the forward declaration of `cbm_project_name_from_path` is gone since `pipeline.h` already provides it. `cli` suite passes locally (311). The earlier `test-windows-guards` failure was the guard's own fixture ("ASCII baseline did not index"), not these tests, which are compiled out on Windows.

PR #1769, opened 2026-08-20
fix(install): honor --clients and reject unrecognized flags

### LINKED ISSUE #1753: install.sh silently drops --clients and configures every detected client

bmcnaboe:
install.sh silently discarded any flag it did not recognize, so `--clients=claude,codex` was dropped and every detected client got configured.

Replace the two lenient parse loops with one strict parser: known flags only (--dir/--dir=/--clients=/--skip-config/--help); an unrecognized flag is an error (exit 2); and `--clients` is passed through to the binary installer, which honors the `=` form and validates the list. `--skip-config` behavior is unchanged. shellcheck clean. Opened as a draft.

Fixes #1753

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -79,26 +79,42 @@ download_file() {
     fi
 }
 
+# Strict argument parser: known flags only. An unrecognized flag is an error,
+# never silently ignored — otherwise a mistyped or unsupported option (a dropped
+# --clients being the motivating case) leaves the caller believing it took effect
+# while the install proceeds with defaults. --clients is passed through to the
+# binary installer, which owns client selection.
+CLIENTS=""
+expect_dir=false
 for arg in "$@"; do
+    if [ "$expect_dir" = true ]; then
+        INSTALL_DIR="$arg"
+        expect_dir=false
+        continue
+    fi
     case "$arg" in
         --dir=*)        INSTALL_DIR="${arg#--dir=}" ;;
+        --dir)          expect_dir=true ;;
+        --clients=*)    CLIENTS="${arg#--clients=}" ;;
         --skip-config)  SKIP_CONFIG=true ;;
         --help|-h)
-            echo "Usage: install.sh [--dir=<path>] [--skip-config]"
-            echo "  --dir PATH     Install directory (default: ~/.local/bin)"
-            echo "  --skip-config  Skip automatic agent configuration"
+            echo "Usage: install.sh [--dir=<path>] [--clients=<c1,c2,...>] [--skip-config]"
+            echo "  --dir PATH        Install directory (default: ~/.local/bin)"
+            echo "  --clients LIST    Comma-separated clients to configure (default: all detected)"
+            echo "  --skip-config     Skip automatic agent configuration"
             exit 0
+            ;;
+        *)
+            echo "error: unknown option '$arg'" >&2
+            echo "Run 'install.sh --help' for usage." >&2
+            exit 2
             ;;
     esac
 done
-# Handle --dir <path> (space-separated)
-prev=""
-for arg in "$@"; do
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you for keeping the installer fix focused on strict flag parsing and correct `--clients` forwarding while preserving `--skip-config`. The contribution queue is quite full, so review may take a little time. We have the PR routed and will return with grounded feedback as soon as possible.

DeusData:
Closing this because the fix has landed — but you were first, and you were not credited. I want to say that plainly before anything else.

You opened this on **20 August**. It sat unreviewed, and on **27 August** I committed `902f85b0` ("fix(install): forward clients selector") under my own name, fixing the same defect you reported in #1753. Seven days, no credit, no comment here. That is the second time today I have found this pattern in my own queue, and it is a failure of my review process, not a coincidence.

## What is on `main` now

The parser was rewritten as a `while [ "$#" -gt 0 ]` loop with proper `shift` consumption, `CLIENTS_SET`/`CLIENTS` are threaded through to `INSTALL_ARGS`, and unknown options exit 2. So your diagnosis — *"install.sh silently discarded any flag it did not recognize, so `--clients=claude,codex` was dropped"* — is exactly what got fixed, and your strictness principle is what shipped.

I looked hard for something in your version worth keeping on top, and in fairness to you I should report what I found, because the current code is ahead on four specific points:

1. **`--dir` value validation.** `main` refuses a value that starts with `-`:
   ```sh
   if [ "$#" -lt 2 ] || [[ "$2" == -* ]]; then
   ```
   Your `expect_dir` flag has no such guard, so `install.sh --dir --skip-config` would set `INSTALL_DIR="--skip-config"` and install into a directory of that name.

2. **`--clients` and `--skip-config` are independent on `main`** — two separate `if` blocks. Yours uses `elif`, so passing both silently drops `--clients`. That is the same bug class this PR set out to kill, which is worth knowing precisely because it is so easy to reintroduce.

3. **`CLIENTS_SET` distinguishes "not passed" from "passed empty"**, which `-n "$CLIENTS"` cannot.

4. **`-*` and bare `*` get separate messages** — "unknown option" vs "unexpected argument".

None of that is a criticism of a patch written in August against August's code. It is the reason I am closing rather than asking you to rebase: there is genuinely nothing left for this diff to add, and asking you to rework it to match would be wasting more of your time on top of the week it already sat.

## What I owe you

I cannot retroactively add a trailer to a commit already on `main` without rewriting public history, which I will not do. So the credit is here, in the record: **the `--clients` selector is honoured by `install.sh` today because @bmcnaboe reported it in #1753 and fixed it in #1769 on 20 August.** If you would like me to note it in the release notes for the version that carries it, say the word and I will.

Thank you for this, and I am sorry it went the way it did. Your #1767 is a genuinely good piece of work and I have reviewed it properly today.

PR #1770, opened 2026-08-20
fix(watcher): stop nested non-git dirs and monorepo siblings retriggering

DeusData:
Distilled from #1001 by @bethzyy, credited via `Co-authored-by`. Addresses the remaining parts of #713, #841 and #937.

`main` independently fixed that PR's two headline problems while it waited (the unbounded non-git file count, and per-poll churn via a dirty-state signature), and the watcher was rewritten onto a supervised argv-spawn harness in between — so this is a reimplementation against the new code rather than a rebase. Three of their findings were still live on `main`, and I verified each one present before touching it.

### 1. Nested non-git directories inherit an unrelated repo's dirt

`git rev-parse --git-dir` walks **up**, so a scratch folder that merely sits under some other repository answered yes. The watcher then inherited that ancestor's dirty state — permanently non-empty, and about entirely different files — and reindexed **on every poll, forever**. Reporters measured this in hundreds of GB of writes per day.

A directory is now git-managed only if it carries its own `.git` (directory or gitlink file) **or** the ancestor actually tracks something inside it. A real monorepo sub-package still qualifies; a scratch or ignored folder does not.

### 2. Monorepo siblings cross-trigger

`git status` reports the whole repository regardless of `-C`, so every package reindexed whenever any *other* package was edited. The status call is now scoped with a `-- .` pathspec.

### 3. Porcelain paths were stat'ed against the wrong root

Porcelain prints **repository-relative** paths, but the dirty signature stat'ed them against the project root. At a repository root the two coincide and the bug is invisible; for a subdirectory project every stat missed, silently degrading the signature to text-only and losing the size/mtime component that makes an edit to an already-dirty file detectable.

Paths now resolve through `rev-parse --show-cdup`. Deliberately **not** `--show-toplevel`: under MSYS/Cygwin git that returns a translated absolute path that will not join onto the native path we hold, whereas a relative hop composes on every platform.

### Tests

Both new tests are **verified binding**, each against its own bug:
- reverting the classification guard reddens `watcher_nested_non_git_dir_does_not_inherit_ancestor_dirt`
- removing the pathspec reddens `watcher_monorepo_subdir_ignores_sibling_changes`

and neither disturbs the other. The monorepo test also asserts the positive case — editing the watched package itself still triggers — so the scoping cannot silently swallow real changes.

### Verification

macOS ASan/UBSan: watcher, git_context, discover, gitignore, incremental, pipeline — **623 passed, 0 failed**. `lint-ci` and the clang-analyzer memory gate both clean.

### DIFF
--- a/src/watcher/watcher.c
+++ b/src/watcher/watcher.c
@@ -75,6 +75,10 @@ typedef struct {
     uint64_t last_dirty_sig;       /* committed dirty-state signature */
     uint64_t pending_dirty_sig;    /* observed at check time */
     char pending_head[CBM_SZ_128]; /* HEAD observed at check time */
+    /* Hop from root_path up to the repository root ("" when they are the same),
+     * from `rev-parse --show-cdup`. Porcelain paths are repository-relative, so
+     * the signature needs this to stat them. Resolved once at baseline. */
+    char repo_cdup[CBM_SZ_4K];
 } project_state_t;
 
 /* ── Watcher struct ─────────────────────────────────────────────── */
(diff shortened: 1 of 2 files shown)


DeusData:
Closing: **this change is already on `main`** as `1af49dfe`, byte-identical to this branch's `a0944974`.

The #1772 branch was stacked on this one, so merging #1772 carried the watcher fix in with it. GitHub cannot see that — the two commits have different SHAs despite identical content — so the PR still shows a live diff. Verified with a content-level comparison rather than by trusting the diff view.

Nothing is lost: `git log origin/main -- tests/test_watcher.c` shows the 130 lines of nested-non-git and monorepo-sibling coverage present on main.

PR #1778, opened 2026-08-21
feat(search): index markdown section bodies for BM25 full-text search (#518)

### LINKED ISSUE #518: Section nodes don't index body text — BM25 can't search markdown content

ShauryaaSharma:
Slice 1 of the #617 split, per the review guidance there: the FTS body infrastructure plus Markdown `Section` indexing. #617 stays open as the ledger; the #519 YAML/JSON Module description promotion follows as a second PR stacked on this one.

Rebuilt against current `main` rather than rebased — the original branch was ~1100 commits behind and `pipeline.c`, `mcp.c` and `store.c` had all been reworked around the touched regions, so the old diff served as the design spec.

## What & why

`search_graph` BM25 only matched node names and headings, so it was blind to markdown prose. `Section` nodes exposed only their heading text and were excluded from BM25 results entirely. This indexes the section body so content is searchable.

Closes #518.

## Changes

**store** — adds a `body` column to the `nodes_fts` FTS5 table. New `cbm_store_fts_rebuild()` drops and recreates the table (which is what upgrades legacy 4-column databases) and backfills `body` from each node's docstring, guarded by `json_valid()` so malformed-JSON rows degrade to name-only indexing instead of aborting the whole `INSERT...SELECT`.

`CBM_SQL_FTS_BODY_EXPR` is exported from `store.h` so every `nodes_fts` write site feeds `body` through one expression.

**pipeline** — the wholesale backfill delegates to `cbm_store_fts_rebuild()` (net −17 lines). The row-level delta-merge insert in `pipeline_delta.c` writes `body` through the same shared expression.

That third write site is the one worth reviewing closely. `nodes_fts` now has five columns, and an INSERT naming only the original four is still perfectly valid SQL that silently leaves `body` NULL — prose arriving through delta merge would have been unsearchable while a full reindex looked correct. It has its own test rather than relying on the full-index path.

**mcp** — stops excluding `Section` from BM25 results in both the search and count queries. `Section` falls in the unboosted `ELSE 0.0` bucket of the label-boost CASE, so code symbols keep their ranking advantage by construction rather than by exclusion. `Module` stays excluded here and is handled in the #519 slice.

**internal/cbm** — captures the markdown section body beneath each heading, stopping at the first nested subsection (each subsection gets its own `Section` node and its own body). Capped at the existing `MAX_COMMENT_LEN` (500) with a UTF-8-safe backoff, reusing the existing docstring property.

## Testing

Seven cases across three suites. CI runs them on ubuntu/macOS/Windows including MSan, TSan and LSan.

| test | guards |
|---|---|
| `markdown_section_body_captured` | body captured, sibling section excluded |
| `markdown_section_no_body` | documented sibling has prose while a bare heading does not |
| `markdown_section_body_capped` | cap holds and the UTF-8 backoff does not split a character |
| `fts_rebuild_indexes_body_content` | body tokens searchable, names still searchable |
| `fts_rebuild_upgrades_legacy_schema` | legacy 4-column table upgraded in place |
| `fts_rebuild_tolerates_malformed_properties` | `json_valid()` guard holds |
| `pipeline_delta_merge_indexes_body` | the incremental write site, driven through `cbm_delta_patch` |

The delta test is the one worth looking at. It drives the real merge — `cbm_delta_preseed`, then `cbm_delta_patch` — rather than reimplementing the INSERT in the test body. Verified that it discriminates: against a four-column revert of `pipeline_delta.c` the body assertions fail while the name assertion still passes, confirming the node merged and only its body was lost.

Writing it against the production path also surfaced an ordering constraint a hand-rolled version could not: `cbm_delta_preseed` lifts the gbuf id watermark above `MAX(id)`, so a node added to the gbuf before that call is silently never merged.

The extractor's grammar assumption was verified directly against `tree_sitter_markdown` rather than assumed: for nested headings the parser produces `section [0-81]` containing the heading, a paragraph, and a nested `section [48-81]`, and the helper correctly stops at byte 48.

## Performance

`scripts/benchmark-fts-body.c` (added here) isolates what the `body` column actually costs. Since the full-index path already did delete-all plus a full re-INSERT, `body` does not add a pass over the graph — it makes an existing rebuild index more text. So the harness measures the two things that genuinely change: per-row body tokenisation and the storage the column adds. It builds one synthetic corpus three ways:

- **A** — 4-column FTS (pre-#518 baseline)
- **B** — 5-column, body for every node (what this PR ships)
- **C** — 5-column, body only for `Section`/`Module` rows (the `WHERE` lever)

Worth knowing when reading the diff: the harness is 477 of the ~934 added lines, so a little over half of this PR is a measurement tool rather than feature code.

### FTS index size

| nodes | A | B | C | B − A |
|---:|---:|---:|---:|---:|
| 100k | 7.2 MB | 10.2 MB | 7.6 MB | **+3.0 MB** (+41%) |
| 200k | 15.0 MB | 19.7 MB | 15.8 MB | **+4.7 MB** (+31%) |
| 500k | 36.6 MB | 48.6 MB | 40.7 MB | **+12.0 MB** (+33%) |

Beyond the 100k point (where fixed page overhead distorts the ratio) the marginal cost settles at **~24 bytes per node**. Extrapolated to the ~8.5M-node reference corpus that is **~200 MB** — inside the +150–400 MB band estimated in review.

### One correction to the estimate

The `WHERE label IN ('Section','Module')` lever was expected to remove ~95% of the cost. Measured, it removes **~66%** of the added storage (at 500k: +12.0 MB → +4.1 MB).

The 95% holds for *row count* — variant C indexes 17.8k bodies against B's 157k, which is 11% — but storage does not follow row count. Markdown section bodies run substantially longer than function docstrings, and the fifth column carries fixed overhead whether or not a given row has prose. So narrowing buys less than it appears to, which seemed worth knowing before trading away function-docstring search for it.

### On timing

I am deliberately not quoting backfill timings. My machine shows 39–44% run-to-run spread, which is large relative to the delta being measured; the harness detects this and prints a warning rather than letting a noisy column pass as authoritative. The size figures above are deterministic and unaffected. Timings on a quieter machine would be meaningful — `scripts/benchmark-fts-body.sh` builds and runs it.

Two methodology notes, since both mistakes produce confident-looking nonsense and I hit them: variants are interleaved rather than grouped (grouped runs let drift land on whichever variant goes last, which reported C as *slower* than B despite C doing strictly less work), and the minimum is reported rather than the mean.

This does not replace `scripts/benchmark-index.sh`, which measures real end-to-end indexing on a real repository. I cannot run that one — it needs a built product binary and shells out to `python3`, and my toolchain is 32-bit MinGW.org GCC 6.3.0 with no `make`; a 32-bit address space also will not hold anything near the reference corpus. Per the split agreed on #617, that half is being run on the reference machine — this harness measures the mechanism portably, the reference run measures the reality.

## Behavioural changes

Two things change for existing callers, both deliberate, neither purely additive:

**`Section` is now returned by BM25 and counted in the total.** That is the intended effect of the direction call on #617 — it changes both the result set and the `total` for existing `search_graph` callers. `Section` sits in the unboosted `ELSE 0.0` bucket of the label-boost CASE, so code symbols keep their ranking advantage by construction rather than by exclusion.

**BM25 scores shift for every row, including rows with no body.** `bm25(nodes_fts)` is called with no column weights, so all columns sit at the default 1.0. FTS5 normalises by document length against the corpus average, and body tokens count toward row length — so `avgdl` moves as soon as `body` is populated, and every row's normalisation moves with it. This is not repairable by weighting: even `body` at 0.0 removes the contribution but not the length.

So the ranking question is not *whether* scores move but whether they move deliberately. I have left the weights alone pending a steer, since the ranking model is the maintainer's call. My reasoning is in the review thread: weights are worth setting to bound how far prose can outrank a symbol match, which argues for `body` below `name`/`qualified_name` rather than at parity.

## Notes

The schema change is backward compatible — additive column, legacy databases keep working read-only and are upgraded by the DROP+recreate at the next index run, which was already a full FTS rebuild. Note that this is a statement about the schema, not about ranking; see the behavioural changes above.

One operational caveat worth stating rather than leaving to be discovered: the DROP+recreate opens a brief window during the schema upgrade where a concurrent `bm25_search` finds no table and degrades to the regex path. It is one-time per database and degrades gracefully, but with daemon-era concurrent sessions it is observable.

No MCP tool changes, no new dependencies, no new `system()`/`popen()`/network calls.

### DIFF
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -177,6 +177,11 @@ typedef enum {
     CBM_LANG_COUNT
 } CBMLanguage;
 
+// Byte budget for any prose attached to a definition as its docstring: leading
+// doc comments, and Markdown section bodies (#518). Shared with tests so the cap
+// is asserted against one definition rather than a copied literal.
+#define CBM_MAX_COMMENT_LEN 500
+
 // --- Extraction result structs ---
 
 typedef struct {
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -3849,7 +3849,8 @@ static const char *qn_safe_segment(CBMArena *a, const char *name) {
     return out;
 }
 
-static void push_simple_class_def(CBMExtractCtx *ctx, TSNode node, char *name, const char *label) {
+static void push_simple_class_def(CBMExtractCtx *ctx, TSNode node, char *name, const char *label,
+                                  const char *docstring) {
     CBMArena *a = ctx->arena;
     CBMDefinition def;
     memset(&def, 0, sizeof(def));
@@ -3860,6 +3861,7 @@ static void push_simple_class_def(CBMExtractCtx *ctx, TSNode node, char *name, c
     def.start_line = ts_node_start_point(node).row + TS_LINE_OFFSET;
     def.end_line = ts_node_end_point(node).row + TS_LINE_OFFSET;
     def.is_exported = true;
+    def.docstring = docstring; // Markdown section body (#518); NULL for other configs
     cbm_defs_push(&ctx->result->defs, a, def);
 }
 
@@ -3958,6 +3960,58 @@ static char *extract_markdown_heading_name(CBMArena *a, TSNode node, const char 
     return trim_heading_name(name);
 }
 
+// Capture the prose body beneath a Markdown heading so BM25 can search the content
+// and not just the heading text (#518). In the tree-sitter-markdown grammar each
+// heading lives inside a `section` node that also holds the body blocks and any
+// nested subsections; the body is the source span between the heading and either
+// the first nested subsection or the end of the section. Nested subsections are
+// excluded because each gets its own Section node and its own body. Returns NULL
+// when there is no enclosing section or no body text. Trimmed, and capped at
+// MAX_COMMENT_LEN (the same budget docstrings use) without splitting a UTF-8
+// sequence.
+static char *extract_markdown_section_body(CBMArena *a, TSNode heading, const char *source) {
+    TSNode parent = ts_node_parent(heading);
+    if (ts_node_is_null(parent) || strcmp(ts_node_type(parent), "section") != 0) {
+        return NULL;
+    }
+    uint32_t body_start = ts_node_end_byte(heading);
+    uint32_t body_end = ts_node_end_byte(parent);
+    // Stop at the first nested subsection — it gets its own Section node + body.
+    uint32_t cc = ts_node_child_count(parent);
+    for (uint32_t i = 0; i < cc; i++) {
+        TSNode ch = ts_node_child(parent, i);
+        if (ts_node_start_byte(ch) >= body_start && strcmp(ts_node_type(ch), "section") == 0) {
+            body_end = ts_node_start_byte(ch);
+            break;
+        }
+    }
+    // Trim surrounding whitespace/newlines. UTF-8 lead and continuation bytes are all
+    // >= 0x80, so a byte-wise <= ' ' test never cuts a multi-byte character.
+    while (body_start < body_end && (unsigned char)source[body_start] <= ' ') {
+        body_start++;
+    }
+    while (body_end > body_start && (unsigned char)source[body_end - 1] <= ' ') {
+        body_end--;
+    }
+    if (body_end <= body_start) {
+        return NULL;
+    }
+    size_t len = (size_t)(body_end - body_start);
+    if (len > MAX_COMMENT_LEN) {
+        len = MAX_COMMENT_LEN;
+        // Back off so the cap never splits a UTF-8 multi-byte sequence: source[start+len]
+        // is the first excluded byte, and a continuation byte there means we landed
+        // mid-character.
+        while (len > 0 && ((unsigned char)source[body_start + len] & 0xC0) == 0x80) {
+            len--;
+        }
+        if (len == 0) {
+            return NULL;
+        }
+    }
+    return cbm_arena_strndup(a, source + body_start, len);
+}
+
 // INI: extract section name from section node.
 static char *find_ini_section_name(CBMArena *a, TSNode node, const char *source) {
     uint32_t nc = ts_node_child_count(node);
@@ -4021,6 +4075,7 @@ static bool extract_config_class_def(CBMExtractCtx *ctx, TSNode node, const char
     CBMArena *a = ctx->arena;
     char *name = NULL;
     const char *label = "Class";
+    const char *docstring = NULL;
 
     if (ctx->language == CBM_LANG_TOML &&
         (strcmp(kind, "table") == 0 || strcmp(kind, "table_array_element") == 0)) {
@@ -4036,6 +4091,7 @@ static bool extract_config_class_def(CBMExtractCtx *ctx, TSNode node, const char
         // label rather than degrade it to match a test. The markdown repro asserts
         // "Class"; that assertion is the inaccurate side and is flagged for review.
         label = "Section";
+        docstring = extract_markdown_section_body(a, node, ctx->source); // #518
     } else if (ctx->language == CBM_LANG_HCL && strcmp(kind, "block") == 0) {
         name = find_hcl_block_name(a, node, ctx->source);
     } else {
@@ -4043,7 +4099,7 @@ static bool extract_config_class_def(CBMExtractCtx *ctx, TSNode node, const char
     }
 
     if (name && name[0]) {
-        push_simple_class_def(ctx, node, name, label);
+        push_simple_class_def(ctx, node, name, label, docstring);
     }
     return true;
 }
@@ -4098,7 +4154,7 @@ static bool extract_sql_ddl_class_def(CBMExtractCtx *ctx, TSNode node, const cha
     if (!name || !name[0]) {
         return false;
     }
-    push_simple_class_def(ctx, node, name, label);
+    push_simple_class_def(ctx, node, name, label, NULL); // SQL tables/views carry no body
     // Must match push_simple_class_def's QN exactly (qn_safe_segment included)
     // or pass_usages cannot find the enclosing def for the lineage source.
     const char *qn =
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -16,7 +16,6 @@
 #include <ctype.h>
 
 // Buffer sizes for local arrays (base classes, params, return types).
-#define MAX_COMMENT_LEN 500
 #define MAX_BASES 16
 #define MAX_BASES_MINUS_1 15
 #define MAX_PARAMS CBM_SZ_32
@@ -1236,12 +1235,12 @@ static bool is_comment_node(const char *kind) {
             strcmp(kind, "line_comment") == 0 || strcmp(kind, "multiline_comment") == 0);
 }
 
-// Extract comment text, truncating to MAX_COMMENT_LEN.
+// Extract comment text, truncating to CBM_MAX_COMMENT_LEN.
 // #1017: snap the cut point back to a complete UTF-8 codepoint boundary.
 static char *extract_comment_text(CBMArena *a, TSNode node, const char *source) {
     char *text = cbm_node_text(a, node, source);
-    if (text && strlen(text) > MAX_COMMENT_LEN) {
-        size_t cut = MAX_COMMENT_LEN;
+    if (text && strlen(text) > CBM_MAX_COMMENT_LEN) {
+        size_t cut = CBM_MAX_COMMENT_LEN;
         while (cut > 0 && ((unsigned char)text[cut] & 0xC0) == 0x80)
             cut--;
         text[cut] = '\0';
@@ -3967,7 +3966,7 @@ static char *extract_markdown_heading_name(CBMArena *a, TSNode node, const char 
 // the first nested subsection or the end of the section. Nested subsections are
 // excluded because each gets its own Section node and its own body. Returns NULL
 // when there is no enclosing section or no body text. Trimmed, and capped at
-// MAX_COMMENT_LEN (the same budget docstrings use) without splitting a UTF-8
+// CBM_MAX_COMMENT_LEN (the same budget docstrings use) without splitting a UTF-8
 // sequence.
 static char *extract_markdown_section_body(CBMArena *a, TSNode heading, const char *source) {
     TSNode parent = ts_node_parent(heading);
@@ -3997,8 +3996,8 @@ static char *extract_markdown_section_body(CBMArena *a, TSNode heading, const ch
         return NULL;
     }
     size_t len = (size_t)(body_end - body_start);
-    if (len > MAX_COMMENT_LEN) {
-        len = MAX_COMMENT_LEN;
+    if (len > CBM_MAX_COMMENT_LEN) {
+        len = CBM_MAX_COMMENT_LEN;
         // Back off so the cap never splits a UTF-8 multi-byte sequence: source[start+len]
         // is the first excluded byte, and a continuation byte there means we landed
         // mid-character.
--- a/scripts/benchmark-fts-body.c
+++ b/scripts/benchmark-fts-body.c
@@ -1,0 +1,446 @@
+/*
+ * benchmark-fts-body.c — isolate the cost of the nodes_fts `body` column (#518).
+ *
+ * WHAT THIS MEASURES, AND WHY IT IS THE RIGHT QUESTION
+ *
+ * The full-index path already performed a wholesale FTS rebuild before #518:
+ * delete-all followed by a full re-INSERT over every node (pipeline.c,
+ * generation_rebuild_fts).  Adding `body` therefore does NOT introduce a new
+ * pass over the graph — it makes an existing rebuild index more text.  So the
+ * honest cost question is narrower than "how much slower is indexing":
+ *
+ *   1. per-row body tokenisation time, and
+ *   2. the storage the extra column adds to the FTS index.
+ *
+ * This harness measures exactly those two, at a range of node counts, by
+ * building the same synthetic corpus three ways:
+ *
+ *   A  4-column FTS (pre-#518 baseline: name, qualified_name, label, file_path)
+ *   B  5-column FTS, body backfilled for EVERY node (what #518 ships)
+ *   C  5-column FTS, body backfilled only for Section/Module rows
+ *      (the "one-line WHERE" lever — loses function-docstring search)
+ *
+ * B minus A is the cost of the feature.  C minus A is the cost if the backfill
+ * is narrowed.  B minus C is what the narrowing would save.
+ *
+ * WHAT IT DOES NOT MEASURE
+ *
+ * Real-corpus end-to-end wall-clock.  That needs a built product binary and a
+ * real repository — use scripts/benchmark-index.sh for it.  This harness
+ * deliberately isolates the FTS write so the numbers are not buried under
+ * parsing, tree-sitter, and I/O.
+ *
+ * FIDELITY NOTES (read before trusting a number)
+ *
+ * - The real backfill wraps `name` in cbm_camel_split(); that function lives in
+ *   the product, not here.  All three variants use the plain name identically,
+ *   so it cancels out of every delta.  Absolute figures for the name column are
+ *   therefore slightly low; the A/B/C deltas are unaffected.
+ * - The body expression is the real one, json_valid() guard included, so
+ *   malformed-JSON rows exercise the same path they do in production.
+ * - Index size is measured as page_count * page_size after the backfill, minus
+ *   the same figure for a database holding only the `nodes` table.  That
+ *   isolates the FTS shadow tables from the base data.
+ *
+ * BUILD
+ *   see scripts/benchmark-fts-body.sh, or:
+ *   cc -O2 -o benchmark-fts-body scripts/benchmark-fts-body.c \
+ *      vendored/sqlite3/sqlite3.c -Ivendored/sqlite3 \
+ *      -DSQLITE_ENABLE_FTS5 -lpthread -lm
+ *
+ * USAGE
+ *   ./benchmark-fts-body [rowcount ...]      (default: 100000 500000 2000000)
+ */
+
+#include "sqlite3.h"
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef _WIN32
+#include <windows.h>
+/* QueryPerformanceCounter rather than GetTickCount64: higher resolution, and it
+ * does not require a Vista-era SDK (older MinGW headers lack the latter). */
+static double now_ms(void) {
+    LARGE_INTEGER f, t;
+    QueryPerformanceFrequency(&f);
+    QueryPerformanceCounter(&t);
+    return (double)t.QuadPart * 1000.0 / (double)f.QuadPart;
+}
+#else
+#include <time.h>
+static double now_ms(void) {
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return (double)ts.tv_sec * 1000.0 + (double)ts.tv_nsec / 1e6;
+}
+#endif
+
+/* ── Corpus shape ─────────────────────────────────────────────────────────
+ * Tuned to the maintainer's estimate that Section+Module are ~5% of nodes, so
+ * the narrow lever removes ~95% of the body backfill.  Adjust and re-run if
+ * your corpus differs — every figure below is sensitive to these. */
+enum {
+    PCT_MODULE = 3,  /* one per indexed file */
+    PCT_SECTION = 3, /* markdown headings */
+    /* remainder: Function/Method/Class/... — the code symbols */
+
+    PCT_CODE_HAS_DOC = 30,    /* share of code symbols carrying a docstring */
+    PCT_SECTION_HAS_BODY = 80,/* share of Sections carrying prose */
+    PCT_MODULE_HAS_DESC = 40, /* share of Modules carrying a description */
+    PCT_MALFORMED = 1,        /* pre-fix rows whose properties JSON is invalid */
+
+    DOC_LEN_CODE = 180,   /* average docstring length, bytes */
+    DOC_LEN_SECTION = 300,/* markdown section bodies run longer */
+    DOC_LEN_MODULE = 120,
+
+    MAX_BODY = 500 /* MAX_COMMENT_LEN — the extractor's cap */
+};
+
+/* The real body expression from store.h (CBM_SQL_FTS_BODY_EXPR). */
+#define BODY_EXPR                                                                                  \
+    " CASE WHEN json_valid(properties)"                                                            \
+    " THEN coalesce(json_extract(properties,'$.docstring'),'') ELSE '' END "
+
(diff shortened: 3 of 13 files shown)


DeusData:
Read in full. The structure is right and slice 1 is complete with no slice-2 leakage. **But the delta test has exactly the hole it was written to close**, and there is a second behavioural change the description does not mention. Both are fixable; neither is a rethink.

## The delta test does not guard the delta call site

This is the important one, and I checked it rather than reading the name and moving on.

`fts_delta_insert_populates_body` never calls `cbm_delta_patch()` or any production entry point. It hand-prepares its own `INSERT INTO nodes_fts (...)` inside the test body — and the copy is not even identical to production, since it uses plain `name` where `pipeline_delta.c` uses `cbm_camel_split(name)`.

So: **revert `pipeline_delta.c`'s INSERT to the original four columns and this test still passes.** It guards `CBM_SQL_FTS_BODY_EXPR`; it does not guard the site that consumes it. That is precisely the failure mode you described better than I did — no compile error, nothing red, `body` silently NULL on the path users hit daily — and the test written to catch it cannot.

The fix is to drive the production delta path: build a base index, add nodes, run the real delta merge, then assert `fts_match_count` finds body text from the newly merged rows. If that is awkward to reach from `test_store_search.c`, it belongs wherever `cbm_delta_patch` is already exercised. A test that reimplements the thing it is testing proves the reimplementation.

## A fourth write site exists

`tests/test_mcp.c:2489` does a `delete-all` plus a four-column `INSERT INTO nodes_fts(rowid, name, qualified_name, label, file_path)`. It stays valid SQL against a five-column table, so nothing goes red — it just leaves `body` NULL and stops mirroring production shape.

Not a correctness bug today, but the `store.h` comment claims `CBM_SQL_FTS_BODY_EXPR` covers *every* `nodes_fts` write site, and that is now inaccurate. Either route that site through the macro too or narrow the claim, because a comment that overstates its own coverage is how the next person gets caught.

## BM25 ranking changes for every query, including nodes with no body

`bm25(nodes_fts)` is called at `mcp.c:3025` and `:3062` with **no column weights**, so every column carries the default 1.0. Adding a fifth column changes the score for every row in the table — including rows whose `body` is empty. Existing `search_graph` result ordering shifts for queries that have nothing to do with prose.

The PR body says "backward compatible — additive column". That is true of the schema and not of the ranking. This is worth an explicit decision rather than an accident: either accept the shift and say so, or pass explicit column weights to `bm25()` so the four original columns keep their current relative influence and `body` is added deliberately. I lean towards explicit weights — it makes the ranking a stated design rather than a side effect of column count — but I would rather hear your reasoning first, since you have been closer to the FTS behaviour than I have.

Alongside that: `Section` is now returned by BM25 *and counted in the total*, which changes both the result set and the count for existing callers. That is the intended effect of the direction call I gave you, so it is correct — it just belongs in the description as a behavioural note.

## Two test weaknesses

- **`markdown_section_no_body` is vacuous.** On `main` every Section docstring is already NULL, so `ASSERT(body == NULL || body[0] == '\0')` passes unchanged with or without your production code. It needs to assert something only true after the change — for instance that a *sibling* section in the same file does have a body while this one does not.
- **`markdown_section_body_capped` hardcodes `500`** rather than `MAX_COMMENT_LEN`, which reintroduces exactly the parallel-constant coupling you removed from production. And its corpus is `"alpha "` repeated — pure ASCII — so the UTF-8 backoff never executes. A multi-byte character straddling the cap is the case that backoff exists for.

## Mechanical

`lint` fails on `src/store/store.h:361`: `CBM_SQL_FTS_BODY_EXPR` is one 273-byte physical line. Valid C via adjacent string concatenation, but past the column limit — split it across continuations and it clears.

Worth knowing that **this is currently blocking everything else**: the `test` job is skipped the moment lint fails, so none of your seven tests have executed in CI even once. Fix the line and you get a real signal.

## What is right, and it is most of it

`CBM_SQL_FTS_BODY_EXPR` as a single definition consumed by both production sites is the correct structural answer — it converts "remember to update both" into "there is only one", which is the durable fix for the class of bug we found. The DDL is likewise single-source. `fts_rebuild_tolerates_malformed_properties` genuinely guards the `json_valid()` wrapper, and `fts_rebuild_upgrades_legacy_schema` proving a four-column database upgrades in place is the test most people would have skipped. Documenting the DROP+CREATE window in the body, the commit message *and* at the function is three places more than it needed to be.

Scope is clean: `Module` stays excluded in both queries, no YAML/JSON promotion, nothing from slice 2 leaked in.

On the benchmark harness — you committed it about twenty minutes after asking whether it was wanted. I have since said yes on #617, so keep it; I mention it only because "asked, then shipped before the answer" is a habit worth not forming. It is 477 of the 859 lines here, which is also worth a sentence in the description so a reviewer knows 55% of the diff is a measurement tool rather than the feature.

Fix the delta test, decide the BM25 weights question, tighten the two tests and split that header line, and I will merge this. Then I will run the real-corpus benchmark on my side as promised.

ShauryaaSharma:
> **Note:** written and posted before I had read your review above — the two crossed. Replying to the review properly in a follow-up; leaving this as posted since the benchmark numbers stand on their own.

Harness is in — `b417ade4`, as `scripts/benchmark-fts-body.c` plus a `benchmark-fts-body.sh` wrapper. The PR description above now carries the full table; this is the part that might change a decision you're weighing.

**Your storage estimate was right.** Beyond the 100k point, where fixed page overhead distorts the ratio, the marginal cost settles at ~24 bytes per node — about **200 MB** extrapolated to 8.5M nodes, inside the +150–400 MB you estimated.

**The 95% figure for the `WHERE` lever does not hold for storage.** Measured, it removes **~66%** of the added index, not ~95%:

| nodes | A (pre-#518) | B (body, all) | C (Section/Module only) | B − A |
|---:|---:|---:|---:|---:|
| 100k | 7.2 MB | 10.2 MB | 7.6 MB | +3.0 MB |
| 200k | 15.0 MB | 19.7 MB | 15.8 MB | +4.7 MB |
| 500k | 36.6 MB | 48.6 MB | 40.7 MB | +12.0 MB |

At 500k the lever takes +12.0 MB down to +4.1 MB. The 95% is correct for *row count* — C indexes 17.8k bodies against B's 157k, which is 11% — but storage doesn't follow row count. Markdown section bodies run substantially longer than function docstrings, and the fifth column carries fixed overhead whether or not a given row has prose. So narrowing buys about two thirds, not nineteen twentieths, and it costs function-docstring search. Given your findability-first call I'd read that as weakening the case for the lever, but it's your trade to make and the harness will re-measure it against whatever corpus shape you think is realistic — the distribution constants are at the top of the file.

**I'm not quoting timings.** My machine shows 39–44% run-to-run spread, which is large next to the delta being measured. The harness detects that and prints a warning rather than letting a noisy column pass as authoritative, so the timing half is there but self-flagged as unreliable here. Size figures are deterministic and unaffected. Your reference-machine run is the one that will produce meaningful wall-clock, and thank you for taking that half.

Two methodology notes, because I got both wrong first and each produced confident-looking nonsense. Running the variants grouped (A,A,A, B,B,B, C,C,C) let drift over the life of the process land entirely on whichever went last, which reported C as *slower* than B despite C doing strictly less work — variants are now interleaved so drift cancels. And the seed was a `unsigned long` xorshift, which is 32 bits on a 32-bit target, so it truncated and the three variants weren't seeing identical corpora; it's `uint64_t` now with a per-variant reset. Both are called out in the file so the next person doesn't reintroduce them.

No rush from my side on slice 2 — it stacks on whatever this lands as, so I'll hold until this has a direction.

ShauryaaSharma:
My previous comment crossed with your review — I posted it without having read this. Apologies; it reads as if I ignored you. Replying properly now.

I checked every finding against the tree rather than taking them on trust. **All six are correct**, and the first one is a real hole in the thing I was most confident about.

## The delta test doesn't guard the delta call site — you're right

`fts_delta_insert_populates_body` contains zero references to `cbm_delta_patch`. It hand-prepares its own INSERT, and as you spotted the copy isn't even faithful — plain `name` where production uses `cbm_camel_split(name)`.

So revert `pipeline_delta.c` to four columns and my test still passes. I wrote a test that proves my own reimplementation works, then described it in the PR body as the guard for a bug whose whole signature is "the tested path works." That's the exact trap I claimed to be closing, and I walked into it while writing the test for it.

Fixing it to drive the real merge: build a base index, add nodes, run `cbm_delta_patch`, then assert `fts_match_count` finds body text from the newly merged rows. If that's awkward from `test_store_search.c` it goes where `cbm_delta_patch` is already exercised. Agreed that a test reimplementing its subject proves the reimplementation.

## Fourth write site — and my comment overstates its own coverage

`tests/test_mcp.c:2489` confirmed: `delete-all` plus a four-column INSERT. Valid SQL, nothing red, `body` left NULL, and it stops mirroring production shape.

The `store.h` comment says the macro covers *every* `nodes_fts` write site, which is now false. I'll route that site through `CBM_SQL_FTS_BODY_EXPR` rather than narrow the claim — the point of hoisting it was to make "there is only one" true, and a test fixture drifting from production shape is how the next person gets a misleading green run. You're right that a comment overstating its coverage is worse than no comment.

## BM25 weights — you asked for my reasoning, and it changes the recommendation slightly

Confirmed: `bm25(nodes_fts)` at `mcp.c:3025` and `:3062`, no weights, so every column sits at the default 1.0.

One thing worth putting on the record before choosing, because it affects what weights can actually deliver: **explicit weights cannot restore the old ordering.** FTS5's bm25 normalises by document length against the corpus average. Column weights scale each column's term contributions, but body tokens still count toward row length and therefore toward the corpus average — so `avgdl` shifts the moment `body` is populated at all, and every row's normalisation shifts with it, including rows with empty bodies. Even `body` weighted 0.0 would not pin the ranking, because a zero weight removes the contribution but not the length.

So "additive column, backward compatible" is wrong about ranking in a way weights can't repair, and I should not have written it. The real choice is not *whether* ranking moves but whether it moves deliberately.

Given that, I agree with your lean, for a slightly different reason than preservation: weights are worth setting because they let us **bound how far prose can outrank a symbol match**, which is the actual risk. A term in a name is a stronger signal of intent than the same term in a paragraph. I'd propose weighting `body` below `name`/`qualified_name` rather than at parity — findability-first still argues for indexing it, but not for letting a passing mention in prose outrank the function actually named that.

I don't want to pick the numbers unilaterally since you own the ranking model, and I'd rather tune them against the label-boost CASE than guess. Happy to propose a concrete set with before/after orderings on a sample corpus if that's useful. And the `Section`-now-returnable behavioural note belongs in the description regardless — that was intended, but intended is not the same as documented.

## The two weak tests

Both fair.

`markdown_section_no_body` is vacuous exactly as you say — `docstring` is NULL for every Section on main, so it passes with my production change reverted. Rewriting it so a sibling section in the same file *does* have a body while this one does not, which is only true after the change.

`markdown_section_body_capped` hardcoding 500 reintroduces the parallel-constant coupling I'd just removed from production — that's the same mistake one layer down, and I didn't notice because the test passed. It'll assert against `MAX_COMMENT_LEN`. And the corpus being `"alpha "` repeated means the UTF-8 backoff has never executed in any run of this suite; I'll put a multi-byte character straddling the cap boundary, which is the only case that code exists for.

## Lint, and what it cost

`store.h:361` is a single 274-character line. My continuations collapsed when I generated the header and I didn't re-check the file's column widths afterwards — I checked the benchmark file and assumed the header was fine.

The expensive part is the one you pointed out: lint failing skipped the test job, so **none of these seven tests have run in CI even once**. Every "verified" claim I've made rests on local runs on a toolchain that stubs TRE regex and can't build the product binary. Splitting that line is the first thing I'll do, so the next push produces a real signal rather than my assurances.

## On shipping the harness before your answer

Fair, and I'd rather you said it. I asked whether it was the evidence you wanted specifically so I wouldn't measure the wrong thing twice, and then didn't wait for the answer — which made the asking decorative. Noted, and I'll wait next time.

I'll add the note that `scripts/benchmark-fts-body.c` is 477 of the 859 lines, so the diff size reads honestly as a measurement tool rather than feature code.

---

Plan, in the order I'll push them: split the header line so CI actually runs, then the delta test against `cbm_delta_patch`, the two test fixes, the fourth write site, and the description notes. The BM25 weights I'd like your steer on before I touch — it's your ranking model, and it's the one item here that's a design decision rather than a defect.

Thanks for checking the delta test by reading what it does rather than what it's called. That's the finding I'd least have caught myself.

ShauryaaSharma:
Five of the six items are pushed. The BM25 weights decision is the one left, and it is yours.

**Lint, and the signal it was hiding.** `store.h:361` is split. My first attempt still failed, and the log was precise about why — I had anchored the backslashes on the macro's last line, but clang-format aligns to the longest line *carrying* a backslash, and a macro's final line carries none. Column 41, not 78, which cross-checks against the existing multi-line macro at `cli.c:2983`.

CI is green now, all 34 checks, and the `test` job has run for the first time on this PR — including MSan, TSan and LSan across ubuntu/ubuntu-arm/macOS/Windows. Every claim I made before this rested on local runs against a toolchain that stubs TRE regex and cannot build the product binary, so this is the first real signal any of it has had. You were right that the skipped test job was the expensive part.

**The delta test.** Replaced with `pipeline_delta_merge_indexes_body` in `test_pipeline.c`, which drives `cbm_delta_preseed` then `cbm_delta_patch` rather than reimplementing the INSERT. I verified it discriminates rather than assuming: against a four-column revert of `pipeline_delta.c` the body assertions (`canary`, `rollback`) fail while the name assertion (`deployment`) still passes — the node merged, only its body was lost, which is exactly the regression.

Driving the production path immediately paid for itself by surfacing something the hand-rolled version could not: `cbm_delta_preseed` lifts the gbuf id watermark above `MAX(id)`, so a node added to the gbuf *before* that call keeps a low temp id and is silently never merged by the `id > max_db_id` predicate. My first attempt did exactly that and merged zero nodes. Documented in the test.

**The fourth write site.** Rather than routing `test_mcp.c:2489` through the macro, it now calls `cbm_store_fts_rebuild` — the production function. A fixture that spells out its own column list has to be remembered on every schema change; one that calls the rebuild cannot drift. That was the last `nodes_fts` INSERT outside the store and the delta patch, so the `store.h` comment no longer overstates its coverage, and I reworded it to say what is actually true.

**The two weak tests.** `markdown_section_no_body` now extracts two sections and asserts the documented sibling *has* prose while the bare one does not — a pair only true once capture works and stops at the heading boundary. `markdown_section_body_capped` asserts against `CBM_MAX_COMMENT_LEN`, which I promoted into `cbm.h` so production and tests share one definition instead of the test copying the literal; a test-layer copy is the same coupling this branch removed from production. Its corpus is now 498 ASCII bytes followed by U+20AC so that character straddles offsets 498–500. Measured result is 498 bytes, not 500 — the backoff branch executes, which it never did once under the old all-ASCII corpus.

**BM25 weights — yours to call, with one correction that narrows it.**

Confirmed no column weights at `mcp.c:3025` and `:3062`. But explicit weights cannot restore the old ordering, which I think changes the decision: FTS5 normalises by document length against the corpus average, and body tokens count toward row length regardless of weight. `avgdl` therefore moves the moment `body` is populated, and every row's normalisation moves with it — including empty-body rows. Even `body` at 0.0 removes the contribution but not the length.

So "additive column, backward compatible" was wrong about ranking in a way weights cannot repair, and I have corrected that in the description rather than leaving it standing. The real choice is not whether scores move but whether they move deliberately.

That is why I still agree with your lean, for a different reason than preservation: weights are worth setting to bound how far prose can outrank a symbol match. A term in a name is a stronger signal of intent than the same term in a paragraph, so I would put `body` below `name`/`qualified_name` rather than at parity. I have not picked numbers — it is your ranking model and it interacts with the label-boost CASE. Happy to propose a set with before/after orderings on a sample corpus if that is useful.

**Description** now carries the `Section`-returnable and ranking-shift notes as behavioural changes rather than burying them, and says plainly that the harness is 477 of the ~934 added lines so the diff size reads honestly.

On shipping the harness before your answer — fair, and noted. I asked precisely so I would not measure the wrong thing twice, then did not wait, which made the asking decorative.

DeusData:
Hi @ShauryaaSharma — thank you, genuinely, and I owe you a straight explanation.

**What's changing:** I'm landing the prose-indexing work in-house rather than through #1778. That reverses what we agreed on 2026-08-21, when you said you still wanted to carry it and I accepted. The reason is scheduling on my side, not anything about your work — I've been clearing a long PR backlog, and I'd rather this ship now than send you through another rebase that my own delay caused. Your commit is credited with `Co-authored-by:` on the change.

**The two things I owed you, answered properly:**

1. **BM25 column weights.** You argued `body` should sit below the identifier columns, and you were right. Rather than guess, I checked: FTS5 computes `f(qi,D) = Σ_c w_c · n(qi,c)` — it applies column weights to per-column term frequency **before** the tf-saturation term, which is exactly what Robertson, Zaragoza & Taylor's BM25F prescribes, so weighting here is principled rather than a hack. And the field-weighting literature is consistent that body text takes the lowest weight (typical title:body 3:1–10:1). Landing with `bm25(nodes_fts, 1.0, 1.0, 1.0, 1.0, 0.3)`.

2. **The benchmark I said I'd run.** Your storage number holds up. I also measured the `avgdl` shift you flagged, on two real indexes: cbm's own (29.2% of nodes carry a docstring) moves 22.68 → 28.76 tokens, **+26.8%**; a Kotlin corpus (7.4%) moves **+4.9%**. Your concern was well-founded — bodiless rows gain about **+9.5%** score at `b=0.75`. Worth knowing: no weight setting removes that, because `w_c` acts on `tf` and never on `|D|`, and FTS5 hardcodes `b=0.75`. Only a separate prose table would, and I judged that machinery not worth it here.

**One thing you'd want to know:** main has since grown a *third* FTS write site at `pipeline_delta.c:528` that didn't exist when you opened #617. It writes four columns, so `body` would land NULL for every node arriving via delta merge, while a full reindex looked perfect. That's folded into the change.

On your `pipeline_incremental.c` hunk — you were right that it targeted deleted code; there are zero `nodes_fts` references in that file now.

For the record: you stayed responsive across two months, you declined to fabricate a benchmark your toolchain couldn't produce, and you flagged your own dead hunk unprompted. That's the behaviour I wish every PR had, and #703 is still load-bearing here. I'd welcome your next one.

I'll leave #617 and #1778 open until this lands, then close both citing the merge SHA.

DeusData:
Landed. **BM25 prose search is on main as of `be7b5782`** (PR #1882), closing #518 and #519.

@ShauryaaSharma — thank you. This was your design and you carried it for two months; the merged commit credits you with `Co-authored-by:`. Closing this ledger PR and #1778 now that the work is in.

**What shipped**, and how it differs from where this thread left off:

- `nodes_fts` gains a fifth `body` column, backfilled from each node's existing `properties.docstring` — no schema change, no `user_version` bump, **no forced reindex**. Legacy four-column databases are detected at runtime and degrade to "no prose" rather than erroring, with tests pinning that.
- Markdown section bodies (#518) and top-level YAML/JSON `description`/`summary`/`purpose` (#519) now get docstrings, and `Section`/`Module` are no longer excluded from BM25 results.
- Column weights landed as **`bm25(nodes_fts, 1.0, 1.0, 1.0, 1.0, 0.3)`** — your instinct that `body` belongs below the identifier columns was right, and the research backs it: FTS5 applies weights to per-column term frequency *before* the saturation term, which is what Robertson/Zaragoza/Taylor's BM25F prescribes, and the field-weighting literature consistently gives body text the lowest weight.

**Two things you raised that turned out to matter more than either of us thought:**

Your concern about scores shifting was well-founded, and it is now measured: cbm's own corpus moves `avgdl` 22.68 → 28.76 tokens (**+26.8%**; 29.2% of nodes carry a docstring), a Kotlin corpus +4.9%. That works out to roughly **+9.5%** score change for a bodiless row. No weight setting removes it — `w_c` acts on `tf`, never on `|D|`, and FTS5 hardcodes `b=0.75`. Only a separate prose table would, and that machinery was judged not worth it here. Documented rather than buried.

And main had grown a **third** FTS write site at `pipeline_delta.c` that this PR predates. Left alone, `body` would have landed NULL for every node arriving via delta merge — the warm path most users hit — while a full reindex looked perfect. All three write sites now route through one shared helper, and the regression test for it fails without the fix.

You were also right that your `pipeline_incremental.c` hunk targeted deleted code.

For the record: you stayed responsive across two months, declined to fabricate a benchmark your toolchain could not produce, and flagged your own dead hunk unprompted. #703 is still load-bearing here — it is the origin of this repo's `cbm_fopen` rule. I would welcome your next one.

PR #1781, opened 2026-08-21
fix(codex): migrate skill to documented user path

agn-7:
## Summary

- install the Codex user skill at the documented `$HOME/.agents/skills` root while keeping config, hooks, and profiles under `$CODEX_HOME`
- migrate byte-identical installer-owned legacy copies from `$CODEX_HOME/skills` only after the canonical copy is written and verified
- preserve foreign, modified, linked, or unsafe content unless `--force` explicitly authorizes replacement
- cover canonical/legacy reconciliation in plan, dry-run, repeat install, failure, force, and uninstall paths

## Rationale

This follows the opportunity recorded when #977 was closed. Current Codex builds discover both `$HOME/.agents/skills` and `$CODEX_HOME/skills`, but the public documentation names `$HOME/.agents/skills` as the user-scoped location. Permanently writing the same named skill to both roots creates duplicate selector entries, so this change converges provably owned installs to one canonical copy instead.

The legacy path is not treated as broken. Unowned legacy content remains in place and prevents creation of a duplicate canonical copy; `--force` is required to replace and migrate it.

## Safety

- destination publication is verified before owned legacy cleanup
- cleanup uses the existing exact-content ownership APIs and preserves sibling files
- equal or same-file roots are deduplicated
- hook-preflight failures cannot let another shared-root client create a second Codex copy
- uninstall checks canonical and legacy roots independently and removes only exact-owned content

## Verification

- `make -f Makefile.cbm cbm`
- `scripts/smoke-test.sh build/c/codebase-memory-mcp --agent-config-only`
- CLI suite: all new Codex tests and the extended preflight regression pass; local aggregate is `286 passed, 1 failed` because the host-installed Hermes CLI triggers the pre-existing `cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index` environment-sensitive failure
- `make -f Makefile.cbm lint-format CLANG_FORMAT=/Library/Developer/CommandLineTools/usr/bin/clang-format`
- security layers pass, including MCP fuzz `32/32` when run outside the filesystem/network sandbox required for local IPC

`scripts/lint.sh` could not run `clang-tidy` or `cppcheck` locally because those executables are not installed; the no-skips and clang-format checks pass.

## Related Work

#1739 also introduces `cleanup_actions_planned` for a separate Codex `AGENTS.md` cleanup. This PR uses the same receipt field for legacy skill cleanup and does not change `AGENTS.md` behavior. If #1739 merges first, the resolution is to retain the shared field and both cleanup action types.

Follow-up to #977. Related merged hardening: #1373.


DeusData:
Thank you for the detailed ownership and migration work. I checked current `main`: Codex configuration honors `CODEX_HOME`, and the installer currently derives the Codex skill directory as `$CODEX_HOME/skills`; install and uninstall both use that same legacy location. The local migration premise is therefore grounded.

I have labeled this as an editor-integration, UX, and installer-security item and routed it for detailed review. The external Codex path contract and the exact-owned cleanup behavior still need to be verified together before any merge decision, especially because this changes where durable user content is written and removed. The PR is one coherent migration despite its size, so I am not asking for an artificial split at triage. Our review queue is full, and this may take a little time. Thank you for covering plan, dry-run, repeat-install, failure, force, and uninstall paths.

DeusData:
Direction confirmed — thank you for doing the homework on this. I re-read Codex's current skills documentation today: the USER scope is `$HOME/.agents/skills`, `$CODEX_HOME/skills` is not among the documented locations at all (only `~/.codex/config.toml` is, for enabling/disabling), symlinked skill folders are followed, and two same-named skills are both shown rather than merged — which is exactly the duplicate a Codex+Zed/OpenHands/Warp user gets from us today, since main already installs the shared root for those clients. So moving our Codex install to `~/.agents/skills`, migrating installer-owned legacy copies, and cleaning both roots on uninstall is the right change, and I verified it on a tree merged with current main: `cli` 312/312, the six new tests go red with main's `cli.c`, the Codex block of the smoke passes standalone against the product binary, lint is clean.

Two things before it merges:

1. **Ownership must cover released skill text, not just the current bytes.** `src/cli/cli.c:7910` judges a legacy `$CODEX_HOME/skills/codebase-memory/SKILL.md` as installer-owned only if it matches the *current* skill contents (`released_contents=NULL`). The skill text changed at v0.10.0 and v0.10.3, and a plain install/update never rewrote an existing skill, so every install older than v0.10.3 is "not installer-owned" under this rule: no migration, a warning on every install (`:8046`), and — the part that turns a no-op into a regression — the shared-root install for Zed/OpenHands/Warp is skipped (`:7823` / `:7929`), order-dependently. Please accept any previously released skill text as owned (the same `released_contents` mechanism the hook documents use), and never let an unowned Codex copy block the other clients' install; the copy can simply be left in place with the one warning.

2. **Rebase** — main changed all four files since your base (`README.md` and `tests/test_cli.c` conflict textually; the `cli.c` and smoke hunks merge clean). Same PR, same sign-off.

Minor, take or leave: a symlinked skill directory now records an install error (`:7875-7908`) where main silently skipped; since Codex itself follows symlinked skill folders, treating a symlink whose target is ours as owned would be friendlier.

With those two this merges on green — and it closes a real duplicate-skill annoyance, not just a doc mismatch.

agn-7:
Implemented the requested changes and rebased this branch onto the latest `main` (`2058d49a`).

### Released-content ownership

- Audited every consolidated `codebase-memory` skill release from `0.6.0` through `v0.10.8`.
- Added the four byte-distinct released payloads to the installer's `released_contents` ownership set:
  - `0.6.0` through `v0.9.0`
  - `v0.9.1-rc.1`
  - `v0.10.0` through `v0.10.2`
  - `v0.10.3` through `v0.10.8`
- Exact released payloads now migrate from the legacy Codex path and upgrade in place at the shared path.
- Modified or otherwise user-owned files remain untouched unless `--force` is used.
- Uninstall recognizes and removes exact historical installer-owned payloads as well as the current payload.

### Shared-root behavior

- Removed the global shared-skill suppression. A foreign Codex legacy skill no longer blocks OpenHands, Zed, Warp, or another shared-root client from installing the canonical skill.
- Migration remains destination-first: the canonical skill is verified before an installer-owned legacy copy is removed.

### Verification

- CLI suite: **326 passed**
- `scripts/smoke-test.sh build/c/codebase-memory-mcp --agent-config-only`: **passed**
- `make -f Makefile.cbm security`: **passed**
- `git diff --check`: **passed**

The optional final-directory symlink case is not included, keeping this update focused on the required compatibility and migration behavior.

agn-7:
CI update: every substantive check passed except `test / test-windows-guards`.

The failing guard is the exact known `section_cold_storm` runner race tracked in #2057:

```text
RED: cold-storm client 0 failed (racing daemon spawn):
codebase-memory-mcp: secure CLI coordination could not be created (endpoint)
```

Failed job: https://github.com/DeusData/codebase-memory-mcp/actions/runs/34943783091/job/104301975858

This branch does not modify daemon coordination, IPC, or `tests/windows/test_daemon_stability.py`. All three PR smoke jobs (macOS, Ubuntu, and Windows), lint, CodeQL, security, sanitizers, diagnostics, and the complete Unix/Windows matrix otherwise passed.

I attempted a failed-only rerun, but GitHub requires upstream repository write/admin permission. Could a maintainer please rerun the failed jobs for run `34943783091`? I have not pushed a no-op commit because #2057 explicitly calls out changing the commit merely to reroll this guard as contrary to the CI-determinism rule.

DeusData:
Thank you for addressing the ownership follow-up and removing the shared-root suppression. Those reported changes move this back to maintainer review; they are not being described as an independently completed security review. We need more time to review the current head and attribute its CI results before an integration decision. No further speculative rebase is requested.

PR #1782, opened 2026-08-21
security: add a hidden-instruction gate and refuse PR changes to CI

DeusData:
Two related holes. The second is what makes the first enforceable.

## What this refuses

**Layer 0, on every tracked file** — hidden-instruction carriers: zero-width sequences, bidirectional overrides, the Unicode Tags block, stray BOMs. Plus obfuscation that defeats pattern matching: Cyrillic homoglyphs, letter-spacing, fullwidth and mathematical-bold compatibility forms. Plus prose smuggled into a generated parser's symbol table. Whole tree, ~23s.

**Pull-request metadata** — title, body and commit messages, which nothing scanned before. Stricter ruleset than files: also refuses collapsed sections, embedded markup, event-handler images, `javascript:`/`data:` URIs, and comment syntax invisible on the rendered page.

**CI-defining paths** — any pull request touching `.github/`, `scripts/`, `test-infrastructure/` or `tests/*.sh`.

## Why patterns, not a wordlist

The persuasion half of an injection is natural language: unbounded, translatable, homoglyph-substitutable. Published refusal rates fall from ~79% in English to ~23% in some low-resource languages, so a keyword list is structurally inadequate rather than merely noisy.

The **carrier** half is finite, language-independent, and has no legitimate use in source. That gates at a false-positive rate near zero.

So this catches **hiding, not persuasion**. A green result means *nothing is concealed from the reviewer* — which is what makes human review trustworthy. It is not "no injection", and the script's own header says so. Measured: **10.1%** against visible prose, **99.1%** against the same text smuggled into a parser symbol table. Same payload, different location, opposite verdict.

## Why it cannot be switched off by a pull request

Every other check here is defined by files *inside* the pull request, and `pull_request` runs the merge commit's definition. A hostile PR doesn't defeat a gate — it edits the gate, or deletes the step calling it. `ci-ok` is no help; its own `needs:` list lives in the PR.

`ci-integrity.yml` runs under `pull_request_target`, so its definition comes from the **base branch** and the PR cannot edit its effect on itself. `actions/checkout` resolves to the base commit, so the scanner executed is main's copy. Pull-request content arrives strictly as **data**: paths are matched with `case`, metadata is written to a file and passed by path, never expanded by the shell. Read-only token, no secrets in scope, no build step, no `ref:` at the PR head — the two changes that would turn this into a pwn-request, called out in the header so nobody adds them later.

## Seven bypasses found by attacking it

Each of these worked against an earlier version and is now pinned in the selftest:

| Bypass | Closed by |
|---|---|
| Cyrillic homoglyphs | mixed-script token detection |
| Letter-spacing | consecutive single-letter runs |
| Fullwidth forms | NFKC folding before matching |
| Mathematical bold | NFKC folding |
| **One invalid byte hid a whole file** | decode with replacement; NUL marks true binary |
| Case-sensitive path matching | POSIX character classes |
| Files API 3000-entry cap | fail closed when counts disagree |

Two were found only because a **test** was wrong rather than the code — the selftest applied parser-scoped rules tree-wide and claimed coverage the gate didn't have.

## False-positive discipline

The allowlist has **one** entry: the deliberate injection fixture in `test_cli.c`, where the literal *is* the payload under test. Everything else was fixed by rewording rather than exempting — including two zero-width spaces in `php_lsp.c` and several of this gate's own comments.

Entries pin the sha256 of a single **line**, located by content rather than number, so unrelated edits never churn them while editing blessed text invalidates it. `--update` emits a placeholder the gate rejects, so an exception cannot be produced mechanically.

Every rule was measured before gating. Rules that couldn't be made clean were **dropped**, not loosened: tree-wide hiding-construct detection (61 legitimate uses) and the phrase layer on vendored source (it fired on sqlite3's `int ignoreJump /* Instruction ... */`).

## This pull request refuses itself, twice

It touches `.github/` and `scripts/`, so the path tripwire refuses it. And two commit messages document tokens the metadata scanner detects, so that refuses it too.

**Both are correct, and this lands via break-glass** — which is exactly the path the gate is designed to force for CI changes. Rewriting the history to dodge it was considered and rejected: the friction is the feature.

## Not done here — needs a maintainer decision

**The tripwire is inert until `ci-integrity` is a required status check.** Related: `required_pull_request_reviews` is currently `none`, so `main`'s entire protection is `dco` + `ci-ok` — both defined in files a pull request can rewrite. Enabling required review with code-owner approval on `.github/` and `scripts/` is what turns these from suggestions into controls.

`scripts/benchmark-injection-detection.py` is non-gating and never touches the network implicitly — it exists so the catch rate stays a measurement rather than a claim.


DeusData:
Closing — rescoped.

This solved the problem in the wrong place. The gate here protected the *repository*; the threat it was built for is an agent reading attacker-controlled text and being persuaded by it, and there the agent is the target. Enforcement belongs at the point of consumption, in the local harness (Claude Code / Codex), not in this project's CI — and a checkpoint here only ever saw content arriving through a PR to this repo, while an agent reads issues, comments and discussions everywhere.

Nothing is lost; the work moves to the harness. Recorded here so the reasoning is findable:

- **Seven bypasses** were found by attacking it rather than reasoning about it — Cyrillic homoglyphs, letter-spacing, fullwidth and mathematical-bold forms, a one-byte invalid-UTF-8 skip that hid an entire file, case-sensitive path matching, and the 3000-entry files-API cap. Two surfaced only because a *test* was wrong rather than the code.
- **Measured, not asserted:** 10.1% against visible prose, 99.1% against the same text smuggled into a parser symbol table. Same payload, different location, opposite verdict.
- **Severity is load-bearing.** A binary refuse/allow model rejected 28 of 40 recent PRs, because our own acknowledgement bot posts an HTML comment into nearly every thread. Hiding *mechanisms* must be noted while attack *signatures* refuse — otherwise the tool is switched off within a day.
- **Redaction is a security property, not formatting.** A report meant to be read by an agent must never echo the payload back, or the scanner becomes the delivery mechanism for the injection it just found.

No branch protection changes are needed after all — the required-check and required-review asks are withdrawn along with the tripwire.

Issue #1718, opened 2026-08-18
`--approve-sensitive` never lifts the sensitive-root refusal for Windows paths under `Program Files`


ezonius:
### Version

codebase-memory-mcp 0.10.5

### Platform

Windows (x64)

### Install channel

npm

### Binary variant

standard

## Summary

On Windows, `codebase-memory-mcp allow-root --approve-sensitive <path>` records a sensitive-root grant, but
`index_repository` (and the HTTP UI indexing route) still refuses the exact same path with:

```
C:/Program Files (x86)/Steam/steamapps/common/Factorio/data: path is a home or credential directory.
To index it anyway, run: codebase-memory-mcp allow-root --approve-sensitive C:/Program Files (x86)/Steam/steamapps/common/Factorio/data
```

The grant appears to be recorded correctly (with the `!` sensitive marker), yet the refusal persists for any
path whose first component under the drive is `Program Files` / `Program Files (x86)` — i.e. the escape hatch
is broken exactly where it is needed most.

## Root cause

The refusal is a **byte-exact string equality check failing across Windows path-separator mismatch**.

1. `index_repository` calls `cbm_workspace_root_allowed` (`src/mcp/mcp.c:8074`). Before that, the repo path is
   canonicalized and its separators are normalized to forward slashes:

   ```c
   // src/mcp/mcp.c:1396-1398 (canonicalize_repo_path_if_exists)
   if (cbm_canonical_path(repo_path, real, sizeof(real))) {
       cbm_normalize_path_sep(real);
       ...
   ```

2. The candidate path is classified sensitive because `Program Files (x86)` matches `WS_WINDOWS_SYSTEM_TREES`
   (`src/foundation/workspace.c:159-164`), returning `CBM_WS_DENY_SENSITIVE` from
   `cbm_workspace_classify_root`.

3. A sensitive refusal can only be lifted by an **exact** sensitive-grant match. In `cbm_workspace_root_allowed`:

   ```c
   // src/foundation/workspace.c:548
   if (verdict == CBM_WS_DENY_SENSITIVE && match.exact_sensitive) {
       return true;
   }
   ```

   `match.exact_sensitive` is set in `ws_match_visit`:

   ```c
   // src/foundation/workspace.c:394-395
   if (sensitive && ws_paths_equal(root, m->candidate)) {
       m->exact_sensitive = true;
   }
   ```

[...]

PR #1786, opened 2026-08-21
fix(workspace): treat '/' and '\' as the same separator in root matching (#1718)

umi008:
## What does this PR do?

`--approve-sensitive` never lifted the sensitive-root refusal on Windows for paths under `Program Files`, because the grant is recorded from `cbm_canonical_path` (backslashes) while the indexer hands the candidate already normalized to `/`. `ws_is_ancestor_or_equal` compared the raw bytes, so the recorded grant never matched the candidate and the escape hatch stayed dead exactly where it is needed most.

The comparison now treats `/` and `\` as the same separator on the component boundary, consistent with `ws_is_sep` everywhere else in this module. The containment rule (a separator byte is never part of a component name) is unchanged.

The `ws_paths_equal` home-directory rule in `cbm_workspace_classify_root` uses the same helper, so home detection becomes robust to separator spelling as well. No classification semantics changed beyond separator equivalence.

## Checklist

### DIFF
--- a/src/foundation/workspace.c
+++ b/src/foundation/workspace.c
@@ -212,7 +212,16 @@ static bool ws_any_component_matches(const char *path, const char *const *names,
 }
 
 /* True when b is a or lives under a. Compares on a separator boundary so
- * "/a/bc" is not treated as living under "/a/b". */
+ * "/a/bc" is not treated as living under "/a/b".
+ *
+ * '/' and '\' are equivalent here, as they are everywhere else in this module
+ * (ws_is_sep). The Windows sensitive-grant escape hatch (#1718) records the
+ * grant from cbm_canonical_path — backslashes — while the indexer candidate
+ * arrives normalized to '/'. A byte-exact comparison made the grant never
+ * match the candidate on Windows, so --approve-sensitive could not lift a
+ * "Program Files" refusal. Comparing on the separator-equivalence boundary
+ * keeps the containment semantics (a separator byte is never "part of a
+ * component name") while erasing the platform spelling difference. */
 static bool ws_is_ancestor_or_equal(const char *a, const char *b) {
     if (!a || !b || !a[0] || !b[0]) {
         return false;
@@ -221,8 +230,13 @@ static bool ws_is_ancestor_or_equal(const char *a, const char *b) {
     while (la > 1 && ws_is_sep(a[la - 1])) {
         la--;
     }
-    if (strncmp(a, b, la) != 0) {
-        return false;
+    for (size_t i = 0; i < la; i++) {
+        if (a[i] == b[i]) {
+            continue;
+        }
+        if (!(ws_is_sep(a[i]) && ws_is_sep(b[i]))) {
+            return false;
+        }
     }
     return b[la] == '\0' || ws_is_sep(b[la]);
 }
--- a/tests/test_workspace.c
+++ b/tests/test_workspace.c
@@ -261,6 +261,96 @@ TEST(ws_manifest_approval_refuses_overbroad_requests) {
     PASS();
 }
 
+/* #1718: the sensitive-root escape hatch broke on Windows separator spelling.
+ * The grant is recorded from cbm_canonical_path — backslashes — while the
+ * indexer hands the candidate already normalized to '/'. A byte-exact
+ * comparison made the grant never match the candidate, so an approved
+ * "Program Files" root stayed refused. '/' and '\' are equivalent everywhere
+ * else in this module (ws_is_sep); the boundary comparison must treat them
+ * equally. The classify_root home rule uses the same ws_paths_equal, so a
+ * mixed-separator home spelling is the string-level regression check; the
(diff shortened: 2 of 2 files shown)


DeusData:
Thank you for finding this separator-normalization hole in the sensitive-root grant path. I checked current `main`: `ws_is_ancestor_or_equal()` trims and checks separators at boundaries, but compares the path prefix byte-for-byte, so a recorded backslash path cannot match the same Windows path normalized with forward slashes. That grounds both the approval and home-path mismatch described here.

I have labeled this as a high-priority Windows security bug and routed it for review. The Windows test shard is currently red, so that failure must be attributed before the PR can advance. Our review queue is full, so detailed review may take a little time. Thank you for keeping the containment boundary intact while narrowing the change to separator equivalence.

DeusData:
Here is the attribution that was asked for, so you don't have to go digging: **the Windows red is your own new test, not pre-existing shard noise.**

```
ws_approve_sensitive_matches_across_separator_styles  FAIL
tests/test_workspace.c:347: ASSERT(strstr(err, "--approve-sensitive") != ((void*)0))
```

`workspace rc=1 pass=16 fail=1` on `test-windows (windows-latest, CLANG64, x86_64, 1/2)`. Everything else in the shard is green, and the other three assertions in your test — the exact-grant round trip, the sibling refusal, and the string-level classify cases — all pass. Only the last block fails.

**The likely cause is one missing `mkdir`, not your production change.** The descendant case builds `%s/.ssh/cfg/sub` but never creates it — `cbm_mkdir_p` is called for `.ssh/cfg` only. That matters because `cbm_canonical_path` resolves through a path that must exist: `realpath()` on POSIX, and on Windows an opened handle plus `GetFinalPathNameByHandleW` (`src/foundation/compat_fs.c:957-994`). When canonicalization can't resolve the candidate, `match.contained` doesn't hold, so `cbm_workspace_root_allowed` takes the containment branch at `src/foundation/workspace.c:563` and emits *"is outside the allowed root"* — never reaching the sensitive-hint branch at `:584` that your assertion is looking for. That also explains why the sibling assertion above it passes: it expects exactly that boundary wording.

So the assertion is right about the intended behaviour; the fixture just doesn't put the directory on disk for the platform that insists on it. Creating `sub` the way you create `cfg` should be the whole of it.

To be straight with you about confidence: I read this statically rather than building your branch, so treat it as a well-evidenced hypothesis rather than a confirmed one. The failing assertion, the passing siblings and the existence requirement are all facts; the causal link between them is the inference.

On the change itself — the production hunk reads correct to me. Replacing the byte-exact `strncmp` prefix with a separator-equivalent walk is consistent with the rest of the module: `ws_is_sep` already treats `\` as a separator unconditionally on every platform, and component splitting has always done so, so this removes an inconsistency rather than introducing a new equivalence. Keeping the boundary rule (`b[la] == '\0' || ws_is_sep(b[la])`) untouched is what preserves `/a/bc` not matching `/a/b`, and I checked that it still does.

Thanks for a genuinely well-scoped fix, and for the diagnosis in the description — the "grant recorded with `\`, candidate arrives with `/`" framing is what made this quick to follow.

DeusData:
The production change is right and I verified it the hard way on a tree merged with current main: `workspace` 21/21 plus `daemon_application`, `mcp`, `mcp_mutation_guard`, `extraction` (724 passed, no sanitizer hits), lint clean; swapping main's `workspace.c` back in makes your new test fail exactly at the separator-style case (`test_workspace.c:394`, ALLOW instead of DENY_SENSITIVE), so it binds; and a boundary probe through the same `ws_paths_equal` path confirms `C:\a\b` ≡ `C:/a/b` in both orders and with mixed or trailing separators, while `C:/a/bc`, `C:/a/b/c` and `C:/a` are still rejected. The #1718 shape is reachable end to end (`resolve_canonical_path` normalizes on Windows before the grant match).

What keeps CI red is the test itself, not the fix. `test / test-windows (CLANG64, x86_64, 1/2)` fails at `tests/test_workspace.c:347` — `ASSERT(strstr(err, "--approve-sensitive"))` — because the descendant path `<cache>/.ssh/cfg/sub` is asserted on but never created. On Windows, `cbm_path_within_root` canonicalizes through `CreateFileW(OPEN_EXISTING)` (`compat_fs.c:968`), which returns nothing for a nonexistent path, so containment fails and the error you get is "outside the allowed root" rather than the sensitive-root hint. Two lines:

1. right after the `/`-conversion loop for `descendant` (around lines 340–344), add `ASSERT_TRUE(cbm_mkdir_p(descendant, 0700));` before the `cbm_workspace_root_allowed(descendant, …)` call;
2. line 302: `fopen(path, "w")` → `cbm_fopen(path, "wb")` — repo rule for anything touching repo/user paths; functionally harmless here since the grant reader strips `\r`.

Keep the sign-off and this merges on green — the Windows shard is the confirmation. Thanks for the clean fix; the separator equivalence at the containment boundary is exactly where it belonged.

Issue #1714, opened 2026-08-18
check_index_coverage: freshness=metadata_changed for every file, immediately after a full re-index (Windows)


ivbrajkovic:
### Version

codebase-memory-mcp 0.10.6 (also reproduced on 0.10.3)

### Platform

Windows (x64) — Windows 11 build 26200, NTFS, system timezone UTC+02:00

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

`check_index_coverage` returns `freshness: "metadata_changed"` with `recommended_action: "read_source_and_reindex"` for **every** path, **always** — including files that were last modified before the index was written, checked seconds after a successful full re-index. `status` stays `no_recorded_issue` and `coverage` is empty. The graph itself is correct and in sync (verified by a create/delete round-trip through the auto-watch daemon and by `get_code_snippet` output matching source byte-for-byte).

Expected: `freshness: "metadata_match"` for unchanged files, as seen on other platforms (e.g. the report in #1613 shows `metadata_match` on macOS).

Impact: on Windows the freshness field carries no signal. An agent following the documented guidance ("read flagged source and qualify claims when metadata is changed") is told to distrust or re-index a graph that is actually current.

### Reproduction

1. Code — one file `src/greet.ts` (dummy snippet) in an empty directory `%TEMP%\cbm-freshness-repro`, no git init:

```ts
export const greet = (name: string): string => `hello, ${name}`;
```

2. Commands (v0.10.6, run back to back):

```
codebase-memory-mcp cli --json index_repository '{"repo_path":"C:/Users/<user>/AppData/Local/Temp/cbm-freshness-repro"}'
codebase-memory-mcp cli --json check_index_coverage '{"project":"C-Users-<user>-AppData-Local-Temp-cbm-freshness-repro","paths":["src/greet.ts"]}'
```

3. Result — index reports `status: indexed, nodes: 6, edges: 5`; the coverage call ~2 s later returns:

```json
[{
  "requested_path": "src/greet.ts",
  "path": "src/greet.ts",
  "status": "no_recorded_issue",
  "freshness": "metadata_changed",
  "recommended_action": "read_source_and_reindex",
  "coverage": []
}]
```

The file's mtime (09:53:36Z) is 12 s **older** than the index generation (09:53:48Z) and the file was not modified between the two commands, so this should be a match. Expected: `metadata_match`.

[...]

PR #1787, opened 2026-08-21
fix(mcp): report freshness with the indexer's own mtime source (#1714)

umi008:
## What does this PR do?

`check_index_coverage` reported `freshness: "metadata_changed"` for **every** file on Windows, always — including files checked seconds after a successful full re-index. The freshness reader recomputed `mtime_ns` from `struct stat` (which on Windows truncates to seconds via `st_mtime`), while the indexer records the FILETIME-derived nanosecond value from `cbm_path_info_utf8`. A byte-identical file therefore never matched, so the freshness field carried no signal and agents following the documented guidance were told to distrust a graph that was actually current.

Both readers now compare against the same source the writer uses:

- `coverage_path_freshness` (mcp.c) derives the comparison mtime/size from `cbm_path_info_utf8` — the exact function the indexer records from — instead of a platform-dependent `stat` conversion. The Windows-only `st_mtime` truncation path is gone.
- The same truncation existed in the pipeline's incremental change-detection (`pipeline_incremental.c`), where it would mark every file changed on Windows; it now uses `cbm_path_info_utf8` too, on all platforms.

## Checklist

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -4204,17 +4204,6 @@ static coverage_path_result_t coverage_normalize_rel(const char *input, bool all
     return written > 0U || allow_root ? COVERAGE_PATH_OK : COVERAGE_PATH_INVALID;
 }
 
-static int64_t coverage_stat_mtime_ns(const struct stat *st) {
-#ifdef __APPLE__
-    return ((int64_t)st->st_mtimespec.tv_sec * (int64_t)CBM_NSEC_PER_SEC) +
-           (int64_t)st->st_mtimespec.tv_nsec;
-#elif defined(_WIN32)
-    return (int64_t)st->st_mtime * (int64_t)CBM_NSEC_PER_SEC;
-#else
-    return ((int64_t)st->st_mtim.tv_sec * (int64_t)CBM_NSEC_PER_SEC) + (int64_t)st->st_mtim.tv_nsec;
-#endif
-}
-
 static const char *coverage_path_freshness(cbm_store_t *store, const char *project,
                                            const char *root_path, const char *rel_path,
                                            bool *outside) {
@@ -4228,8 +4217,13 @@ static const char *coverage_path_freshness(cbm_store_t *store, const char *proje
     if (n < 0 || (size_t)n >= sizeof(abs_path)) {
         return "unavailable";
     }
-    struct stat st;
-    if (stat(abs_path, &st) != 0) {
+    /* #1714: mtime_ns must come from the SAME source the indexer recorded it
+     * with — cbm_path_info_utf8. Recomputing from struct stat truncates to
+     * seconds on Windows (st_mtime), while the file_hashes record carries
+     * FILETIME-resolution nanoseconds, so a byte-identical file never matched
+     * and every path was reported metadata_changed. */
+    cbm_path_info_t info;
+    if (cbm_path_info_utf8(abs_path, &info) != 0) {
         return "missing";
     }
     if (!cbm_path_within_root(root_path, abs_path)) {
@@ -4245,7 +4239,7 @@ static const char *coverage_path_freshness(cbm_store_t *store, const char *proje
     if (rc != CBM_STORE_OK) {
         return "unavailable";
     }
-    bool matches = hash.mtime_ns == coverage_stat_mtime_ns(&st) && hash.size == st.st_size;
+    bool matches = hash.mtime_ns == info.mtime_ns && hash.size == info.size;
     cbm_store_clear_file_hash(&hash);
     return matches ? "metadata_match" : "metadata_changed";
 }
--- a/src/pipeline/pipeline_incremental.c
+++ b/src/pipeline/pipeline_incremental.c
@@ -102,18 +102,6 @@ static const char *itoa_buf(int v) {
     return buf[idx];
 }
 
-/* ── Platform-portable mtime_ns ──────────────────────────────────── */
-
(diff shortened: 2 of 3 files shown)


DeusData:
Thank you for tracing the Windows freshness mismatch to two different timestamp sources. I checked current `main`: `coverage_path_freshness()` converts `struct stat` time, which uses second precision on the Windows branch, while the index path already records `cbm_path_info_utf8` nanoseconds. The incremental comparison has the same platform-dependent `st_mtime` fallback. That grounds both false-change paths described here.

I have labeled this as a high-priority Windows parsing and correctness bug and routed it for review. The current CI surface is green. Our review queue is full, so detailed review may take a little time. Thank you for converging the readers on the writer's existing source of truth.

chao-hsu-14:
Independent Windows confirmation from `codebase-memory-mcp 0.10.8`:

- Immediately after a cold full re-index, `generation_matches` was true and coverage recording was complete, but every sampled path still returned `status: no_recorded_issue` with `freshness: metadata_changed`.
- I compared the persisted `file_hashes` row with the live file: SHA-256, size, and the full FILETIME-derived `mtime_ns` matched. The mismatch was isolated to the coverage reader rebuilding the timestamp as whole-second `st_mtime * 1e9`.
- I reviewed commit `455e4fb`: using `cbm_path_info_utf8` in both `coverage_path_freshness` and incremental `classify_files` aligns both readers with `semantic_manifest_hash_file`, which is the writer of the stored metadata. This matches the observed root cause and also addresses the same precision mismatch in warm incremental indexing.

The PR Windows CI and regression are green. I could not independently rebuild the PR binary locally because this machine does not have the required MSYS2 CLANG64 toolchain, so this is an independent reproduction plus source-level validation, not a separate binary test.

DeusData:
Approved on merit, and **the impact is larger than your description claims** — I want to spell that out, because it strengthens the case rather than weakening it.

## Verified

Both truncations are real and both are Windows-only:

- `mcp.c:4725` `coverage_stat_mtime_ns` — `#elif defined(_WIN32)` returns `st->st_mtime * NSEC_PER_SEC`, seconds resolution.
- `pipeline_incremental.c:107` `stat_mtime_ns` — the same `#elif defined(_WIN32)` branch, same truncation.

macOS uses `st_mtimespec.tv_nsec` and Linux `st_mtim.tv_nsec`, so both agree with the writer and neither is affected. Your diagnosis is exact.

**And the repo already states the rule you are enforcing.** `src/pipeline/artifact.c:534-538`:

> *The disk side MUST be read with `cbm_path_info_utf8`, because that is the [function the indexer] calls … On POSIX any lstat-based helper agrees with it, but on Windows `cbm_path_info_utf8` derives mtime_ns from `ftLastWriteTime`.*

So this is not a new convention — it is an existing, documented invariant that two readers were quietly violating. That is the cleanest possible justification for the change, and I would put that citation in the PR description.

## The part you understated

You describe the `pipeline_incremental.c` half as "it would mark every file changed on Windows". Following it through, `pipeline_incremental.c:693` is:

```c
if (stat_mtime_ns(&st) != h->mtime_ns || st.st_size != h->size) {
    changed[i] = true;
```

With a seconds-truncated left side and a FILETIME-derived right side, those can only agree when the sub-second component is exactly zero. So on Windows **incremental reindex has been degenerating into a full reindex on essentially every run** — that is a performance defect on top of a correctness one, and on a large repository it is the more expensive of the two.

The `check_index_coverage` half is the worse *kind* of bug, though. A tool that reports nothing is annoying; a tool that reports `metadata_changed` for a graph indexed seconds ago actively misleads, and the documented guidance tells agents to act on it. Silent wrongness beats loud absence every time, and this was the wrong side of that.

## Before merge

**Please tick or explain the lint box.** `make -f Makefile.cbm lint-ci` is unchecked in your checklist. If it failed with whole-file `clang-format` drift in files you did not touch, that is a known false positive — our gate requires the **Homebrew LLVM** `clang-format`, and a distro or standalone `clang-format-20` reports spurious differences. Just say so and I will treat it as clean.

**You are 205 commits behind `main`**, so the green on this PR is stale and means little. Normally I would rebase it for you, but our Actions pool is badly backlogged right now (31 runs queued) and every rebase adds a full matrix to that queue, so I am deliberately holding off. I will refresh it once the queue drains, or you can rebase whenever suits you.

Nothing to change in the code. Thank you — this is a well-isolated fix to a bug that made a correctness signal actively harmful on one platform.

DeusData:
Merged — with an apology attached. This was **approved on merit on 2026-09-02** and then sat green for eighteen days. Nothing was outstanding from you; the delay was entirely ours.

Reading it back before merging, the maintainer note from 09-02 holds up: the impact is larger than your own description claimed. `coverage_stat_mtime_ns` derived the timestamp from `struct stat`, whose Windows branch carries **second** precision, while the index path recorded nanoseconds from a different source. Two sources for one comparison is a bug that cannot be tuned away — it is wrong whenever the two disagree, and on Windows they disagree routinely. Taking the value from the same source the indexer recorded it from is the only fix that closes it rather than narrowing the window.

I also want to credit **@chao-hsu-14** here: their independent confirmation on 0.10.8 — `generation_matches` true and coverage recording complete, yet every sampled path still reporting `freshness: metadata_changed` — is what turned this from a plausible diagnosis into a settled one. Two people arriving at the same place from different directions is the strongest evidence a bug report can carry.

Thank you both. Sorry it took so long to act on work that was ready.

Issue #1750, opened 2026-08-19
Bug: Stale AST line coordinates in get_code_snippet / search_code after file modification


0xkhalz:
### Version
v0.10.7

### Platform
Linux (x64)

### Install channel
Built from source

### Binary variant
standard

### What happened, and what did you expect?
Files with `parse_partial` ranges (and modified files generally) can have stale AST line coordinates after the file is edited on disk without re-indexing.

- `get_code_snippet` reads and slices source lines from the live file on disk using the coordinates stored at index time. When the file has shifted due to edits, it returns shifted/corrupted source text or an entirely different function's body for the requested symbol.
- `search_code` runs `grep` against the live file on disk (obtaining current line numbers) but maps matches to symbols using stored (stale) `[start_line, end_line]` node ranges, causing grep hits to be misattributed to adjacent functions, and in `MODE_FULL` returning drifted source lines.

**Expected behavior:**
When file metadata changes on disk (`freshness != "metadata_match"`), tools should detect drift and avoid returning corrupted source slices or misattributed symbols.

### Reproduction

1. **Initial indexed state:**
`check_index_coverage` on `src/mcp/mcp.c` before edit:
```json
{
  "path": "src/mcp/mcp.c",
  "status": "partial",
  "freshness": "metadata_match"
}
```
Indexed coordinates for `maybe_auto_index`: `11769–11819`.

2. **Modification on disk:**
Insert 50 comment lines immediately above `maybe_auto_index` in `src/mcp/mcp.c`. The function's true location on disk moves from `11769–11819` to `11819–11869`. No `index_repository` or re-index is performed.

3. **Query:**
Call `get_code_snippet` for `maybe_auto_index`.

4. **Actual result:**
Returned:
```
start_line: 11769
end_line:   11819
```
The returned `source` field contained the 50 newly inserted comment lines plus the function signature, rather than the function's actual body at its live location. In other queries, it returned the body of `dispatch_tool` instead.

### Confirmations
- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

PR #1788, opened 2026-08-21
fix(mcp): flag source drift in get_code_snippet and search_code (#1750)

umi008:
## What does this PR do?

`get_code_snippet` and `search_code` sliced the live file on disk using line coordinates recorded at index time. After the file was edited without re-indexing those ranges are stale, and the tools returned shifted source text — a grep hit misattributed to an adjacent function, or a snippet of the wrong symbol — presented as though it were current source.

Both tools now consult the same freshness oracle `check_index_coverage` already exposes. When the recorded metadata no longer matches the file on disk (`metadata_changed`/`missing`), they skip the stale live slice and report `source_drift: true` plus the freshness state instead. The healthy path (`metadata_match`) is byte-identical to before; re-indexing remains the remedy.

Tested with two reproduce-first regression tests: a snippet request and a full-mode search after a file is modified on disk (asserting `source_drift` and no stale text), plus the unchanged clean-metadata path.

## Checklist

### DIFF
--- a/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1750/.openspec.yaml
+++ b/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1750/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-21
--- a/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1750/design.md
+++ b/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1750/design.md
@@ -1,0 +1,52 @@
+## Context
+
+See proposal.md for motivation. The relevant current state (verified against source):
+
+- `get_code_snippet` → `build_snippet_response` (src/mcp/mcp.c:8639) → `resolve_snippet_source` (mcp.c:8506) → `read_file_lines(abs_path, start, end)` slices the LIVE file with `node->start_line/end_line` taken from the graph index. No freshness check anywhere on that path.
+- `search_code` → `classify_all_grep_hits` (mcp.c:9576) attributes grep hits to symbols by `find_tightest_node` over indexed node ranges; `attach_result_source` (mcp.c:9071) then reads the live file with `r->start_line/end_line` (mode full, mcp.c:9107) or around `r->match_lines` (context mode, mcp.c:9126).
+- The drift signal already exists: `coverage_path_freshness` (mcp.c:4218) compares the stored `mtime_ns`/`size` (from `cbm_store_get_file_hash`) with the file on disk. It is only consumed by `handle_check_index_coverage`; the read tools never consult it.
+- Both tools already guard containment via `cbm_path_within_root`; the missing guard is metadata drift, not traversal.
+
+## Goals / Non-Goals
+
+**Goals:**
+- `get_code_snippet` reports drift explicitly instead of slicing stale coordinates.
+- `search_code` never attaches `source`/`context` sliced from a drifted file using stale ranges, and marks such rows.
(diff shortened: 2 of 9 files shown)


DeusData:
Thank you for identifying a correctness problem that can make stale coordinates look authoritative, and for covering both snippet and search paths. I checked current `main`: `get_code_snippet` and `search_code` still consume stored start and end lines for live-file slices without first applying the existing freshness result. That grounds the source-drift failure described here.

There is one mechanical scope blocker before code review. This PR adds 196 lines across seven new `openspec/` files, including project configuration, while current `main` has no `openspec/` tree and `CONTRIBUTING.md` requires prior approval for project configuration. Please remove the entire `openspec/` portion and keep the focused MCP change plus tests. I have labeled the underlying problem as a high-priority parsing and correctness bug. Splitting enables review but does not pre-decide the merge outcome. Thank you for the clear reproduction and fail-closed response shape.

DeusData:
The fix is good and I want it. Two things before it can land — one of them is a merge-order hazard you may not have noticed, and it is the more important of the two.

## Blocker: this must not merge before #1787

Your D1 makes `coverage_path_freshness` the single drift oracle:

> `freshness == "metadata_changed" || "missing"` ⇒ drifted → skip the live read

And your own #1787 exists because **that oracle returns `metadata_changed` for every file on Windows, always** — `coverage_stat_mtime_ns` truncates to seconds under `#elif defined(_WIN32)` while the indexer records FILETIME-derived nanoseconds, so a byte-identical file never matches.

So if this lands first, on Windows:

- `get_code_snippet` returns the placeholder for **every** request,
- `search_code` attaches **no** `source` or `context` to **any** result.

Both read tools would go dark on that platform. Today the broken oracle is merely uninformative because only `check_index_coverage` consumes it; this PR makes it load-bearing. That is not a flaw in either change — each is correct alone — it is purely an ordering constraint, and exactly the kind that merges clean and looks fine in review.

**#1787 goes first.** I have approved it; once it lands I will re-run this one. Worth adding a "depends on #1787" line to the description so nobody merges them the other way round.

## Please drop the `openspec/` files

Seven of the nine files here — `openspec/config.yaml`, `openspec/specs/…`, and the `openspec/changes/archive/…` tree — are spec-framework scaffolding, about 196 lines against a fix of `mcp.c +56/-10` plus its test. `openspec/` does not exist on `main`, no other contributor uses it, and adopting it is a project-level decision rather than something a bug fix should carry in.

We have decided not to adopt it, so **please strip `openspec/` from this PR and keep the code and tests**. Same request on #1789 and #1790.

To be clear about what is being declined: **the framework directory, not the thinking.** Your D1–D4 decisions, the Non-Goals, and the Risks section are genuinely good analysis — D4 in particular (per-item rather than per-response, so one drifted file cannot suppress healthy results) is the right call and I would not have wanted it decided the other way. That reasoning is welcome in the PR description itself, where reviewers read it. It is the permanent `config.yaml` and per-change archive tree in a lean C repo that we are saying no to.

## On the fix itself

The defect is well chosen. Slicing a live file with index-time coordinates and presenting the result as current source is the failure mode this project should care most about — not an error, not an absence, but *plausible wrong text*: a snippet of the adjacent function, or a grep hit attributed to the wrong symbol, with nothing in the response to say so.

Three things I checked and agree with:

- **Reusing the existing oracle rather than inventing a second freshness notion** keeps `check_index_coverage` and the read tools from ever disagreeing about whether a file is current. Two oracles would drift apart.
- **Leaving `unavailable` / `not_tracked` on today's behavior.** A file with no hash record cannot be judged, and refusing would regress reads of foreign files. Declining to widen the guard is the right instinct.
- **Raw grep hits keep their content**, because those lines come from live grep output rather than stale index coordinates. That distinction is easy to miss and you got it right.

Strip `openspec/`, note the #1787 dependency, and this is ready.

Issue #1665, opened 2026-08-16
Read-only repository cannot be indexed on 0.10.5 (0.9.0 succeeds), and the error blames repo_path


DeusData:
### Version

0.10.5 (regression against 0.9.0)

### Platform

Linux (arm64, container); path is platform-independent

### What happens

Indexing a **read-only** repository with `--persistence true` fails outright on 0.10.5, where 0.9.0 completes. The error names the wrong cause and gives the user nothing to act on:

```
{"project":"…","status":"error","hint":"Pipeline failed. Check repo_path exists and contains source files. Try mode='fast' for a quicker diagnostic run."}
```

The repo path exists and is full of source files. `CBM_LOG_LEVEL=debug` adds no further detail, and no `pipeline.err` line reaches the user-visible output.

### Reproduction

Identical mount, identical project directory, only the version differs:

```sh
docker run --rm -v /path/to/corpora:/corpus:ro -v /path/to/bins:/bins ubuntu:24.04 bash -c '
  for v in v090 v0105; do
    D=/tmp/$v; mkdir -p $D
    env HOME=$D CBM_CACHE_DIR=$D/.cache /bins/$v/codebase-memory-mcp cli index_repository \
      --repo-path /corpus/django --mode full --name ro-$v --persistence true
    echo "$v rc=$?"
  done'
```

Result:

| | `--persistence true` | `--persistence false` |
|---|---|---|
| 0.9.0 | **rc=0**, `"status":"indexed"` | rc=0 |
| 0.10.5 | **rc=1**, `"status":"error"` | **rc=0** |

`--persistence false` succeeding on 0.10.5 isolates this to the artifact-export path (`.codebase-memory/graph.db.zst` is written *inside the repo*, which a read-only checkout cannot accept).

### What I verified vs. inferred

**Verified:** the version difference, the `--persistence` split, and that a writable copy of the same tree indexes fine on 0.10.5 (rc=0, 55368 nodes).

[...]

PR #1789, opened 2026-08-21
fix(mcp): attribute post-publish failures to artifact export (#1665)

umi008:
## What does this PR do?

With `--persistence true` on a read-only repository, `index_repository` reported `status: "error"` with the hint *"Pipeline failed. Check repo_path exists and contains source files"* — blaming the repository for a failure that is actually the post-publish team-sharing artifact export (`.codebase-memory/` cannot be written inside a read-only checkout). The graph database itself published fine.

The pipeline now snapshots the artifact export error of the run it just executed (a per-pipeline buffer, not the process-global export error which can describe a previous run). `index_repository` consults that snapshot on failure and names the real cause: the artifact export, the error detail, and the remedies (writable checkout or `--persistence false`). Every other failure path keeps the existing hint byte-for-byte.

The fail-hard semantics are unchanged: a failed persistence export still makes the run fail, which is the in-tree documented decision (`export_after_publish`, "intentionally fails the run"). This change only makes an already-failing run report the truth — the issue-1466-era misleading `repo_path` hint is what hid the cause.

Verified with a reproduce-first pipeline test: an artifact directory blocked (`.codebase-memory` as a file), fixture indexed, asserts `rc != 0`, `cbm_pipeline_export_error()` non-empty and naming the artifact phase, the DB still published, and a clean re-run leaving the snapshot empty.

## Checklist

### DIFF
--- a/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1665/design.md
+++ b/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1665/design.md
@@ -1,0 +1,45 @@
+## Context
+
+See proposal.md. Verified state:
+
+- `export_after_publish` (src/pipeline/pipeline.c:2492) runs AFTER the DB publish/rename (pipeline.c:2602-2613): the graph DB is already live at `final_path` when the artifact export runs. On failure it logs `pipeline.err` (phase `artifact_export`, err = `cbm_artifact_export_last_error()`) and returns the non-zero rc, which `cbm_pipeline_run` propagates (pipeline.c:2615-2618).
+- `handle_index_repository` (src/mcp/mcp.c:8335-8351): `rc != 0` ⇒ `status: "error"` + the fixed hint "Pipeline failed. Check repo_path exists and contains source files. Try mode='fast'..." — written regardless of where in the pipeline the failure occurred.
+- The fail-hard policy is an in-tree, cited decision ("A failed persistence export intentionally fails the run; this used to be ignored.", pipeline.c; commit 44d51745 "fix: address review feedback on artifact export error surfacing"). This change does NOT touch that policy.
+- `cbm_artifact_export_last_error()` (src/pipeline/artifact.c:67) is global state — it can carry a PREVIOUS run's error, so the response must not consult it directly; the pipeline must snapshot the error of the run it just executed.
+
+## Goals / Non-Goals
+
+**Goals:**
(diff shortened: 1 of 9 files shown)


DeusData:
Thank you for tracing the misleading error to the post-publish artifact phase and for preserving the existing fail-hard contract. I checked current `main`: artifact export logs its own phase error, but the MCP failure path still returns the generic “Pipeline failed. Check repo_path” hint. That confirms the attribution problem.

Before code review, please remove the 135 lines under the five new `openspec/` files. Current `main` has no `openspec/` tree, and `CONTRIBUTING.md` requires prior approval for project configuration. The focused `mcp.c`, pipeline, header, and regression-test changes can remain together as the single bug fix. I have labeled this as a normal-priority UX bug. Splitting enables review but does not pre-decide the merge outcome. Thank you for making the existing failure report the actual cause.

DeusData:
Approved on merit. One thing to remove first.

## Please drop the `openspec/` files

Five of the nine files here are `openspec/` spec-framework scaffolding. `openspec/` does not exist on `main`, no other contributor uses it, and adopting it is a project-level decision rather than something a bug fix should bring in. We have decided not to adopt it — **please strip `openspec/` and keep `src/` plus `tests/`**. Same request on #1788 and #1790.

The analysis itself is welcome; put it in the PR description, where reviewers actually read it. It is the permanent config and per-change archive tree in a lean C repo that we are declining, not your reasoning.

## On the fix

This is a good defect to have chased, because the bug was **a lie rather than a silence**. `index_repository` reported

> Pipeline failed. Check repo_path exists and contains source files

for a run where the repository was fine and the graph database published successfully — the failure was the post-publish artifact export unable to write `.codebase-memory/` into a read-only checkout. A user following that hint goes and checks their repo path, which is correct, and learns nothing. Misdirection costs more than a bare failure.

Two decisions I want to name because they are what make this safe:

**Snapshotting the artifact export error of the run just executed, rather than reading the process-global export error.** That global could describe a *previous* run, so consulting it would occasionally attach a stale, unrelated cause to a fresh failure — which is the same class of bug you are fixing, reintroduced one level up. A per-pipeline buffer is the right shape.

**Leaving the fail-hard semantics alone.** A failed persistence export still fails the run, per the in-tree `export_after_publish` decision. You changed only what an already-failing run *says*, and said so explicitly. That is the difference between a diagnostics fix and a behavior change smuggled in as one, and keeping every other failure path byte-for-byte is what lets me take this without a broad regression hunt.

Strip `openspec/` and I will merge on green. Our Actions queue is heavily backlogged at the moment (30 runs queued), so expect a wait that has nothing to do with this PR.

Issue #1760, opened 2026-08-20
Stale rendezvous directory (`/tmp/cbm-daemon-<uid>`) permanently blocks `install`/`update` with "active CBM sessions ... could not be stopped safely" after hard-killed CBM processes


hansinator85:
### Version

codebase-memory-mcp 0.10.8

### Platform

macOS (Apple Silicon)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

## Title
Stale rendezvous directory (`/tmp/cbm-daemon-<uid>`) permanently blocks `install`/`update` with "active CBM sessions ... could not be stopped safely" after hard-killed CBM processes

## Summary

After several long-lived `codebase-memory-mcp` MCP server processes were terminated non-gracefully (`pkill`, no clean shutdown), every subsequent `codebase-memory-mcp install` and `codebase-memory-mcp update` call failed permanently with:

```
Stopping active CBM sessions and operations for install...
error: active CBM sessions and operations could not be stopped safely; no activation was committed.
error: something is still using CBM. If an editor or agent is running an MCP server, close it and retry; 'codebase-memory-mcp daemon status' lists the holders when a cbm binary is still installed.
```

This happened even though every process-level and file-lock check came back completely clean:

- `ps aux | grep codebase-memory-mcp` → no matches
- `codebase-memory-mcp daemon status` → `daemon: not running`
- `lsof +D ~/.cache/codebase-memory-mcp` → no open handles
- No `*.lock`, `*.pid`, or `*.sock` files under `~/.cache/codebase-memory-mcp`
- No matching `launchctl list` entries / no LaunchAgent plist
- No stray files under `$TMPDIR` (the per-user `/var/folders/.../T/` directory)

The installer's own diagnostics (`daemon status`, the error message itself) only ever point at "an editor or agent running an MCP server" — i.e. a *process-level* explanation — which sent us down the wrong debugging path for a long time. The actual blocker is **stale on-disk coordination state**, not a live process.

## Environment

- OS: macOS (Apple Silicon / arm64)
- Shell: zsh
- Install method: official `install.sh` (`curl -fsSL https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh | bash`), binary at `~/.local/bin/codebase-memory-mcp`
[...]

Issue #1757, opened 2026-08-20
macOS: markerless stale socket/anchor pair permanently blocks daemon startup


silee9019:
### Version

codebase-memory-mcp 0.10.8

### Platform

macOS on Apple Silicon

### Install channel

GitHub release archive

### Binary variant

standard

### What happened, and what did you expect?

When a runtime namespace contains a hard-linked `.sock` and `.anc` pair but no `.sock.identity` or `.sock.pending`, daemon startup remains permanently blocked.

No process holds the socket and connecting to it returns `ECONNREFUSED`, but both `daemon start` and CLI commands fail until the pair is moved manually.

I expected startup to recover this provably stale state safely, or to report the exact blocking artifacts and provide a supported recovery command.

The public command only reports a 30-second startup timeout. The daemon log ends at `stage=namespace_validation`.

### Reproduction

This reproduction uses isolated runtime and cache directories and does not touch the account daemon.

```sh
#!/bin/sh
set -eu

runtime=$(mktemp -d /private/tmp/cbm-markerless-repro.XXXXXX)
cache=$(mktemp -d /private/tmp/cbm-markerless-cache.XXXXXX)
chmod 700 "$runtime" "$cache"

run_cbm() {
  CBM_RUNTIME_DIR="$runtime" \
  CBM_CACHE_DIR="$cache" \
  codebase-memory-mcp "$@"
}

start_output=$(run_cbm daemon start)
printf '%s\n' "$start_output"

pid=$(printf '%s\n' "$start_output" |
  sed -n 's/.*pid \([0-9][0-9]*\)).*/\1/p')
kill -KILL "$pid"

attempt=0
while kill -0 "$pid" 2>/dev/null && [ "$attempt" -lt 100 ]; do
  sleep 0.1
  attempt=$((attempt + 1))
done
if kill -0 "$pid" 2>/dev/null; then
  printf 'daemon did not exit\n' >&2
  exit 1
fi

namespace="$runtime/cbm-daemon-$(id -u)"
identity=$(find "$namespace" -maxdepth 1 -name '*.sock.identity' -print -quit)
socket=${identity%.identity}
anchor=${socket%.sock}.anc

stat -f '%HT inode=%i links=%l %N' "$socket" "$anchor"
mv "$identity" "$runtime/socket.identity.backup"

set +e
run_cbm daemon start
start_rc=$?
set -e
printf 'markerless_start_rc=%s\n' "$start_rc"
[...]

PR #1790, opened 2026-08-21
fix(cli): recover stale rendezvous publication before declaring BUSY (#1760)

umi008:
## What does this PR do?

After CBM processes are hard-killed (`pkill`/`kill -9`), the runtime dir keeps the published socket/anchor/identity files while every file lock is kernel-released. The CLI activation guard probed generation and treated the `ECONNREFUSED` socket as an active daemon — a deliberate fail-closed choice that protects saturated BSD listeners — so `install`/`update` failed forever with "active CBM sessions ... could not be stopped safely" even though nothing was running. Only daemon-start paths ran the stale cleanup, which the CLI could never reach: deadlock by residue.

`cli_activation_production_reserve` now recognizes the provably-stale combination — generation present, lifetime reservation absent, startup lock already held — and runs the existing `cbm_daemon_ipc_stale_generation_cleanup` repair, then re-probes generation before deciding. The cleanup re-checks both conditions and refuses on a genuinely held reservation, so a live daemon (including a saturated BSD listener) is never disturbed and the guard falls back to the original BUSY verdict. POSIX-only; Windows keeps its rendezvous-record semantics unchanged.

Verified with an end-to-end regression test: a child listens and exits without unlink (hard-kill residue), then the real activation guard installs successfully, removes the residue, and records the `completed` phase.

## Checklist

### DIFF
--- a/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1760/.openspec.yaml
+++ b/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1760/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-21
--- a/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1760/design.md
+++ b/openspec/changes/archive/2026-08-21-oss-codebase-memory-mcp-issue-1760/design.md
@@ -1,0 +1,57 @@
+## Context
+
+See proposal.md for motivation. Verified current state (against source):
+
+- `cli_activation_production_reserve` (src/cli/cli.c:527) holds cohort EX locks + the startup lock, then calls `cbm_daemon_ipc_generation_probe_under_startup_lock` (cli.c:571-572). `generation != 0` ⇒ return BUSY (0) or error (cli.c:573-578).
+- The probe's endpoint-probe branch (ipc.c) is deliberately fail-closed: a present-but-ECONNREFUSED socket counts as generation 1 (to protect saturated BSD listeners whose reserved backlog yields ECONNREFUSED while alive).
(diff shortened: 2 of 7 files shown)


DeusData:
Thank you for the detailed stale-residue analysis and for reusing the existing guarded cleanup path. I checked current `main`: `cli_activation_production_reserve()` probes the published generation under the startup lock and can return the existing BUSY result without attempting stale-generation cleanup. That grounds the activation deadlock described here.

There is one factual contribution-policy blocker before code review. This focused fix also adds 131 lines under five new `openspec/` files, including `openspec/config.yaml`; current `main` has no `openspec/` tree, and `CONTRIBUTING.md` requires prior approval for project configuration. Please remove the entire `openspec/` portion from this PR and leave the focused C change plus regression test. That preserves the useful work while keeping one independently reviewable claim. I have labeled the underlying issue as a high-priority stability bug. Splitting enables review but does not pre-decide the merge outcome. Thank you for the focused production repair.

DeusData:
Not a review yet — this one is parked on our side for an unrelated reason, and I did not want you to find that out only after doing more work.

**Please drop the `openspec/` files here too.** As on #1788 and #1789: `openspec/` does not exist on `main`, no other contributor uses it, and adopting a spec framework is a project-level decision rather than something a bug fix should carry in. We have decided not to adopt it, so the ask is to keep `src/` and `tests/` and strip the rest.

That request stands independently of the review, so you can act on it now across all three without waiting on me.

I will come back to the change itself. Thank you for the batch — #1787 in particular is a good catch and I have approved it.

DeusData:
Thank you for this — you were the first to put a fix in front of us for #1760, and your diagnosis (generation present, lifetime reservation absent, startup lock held, so the cleanup that already exists should run before declaring BUSY) was exactly right.

We went with #1894 (now on main as 760c08f453aa), which does the same recovery one layer lower, inside the daemon's IPC stale-generation cleanup, so it also covers the markerless .sock/.anc pair from #1757 and the worker hot-loop in #2015; it also carries the device/inode and two-link revalidation the unlink needs. Since it subsumes this change, I am closing this PR rather than merging both; the credit for getting there first is yours, and the fix ships in v0.10.9. If you see any residue of the wedge on that release, please open a fresh report and ping me. Thanks again.

PR #1799, opened 2026-08-22
fix(foundation): resume spawn backoff after EINTR (#1616)

### LINKED ISSUE #1616: cbm_nanosleep never resumes after EINTR: every timed wait can be silently truncated

rarepops:
## What does this PR do?

Adds `cbm_nanosleep_full()` to resume POSIX sleeps after `EINTR`, and uses it for subprocess spawn backoff so transient signals cannot silently shorten the documented retry budget. Other sleep call sites retain their existing interruptible behavior.

Adds a POSIX regression test that repeatedly interrupts the deterministic spawn backoff with `SIGALRM` and verifies the full cumulative delay is preserved.

Fixes #1616

## Checklist

- [x] Tests pass locally (`subprocess`: 32 passed)
- [x] Lint passes (`clang-format-20`, `cppcheck 2.20.0`)

### DIFF
--- a/src/foundation/compat.c
+++ b/src/foundation/compat.c
@@ -17,6 +17,20 @@
 #include <sys/stat.h>
 #endif
 
+int cbm_nanosleep_full(const struct timespec *req) {
+#ifdef _WIN32
+    return cbm_nanosleep(req, NULL);
+#else
+    struct timespec remaining = *req;
+    while (nanosleep(&remaining, &remaining) != 0) {
+        if (errno != EINTR) {
+            return -1;
+        }
+    }
+    return 0;
+#endif
+}
+
 /* ── strndup (Windows lacks it) ───────────────────────────────── */
 
 #ifdef _WIN32
--- a/src/foundation/compat.h
+++ b/src/foundation/compat.h
@@ -96,6 +96,9 @@ static inline int cbm_nanosleep(const struct timespec *req, struct timespec *rem
 #define cbm_nanosleep nanosleep
 #endif
 
+/* Sleeps for the full requested duration even when POSIX signals interrupt it. */
+int cbm_nanosleep_full(const struct timespec *req);
+
 /* ── gmtime_r (Windows lacks it) ─────────────────────────────── */
 #ifdef _WIN32
 static inline struct tm *cbm_gmtime_r(const time_t *timep, struct tm *result) {
--- a/src/foundation/subprocess.c
+++ b/src/foundation/subprocess.c
@@ -970,7 +970,7 @@ static void cbm_spawn_backoff(int attempt) {
     int shift = attempt < CBM_SPAWN_BACKOFF_MAX_SHIFT ? attempt : CBM_SPAWN_BACKOFF_MAX_SHIFT;
     long ms = (long)CBM_SPAWN_BACKOFF_BASE_MS << shift;
     struct timespec delay = {ms / 1000L, (ms % 1000L) * 1000L * 1000L};
-    (void)cbm_nanosleep(&delay, NULL);
+    (void)cbm_nanosleep_full(&delay);
 }
 
 /* fork() fails with EAGAIN under the same pressure posix_spawn does, and the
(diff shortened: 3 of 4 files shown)


DeusData:
Thank you for the focused EINTR fix and deterministic regression test. I checked current `main`: POSIX maps `cbm_nanosleep` directly to `nanosleep`, and the subprocess spawn backoff discards its return value while passing no remainder. That confirms a signal can shorten the documented retry budget.

I have labeled this as a high-priority stability bug and routed it for review. CI is green. Our review queue is full, so detailed review may take a little time, but the change is now classified and visible. Thank you for limiting the new full-sleep behavior to the spawn-backoff path.

DeusData:
Merged as `14249349`. That is your third landing today — thank you.

Two things about how this one got verified are worth telling you, because the usual route was not available.

Your CI was 35/35, but the branch was 205 commits behind `main`, so that green no longer described the code that would actually land. Normally I would rebase and re-run — except our Actions pool has been servicing roughly one job at a time against a 35-deep queue all day, so a rebase would have cost you hours and everyone else a slot.

So instead I sized the risk with evidence rather than guessing: exactly **one** commit on `main` had touched any of your four files since your base, and it touched only `tests/test_subprocess.c`, adding an unrelated Windows-only Job Object test. `compat.c`, `compat.h` and `subprocess.c` had no changes at all.

Then I built the **actual merge result** locally — current `main` plus your branch — and ran the affected suites. Clean build, `281 passed, 5 skipped, 0 failed`. When #1779 landed and invalidated that, I redid it against the new `main` and included the `mcp` suite too, since that merge touched `mcp.c`. Still clean. That is stronger evidence than a fresh remote green would have been for the platforms it covers, and it cost the queue nothing.

On the fix itself: the detail that made it easy to accept is that the old call passed `NULL` for the remainder, so the unslept time was **discarded** rather than merely shortened. With the backoff reaching 2.56 s at shift 8, a signal could collapse the wait to almost nothing — tightening the retry loop under precisely the pressure it exists to ride out. Confining the change to the one call site and leaving other sleeps interruptible is what kept the blast radius honest.

The fixing SHA is recorded on #1616.

Issue #1558, opened 2026-08-12
Installer/updater scope and safety: no per-client selector, destructive update step before an interactive prompt, hidden ui_enabled, brittle legacy-marker cleanup


PsyTech-Rob:
**Environment:** v0.10.2, Windows 11. Four related findings from one install/update cycle.

## 1. `install` has no per-client selector

`install` writes MCP config, agents, skills, instructions and hooks for **every** detected client (Claude Code, Codex, OpenCode, Cursor). We only wanted Claude + Codex; the OpenCode/Cursor integrations had to be reverted by hand — and the next `install`/`update` silently recreates them. A `--clients=claude,codex` style selector (or per-client confirmation in interactive mode) would fix this.

## 2. `update -y` ran a destructive step before a blocking interactive prompt

On 0.9.0 → 0.10.2, `update -y` auto-confirmed deletion of other projects' index files (two `.db` indexes, ~59 MB, lost) and **then** aborted on the interactive standard/ui variant chooser ("failed to read input" under a non-interactive shell). Destructive cleanup should be sequenced after the last prompt that can abort the run. #1538 removes the chooser, which fixes this particular abort but not the ordering principle; #1544 is adjacent.

## 3. `ui_enabled` is invisible to `config list` / `config set`

The only way to disable the UI HTTP listener (default `127.0.0.1:9749`) is editing `config.json` (`"ui_enabled": false`) by hand — the key is not shown by `config list`, not settable by `config set`, and the server binary accepts `--ui=false` without effect. It took two debugging sessions to find the knob; meanwhile the daemon exposes a loopback HTTP surface that looked impossible to turn off.

## 4. Legacy hook cleanup aborts on an orphan marker

An earlier duplicated install left a Codex `config.toml` with a `# <<< codebase-memory-mcp SessionStart <<<` closing marker without its `# >>>` opener. `install` then fails that client with `error: agent_config ... op=legacy_hook_cleanup` and stops the activation ("one or more agent configurations failed"). Since the malformed state was produced by a previous installer run, cleanup could tolerate/strip orphan markers instead of aborting (self-heal).

DeusData:
@PsyTech-Rob — apologies for the silence on this one; four distinct, well-isolated findings from a single install cycle deserved a reply much sooner. All four are legitimate. Taking them in the order I would fix them rather than the order you listed them, because two are considerably more serious than the others.

[...]

PR #1800, opened 2026-08-22
fix(help): add --clients to CLI help

kilianpaquier:
## What does this PR do?

Hello 👋

Just a small PR to add `--clients` to CLI's help.

## Checklist

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -902,6 +902,8 @@ static void print_help(void) {
     printf("                                      Run one tool locally, then exit\n");
     printf("  codebase-memory-mcp install [-y|-n] [--force] [--dry-run] "
            "[--dir=<path>] [--skip-config]\n");
+    printf("                                      [--clients=<tokens>]  Run "
+           "'install --clients' to list tokens\n");
     printf("  codebase-memory-mcp uninstall [-y|-n] [--dry-run]\n");
     printf("  codebase-memory-mcp update [-y|-n]\n");
     printf("  codebase-memory-mcp config <list|get|set|reset>\n");


DeusData:
Thank you for the concise help fix. I checked current `main`: the install parser accepts `--clients`, but the install usage line in `src/main.c` omits it. The documentation gap is therefore confirmed.

I have labeled this as a normal-priority documentation and UX correction and queued it for review. CI is green. Our review queue is full, so detailed review may take a little time. Thank you for keeping this to the single help-surface change.

DeusData:
👋 Approved — thank you, this is a real gap and not just tidying.

`--clients=` has been parsed since #1558 (`cli.c:10198`) but `print_help` never mentioned it, so the only way to discover the flag was to read the source or already know it existed. For a flag that decides which agent clients get configured on install, that is a genuine usability defect.

Two details I checked and liked:

- **Pointing at `install --clients` to list the tokens**, rather than trying to enumerate 40-odd client names in the help text. Bare `--clients` really does print the list (`cli.c:10205` handles it alongside `--clients=help`/`list`), so the cross-reference is accurate and stays accurate as clients are added.
- **The wording matches the flag's actual form**, `--clients=<tokens>` with the `=`. That is the form the parser accepts, so nobody follows the help into an error.

I verified the surrounding help block on current `main` is byte-identical to your patch context, so it still applies cleanly and renders as intended, despite the branch being some way behind.

You left the test and lint boxes unticked — entirely reasonable for two `printf` lines, and CI is 34/34 green on it, so nothing needed from you there.

I will merge this once our Actions queue recovers; there is currently one job executing repo-wide against 24 queued, so merges are moving slowly today for reasons unrelated to your PR.

DeusData:
Merged as `c895513f`. Thank you 👋 — small PR, real gap: `--clients=` had been parsed since #1558 but never appeared in `print_help`, so the only way to discover the flag was to read the source.

Pointing at `install --clients` to list the tokens, rather than enumerating client names inline, is what makes it stay correct as clients are added — and as it happens #1802 landed alongside this one and adds registry stable IDs to exactly that listing, so the cross-reference is now complete in both directions.

PR #1802, opened 2026-08-22
fix(cli): apply client selector to registry integrations (#1798)

### LINKED ISSUE #1798: install --clients=<token> accepted but inert in the Windows binary: all detected clients configured, none force-included
### Version

codebase-memory-mcp 0.10.8

### Platform

Windows (x64) — Windows 11 build 26200

### Install channel

### Binary variant

standard

`install --clients=<token>` on the shipped Windows binary is accepted but inert: it neither restricts configuration to the named clients nor force-includes them when undetected. Three experiments (all `--dry-run`, all with the `=` form):

1. **No restriction.** On a machine where Cursor, Qoder CLI, and Rovo Dev CLI are detected, `install --clients=cursor --dry-run` shows full configuration sections (mcp + skill + 3 agents, and for Qoder also hooks) for Qoder CLI and Rovo Dev CLI — not just Cursor.
2. **No inclusion.** With `HOME`/`USERPROFILE` redirected to an empty directory, the same command reports `Detected agents: (none)` and configures nothing — the flag does not add `cursor`.
3. **Parsing works; scope doesn't.** `--clients=x` correctly errors with `unknown client: x` and lists valid tokens, so the flag is parsed. Scope is decided purely by detection: an otherwise-empty home containing only a `.cursor/mcp.json` marker yields `Detected agents: Cursor` and a complete Cursor-only configuration — with or without the flag.

Expected, per the v0.10.4 release notes: "`--clients=claude,codex` now restricts it."

Cross-reference #1753: that report says the binary's own installer "honors the flag correctly" and scopes the bug to the `install.sh` wrapper. On Windows v0.10.8 the binary path is affected too — every experiment above used the binary's `install` subcommand directly (the `=` form). If the flag works on macOS/Linux, this may be Windows-specific argument handling in the binary.

### Reproduction

Windows, v0.10.8 binary at `%LOCALAPPDATA%\Programs\codebase-memory-mcp\codebase-memory-mcp.exe`.

Experiment 1 — flag does not restrict (machine with several MCP clients installed):

```
codebase-memory-mcp install --clients=cursor --dry-run
```

Output (excerpt):

```
Detected agents: Cursor Qoder CLI Rovo Dev CLI

Cursor:
  mcp: <user>/.cursor/mcp.json
  skill: <user>/.cursor/skills/codebase-memory/SKILL.md (1 installed)
  agent: x3
Qoder CLI:
  mcp: <user>/.qoder/settings.json
  skill: <user>/.qoder/skills/codebase-memory/SKILL.md (1 installed)
  agent: x3
  hooks: <user>/.qoder/settings.json (SessionStart + SubagentStart + PostToolUse Read)
Rovo Dev CLI:
  mcp: <user>/.rovodev/mcp.json
  instructions: <user>/.rovodev/AGENTS.md
  skill: <user>/.rovodev/skills/codebase-memory/SKILL.md (1 installed)
  agent: x3
```

Expected: only the Cursor section.

Experiment 2 — flag does not include (empty sandbox home; Git Bash syntax, binary honors both vars):

```
mkdir <sandbox>
HOME=<sandbox> USERPROFILE='C:\<sandbox>' codebase-memory-mcp install --clients=cursor --dry-run
```

Output (excerpt): `Detected agents: (none)` — and no Cursor section, despite `--clients=cursor`.

Expected: Cursor configured via the flag.

Experiment 3 — detection alone drives scope: sandbox containing only `.cursor/mcp.json` with `{"mcpServers": { }}` → `Detected agents: Cursor`, complete Cursor-only configuration, flag irrelevant.

### Logs

stderr is clean; the outputs above are the full signal. One wording nit observed along the way: in `--dry-run` mode the installer prints `Added <dir> to the current-user PATH` directly above the footer `(dry-run — no files were modified)`, which reads as a completed action; "Would add" would match the rest of the dry-run phrasing.

### Project scale

N/A — installer flag behavior, not indexing.

### Confirmations

- [x] I searched existing issues and this is not a duplicate — #1753 covers the `install.sh` wrapper only and asserts the binary honors the flag, which contradicts what is observed here on Windows.

rarepops:
## What does this PR do?

Fixes #1798.

`--clients` previously filtered only the legacy client table, so registry-backed clients such as Qoder and Rovo Dev were still detected and configured. This change makes registry stable IDs part of the selector vocabulary and applies the same restriction to registry detection output and installation.

The selector remains restrictive: it configures only selected clients that are detected, and does not force-enable undetected clients.

Validation covered the sanitizer build and both selector regressions. The full CLI suite passes with 283 tests. clang-format 20.1.8 and `git diff --check` pass; focused cppcheck reports only pre-existing style findings in test helpers and older test code.

## Checklist

- [x] Tests pass locally (`cli`: 283 passed)
- [x] Lint passes (`make -f Makefile.cbm lint-ci`) - upstream `lint / lint` and `lint / lint-mem` checks pass

### DIFF
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -7376,6 +7376,12 @@ static const char *detect_arch(void) {
 
 /* ── Agent config install/refresh (shared by install + update) ── */
 
+/* Set by `install --clients=...` after validation and consumed by both the
+ * legacy detector and the registry-backed client path (#1558, #1798). */
+static const char *g_client_selection = NULL;
+static bool cli_clients_apply_selection(const char *spec, cbm_detected_agents_t *detected);
+static bool cli_clients_selects_registry_client(const char *spec, cbm_agent_client_id_t client_id);
+static void cli_clients_print_list(FILE *out);
 static void print_detected_registry_agents(const char *home, bool *any);
 
 /* Print detected agent names on a single line. */
@@ -8169,7 +8175,10 @@ static void print_detected_registry_agents(const char *home, bool *any) {
     cbm_init_agent_registry_context(home, &registry);
     for (size_t index = 0U; index < cbm_agent_client_count(); index++) {
         const cbm_agent_client_profile_t *profile = cbm_agent_client_at(index);
-        if (profile && cbm_agent_client_detect(profile->id, &registry.options)) {
+        if (profile &&
+            (!g_client_selection ||
+             cli_clients_selects_registry_client(g_client_selection, profile->id)) &&
+            cbm_agent_client_detect(profile->id, &registry.options)) {
             printf(" %s", profile->display_name);
             *any = true;
         }
@@ -8525,7 +8534,10 @@ static void install_agent_client_registry(const char *home, const char *binary_p
     cbm_init_agent_registry_context(home, &registry);
     for (size_t index = 0U; index < cbm_agent_client_count(); index++) {
         const cbm_agent_client_profile_t *profile = cbm_agent_client_at(index);
-        if (!profile || !cbm_agent_client_detect(profile->id, &registry.options)) {
+        if (!profile ||
+            (g_client_selection &&
+             !cli_clients_selects_registry_client(g_client_selection, profile->id)) ||
+            !cbm_agent_client_detect(profile->id, &registry.options)) {
             continue;
         }
         if (!g_install_plan) {
@@ -9365,13 +9377,6 @@ static void install_additional_agent_configs(const cbm_detected_agents_t *agents
     }
 }
 
-/* #1558: set by `install --clients=...` after validation, consumed at the one
- * place detection happens. Validation runs in cbm_cmd_install so an unknown
- * token fails before anything is written, not midway through configuring. */
-static const char *g_client_selection = NULL;
-static bool cli_clients_apply_selection(const char *spec, cbm_detected_agents_t *detected);
-static void cli_clients_print_list(FILE *out);
-
 int cbm_install_agent_configs(const char *home, const char *binary_path, bool force, bool dry_run) {
     g_agent_install_errors = 0;
     cbm_detected_agents_t agents = cbm_detect_agents(home);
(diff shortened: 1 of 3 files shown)


DeusData:
Thank you for the focused selector fix and regression coverage. I checked current `main`: `--clients` is applied to the legacy `cbm_detected_agents_t` set, after which `install_agent_client_registry()` is called separately without that selected-client restriction. That grounds the registry-backed client leak described here.

I have labeled this as a high-priority editor-integration bug and queued it for review. CI is green. Our review queue is full, so detailed review may take a little time, but the PR is now classified and routed. Thank you for preserving the selector's restrictive semantics rather than force-enabling undetected clients.

DeusData:
Approved on merit. I verified the gap on `main` rather than taking it: `cli_clients_apply_selection` is applied to the legacy `probe` at `cli.c:10280`, while `print_detected_registry_agents` (`:8283`, called from `:7510`) iterates registry profiles and calls `cbm_agent_client_detect(profile->id, …)` with no reference to any selection at all. So registry-backed clients genuinely bypassed `--clients` entirely.

That matters more than a missing filter usually would, because it fails in the **configuring** direction. A user writing `--clients=claude` is stating what they want touched; silently configuring Qoder and Rovo Dev anyway is the same class of harm as #1753, where a flag was accepted and had no effect. A selector that under-selects is a bug; one that over-configures is a surprise on someone's machine.

Two things I checked and want to credit:

**The polarity is right.** Both new guards read `!g_client_selection || cli_clients_selects_registry_client(...)` — no selection means no filtering, so the default path is byte-identical to today and only an explicit `--clients` restricts. The inverse form in the skip path is consistent with it. That is the shape that makes this safe to take without auditing every caller.

**You kept the restrictive semantics explicit**: selected *and* detected. Not force-enabling an undetected client is the right call, and stating it in the description saved me from having to derive it from the guards.

**And a nice interaction you may not have noticed:** adding registry `stable_id`s to `cli_clients_print_list` makes `install --clients` list the registry clients too. #1800 is currently open and adds `--clients=<tokens>` to the CLI help, pointing users at exactly that listing to discover valid tokens. Without your change that listing would have been incomplete for precisely the clients this PR fixes. The two are complementary; I have approved both.

## Before merge

You are 205 commits behind `main`, so the 35/35 green here is stale. I am deliberately not rebasing: our Actions pool is servicing roughly one job at a time against 21 queued PR runs, and every rebase adds a full matrix to that. I will refresh it once the pool recovers.

Nothing to change in the code. Thank you — this is the third careful PR of yours I have read today, and closing the `--clients` gap on the registry side completes work that #1753 and #1800 only half-covered.

DeusData:
Merged as `4d8ac9e1`. That is your fourth landing today — #1779, #1799, this, and the delta on #1773. Thank you.

The verification route is worth a note, since your branch was 205 commits behind and the green was stale. Our Actions pool has been servicing roughly one job at a time against a 38-deep queue, so rebasing would have cost you hours and everyone else a slot. Instead I built the **actual merge result** locally — current `main` plus this branch plus #1800, which is file-disjoint and complementary to it — and ran the `cli` and `agent_clients` suites: clean build, **329 passed**. Verifying the combined tree meant one check covered the state after both merges rather than two.

On the fix: what made it easy to accept was the polarity. `!g_client_selection || cli_clients_selects_registry_client(...)` leaves the no-selector path byte-identical, so only an explicit `--clients` can restrict anything — that is what let me take it without auditing every caller. And the fixing SHA is recorded on #1798.

PR #1804, opened 2026-08-22
test.sh: refuse the cli suite against a real HOME

lgerard42:
## Why

The `cli` suite installs and uninstalls agent configurations **for real** — it calls `cbm_cmd_install` and `cbm_cmd_uninstall`, not mocks, across 53 call sites.

`HOME` isolation is per-test rather than global: `tests/test_cli.c` calls `cbm_setenv("HOME", tmpdir, 1)` 58 separate times, and `scripts/test.sh` sets `HOME` zero times. So the suite acts on whatever `HOME` it inherits, and the safety of a run depends on every individual test remembering — including the ones that fail partway.

The observable consequence is that results outside isolation are not trustworthy. Running the suite against a real `$HOME` on my machine gave different failure counts on consecutive runs, and the tests that moved were exactly the agent-config ones — OpenClaw, VS Code, Gemini, Augment, CodeBuddy/Pochi. The same tree under `HOME=$(mktemp -d)` was stable at the baseline every time.

I am not claiming a specific test damages a real config — I tried to reproduce that and could not, twice, against increasingly faithful copies of the home directory in question. The argument here is narrower and I think uncontroversial: a suite that installs and uninstalls agent configurations should not be pointed at a developer's own one by default, and its results there are unreliable regardless.

## What this does

`scripts/test.sh` exits 2 when the `cli` suite is selected and `$HOME` contains `.claude` or `.claude.json`, printing the isolated command to use. Default mode (no `--suites`) is covered too, since it runs every suite. `CBM_ALLOW_REAL_HOME=1` overrides for anyone who wants the current behaviour.

CI is unaffected in the normal case: a runner whose `$HOME` has no `.claude` never trips it, and the override is available if yours does.

Verified in this tree: refuses `--suites cli` on a real HOME with exit 2, permits `--suites discover` there, and permits `--suites cli` under an isolated HOME.

Happy to gate it differently — a warning instead of a refusal, or a different env var name — if you would rather.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -121,6 +121,35 @@ if [ "$TSAN" -eq 1 ] && [ -n "$SUITES" ]; then
     echo "test.sh: --tsan and --suites are separate modes (the TSan leg has its own suite set). Please consult --help." >&2
     exit 2
 fi
+
+# The cli suite drives the real install and uninstall paths. HOME isolation is
+# per-test (tests/test_cli.c sets it 58 times; this script sets it none), so the
+# suite installs and uninstalls agent configurations against whatever HOME it
+# inherits. Its results are also unreliable outside isolation: against a real
+# HOME it reports different failure counts run to run, all in the agent-config
+# install/uninstall tests, where an isolated HOME is stable. Refuse rather than
+# touch the machine's agent configuration. Default mode (no --suites) runs every
+# suite, so it is covered too.
+cbm_cli_suite_selected=0
+if [ -z "$SUITES" ]; then
+    cbm_cli_suite_selected=1
+else
+    for cbm_suite in $SUITES; do
+        if [ "$cbm_suite" = "cli" ]; then
+            cbm_cli_suite_selected=1
+        fi
+    done
+fi
+if [ "$cbm_cli_suite_selected" -eq 1 ] && [ "${CBM_ALLOW_REAL_HOME:-0}" != "1" ] &&
+   { [ -e "$HOME/.claude.json" ] || [ -d "$HOME/.claude" ]; }; then
+    echo "test.sh: refusing to run the cli suite against a real HOME ($HOME)." >&2
+    echo "  That suite installs and uninstalls agent configurations for real, against" >&2
+    echo "  whatever HOME it inherits, and its results are unreliable outside isolation." >&2
+    echo "  Run it isolated instead:" >&2
+    echo "    HOME=\"\$(mktemp -d)\" CCACHE_DIR=\"$HOME/.ccache\" scripts/test.sh $*" >&2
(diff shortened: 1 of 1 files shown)


DeusData:
Thank you for the careful framing here, especially for separating a confirmed isolation problem from damage you could not reproduce. I checked current `main`: `scripts/test.sh` selects and runs the `cli` suite without a runner-level `HOME` isolation guard, while the suite exercises real install and uninstall paths. The safety and result-reliability premise is therefore grounded.

I have labeled this as a high-priority security and maintainer-safety item. The exact policy still needs a maintainer direction decision before detailed review: hard refusal with an escape hatch versus making isolation unconditional at the runner boundary. The current DCO check is also red, so the commit will need a valid `Signed-off-by` line. Our review queue is full, but this is now classified and routed. Thank you for raising it without overstating the evidence.

DeusData:
Following up on the review note from the 24th. Two separate things are holding this, and only one of them is yours:

**Yours — DCO.** That is the only red check; everything else is green. It is a single commit (`c91ae768`), so it should be one round trip: amend it with a sign-off (`git commit --amend -s --no-edit`) and force-push with lease.

**Ours — the policy call.** As noted on the 24th, the shape of the fix is a maintainer direction decision: hard refusal with an escape hatch (what you implemented) versus making `HOME` isolation unconditional at the runner boundary. That is genuinely open on our side and it is queued for a decision, so please do not read the delay as doubt about the report.

The underlying finding stands on its own either way, and the part worth calling out is that you separated a *confirmed* isolation problem from damage you could not reproduce, instead of asserting the stronger claim. That is the right way to file this, and it is what made it easy to classify.

This will not be closed for inactivity.

DeusData:
The direction decision is made, so you are unblocked on the design question.

**Isolate by default, and keep the escape hatch.** Concretely, that is your change plus one step further:

1. `scripts/test.sh` sets an isolated `HOME` for the `cli` suite unconditionally, so the suite cannot act on a real one regardless of who invokes it or how.
2. `CBM_ALLOW_REAL_HOME=1` is retained for the deliberate case, and documented where a reader will find it rather than discovered by accident.

The reasoning for going past a refusal: your report showed the suite producing *different failure counts on consecutive runs* against a real `HOME`, which means the current per-test isolation is genuinely incomplete rather than merely risky. A refusal warns the person who already knows; isolating removes the failure mode for the person who does not. Keeping the hatch preserves the case you were careful not to break.

Two smaller notes for when you pick it back up:

- The commit still needs a `Signed-off-by`; DCO is the only red check.
- Worth stating in the PR description which of the two behaviours a maintainer gets by default, since that is the part future readers will need.

Thank you for filing this the way you did — separating the confirmed isolation problem from the damage you could not reproduce is what made it possible to classify quickly, and it is the reason the decision went the further of the two ways rather than the cautious one.

Issue #1384, opened 2026-07-31
scripts/test.sh hangs indefinitely at the venue-parity contract when ImageMagick is installed (bash runs generate-sbom.py, 'import' becomes ImageMagick)


OmerZeyveli:
**Version:** `main` (`d698db8e`) · **Platform:** Ubuntu 24.04, x86_64, gcc 13.3

## What happens

`scripts/test.sh` hangs forever at Step 0j on any machine that has ImageMagick installed. No error, no timeout — the process just sleeps indefinitely, after printing:

```
=== Step 0j: venue parity contract (one harness, every venue) ===
venue-parity contract OK (22 workflows marker-checked, 10 venue workflows whitelist-walked, 4 deliberate step exemptions)
```

## Root cause

`tests/test_venue_parity_contract.sh` probes every entry point with `bash`:

```sh
for entry in $HELP_ENTRIES; do
    out=$(cd "$ROOT" && bash "$entry" --help 2>&1) && rc=0 || rc=$?
```

`HELP_ENTRIES` contains one Python file, `scripts/ci/generate-sbom.py`. Running it under `bash` makes the shell interpret Python source as shell commands, so the module imports become *commands*:

```
$ ps -o pid,cmd -p 2153369
2153369 import datetime
```

On Linux `import` is **ImageMagick's screen-capture tool**, which blocks waiting for the user to click a window:

```
$ command -v import
/usr/bin/import
$ import --version
Version: ImageMagick 6.9.12-98 Q16 x86_64
```

So the probe waits on an X11 window selection that will never come. `</dev/null` does not help — `import` waits on the display, not stdin.

CI almost certainly does not hit this because container images generally lack ImageMagick. It reproduces reliably on a desktop Linux developer machine, which is exactly where CONTRIBUTING tells contributors to run `scripts/test.sh` before submitting.

## Minimal reproduction

```sh
bash scripts/ci/generate-sbom.py --help     # hangs forever with ImageMagick installed
```

## A second, quieter consequence

Because the probe has always run this file under `bash`, **the contract has never actually verified `generate-sbom.py`'s `--help` behaviour** — it has only ever checked bash's misinterpretation of it. Under the correct interpreter the script does satisfy the contract:

```sh
$ python3 scripts/ci/generate-sbom.py --help
Generate the release SPDX SBOM (sbom.json).
...
$ echo $?
0
```

(`generate-sbom.py:22` handles `-h`/`--help` and exits 0.)

[...]

PR #1810, opened 2026-08-23
feat(lang): TwinCAT / CODESYS (IEC 61131-3 Structured Text) support

### LINKED ISSUE #1805: feat: TwinCAT / CODESYS (IEC 61131-3 Structured Text) language support
## Summary

Industrial automation codebases built with Beckhoff TwinCAT 3 (and CODESYS-family IDEs) are currently invisible to codebase-memory-mcp: PLC source files (`.TcPOU`, `.TcDUT`, `.TcGVL`, `.TcIO`) and project files (`.plcproj`, `.tsproj`) match nothing in the extension table and are silently skipped. This proposal adds first-class IEC 61131-3 Structured Text support, following the two existing patterns the codebase already sanctions: the ObjectScript Studio Export XML transform-only pipeline and the Infra-Pass pattern (CONTRIBUTING.md "Adding or Fixing Language Support").

Related umbrella issue: #592 (Parsing & extraction quality — language/format coverage gaps).

## Design (implementation ready, tested locally)

**New languages:**
1. `CBM_LANG_IEC_ST` — real grammar language for bare `.st`/`.iecst` sources. Grammar: [HeytalePazguato/tree-sitter-iec61131-3-st](https://github.com/HeytalePazguato/tree-sitter-iec61131-3-st) @ `00e24f50f8de` (v0.1.2) — MIT, **ABI 15**, standard-first with full OOP (METHOD/PROPERTY/EXTENDS/IMPLEMENTS), case-insensitive keywords. Vetted against realistic TwinCAT ST: nested METHOD-in-FUNCTION_BLOCK parses with 0 errors; all declaration nodes expose `name` fields. `vendor-maintained` MANIFEST classification (niche vendor language, like objectscript_*).
2. `CBM_LANG_TWINCAT` — transform-only XML container (mirrors `CBM_LANG_OBJECTSCRIPT_EXPORT`): `.TcPOU/.TcDUT/.TcGVL/.TcIO` are XML wrappers carrying ST in `<Declaration>`/`<Implementation><ST>` CDATA. New `internal/cbm/twincat_xml.c` transcoder modeled on `iris_export_xml.c`, re-extracted as `CBM_LANG_IEC_ST`, aggregated via the pass_definitions.c multi-symbol mechanism (both sequential and parallel dispatch).
3. `CBM_LANG_PLCOPEN_XML` — transform-only, content-sniffed (`plcopen.org/xml` namespace in `.xml`), for CODESYS PLCopen exports. Same transcoder family.

**Project files:** `.plcproj`/`.tsproj` handled pass_pkgmap-style: one `Package` node per `.plcproj` (library refs → `DEPENDS_ON`, `Compile Include=` members → `CONTAINS_FILE`), `.tsproj` → `DEPENDS_ON` its PLC projects; both registered as incremental control files.

**Node mapping:** FUNCTION_BLOCK/PROGRAM → Class, FUNCTION → Function, METHOD/ACTION/PROPERTY accessors → Method (parent_class + DEFINES_METHOD), INTERFACE → Interface, TYPE → Type, GVL entries → Variable, EXTENDS → INHERITS, IMPLEMENTS → IMPLEMENTS.

**All hard gates handled:** vendored-checksums.txt, MANIFEST.md provenance row, repro_language_registry.c ledger + partition counts, call-argument matrix B row, grammar_cases/label-goldens/lang_ext test layers.

## Motivation

PLC/industrial-automation is a sizeable underserved niche: TwinCAT projects are Git-versioned XML+ST monorepos that current code-intelligence tooling ignores entirely. Graph queries (FB → method call chains, GVL usage tracing, library dependency mapping) are exactly what large machine-control codebases lack today.

I have the implementation largely complete on a fork (branch `feat/twincat-iec61131-st`) and can open the PR (DCO-signed, phased commits) as soon as maintainers confirm the direction.

ysfsmet:
Closes #1805.

Industrial-automation repositories built with Beckhoff TwinCAT 3 (and CODESYS-family IDEs) were entirely invisible to the indexer: PLC sources (`.TcPOU`, `.TcDUT`, `.TcGVL`, `.TcIO`) and project files (`.plcproj`, `.tsproj`) matched nothing in the extension table and were silently skipped. This adds first-class IEC 61131-3 Structured Text support, following the patterns the codebase already sanctions.

Everything below was validated against a real 130-file / 66-POU TwinCAT solution, not only fixtures — that is where most of the interesting defects came from.

## What lands

**Three new languages.**

- `CBM_LANG_IEC_ST` — the grammar language, for plain `.st`/`.iecst` sources.
- `CBM_LANG_TWINCAT` — transform-only container for `.TcPOU`/`.TcDUT`/`.TcGVL`/`.TcIO`, mirroring `CBM_LANG_OBJECTSCRIPT_EXPORT`: `internal/cbm/twincat_xml.c` recomposes the `<Declaration>`/`<Implementation><ST>` CDATA into textual ST, newline-padded so definition lines still point into the physical file, and the generated units are re-extracted as `CBM_LANG_IEC_ST`.
- `CBM_LANG_PLCOPEN_XML` — transform-only container for CODESYS/PLCopen TC6 exports, detected by a content sniff on the `plcopen.org/xml` namespace. PLCopen declares variables as structured XML rather than CDATA, so `internal/cbm/plcopen_xml.c` *synthesizes* the declaration text from `<inputVars>`/`<outputVars>`/`<variable>`.

**A dedicated project-file pass** (`src/pipeline/pass_tcproj.c`): each `.plcproj` becomes a `Package` node with `DEPENDS_ON` edges to its library references (`Tc2_Standard`, `Tc3_Module`, …) and `CONTAINS_FILE` edges to its members; a `.tsproj` gets `DEPENDS_ON` to the PLC projects it references and `CONFIGURES` to their `.xti` device descriptions. Both are registered as incremental control files.

**Graph mapping.** `FUNCTION_BLOCK`/`PROGRAM` → `Class`; `FUNCTION` → `Function`; `METHOD`/`PROPERTY` → `Method` with `DEFINES_METHOD`; `INTERFACE` → `Interface`; DUT `TYPE` → `Type`; GVL entries → `Variable`; `EXTENDS` → `INHERITS`; `IMPLEMENTS` → `IMPLEMENTS`.

## The grammar is a fork, and that was necessary

The upstream grammar ([HeytalePazguato/tree-sitter-iec61131-3-st](https://github.com/HeytalePazguato/tree-sitter-iec61131-3-st), MIT, ABI 15) is standard-only, and standard-only does not parse production TwinCAT: **49 of 130 files parsed partially**, and every partial parse cost symbols — whole function blocks reached the graph without their methods.

It is forked into `tools/tree-sitter-iec-st/` (grammar.js plus generated parser, same layout as `tools/tree-sitter-form`). Each rule was derived from a real failing construct and verified against the same corpus: modifiers *after* the POU keyword (`FUNCTION_BLOCK PUBLIC FINAL FB_X`), wildcard located addresses (`AT %I*`, spaced or glued, on VAR entries and DUT fields), bit-in-word access (`wError.0 := TRUE`), a base type after an enumerator list (`(Idle, Run) UINT;`), `REFERENCE TO` beside `REF_TO`, function-block instance argument lists (`fb : FB_T()`, `ByteBuffer(ADR(x), SIZEOF(x))`), the terminator TwinCAT omits after a block-shaped `TYPE`, attribute pragmas above members, `__TRY/__CATCH/__FINALLY/__ENDTRY`, `;`-terminated interface prototypes, and tolerance for a stray `;`.

**Result on the same solution: 49 partially-parsed files → 8, and 98 more methods reach the graph.** ABI stays 15; generation is reproducible (`tree-sitter generate`, CLI 0.26.x).

## Defects found and fixed along the way

Three of these are worth calling out because they were silent — no error, no failing test, just wrong or missing data:

- **POU names became `PUBLIC`.** The grammar bound the access modifier to the declaration's `name` field and dropped the real identifier into an ERROR node: 13 function blocks named `PUBLIC` on the real solution. Recovered structurally (header line only, so a body-level parse error can never rename its POU), which serves all three mis-bound POU kinds without needing the source buffer.
- **Project files minted one `Class` per XML element.** Mapping `.plcproj`/`.tsproj` to `CBM_LANG_XML` so they would get File nodes also ran the XML grammar over their markup: 120 nodes named `Project`, `PropertyGroup`, `Name`, `BitSize` buried the 64 genuine POU classes. The definition passes now skip grammar extraction for them — in **both** the sequential and parallel dispatch.
- **A stale grammar object linked silently.** `grammar_*.c` is a one-line wrapper that `#include`s its vendored parser, and make cannot see through an include. Harmless for grammars that never change; for a fork it meant the rebuilt binary kept the old grammar while the source tree showed the new one. `Makefile.cbm` now names the vendored parser as an explicit prerequisite (prod/test/tsan).

Also fixed: `pass_semantic_edges` passed NULL to `qsort` when a project has `Class`/`Method` nodes but no `Function` node — formal UB that UBSan flags, and a TwinCAT solution is exactly that shape.

## Testing

Every registry gate is updated: the capability ledger and its partition counts, the call-argument matrix, the call-node manifest, label goldens, `vendored-checksums.txt`, and the MANIFEST provenance record (fork noted, so a future re-vendor re-applies the dialect rules instead of reverting them).

New tests cover the transcoders (multi-POU, declaration synthesis, graphical-body tolerance, BOM, UTF-16 rejection, entity decoding, prefix near-misses), the access-modifier recovery, the project-markup exclusion, and an end-to-end `pipeline_twincat_project_graph` asserting the Package/DEPENDS_ON/CONTAINS_FILE shape plus a transcoded POU's `DEFINES_METHOD` chain.

All suites pass; `make -f Makefile.cbm cbm` links.

Two notes for reviewers:

1. `tests/test_venue_parity_contract.sh` hangs on any machine with ImageMagick installed — it probes `scripts/ci/generate-sbom.py` with `bash`, so `import datetime` runs ImageMagick's screen grabber. That is #1384, not this branch; the suites here were run with `DISPLAY` unset to get past it.
2. On this machine `tests/test_cli.c:1748` (`cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index`) fails identically on pristine `main` — verified in a worktree at the merge base. Not introduced here.

## Known limitations (documented in README)

Cross-file GVL/DUT references resolve by name rather than declared type; a function-block instance call resolves on the instance identifier rather than the block's type; inside a TwinCAT container the POU's own body line numbers can land after its methods (declaration lines are exact); PLCopen array/pointer-typed variables are dropped rather than approximated, and resource-level `<globalVars>` outside a POU interface are not transcoded. Eight of the 130 files still parse partially — the remaining constructs are listed in the fork's notes.

https://claude.ai/code/session_011BDg22SxMEYCXxGQicWgAc


adfjadfj16-a11y (dismissed):

DeusData:
Thank you for this — it is a serious piece of work, and validating against a real 130-file / 66-POU TwinCAT solution rather than only fixtures is exactly the evidence that makes a language contribution credible. Several details show you read how this codebase already does things: mirroring `CBM_LANG_OBJECTSCRIPT_EXPORT` for the transform-only container, newline-padding the recomposed ST so definition lines still point into the physical file, and the explicit `Makefile.cbm` dependency so a regenerated grammar cannot silently relink a stale object.

**The maintainer decision is that this needs to be split: one PR per language, each reviewed on its own merit.**

I want to be straight with you about what that does and does not mean: **splitting is not a commitment to merge any of them.** Each will be judged separately, and any one could still be declined. I would rather say that now than have you do the work on an implied promise.

## The split we are asking for

The dependency order makes itself:

1. **`CBM_LANG_IEC_ST` alone** — the grammar language for `.st` / `.iecst`, the vendored grammar, and the extraction work in `extract_defs.c` / `lang_specs.c`. This is the foundation; both containers re-extract as `IEC_ST`, so nothing else can be judged before it.
2. **`CBM_LANG_TWINCAT`** — the `.TcPOU` / `.TcDUT` / `.TcGVL` / `.TcIO` container and `twincat_xml.c`.
3. **`CBM_LANG_PLCOPEN_XML`** — the CODESYS/PLCopen container and `plcopen_xml.c`.

`pass_tcproj.c` and the `.plcproj` / `.tsproj` handling can ride with (2) or come as a fourth — your call, whichever reviews more cleanly.

The practical reason is reviewability. At +282,044 lines over 61 files, no reviewer can give the extraction logic, two XML transforms, a new pipeline pass and a 1,253-line grammar the attention each deserves in one pass — and a self-maintained grammar fork is a long-term maintenance commitment that should be decided on its own, not as a by-product of accepting TwinCAT support.

## Things to fix in whichever PR vendors the grammar

- **`THIRD_PARTY.md` is not updated.** Our vendoring policy requires the LICENSE vendored *with* the code **and** `THIRD_PARTY.md` plus the SBOM updated **in the same change** — notices ship inside the binary archives, so a missing entry is a real gap rather than paperwork. You did update `MANIFEST.md` and `scripts/vendored-checksums.txt` and vendor the LICENSE, so this is the one that got away.
- **The published language count is gated.** Each new language has to be reflected on every surface the contract test checks, and adding three at once against a moving count is exactly how two branches each write `N+1` and merge to a wrong total.
- **This currently conflicts with `main` and has no CI results at all.** Smaller PRs will also get through our queue far better — we are heavily backlogged right now, and a 61-file PR is the worst possible shape for that.

## On the security review

Nothing to worry about there, so you do not need to pre-empt it: the grammar is MIT, the new `tools/tree-sitter-iec-st/package.json` has no install hooks and its dependencies are ordinary tree-sitter tooling in an authoring directory our build does not install, and the `Makefile.cbm` change is a legitimate build-dependency fix. I checked all three.

I know splitting a landed piece of work is unrewarding, and I am sorry to ask for it without a guarantee at the end. But per-language review is the only way this gets a fair hearing rather than a nervous one.

DeusData:

DeusData:
Heads-up, with thanks: one small defect fix bundled in this PR is too useful to wait for the language review, so it is split out as #2083 with you credited as co-author: the semantic-edges pass called the sort with a null array whenever a project has no functions or methods, which is undefined behaviour that the Linux sanitizer lane reports today. When you next rebase this branch that hunk will drop out on its own; the language addition itself stays under review here as before.

DeusData:
Status note: the empty-graph function-sort guard that surfaced while reviewing this PR (a qsort over zero functions) landed on main as 8ec859bb8975 (#2083) and ships in v0.10.9, so it is off this PR's plate. The language support itself is still under review on its own merits. Thank you for the patience on this one.

SarletteD:
Hi @DeusData, hi @ysfsmet,

I've been working on the same topic independently, on
[`SarletteD/codebase-memory-mcp` @ `feature/twincat-structured-text`](https://github.com/SarletteD/codebase-memory-mcp/tree/feature/twincat-structured-text),
and use it daily on a real TwinCAT 3 code base (about 1,300 object files spread over several
libraries; the project itself is proprietary, so I can only share numbers). I don't want to open a
competing PR. #1810 was here first, and you've already given it a split plan. So I'm writing here
to ask how this work could help instead.

**Where the two approaches differ**

- One language, `CBM_LANG_ST`, on the **unmodified** upstream grammar
  (`HeytalePazguato/tree-sitter-iec61131-3-st` @ `00e24f50f8de`, vendored via
  `scripts/vendor-grammar.sh`). TwinCAT XML is `TRANSFORM_ONLY`: `twincat_xml.c` reassembles each
  POU and normalizes the dialect the grammar excludes. Every pass puts back the newlines it
  removes, so definitions and usages map back to exact XML lines. Across 1,380 real files that
  brings the parse-error rate from 84.9 % down to 4.8 %, with no grammar fork to maintain.
  PLCopen XML and `.plcproj` project nodes are **not** covered. #1810 has those.

**What the fork adds that I couldn't find in #1810's diff** (each measured against an XML ground
truth on that code base):

1. **Namespace-qualified bases.** `EXTENDS Lib_A.FB_Base` / `IMPLEMENTS Lib_A.I_X` resolve through
   the `.plcproj` placeholder chain (source → nearest `.plcproj` → alias → placeholder title → the
   library that declares it), in both resolution venues. It never guesses by the last segment,
   because the same short name exists in several libraries. Result: 954/954 resolvable base
   references get an edge, 0 point at the wrong library (IMPLEMENTS 475 → 576, INHERITS 295 → 378).
2. **DUT internals as nodes.** Enum members and struct/union fields become `Field` nodes under their
   `Type` (`enum_value`, `return_type`, `parent_class`). `E_X.MEMBER`, `Alias.E_X.MEMBER` and struct
   fields reached through a declared variable resolve to member-level `USAGE` edges: 2,122 fields,
   6,315 member edges on that tree, and spot checks match a text search.
3. **`line` and `count` on every Structured Text `USAGE` edge.**
4. `get_architecture` lists Structured Text / TwinCAT in its language table.

All of it is covered by tests in `test_extraction.c` / `test_edge_structural.c`. Full
`scripts/test.sh`: 141 suites, 8061 passed, 0 failed. `lint.sh --ci` with clang-format 20.1.8 and cppcheck 2.20.0
reports nothing in the touched files, and `make security` passes.

**What I'd propose**

Once the IEC_ST foundation from #1810's split is settled, I could port items 1–4 onto it as small,
separate issues + PRs (<500 lines each, DCO-signed). The namespace resolution (1) and the DUT
members (2) are each independent of how the TwinCAT XML is transcoded. If you'd rather take the
single-language approach above as the base instead, I can also split that into reviewable steps.
Either way, I won't open anything until you say which

DeusData:
Thank you for the grammar and normalization follow-up. There are now competing foundation choices for the IEC ST work, and that direction needs a maintainer decision before more porting effort. We need more time to compare them; please do not interpret this as a request to rework the stack against an unchosen foundation.

ysfsmet:
I’m completely fine with moving forward with @SarletteD ’s approach if that’s the direction the maintainers would prefer for the IEC ST foundation.

I’ve been dealing with a short-term increase in workload recently, so unfortunately I haven’t been able to dedicate enough time to continue development on this PR.

If you’re willing, @SarletteD, I’d really appreciate it if you could split your implementation into the smaller, focused PRs suggested by @DeusData and submit them separately. I’m happy for the work to continue in that direction, and I can contribute again once my availability improves.

Thanks for picking this up and for the additional work on the TwinCAT/ST support.

Issue #1682, opened 2026-08-16
TypeScript: CALLS edges silently missing for some directly-imported functions — trace_path reports callers_total: 0 (v0.10.5)


softwaredeveto:
## Summary

In a TypeScript monorepo, `CALLS` edges are missing for **some** directly-imported function calls, while structurally identical calls to a sibling helper resolve correctly. Because the misses are silent, `trace_path(direction="inbound")` returns `callers_total: 0`, which is easy to read as "this function has no callers" when it actually has several.

I'm not reporting "call resolution is broken" — most of the graph looks good. I'm reporting that the **failure is indistinguishable from a true negative**, and that it isn't explained by any call-shape rule I could find.

## Environment

- `codebase-memory-mcp@0.10.5` (npm), Windows 11
- TypeScript monorepo: NestJS API + React (Vite) + React Native, pnpm workspaces
- Index: 21,937 nodes / 115,747 edges, ~56 s
- Queried via `cli query_graph` and the MCP `trace_path` / `query_graph` tools (same results)

## Observed

Three symbols, same repo, same index, same query shape:

| # | symbol | how it is called | graph | actual (grep) |
|---|---|---|---|---|
| A | `helperA` — exported function in `modules/pricing/util.ts` | `import { helperA }` + direct call | ✅ 5 production callers, function-level | matches |
| B | `helperB` — exported function in `modules/campaign/engine.ts` | `import { helperB }` + direct call | ❌ **0** `CALLS` (only `DEFINES`) | **3 production call sites** |
| C | `ServiceX.methodY` — class method | `this.injected.methodY(...)` (NestJS DI) | ❌ **0** production; 13 `USAGE` edges, all in `*.spec.ts` | **11 call sites / 6 files** |

**A and B have the identical call shape** — plain named import, direct invocation, callers are class methods in other modules. One resolves, the other does not. That's what makes this hard to work around: there is no rule I can give a user (or myself) for when to trust a zero.

Query used (same for all three):

```cypher
MATCH (a)-[e]->(b) WHERE b.name = '<symbol>'
RETURN type(e) AS edge, a.name AS source, a.file_path AS file
```

For B the only inbound edges are `DEFINES`. For C the only inbound edges are `USAGE` from test files — interestingly the test mocks (`{ methodY: jest.fn() }`) resolve **perfectly** (13/13, exactly matching grep), while none of the 11 real call sites do.

## What I ruled out

[...]

PR #1811, opened 2026-08-23
fix(pipeline): preserve dotted TypeScript import basenames

pcristin:
## What does this PR do?

Preserve extensionless dotted JavaScript/TypeScript module basenames such as
`featureX.engine` during relative-import resolution.

- Strip only recognized JS/TS module file extensions instead of treating every
  final dot suffix as an extension.
- Avoid applying generic module-extension stripping a second time after the
  relative resolver has normalized the path.
- Add a five-file TypeScript regression reproducing #1682 and assert that the
  `helperB` and `SomeType` `IMPORTS` edges both target `featureX.engine.ts`.

Refs #1682.

## Verification

- `make -f Makefile.cbm test-focused TEST_SUITES='fqn edge_imports'` — 143 passed
- `make -f Makefile.cbm test-par` — 7,577 passed, 0 failed, 8 skipped across 139 suites
- `git diff --check`
- Changed files pass `clang-format --dry-run --Werror`
- Diff-scoped clang-tidy passes

The repository-wide `lint-ci` baseline is not clean in untouched files: cppcheck
reports existing style findings in `src/store/store.c` and `src/cypher/cypher.c`,
and the formatter reports existing findings in `src/mcp/mcp.c` and
`src/pipeline/pipeline_incremental.c`. No unrelated lint cleanup is included here.

## Checklist


pcristin:
CI triage: the only substantive red job is `test / test-windows-guards`, where `section_cold_storm` lost one of six concurrent clients while they raced to spawn the ephemeral daemon. All dedicated Windows test shards, sanitizers, lint, security, and DCO jobs passed.

This PR changes only FQN/import resolution and its regression tests; it does not touch daemon or Windows-guard code. The same cold-storm timeout occurred on the unrelated [PR #1245 workflow run](https://github.com/DeusData/codebase-memory-mcp/actions/runs/32657966944/job/97239853174), while surrounding runs of this guard passed, so this matches an
existing intermittent guard failure rather than this patch.

I tried to rerun only the failed jobs, but GitHub rejected it because contributors do not have repository Actions rerun permission. Could you @DeusData rerun the failed jobs? I have deliberately not added an unrelated code change or empty commit just to retrigger CI.

DeusData:
Thank you for the careful CI attribution. I checked the diff and current main: this PR is limited to FQN and import resolution plus its regression tests, while the red job is the Windows daemon guard. That failure is outside this change path and matches the existing startup-flake family you linked.

You are also correct that contributors cannot rerun repository jobs. This should be rerun from the maintainer side, and there is no need to create an unrelated commit. Thank you for documenting the evidence instead of treating the red aggregate as proof against the patch.

adfjadfj16-a11y (dismissed):

adfjadfj16-a11y (dismissed):

DeusData:
Approved on merit. **The one red is not yours**, and I can be specific about that.

## The failure

`test / test-unix (macos-15-intel)` fails at:

```
lock_registry_absolute_deadline_survives_repeated_wakes
  FAIL tests/test_lock_registry.c:1153: ASSERT(tail_queued)
```

with run totals `7617 passed, 1 failed`. Your change touches `src/pipeline/fqn.c` and `src/pipeline/pass_pkgmap.c` — the import-path resolver. It cannot reach the lock registry.

This is a known flake we have already diagnosed: the assertion's verdict is fed by a **100 ms wall-clock window**, so a loaded runner decides the result rather than the code. We saw the identical signature — same test, same line, same platform — on 28 August on an unrelated NetBSD-only `#ifdef` change. Your PR is the second unrelated diff to be reddened by it, which I have recorded; it makes fixing it properly a priority rather than a curiosity. Nothing for you to do.

## On the change

The bug is a good one to have found, because the old code encoded a false rule: *the text after the final dot is an extension*. For `featureX.engine` that silently rewrites the module identity, and the resulting `IMPORTS` edges point at a module that does not exist. Stripping only **recognised** JS/TS module extensions replaces a guess with a fact, which is the right direction.

Two things I want to credit:

**Removing the second, generic strip after the relative resolver has normalised the path.** Double-stripping is the kind of bug that hides behind a passing test suite for years, because it only bites inputs that survive the first pass with a dot still in them. Noticing that the two stages overlapped is more valuable than the extension list itself.

**Asserting both edges in the regression** — `helperB` *and* `SomeType` both targeting `featureX.engine.ts`. A value import and a type import take different paths through the resolver, so pinning only one would have left half the fix unguarded.

## On your lint note

> the formatter reports existing findings in `src/mcp/mcp.c` and `src/pipeline/pipeline_incremental.c`

That is a **false positive on your side, not drift on ours**, and you are the third contributor to hit it this week. Our gate requires the Homebrew LLVM `clang-format`; a distro or standalone `clang-format-20` reports whole-file differences on exactly those large files. I ran `make -f Makefile.cbm lint-ci` against current `main` today with Homebrew `clang-format 22.1.8` and it exits clean. So your instinct to include no unrelated lint cleanup was exactly right — there was nothing to clean.

I will raise making that explicit in CONTRIBUTING, since three people have now spent time on it.

## Before merge

Four checks are still queued. Our Actions pool is servicing roughly one job at a time against 21 queued PR runs, so this is slow for reasons unrelated to your PR. I will merge once it settles, and the `lock_registry` red is not a blocker.

DeusData:

pcristin:
The latest red check appears unrelated to this PR's #1682 changes and is caused by an upstream toolchain change.

The failing job is `test / test-lsan-macos`: https://github.com/DeusData/codebase-memory-mcp/actions/runs/33745916176/job/100622049644

It fails during compilation with:

```
src/cli/cli.c:1768:21: error: variable 'g_mcp_command_path_probe_counter' set but not used [-Werror,-Wunused-but-set-global]
```

The previous LSan run on this PR installed Homebrew LLVM 22.1.8 and passed. This run installed the newly available LLVM 23.1.0 through the workflow's unpinned `brew install llvm ccache`, which exposes the existing POSIX test-build diagnostic.

The PR branch's `src/cli/cli.c` is identical to current upstream `main`; the PR diff remains limited to the intended FQN/import files. The latest merge from `main` also changed only `src/mcp/mcp.c` and `tests/test_mcp.c`.

This likely needs an upstream fix to make the test-only probe counter platform-correct, or a temporary LLVM pin. Another branch update or unrelated change in this PR should not be necessary.

DeusData:
Thank you — that diagnosis is exactly right, and it was the first sighting of a real problem.

Run 33745916176 poured Homebrew LLVM 23.1.0 and died at `src/cli/cli.c:1768` under `-Wunused-but-set-global`; the run on your current head drew the 22.1.8 bottle and got past the compile, which is why the two runs look different. The compile red is `main`'s, not yours: the counter's only read is Windows-only, so the POSIX store is genuinely dead and Clang 23 is correct to say so. The fix is #2027.

The remaining red on your current head (`subprocess_run_spawn_failure`, `test_subprocess.c:272`) is the sanitizer-lane spawn-starvation family I described above — also not yours. I have re-queued the full run; once #2027 is on `main` I will make sure this gets a run against it, and it merges on green. Nothing needed from you.

DeusData:
Merged as 3bbabf18 — thank you, and thanks for the patience across the reruns and the LLVM 23 compile detour (that one was ours, fixed in #2027). Dotted TypeScript import basenames now keep their full name through `fqn.c` and the package map, so `foo.service` no longer collapses to `foo` and the IMPORTS edges land on the right module. Really appreciate you seeing #1682 through.

PR #1817, opened 2026-08-24
fix(ui): allow graph schema through UI RPC

### LINKED ISSUE #1663: Graph UI stats panel broken in v0.10.5: /rpc returns 403 for get_graph_schema ('UI RPC method is not allowed')
## Environment
- Windows 11, v0.10.5 (exe install), UI at http://127.0.0.1:9749
- Agents connected via MCP (Claude Code, Codex, Hermes)

## Symptom
The Projects tab's summary tiles show `Projects: 45, nodes: 0, edges: 0` and each project row shows 0 nodes / 0 edges. The general/stats panel previously showed correct totals and language counts. The MCP channel is fully functional: `list_projects` returns per-project node/edge counts, and `index_status` reports `ready` for every project.

## Root cause
The SPA (served from the same binary, assets/index-*.js) fetches the projects list and then calls, per project:

```js
POST /rpc
{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"get_graph_schema","arguments":{"project":"<name>"}}}
```

The server responds `403` with:

```json
{"jsonrpc":"2.0","error":{"code":-32601,"message":"UI RPC method is not allowed"},"id":null}
```

Only `list_projects` is permitted over the HTTP /rpc route; `get_graph_schema` and `get_code_snippet` are both rejected. The SPA swallows the failure (`schema: null` → contributes 0), so the panel sums to zero without any visible error.

## Repro
```
curl -s -X POST http://127.0.0.1:9749/rpc -H "Content-Type: application/json"   -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"get_graph_schema","arguments":{"project":"<any indexed project>"}}}'
# -> HTTP 403, "UI RPC method is not allowed"

# same request with "list_projects" -> 200 OK
```

## Notes
- `config list` exposes only `auto_index / auto_index_limit / auto_watch / ui-lang / ui_enabled / ui_port` — no allowlist or auth knob.
- If the HTTP-rpc lockdown is intentional loopback hardening (the UI listener replies with `Access-Control-Allow-Origin: *`), the UI still needs a sanctioned path for its own stats queries — either allow `get_graph_schema`/read-only graph tools on loopback, or serve the aggregate through a dedicated `/api` endpoint.
- This regressed between v0.9.0 (panel worked) and v0.10.5.

astandrik:
## What does this PR do?

The Graph UI already calls `get_graph_schema` while loading projects, but the loopback UI RPC allowlist rejected that call with HTTP 403.

This adds only `get_graph_schema` to the existing `list_projects` / `get_code_snippet` allowlist. The regression test now covers all three allowed tools and keeps `search_graph`, `delete_project`, `manage_adr`, `ingest_traces`, and `index_repository` blocked. The existing `initialize` and duplicate `params.name` checks remain unchanged.

Fixes #1663

## Verification

- Reproduce-first HTTPD run: applying the test change alone fails the target allowlist case with HTTP 403.
- `scripts/test.sh --suites httpd`
- `scripts/test.sh`
- `scripts/lint.sh --ci`
- `scripts/build.sh --with-ui BUILD_DIR=build/issue1663-ui CC=gcc CXX=g++`
- `SMOKE_REQUIRE_UI=1 scripts/smoke-local.sh build/issue1663-ui/codebase-memory-mcp ui`
- `make -j32 -f Makefile.cbm security BUILD_DIR=build/issue1663-security`

The full Linux gate passed on the remote builder against the exact commit tree. Windows runtime behavior remains for CI to confirm.

## Checklist

- [x] Tests pass (`scripts/test.sh`)
- [x] Lint passes (`scripts/lint.sh --ci`)
- [x] New behavior is covered by a reproduce-first regression test

### DIFF
--- a/src/ui/http_server.c
+++ b/src/ui/http_server.c
@@ -1756,7 +1756,8 @@ static bool rpc_is_allowed_for_ui(const char *body, size_t body_len) {
     const char *name_text = yyjson_is_str(name) ? yyjson_get_str(name) : NULL;
     bool allowed =
         method_text && strcmp(method_text, "tools/call") == 0 && name_text &&
-        (strcmp(name_text, "list_projects") == 0 || strcmp(name_text, "get_code_snippet") == 0);
+        (strcmp(name_text, "list_projects") == 0 || strcmp(name_text, "get_code_snippet") == 0 ||
+         strcmp(name_text, "get_graph_schema") == 0);
     yyjson_doc_free(document);
     return allowed;
 }
--- a/tests/test_httpd.c
+++ b/tests/test_httpd.c
@@ -1169,22 +1169,31 @@ TEST(ui_server_mutations_require_json_content_type) {
 TEST(ui_server_rpc_allows_only_ui_read_tools) {
     th_server_t ts;
     ASSERT_EQ(th_server_start(&ts), 0);
-    const char *body = "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"tools/call\","
-                       "\"params\":{\"name\":\"list_projects\",\"arguments\":{}}}";
     char req[1024];
-    snprintf(req, sizeof(req),
-             "POST /rpc HTTP/1.1\r\n"
-             "Content-Type: application/json\r\n"
-             "Content-Length: %d\r\n\r\n%s",
-             (int)strlen(body), body);
     char resp[8192];
-    int n = th_http(cbm_http_server_port(ts.srv), req, resp, sizeof(resp));
-    ASSERT_GT(n, 0);
-    ASSERT_EQ(th_status(resp), 200);
-    ASSERT_NOT_NULL(strstr(resp, "\"jsonrpc\""));
+    int n;
 
-    static const char *blocked_tools[] = {"delete_project", "manage_adr", "ingest_traces",
-                                          "index_repository"};
+    static const char *allowed_tools[] = {"list_projects", "get_code_snippet", "get_graph_schema"};
+    for (size_t i = 0; i < sizeof(allowed_tools) / sizeof(allowed_tools[0]); i++) {
+        char allowed_body[512];
+        snprintf(allowed_body, sizeof(allowed_body),
+                 "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"tools/call\","
+                 "\"params\":{\"name\":\"%s\",\"arguments\":{}}}",
+                 allowed_tools[i]);
+        snprintf(req, sizeof(req),
+                 "POST /rpc HTTP/1.1\r\n"
+                 "Content-Type: application/json\r\n"
+                 "Content-Length: %zu\r\n\r\n%s",
+                 strlen(allowed_body), allowed_body);
+        n = th_http(cbm_http_server_port(ts.srv), req, resp, sizeof(resp));
+        ASSERT_GT(n, 0);
+        ASSERT_EQ(th_status(resp), 200);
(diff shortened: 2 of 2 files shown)


DeusData:
Thank you for isolating this to the UI RPC allowlist and for covering both the allowed and blocked methods. Adding `get_graph_schema` changes the loopback UI RPC security boundary, so we need more time to review the route guarantees and the complete allowlist before making a merge decision. We will come back with a decision; no additional work is requested from you at this stage.

DeusData:
Approved on merit. **All 24 red checks are one environmental cause and none of them are yours** — details below, because 24 failures on a two-line change is alarming until you know why.

## The change

This touches a security allowlist, so I looked at it as one. `http_server.c:1757-1759` currently gates the loopback UI RPC on `method == "tools/call"` and `name ∈ {list_projects, get_code_snippet}`.

The argument that settles it: **`get_code_snippet` is already allowed, and it returns actual source code.** `get_graph_schema` returns node and edge type names. Adding it is strictly *less* exposure than what the allowlist already permits, over the same loopback-only channel, for a call the UI already makes and currently gets a 403 for.

Your test keeps `search_graph`, `delete_project`, `manage_adr`, `ingest_traces` and `index_repository` blocked, which is the half that matters — the mutating tools stay out, and the negative cases are what stop this allowlist growing by accident later.

Doing it reproduce-first, so the test change alone fails with the 403 before the source change, is exactly right for an allowlist addition: it proves the guard was actually the thing rejecting the call, rather than something else upstream.

## Why 24 checks are red

Every failing leg dies at the same preflight, before any test runs:

```
ERROR: repository build/ ancestor must be absent or a real directory
```

That is `scripts/path-safety.sh:30`, which refuses to proceed when the repository's `build/` is a **symlink** or exists as a non-directory. It is a guard against `rm`-ing outside the repo, and it fires on the state of the runner's working tree — not on anything in a diff. A two-line change to `src/ui/http_server.c` cannot produce it.

That single cause explains the whole board: every `test-unix` shard, every `test-tsan`, the sanitizer legs, `pr-smoke` on all three platforms, `shard-completeness` (its dependencies died), and `ci-ok` (it aggregates). One environmental fault, twenty-four red checks.

I have recorded the signature so this is recognised immediately next time rather than re-diagnosed.

## Before merge

You are only **8 commits behind `main`**, which is the freshest of anything in my queue today — thank you for keeping it current.

I would normally rebase to trigger a clean re-run, but our Actions pool is in a bad state: there is exactly **one** job executing repo-wide against **24 queued runs**, so every new run I start makes the wait worse for everyone. I am deliberately holding off and will refresh this once the pool recovers.

Your local verification list is unusually thorough — the UI build, `SMOKE_REQUIRE_UI=1 smoke-local.sh`, and the security target — and saying plainly that "Windows runtime behavior remains for CI to confirm" is the right way to state a limit rather than imply coverage you do not have.

DeusData:
An update that you are owed an honest explanation for, because this PR was approved three days ago. A direction decision has landed since: the graph UI is being reimplemented from the ground up, and the current app together with its loopback RPC layer — the allowlist this change edits — will be replaced wholesale rather than maintained. With that settled, merging a change into the surface that is about to disappear gives it a very short life, and we would rather be straight with you than land it and delete it in the next release cycle. So we are closing this one, with #1663 closed alongside it as superseded.

What carries forward: the stats panel's schema request is on the requirements list for the new version, and so is your point that the allowlist should be tested from both sides — the allowed and the blocked method — which was the part of this PR that made it a good one. Thank you for isolating the bug precisely and for a two-line fix with more test than change; that is exactly the shape we like, and you are very welcome back.

PR #1819, opened 2026-08-24
fix(ui): load all paginated projects (#1812)

### LINKED ISSUE #1812: Management UI only displays the first 50 projects and ignores list_projects.has_more
### Version

codebase-memory-mcp 0.10.8

### Platform

macOS (Apple Silicon)

### Install channel

### Binary variant

standard

The Codebase Memory management UI only displays the first 50 indexed projects.

I currently have 86 indexed projects. The `list_projects` API/CLI correctly returns the following pagination metadata:

- total: 86
- offset: 0
- limit: 50
- returned: 50
- has_more: true

However, the management UI only requests and displays the first page. It does not provide pagination, a “Load more” action, or automatically request the remaining pages.

As a result, projects after the first 50 are indexed and accessible through the CLI, but they cannot be found in the management UI.

This can be verified by requesting the second page:

codebase-memory-mcp cli --json list_projects --offset 50 --limit 50 --include-details true

Expected behavior:

The management UI should either:

1. Support pagination;
2. Provide a “Load more” action; or
3. Automatically fetch subsequent pages while `has_more` is true.

Environment:

- codebase-memory-mcp version: 0.10.8
- Total indexed projects: 86
- Management UI port: 9749

### Reproduction

The Codebase Memory management UI only displays the first 50 indexed projects.

I currently have 86 indexed projects. The `list_projects` API/CLI correctly returns the following pagination metadata:

- total: 86
- offset: 0
- limit: 50
- returned: 50
- has_more: true

However, the management UI only requests and displays the first page. It does not provide pagination, a “Load more” action, or automatically request the remaining pages.

As a result, projects after the first 50 are indexed and accessible through the CLI, but they cannot be found in the management UI.

This can be verified by requesting the second page:

codebase-memory-mcp cli --json list_projects --offset 50 --limit 50 --include-details true

Expected behavior:

The management UI should either:

1. Support pagination;
2. Provide a “Load more” action; or
3. Automatically fetch subsequent pages while `has_more` is true.

Environment:

- codebase-memory-mcp version: 0.10.8
- Total indexed projects: 86
- Management UI port: 9749

### Logs

```text

```

```text

```

_No response_

### Confirmations

rarepops:
## What does this PR do?

Fixes #1812.

- follows `list_projects` pagination until `has_more` is false
- publishes the complete project list before schema hydration finishes
- caps schema hydration at four concurrent requests
- rejects non-advancing pagination metadata instead of looping indefinitely
- adds regressions for a 51-project result and bounded schema concurrency

## Validation

- `cd graph-ui && npm test` (41 passed)
- `cd graph-ui && npm run build`
- `git diff --check`

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) (full native suite not run; focused UI suite passed)
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) (not run locally)
- [x] New behavior is covered by a reproduce-first test


rarepops:
CI note: the fresh run's `test-windows-guards` failure is outside this PR's two TypeScript files and appears to cascade from guard setup state.

The raw job log shows:

1. `test_non_ascii_path.py` first reports `SETUP FAIL: ASCII baseline did not index` after its Unicode relocation check passed.
2. The next daemon lifecycle guard then finds an active daemon with a different cache/build and cannot start its isolated daemon.
3. The daemon stability guard fails for the same active-cache/build mismatch.
4. Both regular Windows shards pass, as do macOS TSan (the prior unrelated failure), Linux TSan, UI build, lint, DCO, and security checks.

The branch is current with `main`, mergeable, and changes only:

- `graph-ui/src/hooks/useProjects.ts`
- `graph-ui/src/hooks/useProjects.test.tsx`

I do not have permission to rerun repository jobs. A maintainer rerun of `test-windows-guards` should distinguish this guard-state cascade without adding unrelated churn to the PR.

DeusData:
Thank you for the reproduce-first pagination coverage and for bounding schema hydration concurrency while publishing the project list early. The contribution queue is quite full, so the full review may take a little time. We have the PR routed and will come back with grounded feedback as soon as possible.

DeusData:
Approved. The pagination loop is written with the right posture, which is what makes this safe rather than merely correct.

**You treat the server's own pagination metadata as untrusted.** That is the part I would have asked for:

```ts
const pageOffset = Number.isSafeInteger(page.offset) ? page.offset! : offset;
const returned   = Number.isSafeInteger(page.returned) ? page.returned! : pageProjects.length;
const nextOffset = pageOffset + returned;
if (returned <= 0 || nextOffset <= offset) {
  throw new Error("list_projects pagination did not advance");
}
```

Two distinct ways a `while (true)` over a paginated API spins forever — a page claiming `has_more` while returning nothing, and an offset that fails to advance or goes backwards — and both are caught, with non-numeric metadata falling back to locally known values rather than poisoning the arithmetic. A loop that trusted `offset` and `returned` would have been shorter and would hang the UI on a server bug instead of surfacing it.

**Publishing the complete list before schema hydration finishes** is the right ordering: the thing the user asked for appears as soon as it is known, and the enrichment fills in behind it. Capping hydration at four concurrent requests keeps that from turning a 51-project workspace into 51 simultaneous calls.

**And the regression is sized to the bug** — 51 projects against a 50-item page is exactly the boundary that was broken, rather than a round number that happens to exceed it.

## Your CI red is not yours

`test / test-windows-guards` fails, and it cannot be this change: a `graph-ui/src/hooks/*.ts` edit does not make a Windows fixture repository index with null counts. The job's own setup could not index a plain ASCII fixture:

```
SETUP FAIL: ASCII baseline did not index:
  {'nodes': None, 'edges': None, 'definition_nodes': 5}
```

Everything downstream then skipped on a failed precondition — including `test_ui_drive_listing.py`'s "Is this a UI build?", which is part of the same cluster and fails on non-UI pull requests too, so it is not evidence about your change either.

I have now confirmed this identical cluster, with that identical diagnostic, on five unrelated pull requests spanning several days. It is ours to fix. Please do not chase it.

**On the unticked boxes:** not running the full native suite or `lint-ci` for a change confined to `graph-ui/` is a reasonable call, and saying so beats ticking them. The focused UI suite and the build are the relevant gates here.

**Please rebase** — `main` moved three times yesterday: broken by a duplicate-symbol merge, repaired by #1993, then #1703 landed.

DeusData:
This one needs an honest explanation, because it was approved and was next in the merge order. Since then a direction decision has landed: the graph UI is being reimplemented from the ground up, and the current React app — including the projects hook this fixes — will be replaced wholesale rather than maintained. Merging into a surface that is about to disappear gives the fix a very short life, so we are closing this and #1812 as superseded rather than landing it and deleting it within a release cycle.

What carries forward is the posture rather than the code: treating the server's own pagination metadata as untrusted, bounding the hydration concurrency, and publishing the list early are going in as requirements for the new projects view, with this PR as the reference. Thank you for a reproduce-first fix with more test than change — that is the shape we like — and for the careful CI attribution on the Windows guards leg, which was correct. You are very welcome back.

PR #1822, opened 2026-08-24
fix(discover): report a symlink skip through the ignored-files taxonomy

### LINKED ISSUE #1815: `index_repository` silently skips all files reached via a directory symlink or NTFS junction
### Version

codebase-memory-mcp 0.10.8

### Platform

Windows (x64)

### Install channel

### Binary variant

ui

On Windows (and probably also Linux), `index_repository` does not follow directory symlinks or NTFS junctions when walking the repo tree. Files reachable only through such a reparse point are completely absent from the resulting graph — they're not indexed, and they're also not reported in `skipped`, `parse_partial`, or `not_indexed` in the response or in `index_status`. From the tool's perspective, the directory (and everything under it) simply doesn't exist. This makes it impossible to index a codebase that uses linked/shared source directories without physically duplicating or hard-linking files.

**Environment**

- OS: Windows 10/11
- Filesystem: NTFS
- Index mode tested: `fast`, `moderate` (same result in both)

**Additional Information**

- Tested both a directory **symlink** and an NTFS **junction** (`mklink /J`) as the link type — both produce identical (null) results, so this isn't specific to one reparse-point type.
- Tested indexing from multiple `repo_path` levels (the linked dir itself, its parent, and its grandparent) — result is the same regardless of which ancestor is passed as `repo_path`. The walker appears to prune the reparse point wherever it's encountered during the walk.
  - One exception: when the linked directory itself is passed as `repo_path`, it is resolved to its target directory and the target files are indexed successfully. As stated, the issue onlyaffects traversal through a reparse point below the selected `repo_path`.
- As a control: replacing the directory symlink with a **hard link of the individual file** (`New-Item -ItemType HardLink`) — i.e. a real directory entry pointing at the same file content, not a reparse point — makes the file show up correctly in the graph. This confirms the walker specifically special-cases/skips reparse points (symlink and junction), rather than e.g. following a max-depth or `.gitignore`-driven exclusion.
- `.gitignore`/`.cbmignore` are not the cause: `git check-ignore` on files behind the link correctly resolves per the gitignore rules — the exclusion is happening below the git layer, at the file-walk stage.

**Impact**

Any project layout using symlinked/junctioned shared source directories (common for legacy code shares, vendored/linked libraries, or monorepo cross-project sharing) gets silently and invisibly incomplete indexing. Because there's no coverage signal for it, it's easy to trust `index_status`'s "clean" coverage report while a whole subtree is missing.

## Suggested fix

At minimum, report skipped symlinks and junctions in `index_status`, for example:

```json
{
  "path": "linked",
  "reason": "reparse-point"
}
```

Ideally, add an opt-in option to `index_repository`, such as:

```json5
"follow_symlinks": true
```

This should allow legitimate linked source directories to be traversed while keeping the current safe default behavior.

### Reproduction

1. Create a minimal repo layout ( I used the D-drive for fewer permission issues, you may adjust it to C:):
   ```
   D:\repro\
     legacy\
       shared\
         Bar.cs          <-- contains: public enum Bar { A, B }
     project\
       src\
         Foo.cs          <-- ordinary file, references Bar somehow or not, doesn't matter
       linked            <-- directory symlink OR NTFS junction -> D:\repro\legacy\shared
   ```
   Create the link with either:
   ```powershell
   # symlink (needs elevation / Developer Mode)
   New-Item -ItemType SymbolicLink -Path D:\repro\project\linked -Target D:\repro\legacy\shared

   # OR junction (no elevation needed) — same result
   cmd /c mklink /J D:\repro\project\linked D:\repro\legacy\shared
   ```
  <details>
  <summary>Script to set up this repo</summary>

   ```cmd
   @echo off
   set "ROOT=D:\repro"
   
   mkdir "%ROOT%\legacy\shared" "%ROOT%\project\src" 2>nul
   
   >"%ROOT%\legacy\shared\Bar.cs" echo namespace Shared;
   >>"%ROOT%\legacy\shared\Bar.cs" echo.
   >>"%ROOT%\legacy\shared\Bar.cs" echo public enum Bar { A, B }
   
   >"%ROOT%\project\src\Foo.cs" echo namespace Project;
   >>"%ROOT%\project\src\Foo.cs" echo.
   >>"%ROOT%\project\src\Foo.cs" echo public static class Foo { }
   
   if exist "%ROOT%\project\linked" (
       echo Link already exists: %ROOT%\project\linked
       exit /b 1
   )
   
   mklink /J "%ROOT%\project\linked" "%ROOT%\legacy\shared"
   
   echo.
   echo Repro created at %ROOT%\project   
   ```

   </details>

2. Index the repo:
   ```
   index_repository({ "repo_path": "D:/repro/project",  "name": "repro-junction",  "mode": "moderate",  "persistence": false })
   ```
3. Inspect the result:
   ```
   search_graph({ "project": "repro-junction",  "name_pattern": ".*Bar.*",  "limit": 50,  "format": "json" })
   index_status(project="repro-junction")
   ```

**Expected behavior**

Either:
- `Bar.cs` (reached via `project/linked/Bar.cs`) is indexed like a normal file, or
- if symlinks are intentionally not followed, the file/directory should show up in `index_status` under `not_indexed` or `skipped` with a clear reason (e.g. `"reason": "symlink"` / `"reason": "reparse-point"`), so it's at least *visible* that content was excluded.

**Actual behavior**

- `search_graph` returns zero results for anything under `linked/`.
- `index_status` node/edge counts don't change whether the link exists or not.
- `linked/` and everything beneath it is **absent from every coverage report** (`not_indexed`, `skipped`, `parse_partial`) — there is no signal anywhere that content was excluded. It looks identical to "this directory doesn't exist," not "this directory was intentionally skipped."

In the reproduction, `index_repository` returned:

```json
{
  "skipped_count": 0,
  "parse_partial_count": 0,
  "not_indexed_files_count": 0
}
```
`index_status` also returned empty lists for all three coverage categories.

Note that git itself follows the junction:

```bash
git add -A ; git status --short
A  linked/Bar.cs
A  src/Foo.cs
git check-ignore -v linked/Bar.cs → exit 1 (not ignored)
```

### Logs

```text
=== STDERR ===
Preparing one-shot local CBM command...
level=warn msg=mem.allocator.preloading_completed still_preloading=false detail=allocator_was_still_preloading,_so_arena_creation_and_purging_were_disabled;_process_init_completed_explicitly
level=info msg=mem.allocator.owned classes=all
level=info msg=mem.init budget_mb=11371 total_ram_mb=32489 source=ram_fraction
Running index_repository locally...
Completed index_repository (2749 ms)
=== STDOUT ===
{"project":"repro-logcapture","not_indexed_files_count":0,"skipped_count":0,"parse_partial_count":0,"nodes":6,"edges":5,"expected_nodes":6,"expected_edges":5,"adr_present":false,"adr_hint":"Project indexed. Consider creating an Architecture Decision Record: explore the codebase with get_architecture(aspects=['all']), then use manage_adr(mode='update') to persist architectural insights across sessions.","artifact_present":false,"status":"indexed"}
```

```text

```

_No response_

### Confirmations

AmirF194:
## What does this PR do?

`walk_dir_process_entry` rejects a directory entry when `safe_stat()` skips it as a symlink or reparse point, but that branch returns with a bare `return;` and never calls `file_list_add_ignored`, unlike its two sibling skip paths a few lines below (directory exclusion, file `skip_reason`). A file reachable only through a symlinked directory disappears from the index with no trace in `not_indexed`/`skipped`/`parse_partial`, so a symlink-heavy subtree looks identical to a complete index (#1815).

This is the narrower, buildable half of #1815 that DeusData already called out separately from the traversal-contract question (whether symlinks should be followed), which they said needs more design time. This PR does not touch that decision.

`safe_stat()` now takes an `is_symlink` out-param reporting whether a rejection was specifically the symlink/reparse-point check (as opposed to some other stat failure). `walk_dir_process_entry` records the entry via `file_list_add_ignored(out, rel_path, "symlink")` when so. The `count_only` path is untouched.

**Rebase risk, not a conflict of claims**: open PR #1803 also changes `safe_stat()`'s signature (adds its own `is_link` out-param) to fix a different bug, `count_only` aborting discovery entirely on any symlink. If #1803 merges first, this branch will need a rebase to reuse its out-param instead of adding a second one; read its full diff before assuming a collision, it does not add any ignored-file reporting itself.

## Verification

- New regression test `discover_symlink_skip_is_reported` (`tests/test_discover.c`) fails on unmodified `main` (`ignored_total == 0`, expected 1) and passes on this branch; ran both in a clean `ubuntu:24.04` container via `scripts/test.sh --suites discover`, 114/114 passed, no leaks under the default ASan/LSan build.
- Full `scripts/test.sh` (all suites) passes in the same container, with one exception: `tests/test_parallel_harness_contract.sh`'s `stubborn_tree` case fails without `docker run --init`; confirmed it passes identically with `--init` on a fresh container, so it is a container-PID-1 artifact of my verification setup, not related to this diff (it does not touch anything under `src/discover/`).
- `scripts/lint.sh --ci` with the exact versions CI pins (cppcheck 2.20.0 built from source, `clang-format-20`): both changed files are clean. The overall script still exits non-zero because of pre-existing `clang-format-20` drift in `src/mcp/mcp.c` and `src/pipeline/pipeline_incremental.c`, both untouched by this diff, confirmed present on unmodified `main`.
- Not verified: the Windows branch of `safe_stat()` (also updated, same shape) has no Windows environment available here; and `lint-mem`'s path-sensitive clang-tidy analyzer (CI job, 60-minute budget, whole-repo) was not run locally.

Fixes #1815


AmirF194:
Checking in on this one, no rush. Happy to adjust the fix if the reporting taxonomy should work differently than the two existing skip paths it mirrors.

DeusData:
Thank you for keeping this to the ignored-file reporting half of #1815 and for calling out the exact overlap with #1803. That makes the intended change and rebase risk clear. The contribution queue is quite full, so review may take a little time, but we have the PR routed and will return with feedback as soon as possible.

DeusData:
Approved. And thank you for the paragraph about #1803 — that is the check I would otherwise have had to do, and I have just been bitten by not doing it on another pair of PRs, so it is worth saying it landed well.

**I verified it and you characterised it correctly.** #1803 is open, touches the same two files, changes `safe_stat()`'s signature for a different bug — a symlink aborting a bounded count — and adds **zero** `file_list_add_ignored` calls. "Rebase risk, not a conflict of claims" is exactly right: the two are complementary, not competing.

**Suggested ordering**, subject to #1803's own review: let it land first, since its symptom is that any repository containing a symlink can never auto-index at all, and then rebase this to reuse its `is_link` out-param rather than adding a second one. If that ordering turns out to be inconvenient, say so — it is a preference, not a requirement.

## The bug

A bare `return;` where the two sibling skip paths a few lines below both call `file_list_add_ignored` is the whole defect, and the consequence is the bad kind: a file reachable only through a symlinked directory vanishes with no trace in `not_indexed`, `skipped` or `parse_partial`. **A symlink-heavy subtree then looks identical to a complete index.** That is worse than an error, because nothing invites anyone to look — the same shape as several other defects moving through this queue at the moment.

Threading an `is_symlink` out-param so the caller can distinguish "rejected specifically as a symlink or reparse point" from "some other stat failure" is the right granularity. Reporting every `safe_stat` rejection as `"symlink"` would have been easier and wrong.

**And you held the line on scope.** Not touching whether symlinks should be followed — the traversal-contract question that was explicitly deferred for design time — is what makes this reviewable now instead of later. Leaving `count_only` alone is consistent with that.

## The verification section is a model

Three things in it are worth other contributors copying:

- The regression **fails on unmodified `main`** with the exact observed value (`ignored_total == 0`, expected 1), so the test demonstrably tests the fix.
- The `stubborn_tree` failure is attributed rather than waved away: a container-PID-1 artifact that passes identically with `docker run --init`, on a path this diff does not touch. That is a checkable claim, not a reassurance.
- The **"Not verified" list is explicit** — the Windows branch of `safe_stat()` with no Windows environment available, and `lint-mem`'s whole-repo analyzer. Naming what you could not test is more useful than a checklist of ticks, and it tells me exactly where CI is doing the work you could not.

The pre-existing `clang-format-20` drift in `src/mcp/mcp.c` and `src/pipeline/pipeline_incremental.c` is real and not yours; both files are untouched here.

**Status:** cleared and approved, `CLEAN` and green. I will re-run against current `main` before merging — it moved three times yesterday.

AmirF194:
That ordering works for me. I'll hold this until #1803 lands, then rebase onto its `is_link` out-param instead of adding a second one.

PR #1823, opened 2026-08-24
fix(ci): stop kill-grace from bounding Windows helper spawns

tthayer:
Split out of #1426 at maintainer request — process-control change, reviewed on its own.

## Problem

The Windows shard failed Step 0i (parallel suite scheduler contract) with `cleanup failed: hang_after_summary: leader exited leaving live descendants` while its sibling shard passed. Nothing was actually leaking a process.

`run-test-wave.py` passed `--kill-grace` as the subprocess timeout for two external Windows helpers: `taskkill.exe /T /F` and the `powershell.exe` `Get-CimInstance` descendant probe. Those are different quantities. `kill_grace` budgets how long a doomed process may take to die; the helpers also pay process spawn plus, for PowerShell, CIM startup. Contract fixtures run with `--kill-grace 1`, which is not reliably enough to launch either helper on a loaded runner.

The two timeouts compounded: a timed-out `taskkill` is reported as "could not prove process-tree cleanup", which raises out of the wave loop; the finally-block cleanup re-enters with the leader already dead, falls to the descendant probe, which times out on the same one-second budget and takes its "cannot prove absence -> assume the worst" branch. Phantom descendants, red wave.

## Change

> **Rebased onto main after #2055.** That PR gave the PowerShell descendant probe its own budget (`WINDOWS_DESCENDANT_PROBE_SECONDS`, retried), which supersedes the probe half of this change. This PR is now narrowed to the `taskkill.exe` helper, which main still times with `--kill-grace`.

- `scripts/run-test-wave.py`: the `taskkill.exe /T /F` invocation gets its own floor, `max(kill_grace, WINDOWS_HELPER_TIMEOUT_SECONDS=30)`. The descendant probe keeps the budget it got in #2055. `kill_grace` still governs every actual death wait, so nothing fails open — the timeout-race contract still refuses with rc=2 over a genuinely surviving descendant, now because PowerShell answered rather than because it timed out. The production path already passed `--kill-grace 15` and is unchanged in behaviour.
- `tests/test_parallel_harness_contract.sh`: the scheduler's refusal budget (a flat 120s on main) now reads the taskkill floor and the probe budget out of the scheduler on Windows rather than restating them — hard-coding turns a slow runner into a harness failure the moment the numbers drift. The refusal path can spend both twice (wave loop, then cleanup re-entry), so it allows both passes plus interpreter startup. POSIX is unchanged at 8s: the refusal there is signal-driven.

The two halves ship together because the contract imports `WINDOWS_HELPER_TIMEOUT_SECONDS` (and the probe constants) from the scheduler.

## Validation

`tests/test_parallel_harness_contract.sh` passes locally on darwin arm64. The Windows behaviour it guards is exercised by the Windows shard in CI.


DeusData:
Thank you for splitting the Windows process-control fix into a focused PR and for documenting the difference between helper startup time and process death grace. The contribution queue is quite full, so the full review may take a little time. We have the PR routed and will come back with grounded feedback as soon as possible.

DeusData:
Approved, and this one is worth more than its 39 lines: it closes **#1963**, which was filed from this queue after the same failure reddened an unrelated pull request.

**Your causal chain is exactly right, and it is the part most people would not have found.** A timed-out `taskkill` reports "could not prove process-tree cleanup", which raises out of the wave loop; the finally-block cleanup then re-enters with the leader already dead, falls through to the descendant probe, times out on the *same* one-second budget, and takes its "cannot prove absence → assume the worst" branch. Two independent timeouts compounding into a phantom-descendant report, while the sibling shard passes. That is why it looked like flakiness rather than a bug.

And the underlying confusion is stated in one sentence: `kill_grace` budgets how long a doomed process may take to die, but the helpers also pay process spawn and, for PowerShell, CIM startup. Those are different quantities that happened to share a variable.

**I verified the safety claim rather than taking it.** On `main`, `kill_grace` bounds six things. Your change touches exactly two — the `windows_descendants` probe at `:170` and the `taskkill` invocation at `:188`, both *helper spawns*. All four genuine death waits are untouched:

```
:196  process.wait(timeout=kill_grace)
:203  process.wait(timeout=kill_grace)
:233  wait_for_group_exit(... + kill_grace)
:239  wait_for_group_exit(... + kill_grace)
```

So the timeout-race contract still refuses with rc=2 over a genuinely surviving descendant — now, as you say, because PowerShell answered rather than because it timed out. Nothing fails open, and the production path at `--kill-grace 15` is behaviourally unchanged.

**Having the contract read the floor out of the scheduler is the right call.** `module.WINDOWS_HELPER_TIMEOUT_SECONDS * 2 + 10` rather than a restated `8` means the two cannot drift into a harness failure that looks like a product failure. Your `* 2` reasoning is sound too — the refusal path really can spend the floor twice, proving descendants and then re-entering cleanup. That is the same class of fix as deriving a loop bound from its own array instead of borrowing an unrelated constant.

Shipping both halves together is correct given the import.

## Status

**Clearance** is `REVIEW(1)` on `scripts/run-test-wave.py` — the established shape for a test-harness script here. It needs a maintainer marker, not a change from you; I will handle it.

This is `CLEAN` and green, so it should move quickly once cleared — it is CI-unblocking, and a fix for something currently reddening other people's work goes ahead of ordinary queue order.

One clarification so the scope stays honest: this addresses `test-windows` **shard 1** and its parallel-scheduler contract. There is a separate cluster on `test-windows-guards` — preconditions failing because the guards job's own setup cannot index an ASCII fixture — which I have seen on four unrelated PRs and which this does not touch. Different failure, also ours.

DeusData:
Thank you for this, and sorry it sat while the harness moved under it. The diagnosis was right, and two of the three things it fixes are now on main through #2055 (`202c8e42`), which I wrote after the same failure reddened yet another unrelated PR: the descendant probe has its own 15 s × 2 budget and names what it could not prove, and the contract test now loads the scheduler with `importlib` and waits on a wide `communicate(timeout=120)`.

One piece of #1963 is still open, and it is the piece only your PR addresses: `scripts/run-test-wave.py` still passes `timeout=kill_grace` to the `taskkill.exe /T /F` spawn (line 221 on main). With the contract's `--kill-grace 1`, a slow helper start raises out of the wave loop and re-enters cleanup with the leader already force-killed — the first half of the compound failure you described. So this PR is worth finishing, and it can still close #1963.

Since main rewrote the surrounding code, a rebase will conflict on exactly the hunks #2055 covered. The smallest resolution is to drop those and keep only the taskkill site, in #2055's shape — this is the whole remaining change (it passes `py_compile`, the guard is red on pristine main and green with the change, and the full contract is green on macOS):

```diff
--- a/scripts/run-test-wave.py
+++ b/scripts/run-test-wave.py
@@ -41,6 +41,14 @@ POLL_SECONDS = 0.05
 WINDOWS_DESCENDANT_PROBE_SECONDS = 15
 WINDOWS_DESCENDANT_PROBE_ATTEMPTS = 2
+# WHY: the same distinction holds for `taskkill.exe /T /F`. --kill-grace bounds
+# how long the tree may take to *die* after taskkill has issued the kills; the
+# taskkill invocation itself pays a helper spawn plus a tree walk, which is a
+# stable-state cost unrelated to the tree's stubbornness. Timing the spawn with
+# a 1s --kill-grace turned a slow helper start into "taskkill could not prove
+# process-tree cleanup", which raises out of the wave loop and lands the finally
+# pass on the descendant probe with the leader already force-killed.
+WINDOWS_TASKKILL_SECONDS = 15
@@ -218,7 +226,7 @@ def terminate_process_tree(active: ActiveSuite, kill_grace: int) -> None:
                 stderr=subprocess.DEVNULL,
-                timeout=kill_grace,
+                timeout=WINDOWS_TASKKILL_SECONDS,
             )
--- a/tests/test_parallel_harness_contract.sh
+++ b/tests/test_parallel_harness_contract.sh
@@ -45,6 +45,23 @@ if [[ "$probe_sites" == *kill_grace* ]]; then
     exit 1
 fi
+# The same holds for the `taskkill.exe /T /F` spawn on a live leader: its own
+# invocation must not be timed by --kill-grace either. Grab the call as a block
+# (from the executable name to the closing paren of subprocess.run) so the
+# check does not depend on which line the timeout keyword lands on.
+taskkill_call=$(awk '/"taskkill.exe"/{grab=1} grab{print} grab && /^ *\)$/{exit}' "$scheduler")
+if [[ -z "$taskkill_call" ]]; then
+    echo "FAIL: could not locate the Windows taskkill invocation in the scheduler" >&2
+    exit 1
+fi
+if [[ "$taskkill_call" == *kill_grace* ]]; then
+    echo "FAIL: the Windows taskkill helper spawn is still timed by --kill-grace" >&2
+    exit 1
+fi
```

Your `max(kill_grace, 30)` shape works just as well if you prefer it — naming is yours. Keep the sign-off on the rebased commit and this merges on green; the Windows leg is the live proof, as before.

PR #1833, opened 2026-08-25
fix: persist daemon watches across client disconnects

LeoLan-Digital:
## Summary

Keep project watches alive in permanent daemon mode after the last client session disconnects.

This allows daemon-owned background file watching and indexing to continue independently of client session lifetime.

## What changed

- Preserve the physical project watch when the application runs in permanent mode
- Keep non-permanent behavior unchanged
- Allow watcher-triggered background indexing without an active client session
- Retain daemon ownership of watcher-triggered jobs until completion
- Add regression coverage for permanent daemon watch persistence

## Tests

Focused daemon application suite:

- 50 passed

Full CLI baseline comparison:

- baseline: 280 passed, 1 failed
- remaining known failure: `tests/test_cli.c:1748`
- the same failure reproduces on the unmodified baseline and is unrelated to this change

`git diff --check` passes.

## Behavior

Non-permanent applications retain the existing session-scoped watch lifecycle.

Permanent daemon mode now treats the physical watch as daemon-owned rather than session-owned, while project pruning and daemon shutdown still remove the watch through their existing lifecycle paths.

### DIFF
--- a/src/daemon/application.c
+++ b/src/daemon/application.c
@@ -149,11 +149,11 @@ struct cbm_daemon_application_job {
     cbm_daemon_application_job_t *next;
 };
 
-/* A watcher-triggered physical job is owned by the exact live sessions that
- * currently subscribe to its project/root watch. The callback waiting for the
- * job is only a storage waiter; it is deliberately not an ownership
- * subscription, so the worker is cancelled when the last matching session
- * disconnects even while unrelated daemon sessions remain alive. */
+/* A watcher-triggered physical job is normally owned by the exact live
+ * sessions subscribed to its project/root watch. In permanent daemon mode the
+ * watcher itself additionally owns the physical job so background indexing can
+ * continue while no client session is connected. The callback remains a
+ * storage waiter in both modes and releases daemon ownership on completion. */
 struct cbm_daemon_application_watch_job_subscription {
     cbm_daemon_application_session_t *session;
     cbm_daemon_application_job_t *job;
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for keeping the change focused on permanent-daemon watch ownership and preserving the non-permanent lifecycle. Session disconnects, watcher ownership, pruning, and background jobs form a sensitive lifecycle boundary, so we need more time to review it carefully. The queue is quite full, but we will come back with grounded feedback as soon as possible.

DeusData:
Reviewed. The change is sound and the design question it turns on is answered well. Two lint items stand between it and merge, and the second one deserves care rather than the obvious fix.

**The ownership shift is the right shape.** Treating the physical watch as daemon-owned rather than session-owned *only* in permanent mode, with non-permanent behaviour untouched, keeps the blast radius to the mode that actually wants it. And you answered the question I would have asked — if the watch outlives the session, who removes it — by leaving project pruning and daemon shutdown as the lifecycle paths that still tear it down. A watch nothing owns is a leak; a watch owned by the daemon in permanent mode is the point.

## The two lint failures

**`clang-format` at `src/daemon/application.c:3390-3392`** — mechanical. `git clang-format` on the changed lines clears it. Note the CI formatter is the Homebrew LLVM build; a standalone `clang-format-20` produces spurious whole-file drift here.

**`cppcheck: Variable 'watcher_owns_job' is assigned a value that is never used [unreadVariable]` at `:3460`** — please do not fix this by deleting the assignment.

The flag itself is load-bearing: it is read at the `if (watcher_owns_job)` immediately above, and that read is what makes the unsubscribe conditional. cppcheck is objecting only to the `watcher_owns_job = false;` that follows the unsubscribe.

The reason to be careful is that this sits inside the `for (;;)` loop that begins around `:3451`. If control can re-enter that block on a later iteration with the flag still `true`, the reset is what stops a second `application_job_unsubscribe_locked(job)` on the same job — and removing it to quiet the linter would introduce a double-unsubscribe rather than remove dead code.

So the question to settle is narrow: **can the loop re-reach that `if` after taking it once?**

- If it cannot — every path through the block breaks or returns — the assignment really is dead and deleting it is correct.
- If it can, cppcheck is wrong here, and the house rule applies: refactor so the intent is visible to the analyser, or keep it with a `// WHY:` comment explaining the re-entry. Never silence a linter by removing a line that is doing work.

You are much better placed to answer that than I am from the diff, since the loop body is mostly context here.

## Otherwise

Your baseline comparison is the right method — `tests/test_cli.c:1748` reproducing on the unmodified base is what makes "unrelated" a measurement rather than an assertion. That failure is environmental and well known here; it depends on the coding agents installed on the machine.

**Please also rebase**: `main` moved three times yesterday — broken by a duplicate-symbol merge, repaired by #1993, then #1703 landed.

Issue #1557, opened 2026-08-12
Kotlin interface nodes get label Class; companion-object members and TS re-export aliases produce no nodes


PsyTech-Rob:
**Environment:** v0.10.2, Windows 11, `--mode full`. Kotlin + TypeScript monorepo.

Three related fidelity gaps that make label/name-based queries lie on Kotlin-heavy codebases:

## 1. Kotlin `interface` gets label `Class`

`interface OrganizacaoRepository` (a plain Kotlin domain port) is stored with label `Class`. All **199** `Interface`-labeled nodes in our graph are TypeScript. Consequence: `search_graph(label="Interface")` on the Kotlin side returns nothing while interfaces exist everywhere — an agent that trusts the label concludes "no interfaces here". (The `INHERITS` edges themselves are fine and correctly link implementations to these `Class`-labeled interfaces.)

## 2. `companion object` members produce no Method node

`TipoLogradouro` declares `create(...)` and `rehydrate(...)` inside `companion object`. Neither exists as a node; `DEFINES_METHOD` for the class omits them, and call sites (`TipoLogradouro.create(...)` from other modules) resolve to a class-level `USAGE` only. Method-level queries ("who calls create?") return zero.

## 3. TS re-export aliases produce no node

`export { diaCivilNo as hojeNaClinica } from './datas'` — searching `hojeNaClinica` yields zero hits with no pointer to `diaCivilNo`, although the alias is what the rest of the codebase imports.

All three are "absence that isn't absence": the graph's answer is empty while the code isn't. A node for the alias/companion member (or at minimum a documented limitation list echoed on empty results) would prevent false negative conclusions.

DeusData:
@PsyTech-Rob, thank you for the concrete graph-level examples. These are three independent extraction contracts and need separate issues so they can be reproduced and fixed without coupling unrelated language paths.

Current `main` supports that split: Kotlin declarations are classified through the generic class-label path, companion objects have their own Kotlin extraction/recovery handling, and TypeScript re-export resolution is a separate import/export path that explicitly fails closed when the exported alias name is unavailable.

Please split this into:

1. Kotlin `interface` declarations labeled as `Class`.
2. Missing methods inside Kotlin companion objects.
3. Missing TypeScript re-export alias nodes.

One small public fixture per issue is enough; link each new issue back here and we will route them independently. We can keep this report open as the index until the split is complete. Thank you again for identifying false absences rather than treating an empty graph answer as proof.

Issue #1556, opened 2026-08-12
TESTS_FILE edges are never created for JVM (Kotlin/JUnit) tests


PsyTech-Rob:
**Environment:** v0.10.2, Windows 11, `--mode full`. Kotlin (JUnit5/Testcontainers) + TypeScript (vitest) monorepo.

## Symptom

`TESTS_FILE` edges are created only for the TypeScript side. Measured on a full index:

- Files matching `.*(Test|IT)\.kt` or `.*\.test\.tsx?`: **283**
- `TESTS_FILE` edges: **92** — all of them `.test.tsx` (66) / `.test.ts` (26)
- JVM tests linked: **0**, including trivially-conventional pairs like `AgendarUseCaseTest.kt` → `AgendarUseCase.kt` (same package, sibling `src/test/kotlin` source set) and `JooqCompromissoRepositoryIT.kt` → `JooqCompromissoRepository.kt`.

## Impact

Any "which tests cover this class?" query over the graph silently answers "none" for the whole JVM half of a polyglot repo. We now instruct agents to fall back to naming convention + `OVERRIDE` edges of test fakes for Kotlin, but the edge type advertises coverage it doesn't have.

## Suggestion

Apply the same filename/convention heuristic used for `*.test.ts(x)` to JVM conventions (`XTest`/`XIT`, `src/test/{kotlin,java}` mirroring `src/main/...`), or document `TESTS_FILE` as TS-only.

DeusData:
Thank you @PsyTech-Rob. Confirmed in `pass_tests.c`: `XTest.kt` is recognized as a test path, but `test_to_prod_path()` only derives production paths for Go, Python, and JavaScript/TypeScript naming conventions. It has no Kotlin/Java source-set mirror and no `XIT.kt` mapping, so JVM `TESTS_FILE` edges cannot be created by the current pass.

Routed priority/high under the parsing and extraction epic #592. The measured TS/JVM split is a useful regression fixture.

PR #1840, opened 2026-08-26
fix(kotlin): resolve member calls through typed properties instead of by name

aarya-brex:
## What does this PR do?

Makes the declared receiver type available to Kotlin call resolution, so a member
call through a constructor-injected property resolves by type instead of by name.
Refs #1555 (partial — see *What this does not fix*).

Two separable commits: the first makes property types available, the second stops
properties from winning the name race for calls.

### Root cause

`push_var_def_qn` never populated `return_type`, so **every** Kotlin property
cross-registered as an untyped placeholder. `kotlin_lookup_method` then had no
receiver type to match, missed, and resolution fell through to the name-only
strategies (`suffix_match`, `unique_name`, `field_type_hint`, `fuzzy`).

Primary-constructor properties were missing entirely: `class Foo(private val bar: Bar)`
declares a member reachable from every instance method, but it sits on the class
line rather than in the class body, so `extract_class_variables` never saw it.
That is precisely the constructor-injected-port shape in #1555.

This matches your own reading on that issue — *"receiver/parameter type and
existing `OVERRIDE` relationships are not consistently part of that final
candidate decision"*. This PR addresses the receiver-type half.

### Why narrowing the candidate pool, not vetoing the winner

`cbm_label_is_registry_symbol` admits `Variable` and `Field` so that reads and
writes can resolve. A consequence is that every property name is also a bare-name
candidate during *call* resolution, so a property can outrank the function a call
actually meant.

My first attempt vetoed the edge when a weak strategy landed on a property. That
is wrong, and measurably so: wherever a property shadows a function name — common
in Kotlin, since a constructor `val` and the method that reads it share
vocabulary — it deletes a genuine edge. On JetBrains/Exposed the veto form cost
**96 real method-target edges** (e.g. `AbstractQuery.comment` 24 → 2 edges,
shadowed by `val comment` in `ColumnDiff`/`ColumnMetadata`; `Query.where` 625 →
598, shadowed by `AbstractQuery.where` and the statement classes' `val where`).

So the split is by question, not by strategy:

- `cbm_label_is_registry_symbol` — *may this declaration be found by name?*
- `cbm_label_is_callable_target` (new) — *may a call bind to it?* (`Function`,
  `Method`, `Constructor`, `Class`)

`cbm_registry_resolve_callable` applies the narrower predicate **before** a winner
is picked, and only to strategies 3/3.5/4 — `import_map` and `same_module` are
untouched, since those already have real evidence. If a weak match lands on a
property, resolution is retried against callable candidates only; the edge is
dropped only if that also finds nothing.

Uncached, for the same reason `cbm_registry_resolve_lineage` is: `_resolve_cache`
is keyed on `callee_name` alone, so it cannot hold two different label policies.
`cbm_pipeline_node_is_callable_target` now delegates rather than repeating the
label list.

### Measurements

`JetBrains/Exposed` (~800 `.kt` files), full index, v0.10.2 baseline at `010569f`:

| | baseline | veto (rejected) | this PR |
|---|---|---|---|
| `lsp_kt_method` edges | 472 | 650 | **650** |
| `CALLS` → `Variable` (fabricated) | 874 | 188 | **188** |
| total `CALLS` rows | 30772 | 29965 | **30378** |
| `AbstractQuery.comment` | 24 | 2 | **24** |
| `Query.where` | 625 | 598 | **625** |
| `substring` / `init` | 33 / 11 | 32 / 6 | **33 / 11** |

Typed resolution up ~38%, name-only fabrications down ~78%, and no genuine edge
lost. I verified retry *targets* against source rather than just counting:
`comment` → `AbstractQuery.comment` (`AbstractQuery.kt:195`), `where` →
`Query.where` (exposed-jdbc `Query.kt:118`).

The 13 remaining method-target drops are all `kotlin.Any.equals` via
`lsp_kt_operator`, and that is an upgrade rather than a loss: baseline put 14
edges on the synthetic `kotlin.Any.equals` placeholder, this PR puts 0 there
while real overrides gain (`Column.equals` 1 → 3, `ResultRow.get` 3 → 4, plus new
`LinkedIdentityHashSet.contains` and `EntityClass.get`). The node exists in both
graphs — receivers are now typed, so `==` binds to the class's own override.

Also run against a large private Kotlin monorepo (~13k nodes) to check the
behaviour holds off the benchmark repo: `lsp_kt_method` 74 → 801, `CALLS` →
`Variable` 739 → 159, zero genuine edges lost.

### Testing

- `scripts/test.sh` (ASan + UBSan, full suite): **8435 assertions, 0 failures**,
  exit 0. The three nonzero-rc suites in the log are
  `test_parallel_harness_contract.sh`'s deliberate hang fixtures.
- `make -f Makefile.cbm test-focused TEST_SUITES=registry`: 60/60.
- New: `TEST(callable_label_predicate_is_narrower_than_registry_membership)`
  asserts the callable set is a strict subset of the registry-admitted set, which
  is the invariant the whole change rests on.
- Renamed `kotlin_suppress_weak_call_to_property` →
  `kotlin_weak_match_is_uncallable`: it now reports a fact ("this match cannot be
  called") rather than prescribing a drop, since the caller re-resolves.

One gap to flag honestly: `make -f Makefile.cbm lint-format` and `lint-no-suppress`
both pass locally, but I could not run `lint-cppcheck` (cppcheck isn't installed
on this machine), so that gate is unverified until CI runs it.

### What this does not fix

- **`OVERRIDE`/`INHERITS`-based disambiguation**, the other suggestion in #1555.
  A call through an interface-typed property still resolves to the interface
  method, not the implementation.
- **Inferred types beyond direct construction.** `val x: T` and `val x = Foo()`
  are handled; a factory or call chain returns NULL and falls through as before.
  Inferring through an expression is the type pass's job, not the extractor's.
- **Java is unaffected** — it already resolves these via `lsp_type_dispatch` at
  0.95. This is Kotlin-only.
- Adjacent and untouched: #1557 (companion-object members produce no nodes;
  `interface` labelled `Class`), #1556, #1732.
- The honest ceiling: `unique_name` + `suffix_match` are still the majority of
  `CALLS` edges on a large Kotlin repo even with this patch. This closes a
  specific typing gap; it does not make the Light Semantic Pass a compiler
  frontend.

One observation while validating: `EVALUATION_PLAN.md` measures precision/recall
for cross-repo (X1) and semantic (S1/S2) edges, but not call-resolution
correctness — so neither the original bug nor the regression my first attempt
introduced would have been caught by the existing harness. The strategy-mix
before/after A/B above is the check I fell back on; a fixture-based version of it
might be worth adding as a gate. Happy to open a separate issue if useful.

## Checklist

- [x] Tests pass locally (`scripts/test.sh`, ASan+UBSan — 8435 asserts, 0 failures)
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) — `lint-format` and
      `lint-no-suppress` pass; `lint-cppcheck` not run locally (cppcheck
      unavailable), relying on CI

### DIFF
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -770,4 +770,15 @@ bool cbm_label_is_relation(const char *label);
 // `label` may be NULL (returns false). Defined in helpers.c.
 bool cbm_label_is_registry_symbol(const char *label);
 
+// True for labels that can be the target of an invocation: Function, Method,
+// Constructor, and Class (a class name in call position is a construction).
+// Registry membership is deliberately wider than this — Variable and Field are
+// admitted so that reads and writes can resolve — which also makes every
+// property name a bare-name candidate during call resolution. Consumers
+// resolving a CALL use this predicate to narrow the candidate pool to
+// declarations that can actually be called, so a property that merely shares a
+// function's name does not win the name race. `label` may be NULL (returns
+// false). Defined in helpers.c.
+bool cbm_label_is_callable_target(const char *label);
+
 #endif // CBM_H
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -195,6 +195,9 @@ static void extract_variables(CBMExtractCtx *ctx, TSNode root, const CBMLangSpec
 static void extract_var_names(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec);
 static void extract_class_variables(CBMExtractCtx *ctx, TSNode class_node, const char *class_qn,
                                     const CBMLangSpec *spec);
+static void push_var_def_typed(CBMExtractCtx *ctx, const char *name, TSNode node,
+                               const char *type_text);
+static const char *resolve_kotlin_var_type(CBMArena *a, TSNode node, const char *source);
 static void extract_rust_impl(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec);
 static void extract_class_methods(CBMExtractCtx *ctx, TSNode class_node, const char *class_qn,
                                   const CBMLangSpec *spec);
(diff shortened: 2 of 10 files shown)


aarya-brex:
Closing this — on further review I found my own patch regresses correct edges, so it is not ready to offer.

Specifically, the extractor commit registers Kotlin primary-constructor properties with a module-level qualified name that omits the parent class. That makes a property a same-module candidate for calls to a same-named function, and in one case a top-level extension function and a constructor `val` collide on the same qualified name so the callable node is lost entirely. My before/after measurements also turned out to be parsed incorrectly, so several figures in the description above are wrong.

Apologies for the noise. The underlying gap in #1555 is real and the typed-property direction did measurably reduce name-only fabrications, so I may return with a smaller, correctly-measured change scoped to the extractor alone — but I did not want to leave an inaccurate PR sitting in the queue in the meantime.

PR #1843, opened 2026-08-26
fix: extract structure from Vue inline scripts

### LINKED ISSUE #1410: Vue SFC `<script setup>` blocks are never parsed — silent empty graph for `.vue` files, causing false-positive dead code
### Version

`codebase-memory-mcp 0.9.0` (Linux x86_64, standalone binary)

### Platform

Linux (x64)

### Install channel

### Binary variant

standard

## Summary

`.vue` files are discovered and registered as `File` and `Module` nodes, and
`get_architecture` reports them under `languages`. But their `<script setup>`
contents are never parsed: **no `Function`/`Method` nodes are created for
functions defined in a `.vue` file, and no `CALLS` edges are created for calls
originating from one.**

The index therefore looks healthy while being structurally empty for every
`.vue` file in the repo.

## Impact

This makes `search_graph(max_degree=0, exclude_entry_points=true)` unsafe on any
Vue project. Every utility called only from components reports `in_degree: 0`
and is presented as dead code.

On a 9.2k-line Vue/TS codebase the dead-code query returned 29 functions.
**6 of the 11 I verified were false positives** — all called from `.vue` files:

| Reported dead | Actual call sites |
| --- | --- |
| `showModal` | 5 |
| `hideModal` | 6 |
| `copyToClipboard` | 2 |
| `showToast` | 1 |
| `hideToast` | 1 |
| `now` | 1 |

Acting on that list deletes working code. `fan_in` in `get_architecture`
hotspots is understated for the same reason (reported `clone` at `fan_in: 3`;
actual is 6).

## Minimal reproduction

Four files. The `.ts` caller and the `.vue` caller do the same thing.

```
repro/
  package.json
  tsconfig.json
  src/util.ts
  src/caller.ts
  src/components/Caller.vue
```

**`src/util.ts`**

```ts
export function fnRelTs(x: number): number { return x + 1; }
export function fnAliasTs(x: number): number { return x + 2; }
export function fnRelVue(x: number): number { return x + 3; }
export function fnAliasVue(x: number): number { return x + 4; }
export default { fnRelTs, fnAliasTs, fnRelVue, fnAliasVue };
```

**`src/caller.ts`**

```ts
import uRel from './util';
import uAlias from '~/util';

export function runTs(): number {
  return uRel.fnRelTs(1) + uAlias.fnAliasTs(2);
}
```

**`src/components/Caller.vue`**

```vue
<script setup lang="ts">
import uRel from '../util';
import uAlias from '~/util';

const runVue = (): number => {
  return uRel.fnRelVue(1) + uAlias.fnAliasVue(2);
};
</script>

<template>
  <div>{{ runVue() }}</div>
</template>
```

**`tsconfig.json`**

```json
{ "compilerOptions": { "target": "ES2022", "module": "ESNext", "moduleResolution": "bundler", "strict": true, "baseUrl": ".", "paths": { "~/*": ["src/*"] } } }
```

**`package.json`**

```json
{ "name": "repro", "version": "1.0.0", "type": "module" }
```

Then:

```
index_repository(repo_path="…/repro", mode="full", name="repro")
query_graph(project="repro",
  query="MATCH (f) WHERE (f:Function OR f:Method) RETURN f.name, f.file_path, f.in_degree ORDER BY f.name")
```

### Actual

```
fnAliasTs   src/util.ts   1
fnAliasVue  src/util.ts   0     ← called from Caller.vue
fnRelTs     src/util.ts   1
fnRelVue    src/util.ts   0     ← called from Caller.vue
runTs       src/caller.ts 0
```

`runVue` is **absent from the graph entirely** — no node is created for it.

```
MATCH (a)-[r:CALLS]->(b) RETURN a.name, a.file_path, b.name
→ runTs, src/caller.ts, fnRelTs
  runTs, src/caller.ts, fnAliasTs
```

Only two edges. Nothing originates from `Caller.vue`.

### Expected

`fnRelVue` and `fnAliasVue` at `in_degree: 1`, a `Function` node for `runVue`
at `src/components/Caller.vue`, and two additional `CALLS` edges.

## What this is *not*

I first suspected namespace default imports (`import u from './util'` then
`u.fn()`) or path-alias resolution (`~/util`). **Both work correctly** — the
2×2 above isolates it: `caller.ts` resolves relative *and* aliased namespace
calls fine. The only failing variable is the caller being a `.vue` file.

## Why it is easy to miss

`get_architecture` reports the files as present, which reads as success:

```json
"languages": [ { "language": "Vue", "file_count": 34 }, … ]
```

Aggregate edge density is the tell — on the same machine:

| Repo | Nodes | Edges | Edges/node |
| --- | --- | --- | --- |
| Python, 3.7k LOC | 409 | 1665 | **4.07** |
| Vue+TS, 9.2k LOC | 480 | 924 | **1.93** |

And on the Vue repo:

```
MATCH (f) WHERE (f:Function OR f:Method) AND f.file_path ENDS WITH '.vue'
RETURN count(f)                                            → 0

MATCH (a)-[r:CALLS]->(b) WHERE a.file_path ENDS WITH '.vue'
RETURN count(r)                                            → 0

MATCH (f:File) WHERE f.path ENDS WITH '.vue' RETURN count(f) → 34
```

34 file nodes, zero structural content.

## Suggested direction

Extract the `<script>` / `<script setup>` block from the SFC before handing it
to the TS parser, preserving a line offset so `file_path`/line attribution stays
correct. Template-block references (`@click="handler"`, `{{ fn() }}`) are a
larger job and could be a follow-up — script-block extraction alone would
recover the bulk of the missing edges.

Two smaller asks, if useful:

1. Until this lands, consider having `index_repository` warn when a language is
   detected but yields zero function nodes — that single signal would have made
   this obvious immediately.
2. `exclude_entry_points=true` in `search_graph(max_degree=0, …)` implies a
   confidence the index cannot support for partially-parsed languages. A
   per-language "structural coverage" figure in `index_status` would let callers
   know when not to trust a dead-code result.

## Environment

- `codebase-memory-mcp` 0.9.0, standalone binary
- Linux 7.0.0-28-generic, x86_64
- Repos: Vue 3 + TypeScript (Vite, `~/*` path alias), and Python 3 for the
  comparison baseline
- `mode="full"` in all runs; `skipped_count: 0`, `expected_nodes` matched
  `nodes` in every index (no reported extraction failures)

### Logs

```text

```

```text

```

_No response_

### Confirmations

DeusData:
## Summary

- Extract definitions and calls from each inline Vue `<script>` / `<script setup>` block while retaining the file's single existing `Module` accumulator.
- Parse each eligible block with a fresh parser/tree and one absolute included range over the complete original file buffer, preserving source coordinates.
- Accept absent/`js`/`javascript` and `ts`/`typescript` languages; skip `src` scripts, unsupported preprocessors, and scriptless files.
- Keep the public definitions wrapper intact, add only an internal module-free entry point, and preserve the single Vue import path.

Fixes #1410.

## Design choice and trade-offs

The selected design uses one fresh parser/tree per inline block with an included range derived solely from the embedded content node. Parsing the complete original buffer keeps byte, row, and column coordinates correct for CRLF, non-ASCII content, and same-line tags. Per-block isolation also prevents a malformed or failed block from leaking partial output into another block; every allocation, language selection, included-range setup, and parse path is checked and cleaned up.

Material alternatives considered:

- **Parse sliced script strings and remap positions afterward:** somewhat simpler parser input, but requires fragile byte/row/column translation and is especially error-prone for CRLF and UTF-8.
- **Reuse one parser/tree across all embedded blocks:** fewer short-lived allocations, but couples discontiguous block state and failure handling and makes duplicate/module behavior harder to prove.

The chosen approach spends a small parser allocation per inline script block in exchange for coordinate correctness, failure isolation, and a simpler no-duplicates contract. Vue files have a bounded number of script blocks, so that trade-off is appropriate here.

## Verification

Reproduction and binding proof:

- Current production failed deterministically three times with zero Vue functions where two were expected.
- Frozen-base GREEN: `extraction` 289/289 and `edge_structural` 33/33.
- Production-only revert restored the exact RED; restoring the implementation returned both suites to GREEN.
- After rebasing onto `909051ccee2d070f33e15fda71bfe61c9076eea6`: `extraction` 296/296 and `edge_structural` 33/33.
- Independent exact-commit review of `531f88fb2ffa190b0a60b189fb0ef90361b6a27e` passed with no blocking findings.

Platform gates:

- **Linux arm64:** full canonical leg green — 7425 passed, 0 failed, 8 skipped; required watchdog/guard steps passed.
- **Windows ARM64:** relevant suites green (`extraction` 296/296, `edge_structural` 33/33; daemon/runtime, CLI, MCP, and pipeline suites also green). The full parallel run recorded 7470 passed, 0 assertion failures, 63 skipped; only the unrelated `complexity` job hit its scheduler timeout, while a focused exact-SHA rerun passed 4/4. Windows guards passed.
- **Windows smoke:** the first canonical run stalled in the allocator probe; the exact-SHA focused shipped-binary probe then passed with the expected allocator-owned result. A canonical retry cleared that probe and passed indexing, query/MCP contracts, install/uninstall/adversarial checks, binary security, process lifecycle, update, download/checksum, and installer E2E. Its final UI phase stalled in the harness's unbounded `curl` readiness call, so the run was stopped at the attributed Windows environment/harness backstop rather than retried again.
- **macOS:** full suites green — 7587 passed, 0 failed, 8 skipped — and the parent watchdog passed. The separate worker watchdog was killed by the environment (`Killed: 9`, rc 126) twice and was parked at its backstop.
- **Linux amd64:** Rosetta/ASan startup and `daemon_runtime` failures reproduced on the exact parent `909051cc`, attributing them to the baseline rather than this change.
- `make -f Makefile.cbm lint-format` passed. Repository-wide lint remains red only on pre-existing diagnostics outside the files changed here.

Tests assert one Vue `Module`, exact function/call counts and identities, original coordinates (including CRLF, non-ASCII, same-line tags), normal plus setup blocks, no duplicates, and exact pipeline `CALLS` identity. Negative controls cover unsupported languages, `src` scripts, and scriptless Vue files; plain TypeScript, Svelte, HTML, Astro, and existing Vue import behavior remain covered.

### DIFF
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -727,6 +727,9 @@ void cbm_channels_push(CBMChannelArray *arr, CBMArena *a, CBMChannel ch);
 // --- Sub-extractor entry points ---
 
 void cbm_extract_definitions(CBMExtractCtx *ctx);
+/* Internal companion for embedded-language trees that contribute definitions
+ * to an existing host-file Module rather than minting a second Module. */
+void cbm_extract_definitions_without_module(CBMExtractCtx *ctx);
 // dbt lineage for Jinja-templated SQL models: emits a Model def plus one usage
 // per ref()/source() call. No-op unless the file parses as SQL and actually
 // contains a dbt builtin call. Defined in extract_dbt.c.
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -7524,6 +7524,19 @@ static void walk_defs(CBMExtractCtx *ctx, TSNode root, const CBMLangSpec *spec, 
     free(s.data);
 }
 
+void cbm_extract_definitions_without_module(CBMExtractCtx *ctx) {
+    const CBMLangSpec *spec = cbm_lang_spec(ctx->language);
+    if (!spec) {
+        return;
+    }
+
+    // Walk AST for function/class definitions
+    walk_defs(ctx, ctx->root, spec, 0);
+
+    // Extract module-level variables
+    extract_variables(ctx, ctx->root, spec);
+}
+
 void cbm_extract_definitions(CBMExtractCtx *ctx) {
     const CBMLangSpec *spec = cbm_lang_spec(ctx->language);
     if (!spec) {
@@ -7545,9 +7558,5 @@ void cbm_extract_definitions(CBMExtractCtx *ctx) {
     mod.is_test = ctx->result->is_test_file;
     cbm_defs_push(&ctx->result->defs, a, mod);
 
-    // Walk AST for function/class definitions
-    walk_defs(ctx, ctx->root, spec, 0);
-
-    // Extract module-level variables
-    extract_variables(ctx, ctx->root, spec);
+    cbm_extract_definitions_without_module(ctx);
 }
--- a/internal/cbm/extract_imports.c
+++ b/internal/cbm/extract_imports.c
@@ -1395,8 +1395,13 @@ static void parse_spec_imports(CBMExtractCtx *ctx) {
 // that the main parser uses.  Adding another host language is a one-line
 // declaration in lang_specs.c.
 
+typedef struct {
+    TSNode script;
+    TSNode content;
+} CBMEmbeddedBlock;
+
 static void embedded_collect_content_nodes(TSNode root, const CBMEmbeddedLangSpec *spec,
-                                           TSNode *out, int *out_count, int max_out) {
+                                           CBMEmbeddedBlock *out, int *out_count, int max_out) {
     /* Iterative DFS so deeply-nested script blocks are still found.  Cap the
      * stack to a sane bound (host grammars do not have million-deep markup
      * trees) — no need to introduce TSNodeStack here. */
@@ -1412,7 +1417,8 @@ static void embedded_collect_content_nodes(TSNode root, const CBMEmbeddedLangSpe
             for (uint32_t k = 0; k < cc; k++) {
                 TSNode c = ts_node_child(node, k);
                 if (strcmp(ts_node_type(c), spec->content_node_type) == 0) {
-                    out[(*out_count)++] = c;
+                    out[*out_count] = (CBMEmbeddedBlock){.script = node, .content = c};
+                    (*out_count)++;
                     if (*out_count >= max_out) {
                         return;
                     }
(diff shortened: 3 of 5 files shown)


DeusData:
Superseded by #1852.

This PR's only commit carried no `Signed-off-by` trailer, so the `dco` check failed while all 33 other checks passed. Rather than rewrite history on the branch, the change was re-committed with a sign-off on a fresh branch.

The content is unchanged — both commits produce tree `befebbbb`, verified by hash, so #1852 is byte-identical apart from the trailer and a fuller commit message.

Nothing was lost; please continue on #1852.

PR #1862, opened 2026-08-27
feat(cli): persist install hook opt-out

### LINKED ISSUE #1857: feat(cli): persistent opt-out for install hooks
## Problem

PR #753 identified measurable per-call hook process-spawn overhead. Maintainer direction there was to keep hooks enabled by default, add an opt-out, and persist the user's choice so later installs and updates respect it.

## Proposed behavior

- `install --no-hooks` disables CBM-managed lifecycle/tool hooks.
- `install --hooks` re-enables them.
- Bare `install` keeps hooks enabled by default and reads the persisted preference.
- The preference is stored as `install_hooks` in the existing CLI config database.
- All current hook-capable install surfaces obey the same policy.
- MCP, skills, instructions, profiles, and other non-hook integration remain unchanged.
- Plan and dry-run reflect the effective policy without persisting preview flags.
- Existing foreign/user hooks remain untouched.

This follows the maintainer-approved direction in #753.

Andy11-cpu:
Fixes #1857

Supersedes #753.

## What

- Keeps install hooks enabled by default.
- Adds persistent `install --no-hooks` and `install --hooks`.
- Applies the policy across current hook-capable installer surfaces.
- Preserves MCP, skills, instructions, profiles, and foreign hook state.
- Keeps plan/dry-run non-mutating.

## Why

#753 established the performance cost of per-call hook process spawning. This implements the maintainer-approved direction from that discussion: retain graph-first defaults while giving users a persistent opt-out.

## Verification

- `scripts/test.sh`
- `scripts/lint.sh --ci CLANG_FORMAT=clang-format-20`
- `scripts/ci/lint-mem.sh clang-tidy-22`
- `make -f Makefile.cbm security`
- agent-config smoke
- focused hook-policy tests
- `git diff --check`

All pass locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,8 @@ The `install` command automatically strips macOS quarantine attributes and ad-ho
 
 The `install` command auto-detects installed coding agents and configures their documented MCP entries plus durable instructions, skills, and lifecycle hooks where supported.
 
+Hooks are enabled by default. Run `codebase-memory-mcp install --no-hooks` to remove CBM-managed hooks while keeping MCP registration and other integration files; the choice persists for later installs and updates. Run `codebase-memory-mcp install --hooks` to re-enable them, or inspect the current preference with `codebase-memory-mcp config get install_hooks`.
+
 ### Session Coordination Daemon
 
 CBM automatically shares one per-account coordination daemon across Claude Code, Codex, OpenCode, and every other configured client. There is no opt-in setting for MCP servers or hook clients: the first daemon-backed CBM session starts it, each session registers its own work, and the final session shuts it down. The daemon owns long-lived background services such as watchers, shared indexing jobs, and the optional UI. Closing one session cancels work owned only by that session, while work still needed by another session continues.
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -79,6 +79,7 @@ codebase-memory-mcp config set auto_index true
 codebase-memory-mcp config set auto_index_limit 50000
 codebase-memory-mcp config set watcher_enabled false
 codebase-memory-mcp config reset auto_index
+codebase-memory-mcp config get install_hooks
 ```
 
 Current keys:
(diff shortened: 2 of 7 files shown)


DeusData:
Thank you for implementing the previously discussed persistent hook opt-out and for preserving foreign hook state, dry-run behavior, and the remaining installer surfaces. The contribution queue is currently quite full, so the full review may take a little time. We have the PR routed and will come back with grounded feedback as soon as possible.

DeusData:
Reviewed. The direction is settled — this implements the approved outcome from #753 — so this is about one build break, and it is a small one.

## The Windows failure is real, and it is a missing platform guard

Both `test-windows` shards, `pr-smoke (windows-latest)` and `test-windows-guards` fail for the same reason: the tree does not compile on Windows.

```
src/cli/cli.c:8846: error: call to undeclared function 'cbm_remove_gemini_coverage_hook'
                    did you mean 'cbm_remove_augment_coverage_hook'?
```

The compiler's suggestion is a red herring. `cbm_remove_gemini_coverage_hook` does exist — at `src/cli/cli.c:5769` — but its definition sits inside `#ifndef _WIN32`, which opens a few lines above it. On Windows it is compiled out entirely.

The existing call site already knows this. At `cli.c:11104`:

```c
#ifndef _WIN32
        if (cbm_remove_gemini_coverage_hook(cp, installed_binary) != CLI_OK) {
            ...
        }
#endif
```

Your new call is not wrapped. Adding the same `#ifndef _WIN32` / `#endif` pair around it should be the whole fix.

**Why only Windows caught it:** clang treats an implicit function declaration as an error, and the CLANG64 legs are the ones building with clang. On POSIX the function is present, so every other leg compiles cleanly and says nothing. Worth knowing, because a green Linux and macOS run tells you nothing about this class of mistake.

**One knock-on to check while you are there:** `cbm_upsert_gemini_coverage_hook` lives under the same guard at `:5753`. If the persistent-policy work added an upsert call anywhere, it needs the same treatment — the compiler only reports the first error.

## The rest

The shape is right: hooks stay enabled by default, `--no-hooks` / `--hooks` persist the choice, plan and dry-run stay non-mutating, and foreign hook state along with MCP, skills, instructions and profiles is preserved. Keeping the graph-first default while giving a persistent opt-out is what #753 concluded, and doing it as a persisted policy rather than a per-invocation flag is what actually solves the reported problem.

**Clearance** is `REVIEW(1)` on `scripts/smoke-test.sh` — the established shape here. It needs a maintainer marker, not a change from you.

**Also please rebase** once the guard is in: `main` moved three times yesterday (broken by a duplicate-symbol merge, repaired by #1993, then #1703 landed).

PR #407, opened 2026-06-02
fix(mcp): exit stdio server on parent death

nvt-pankajsharma:
Fixes #406

## Summary
- Add a POSIX parent-process watchdog for the default stdio MCP server path.
- Reuse the existing shutdown path from signals and the watchdog so orphaned servers close stdin, stop active work, and exit.
- Add a regression test that kills a wrapper parent and verifies the child MCP server exits instead of remaining orphaned.

## Scope
- POSIX parent-death detection via `getppid()` polling.
- Windows is left unchanged; the watchdog regression skips Windows-like shells.
- This PR is intentionally limited to the parent-exit issue to keep one issue per PR.

## Validation
- `make -f Makefile.cbm cbm && bash tests/test_parent_watchdog.sh`
- `scripts/test.sh`
- `scripts/lint.sh --ci`
- `scripts/build.sh CC=cc CXX=c++`
- `make -f Makefile.cbm security`
- `git diff --cached --check`

Note: the security target prints stale absolute paths during vendored integrity scanning, but it exits successfully with `Vendored integrity check passed` and `All security checks passed`.

## Local hook note
`git commit` without `--no-verify` ran the repository pre-commit hook and failed in unrelated existing `clang-tidy` findings outside this PR's files, including `internal/cbm/cbm.c`, `src/store/store.c`, `src/watcher/watcher.c`, and `src/pipeline/pass_lsp_cross.h`. The contribution checklist commands above passed on this branch.

Issue #185, opened 2026-03-30
codebase-memory-mcp.exe running after quitting VS Code.


testsubject8383:
In Win10 task manager I have 5 codebase-memory-mcp.exe running. Now I have 6 after I restarted VS Code and loaded a repo.

Using VS Code, OpenCode and OpenChamber GUI, all latest versions.

DeusData:
Thanks for the report. Each MCP client (VS Code, OpenCode, OpenChamber) launches its own `codebase-memory-mcp` stdio server, so a few concurrent processes is expected — but they should each exit when their client closes the stdio pipe. Lingering instances mean a process isn't detecting that its parent went away.

Looking at the server lifecycle: on Windows the event loop relies on stdin reaching EOF (the client closing the pipe) to terminate. That works for a **graceful** client shutdown, but if a client is force-killed or crashes and leaves the pipe write-handle dangling, the server can be orphaned — there's currently no parent-process-death watchdog as a backstop.

A fair amount of MCP-loop and Windows work has landed since this was filed (v0.7.0): MCP `ping`, a reworked stdin poll path, and Windows-specific I/O fixes. Could you retest on **v0.7.0** and let me know:

1. Does the problem still occur on v0.7.0?
2. When you see orphans, was the client closed **gracefully** (normal quit) or force-killed/crashed?
3. Do the orphaned processes eventually exit on their own (e.g. after a minute), or do they stay forever?
4. Which client's processes linger — is it one specific client or all three?

That'll tell me whether to add a Windows parent-death watchdog or whether it's a specific client not closing its pipe. Thanks!

DeusData:
Closing for inactivity — no response to the retest request. The MCP loop got `ping`, a reworked stdin poll path, and Windows I/O fixes in v0.7.0; current is **v0.8.1**.

If you still see orphaned `codebase-memory-mcp.exe` processes after quitting VS Code on v0.8.1, comment with the answers to the four questions above (graceful quit vs force-kill, do they exit eventually, which client) and I'll reopen right away. Thanks!

PR #1874, opened 2026-08-28
fix(main): exit Windows stdio servers when the spawning client dies (#914)

### LINKED ISSUE #914: Windows v0.8.1: stale graph + Permission denied on delete_project after orphan cbm process (reconfirm #277)

LazyXuan:
Fixes #914 (also the lingering-server half of the #185 report; eases the WAL bloat in #1083).

## Problem

When an MCP stdio client on Windows (ZCode, Codex, OpenCode, VS Code, ...) is force-killed, the `codebase-memory-mcp.exe` it spawned survives as an orphan, blocked on stdin forever.

The parent-death watchdog from #407 was built POSIX-only, with the comment *"Windows is unaffected (job objects handle this)"*. That assumption only covers half the process tree:

- the `KILL_ON_JOB_CLOSE` job in `subprocess.c` wraps processes **CBM spawns itself** (index workers);
- an MCP stdio server is spawned **BY the client**, as the client's child — and Windows, unlike POSIX, never propagates a parent's termination to its children.

So the orphan lingers, holding SQLite WAL read locks that block checkpoints (#1083) and surface as `delete_project` permission errors (the #914 symptom chain).

## Fix

The Windows branch of the watchdog waits on a handle instead of polling a ppid:

- At startup (same point POSIX captures `getppid()`), the client resolves its parent PID via a Toolhelp snapshot and opens a `SYNCHRONIZE` handle to it.
- The watchdog thread waits on that handle with a 500 ms timeout (the timeout only re-checks `g_shutdown`, mirroring the POSIX poll cadence). The kernel signals a process handle exactly once, on termination, and the held handle pins the process object — so PID reuse cannot fool the wait, unlike re-reading a ppid.
- A signaled parent takes the same deliberate `_exit(0)` as POSIX: with the owning client gone, kernel handle reclamation is the only trustworthy release for the daemon connection, file locks, and the WAL read lock an orphan would otherwise pin.
- No trustworthy parent signal at startup (snapshot failed, reserved PID, parent already gone, handle unopenable) → keep running and lean on the stdin EOF path, exactly the POSIX `initial_ppid <= 1` fail-open choice; a thread-creation failure stays fail-closed.
- Workers are untouched: they run inside CBM's own kill-on-close job, which is the containment the old comment assumed everyone had.

This only affects CBM's own lifecycle: the watchdog terminates this process when its own parent dies. It never touches, signals, or enumerates any other process beyond reading the parent PID once.

## Test

`tests/test_parent_watchdog.sh` now runs on MSYS2 instead of skipping. The Windows arm differs from the POSIX arm for measured reasons, each noted in the script:

- stdin is an **anonymous pipe** (an MSYS FIFO is not readable by native binaries — the server blocks forever on it), with a writer helper that holds the write end open, so the observed exit can only come from the watchdog, never an EOF;
- the kill resolves the child's **Windows-physical parent** (the pipeline subshell — the watchdog watches the Toolhelp `ParentProcessId`, which is not the wrapper script process) and `TerminateProcess`es exactly that one, because MSYS `kill -9` does not reliably terminate the Windows process behind an MSYS pid.

Discrimination proof (local Windows run of the same test on the same tree): the pre-fix binary leaves the server running after the parent dies; this build exits within one watchdog tick. `clang-format` and `cppcheck` are clean on the touched region; the full suite is left to CI (local Git Bash has known spawn-isolation flakiness in daemon suites).

## Notes

- Scope deliberately mirrors POSIX semantics, including the "parent already dead at startup" fail-open (POSIX cannot distinguish that from ppid==1 either); the stdin EOF path covers it.
- A shell-launched server still watches the shell, not the ultimate client — the same limitation the POSIX ppid poll has.


DeusData:
Approved. The diagnosis corrects a documented assumption, and the mechanism you chose is better than the one it mirrors.

**The half-truth in the #407 comment is the crux.** *"Windows is unaffected (job objects handle this)"* is true for processes **CBM spawns** — those sit in its own `KILL_ON_JOB_CLOSE` job. It is simply not true for an MCP stdio server, which is spawned **by the client**, as the client's child, and Windows does not propagate a parent's termination downward. Splitting the process tree into those two halves is what makes this obviously a real gap rather than a theoretical one, and the symptom chain you trace — orphan holds a SQLite WAL read lock, checkpoints blocked (#1083), `delete_project` surfaces a permission error (#914) — explains why it looked like three unrelated bugs.

**Waiting on a handle rather than polling a ppid is the right primitive, for the reason you give.** An open `SYNCHRONIZE` handle pins the process object, so the PID cannot be recycled underneath you; the kernel signals it exactly once, on termination. A ppid poll can be fooled by reuse and has to re-read state it does not own. That is the same category of improvement as using a Job Object for tree membership — picking the primitive that makes the race impossible instead of narrowing the window.

**The fail-open/fail-closed asymmetry is right and I checked it.** No trustworthy parent signal — snapshot failure, reserved PID, parent already gone, `OpenProcess` refusing — returns success and leaves the process running on the stdin EOF path, which is exactly POSIX's `initial_ppid <= 1` choice. A thread-creation failure returns false. Those are the correct directions: an unknowable parent should not kill a working server, but a watchdog that cannot start should not be reported as started.

**And the test earns its result.** The writer holding the pipe's write end open means stdin never sees EOF, so the observed exit **can only** come from the watchdog. Without that, a passing test would be equally consistent with an EOF exit and would prove nothing. Resolving the child's Windows-physical parent and calling `TerminateProcess` on exactly that one — rather than trusting MSYS `kill -9` — is the same discipline applied to the kill side.

Both deviations from the POSIX arm are explained in the script where a future reader will meet them, which is where they belong.

## Process note

Clearance came back `REVIEW(3)`, all three in `tests/test_parent_watchdog.sh`: `exec sleep 3600`, and two `chmod +x`. I read them — they are the wrapper helper the test writes into its own temp directory from a quoted heredoc, then marks executable. No exemption, no network, nothing touched outside the tmpdir. The scanner is flagging ordinary scaffolding in a shell file, which is what it is supposed to do; I will get the marker written before merge.

**Please rebase** — `main` moved three times today (broken by a duplicate-symbol merge, repaired by #1993, then #1703 landed).

Being explicit that a shell-launched server still watches the shell rather than the ultimate client — the same limitation the POSIX poll has — is the right way to leave a known boundary. Thank you.

DeusData:
Merged as 5e5405c6e656 — thank you! Resolving the parent through Toolhelp, waiting on a SYNCHRONIZE handle with a bounded loop, and failing open when no trustworthy parent handle exists is the right shape for the Windows half of the watchdog, and keeping workers out of it avoids the job-object interplay. Running test_parent_watchdog.sh on MSYS2 too, with the ps -W guard for POSIX cleanup, closes the coverage gap that let #914 linger for so long. This ships in v0.10.9.

PR #1816, opened 2026-08-24
fix(windows): support Unicode repository paths

z2250389:
## What does this PR do?

Fixes #1240.

Windows still passed UTF-8 repository paths to narrow CRT filesystem APIs at three boundaries:

- Git root existence checks used `stat()`, so valid Unicode repositories reported `root_exists=false`.
- The Graph UI browser used `opendir()` / `readdir()`, so Unicode folders were omitted or could not be opened.
- Daemon session-context and background-index admission used `stat()`, so UI indexing failed even after navigation succeeded.

This replaces those calls with the existing UTF-8-aware filesystem helpers (`cbm_is_dir`, `cbm_opendir`, `cbm_readdir`, and `cbm_closedir`) and adds reproduce-first regression tests for Git discovery, UI browsing, MCP session context, and UI indexing.

## Validation

- ASan/UBSan targeted suites: `daemon_application httpd git_context` — 113 passed, 6 Windows platform skips
- Security audit: passed
- Graph UI production build: 738 modules built; 0 npm vulnerabilities
- Windows UI smoke test: browsed `C:/Users/〈Path to folder〉/道具箱/〈repo〉`, indexed successfully, and detected branch `main`
- Live graph result: 4,560 nodes / 23,863 edges

The full Windows cppcheck sweep was timeboxed after 35 minutes; it reported four pre-existing findings on untouched lines and no finding on the changed lines.

## Checklist

- [x] Every commit is signed off (`git commit -s`) — required, CI rejects unsigned commits ([DCO](../DCO), see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [ ] Tests pass locally (`make -f Makefile.cbm test`) — targeted affected suites pass; the full Windows suite has existing platform failures
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) — diff-scoped checks pass; full Windows cppcheck reports existing findings and exceeded the 35-minute local timebox
- [x] New behavior is covered by a test (reproduce-first for bug fixes)

PR #1876, opened 2026-08-28
Fix Unicode daemon session roots on Windows

ertankucukoglu:
## Summary

On Windows, canonicalization returns UTF-8 paths, but daemon validation passed them to narrow `stat()`. Non-ASCII working directories were therefore rejected before MCP initialization.

- Validate session and background-index roots through the UTF-8 filesystem API.
- Cover Unicode working-directory startup and non-ASCII path variants.

## Tests

- `scripts/test.sh --suites daemon_application SANITIZE=` — 50 passed
- Production build and MCP initialize from a Unicode working directory
- Latin, Cyrillic, CJK, and Greek variants matched the ASCII baseline

### DIFF
--- a/src/daemon/application.c
+++ b/src/daemon/application.c
@@ -374,6 +374,11 @@ static bool application_regular_db_exists(const char *project) {
     return stat(path, &status) == 0 && S_ISREG(status.st_mode);
 }
 
+static bool application_canonical_directory_exists(const char *path) {
+    cbm_path_info_t info = {0};
+    return cbm_path_info_utf8(path, &info) == 0 && info.is_directory;
+}
+
 static cbm_daemon_application_watch_t *application_find_watch_locked(
     cbm_daemon_application_t *application, const char *project) {
     for (cbm_daemon_application_watch_t *watch = application->watches; watch; watch = watch->next) {
@@ -2354,9 +2359,7 @@ static cbm_daemon_runtime_application_status_t application_set_context(
     if (canonical && allowed_present) {
         canonical = cbm_canonical_path(allowed, canonical_allowed, sizeof(canonical_allowed));
     }
-    struct stat root_status;
-    canonical =
-        canonical && stat(canonical_root, &root_status) == 0 && S_ISDIR(root_status.st_mode);
+    canonical = canonical && application_canonical_directory_exists(canonical_root);
     bool set =
         canonical && cbm_mcp_server_set_session_context(session->mcp, canonical_root,
                                                         allowed_present ? canonical_allowed : NULL);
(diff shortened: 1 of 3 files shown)


DeusData:
Reviewed and cleared. The fix is right and it belongs to a bug class this repo already knows.

Windows canonicalization hands back UTF-8, and passing that to a narrow `stat()` is the same shape as passing it to a narrow `fopen()` — which is why this repo has `cbm_fopen()`, and why SQLite opens on user paths go through their own normalization. A non-ASCII working directory being rejected *before MCP initialization* is a particularly unhelpful place for it to bite, because there is no session yet to report the error into.

Covering Latin, Cyrillic, CJK and Greek against an ASCII baseline is the right breadth — one non-ASCII case can pass by luck of the code page.

## Two things, one yours and one not

**Yours: the commit has no sign-off.** I checked with git's trailer parser rather than a text search, because a blank line before a trailer can hide it from git while `grep` still finds it. Commit `31066f8c` has none at all:

```
git push --force-with-lease
```

**Not yours: `test / test-windows-guards`.** Your own new test passed:

```
=== tests\windows\test_non_ascii_path.py ===
GREEN (tests\windows\test_non_ascii_path.py)
GREEN (tests\windows\test_non_ascii_cache_dump.py)
```

What failed is a cluster of *preconditions* in unrelated cases — `test_hook_augment.py` ("index did not run"), `test_ui_drive_listing.py` ("HTTP server did not start… Is this a UI build?"), and `test_cli_non_ascii_arg.py` ("ASCII control did not index via CLI"). The UI one is plainly a build-configuration precondition with nothing to do with Unicode paths.

I have seen that exact cluster today on an unrelated Swift extraction PR as well, so it is a property of that lane rather than of either change. It is ours to sort out, not yours, and I am recording it as such.

**Also please rebase** — `main` moved three times today (broken by a duplicate-symbol merge, repaired by #1993, then #1703 landed), and this is currently showing as conflicting.

Sign-off plus a rebase and this is ready. Thanks for finding it at the right layer rather than special-casing the path.

DeusData:
Merged as `2a0e9b9b`. Thank you for the sign-off and the rebase — the substance was approved a while ago and everything since was procedural on our side.

This is the same class as #1787 and #1816: Windows Unicode repository paths breaking because a narrow CRT call sits where a UTF-8-aware one belongs. Session-context roots were one more place it reached, and the daemon is a particularly bad one to lose them in, because the failure appears as "your project is not indexed" rather than as a path error.

Before merging I confirmed the 22-commit gap to `main` was inert: no commit had touched `src/daemon/application.c`, `tests/test_daemon_application.c` or `tests/windows/test_non_ascii_path.py` since your base, so the green described exactly what landed.

That is your second merge today, after #1836.

PR #1877, opened 2026-08-28
fix(confidence): stop an unreadable confidence from reading as 0.0

### LINKED ISSUE #1980: An unreadable confidence value is recorded as 0.0, and displaces a clean stored blob
### Version

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

`strtod` answers `0.0` for text it cannot read. Two places treat that answer as a real
confidence, so a property blob carrying a malformed value — `"confidence":null`, an empty
string, anything non-numeric — is recorded as a confidence of **zero**. Zero is a
meaningful value in both places, so nothing looks wrong.

The concrete harm is in `src/graph_buffer/graph_buffer.c`. `edge_props_confidence` answers
`CBM_EDGE_CONF_ABSENT` (`-1`) for an edge carrying no confidence, so that any real
confidence outranks it. The merge in `edge_props_should_replace` is a plain comparison:

```c
if (inc != cur) {
    return inc > cur;
}
```

A malformed value answers `0.0`, and `0.0` beats `-1`. So a blob whose confidence the code
cannot even read **displaces a clean stored blob** that simply carried no confidence at
all — and the stored `"strategy"` goes with it. The function's own doc comment already
promises the opposite behaviour.

I expected an unreadable confidence to be treated as absent, exactly as a missing one is,
so it never outranks stored data.

### Reproduction

1. Index any project so at least one edge carries a `confidence` property.
2. Present a second property blob for the same edge whose confidence is malformed — for
   example `{"confidence":null,"strategy":"x"}`, or an empty string, or `"high"`.
3. Result: the malformed blob wins the merge and replaces the stored one, taking the
   stored `strategy` with it. Expected: the malformed blob is treated as carrying no
   confidence, so it does not displace stored data.

The same `strtod` assumption appears in a second place, so a fix should cover both call
sites rather than the merge alone.

### Logs

```text
No error or warning is emitted — the malformed value is accepted as 0.0.
```

```text
```

Reproduces at any scale.

### Confirmations

CaptainMittens:
## What this fixes

`strtod` answers `0.0` for text it cannot read. Two places treat the number
it returns as a real confidence, so a property blob carrying a malformed
value — `"confidence":null`, an empty string, anything non-numeric — came
back as a recorded confidence of **zero**. Zero is a meaningful value in
both places, so nothing looked wrong.

### 1. `src/graph_buffer/graph_buffer.c` — the wrong blob wins a merge

`edge_props_confidence` answers `CBM_EDGE_CONF_ABSENT` (`-1`) for an edge
that carries no confidence, so that any real confidence outranks it. The
merge in `edge_props_should_replace` is a plain comparison:

```c
if (inc != cur) {
    return inc > cur;
}
```

A malformed value answered `0.0`, which beats `-1`. So a blob whose
confidence the code cannot even read **displaced a clean stored blob** that
simply carried no confidence at all. The stored `"strategy"` went with it.

The function's own doc comment already promised the correct behaviour:

```c
/* Read "confidence":<double> out of an edge property blob. Absent/unparseable
 * reads as -1 ... */
```

Only the absent half was true.

### 2. `src/mcp/mcp.c` — "not recorded" published as `0.00`

`bfs_edge_evidence_for_hop` sets the confidence to `-1.0` first, and the
emitter publishes any value of `0` or more as a recorded number
(`%.2f` in the text renderer, a real in the JSON renderer) and `-` / `null`
below zero. A malformed value therefore printed as `0.00`.

That is the one number the surrounding code works to keep meaningful. A
caller reading `0.00` cannot tell "the resolver was certain this call is
wrong" from "nobody wrote a number here".

## The fix

Both sites now pass an end pointer to `strtod` and keep the absent sentinel
when the pointer never moved — nothing was read. This is the shape
`src/store/store.c:402` already uses.

```c
const char *value = p + sizeof(conf_key) - SKIP_ONE;
char *end = NULL;
double conf = strtod(value, &end);
if (end == value) {
    return CBM_EDGE_CONF_ABSENT;
}
return conf;
```

Only the unreadable case changes. A real `0.0` still reads as `0.0`
everywhere, and every existing confidence test passes untouched.

## Tests

Two tests come with the change. Both were **seen failing before the fix and
passing after** — not written after the fact.

| Test | File |
|---|---|
| `gbuf_edge_props_unreadable_confidence_does_not_displace_absent` | `tests/test_graph_buffer.c` |
| `tool_trace_path_unreadable_confidence_reports_not_recorded` | `tests/test_mcp.c` |

Each carries positive controls, so a later failure points at the confidence
and not at a broken request. The mcp test asserts the hop and its readable
`strategy` class still come through, then asserts the output holds no
`0.00`.

Red, before the fix:

```
258 passed, 2 failed, 6 skipped
  FAIL tests/test_mcp.c:3398: strstr(ev_txt, "0.00") is not NULL
  FAIL tests/test_graph_buffer.c:261: ASSERT(strstr(edges[0]->properties_json, "\"strategy\":\"lsp\"") != NULL)
```

Green, after the fix:

```
260 passed, 0 failed, 6 skipped   (EXIT=0)
```

## Checks run

| Command | Result |
|---|---|
| `make -f Makefile.cbm test-focused TEST_SUITES="graph_buffer mcp"` | `260 passed, 0 failed, 6 skipped` — exit 0 |
| `make -f Makefile.cbm test` | `7633 passed, 2 failed, 8 skipped` |

The two full-suite failures are in `tests/test_cli.c` (lines 1749 and 6725)
and reproduce on a clean tree without this change. Both print
`error: one or more agent cleanup operations failed`, so they depend on the
coding agents installed on the machine rather than on anything here.

## How this was found

By scanning for siblings of the parse bug fixed in #1875 — the same shape
of "a parse reports success while the input stays unread". These two are
the confidence-sentinel pair from that scan. Four more candidates remain
and will come as separate pull requests.

## Checklist

Fixes #1980

### DIFF
--- a/src/graph_buffer/graph_buffer.c
+++ b/src/graph_buffer/graph_buffer.c
@@ -1045,7 +1045,18 @@ static double edge_props_confidence(const char *props_json) {
     if (!p) {
         return CBM_EDGE_CONF_ABSENT;
     }
-    return strtod(p + sizeof(conf_key) - SKIP_ONE, NULL);
+    /* strtod answers 0.0 for text it cannot read, and 0.0 is a real
+     * confidence that beats CBM_EDGE_CONF_ABSENT. So a blob carrying
+     * "confidence":null used to outrank a clean blob that carries no
+     * confidence at all, and displace it. Ask strtod where it stopped: an
+     * end pointer that never moved means it read nothing. */
+    const char *value = p + sizeof(conf_key) - SKIP_ONE;
+    char *end = NULL;
+    double conf = strtod(value, &end);
+    if (end == value) {
+        return CBM_EDGE_CONF_ABSENT;
+    }
+    return conf;
 }
 
 /* Decide whether an incoming property blob replaces the stored one on a
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -6498,7 +6498,16 @@ static bool bfs_edge_evidence_for_hop(cbm_traverse_result_t *tr, int64_t hop_nod
         if (conf) {
             const char *colon = strchr(conf, ':');
             if (colon) {
-                *confidence_out = strtod(colon + 1, NULL);
+                /* strtod answers 0.0 for text it cannot read, and the caller
+                 * publishes any value >= 0 as a recorded confidence. So a
+                 * malformed value used to print as 0.00 — the one number the
+                 * surrounding code works to keep meaningful. Keep the -1 when
+                 * the end pointer never moved: nothing was read. */
+                char *end = NULL;
+                double parsed = strtod(colon + 1, &end);
+                if (end != colon + 1) {
+                    *confidence_out = parsed;
+                }
             }
         }
         return true;
--- a/tests/test_graph_buffer.c
+++ b/tests/test_graph_buffer.c
@@ -236,6 +236,34 @@ TEST(gbuf_edge_props_merge_prefers_higher_confidence) {
     PASS();
 }
 
+/* A confidence the code cannot read is not evidence of anything, so it must
+ * not outrank an edge that simply carries no confidence at all.
+ *
+ * edge_props_confidence answers -1 for "absent" so that any real confidence
+ * beats it. strtod answers 0.0 for text it cannot read, so an unreadable
+ * value used to come back as a real confidence of zero -- which beats -1 and
+ * displaced the stored blob. The function's own comment already promised
+ * that "absent/unparseable reads as -1"; only the absent half was true. */
+TEST(gbuf_edge_props_unreadable_confidence_does_not_displace_absent) {
+    const char *no_conf = "{\"callee\":\"f\",\"strategy\":\"lsp\"}";
+    const char *bad_conf = "{\"callee\":\"f\",\"confidence\":null,\"strategy\":\"registry\"}";
+
+    cbm_gbuf_t *gb = cbm_gbuf_new("test", "/tmp");
+    int64_t a = cbm_gbuf_upsert_node(gb, "Function", "a", "pkg.a", "f.go", 1, 5, "{}");
+    int64_t b = cbm_gbuf_upsert_node(gb, "Function", "b", "pkg.b", "f.go", 6, 10, "{}");
+    cbm_gbuf_insert_edge(gb, a, b, "CALLS", no_conf);
+    cbm_gbuf_insert_edge(gb, a, b, "CALLS", bad_conf); /* unreadable, arrives last */
+
+    const cbm_gbuf_edge_t **edges = NULL;
+    int count = 0;
+    cbm_gbuf_find_edges_by_type(gb, "CALLS", &edges, &count);
+    ASSERT_EQ(count, 1);
+    ASSERT_TRUE(strstr(edges[0]->properties_json, "\"strategy\":\"lsp\"") != NULL);
+
+    cbm_gbuf_free(gb);
+    PASS();
+}
+
 /* An empty incoming blob must never displace real stored properties. */
 TEST(gbuf_edge_props_merge_keeps_existing_on_empty) {
     const char *lsp = "{\"callee\":\"f\",\"confidence\":0.95,\"strategy\":\"lsp\"}";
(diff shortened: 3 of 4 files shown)


CaptainMittens:
Context that is not visible from this PR on its own: it is one of four from a single scan.

After #1875 I went looking for siblings of its shape — a parse reports success while the input stays unread. The scan found six. Four are filed: #1875 (the cypher parser), this one (`strtod`), #1880 (`atoi`/`atol` on environment settings), and #1881 (`strtoll` on span timestamps).

They make the same argument as #1922: an input the code cannot read becomes a plausible value, and nothing downstream can tell it from a real one. The silence is the harm.

Here that value is `0.0`, and it is worse than a wrong number. `0.0` outranks the `-1` absent sentinel in `edge_props_should_replace`, so a blob whose confidence could not be read **displaced a clean stored blob** that simply carried none — and took the stored `strategy` with it. The function's own doc comment already promised the right behaviour; only the absent half of it was true.

Two more candidates from that scan are not filed yet. Would you rather they came as their own PRs, or folded into one of these?

DeusData:
Approved. The strongest thing in this PR is that the code disagreed with its own documented contract, and you found it by reading the comment.

**I checked it and it is exact.** `graph_buffer.c:1036`:

> Read "confidence":\<double\> out of an edge property blob. **Absent/unparseable** reads as -1 so any edge that carries a confidence outranks one that does not.

`CBM_EDGE_CONF_ABSENT` is `-1.0`, and only the absent half was ever implemented. "Only the absent half was true" is precisely right.

**And the consequence is graph corruption, not just a wrong number.** `strtod` answers `0.0`, `0.0 > -1.0`, so `edge_props_should_replace` concludes the incoming blob outranks the stored one — an edge whose confidence *cannot be read* displaces a clean edge that simply had none, and the stored `"strategy"` goes with it. That is a silent quality loss in the graph, and nothing about it looks wrong afterwards because zero is a legitimate confidence.

**The MCP half is the honesty half.** `0.00` published for an unreadable value is indistinguishable from "the resolver was certain this call is wrong", which is exactly the distinction `-1` / `null` / `-` exists to preserve. Reusing the `store.c:402` end-pointer shape rather than inventing a third convention is right.

**Only the unreadable case changes**, a real `0.0` still reads as `0.0`, and the existing confidence tests pass untouched — that scoping is what makes this safe to take.

**Your tests were seen red first, with positive controls.** Asserting the hop and its readable `strategy` still come through *before* asserting the output holds no `0.00` means a later failure points at the confidence rather than at a broken request. That is the shape that keeps a regression test useful in two years.

## One thing to do

**This is now `DIRTY`, and that is my doing rather than yours.** #1703 merged an hour ago and also touches `src/mcp/mcp.c`. A rebase onto current `main` should be mechanical — the two changes are in different functions. `main` also went briefly broken and was repaired by #1993 earlier today, so rebase now rather than onto anything older.

Once it is clean I will re-run and merge. Nothing else outstanding.

Working the class from #1875 rather than the individual bug — confidence here, timestamps in #1881, environment settings in #1880 — is why these read as one audit instead of three coincidences.

CaptainMittens:
Rebased onto current `main` (`5fbab7bb`) — now `62b44519`. `MERGEABLE` again.

One conflict, and it was the adjacent-line kind: `#1703` added `RUN_TEST(tool_trace_path_evidence_columns_match_header_issue1542)` on the line this branch adds `RUN_TEST(tool_trace_path_unreadable_confidence_reports_not_recorded)`. Both registrations belong, so both are kept, yours first. I confirmed each of the two has a `TEST()` definition and a `RUN_TEST()` registration afterwards, rather than assuming the resolution was right.

`src/mcp/mcp.c` auto-merged — the two changes are in different functions, as you predicted. I read the merged hunk to confirm the end-pointer check landed intact.

Built and ran the affected suites on the rebased tree: `mcp` and `graph_buffer` together, **303 passed, 4 skipped, 0 failed**.

Thank you for checking the `graph_buffer.c:1036` comment against the code rather than taking my word for it. The contract being written down and only half-implemented is what made this findable.

DeusData:
Merged as `7910de59`. Thank you — and sorry for the delay; the conflict you had to resolve was caused by our own #1703 merge, not by anything in your change.

The defect is a good one to have caught: an unreadable confidence reading as `0.0` is indistinguishable, downstream, from a genuine zero-confidence result. Every consumer that treats low confidence as "weak evidence" would quietly treat a parse failure as strong evidence of weakness. A sentinel that cannot be confused with a real value is the only way to keep those two cases apart.

Before merging I verified the combined tree, since #1836 landed in the same batch and also touches `src/mcp/mcp.c` and `tests/test_mcp.c` — `main` plus all three, `mcp cypher graph_buffer`, **493 passed, 0 failed**.

That is your third merge today, after #2008 and #1986.

PR #1880, opened 2026-08-28
fix(env): stop an unreadable environment setting from picking a wrong default

### LINKED ISSUE #1981: A malformed CBM_* environment value silently picks a wrong setting at three sites
### Version

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

`atoi` and `atol` answer `0` for text they cannot read, and `0` is a real setting at three
places in this project. So a typo, a trailing unit such as `30s`, or a stray space
silently chooses a value nobody asked for, and nothing on screen says the setting was
dropped.

| Site | Setting | What a typo does |
|:--|:--|:--|
| `src/cli/hook_augment.c` | `CBM_HOOK_DEADLINE_MS` | returns the **shortest** deadline the setting allows |
| `src/mcp/mcp.c` | `CBM_INDEX_MAX_RESTARTS` | keeps 100, and so does an explicit `=0` |
| `src/mcp/index_supervisor.c` | `CBM_INDEX_WORKER_TIMEOUT_S` | keeps the 15-minute default |

The first is the sharpest. The code reads:

```c
int v = atoi(env);
if (v < HA_DEADLINE_MIN_MS) {
    return HA_DEADLINE_MIN_MS;
}
```

`atoi` answers `0`, `0` is below `HA_DEADLINE_MIN_MS`, and the clamp hands back **50 ms**
— for a setting whose only purpose is to give the hook *more* room. The comment directly
above that function records a hunt for hook runs that never finished, 0 of 24 real
sessions, which is the exact symptom a silently-shortened deadline produces.

I expected an unreadable setting to be rejected with a message naming the variable, or at
minimum to fall back to the documented default rather than to the opposite extreme.

### Reproduction

1. Set a deliberately malformed value, e.g. `export CBM_HOOK_DEADLINE_MS=30s` (a plausible
   typo — the variable name ends in `_MS`, so a unit suffix reads as natural).
2. Run the hook path that consumes it.
3. Result: the deadline becomes 50 ms, the shortest the setting allows, with no message.
   Expected: either an error naming `CBM_HOOK_DEADLINE_MS`, or the documented default.

The other two sites reproduce the same way with `CBM_INDEX_MAX_RESTARTS` and
`CBM_INDEX_WORKER_TIMEOUT_S`. Note that at `CBM_INDEX_MAX_RESTARTS` an explicit `=0` and a
typo are indistinguishable today, so a user cannot switch restarts off.

### Logs

```text
No message is emitted for a rejected value — that is the defect.
```

```text
```

Reproduces at any scale.

### Confirmations

CaptainMittens:
## What this fixes

`atoi` and `atol` answer `0` for text they cannot read, and `0` is a real
setting at three places in this project. So a typo, a trailing unit such as
`30s`, or a stray space silently chose a value nobody asked for — and nothing
on screen said the setting had been dropped.

| Site | Setting | What a typo did |
|---|---|---|
| `src/cli/hook_augment.c` | `CBM_HOOK_DEADLINE_MS` | returned the **shortest** deadline the setting allows |
| `src/mcp/mcp.c` | `CBM_INDEX_MAX_RESTARTS` | kept 100, and so did `=0` |
| `src/mcp/index_supervisor.c` | `CBM_INDEX_WORKER_TIMEOUT_S` | kept the 15-minute default |

### 1. `CBM_HOOK_DEADLINE_MS` — a typo bought the shortest possible deadline

```c
int v = atoi(env);
if (v < HA_DEADLINE_MIN_MS) {
    return HA_DEADLINE_MIN_MS;
}
```

`atoi` answers `0`, `0` is below `HA_DEADLINE_MIN_MS`, and the clamp handed
back **50 ms** — for a setting whose only purpose is to give the hook more
room. The comment directly above that function records a hunt for hook runs
that never finished, **0 of 24 real sessions**, which is the exact symptom a
silently-shortened deadline produces.

### 2. `CBM_INDEX_MAX_RESTARTS` — the setting did the opposite of the request

```c
int v = atoi(cap_env);
if (v > 0) {
    cap = v;
}
```

Two ways to lose. A typo kept the default of 100. And
`CBM_INDEX_MAX_RESTARTS=0` — which reads as "do not restart" to anybody who
sets it — also kept 100.

### 3. `CBM_INDEX_WORKER_TIMEOUT_S` — a 30-second test hung for 15 minutes

The comment above this override says it exists to tighten the timeout for
tests. `atol` answered `0` for `30s` or a stray space, `0` failed the `> 0`
test, and the 15-minute default came back with nothing to explain it.

## The fix

One helper, not three copies of the same check:

```c
bool cbm_env_long(const char *name, long *out);
```

It answers true only when the variable is set, is not empty, and reads cleanly
from its first character to its last. It holds **no policy** — no minimum, no
maximum, no default — because the three sites disagree on all three, and a
helper that guessed would be wrong at two of them.

The shape is the one `src/main.c:1104` already uses for `--port=`: an end
pointer, `errno`, and a check that nothing was left over. It adds one thing on
top — a refusal of a leading blank, because `strtol` steps over blanks of its
own accord, so `" 5"` would otherwise read as the number 5. The `cli` suite
caught that during this work.

Each site keeps its own rule:

| Site | New behaviour on text it cannot read |
|---|---|
| worker timeout | keeps the 15-minute default AND logs the value it dropped |
| restart cap | `0` now means no restarts; unreadable keeps 100 AND logs the dropped value |
| hook deadline | yields `HA_DEADLINE_DEFAULT_MS` (2000), never the floor |

`hook_augment.c` stays silent on purpose: it includes no log header and writes
no stderr, because its output is hook protocol.

The restart-cap parse was lifted out of a very large function into a named
`index_restart_cap()`, so it can be read and reached on its own.

## Tests

Four tests. The two that pin the user-visible behaviour were **seen failing
before the fix**:

```
cli_hook_deadline_ignores_an_unreadable_value   unreadable value "abc" gave 50 ms
  FAIL tests/test_cli.c:402: ms == 50, expected HOOK_DEADLINE_DEFAULT == 2000
cli_index_restart_cap_honours_zero_and_refuses_junk
  FAIL tests/test_cli.c:434: cbm_index_restart_cap_for_testing() == 100, expected 0 == 0
```

Each carries a positive control first — a good value is still read — so a later
failure points at the unreadable case and not at a reader broken outright.

| Test | File |
|---|---|
| `platform_env_long_reads_a_clean_number` | `tests/test_platform.c` |
| `platform_env_long_refuses_what_it_cannot_read` | `tests/test_platform.c` |
| `cli_hook_deadline_ignores_an_unreadable_value` | `tests/test_cli.c` |
| `cli_index_restart_cap_honours_zero_and_refuses_junk` | `tests/test_cli.c` |

Two `_for_testing` wrappers reach the statics, following the convention already
used 25 times in `src/cli/cli.h`. The deadline wrapper is POSIX-only, matching
the `#ifndef _WIN32` block the function lives in.

## Checks run

| Command | Result |
|---|---|
| `make -f Makefile.cbm test-focused TEST_SUITES="platform cli mcp"` | `518 passed, 2 failed, 6 skipped` |
| `make -f Makefile.cbm test` | `7635 passed, 2 failed, 8 skipped` |

The two failures are in `tests/test_cli.c` (lines 1826 and 6802), print
`error: one or more agent cleanup operations failed`, and reproduce on a clean
tree without this change. They depend on the coding agents installed on the
machine.

## How this was found

By scanning for siblings of the parse bug fixed in #1875 — "a parse reports
success while the input stays unread". #1877 fixed the confidence pair from the
same scan. This is the environment-variable group.

## Checklist

Fixes #1981

### DIFF
--- a/src/cli/cli.h
+++ b/src/cli/cli.h
@@ -525,6 +525,13 @@ char *cbm_hook_augment_lifecycle_json_for(const char *input, const char *forced_
 /* Thin daemon frontend support: preserve the hook's bounded stdin read and
  * hard fail-open deadline without constructing a local MCP/store instance. */
 void cbm_hook_augment_arm_deadline(void);
+
+/* The in-process deadline in milliseconds, as CBM_HOOK_DEADLINE_MS resolves it.
+ * Exposed so a test can check what an unreadable value falls back to. POSIX
+ * only: the Windows path arms a fixed timer and reads no environment value. */
+#ifndef _WIN32
+int cbm_hook_augment_deadline_ms_for_testing(void);
+#endif
 char *cbm_hook_augment_read_stdin(void);
 /* Pure no-op gate for the hook-client fast path (see hook_augment.c). */
 bool cbm_hook_augment_input_is_noop_bash(const char *input);
--- a/src/cli/hook_augment.c
+++ b/src/cli/hook_augment.c
@@ -20,6 +20,7 @@
 #include "foundation/compat_fs.h"
 #include "foundation/constants.h"
 #include "foundation/mem.h"
+#include "foundation/platform.h"
 #include "mcp/mcp.h"
 #include "pipeline/pipeline.h"
 #include "yyjson/yyjson.h"
@@ -70,18 +71,21 @@
  * hook "timeout" remains the outer backstop (and alone governs Windows,
  * where this whole in-process deadline block is compiled out). */
 static int ha_deadline_ms(void) {
-    const char *env = getenv("CBM_HOOK_DEADLINE_MS");
-    if (!env || !env[0]) {
+    /* A value this reader cannot read gets the DEFAULT, never the floor. atoi
+     * used to answer 0 for a typo, 0 is below the minimum, and the clamp then
+     * handed back the shortest deadline the setting allows — the opposite of
+     * what somebody raising CBM_HOOK_DEADLINE_MS is asking for. */
+    long v = 0;
+    if (!cbm_env_long("CBM_HOOK_DEADLINE_MS", &v)) {
         return HA_DEADLINE_DEFAULT_MS;
     }
-    int v = atoi(env);
     if (v < HA_DEADLINE_MIN_MS) {
         return HA_DEADLINE_MIN_MS;
     }
     if (v > HA_DEADLINE_MAX_MS) {
         return HA_DEADLINE_MAX_MS;
     }
-    return v;
+    return (int)v;
 }
 
 static int g_ha_crumb_fd = -1;
@@ -121,6 +125,10 @@ static void ha_open_crumb_log(int deadline_ms) {
                      "CBM_HOOK_DEADLINE_MS)\n",
                      deadline_ms, (long)getpid());
     g_ha_crumb_len = (n > 0 && n < (int)sizeof(g_ha_crumb_msg)) ? (size_t)n : 0;
+}
+
+int cbm_hook_augment_deadline_ms_for_testing(void) {
+    return ha_deadline_ms();
 }
 
 void cbm_hook_augment_arm_deadline(void) {
--- a/src/foundation/platform.c
+++ b/src/foundation/platform.c
@@ -8,6 +8,8 @@
 #include "foundation/compat.h"
 #include "foundation/constants.h"
 #include "foundation/platform_internal.h"
+#include <ctype.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <stdint.h>
 #include <stdio.h>
@@ -433,6 +435,33 @@ const char *cbm_safe_getenv(const char *name, char *buf, size_t buf_sz, const ch
         return platform_copy_environment_value(buf, buf_sz, fallback);
     }
     return NULL;
+}
+
+/* See platform.h. The shape here is the one src/main.c:1104 already uses for
+ * --port=: an end pointer says where the read stopped, errno catches a number
+ * too large, and *end == '\0' catches anything left over. */
+bool cbm_env_long(const char *name, long *out) {
+    if (!out) {
+        return false;
+    }
+    char raw[CBM_SZ_64] = {0};
+    if (!cbm_safe_getenv(name, raw, sizeof(raw), NULL) || !raw[0]) {
+        return false;
+    }
+    /* strtol skips leading blanks of its own accord, so " 5" would read as 5.
+     * A blank in front of a setting is a slip, not a number, so refuse it here
+     * rather than let strtol quietly step over it. */
+    if (isspace((unsigned char)raw[0])) {
+        return false;
+    }
+    char *end = NULL;
+    errno = 0;
+    long value = strtol(raw, &end, CBM_DECIMAL_BASE);
+    if (errno != 0 || !end || end == raw || *end != '\0') {
+        return false;
+    }
+    *out = value;
+    return true;
 }
 
 /* ── Home directory (cross-platform) ───────────────────── */
--- a/src/foundation/platform.h
+++ b/src/foundation/platform.h
@@ -121,6 +121,18 @@ int cbm_default_worker_count(bool initial);
  * Returns NULL when the variable is unset and fallback is NULL. */
 const char *cbm_safe_getenv(const char *name, char *buf, size_t buf_sz, const char *fallback);
 
+/* Read an environment variable as a whole number.
+ *
+ * Answers true only when the variable is set, is not empty, and reads cleanly
+ * from its first character to its last. Anything else — a typo, a trailing
+ * unit such as "30s", a leading or trailing space, or a number too large for a
+ * long — answers false and leaves *out untouched, so the caller picks its own
+ * fallback and can say that it did.
+ *
+ * This exists because atoi and atol answer 0 for text they cannot read, and 0
+ * is a real setting at every call site in this project. */
+bool cbm_env_long(const char *name, long *out);
+
 /* ── Home directory ─────────────────────────────────────────────── */
 
 /* Cross-platform home directory: tries HOME first, then USERPROFILE (Windows).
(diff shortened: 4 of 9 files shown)


CaptainMittens:
Same shape as #1922: an input the code cannot read becomes a plausible value, and nothing downstream can tell it from a real one.

Here it is `atoi` and `atol` answering `0`, and `0` is a real setting at all three sites. The one worth a second look is `CBM_HOOK_DEADLINE_MS`. An unreadable value reads as `0`, falls below the floor, and the clamp hands back **50 ms** — the shortest deadline the setting allows, for a setting whose only purpose is to give the hook more room. The comment above that function records a hunt for hook runs that never finished, 0 of 24 real sessions, which is the symptom a silently-shortened deadline produces.

`CBM_INDEX_MAX_RESTARTS=0` has the same flavour: it reads as "do not restart" to anyone who sets it, and kept the default of 100.

This is one of four from a single scan for siblings of #1875's shape — #1875, #1877, #1880, #1881. Context for the set, and a question about the two still unfiled, is on #1877.

DeusData:
Thank you for the detailed reproduce-first analysis and for keeping policy at the individual call sites instead of hiding it in the shared parser. This changes hook deadlines, worker timeout handling, and restart policy across the CLI and MCP paths, so we need more time to review compatibility and failure behavior before making a decision. No additional information is requested from you at this stage.

DeusData:
Approved. Three real bugs, and the first one is nastier than the table conveys.

**I checked your headline evidence and it is exact.** The comment above `ha_deadline_ms` says augmentation "never appeared in real sessions (0/24 observed)" because the original 300 ms budget self-terminated on cold starts. And then:

```c
int v = atoi(env);
if (v < HA_DEADLINE_MIN_MS) { return HA_DEADLINE_MIN_MS; }   /* 50 */
```

So a typo in the one setting that exists to make that budget *more* generous hands back **50 ms** — six times shorter than the 300 ms already documented as too short. Someone raising the deadline because their hook never fires would make it strictly worse and see nothing.

**`cbm_env_long` holding no policy is the right call and I verified it does.** No minimum, no maximum, no default — it only answers whether the text read cleanly. You are right that a helper which guessed would be wrong at two of the three sites; the worker timeout keeps 15 minutes, the restart cap keeps 100, and the hook deadline yields 2000, and those are genuinely three different answers.

**And `CBM_INDEX_MAX_RESTARTS=0` now meaning no restarts** is the fix I would have missed. A typo keeping the default is bad; an explicit `0` keeping 100 is the setting doing the opposite of what it says.

**One thing I chased and it comes out clean.** `cbm_env_long` reads into a 64-byte buffer, so I checked what happens to a longer value: `platform_copy_environment_value` returns NULL and empties the buffer when `length >= buf_sz` rather than truncating, so `cbm_safe_getenv` fails, `cbm_env_long` returns false, and the site keeps its documented default. Fail-closed, no silent truncation into a wrong-but-parseable number. Worth stating because the alternative would have been a real hole.

**Smaller things done right:** the leading-blank refusal (`strtol` steps over blanks, so `" 5"` would read as 5) with the note that the `cli` suite caught it; positive controls first in each test so a later failure points at the unreadable case rather than a reader broken outright; `hook_augment.c` staying silent on purpose because its output is hook protocol; and lifting the restart-cap parse into a named `index_restart_cap()` so it can be reached at all. Your `_for_testing` count is right too — cli.h has exactly 25.

**Status:** cleared, reviewed, approved. `main` moved twice today (broken, repaired by #1993, then #1703 merged), so I will update this branch and re-run before merging. Nothing needed from you.

Working the bug class from #1875 rather than the bug — confidence pair in #1877, timestamps in #1881, environment here — is why these three read as one piece of work instead of three coincidences.

CaptainMittens:
Rebased onto `c38aa35c` and pushed as `6b801eec`. The clean merge-tree simulation held — one commit replayed, no conflicts.

Posting the local test result too, because it looks alarming and is not. If you run this branch's suite yourself you will see failures, and I would rather you had the comparison up front than spend time on it.

On this machine, same worktree, same `TEST_SUITES="mcp cli"`:

| Tree | Result |
|:--|:--|
| `origin/main` at `c38aa35c` | 536 passed, 6 failed |
| this branch at `6b801eec` | 542 passed, 2 failed |

The branch fails fewer than the base it sits on, and its two are a strict subset of main's six. They are the same two test functions — this branch adds 81 lines to `tests/test_cli.c`, which shifts both sites by 77:

| Test | on main | on branch |
|:--|:--|:--|
| `cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index` | `1749` | `1826` |
| `cli_agent_client_registry_routes_plan_install_and_uninstall` | `6725` | `6802` |

Main's other four are `tests/test_mcp.c:12521` and `:12601` (index recovery, child killed by signal 9 on the alarm), plus `tests/test_cli.c:9390` and `:9491`.

Two machine conditions explain them, and neither exists on your runners:

**OpenCode is installed here.** The failing runs print `Detected agents: OpenCode …` from inside the test's own fixture. `cbm_find_cli` searches `/usr/local/bin` and `/opt/homebrew/bin`, so agent detection escapes the fixture's `HOME` and `PATH`. Three of main's six print that line — `6725`, `9390` and `9491`.

**No permanent daemon is running.** The agent-client install/uninstall and cohort-drain assertions cannot drain a cohort cleanly without one, which is why the failing set is not even stable between runs on the same tree.

I have opened #2019 for the first of those, which fixes the `6725`/`6802` case only. Its description says plainly that `9390` and `9491` are the same class and untouched — the right fix for each depends on what its own assertion means, and I did not want to guess that on your behalf.

Nothing needed from you on this thread. #1880 itself is unchanged apart from the rebase.

DeusData:
Merged — thank you! 🙏

Making the three `CBM_*` integer readers strict (a malformed value is now rejected and the documented default is used, instead of `atoi` silently turning `"4x"` into 4 or `"abc"` into 0) is exactly the kind of fix that removes a whole class of "why is my setting ignored?" reports. The shared parser in `platform.c` with tests on both the CLI and the platform layer makes it easy to keep every future env knob on the same path.

Landed on `main` as merge `54a434743ea6`; it will be in the next release.

PR #1881, opened 2026-08-28
fix(traces): stop an unreadable span timestamp from becoming a real duration

### LINKED ISSUE #1982: An unreadable span timestamp becomes a real-looking duration in ingest_traces
### Version

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

`strtoll` answers `0` for text it cannot read, and `cbm_parse_duration` hands that `0`
straight into its subtraction:

```c
int64_t start = strtoll(start_nano, NULL, CBM_DECIMAL_BASE);
int64_t end = strtoll(end_nano, NULL, CBM_DECIMAL_BASE);
return (end > start) ? (end - start) : 0;
```

The obvious harm is a false duration of zero: a span that genuinely took no time and a
span whose timestamps were garbage become the same row.

It is worse than that. An unreadable **start** time reads as `0`, so the span reports the
whole **end** time as its duration. A test pinning this fails on unfixed code like so:

```
FAIL tests/test_traces.c:340: info.duration_ns == 1050000000,
  expected CBM_DURATION_UNKNOWN == -1
```

That is 1.05 seconds of measured time for a span whose start was never read, and nothing
downstream can tell that number from a real measurement.

I expected an unreadable timestamp to produce an explicit unknown rather than a plausible
duration.

### Reproduction

1. Call `ingest_traces` with a span whose `start_nano` is unreadable — an empty string,
   `null`, or any non-numeric text — and a valid `end_nano` of, say, `1050000000`.
2. Read the span back.
3. Result: `duration_ns` is `1050000000`, indistinguishable from a real 1.05 s span.
   Expected: an explicit unknown, so a consumer can tell measured time from missing time.

The zero-duration case reproduces the same way with both timestamps unreadable.

### Logs

```text
No error is emitted — the unreadable timestamp is accepted as 0.
```

```text
```

Reproduces at any scale.

### Confirmations

CaptainMittens:
## What this fixes

`strtoll` answers `0` for text it cannot read, and `cbm_parse_duration` hands
that `0` straight into its subtraction:

```c
int64_t start = strtoll(start_nano, NULL, CBM_DECIMAL_BASE);
int64_t end = strtoll(end_nano, NULL, CBM_DECIMAL_BASE);
return (end > start) ? (end - start) : 0;
```

The obvious harm is a false duration of **zero** — a span that took no time and
a span whose timestamps were garbage become the same row.

**It is worse than that.** An unreadable *start* time reads as `0`, so the span
reports the whole *end* time as its duration. The test that pins this failed
like so before the fix:

```
FAIL tests/test_traces.c:340: info.duration_ns == 1050000000,
  expected CBM_DURATION_UNKNOWN == -1
```

That is 1.05 seconds of measured time for a span whose start was never read.
Nothing downstream can tell that number from a real one.

## The fix

`cbm_parse_duration` does not move. It is public, it is declared in
`traces.h:66`, and eight tests pin its answers — `0` for a NULL argument, `0`
for an end at or before the start. All of that stays exactly as it is.

A companion carries the extra answer instead:

```c
/* Same, and also says whether both timestamps could be read. */
int64_t cbm_parse_duration_checked(const char *start_nano, const char *end_nano, bool *ok);
```

Both reads go through one small reader in the `src/main.c:1104` shape — an end
pointer, `errno`, and a check that nothing was left over — plus a refusal of a
leading blank, which `strtoll` would otherwise step over.

`cbm_extract_http_info` now uses the companion and writes `CBM_DURATION_UNKNOWN`
(`-1`) when the timestamps do not read. It still returns `true`: the method and
path on that span are still good data, and dropping them would punish them for
a fault they had no part in.

`-1` as "not recorded" is the sentinel this codebase already uses for the same
question — `CBM_EDGE_CONF_ABSENT` in `src/graph_buffer/graph_buffer.c`, which is
what #1877 is about. The `duration_ns` field comment in `traces.h` now says so,
so a reader who meets `-1` has something to read.

## Scope note

`cbm_extract_http_info` has **no production caller** on this tree. `rg` finds it
only in `src/traces/traces.c`, `src/traces/traces.h`, and `tests/test_traces.c`,
and `handle_ingest_traces` in `src/mcp/mcp.c` never touches a timestamp or a
duration. So this is a defect in a public, tested function rather than one
putting bad rows into a graph today. Worth stating plainly rather than
overselling the impact.

## Tests

Two tests. The one that pins the behaviour was **seen failing before the fix** —
the `FAIL` line above is from that run.

| Test | Covers |
|---|---|
| `traces_parse_duration_checked_reports_unreadable_timestamps` | `abc`, empty, `" 100"`, `100ns`, `1e9`, and NULL in either position; plus the positive controls that a good pair, an equal pair, and an end-before-start pair all read fine |
| `traces_extract_http_info_marks_unreadable_duration` | the span keeps its method and path and reports `CBM_DURATION_UNKNOWN` |

The eight existing `cbm_parse_duration` tests pass untouched.

## Checks run

| Command | Result |
|---|---|
| `make -f Makefile.cbm test-focused TEST_SUITES="traces"` | `32 passed, 0 failed` — exit 0 |
| `make -f Makefile.cbm test` | `7633 passed, 2 failed, 8 skipped` |

The two failures are in `tests/test_cli.c` (lines 1749 and 6725), print
`error: one or more agent cleanup operations failed`, and reproduce on a clean
tree without this change.

## How this was found

By scanning for siblings of the parse bug fixed in #1875 — "a parse reports
success while the input stays unread". #1877 fixed the confidence pair and the
companion PR to this one fixes the environment-variable group.

## Checklist

Fixes #1982

### DIFF
--- a/src/traces/traces.c
+++ b/src/traces/traces.c
@@ -6,6 +6,8 @@
 #include "foundation/constants.h"
 
 enum { TRACE_PATH_SLASHES = 3, TRACE_NOT_FOUND = -1 };
+#include <ctype.h>
+#include <errno.h>
 #include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
@@ -63,6 +65,38 @@ const char *cbm_extract_path_from_url(const char *url, char *buf, size_t buf_sz)
 
 /* ── parseDuration ───────────────────────────────────────────────── */
 
+/* Read one nanosecond timestamp. Answers false for text that does not read
+ * cleanly from its first character to its last, the way src/main.c:1104 does
+ * it: an end pointer says where the read stopped, errno catches a number too
+ * large, and *end == '\0' catches anything left over. A leading blank is
+ * refused too, because strtoll would otherwise step over it. */
+static bool trace_read_nano(const char *text, int64_t *out) {
+    if (!text || !text[0] || isspace((unsigned char)text[0])) {
+        return false;
+    }
+    char *end = NULL;
+    errno = 0;
+    long long value = strtoll(text, &end, CBM_DECIMAL_BASE);
+    if (errno != 0 || !end || end == text || *end != '\0') {
+        return false;
+    }
+    *out = (int64_t)value;
+    return true;
+}
+
+int64_t cbm_parse_duration_checked(const char *start_nano, const char *end_nano, bool *ok) {
+    int64_t start = 0;
+    int64_t end = 0;
+    bool read_both = trace_read_nano(start_nano, &start) && trace_read_nano(end_nano, &end);
+    if (ok) {
+        *ok = read_both;
+    }
+    if (!read_both) {
+        return 0;
+    }
+    return (end > start) ? (end - start) : 0;
+}
+
 int64_t cbm_parse_duration(const char *start_nano, const char *end_nano) {
     if (!start_nano || !end_nano) {
         return 0;
@@ -115,7 +149,14 @@ bool cbm_extract_http_info(const cbm_trace_span_t *span, const char *service_nam
         return false;
     }
 
-    out->duration_ns = cbm_parse_duration(span->start_time, span->end_time);
+    /* A timestamp nobody can read is not a measurement. It used to read as 0,
+     * so an unreadable START time reported the whole end time as the duration
+     * -- a made-up number that looks like a real one. The HTTP method and path
+     * on this span are still good, so the span is still returned and only the
+     * duration says it is missing. */
+    bool timed = false;
+    int64_t duration = cbm_parse_duration_checked(span->start_time, span->end_time, &timed);
+    out->duration_ns = timed ? duration : CBM_DURATION_UNKNOWN;
     return true;
 }
 
--- a/src/traces/traces.h
+++ b/src/traces/traces.h
@@ -43,6 +43,8 @@ typedef struct {
     const char *path;
     const char *status_code;
     int span_kind;
+    /* Nanoseconds, or CBM_DURATION_UNKNOWN when the span's timestamps could
+     * not be read. A duration of 0 means the span really took no time. */
     int64_t duration_ns;
 } cbm_http_span_info_t;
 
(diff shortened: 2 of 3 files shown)


CaptainMittens:
Same shape as #1922: an input the code cannot read becomes a plausible value, and nothing downstream can tell it from a real one.

Here `strtoll` answers `0` for a timestamp it cannot read. The obvious harm is a false duration of zero. The real one is an unreadable *start*: it reads as `0`, so the span reports the whole end time as its duration. The test that pins this measured **1.05 seconds** for a span whose start was never read.

The scope note in the description is deliberate — `cbm_extract_http_info` has no production caller on this tree, so this is a defect in a public, tested function rather than one putting bad rows into a graph today.

The `-1` it writes is the sentinel this codebase already uses for the same question, `CBM_EDGE_CONF_ABSENT` — which is what #1877 is about.

One of four from a single scan for siblings of #1875. Context for the set is on #1877.

DeusData:
Thank you for the careful sentinel analysis and for explicitly documenting that `cbm_extract_http_info` has no production caller on the current tree. This still changes a public trace helper and its duration semantics, so we need more time to review the API and downstream compatibility before making a decision. No additional information is requested from you at this stage.

DeusData:
Approved. This is a good fix and an unusually honest PR description.

**The bug is worse than the headline, and you led with that.** A false zero is bad; an unreadable *start* silently reporting the entire end timestamp as the duration is worse, because 1.05 seconds looks exactly like a real measurement. `expected CBM_DURATION_UNKNOWN == -1, got 1050000000` is the right thing to put at the top.

**Not moving the public function was the right call, and I verified it holds.** The diff deletes three real lines — the `duration_ns = cbm_parse_duration(...)` call site and a header comment. `cbm_parse_duration` itself is untouched, so its eight pinned answers (`0` for NULL, `0` for end-at-or-before-start) keep their contract exactly. A companion with an `ok` out-param is the correct shape when the existing return value has no room left to say "unreadable".

**Reusing `-1` rather than inventing a sentinel** is right too, and citing `CBM_EDGE_CONF_ABSENT` as the precedent — plus putting it in the `duration_ns` field comment so a reader who meets `-1` has somewhere to go — is what makes it a convention instead of a magic number.

**Returning `true` while marking the duration unknown** is the right trade. The method and path on that span are still good data and dropping them would punish them for a fault they had no part in.

**And I checked your scope note rather than taking it.** `cbm_extract_http_info` really does appear only in `src/traces/traces.c`, `src/traces/traces.h` and `tests/test_traces.c` — no production caller, exactly as you said. Deflating your own PR's impact in the description is rare and it makes the rest of it easier to trust.

One small thing worth noting for the record, not a change request: refusing a leading blank is a tightening relative to `strtoll`, which would step over it. It is the right choice for a nanosecond timestamp field, and it only applies inside the new checked path, so no existing caller's contract moves.

**Status:** cleared, reviewed, approved. One correction to your PR's timeline — the last check on this head completed on 30 August, so this green predates both today's `main` breakage and its repair (#1993). I will update the branch against current `main` and re-run before merging. Nothing needed from you.

Finding this by scanning for siblings of #1875 — "a parse reports success while the input stays unread" — is a good way to work a bug class rather than a bug.

DeusData:
Merged as 9d0c6a1f — thank you! A tidy, well-bounded fix: an unreadable span timestamp now degrades to "no duration" instead of manufacturing a real-looking one, and the test pins exactly that contract. Closes #1982 with it. Much appreciated, as always.

PR #1883, opened 2026-08-28
feat(pipeline): index-time importance scoring (weighted-degree)

DeusData:
Distilled in-house from #879, with `Co-authored-by:` credit to @petercoxphoto. The design — index-time weighted-degree importance — is theirs and was accepted on 2026-08-20. This is a fresh implementation against current main, because the original no longer applied (3 of 6 files conflicting).

## The blocker that had to be fixed first

The original `name_distinct_file_count` was **cubic in the size of a same-name group**. The kernel hides this — its worst C name is `main` at k≈1064 — but Java does not. Measured on `perf-bench/java` (elasticsearch), one `-O2` binary, cold full index, back to back, node count identical at 693,254 in every leg:

| leg | total index time |
|---|---|
| pass disabled | 85 s |
| **linear (shipped)** | **80 s** |
| cubic (original) | **483 s** |

**≈6.0x.** The shipped version is free within run-to-run noise.

The fix computes the distinct-file count **once per distinct name** (memoized) and deduplicates paths through a hash set instead of pairwise `strcmp`, taking the work from Σk³ per group to **Σk = O(N)**.

## Three traps closed, each proved by revert

1. **`PREDUMP_PASS_COUNT` is now structural** — `enum { PREDUMP_PASS_COUNT = (int)(sizeof(passes)/sizeof(passes[0])) }`. Main had already grown a 7th pass; a hand-written `= 7` beside an 8-entry table would have left importance — registered last — **silently never running, with every test green**. Restoring `= 7` produces 3 failures.
2. **The property append is idempotent.** Rehydrated nodes already carry `"importance"`, so an append-only write yields duplicate JSON keys. Forcing append-only prints the real corruption: `...,"importance":1.000000,"importance":1.000000}`. The key is located in **key position** (preceded by `{`/`,`, followed by `:`) rather than by `strstr`, which would also match text inside a string value.
3. **A new *gated* complexity counter**, not a recorded one. The existing gates structurally cannot observe this pass, and the `g_lsp_tail_*` precedent is explicitly "recorded, deliberately NOT gated". Measured `imp_name_visits` ratio: **1.92 linear** vs **3.96 cubic**, with the gate proved to fire. Two non-vacuity floors mean a skipped pass fails rather than passing on zeroed counters. Ratios only, never wall time.

## No reindex

`CBM_INDEX_FORMAT_VERSION` is unchanged — `src/store/store.h` is not in this diff at all. `importance` is a JSON key inside the **existing** `properties_json` column: no schema change, and legacy indexes simply lack the key, so consumers must tolerate its absence.

## Honest scope note

**Nothing reads this score yet.** `compute_search_score` is deliberately untouched; the consumer is #880. This lands the production of the signal, not its use.

## Verification (macOS, ASan+UBSan)

- `importance complexity pipeline` → **271 passed, 0 failed** (11 + 5 + 255; `pipeline` is an exact baseline match).
- 10 store/index/graph suites → **296 passed**.
- `complexity` is now 5/5; existing gates unmoved (nodes 1.93, edges 2.06, per-lang 1.98, perfile_defs 2.00).
- `make -f Makefile.cbm lint-ci` passes.
- No raw `fopen(` in the diff. Every assertion is on state or counters — no timeout decides a verdict.

## Known limitations, recorded not hidden

- On the closure-delta incremental route the graph buffer holds only the delta, so importance there is computed over a partial graph. `similarity`, `semantic_edges`, `configlink` and `decorator_tags` already carry exactly this limitation on that path; consistency was preferred over inventing new behaviour. Making the delta route exact is a separate design question.
- `append_complexity_props()` in `pass_complexity.c` has the identical append-only shape. It is safe today only because that pass is not wired into the incremental post-passes — if it ever is, it inherits the duplicate-key bug. Worth its own issue.
- `cbm_is_test_path()` still lacks a `_test.c` / `_test.h` case, so the test-penalty under-fires on C (~640 kernel files unclassified). Pre-existing, deliberately out of scope, deserves its own issue.

### DIFF
--- a/Makefile.cbm
+++ b/Makefile.cbm
@@ -354,6 +354,7 @@ PIPELINE_SRCS = \
     src/pipeline/pass_similarity.c \
     src/pipeline/pass_semantic_edges.c \
     src/pipeline/pass_complexity.c \
+    src/pipeline/pass_importance.c \
     src/pipeline/pass_cross_repo.c \
     src/pipeline/artifact.c \
     src/pipeline/pass_ensemble_routing.c \
@@ -560,7 +561,7 @@ TEST_DISCOVER_SRCS = \
 
 TEST_GRAPH_BUFFER_SRCS = tests/test_graph_buffer.c
 
-TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_cross_repo.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_index_resilience.c tests/test_index_format.c tests/test_call_reference_contract.c tests/repro/repro_call_scope_usages.c tests/repro/repro_call_argument_usages.c tests/repro/repro_reference_precision.c tests/repro/repro_lexical_binding_precision.c tests/repro/repro_call_argument_matrix_a.c tests/repro/repro_call_argument_matrix_b.c tests/repro/repro_call_node_behaviors.c tests/repro/repro_language_registry.c tests/repro/repro_call_node_manifest.c tests/repro/repro_lsp_ordered_signatures.c tests/repro/repro_lsp_ordered_local.c tests/repro/repro_ts_overload_return_chains.c tests/repro/repro_harness_cleanup.c tests/repro/repro_runner_filter.c
+TEST_PIPELINE_SRCS = tests/test_registry.c tests/test_pipeline.c tests/test_importance.c tests/test_cross_repo.c tests/test_fqn.c tests/test_route_canon.c tests/test_path_alias.c tests/test_configlink.c tests/test_infrascan.c tests/test_worker_pool.c tests/test_parallel.c tests/test_index_resilience.c tests/test_index_format.c tests/test_call_reference_contract.c tests/repro/repro_call_scope_usages.c tests/repro/repro_call_argument_usages.c tests/repro/repro_reference_precision.c tests/repro/repro_lexical_binding_precision.c tests/repro/repro_call_argument_matrix_a.c tests/repro/repro_call_argument_matrix_b.c tests/repro/repro_call_node_behaviors.c tests/repro/repro_language_registry.c tests/repro/repro_call_node_manifest.c tests/repro/repro_lsp_ordered_signatures.c tests/repro/repro_lsp_ordered_local.c tests/repro/repro_ts_overload_return_chains.c tests/repro/repro_harness_cleanup.c tests/repro/repro_runner_filter.c
 
 TEST_WATCHER_SRCS = tests/test_watcher.c
 
--- a/src/pipeline/pass_importance.c
+++ b/src/pipeline/pass_importance.c
@@ -1,0 +1,396 @@
+/*
+ * pass_importance.c — Index-time per-symbol importance score (weighted degree).
+ *
+ * Predump pass. Computes an importance score for every Function/Method/Class
+ * node and stores it as a numeric "importance" key inside the node's EXISTING
+ * properties_json blob — no new column, no schema change, and therefore no
+ * CBM_INDEX_FORMAT_VERSION bump. Indexes written by older builds simply lack
+ * the key; every consumer must tolerate its absence.
+ *
+ * Model (weighted degree, Aider-repomap shaped):
+ *
+ *   importance = sqrt(num_refs) * priv * generic * distinct * test_penalty
+ *
+ *     num_refs      incoming CALLS + USAGE edges. 0 -> sqrt(0) = 0.
+ *     priv     0.1  name is private (leading underscore)
+ *     generic  0.1  the name is defined in >= 5 DISTINCT files
+ *     distinct 10   name is snake_case or camelCase AND len >= 8
+ *     test_penalty
+ *              0.1  the symbol is test scaffolding — either the target of an
+ *                   incoming TESTS edge (a production helper exercised by a
+ *                   test), or it lives in a test file per the graph's own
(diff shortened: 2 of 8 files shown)


DeusData:
Converting to draft — review turned up a correctness bug in the incremental wiring that I under-described in the PR body above. Recording it here rather than quietly fixing it, because the original framing was wrong.

**What the body says:** that on the closure-delta route importance is computed over a partial graph, and that this matches a limitation `similarity`, `semantic_edges`, `configlink` and `decorator_tags` already carry.

**Why that framing was too soft.** The shared-buffer part is accurate — all those passes do run on the same buffer. But there is an asymmetry that matters: for edge-producing passes a partial graph yields *fewer edges*, a coverage gap. For importance it yields a **wrong scalar that is then persisted over a correct one**. A symbol referenced 500 times project-wide, sitting in a file you just edited, drops to near-zero. "Confidently unimportant" is a worse failure mode than "missing".

**Verified mechanism:**
- `run_postpasses` has two call sites. At `pipeline_incremental.c:2716` (legacy-partial) `ctx.gbuf` is genuinely rehydrated via `cbm_gbuf_load_from_db` at `:2568`.
- At `:2161`, inside `run_closure_delta`, `ctx.gbuf` is the buffer created **empty** by `cbm_gbuf_new` at `:1992` and never loaded from the DB. The in-code comment asserting a "rehydrated whole-project buffer" is false at that site.
- `cbm_delta_patch(staging, project, gbuf, …)` at `~:2181` then persists those scores.
- `closure_try_plan` runs first at `:2447`, so this is the **preferred warm path**, not an edge case.

**Impact today is nil** — nothing reads the score yet; #880 is the consumer. That is exactly why it is worth fixing now rather than letting #880 inherit a signal that is wrong on the primary warm path.

**Fix in progress:** recompute importance at SQL level over the staging store after `cbm_delta_patch` merges, where the complete graph actually exists. Explicitly *not* loading the full graph on the delta path — that would destroy the warm-reindex win this route exists for.

Landing with it: a test asserting a full index and an incremental index that reach the same final state produce **identical** importance values (the guard against the two scoring implementations drifting), a closure-route regression test that asserts the route was actually taken so it cannot pass vacuously, the corrected comment, and re-measured Java timings so the added warm-path cost is visible rather than absorbed.

Everything else in the PR stands as verified: the O(k³)→O(N) fix (483s → 80s), the structural `PREDUMP_PASS_COUNT`, the idempotent append, and the gated complexity counter.

DeusData:
**The delta-route bug flagged above is fixed, and the warm-path cost is now measured rather than estimated.**

## The fix

Importance is recomputed at SQL level over the **staging store after `cbm_delta_patch` merges**, which is the first moment the complete graph exists on the closure-delta route. Not scoped to changed files, and not by materialising the full graph on the delta path — the first is provably wrong, the second would destroy the reason that route exists.

**Divergence is prevented by construction, not merely tested against.** `cbm_pipeline_importance_score()` is a single definition (`pass_importance.c:309`); the SQL path reaches it through a SQLite user-defined function rather than reimplementing the rule, so the two routes cannot drift. SQL contributes only the three aggregates (in-degree, TESTS presence, distinct-files-per-name).

## Why a changed-files-only rescore would be wrong

The equivalence test demonstrates it concretely. Editing `c0.py`'s **body** adds a call to `helper`, which lives in **unchanged** `hot.py`:

```
MISMATCH hot.helper: full=3.605551  incremental=3.464102
```

That is √13 vs √12 — in-degree short by exactly the added caller, on a symbol in a file that did not change. Global inputs move when unrelated files are edited, so the rescore has to be project-wide.

## Measured cost — kernel, 8,529,900 nodes / 15,986,598 edges

Corpus held fixed for every leg, one `-O2` binary, serialized on a quiet machine (86-90% free memory), all four warm legs asserting `closure_repair`, identical node counts:

| leg | wall |
|---|---|
| cold full index | 301.3 s |
| warm **without** recompute | **48.5 s** (44, 53) |
| warm **with** recompute | **65.5 s** (65, 66) |
| **delta** | **+17.0 s** |

Corroborated three independent ways: wall-clock A/B, the recompute's own internal timing (20.5 / 20.7 s), and a standalone measurement on a separate kernel-scale DB (16.6-17.0 s). It scales **sublinearly** — Java at 693k nodes costs +6 s, so 12.3× the nodes yields 2.8× the cost, because the work is two indexed `GROUP BY`s and a 1.13M-row `UPDATE` rather than per-node traversal.

**Accepted as the price of correct scores.** Worth stating plainly: nothing reads this score until #880, so the cost lands before the benefit.

## Two properties the kernel run proved for free

- **1,131,739 scored symbols, all 1,131,739 carrying the key** — 100% coverage at scale.
- **Zero rows with a duplicate `importance` key after four consecutive recomputes** — the idempotence contract holding at 1.1M rows rather than in fixtures.

## Tests

`importance complexity pipeline` → **273 passed**. Both new tests confirmed to have actually executed by name, and both revert-checked RED:

- `importance_closure_delta_keeps_referenced_symbol_sane` — a symbol with 25 callers scored `0.000000` before the fix, `5.000000` after, matching the full index.
- `importance_full_and_incremental_converge_on_same_scores` — 26 symbols compared, 0 mismatches.

Both assert the route was genuinely `CBM_INCREMENTAL_ROUTE_CLOSURE_REPAIR`. That assertion earned itself: an earlier fixture added a *new definition*, which changed the file's surface and made the planner decline to `FORCED_FULL` — the test would have passed on the wrong path without it.

## Known follow-up

The same scoring costs **2.2 s in-memory on the cold path vs ~20.6 s in SQL on the warm path**. The cost is write volume, not computation — a 1.13M-row UPDATE where a one-file edit changes very few scores. Updating only rows whose value actually moved would preserve exact equivalence while cutting most of that. Filed separately; not attempted here.

PR #1894, opened 2026-08-29
fix: recover markerless stale rendezvous sockets during activation

### LINKED ISSUE #1760: Stale rendezvous directory (`/tmp/cbm-daemon-<uid>`) permanently blocks `install`/`update` with "active CBM sessions ... could not be stopped safely" after hard-killed CBM processes

mvanhorn:
## What does this PR do?

Extend the existing POSIX stale-generation cleanup in `src/daemon/ipc.c` to recognize the markerless hard-linked `.sock`/`.anc` pair only while it retains the matching startup lock and a temporary exclusive lifetime reservation. Require both paths to be secure sockets for the same device/inode with the exact two-link shape, revalidate those identities immediately before inode-matched unlinking, and continue to refuse unanchored, mismatched, malformed, replaced, or actively reserved endpoints. Make the activation-only generation probe invoke this serialized cleanup before classifying a remaining endpoint, so `install` and `update` use the recovery path already shared by daemon startup without weakening the conservative `ECONNREFUSED` behavior for live or saturated listeners.

Hard-killing a POSIX daemon or MCP process can leave the owner-private rendezvous namespace with a hard-linked stable socket and anchor but without the pending or identity record that normally authorizes cleanup. The kernel releases the process-owned startup, lifetime, and cohort locks, yet activation's generation probe treats the refused socket as active and permanently blocks `install` and `update`. Deleting `/tmp/cbm-daemon-<uid>` works around the problem, confirming that durable transport artifacts—not a live CBM session—cause the refusal. The current branch already documents `CBM_RUNTIME_DIR`, distinguishes coordination failures from busy sessions, and returns a nonzero activation status, so this change is limited to safe automatic recovery of the reported crash shape.

Fixes #1760

## Checklist

  Not run: no test command resolved in this workspace, so nothing was executed to pass.
  Not run: no test command resolved in this workspace, so nothing was executed to pass.
  Not run: no test command resolved in this workspace, so nothing was executed to pass.


DeusData:
Thanks for this — and apologies that the only thing you have heard so far was the automated acknowledgement.

Status: the one thing blocking it is DCO. Every other required check is green, which is a better starting position than most PRs arrive in.

It is a single commit (`498f69b4`), so it should be one round trip: amend it with a sign-off (`git commit --amend -s --no-edit`) and force-push with lease.

The substance still needs a proper read against #1760 before anything is promised — that is queued, not skipped. But the sign-off is worth doing now so the review is not waiting on a formality when it comes up.

DeusData:
Review done, and the hard part holds up: I had two independent passes on the cleanup path, one of them adversarial. The only unlink primitive reached is `posix_socket_path_unlink_inode_if_matches` — `fstatat(AT_SYMLINK_NOFOLLOW)`, socket type, owner == euid, mode 0600, exact link count, then dev+ino compared against the identity captured just before, then `unlinkat(dir_fd, name, 0)` — inside a runtime directory opened by the component-wise `O_DIRECTORY|O_NOFOLLOW` walk, with bare 16-hex basenames. The markerless branch runs only under the startup lock and only after taking the lifetime reservation itself, and `.anc` plus the lifetime lock were born in the same commit, so a same-inode nlink-2 pair with no lifetime holder really is a dead owner and a legacy bare socket is preserved. Attempts to unlink a live daemon's socket, another user's socket, a symlink target, or to race the identity capture all hit an explicit check. On a tree merged with current main the nine daemon/IPC/CLI suites pass (542/0), and with main's `ipc.c`/`ipc.h` under your tests exactly the three new ones go red — they bind. Your sign-off is in place (the earlier ask was about the older commit), lint and the no-skips check are clean, and I've refreshed the branch from main so the MSan leg (which died building its Docker image, not on your code) gets a real run.

One line before it merges — a test-isolation defect, not a logic one: your new cli test resets the activation runtime parent with `cbm_cli_set_activation_runtime_parent_for_test(NULL)` (`tests/test_cli.c`, the new test around line 1424). Main's suite contract expects `g_cli_suite_runtime_parent` to be restored there (see the sibling at `:1310` and the contract at `:61–65`); with `NULL`, every later cli test runs the now-mutating activation probe against the developer's real `/tmp/cbm-daemon-<uid>` — a host-dependent verdict that can stop a live daemon. Restore `g_cli_suite_runtime_parent` instead of `NULL`, keep the sign-off, and I merge on green.

For the record on #1760: this fix covers both the markerless shape and the marker-present residue for every probe caller, so it supersedes #1790 (whose own test passes on your tree unchanged); I'll close that one with credit once yours lands. Thanks for the careful work — the nlink/identity proof is the kind of thing that makes a cleanup path trustworthy.

DeusData:
One more note so you do not chase it: the `test-lsan-macos` red on the refreshed branch is the watcher suite (`tests/test_watcher.c:1719`, a known nondeterministic case on that runner, with the leak report being the failed test's own teardown), nothing in your files. The only thing this PR still needs is the one-line seam restore above; CI re-runs on your push.

DeusData:
Merged as 760c08f453aa — thank you, and thanks for the seam-restore follow-up commit. Recognising the markerless hard-linked .sock/.anc pair only while holding the startup lock plus a temporary exclusive lifetime reservation, and revalidating device/inode and the exact two-link shape before unlinking, keeps the recovery inside the daemon's own ownership boundary; that is what made this mergeable where a plain "delete stale files" fix would not have been. With nine independent reports on #1757/#1760 this was the widest-reaching install/daemon bug in the queue. This ships in v0.10.9.

PR #1895, opened 2026-08-29
Preserve separate C# attribute groups

### LINKED ISSUE #1865: decorators/query_graph drops one attribute when a method has multiple separate C# attribute brackets (e.g. [HttpPost] + [Route(...)])
## Summary

When a C# method carries two separate attribute brackets on the same declaration — e.g. an HTTP-verb attribute and a `[Route(...)]` attribute as distinct `[...]` groups rather than combined in one bracket — the `decorators` property returned via `query_graph` for that `Method` node only contains one of the two attributes. The other is silently dropped.

```csharp
[HttpPost]
[Route("orders/{id}/confirm")]
public async Task Confirm(Guid id)
{
    ...
}
```

We queried this via:

```cypher
MATCH (m:Method)
WHERE m.decorators IS NOT NULL
RETURN m.name, m.file_path, m.decorators
```

For methods shaped like the example above, `m.decorators` came back with only the `[HttpPost]` entry — `[Route("orders/{id}/confirm")]` was missing — even though the source file on disk clearly has both attributes present. We confirmed this by diffing CBM's output against the raw source for the same file/method.

## Impact

We index several ASP.NET Core / ABP-style .NET monoliths where this two-bracket style (verb attribute + route attribute as separate lines) is the dominant convention. Because of this, CBM's built-in `:Route` node ends up populated for only a small fraction of the routes we'd expect across our indexed repos (on the order of a handful out of tens of thousands of methods with HTTP attributes). We ended up writing a separate regex-based fallback to reconstruct API routes (merging class-level `[Route]` with method-level `[HttpGet]`/`[HttpPost]`) rather than relying on CBM's own decorator/route extraction, since the built-in extraction wasn't usable for this.

## Version

Pinned to **v0.8.1**. We're stuck on this version for an unrelated reason (a separate regression where `query_graph`'s output changed to a human-readable text table with no JSON path in later releases we tried), so we haven't been able to confirm whether this specific decorator-dropping issue still reproduces on `main` / the latest release. Flagging it here in case it's still present.

## Expected behavior

`m.decorators` (or whatever the AST layer collects for a method's attributes) should include *all* attribute brackets attached to a method declaration, not just the first/last one found.

taljeon:
## What does this PR do?

Fixes #1865.

C# declarations can contain multiple `attribute_list` children when attributes use separate bracket groups. Decorator extraction previously looked up only one wrapper, so later groups were omitted from the definition's `decorators` array.

This change counts and collects decorators from every annotation wrapper in source order. It also adds a focused regression using separate `[HttpPost]` and `[Route(...)]` groups.

## Validation

- `scripts/test.sh --suites extraction` — 308 passed
- `CBM_TEST_PAR_JOBS=4 scripts/test.sh` — 7,659 passed, 0 failed, 8 platform skips; all prod guards passed
- `scripts/lint.sh --ci` with the CI clang-format 20 and cppcheck toolchain
- `make -f Makefile.cbm security` — all security checks passed, including 32/32 MCP robustness cases
- `git diff --check`

Development note: OpenAI Codex assisted with implementation and verification; I reviewed the change and validation results.

## Checklist


DeusData:
Approved. Small, correct, and refactored in the way that makes it safe.

**The symmetry is the part that matters.** `count_wrapped_decorators` and `collect_wrapped_decorators` walk the children with the identical shape — same `ts_node_child_count` loop, same `wrapper_kind` comparison. That is what stops the two disagreeing, and a count/collect mismatch here would either overrun `result` or silently drop entries. Splitting `annotation_wrapper_kind()` out as a pure language→kind mapping, and leaving `find_jvm_modifiers` as a thin wrapper over it rather than deleting it, keeps its other caller working; I checked and it still has two.

**One thing worth stating that the title does not:** this is not only a C# fix. The wrapper path serves Java, Kotlin, Swift, C# and PHP, and the old code took the first wrapper for all of them. So PHP 8 gains the same repair — `#[A] #[B]` as separate `attribute_group`s under separate `attribute_list` nodes had the same truncation. Java, Kotlin and Swift are unaffected in practice since `modifiers` is a single node there.

That is a correct broadening of the same bug class rather than scope creep, but it deserves to be in the description so nobody later wonders why a "C# attribute groups" commit changed PHP behaviour.

**And the disclosure is appreciated.** "OpenAI Codex assisted with implementation and verification; I reviewed the change and validation results" is exactly the right note to leave — it is welcome here and it tells a reviewer what kind of scrutiny the change has already had.

**Status:** cleared, reviewed, approved. `main` was briefly broken earlier today and repaired by #1993; your green predates that, so I will update the branch against current `main` and re-run before merging. Nothing needed from you.

DeusData:
Closing this in favour of **#1759**, which fixes the same bug more completely. Two things I owe you before the reasoning: I approved this PR, and I got a detail wrong in that approval. Both are mine, not yours.

**I approved it without checking whether another open PR fixed the same thing.** #1759 was open at the time. That is a review failure on my side, and it is the reason this is being closed after being told it was good rather than before.

## Why #1759 wins on the merits

`find_jvm_modifiers()` has two callers:

```
extract_defs.c:1714   inside scan_route_annotations   (route extraction)
extract_defs.c:1981   inside extract_decorators
```

Your change adds the wrapper-iterating helpers inside `extract_decorators` and leaves `find_jvm_modifiers` returning a single node — so `:1714` still sees only the first `attribute_list`. #1759 fixes the helper itself, so both callers get every wrapper.

That difference is not academic. The symptom reported in #1692 is `[ServiceFilter(...)]` followed by `[Route("api/[controller]")]`, where the route attribute is invisible — and route extraction runs through `scan_route_annotations`. Your fix would have corrected the `decorators` array while leaving the reported symptom in place.

Fixing the helper is also the better shape generally: the next caller of `find_jvm_modifiers` gets correct behaviour without having to know it needed to iterate.

## A correction I owe you

In my earlier review I said this change also repaired PHP 8, on the grounds that `#[A] #[B]` produces separate `attribute_list` nodes. I asserted that without checking the grammar, and I now believe it is wrong: PHP's grammar has `attribute_list` as the container and `attribute_group` per `#[...]` group, so a stack is one list with several groups and the "first wrapper only" bug never applied there. Please disregard that part of what I told you.

## What was genuinely good here

The property you protected is the one that matters in this kind of change: `count_wrapped_decorators` and `collect_wrapped_decorators` walk the children with an identical shape, so the two cannot disagree. A count/collect mismatch would either overrun `result[]` or silently drop entries, and keeping them symmetrical is what makes the change safe to read. Splitting `annotation_wrapper_kind()` out as a pure mapping while leaving `find_jvm_modifiers` working for its other caller was careful too.

None of that is wasted — it is the same diagnosis, found independently, and you were right about the bug.

Thank you for it, and sorry for the wasted round trip on my account.

PR #1896, opened 2026-08-29
fix(extraction): reach Swift call arguments through call_suffix

### LINKED ISSUE #1892: Swift: no client-side Route or HTTP_CALLS emitted, so cross-repo-intelligence returns 0 edges (URLSession is already in the pattern table)
### Version

`dev` build from source at `4da8f877` (tag `v0.10.8` lineage). I am 19 commits behind `origin/main`; none of those 19 touch `internal/cbm/service_patterns.c`, `src/pipeline/pass_cross_repo.c`, or any Swift front-end file, and my own local commits do not touch them either. So this should reproduce on stock `main`.

### Platform

macOS (Apple Silicon) — macOS 26.6.2, arm64

### Install channel

Built from source

### Binary variant

standard

A Swift client that calls an HTTP endpoint produces **no `HTTP_CALLS` edge and no client-side `Route` node**, so `mode=cross-repo-intelligence` can never link it to the server that serves that path. A TypeScript client making the byte-identical call links correctly.

Swift is not missing from the pattern table. `internal/cbm/service_patterns.c` lines 109-112 already carry:

```c
    /* Swift */
    {"Alamofire", CBM_SVC_HTTP, NULL},
    {"Moya",      CBM_SVC_HTTP, NULL},
    {"URLSession", CBM_SVC_HTTP, NULL},
```

The gap is one step earlier: the Swift front end never emits the client-side `Route` node + `HTTP_CALLS` edge carrying a `url_path` that the cross-repo matcher joins on.

I expected the Swift client to behave like the TypeScript one — an `HTTP_CALLS` edge with `url_path=/api/v1/widgets`, and a `CROSS_HTTP_CALLS` edge naming the Go handler.

**Note on #523**, which asks "what is the minimum a call and a route need to share to be linked?" — the TypeScript half of this repro answers that, and shows the matcher itself is working. This report is the opposite failure: the matcher is fine, one language front end never feeds it. Not a duplicate of #523, #678, #706 (those are matcher/extractor bugs where the call side does produce an edge), nor of #551 (SwiftPM manifest resolution for package-graph edges, a different pass).

### Reproduction

Three tiny repos, all declaring the same path `/api/v1/widgets`. The only variable is the client language. Script attached below; it writes the fixtures and prints the five MCP calls to make.

Results:

| Repo | Language | `Route` nodes | `HTTP_CALLS` | `CROSS_HTTP_CALLS` to server |
|:--|:--|--:|--:|--:|
| `srv-go` | Go + chi | 2 | — | — |
| `cli-ts` | TypeScript + axios | 2 | 2 | **2** |
| `cli-swift` | Swift + URLSession | 0 | 0 | **0** |

TypeScript, working — `MATCH ()-[r:CROSS_HTTP_CALLS]->() RETURN ...`:

```
url_path          target_project   target_file   target_function   transport
/api/v1/widgets   xtest-srv-go     main.go       listWidgets       GET
/api/v1/widgets   xtest-srv-go     main.go       createWidget      POST
```

Swift, same path: `cross_http_calls: 0, total_cross_edges: 0`.

Graph schema for the Swift fixture shows no `Route` label and no `HTTP_CALLS` edge type. Its only edge for the URLSession call is a bare `USAGE` — no `CALLS` edge is emitted at all.

<details>
<summary>repro-swift-http.sh</summary>

```bash
#!/usr/bin/env bash
# Reproduce: a Swift client produces no HTTP_CALLS, so cross-repo-intelligence
# finds no edges. A TypeScript client with the identical path produces both.
# Builds three tiny repos, then index each and run cross-repo-intelligence.
set -euo pipefail
B="${1:-/tmp/xrepo}"
rm -rf "$B"; mkdir -p "$B"/srv-go "$B"/cli-ts "$B"/cli-swift

cat > "$B/srv-go/go.mod" <<'EOF'
module example.com/srv
go 1.22
EOF
cat > "$B/srv-go/main.go" <<'EOF'
package main

import "github.com/go-chi/chi/v5"

func main() {
	r := chi.NewRouter()
	r.Get("/api/v1/widgets", listWidgets)
	r.Post("/api/v1/widgets", createWidget)
	http.ListenAndServe(":8080", r)
}

func listWidgets(w http.ResponseWriter, r *http.Request)  {}
func createWidget(w http.ResponseWriter, r *http.Request) {}
EOF

cat > "$B/cli-ts/package.json" <<'EOF'
{ "name": "cli-ts", "version": "1.0.0", "dependencies": { "axios": "^1.6.0" } }
EOF
cat > "$B/cli-ts/client.ts" <<'EOF'
import axios from "axios";

export async function listWidgets() {
  return axios.get("/api/v1/widgets");
}

export async function createWidget(body: unknown) {
  return axios.post("/api/v1/widgets", body);
}
EOF

cat > "$B/cli-swift/Package.swift" <<'EOF'
// swift-tools-version:5.9
import PackageDescription

let package = Package(
    name: "CliSwift",
    targets: [.target(name: "CliSwift", path: "Sources/CliSwift")]
)
EOF
mkdir -p "$B/cli-swift/Sources/CliSwift"
cat > "$B/cli-swift/Sources/CliSwift/Client.swift" <<'EOF'
import Foundation

final class Client {
    func listWidgets() async throws -> Data {
        let url = URL(string: "https://example.com/api/v1/widgets")!
        let (data, _) = try await URLSession.shared.data(from: url)
        return data
    }

    func createWidget(body: Data) async throws -> Data {
        var request = URLRequest(url: URL(string: "https://example.com/api/v1/widgets")!)
        request.httpMethod = "POST"
        let (data, _) = try await URLSession.shared.upload(for: request, from: body)
        return data
    }
}
EOF

for d in srv-go cli-ts cli-swift; do
  (cd "$B/$d" && git init -q . && git add -A \
    && git -c user.email=t@t -c user.name=t commit -qm init)
done
echo "Fixtures built in $B"
echo
echo "Now, through the MCP server:"
echo "  index_repository  repo_path=$B/srv-go     name=xtest-srv-go     mode=full"
echo "  index_repository  repo_path=$B/cli-ts     name=xtest-cli-ts     mode=full"
echo "  index_repository  repo_path=$B/cli-swift  name=xtest-cli-swift  mode=full"
echo "  index_repository  repo_path=$B/cli-ts     name=xtest-cli-ts     mode=cross-repo-intelligence  target_projects=[\"xtest-srv-go\"]"
echo "  index_repository  repo_path=$B/cli-swift  name=xtest-cli-swift  mode=cross-repo-intelligence  target_projects=[\"xtest-srv-go\"]"
```

</details>

### On a real project

Same shape at scale, on a 462-file / 64k-line Swift app talking to a Go backend:

| | Go + TS backend repo | Swift client repo |
|:--|--:|--:|
| `Route` nodes | 332 | label absent |
| `HTTP_CALLS` edges | 215 | edge type absent |
| `Method` nodes | 2,877 | 3,766 |
| `CALLS` edges | 20,650 | 12,017 |

Swift parsing itself is healthy — 3,766 methods, 12,017 call edges. Only the HTTP layer is missing. On the backend side the 215 `HTTP_CALLS` include its TypeScript frontend (`frontend/lib/api/classes/items.ts` alone contributes 20), so TS client extraction works in a real tree too.

In the real Swift repo the failure takes a slightly different route to the same place. A `CALLS` edge for the URLSession call *does* exist, but resolves onto a local property:

```
callee:     URLSession.shared.data
strategy:   suffix_match    confidence: 0.21
resolved:   <project>.HomeboxKit.Sources.HomeboxUI.PickedFile.data
file:       HomeboxKit/Sources/HomeboxFeatures/Chores/ChoresDetailViewModel.swift:105
```

Because `cbm_service_pattern_*` matches the library identifier in the **resolved** qualified name, and the resolved name is a local `PickedFile.data`, the check cannot fire. The Swift graph also has no `Package` label at all, so no framework identifier can ever appear in a resolved name — the Go/TS project has 215 `Package` nodes with an `external` flag.

That `suffix_match` at 0.21 may be worth a look on its own: Swift calls into system frameworks are being resolved onto same-named local members, which affects more than the HTTP case.

Repro fixtures: 6 files. Real project: 462 Swift files / 64,101 lines against a 1,485-file Go+TS backend.

CaptainMittens:
## What does this PR do?

Swift calls lost every argument, so no Swift HTTP call could ever produce an
`HTTP_CALLS` edge or a `Route` node.

`handle_calls` reads a call's arguments through one tree-sitter field lookup:

```c
TSNode args = ts_node_child_by_field_name(node, TS_FIELD("arguments"));
```

The vendored Swift grammar declares no `"arguments"` field at all — its own
`ts_field_names[]` table has zero occurrences of the string, against one in Go
and two in TypeScript. Swift models a call as a `target` expression plus a
`call_suffix`, and the arguments hang off the suffix as `value_arguments`.

So `args` was null for **every Swift call ever parsed**, `first_string_arg` was
never populated, and everything downstream that needs the URL could not fire.
`Alamofire`, `Moya` and `URLSession` have been in the service-pattern table
(`internal/cbm/service_patterns.c`) the whole time and match the callee text
correctly — the URL simply never arrived.

Three changes, all in `internal/cbm/extract_calls.c`:

1. `swift_call_args()` reaches the argument list through `call_suffix`, in the
   same shape as the existing `objectscript_call_args()` fallback, and is used
   from the same place. A trailing closure has no `value_arguments`, so it
   returns a null node and that call behaves exactly as before.
2. `extract_url_or_topic_arg()` unwraps Swift's per-argument `value_argument`
   node, stepping past a leading `value_argument_label`. Without this,
   `dataTask(with: "/api/v1/widgets")` yields the label `with` rather than the
   path. PHP and C# already had the equivalent unwrap for their `argument` node.
3. `is_string_like()` gains `line_string_literal`, which is what Swift calls an
   ordinary `"…"` literal. The list already held `raw_string_literal`, so only
   Swift's common case was missing.

### What this does not do

It makes a **literal** URL argument reachable. It does not resolve
`URLSession.shared.data(from: URL(string: "…")!)`, where the literal sits inside
a nested constructor — `extract_positional_url` handles a literal, a template
string, a concatenation, or a named constant, and that nested shape is none of
those. Issue #1892 is therefore not fully closed by this change; it is the
blocking layer underneath it.

Also worth recording for whoever picks that up: Alamofire's `AF.request(…)`
shorthand does not match the pattern table, because the table matches the
library name in the callee text and `AF.request` contains none. That is why the
pipeline test below uses `URLSession`.

No `Package.swift` handling is touched — that is #551, already scoped elsewhere.

Fixes #1892

## Validation

Reproduce-first. With the fix reverted and the tests kept, all three fail:

```
swift_call_string_arg_issue1892           FAIL tests/test_extraction.c:3752: c->first_string_arg is NULL
swift_labeled_call_string_arg_issue1892   FAIL tests/test_extraction.c:3769: c->first_string_arg is NULL
pipeline_swift_http_call_makes_route_issue1892  FAIL tests/test_pipeline.c:5028: HTTP_CALLS (0) not >= 1
```

With the fix:

| Suite | Result |
|:--|:--|
| `extraction` | 309 passed (307 before, plus the 2 new) |
| `pipeline` | 256 passed (255 before, plus the 1 new) |
| full `scripts/test.sh` | 7633 passed, 28 failed, 8 skipped (140 suites) |

The 28 failures are all in `tests/test_cli.c` (client install/uninstall) and are
pre-existing in my environment, not caused by this change. Measured rather than
assumed: I reverted only `internal/cbm/extract_calls.c` to `origin/main`, kept
the new tests, rebuilt, and re-ran that suite — `263 passed, 28 failed`, the same
count, and every failure recorded in the full run fails on the baseline too. If
they are green on your runners, they are environmental here and worth ignoring
for this PR.

Lint: `make -f Makefile.cbm lint-cppcheck` and `lint-no-suppress` both clean.
`clang-format` reports no change wanted on any line this PR adds. I could not run
`clang-tidy` locally — it is not in my toolchain — so that one is unverified on
my side; CI's `lint-ci` covers cppcheck and clang-format, which I did run.

Built with `-Wall -Wextra -Werror`, no new warnings.

## Checklist

- [x] Lint passes (`make -f Makefile.cbm lint-ci`) — cppcheck and clang-format run clean; clang-tidy not installed locally

### DIFF
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -61,7 +61,8 @@ static const char *lookup_url_builder(const CBMExtractCtx *ctx, const char *name
 static int is_string_like(const char *kind) {
     return (strcmp(kind, "string") == 0 || strcmp(kind, "string_literal") == 0 ||
             strcmp(kind, "interpreted_string_literal") == 0 ||
-            strcmp(kind, "raw_string_literal") == 0 || strcmp(kind, "string_content") == 0);
+            strcmp(kind, "raw_string_literal") == 0 || strcmp(kind, "string_content") == 0 ||
+            strcmp(kind, "line_string_literal") == 0);
 }
 
 /* Strip surrounding quotes from a string, return arena-allocated copy */
@@ -2286,6 +2287,18 @@ static const char *extract_url_or_topic_arg(CBMExtractCtx *ctx, TSNode args) {
         if (strcmp(ts_node_type(arg), "argument") == 0 && ts_node_named_child_count(arg) > 0) {
             arg = ts_node_named_child(arg, 0);
         }
+        /* Swift wraps each argument in a value_argument that may lead with its
+         * label, so `data(from: url)` would otherwise yield the label `from`
+         * rather than the value. Step past a leading value_argument_label. */
+        if (strcmp(ts_node_type(arg), "value_argument") == 0 &&
+            ts_node_named_child_count(arg) > 0) {
+            TSNode val = ts_node_named_child(arg, 0);
+            if (strcmp(ts_node_type(val), "value_argument_label") == 0 &&
+                ts_node_named_child_count(arg) > 1) {
+                val = ts_node_named_child(arg, 1);
+            }
+            arg = val;
+        }
         const char *ak = ts_node_type(arg);
 
         if (strcmp(ak, "keyword_argument") == 0 || strcmp(ak, "pair") == 0) {
@@ -3045,6 +3058,19 @@ static TSNode objectscript_call_args(TSNode node) {
                                            : cbm_find_child_by_kind(macro_function, "method_args");
 }
 
+/* Swift models a call as a target expression plus a call_suffix, and its grammar
+ * declares no "arguments" field at all, so the generic field lookup finds
+ * nothing for every Swift call. Reach the argument list through the suffix
+ * instead. A trailing closure has a call_suffix with no value_arguments, which
+ * returns a null node and leaves the call without a string argument, as before. */
+static TSNode swift_call_args(TSNode node) {
+    TSNode suffix = cbm_find_child_by_kind(node, "call_suffix");
+    if (ts_node_is_null(suffix)) {
+        return (TSNode){0};
+    }
+    return cbm_find_child_by_kind(suffix, "value_arguments");
+}
+
 static bool node_has_token(TSNode node, const char *token) {
     uint32_t count = ts_node_child_count(node);
     for (uint32_t i = 0; i < count; i++) {
@@ -3633,6 +3659,10 @@ CBMInvocationDescriptor handle_calls(CBMExtractCtx *ctx, TSNode node, const CBML
             // generic "arguments" field; macro arguments add one wrapper.
             if (ts_node_is_null(args) && is_objectscript_language(ctx->language)) {
                 args = objectscript_call_args(node);
+            }
+            // Swift has no "arguments" field either; its args hang off call_suffix.
+            if (ts_node_is_null(args) && ctx->language == CBM_LANG_SWIFT) {
+                args = swift_call_args(node);
             }
             if (!ts_node_is_null(args)) {
                 call.first_string_arg = extract_url_or_topic_arg(ctx, args);
--- a/tests/test_extraction.c
+++ b/tests/test_extraction.c
@@ -4027,6 +4027,41 @@ static const CBMCall *find_call_by_callee(CBMFileResult *r, const char *callee) 
     return NULL;
 }
 
+/* #1892: the Swift grammar declares no "arguments" field, so the generic field
+ * lookup read nothing and every Swift call lost its arguments. Without the URL
+ * the service-pattern table cannot raise an HTTP_CALLS edge or a Route node,
+ * even though Alamofire/Moya/URLSession are already listed in it. */
+TEST(swift_call_string_arg_issue1892) {
+    CBMFileResult *r =
+        extract("func listWidgets() { AF.request(\"https://example.com/api/v1/widgets\") }\n",
+                CBM_LANG_SWIFT, "t", "Client.swift");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    const CBMCall *c = find_call_by_callee(r, "AF.request");
+    ASSERT_NOT_NULL(c);
+    ASSERT_NOT_NULL(c->first_string_arg);
+    ASSERT_STR_EQ(c->first_string_arg, "https://example.com/api/v1/widgets");
+    cbm_free_result(r);
+    PASS();
+}
+
+/* Swift labels its arguments, and each one sits in a value_argument node that
+ * leads with the label. Reading the first child alone would return `with`
+ * rather than the path. */
+TEST(swift_labeled_call_string_arg_issue1892) {
+    CBMFileResult *r =
+        extract("func fetch() { URLSession.shared.dataTask(with: \"/api/v1/widgets/1\") }\n",
+                CBM_LANG_SWIFT, "t", "Fetch.swift");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    const CBMCall *c = find_call_by_callee(r, "URLSession.shared.dataTask");
+    ASSERT_NOT_NULL(c);
+    ASSERT_NOT_NULL(c->first_string_arg);
+    ASSERT_STR_EQ(c->first_string_arg, "/api/v1/widgets/1");
+    cbm_free_result(r);
+    PASS();
+}
+
 /* Issue #1009: URL-builder helper pattern — a function returning a URL-shaped
  * literal, consumed as client(buildPath(id)). The builder's URL is recorded in
  * the per-file constant map and resolved at the call site, for both return
@@ -6782,6 +6817,8 @@ SUITE(extraction) {
     RUN_TEST(swift_constructor_call);
     RUN_TEST(swift_chained_call);
     RUN_TEST(swift_force_unwrap_scanner_shift);
+    RUN_TEST(swift_call_string_arg_issue1892);
+    RUN_TEST(swift_labeled_call_string_arg_issue1892);
     RUN_TEST(objc_interface);
     RUN_TEST(objc_implementation);
     RUN_TEST(dart_top_level_function);
--- a/tests/test_pipeline.c
+++ b/tests/test_pipeline.c
@@ -5480,6 +5480,61 @@ TEST(pipeline_native_fetch_classified_as_http_calls) {
     PASS();
 }
 
+/* #1892: Swift produced no Route node and no HTTP_CALLS edge, because the
+ * Swift grammar has no "arguments" field and the generic lookup therefore read
+ * no call arguments at all. Alamofire/URLSession were already in the service
+ * pattern table; the URL simply never reached it. This is the Swift twin of
+ * the TypeScript fetch case above. */
+TEST(pipeline_swift_http_call_makes_route_issue1892) {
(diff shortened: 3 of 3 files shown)


CaptainMittens:
#1976 builds on this one — it reaches the `URL(string: "…")!` constructor that wraps the string, which is the shape #1892 actually reported. Reviewing this PR first makes that one narrow to its own two commits, since GitHub shows this commit on it as well.

DeusData:
Reviewed and approved. This is the base of your stack and it holds up on its own.

`swift_call_args` is gated twice — on `ctx->language == CBM_LANG_SWIFT`, and on `args` already being null — so no other language can reach it. Following the `objectscript_call_args` shape that was already there for the same "grammar declares no arguments field" problem is the right call; it makes the next reader see two instances of one pattern rather than two ad-hoc special cases. Noting in the comment that a trailing closure yields a `call_suffix` with no `value_arguments`, and that this returns null and leaves behaviour unchanged, is the detail that makes the guard reviewable instead of just plausible.

One thing I checked rather than assumed, because it was the only part of this that is not language-gated: `is_string_like` is shared across every language, so adding `line_string_literal` to it could have changed behaviour well outside Swift. It does not — `line_string_literal` appears in exactly one of the 159 vendored grammars, and that one is Swift. So the widening is inert everywhere else. Worth stating explicitly since the diff does not make it obvious.

**What happens next, so you are not left guessing:** `main` does not currently compile — two changes landed the same type short-name index into `type_registry.h` and git merged them into duplicate struct members. Your green here predates that, so it is stale rather than wrong. #1993 is the repair. Once it lands I will re-run this PR's checks against a fixed `main` and merge it if it is green — there is nothing for you to do, and no rebase needed unless the re-run says otherwise.

That also unblocks the sequencing on #1976, which narrows to its own commit once this is in.

Thanks for tracing this to the grammar's shape rather than special-casing the symptom — "Swift declares no arguments field at all" is the sentence that explains why every Swift call was missing its URL, and it took finding to write.

CaptainMittens:
Understood — nothing for me to do here, and I will leave the branch alone so the re-run tests what you reviewed.

On the `is_string_like` check: that was the one part I could not prove from inside the diff, so thank you for counting it. I had reasoned that `line_string_literal` was Swift-only from the grammar I was reading, which is not the same thing as knowing it appears in one of 159. Your check is the evidence; mine was an assumption that happened to be right.

`main` is now repaired — #1993 is in as `17786374`, and this branch is 3 commits behind it with no conflict, so a re-run should be clean whenever you get to it.

DeusData:
Status update: **your blocker has landed.**

This was recorded on our side as approved and waiting on a fix to `main`. That fix is in — #1986 merged today as `620613ed`, carrying the Swift scanner shift correction and its `MANIFEST.md` row.

The consequence is that this branch is now `DIRTY` and needs a rebase onto current `main`. `main` has taken well over a dozen merges today, so the conflict is expected rather than a sign of anything wrong with the change.

Worth knowing about the chain: **#1976** ("reach a Swift URL built by a constructor") sits on top of this one and is `CONFLICTING` for the same reason. If you rebase this first, that one rebases cleanly on top. Either order works — they touch the same file and neither is large — I only want the sequencing to be deliberate rather than a surprise.

The change itself is approved and needs nothing further from you beyond the rebase. Its three current red checks are against a stale base and will be re-run once it is refreshed.

DeusData:
Merged as `5802ccdd` — thank you for this one, and for the patience through the #1977 → #1986 detour it had to wait out.

The change is exactly what the review endorsed: Swift calls reach their argument list through `call_suffix` where the generic `arguments` field lookup finds nothing, the leading `value_argument_label` is stepped past so `data(from: url)` yields the value and not the label, and `line_string_literal` joins the string-like kinds. Gated on Swift, so no other language moves. Re-diffed before merging: one commit, `extract_calls.c` plus the two tests, no drift from what was reviewed.

#1976 now narrows to its single URL-constructor commit on top of this; picking that up next.

PR #1897, opened 2026-08-29
fix(registry): refuse a name-only match the receiver chain contradicts

### LINKED ISSUE #1893: Swift: Foundation calls bind to same-named project members (unique_name, confidence 0.75) — two of three land on stored properties, so CALLS edges target :Variable nodes
### Version

`dev` build from source at `4da8f877` (tag `v0.10.8` lineage). 19 commits behind `origin/main`; none of those 19 touch the Swift front end or the call-resolution passes, and my own local commits do not either. Should reproduce on stock `main`.

### Platform

macOS (Apple Silicon) — macOS 26.6.2, arm64

### Install channel

Built from source

### Binary variant

standard

In a Swift project, calls into Foundation are bound to **project-local members that merely share the last path component**. The call graph then reports edges that do not exist.

Two of the three in the minimal repro below resolve onto **stored properties** — the graph contains a `CALLS` edge whose target is a `:Variable` node, which cannot be called at all.

This is not confined to a low-confidence tier. The minimal case produces all three at `strategy=unique_name, confidence=0.75`; in the real project the same failure also appears via `suffix_match` at 0.14, 0.21 and 0.55. So a confidence threshold alone would not catch it.

I expected these calls either to be left unresolved or to be marked as external, not bound to unrelated project symbols.

**Same class as the already-recognised family**, filed per-language the way this project already does: `#476` Perl (closed, fixed), `#606` PHP (open), `#1276` Python (open), `#1572` Python/TS cross-language via `unique_name`, `#1355` same-named symbol binding. Swift is not yet reported.

Worth noting alongside `#43` (*Swift: CALLS edges not resolved*), which is **closed as completed**. That fix is working — Swift now produces 12,017 `CALLS` edges in my real project. This report is the accuracy gap that arrived with them.

Related to my `#1892` only in that both are Swift front-end gaps; the two are independent.

### Reproduction

One Swift file. Three project types whose member names collide with Foundation members; three calls that are unambiguously Foundation's.

```swift
import Foundation

struct PickedFile {
    let data: Data          // a stored property, not callable
}

struct DateOnly {
    static func string(from date: Date) -> String { "" }
}

struct Point {
    let date: Date?         // a stored property, not callable
}

final class Caller {
    func fetch() async throws -> Data {
        let url = URL(string: "https://example.com/x")!
        let (data, _) = try await URLSession.shared.data(from: url)
        return data
    }

    func readSetting() -> String? {
        return UserDefaults.standard.string(forKey: "someKey")
    }

    func tomorrow(from now: Date) -> Date? {
        return Calendar.current.date(byAdding: .day, value: 1, to: now)
    }
}
```

With a minimal `Package.swift`, `git init`, then `index_repository mode=full`.

`MATCH (a)-[r:CALLS]->(b) RETURN r.callee, r.strategy, r.confidence, b.qualified_name, b.start_line`:

| callee | strategy | confidence | resolved to | target line | correct? |
|:--|:--|--:|:--|--:|:--|
| `URLSession.shared.data` | unique_name | 0.75 | `Main.data` | 5 | no — `let data: Data` |
| `UserDefaults.standard.string` | unique_name | 0.75 | `Main.DateOnly.string` | 9 | no — `string(from:)` is unrelated |
| `Calendar.current.date` | unique_name | 0.75 | `Main.date` | 13 | no — `let date: Date?` |

And the structural invariant that is violated — `MATCH (a)-[r:CALLS]->(b:Variable)`:

```
callee                        resolved_to    line   parent
URLSession.shared.data        Main.data      5      Main.PickedFile
Calendar.current.date         Main.date      13     Main.Point
```

Rejecting any `CALLS` edge whose target node is not callable would catch two of the three here without touching the naming heuristics.

### On a real project

A 462-file / 64k-line Swift app. Every call into a system framework, with each resolution checked against the source by hand:

| callee | strategy | conf | resolved to | correct? |
|:--|:--|--:|:--|:--|
| `URLSession.shared.data` | suffix_match | 0.21 | `HomeboxUI.PickedFile.data` | no |
| `UserDefaults.standard.string` ×3 | unique_name | 0.75 | `AuthDTOs.HomeboxDateOnly.string(from:)` | no |
| `JSONEncoder().encode` | suffix_match | 0.14 | `MaintenanceDTOs.MaintenanceEntryCreate.encode(to:)` | no |
| `Calendar.current.date` ×2 | suffix_match | 0.55 | `StatisticsDTOs.date` (a `public let date: Date?`) | no |
| `Calendar.utcGregorian.startOfDayUTC` | unique_name | 0.75 | `AuthDTOs.Calendar.startOfDayUTC` | **yes** |

Seven of eight wrong. The eighth is genuinely correct — the project really does declare `extension Calendar { static let utcGregorian; func startOfDayUTC(for:) }`, so I am not counting it against the resolver.

### Likely root: Swift has no import or LSP tier

Resolution strategies actually used, same binary, two projects indexed minutes apart:

| strategy | Swift project (12,017 CALLS) | Go + TS project (20,650 CALLS) |
|:--|--:|--:|
| `import_map` | **0** | 5,381 |
| `lsp_*` (11 variants) | **0** | 2,478 |
| `unique_name` | 5,300 | 4,933 |
| `suffix_match` | 3,827 | 3,348 |
| `field_type_hint` | 1,673 | 51 |
| `qualified_suffix` | 723 | 3 |
| `same_module` | 494 | 4,246 |

Swift uses 5 strategies; Go/TS uses 21. With no import map and no LSP tier, there is nothing that can say `URLSession` is external, so name matching is the only thing left. The Swift graph also carries **no `Package` label at all**, where the Go/TS project has 215 `Package` nodes with an `external` flag.

### Scale, stated honestly

`suffix_match` confidence spread in the Swift project, all 22 distinct values: 1,151 edges at 0.55, the remaining **2,676 below 0.5** — 70% of `suffix_match` edges, 22% of all `CALLS` edges.

I am **not** claiming those 2,676 are all wrong. Sampling the 0.04 tier turned up `vm.load` → `EntityListViewModel.load`, which is plausibly correct despite the confidence. The demonstrable errors are the system-framework ones above. The confidence figures are here to show that a threshold is not the fix — the wrong edges sit at 0.75.

Repro: 2 files. Real project: 462 Swift files / 64,101 lines, 3,766 `Method` nodes, 12,017 `CALLS` edges.

CaptainMittens:
## What does this PR do?

A Swift project's `URLSession.shared.data` bound to its own `PickedFile.data`,
`UserDefaults.standard.string` to a DTO's `string`, and `JSONEncoder().encode`
to a model's `encode`. "Who calls this?" then answers with calls that never
happened.

The cause is that the two name-only resolution strategies match on the final
segment alone. A dotted callee whose **first** segment starts upper-case names a
type — `URLSession`, `Calendar`, `JSONEncoder` — and that receiver chain is
evidence the scorers were discarding.

`receiver_chain_admits()` now requires the candidate's own parent segment to
appear somewhere in that chain:

| callee | candidate | in the chain? | outcome |
|:--|:--|:--|:--|
| `URLSession.shared.data` | `HomeboxUI.PickedFile.data` | `PickedFile` — no | refused |
| `Calendar.utcGregorian.startOfDayUTC` | `AuthDTOs.Calendar.startOfDayUTC` | `Calendar` — yes | kept |

Three shapes pass through untouched, so ordinary code is unaffected:

- a callee with no separator, which has no chain to judge;
- a lower-case root, which names a *value* whose declared type the chain does
  not show — `vm.load`, `http.Get`, `os.path.join`;
- a name in capitals with an underscore, which is a constant holding a value
  rather than a type. This carve-out is measured, not guessed: without it the
  gate refused `ISO_4217_URL.lower` → `builtins.str.lower`, which is correct.
  `JSON` and `URL` carry no underscore and stay guarded.

The gate applies only at the two name-only exits of `resolve_name_lookup`.
`import_map`, `same_module` and `qualified_suffix` already carry real evidence
and are left alone.

Language agnostic by design — the registry holds no language, and every language
that writes receiver chains gains the same protection.

### Why the confidence score could not do this instead

The clearest errors were the **highest**-confidence ones. Swift's import map is
empty (system frameworks are never indexed nodes, so `cbm_pxc_build_import_map`
drops them), and the 0.5 penalty in `resolve_name_lookup` only applies when an
import map exists. So the worst calls kept the full 0.75. A threshold would have
cut correct edges and left these.

A "never point a call at a `Variable`" guard is also wrong: `Variable` is a legal
target on purpose (`cbm_label_is_registry_symbol`), one project here has 645 such
edges and another 984, and `Variable` nodes carry no `type` property, so nothing
in the graph separates a callable variable from stored `Data`.

Fixes #1893

## Validation

Reproduce-first. With the fix reverted and the five tests kept, exactly the two
refusal tests fail and the three guard tests already pass — which is the point,
since a guard test that only passes with the change proves nothing:

```
registry_receiver_chain_refuses_library_unique_name_issue1893   FAIL tests/test_pipeline.c:8799: r.qualified_name is not NULL
registry_receiver_chain_refuses_library_suffix_match_issue1893  FAIL tests/test_pipeline.c:8814: r.qualified_name is not NULL
registry_receiver_chain_keeps_project_extension_issue1893       PASS
registry_receiver_chain_ignores_lowercase_root_issue1893        PASS
registry_receiver_chain_ignores_bare_name_issue1893             PASS
```

### Blast radius, measured on three real codebases

The gate changes resolution for every language, so I counted what it refuses
rather than reasoning about it. Read from live graphs, applying the rule to every
`unique_name` and `suffix_match` edge:

| project | languages | `unique_name` | refused | `suffix_match` | refused |
|:--|:--|--:|--:|--:|--:|
| homebox-interface | Swift | 5300 | 278 (5.2%) | 3827 | 295 (7.7%) |
| homebox | Go, TypeScript, Python | 4933 | 14 (0.3%) | 3348 | 2 (0.06%) |
| Home_Server | Python, shell | 190 | 2 (1.1%) | 37 | 0 |

The Swift project is where the bug lives and it moves most. The Go/TypeScript
project loses 16 edges out of 8281 name-only ones — 0.2% — because its import
map already resolves most calls before the gate is reached.

I read every non-Swift refusal by hand. They are all wrong bindings the gate
should refuse:

```
Date.now            -> homebox.scripts.asc-build-state.now
JSON.parse          -> homebox.frontend.lib.datelib.datelib.parse
Array.from          -> homebox.backend.pkgs.mailer.test-mailer-template.from
Promise.all         -> homebox.docs...starlightChangelogs.versions.all
Object.entries      -> homebox.frontend.pages.checkbooks.[id].entries
Body.Close          -> homebox.backend.internal.data.ent.Close
Path(__file__).resolve -> Home_Server.homelab.scripts.withsecret.resolve
```

`import_map`, `same_module` and `qualified_suffix` counts are unchanged
everywhere, since the gate never runs on those paths.

### Suites

| Suite | Result |
|:--|:--|
| `pipeline` | 260 passed, 0 failed (255 before, plus the 5 new) |
| full `make -f Makefile.cbm test` | 7635 passed, 28 failed, 8 skipped |

The 28 failures are all in `tests/test_cli.c` (client install/uninstall) and are
pre-existing in my environment. Measured rather than assumed: `origin/main` here
runs 7630 passed / 28 failed, and this branch adds exactly the 5 new tests with
the same 28 failures in the same file. If they are green on your runners they
are environmental on mine.

Lint: `make -f Makefile.cbm lint-cppcheck` and `lint-no-suppress` both exit 0.
`clang-format` wants no change on any line this PR adds. I could not run
`clang-tidy` locally — it is not in my toolchain — so that one is unverified on
my side.

## Checklist

### DIFF
--- a/src/pipeline/registry.c
+++ b/src/pipeline/registry.c
@@ -888,6 +888,87 @@ static const char *qualified_suffix_match(const qn_array_t *arr, const char *cal
     return match;
 }
 
+/* A dotted callee whose FIRST segment starts upper-case names a type — URLSession,
+ * Calendar, JSONEncoder. That receiver chain is evidence the bare-name scorers
+ * throw away, and throwing it away binds Foundation's URLSession.shared.data to
+ * a project's own PickedFile.data: high confidence, and nothing in the graph
+ * shows it is wrong. Require instead that the candidate's own parent segment
+ * appears somewhere in the chain. Calendar.utcGregorian.startOfDayUTC resolving
+ * to AuthDTOs.Calendar.startOfDayUTC passes, because Calendar is in the chain.
+ *
+ * Only an upper-case first segment is guarded. A lower-case root names a value
+ * (vm.load, http.Get, os.path.join) whose declared type the chain does not
+ * show, so the chain proves nothing there and the call passes through
+ * unchanged. A callee with no separator passes through as well.
+ *
+ * Language agnostic by design: the registry holds no language, and every
+ * language that writes receiver chains gains the same protection. */
+static bool receiver_chain_admits(const char *callee_name, const char *candidate_qn) {
+    /* Normalize "::" -> "." so the chain composes with dotted candidate QNs,
+     * the same way qualified_suffix_match does. */
+    char dotted[CBM_SZ_512];
+    size_t w = 0;
+    for (const char *s = callee_name; *s && w + SKIP_ONE < sizeof(dotted);) {
+        if (s[0] == ':' && s[1] == ':') {
+            dotted[w++] = '.';
+            s += 2;
+        } else {
+            dotted[w++] = *s++;
+        }
+    }
+    dotted[w] = '\0';
+
+    const char *last_dot = strrchr(dotted, '.');
+    if (!last_dot) {
+        return true; /* bare name — no receiver chain to judge */
+    }
+    if (dotted[0] < 'A' || dotted[0] > 'Z') {
+        return true; /* lower-case root names a value, not a type */
+    }
+    /* A name written in capitals with underscores is a constant holding a
+     * value, not a type: ISO_4217_URL.lower is a string's own method. JSON and
+     * URL carry no underscore and stay guarded. */
+    int has_underscore = 0;
+    int all_caps = 1;
+    for (const char *c = dotted; c < last_dot && *c != '.'; c++) {
+        if (*c == '_') {
+            has_underscore = 1;
+        } else if (*c >= 'a' && *c <= 'z') {
+            all_caps = 0;
+            break;
+        }
+    }
+    if (all_caps && has_underscore) {
+        return true;
+    }
+
+    /* The candidate's parent segment: the one before its final name. */
+    const char *cand_last = strrchr(candidate_qn, '.');
+    if (!cand_last || cand_last == candidate_qn) {
+        return true; /* top-level candidate — no parent to look for */
+    }
+    const char *parent = cand_last;
+    while (parent > candidate_qn && parent[-1] != '.') {
+        parent--;
+    }
+    size_t parent_len = (size_t)(cand_last - parent);
+
+    /* Walk the chain — every segment before the final callee name. A trailing
+     * "()" is dropped so JSONEncoder().encode reads as JSONEncoder. */
+    for (const char *seg = dotted; seg < last_dot;) {
+        const char *end = strchr(seg, '.');
+        size_t len = (size_t)(end - seg);
+        if (len >= 2 && seg[len - 2] == '(' && seg[len - 1] == ')') {
+            len -= 2; /* an empty "()" — JSONEncoder().encode names JSONEncoder */
+        }
+        if (len == parent_len && strncmp(seg, parent, parent_len) == 0) {
+            return true;
+        }
+        seg = end + SKIP_ONE;
+    }
+    return false;
+}
+
 /* Strategy 3+4: Name lookup + suffix match */
 static cbm_resolution_t resolve_name_lookup(const cbm_registry_t *r, const char *callee_name,
                                             const char *module_qn, const char **import_vals,
@@ -913,6 +994,9 @@ static cbm_resolution_t resolve_name_lookup(const cbm_registry_t *r, const char 
 
     /* Strategy 3: unique name */
     if (arr->count == SKIP_ONE) {
+        if (!receiver_chain_admits(callee_name, arr->items[0])) {
+            return empty_result();
+        }
         double conf = CONF_UNIQUE_NAME;
         if (import_vals && import_count > 0 &&
             !is_import_reachable(arr->items[0], import_vals, import_count)) {
@@ -927,6 +1011,9 @@ static cbm_resolution_t resolve_name_lookup(const cbm_registry_t *r, const char 
     }
     const char *best = best_by_import_distance((const char **)arr->items, arr->count, module_qn);
     if (best) {
+        if (!receiver_chain_admits(callee_name, best)) {
+            return empty_result();
+        }
         double conf = candidate_count_penalty(CONF_SUFFIX_MATCH, arr->count);
         return (cbm_resolution_t){best, "suffix_match", conf, arr->count};
     }
--- a/tests/test_pipeline.c
+++ b/tests/test_pipeline.c
@@ -9275,6 +9275,78 @@ TEST(registry_confidence_suffix_match) {
     PASS();
 }
 
+/* Issue #1893: a call on a library type bound to a same-named project member.
+ * URLSession is Foundation's, not this project's, so PickedFile.data is the
+ * wrong target — and with one candidate it won the top name-only confidence. */
+TEST(registry_receiver_chain_refuses_library_unique_name_issue1893) {
+    cbm_registry_t *reg = cbm_registry_new();
+    cbm_registry_add(reg, "data", "HomeboxUI.PickedFile.data", "Variable");
+
+    cbm_resolution_t r =
+        cbm_registry_resolve(reg, "URLSession.shared.data", "HomeboxUI.Net", NULL, NULL, 0);
+    ASSERT_NULL(r.qualified_name);
+
+    cbm_registry_free(reg);
+    PASS();
+}
+
+/* The same refusal on the other name-only exit, where several candidates share
+ * the final name and import distance picks the winner. */
+TEST(registry_receiver_chain_refuses_library_suffix_match_issue1893) {
+    cbm_registry_t *reg = cbm_registry_new();
+    cbm_registry_add(reg, "data", "HomeboxUI.PickedFile.data", "Variable");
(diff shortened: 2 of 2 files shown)


DeusData:
Approved, and I want to be explicit that this cleared a bar most changes of its shape do not.

There is a standing position here that a precision fix restricting receiver resolution **globally** is the wrong shape — suppressors are written as per-language helpers gated on the file language, and all three existing ones (Perl, TS/JS, cross-language suffix) are built that way. This PR is global. It was accepted anyway, on four things it does that the usual version of this change does not:

**The upstream strategies are provably untouched.** I checked rather than took it: `resolve_import_map` and `resolve_same_module` both run *before* `resolve_name_lookup` and win when they hit, so gating only the two name-only exits leaves `same_module`, `import_map` and `qualified_suffix` alone. That matters more than it looks — the specific casualty this rule exists to prevent is a Go receiver method, which carries a flat package QN and resolves *only* through the `same_module` suffix branch. Your gate never reaches it.

**You measured the blast radius instead of arguing it.** Three real codebases, split by language, with every non-Swift refusal read by hand. 16 edges out of 8281 name-only ones on the Go/TypeScript project is the number that settles this, and `Date.now → asc-build-state.now`, `Body.Close → ent.Close`, `Promise.all → starlightChangelogs.versions.all` are all plainly wrong bindings. A blocked high-confidence match does not vanish here, it just stops being invented.

**Your reason for going global is factually right.** `cbm_registry_resolve` takes no language parameter — I confirmed the signature. The per-language alternative is a post-resolution suppressor at the `pass_calls.c` / `pass_parallel.c` seam, and that would silently miss any other consumer of the resolver.

**And you ruled out the cheaper fix with evidence.** "The clearest errors were the highest-confidence ones" is the sentence that closes off the threshold approach: Swift's import map is empty, so the 0.5 penalty never applies and the worst edges kept the full 0.75. A confidence cut would have removed correct edges and left these.

## Two smaller things worth naming

**Your guard tests pass without the change, deliberately.** "A guard test that only passes with the change proves nothing" is exactly right, and showing the reverted run with 2 FAIL / 3 PASS is the clearest way to demonstrate a gate that refuses the right things without over-refusing. Most reproduce-first evidence stops at the failing half.

**The `ISO_4217_URL` carve-out is measured, not guessed.** Excluding capitals-with-underscore because the gate otherwise refused `ISO_4217_URL.lower → builtins.str.lower`, while noting `JSON` and `URL` carry no underscore and stay guarded, is the kind of boundary that only comes from running it.

## Status

Cleared, reviewed, approved. `main` was briefly broken earlier today and repaired by #1993; your green predates that, so I will re-run this against current `main` before merging. Nothing needed from you.

Thank you — this is the most carefully evidenced call-resolution change we have had.

DeusData:
Merged as 91a17231 — thank you! Refusing a name-only match that the receiver chain contradicts is the right call for graph quality: a wrong CALLS edge is worse than a missing one, and this keeps the suppressor local to the receiver logic rather than a global recall trade. #1893 closes with it. Great work.

PR #1907, opened 2026-08-29
fix(pipeline): suppress weak short-name matches for Go selector calls

### LINKED ISSUE #1906: Go: weak short-name strategies fabricate CALLS edges for selector calls (stdlib receivers hijack project symbols)
**Status: fix open — PR #1907** (extractor `is_method` flag for Go selector calls + Go-gated `cbm_go_suppress_weak_method_match`). Full post-fix census inline below.

### Version

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

A Go selector call whose receiver the Go LSP cannot type falls through to the generic registry, which binds it by bare short name to an arbitrary same-named **project** symbol (`suffix_match` / `unique_name`). Stdlib/third-party calls are the worst case: the real callee is not in the tree, so every candidate is wrong by construction. Go never sets `is_method`, so the receiver-aware guard TS/JS got in #592/#606 (and Perl in #476) never fires.

Measured (private Go monorepo, ~19.6k CALLS): `suffix_match` 3817 @ avg 0.30, `unique_name` 3162 @ avg 0.53 — 36% of all CALLS from name guesses. One 14-line checksum helper touching only `os`/`io`/`crypto/sha256` got 3/3 false outbound edges: `f.Close()` → a project `Close` (conf 0.11, 15 candidates), `sha256.New()` → an unrelated `New`, `io.Copy` → the sole project `Copy` (`unique_name` 0.375). `trace_path` surfaces all of it unfiltered, so `Close`/`New`/`Run` accumulate dozens of phantom callers.

Expected: no CALLS edge for a call the resolver cannot place; typed, local, and import-qualified calls keep resolving.

### Reproduction

Dummy fixture: `go.mod` (`module example.com/myapp`) + `storage/storage.go` (the lone project `Close` method + a typed same-package `Boot` caller) + `hash/hash.go`:

```go
package hash

import (
	"os"

	"example.com/myapp/util"
)

func FileLen(path string) int64 {
	f, err := os.Open(path)
	if err != nil {
		return 0
	}
	defer f.Close() // ← false CALLS edge: hash.FileLen -> storage.Close
	st, err := f.Stat()
	if err != nil {
		return 0
	}
	return st.Size()
}

func UsesUtil() string { return util.Tag() } // import-qualified: must keep resolving
```

Actual: `FileLen | Close | {"strategy":"unique_name","confidence":0.75,…}` (with ≥2 candidates: `suffix_match` @ 0.1–0.4). Expected: no `FileLen → Close`; `Boot → Close`, bare local calls, and `UsesUtil → Tag` all survive.

### Two Go-specific constraints (why the TS/JS drop-list can't be reused as-is)

- `field_type_hint` is kept for Go — but see #1927: its receiver-awareness premise did not hold and is fixed separately (#1936).
- `unique_name` drops **only when penalized** (`CONF_UNIQUE_NAME` scaled 0.75 → 0.375 when the lone candidate is outside the caller's import closure — the stdlib-hijack shape). Unpenalized lone candidates are genuinely typed and must keep resolving.

### Post-fix strategy census (same tree indexed with v0.10.8 vs Stack A = #1907+#1913+#1915)

| strategy | v0.10.8 | with the fixes | Δ |
|---|---|---|---|
| `lsp_strategy_cross_file` | 5273 | 6468 | +23% |
| `lsp_direct` | 3033 | 3372 | +11% |
| `lsp_type_dispatch` | 137 | 687 | +401% |
| `lsp_interface_dispatch` | 17 | 79 | +365% |
| `lsp_embed_dispatch` | — | 42 | new |
| `import_map` | 284 | 88 | −69% |
| `field_type_hint` | 493 | **1484** | **+201%** — see #1927/#1936 |
| `unique_name` | 3162 (avg 0.53) | 1573 (avg 0.68) | −50% |
| `suffix_match` | 3817 (avg 0.30) | **113** | **−97%** |
| `same_module` | 3267 | **224** | **−93%** |
| **CALLS total** | **19653** | **14316** | −27% |

Weak short-name share (`suffix_match`+`unique_name`+`callee_suffix`): 36.0% → **12.4%**, with `lsp_*` rising 8481 → 10672 in absolute terms — the edges were re-resolved, not merely deleted. The `field_type_hint` growth is the direct consequence tracked as #1927 (guesses moved from a visibly-weak 0.30 bucket into a 0.85 one); fixed by #1936.

Related: #592/#606 (TS/JS), #476 (Perl), #1893 (Swift), #1276 (Python), #1114; tracked in #1932.

### Confirmations

ilyabrykau-orca:
## What does this PR do?

Fixes #1906.

Extends the TS/JS receiver-aware weak-match guard (#592/#606) to Go: a selector call `x.foo()` whose receiver the Go LSP cannot type must not be bound by a receiver-blind short-name strategy to an arbitrary same-named project symbol (`f.Close()` on an `*os.File` → a project `Close`; on the repo measured in #1906 these strategies were 36% of all CALLS edges, and one 14-line stdlib-only helper had 3/3 false outbound edges).

Mirrors how TS/JS joined the Perl guard (#476):

1. `internal/cbm/extract_calls.c` — flag Go `call_expression` with a `selector_expression` callee as `is_method`, same block as the TS/JS `member_expression` flag (Go has no `this`/`super` analog to exempt).
2. `src/pipeline/registry.c` + `pipeline.h` — new `cbm_go_suppress_weak_method_match(is_go, is_method, strategy, confidence)`. Drops `suffix_match`/`fuzzy` always; drops `unique_name` only when its confidence carries the import-unreachability penalty (< `CONF_UNIQUE_NAME`) — the stdlib-hijack shape. `field_type_hint` is deliberately **kept**: Go struct fields carry declared types, so the hint is receiver-aware there (`lrp_go_s8_field_type_hint` stays GREEN).
3. `src/pipeline/pass_calls.c` / `pass_parallel.c` — feed the Go gate next to the TS/JS one. The drop defers to the emit path exactly like TS/JS, so service/route/HTTP/CONFIG edges stay main-identical. `same_module`, `import_map`, `qualified_suffix`, `callee_suffix` and all `lsp_*` strategies are untouched: typed receivers, same-package calls, and import-qualified calls resolve exactly as before.
4. Tests (reproduce-first):
   - `pipeline_go_receiver_suppresses_weak_method_edge` — RED without the extractor flag (`FAIL: ASSERT(!(cross_file_call_exists(s, project, "FileLen", "Close")))`), GREEN with it. Asserts the stdlib `f.Close()` edge is gone while the typed same-package `s.Close()`, a bare local call, and an import-qualified cross-package call keep resolving. The fixture carries a `go.mod` — import reachability (the `unique_name` penalty) depends on it, like every real Go repo.
   - `extract_go_selector_call_flags_is_method` pins the extractor flag; the old flag-exempt contract test keeps its guard using Python as the exempt language.
   - `go_suppress_drops_weak_selector_matches` / `go_suppress_keeps_typed_and_import_aware_matches` unit-pin the drop-list, the 0.75-vs-0.375 `unique_name` split, and the `lsp_*` keeps.

Perf/memory: one `ts_node_child_by_field_name` + two `strcmp` per Go call site at extraction, one short-circuited predicate per resolved Go call, zero allocations, no new passes.

Known residual (documented in #1906, out of scope here): a Go file whose import table maps zero project packages has no penalty signal, so a lone same-named project symbol can still capture a stdlib method call there; `same_module` can still mis-bind interface-typed calls inside one package. Both need receiver typing, not name heuristics.

## Checklist

- [x] Lint passes (`git clang-format --diff HEAD` clean on changed lines; clang-tidy/cppcheck via CI)


ilyabrykau-orca:
Rebased onto current main (post-#1903). #1903 generalized the TS/JS guard into `cbm_suppress_weak_member_match` with per-site language gates — exactly the shape this PR anticipated — so the Go integration now composes with that helper at both call sites via its own predicate (`cbm_go_suppress_weak_method_match`; different drop-list: field_type_hint kept, unique_name only when import-unreachability-penalized). The flag-exempt extraction contract test moved from Go to Rust, since Python joined the flagged set in #1903 and Go joins here. Full `scripts/test.sh` leg green on the rebased stack.

ilyabrykau-orca:
Retriggered CI with a no-op amend (identical tree, new SHA — I lack rerun rights on this repo). The two failures on the previous run look environmental, not change-related: `test_watcher.c:2526 index_call_count == 0, expected 1` on macos-15-intel (a watcher-timing probe; this PR touches no watcher code), and windows-guards dying in its own harness setup (`SETUP FAIL: ASCII baseline did not index`) before reaching any guard. No semantic changes; the local full venue leg is green.

DeusData:
Reviewed. This is the right shape, and the care in it is above the bar for this class of change. One interaction with your own #1913 needs measuring before either lands.

## The guard itself

`cbm_go_suppress_weak_method_match` is exactly the form this repo wants: a pure per-language predicate gated on the file language, wired at **both** `pass_calls.c` and `pass_parallel.c` with the lockstep note, and — the part that matters most — it leaves `same_module`, `import_map`, `qualified_suffix`, `callee_suffix` and every `lsp_*` strategy alone. High-confidence paths are untouched.

Two decisions in it are better than the guard you mirrored:

**Keeping `field_type_hint` for Go.** The TS/JS guard drops it as a heuristic; you kept it with a reason — a Go struct field carries a declared type, so the hint is receiver-aware there rather than a guess. That is the right distinction and `lrp_go_s8_field_type_hint` holds you to it.

**Dropping `unique_name` only when penalized.** This is the subtle one, and the comment earns it: an unpenalized `unique_name` target sits inside the caller's import closure, so dropping it would kill genuinely-typed lone-candidate calls that never reach the field-type-hint upgrade. Gating on `confidence < CONF_UNIQUE_NAME` rather than on the strategy name alone is what makes this a receiver-blindness guard instead of a blunt recall cut.

Documenting the residual — a file whose import table maps zero project packages has no penalty signal — instead of implying the class is closed, is the right way to leave it.

## The measurement I want, and it is about this PR *plus* #1913

These two are in the same stack and both narrow Go call resolution, and I do not think either PR's numbers capture what they do **together**.

Walk one call through both:

1. Today, a Go method call can resolve via `same_module` at confidence 0.90.
2. **#1913** receiver-qualifies method QNs, so `resolve_same_module`'s `module_qn + "." + callee_name` candidate no longer matches — that call falls through to `resolve_name_lookup`.
3. `resolve_name_lookup` may return `suffix_match`.
4. **This PR** drops `suffix_match` outright for Go selector calls.

So a call that was a confident edge before the stack becomes **no edge at all** after it — and neither PR sees that on its own. #1913 measures the false edges it removes; this one measures the weak edges it removes. Neither measures the calls that arrive at step 4 only *because* of step 2.

I have asked #1913 for a Go CALLS census before/after with a strategy histogram. What I want here is the same measurement taken on **the stack as a whole**, not slice by slice: Go edge count and strategy histogram on `main`, then with #1907 alone, then with #1907+#1913. If the compounding loss is small, that settles it and both merge on the numbers. If a meaningful population lands in that gap, it is much cheaper to know now than after a reindex.

The per-language census with its dumped per-edge sets is the instrument, and the Go leg is the baseline.

None of this is a doubt about the direction. `f.Close()` on an `*os.File` binding to a project `Close` is plainly wrong, 36% of CALLS edges arriving through receiver-blind strategies is plainly too many, and a 14-line stdlib-only helper with 3/3 false outbound edges is a good demonstration. I want the stack's net effect on the record, not just each slice's gross.

## Status

`ms` is `DIRTY` — this and both stack members above it (#1913, #1915) have conflicts. `main` was briefly broken today and was repaired by #1993, so this is a good moment to rebase the whole stack.

Cleared. Approved on shape, pending the stack-level numbers.

ilyabrykau-orca:
Rebased onto current main. Only conflict was test insertion-anchor drift in tests/test_pipeline.c (both merges added tests at the same anchor); every source file auto-merged. Full scripts/test.sh (ASan+UBSan) is green on the rebased stack head (#1936, which contains this commit).

ilyabrykau-orca:
Stack-level census is up — full table and per-edge migration matrix posted on #1913 (they share one instrument run, so the numbers live in one place). The short version for the interaction you asked about:

**The step-2→step-4 cascade is small.** Across the #1913 leg, `suffix_match` grows by only 25 keys (155 → 180) — the calls that fall out of `same_module` under receiver-qualified QNs do not arrive at the suffix fallback in any volume, so this PR's drop-list removes almost nothing that #1913 newly routed there. Of #1907's 929 `same_module` edges: 108 stay, 444 re-land (mostly `field_type_hint` — receiver-typed, and #1936 later re-judges those by owning segment), 377 vanish — and the vanished population is the untyped-receiver binds #1913 exists to stop trusting.

**The compounding loss the numbers did surface was elsewhere**: a third QN site (go_lsp's enclosing-function QN) that #1913 had not moved, which cost method-body callers their LSP caller join (568 `lsp_type_dispatch` + 105 `lsp_direct` keys in the first measurement). Fixed inside #1913 with a reproduce-first test; with the fix the stack's LSP strategies end **above** the pre-stack baseline (`lsp_direct` 8896 → 9124, `lsp_type_dispatch` 2551 → 2640 at the stack head).

Whole-stack net: 21018 → 13815 Go CALLS, of which −6403 is this PR's weak-edge removal and −1019 is #1936's substring-hint removal; receiver-blind strategy share drops 43% → 10.5%. Stack rebased and force-pushed, full scripts/test.sh green on the head.

DeusData:
Apologies — I assessed this some time ago and **never posted the assessment**. There is no comment and no review on this PR, so as far as you could tell it had been ignored. That is my error, and here is the review it should have had.

## Approved on shape

This is the correct shape for a receiver-precision fix in this project, and it matches every constraint we hold guards to:

- **Language-gated**, `cbm_go_suppress_weak_method_match(is_go, …)`, so no other language is affected.
- **Wired at both `pass_calls.c` and `pass_parallel.c`.** That twin-resolver split is the standing trap in this area — a guard added to one and not the other behaves differently depending on file count — and you covered both.
- **Mirrors the established lineage** (Perl → TS/JS via #476/#592/#606) rather than inventing a fourth pattern.
- **`same_module` preserved.** That branch is what Go receiver methods resolve through, because they carry a flat qualified name (`pkg.Area`, not `pkg.Circle.Area`), so gating it off would lose correct edges.

**Keeping `field_type_hint` is the detail that convinced me.** Dropping every weak strategy would have been the easy move; recognising that Go struct fields carry declared types, so the hint is genuinely receiver-aware there, is the harder and more accurate call — and `lrp_go_s8_field_type_hint` staying green is the evidence rather than the assertion.

**And the `unique_name` treatment is properly conditional**: dropped only when its confidence carries the import-unreachability penalty, which is the stdlib-hijack shape, rather than always. That is the difference between a precision fix and a recall cut.

Your measurement answers what I was going to ask for: these strategies accounted for 36% of all CALLS edges on the measured repo, and a 14-line stdlib-only helper had 3 out of 3 false outbound edges. That is a census, not an anecdote.

## Your two reds are ours

`test-windows-guards` fails on `test_daemon_stability.py` with the signature we now see across several unrelated PRs — a `REGRESSION` line preceded by:

```
```

Those two setup failures appear in **every** instance of this red, on three different sections across unrelated changes, which points at one shared setup fault rather than a daemon regression your diff could have caused. `ci-ok` only aggregates it.

Nothing to change. I will merge once that guard is settled or the run comes back clean — and again, sorry for the silence.

DeusData:
Decision posted on #1913 — direction accepted on our own census. Please rebase this one first (it is the base of the stack), then #1913. Thanks again for the patience.

ilyabrykau-orca:
Rebased onto current main (8972ea69, post-memory-core) per the decision on #1913. The only source conflict was additive: main gained the Python local-binding guard (`cbm_suppress_weak_local_binding_call`) at the same anchor points, so `drop_plain_call` now chains all three guards in both resolvers, and the new upstream tests are kept alongside this PR's. Full `scripts/test.sh` green locally (ASan+UBSan), `git clang-format --diff` clean. #1913 follows next.

DeusData:
Heads-up: I cancelled this PR's CI run on purpose a minute ago. It was sharing the hosted-runner pool with tonight's v0.11.0 release build, and the release needed the slots. Nothing about the PR is wrong. I will re-run the full CI as soon as the release is out, so any red you see right now is the cancellation, not your change. Thanks for your patience, and for the steady stream of Go fixes.

PR #1913, opened 2026-08-29
fix(extract): receiver-qualify Go method QNs

### LINKED ISSUE #1909: Go: method QNs ignore the receiver — every same-name method in a package collapses onto one node
**Status: fix open — PR #1913** (receiver-qualified Go method QNs, mirroring the C++ out-of-line block; stacked on #1907). Post-fix: `Close` methods 15 → 86 nodes, `Process` 12 → 47, `Task` 1 → 8; one generated dir recovered 1 → 63 `Close` methods (source 66; the residual 3 are #1911's twins). Generic receivers verified on real code: 45 methods across 8 generic receivers carry clean receiver-qualified QNs (`pkg.Cache.Add` for `func (c *Cache[K, V]) Add`) and zero node QNs contain a `[`. Changes extracted QNs for Go methods → reindex needed (flagged per CONTRIBUTING).

### Version

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

Go concrete methods get a flat, package-scoped QN (`proj.pkg.method` — receiver ignored), so every same-name method in a package collides on one QN; `cbm_gbuf_upsert_node` keeps exactly one node and the losers vanish silently (parse coverage stays clean, so `check_index_coverage` cannot see it).

Measured (private Go repo): 20 `Process`/`Name` declarations across 15 files kept **2** nodes; 9 distinct `Task()` methods fused into one chimera node carrying all 9 bodies' CALLS edges (26 with line numbers outside the surviving body); 19 structs point `DEFINES_METHOD` at the same single `Process`; a `_test.go` mock `Close` won the smallest-file-path tie-break over the production `Close`, so callers of the "production" symbol got filtered as test noise.

Interface members already get `proj.pkg.Iface.method`; C++ out-of-line methods already get `proj.file.Class.method`; the extractor already computes the receiver type for `parent_class` — it just doesn't put it in the QN.

Expected: `func (s *Storage) Close()` → `proj.pkg.Storage.Close`, one node per receiver.

### Reproduction

```go
// x.go
package m

type Storage struct{}
type Cache struct{}

func (s *Storage) Close() {}
func (c Cache) Close() {}
```

Actual: **one** Method node `m.Close`, both types' `DEFINES_METHOD` pointing at it. Expected: `m.Storage.Close` + `m.Cache.Close`.

### Secondary effect this also fixes

`resolve_same_module` (conf 0.90) is an exact `module.name` lookup. Flat method QNs occupy that key, so `pipeline.Process(event)` binds whatever unrelated `Process` merged into the caller's package — a high-confidence false edge no filter can remove. Receiver-qualified QNs delete the flat key; those calls fall through to the typed/guarded paths.

### Fix notes (all consumers already agree)

`pxc_build_lsp_def` passes def QNs verbatim (LSP registries follow automatically); `check_go_class_implements` already supports class-qualified method QNs; the call-scope side (`compute_func_qn`) mirrors the def formula via one shared receiver-type helper.

Related: #1906 (the merge made weak edges land on high-traffic symbols), #1114; tracked in #1932.

### Confirmations

ilyabrykau-orca:
## What does this PR do?

Fixes #1909. **Stacked on #1907** (shares the Go test scaffolding in `tests/test_extraction.c`; the first commit here is #1907's — review only the last commit).

Go concrete methods carried a flat package QN (`proj.pkg.method`, receiver ignored), so every same-name method in a package collided on one QN and the graph upsert kept exactly one node — on the repo measured in #1909: 20 `Process`/`Name` methods → 2 nodes, 9 `Task()` methods fused into one chimera node carrying all nine bodies' call edges, 19 structs `DEFINES_METHOD`-linked to a single shared node, and a `_test.go` mock outranking the production method in the dedupe tie-break. The upsert's own comment calls kind-disambiguated QNs "the real cure".

The change — same shape as Go interface members and the C++ out-of-line method path directly below it in `extract_func_def`:

1. `internal/cbm/extract_defs.c` — when the receiver type resolves, `def.qualified_name = parent_class + "." + name` (`proj.pkg.Storage.Close`); `go_receiver_type_name` becomes the shared `cbm_go_receiver_type_name` (declared in `helpers.h`, following the `cbm_cpp_out_of_line_parent_class` precedent) so both sides of the contract use one formula.
2. `internal/cbm/extract_unified.c` (`compute_func_qn`) — mirror branch for Go `method_declaration`, so method-body calls keep exact source attribution instead of degrading to the `calls_find_source` File-node fallback.
3. No resolver changes needed — consumers already agree:
   - `pxc_build_lsp_def` passes the def QN and `parent_class` (→ `receiver_type`) verbatim into the Go LSP registries, so `lsp_type_dispatch` / `lsp_embed_dispatch` / interface-satisfaction emissions follow the new QN automatically (all `lrp_go_s*` probes stay GREEN, untouched).
   - `check_go_class_implements` explicitly supports class-qualified method QNs (its path (b) reconstructs `<ClassQN>.<method>`; its comment describes the flat QN as the anomaly).

Side effect (also #1909): `resolve_same_module`'s exact `module.name` hash no longer matches concrete methods, which removes the confidence-0.9 false edges where an interface-typed call (`pipeline.Process(event)`) bound to an unrelated same-package method.

Tests:
- `extract_go_method_receiver_qualified_qn` — two same-name methods on different receivers get distinct, receiver-qualified QNs with matching `parent_class`; free functions keep the flat QN.
- `extract_go_no_filename_in_module_qn` updated: the method expectation becomes `proj.myapp.db.Conn.Query` (its actual contract — no filename segment in the QN — still asserted).
- Full `scripts/test.sh` leg green (ASan+UBSan, "All tests passed"), including all `lsp_resolution_probe` Go scenarios and the #1907 pipeline test, with zero probe changes.

Note for maintainers: this changes extracted QNs for Go methods, so existing graphs need a reindex to pick up the new shape (flagged in #1909 per CONTRIBUTING's indexing-change rule).

## Checklist

- [x] Lint passes (`git clang-format --diff` clean on changed lines; clang-tidy/cppcheck via CI)


ilyabrykau-orca:
Rebased onto current main (stacked on the rebased #1907). No semantic changes; full test leg green. Post-fix field numbers for this change are in the #1906 census comment (method-node recovery incl. generic receivers: https://github.com/DeusData/codebase-memory-mcp/issues/1906#issuecomment-5468064499).

DeusData:
Reviewed the last commit. The bug is real and badly worth fixing, the shape of the fix is right, and there is one measurement I want before it lands.

**The problem is as bad as you say.** `store.c:257` declares `UNIQUE(project, qualified_name)` and the node upsert is `ON CONFLICT(project, qualified_name) DO UPDATE SET`, so a flat `proj.pkg.method` genuinely makes every same-name method in a package one row. Nine `Task()` methods fusing into a single node carrying all nine bodies' call edges is not a cosmetic count problem — that node is a chimera whose outgoing edges are attributed to code that never made them. And a `_test.go` mock outranking the production method in the dedupe tie-break is the kind of thing that quietly poisons every answer downstream.

Receiver-qualifying is the right cure, matching the Go interface-member and C++ out-of-line paths already in `extract_func_def`, and lifting `go_receiver_type_name` to a shared `cbm_go_receiver_type_name` so both sides of the contract compute one formula is exactly right — two copies of that formula would drift.

## The measurement I want

You describe this as a side effect, in one line:

> `resolve_same_module`'s exact `module.name` hash no longer matches concrete methods, which removes the confidence-0.9 false edges where an interface-typed call bound to an unrelated same-package method.

I think that line is doing more work than it looks. `resolve_same_module` (`registry.c:790`) builds its candidate as `module_qn + "." + callee_name`. Before this change a call to `Close` inside `proj.pkg` hit `proj.pkg.Close` at confidence 0.90. After it, the method is `proj.pkg.Storage.Close` and that candidate misses — and the `suffix` retry at `:795` builds the same shape, so it misses too.

So `same_module` stops resolving **every** Go concrete-method call, not only the wrong ones. The true ones do not disappear; they fall through to `resolve_name_lookup`, which picks a winner by import distance project-wide at a much lower confidence. That is the failure mode this repo has been bitten by before: a blocked high-confidence match can lose a correct edge and mint a wrong one in the same step, and the net can be negative even when every individual removal was justified.

You measured the false edges removed. **I would like the other half measured too**, and there is already an instrument for it — the per-language CALLS census with its dumped per-edge sets. The Go leg of that is the right baseline. Concretely, what I am asking for is:

- Go CALLS edge count before and after, and
- the strategy histogram — specifically how many edges moved out of `same_module` and where they landed.

If the answer is "the LSP type-dispatch path already covers them, so the demotions are few and the removals are many", that settles it and the PR merges on the strength of the numbers rather than on the argument. If a meaningful number of true edges demote to import-distance guessing, that is worth knowing before this is in `main` rather than after.

The `lrp_go_s*` probes staying green is real evidence and I am not dismissing it — but they are a fixed scenario set, and the population this affects is every Go method call in a repo without LSP coverage on the receiver.

## Two smaller things

**This shifts a baseline other PRs are being judged against.** There is an open cluster of call-resolution PRs that we have decided to judge against one shared census precisely because whichever lands first makes the rest unreadable on their own numbers. A change to Go method QNs moves that baseline for the Go leg. Not an objection — just something I would rather sequence deliberately than discover afterwards.

**Your reindex note is right and needs to travel further than the PR.** A QN-shape change means existing graphs pick up the new shape only on rebuild, and that belongs in the release notes, not only in #1909. I will make sure it gets there.

## Status

`ms` is `DIRTY` — this and both its stack neighbours (#1907, #1915) have conflicts. `main` was broken for a few hours today and was repaired an hour ago by #1993, so now is a good moment to rebase the stack.

Thanks for the measurements you did bring — "20 methods → 2 nodes, 19 structs linked to one shared node" is what turns this from a plausible tidy-up into an obvious defect.

ilyabrykau-orca:
Rebased onto current main as part of the 1907→1913→1915→1936 stack; no source conflicts, test anchors re-anchored. Full scripts/test.sh green on the stack head.

ilyabrykau-orca:
Here is the measurement — and it caught something real, so thank you for insisting on it.

## The census (orca-runtime-sensor, Go→Go CALLS, isolated one-shot indexes)

| | main | +#1907 | +#1907+#1913 | full stack (+#1915+#1936) |
|---|---|---|---|---|
| **total** | 21018 | 14615 | **14812** | 13815 |
| `lsp_direct` | 8896 | 8896 | **9130** | 9124 |
| `lsp_type_dispatch` | 2551 | 2551 | **2640** | 2640 |
| `lsp_interface_dispatch` | 274 | 274 | 287 | 287 |
| `same_module` | 931 | 931 | **117** | 145 |
| `field_type_hint` | 676 | 676 | 1205 | 186 |
| `unique_name` | 2944 | 940 | 936 | 936 |
| `suffix_match` | 4554 | 155 | 180 | 180 |

Per-edge migration of #1907's 929 `same_module` keys (caller + callee-name identity, so chimera retargeting doesn't count as loss): **108 kept** `same_module` (receiver still resolvable flat), **444 re-landed** (308 `field_type_hint`, 97 `lsp_strategy_cross_file`, 39 `unique_name`), **377 lost entirely**. The lost population is exactly the calls whose receiver cannot be typed — the ones `same_module` was binding at 0.90 by package-coincidence, which is this PR's thesis. The LSP strategies keep ≥96% of their edges and end **above** the pre-#1913 baseline. Net caller→callee pairs across the leg: +41.

## What the first run of this census caught

My first measurement showed `lsp_type_dispatch` losing 568 edges and `lsp_direct` 105 — every sample was a **method-body caller** (`(r *Repo) Install` calling `r.helper()`). Root cause: `go_lsp.c`'s `process_function` still set the enclosing-function QN to the bare `package.func` form; its own comment declared the contract ("must be the QN the textual call events and the defs pass produce"), and this PR had moved that shared form to `package.Type.method` on the other two sides but not this one. Every LSP resolution sourced from inside a method lost its caller join in `cbm_pipeline_find_lsp_resolution` and fell back — or died.

Amended into this PR (it is the same one-formula claim): `process_function` now receiver-qualifies via the shared `cbm_go_receiver_type_name`, with a reproduce-first pipeline test (`pipeline_go_method_caller_keeps_lsp_join`: same-receiver method call with a colliding decoy method in another package, RED without the go_lsp hunk, GREEN with it). The table above is measured with the fix; without it the leg read 14156 total with `lsp_direct` at 8263.

One upstream test adopted the new shape: `parallel_go_cross_package_field_chain_resolves` matched edges by the flat `handler.PlaceOrder` fragment; the edge it guards is intact — `OrderHandler.PlaceOrder → OrderService.PlaceOrder`, still `lsp_type_dispatch` at 0.95, now receiver-qualified on both endpoints (which also confirms the cross-package chain path under the join fix) — so its fragments moved to the qualified form.

On the baseline-shift point: agreed — these numbers are from the stack in its merge order (#1907 → #1913 → #1915 → #1936), so whichever leg lands first, the others' columns above stay the readable reference.

Full scripts/test.sh green on the amended stack head; branch force-pushed with the rebase.

DeusData:
Census received, and thank you for insisting on measuring rather than arguing it — it caught exactly the thing I was worried about.

Where this stands: the result changes the shape of Go method qualified names (`pkg.Type.method` where `main` has `pkg.method`) and drops 377 `same_module` edges in exchange for the LSP-side gains. That is a graph-contract decision, not a review nit, and the standing rule here is that call-resolution changes are judged per language against one census rather than one PR at a time. So this is now a direction call on our side, with your census as the primary input — I will bring the answer back to this thread.

**Please hold any further rebase of the stack (#1907 → #1913 → #1915 → #1936) until that call is made.** `main` has moved again since your last one and the stack conflicts, but rebasing now would cost you a pass that may need redoing depending on the decision. Nothing else is needed from you meanwhile.

DeusData:
Thank you for holding the rebase while we measured this ourselves — here is the result, and the decision.

**We ran our own census** on kubernetes/kubernetes @ `8c078bc3` with four builds (the merge-base, #1907 alone, #1907+#1913, the whole stack), diffed the full Go CALLS edge sets, and hand-classified 150 sampled edges against the source.

| build | Go CALLS | what changed |
|---|---|---|
| base | 318,635 | |
| #1907 alone | 237,347 (−25.5%) | purely subtractive; all `suffix_match`/`unique_name`; **80% of sampled removed edges were false** (`os.ReadFile` → a struct field, `fmt.Errorf` → a test fake, `time.Sleep` → `*SleepAction`) |
| + #1913 | 254,821 | **+32,007 Method nodes** the base had fused into one per name (we found base edges whose recorded call lines lie past end-of-file — the fusion bug in the flesh). On the lost side 18/25 sampled drops were false and 3 were actively corrected. |
| whole stack | 235,519 | #1936's last commit halves `field_type_hint`, lifting gained-edge precision from 48% to 64% |

**Decision: the direction is accepted.** Please rebase **#1907 first, then #1913** onto main; #1915 and #1936 follow once those two are in. We'll take each through the normal gates as it goes green.

Two things from the measurement, so you have the full picture:

1. **We could not reproduce the 377.** Our same_module true-loss for #1913 vs base is 8,584 (net −7,207 after 1,377 gained), and no counting definition we tried — same-file only, same-package-dir only, bare callees, distinct targets — lands anywhere near 377. Could you say which corpus and which exact query produced it? It would tell us whether one of the two instruments is off.
2. **Five defect shapes surfaced in the #1913 sample** that are not visible at the 377 level. Two are precision regressions (a cross-package `StatusStrategy.ValidateUpdate` suffix collision; `metav1.Object.SetNamespace` rebinding to `meta.MetadataAccessor.SetNamespace` with the wrong arity — seen twice) and three are recall losses (type-alias receivers such as `type FeatureSet = bitmap`; an interface-typed var holding a local fake; build-tag duplicate methods defeating `unique_name`). **None of these block your PRs** — we are taking them ourselves as follow-ups on top of #1913, with the census as the test fixture source, and will credit this work in them.

The full dumps stay on our side if you want any of the sampled edges with file:line.

ilyabrykau-orca:
Here is the exact instrument behind the 377, plus two definitional differences I think account for most of the gap between our numbers.

## Corpus

A private production Go+C repository (~1,150 Go files; the same corpus as every census I have posted on this stack). Baseline Go CALLS: **21,018** — about 15x smaller than your kubernetes run (318,635), so absolute counts were never going to line up; fractions might.

## Exact query

Both builds are indexed into fresh isolated homes, then each DB is read with:

```sql
SELECT s.name || '|' || s.file_path || '|' || t.name || '|' || t.file_path,
       json_extract(e.properties, '$.strategy')
FROM edges e
JOIN nodes s ON s.id = e.source_id
JOIN nodes t ON t.id = e.target_id
WHERE e.type = 'CALLS'
  AND s.file_path LIKE '%.go'
  AND t.file_path LIKE '%.go';
```

The rows are folded into a dict keyed on that 4-tuple string (first-seen strategy wins for a duplicate key). For every key whose strategy in build A is `same_module`:

- **stayed** — key present in build B with strategy `same_module` (108)
- **moved** — key present in build B under any other strategy (444)
- **LOST** — key absent from build B under *every* strategy (377)

929 = 108 + 444 + 377.

## Three definitional differences that matter

1. **The key is (name, file) on both endpoints, deliberately not the QN** — #1913 changes the QN shape of every Go method (`pkg.method` → `pkg.Type.method`), so any diff keyed on qualified names counts every method-caller edge as lost-plus-gained even when the edge is intact. If your 8,584 is keyed on QNs (or on node ids, which embed them), that alone would dominate the number.
2. **Distinct keys, not edge instances.** The `setdefault` fold means ten call sites of the same (caller name, caller file) → (callee name, callee file) pair count once. If your count is edge rows, a corpus like kubernetes multiplies each lost pair by its call-site frequency. Worth noting: my lost *fraction* is 377/929 = **40.6%** of same_module keys; if your same_module baseline count is around 21k edge rows, 8,584 lost is the same ~41% — which would mean the two instruments agree and only the units differ. What was your same_module baseline count?
3. **Baseline build.** My 929/377 was measured between build(#1907) and build(#1907+#1913) — the stack order — not between merge-base and #1913. #1907 does not touch `same_module`, so this should be minor, but it is not zero: edges that #1907's guard removes cannot appear in either side of my same_module diff.

If it is useful, I can re-run my matrix keyed your way (QN or edge-instance) on my corpus, or you can re-key yours on (name, file) pairs — either should collapse the discrepancy to noise if the hypothesis above is right. The classifier verdicts in your 150-edge sample (18/25 lost-sampled false) match what I saw qualitatively on the 377: overwhelmingly untyped receivers whose flat-QN match was a coincidence of spelling.

ilyabrykau-orca:
Rebased onto current main behind #1907 (stack order per your decision). Conflicts were test-anchor drift only: the extraction fixture now sits after main's new Python local-binding tests, and the parallel field-chain assertion keeps the receiver-qualified fragments (`OrderHandler.PlaceOrder` → `.service.OrderService.PlaceOrder`). Full `scripts/test.sh` green locally (ASan+UBSan), `git clang-format --diff` clean. The 377-instrument answer is two comments up.

DeusData:
Heads-up: I cancelled this PR's CI run on purpose a minute ago. It was sharing the hosted-runner pool with tonight's v0.11.0 release build, and the release needed the slots. Nothing about the PR is wrong. I will re-run the full CI as soon as the release is out, so any red you see right now is the cancellation, not your change. Thanks for your patience, and for the steady stream of Go fixes.

Issue #495, opened 2026-06-17
cfg-gated twin functions collapse into one node; get_code_snippet returns the inactive branch's body


metalmon:
## Bug Description

When two items share a qualified name and are distinguished only by `#[cfg(...)]` attributes (the common Rust `#[cfg(feature = "x")]` / `#[cfg(not(feature = "x"))]` pair), the indexer stores a **single** graph node. `get_code_snippet` then returns the body of **one arbitrary branch** — in practice the *inactive*/fallback one — silently hiding the real implementation. An agent asking "what does `f` do?" gets a misleading answer (e.g. "it's a no-op returning `None`") when the feature-enabled build has a full implementation.

## Steps to Reproduce

Index any Rust crate containing a cfg-gated twin, e.g.:

```rust
#[cfg(feature = "rag-pdf")]
fn try_extract_pdf_text(bytes: &[u8]) -> Option<String> {
    if bytes.len() < 5 || &bytes[..5] != b"%PDF-" { return None; }
    /* real extraction */
}

#[cfg(not(feature = "rag-pdf"))]
fn try_extract_pdf_text(_bytes: &[u8]) -> Option<String> { None }
```

```
query_graph: MATCH (n) WHERE n.name = 'try_extract_pdf_text'
             RETURN n.qualified_name, n.start_line, n.end_line, n.signature
```

## Observed

- `total: 1` — only one node exists for the two definitions.
- It is the `#[cfg(not(...))]` stub: `start_line 385`, `end_line 387`, `signature (_bytes: &[u8])`.
- `get_code_snippet` returns `fn try_extract_pdf_text(_bytes: &[u8]) -> Option<String> { None }` — the real `#[cfg(feature = "rag-pdf")]` impl (lines 372–383) is absent from the graph entirely.

## Expected

Either (a) keep both definitions as distinct nodes tagged by their `cfg` predicate, or (b) at minimum surface that the symbol has multiple cfg-gated definitions so `get_code_snippet` can't silently return a non-representative branch. Returning the inactive fallback as if it were the implementation is the worst outcome.

## Impact

- `get_code_snippet` — a core tool — can return source that is misleading or the opposite of the real behavior.
- Complexity/metrics (`complexity`, `cognitive`, `bt`) are computed from whichever branch wins, so they misrepresent the active build.
- Common in real Rust: `#[cfg(feature)]`, `#[cfg(test)]`, `#[cfg(unix)]`/`#[cfg(windows)]` twins.

[...]

PR #1915, opened 2026-08-29
fix(extract): give each Go init() its own QN

### LINKED ISSUE #1910: Go: multiple init() per package collapse onto one QN — bodies and registration edges silently dropped
**Status: fix open — PR #1915** (file-stem+line `#`-suffix QNs for Go `init`, #495 pattern; stacked on #1913). Post-fix: `init` nodes 7 → **29**, matching source exactly (source count: 29).

### Version

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

Go allows any number of `func init()` per package (several per file included); all run at program start. Every one gets the same QN (`proj.pkg.init`), so the upsert keeps **one node per package** and silently drops the rest with their bodies.

Measured: an event-source package registering each decoder in its own file's `init()` — 21 `init` functions collapsed into 1 node; repo-wide 29 in source, 7 in the graph. The registration topology (which decoder handles which event) is invisible; the surviving node's edges are a blend of merged bodies.

### Reproduction

```go
// a.go
package reg

var handlers = map[string]func(){}

func init() { handlers["a"] = handleA }

func handleA() {}
```

```go
// b.go
package reg

func init() { handlers["b"] = handleB }

func handleB() {}
```

Actual: one node `reg.init` (a.go); b.go's body and its `handleB` edge gone or misattributed. Expected: two nodes.

### Fix (existing #495 pattern)

Rust cfg twins already fold a `#`-suffix into the QN (`rust_cfg_qualified_name`); GoogleTest macros derive unique names (#1266). Same move: `proj.pkg.init#a.go:L5`, mirrored on the call-scope side. Nothing joins on the plain QN — calling `init` explicitly is illegal in Go — so no resolution path is affected.

Related: #1909, #1911 (same silent-upsert-collapse family); tracked in #1932.

### Confirmations

ilyabrykau-orca:
## What does this PR do?

Fixes #1910. **Stacked on #1913** (which stacks on #1907) — same extraction area and shared test scaffolding; review only the last commit.

Go allows any number of `init()` functions per package — even several in one file — and all of them run at start-up. On the flat QN (`proj.pkg.init`) they all collided and the graph upsert kept **one** node per package, silently dropping the rest (measured in #1910: 21 decoder-registration `init`s in one package → 1 node; 29 in the repo → 7 nodes).

The change, following the #495 Rust cfg-twin pattern (and the #1266 GoogleTest name-derivation precedent):

1. `internal/cbm/extract_defs.c` — suffix the QN with the file basename and line: `proj.pkg.init#a.go:L5`. Calling `init` explicitly is illegal in Go, so nothing ever joins on the plain QN — no resolution path is affected.
2. `internal/cbm/extract_unified.c` (`compute_func_qn`) — mirror the exact formula, so init-body calls keep their source attribution instead of degrading to the `calls_find_source` File-node fallback.

Tests (reproduce-first):
- `pipeline_go_multi_init_nodes_survive` — two files, one package, one `init()` each: RED without the def-side change (`ic == 1, expected 2`), GREEN with it.
- `extract_go_multiple_init_disambiguated` — two `init`s in ONE file get distinct suffixed QNs (the line component disambiguates within a file).

Full `scripts/test.sh` leg green (ASan+UBSan, "All tests passed"); `git clang-format --diff` clean on changed lines.

## Checklist

- [x] Lint passes (`git clang-format --diff` clean on changed lines; clang-tidy/cppcheck via CI)

### DIFF
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -3619,6 +3619,24 @@ CBMInvocationDescriptor handle_calls(CBMExtractCtx *ctx, TSNode node, const CBML
                     }
                 }
             }
+            // Go receiver-aware guard (same direction as the TS/JS flag above).
+            // Flag a selector call x.foo(). The Go AST cannot separate a method
+            // call on a value from a package-qualified call — but every selector
+            // call the Go LSP or the import/qualified registry strategies CAN
+            // place never reaches the weak short-name guards, so the flag only
+            // bites on unresolvable receivers (`f.Close()` on an os.File,
+            // `sha256.New()` behind an unindexed import), where a project-wide
+            // short-name match fabricates an edge to an unrelated project
+            // symbol sharing the name. Bare calls (helper()) keep
+            // is_method=false and resolve same-module/import paths as before.
+            if (ctx->language == CBM_LANG_GO &&
+                strcmp(ts_node_type(node), "call_expression") == 0) {
+                TSNode gofn = ts_node_child_by_field_name(node, TS_FIELD("function"));
+                if (!ts_node_is_null(gofn) &&
+                    strcmp(ts_node_type(gofn), "selector_expression") == 0) {
+                    call.is_method = true;
+                }
+            }
 
             TSNode args = ts_node_child_by_field_name(node, TS_FIELD("arguments"));
             // ObjectScript stores args under oref_method/method_args, not the
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -3739,6 +3739,25 @@ static void extract_func_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec 
         }
     }
 
+    /* Go allows any number of init() functions per package — even several in
+     * one file — and they all run at start-up. On the flat QN they all
+     * collided and the graph upsert kept ONE node per package, silently
+     * dropping the rest (#1910). Disambiguate with the #495 cfg-twin pattern:
+     * fold the file basename and the init's per-file ordinal into the QN.
+     * The ordinal (not the line) keeps the QN stable when code is edited
+     * above the function — the QN is node identity, and a line suffix would
+     * churn nodes on every unrelated insertion. Nothing ever joins on the
+     * plain QN — calling init explicitly is illegal in Go — and the
+     * call-scope side (compute_func_qn in extract_unified.c) computes the
+     * same cbm_go_init_ordinal so init-body calls keep their attribution. */
+    if (ctx->language == CBM_LANG_GO && strcmp(def.label, "Function") == 0 &&
+        strcmp(name, "init") == 0) {
+        const char *base = strrchr(ctx->rel_path, '/');
+        base = base ? base + 1 : ctx->rel_path;
+        def.qualified_name = cbm_arena_sprintf(a, "%s#%s:%d", def.qualified_name, base,
+                                               cbm_go_init_ordinal(node, ctx->source));
+    }
+
     // C++/CUDA: out-of-line method definition (`Foo::bar` in a .cc/.cpp). The
     // class body in the header is declaration-only, so without this the
     // definition is recorded as a free Function. Promote it to a Method whose QN
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -3508,7 +3508,7 @@ static void set_def_complexity(CBMDefinition *def, TSNode body, const CBMLangSpe
  * Walks to the parameter_declaration's `type` field, unwrapping pointer_type
  * and generic_type, and returns the type_identifier text (e.g. "OrderService").
  * Returns NULL if no type_identifier is found. */
-static char *go_receiver_type_name(CBMArena *a, TSNode recv, const char *source) {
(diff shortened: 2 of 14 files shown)


ilyabrykau-orca:
Rebased onto current main (stacked on the rebased #1913). No semantic changes; full test leg green. Post-fix init-node recovery (29/29) is in the #1906 census comment.

DeusData:
Reviewed the last commit. The diagnosis is right, the safety argument holds, and I have one substantive concern about the suffix itself.

**The diagnosis is confirmed by the schema.** `src/store/store.c:257` declares `UNIQUE(project, qualified_name)` and the node upsert at `:2364` is `ON CONFLICT(project, qualified_name) DO UPDATE SET`. So a flat `proj.pkg.init` really does mean every `init` in a package is the same row, and the last writer wins — which is exactly your 21 → 1.

**Your safety claim holds, and I checked it both ways.** Nothing in `internal/cbm/`, the LSP resolvers, or `src/pipeline/` matches on the literal `"init"`, so no resolver joins on the plain QN. And since calling `init` explicitly is illegal in Go, no call site produces `init` as a callee either. Both directions are clear.

**The precedents you cite are real.** `extract_defs.c:2079` is the Rust cfg twin using `"%s#%s"`, and `:3544`–`:3613` is the GoogleTest derivation, whose comment names the same failure — "causing qualified-name collisions and node loss (#1266)". The shape is established house practice for this problem class.

## The concern: the suffix is not stable, and both precedents are

This is the part I would like you to weigh, because it is the one property those two precedents share that yours does not.

The Rust cfg twin derives its suffix from the `cfg(...)` predicate text, and says why in the code:

> Build a compact predicate suffix from the cfg(...) text, dropping whitespace and quotes so the QN stays readable and **stable**.

GoogleTest derives its from the macro arguments — `TEST(Suite, Case)` → `TEST_Suite_Case`. Both are semantic content that does not move when the file is edited.

`proj.pkg.init#a.go:L5` embeds a **line number**. Add an import above that `init` and its qualified name changes — and since the QN *is* the node's identity under that UNIQUE constraint, an incremental reindex sees the old node vanish and a new one appear, for an edit that did not touch the function. That is node churn on every unrelated edit above any `init`, and it lands on the delta-merge path rather than a full rebuild.

The file basename alone will not do it — your own `extract_go_multiple_init_disambiguated` test requires two `init`s in one file to differ, and only the line separates them today. But a **per-file ordinal** would: `proj.pkg.init#a.go:1`, `#a.go:2`. That still satisfies that test, and it only churns if you reorder the `init`s in a file rather than every time you add a line above one.

I am raising this rather than ruling on it — you may have considered ordinals and rejected them for a reason I cannot see from here. If so, say why and I will take it.

## Also

**The whole stack is currently `DIRTY`** — #1907, #1913 and this one all have conflicts. `main` was broken for a few hours today (duplicate struct members merged into `type_registry.h`) and was repaired an hour ago by #1993, so now is a good moment to rebase all three.

Thanks for measuring the loss rather than describing it — "21 decoder-registration `init`s in one package → 1 node; 29 in the repo → 7 nodes" is what makes this obviously worth fixing.

ilyabrykau-orca:
Rebased onto current main as part of the 1907→1913→1915→1936 stack; only test insertion-anchor drift in tests/test_pipeline.c, re-anchored. Full scripts/test.sh green on the stack head.

ilyabrykau-orca:
You are right, and the ordinal is strictly better — implemented, not argued with.

I had not weighed the incremental path when I chose the line number: on a full rebuild the line suffix is merely ugly, but under delta-merge the QN is node identity, so an insertion above any `init` would delete-and-recreate the node on an edit that never touched it. Both precedents being edit-stable is not a coincidence; it is the property that makes a QN suffix safe.

Now: `pkg.init#file.go:N`, where N is the 1-based source-order ordinal among the file's top-level `init` function_declarations. One formula, computed from the tree by a shared `cbm_go_init_ordinal` (helpers.c) on both the definition side and the call-scope side (compute_func_qn), so the two cannot drift. Reordering the `init`s within a file still churns — that is inherent to any content-free discriminator — but nothing else does.

The test now pins the property you asked for, not just distinctness: exact QNs `t.init#x.go:1` / `t.init#x.go:2`, then the same file re-extracted with an import and a declaration inserted **above both** inits, asserting both QNs are byte-identical to before.

Branch force-pushed with the stack rebase; full scripts/test.sh green on the stack head.

DeusData:
Status pointer, so this thread is not left hanging: the stack is now waiting on a direction call about the Go qualified-name shape surfaced by the #1913 census — details on #1913. Please hold the rebase of this branch along with the rest of the stack until that lands; I will update here.

DeusData:
The update I promised on 3 September, late and with apologies for that: the direction call on #1913 landed on 15 September (the receiver-qualified shape stays, stack order 1907 → 1913 as you already rebased it), so nothing is waiting on a decision any more. #1907 and #1913 are next in line for a branch update and a fresh CI run as soon as the hosted-runner queue drains from tonight's merges; once #1913 is in I will ping here, and a rebase of this branch onto it is then the only step left before review. Thank you for holding the stack steady through all of this, and for the ordinal fix — that reasoning about delta-merge identity was exactly right.

PR #1918, opened 2026-08-29
fix(cypher): RETURN * must read the live scope, not the query pattern

### LINKED ISSUE #1983: Cypher RETURN * reads the parse-time pattern, not the live scope — duplicate columns, and empty values after WITH
### Version

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

`RETURN *` builds its columns from the variables the **query pattern** named, not from the
bindings it is about to project. One line causes two separate wrong answers, and neither
reports an error — so a caller cannot tell a broken query from an empty graph.

Measured against a real indexed project:

| Query | Columns | Values |
|:--|:--|:--|
| `MATCH (f:Function) RETURN *` | 4, correct | real |
| `MATCH (f:Function) OPTIONAL MATCH (f)-[:CALLS]->(g) RETURN *` | 12 — `f` appears twice | real |
| the same, plus `WITH f.name AS caller, g.name AS callee` | the same 12 | **all empty** |

`execute_return_star` calls `collect_pattern_vars(q, ...)`, which reads the parse-time
pattern. Two consequences follow from that one call:

1. After a `WITH`, the pattern variables are out of scope. The `WITH` replaced them with
   the names it made, and the live bindings are keyed by those names. Asking for `f` finds
   nothing, so every column is written as `""`.
2. `collect_pattern_vars` appends each pattern in turn with no repeat check, so a variable
   named in two patterns gets its columns emitted twice.

I expected `RETURN *` to project the live scope — the bindings that actually exist at that
point in the query.

### Reproduction

1. Index any project with functions and `CALLS` edges.
2. Run each of the three queries in the table above through `query_graph`.
3. Result: query 2 returns 12 columns with `f` duplicated; query 3 returns 12 columns all
   of which are empty strings, with no error. Expected: query 2 returns 8 distinct
   columns, and query 3 returns the two columns the `WITH` named, with real values.

### Logs

```text
No error is emitted in either failing case — empty columns are returned as if the graph
had no matching data.
```

```text
```

Reproduces on any indexed project with at least one CALLS edge.

### Confirmations

CaptainMittens:
PROBLEM

`RETURN *` built its columns from the variables the query pattern named,
never from the bindings it was about to project. One line caused two
separate wrong answers, and neither reported an error — so a caller
cannot tell a broken query from an empty graph.

Measured against a real indexed project before the fix:

| Query | Columns | Values |
|---|---|---|
| `MATCH (f:Function) RETURN *` | 4, correct | real |
| `MATCH (f:Function) OPTIONAL MATCH (f)-[:CALLS]->(g) RETURN *` | 12 — `f` appears twice | real |
| the same, plus `WITH f.name AS caller, g.name AS callee` | the same 12 | ALL empty |

CAUSE

`execute_return_star` called `collect_pattern_vars(q, ...)`, which reads
the parse-time pattern. Two consequences fall out of that one call.

After a `WITH`, the pattern variables are out of scope — the `WITH`
replaced them with the names it made, and the live bindings are keyed by
those names. Asking for `f` then finds nothing, and every column is
written as `""`.

Separately, `collect_pattern_vars` appended each pattern in turn with no
repeat check, so a variable named in two patterns got its four columns
twice. The `OPTIONAL MATCH` above names `f` twice.

FIX

- New `execute_return_star_after_with`: when a `WITH` is present, columns
  come from its items. A name a `WITH` makes holds one value rather than
  a node, so it is ONE column, not the four a node variable gets.
- `collect_pattern_vars` gained a repeat check. It has exactly one
  caller, so nothing else is affected.

The query above now answers two columns, `caller` and `callee`, holding
their values.

TESTS

Two, both failing against the old code:

```
cypher_return_star_dedups_repeated_pattern_var   col_count == 12, expected 8
cypher_return_star_after_with_names_aliases      col_count == 8,  expected 2
```

Before: 183 passed, 2 failed. After: 185 passed.

GATES RUN

- `scripts/lint.sh --ci` — exit 0, all linters passed
- `scripts/test.sh --suites cypher` — exit 0, 185 passed

NOT IN THIS PR

A second fault found alongside this one, kept out to keep the PR to one
change: a variable a `WITH` drops is still accepted afterwards and
renders empty, because nothing checks a projected name against the live
scope. `MATCH (f:Function)-[:CALLS]->(g) WITH f.name AS caller RETURN
caller, g.name` runs and prints a `g.name` column of nothing, where real
Cypher refuses the query and names `g`. Happy to raise it as its own
issue.

Fixes #1983

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -4193,17 +4193,34 @@ static void execute_with_clause(cbm_query_t *q, binding_t **bindings_ptr, int *b
 
 /* Project RETURN * — all bound variable properties */
 /* Collect all variable names from query patterns */
+/* Has this variable already been collected? A query may name the same variable
+ * in more than one pattern, and RETURN * must give it one set of columns. */
+static bool star_var_seen(const char **vars, int vc, const char *name) {
+    for (int i = 0; i < vc; i++) {
+        if (strcmp(vars[i], name) == 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
+/* Collect the variables a RETURN * projects, in the order the query names them
+ * and with no repeats. Without the repeat check, `MATCH (f) OPTIONAL MATCH
+ * (f)-[:CALLS]->(g)` names f in two patterns and f gets its four columns
+ * twice. */
 static int collect_pattern_vars(cbm_query_t *q, const char **vars, int max_vars) {
     int vc = 0;
     for (int pi = 0; pi < q->pattern_count; pi++) {
         for (int ni = 0; ni < q->patterns[pi].node_count && vc < max_vars; ni++) {
-            if (q->patterns[pi].nodes[ni].variable) {
-                vars[vc++] = q->patterns[pi].nodes[ni].variable;
+            const char *var = q->patterns[pi].nodes[ni].variable;
+            if (var && !star_var_seen(vars, vc, var)) {
+                vars[vc++] = var;
             }
         }
         for (int ri = 0; ri < q->patterns[pi].rel_count && vc < max_vars; ri++) {
-            if (q->patterns[pi].rels[ri].variable) {
-                vars[vc++] = q->patterns[pi].rels[ri].variable;
+            const char *var = q->patterns[pi].rels[ri].variable;
+            if (var && !star_var_seen(vars, vc, var)) {
+                vars[vc++] = var;
             }
         }
     }
@@ -4255,8 +4272,41 @@ static void project_star_row(binding_t *b, const char **vars, int vc, const char
     }
 }
 
+/* RETURN * after a WITH.
+ *
+ * The pattern's variables are out of scope by this point — the WITH replaced
+ * them with the names it made. Each of those names holds one value, not a
+ * node, so each is ONE column rather than the four a node variable gets.
+ *
+ * Reading the pattern here instead is the fault this function exists to avoid:
+ * it named variables the bindings no longer hold, found nothing for every one
+ * of them, and answered a full result of empty strings with no error. */
+static void execute_return_star_after_with(cbm_query_t *q, binding_t *bindings, int bind_count,
+                                           int max_rows, result_builder_t *rb) {
+    cbm_return_clause_t *wc = q->with_clause;
+    char name_bufs[CYP_MAX_VARS][CBM_SZ_128];
+    const char *cols[CYP_MAX_VARS];
+    int col_n = wc->count < CYP_MAX_VARS ? wc->count : CYP_MAX_VARS;
+    for (int i = 0; i < col_n; i++) {
+        cols[i] = resolve_item_alias(&wc->items[i], name_bufs[i], sizeof(name_bufs[i]));
+    }
+    rb_set_columns(rb, cols, col_n);
+    for (int bi = 0; bi < bind_count && rb->row_count < max_rows; bi++) {
+        const char *vals[CYP_MAX_VARS];
+        for (int i = 0; i < col_n; i++) {
+            cbm_node_t *vn = binding_get(&bindings[bi], cols[i]);
+            vals[i] = vn && vn->name ? vn->name : "";
+        }
+        rb_add_row(rb, vals);
+    }
+}
+
 static void execute_return_star(cbm_query_t *q, binding_t *bindings, int bind_count, int max_rows,
                                 result_builder_t *rb) {
+    if (q->with_clause) {
+        execute_return_star_after_with(q, bindings, bind_count, max_rows, rb);
+        return;
+    }
     const char *vars[CBM_SZ_32];
     int vc = collect_pattern_vars(q, vars, CBM_SZ_32);
     build_star_columns(rb, vars, vc);
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -1802,8 +1802,16 @@ static int parse_return_or_with(parser_t *p, cbm_return_clause_t **out, bool is_
     /* Projection is materialized per row into fixed-width stack arrays sized at
      * CBM_SZ_32 columns (execute_return_simple and its siblings). Bound the
      * parsed item count to that width so an over-wide RETURN is rejected here
-     * instead of writing past those arrays downstream. */
-    if (r->count > CBM_SZ_32) {
+     * instead of writing past those arrays downstream.
+     *
+     * WITH is bounded tighter, by CYP_MAX_VARS. Every item a WITH projects
+     * becomes one variable of the binding that carries the rest of the query,
+     * and binding_t holds exactly CYP_MAX_VARS variables. A wider WITH used to
+     * parse, then lose every alias past the 16th in with_add_vbinding_var and
+     * answer with silently blank columns. Refuse it here, the same way an
+     * over-wide RETURN is refused, so the caller sees an error instead of a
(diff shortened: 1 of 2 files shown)


DeusData:
Reviewed. Both fixes are correct, and the diagnosis is the good part — "a caller cannot tell a broken query from an empty graph" is exactly why this mattered more than a wrong column count.

**The scope claim checks out.** `collect_pattern_vars` really does have exactly one caller (`execute_return_star`, line 4224) and none anywhere else in the tree, so adding the repeat check cannot reach anything you did not intend. I verified that rather than taking it, because "only one caller" is the kind of claim that quietly stops being true.

**The `WITH` split is the right shape.** Reading `wc->items` instead of the pattern is the fix, and the one-column-per-`WITH`-name rule against four-for-a-node is the detail that makes the column count right rather than merely different. Keeping first-occurrence order in the dedup matters too — `RETURN *` column order is part of the answer.

## One finding, same fault class as the one you fixed

`execute_return_star_after_with` caps columns at `CYP_MAX_VARS`:

```c
int col_n = wc->count < CYP_MAX_VARS ? wc->count : CYP_MAX_VARS;
```

`CYP_MAX_VARS` is 16, but `wc->count` is not bounded by it. The return-item parser grows its array — `if (r->count >= cap) { cap *= PAIR_LEN; r->items = safe_realloc(...); }` — so a `WITH` naming 17 or more items parses fine, and this then answers 16 columns and drops the rest **with no error**.

That is the same shape as the bug this PR exists to remove: a query that silently returns a wrong-but-plausible result. It is reachable by a legitimate query, and a caller cannot tell.

Not asking for a redesign — either clamping loudly or sizing from `wc->count` would do. Worth handling here rather than as a follow-up, precisely because the PR's argument is that silent wrong answers are the problem.

For context, so this does not read as a standard you are being held to alone: the pattern path has a cap of its own (`collect_pattern_vars(q, vars, CBM_SZ_32)`), so the shape is pre-existing. Yours is just tighter and newer.

## The other fault you found

Yes please, raise it as its own issue — a variable a `WITH` drops still being accepted and rendering empty is worth tracking, and `MATCH (f:Function)-[:CALLS]->(g) WITH f.name AS caller RETURN caller, g.name` is a clean reproduction to put in it. Keeping it out of this PR was the right call.

## Status

Cleared, and approved once the column cap is dealt with.

**One heads-up:** `main` does not currently compile — two changes landed the same type short-name index into `type_registry.h` and git merged them into duplicate struct members. #1993 is the repair. Your green here predates it and is stale rather than wrong; wait for #1993 before rebasing.

CaptainMittens:
Both points addressed. Pushed `738174f7`.

### The column cap

Fixed, but not the way it looks like it should be fixed, so it is worth saying why.

The obvious repair is to widen the arrays in `execute_return_star_after_with` from
`CYP_MAX_VARS` to `CBM_SZ_32`, since the parser already allows 32 items. That would make
the result worse. A binding holds exactly `CYP_MAX_VARS` variables
(`cypher.c:2209-2210`), and `with_add_vbinding_var` (`cypher.c:3947`) drops any alias past
the 16th on the way in. Widening the projection to 32 would answer a 20-alias `WITH` with
20 columns of which 4 are permanently blank — silent wrong data in place of a silent short
answer. The 16 was not an undersize; it matched what the binding can actually carry.

So the bound moved to parse time instead, beside the one that already refuses an over-wide
`RETURN`:

```c
if (r->count > (is_with ? CYP_MAX_VARS : CBM_SZ_32)) {
```

The clamp you flagged stays, with a comment naming where the bound comes from. It can no
longer fire, and it is what keeps that function correct if the bound ever moves.

`cypher_wide_with_refused_not_truncated` asserts both halves — a 20-alias `WITH` is
refused, and a 16-alias `WITH` still succeeds with 16 columns — so the guard rejects only
what the binding genuinely cannot hold. Verified failing before the change and passing
after; the cypher suite is 186 passed. No product code emits a Cypher `WITH` at all, and
the widest in the test suite is 3 items.

### The issue you asked for

That one is already fixed on `main`, by #1922, which merged on 29 August. Your
reproduction now answers correctly:

```
MATCH (f:Function)-[:CALLS]->(g) WITH f.name AS caller RETURN caller, g.name
  → variable 'g' is not in scope for RETURN — the WITH clause did not carry it through
```

This branch predates that commit, which is why the fault still looks live from here.

What I did file is **#1995**, for a gap in that same check. `check_projection_scope` models
declared names in a fixed 32-entry array and returns `NULL` — no check — when a query
declares more than that. Nothing bounds how many variables a pattern can declare, so the
scope check silently switches itself off on a wide query and the old blank-column silence
comes back:

```
10 declared names, RETURN zzz.name  → refused, 'zzz' is not in scope
35 declared names, RETURN zzz.name  → accepted, answers a zzz.name column
```

Same root cause as the pre-existing `execute_default_projection` truncation you mentioned:
`pat->node_count` grows on demand with no cap. One parse-time bound on the pattern's
variable count would close all of it. I left that out of this PR deliberately — it is a
different change from the one you reviewed.

DeusData:
The column-cap finding is addressed — `738174f7` ("refuse a WITH wider than a binding can carry") is exactly what I was asking for. Thank you.

**Neither of your two red checks is yours:**

- `test-unix (macos-14)` — the suite passed outright: **7618 passed, 0 failed, 7 skipped**. The job died afterwards at the *"Shard manifest (completeness proof)"* step, which is harness bookkeeping about proving every suite ran, not a test result.
- `test-windows-guards` — `test_daemon_stability.py` regression plus two `SETUP FAIL` lines. That is a known standing red on our side; it has reddened several unrelated PRs this week.
- `ci-ok` just aggregates those.

## One thing you will want to know: this conflicts with your own #1998

I checked rather than guessed — merging `main` + #1918 + #1998 produces:

```
CONFLICT (content): Merge conflict in tests/test_cypher.c
```

`src/cypher/cypher.c` auto-merges; the test file does not. And the two are the same defect family: #1998's `check_pattern_var_capacity` refuses a query naming more pattern variables than a binding holds, while this one refuses a `WITH` wider than a binding can carry. Together with #1875, which landed earlier today, that is three PRs on the theme of *refusing a query the binding cannot honestly answer*.

They are each correct alone, which is precisely why this is worth flagging — nothing surfaces the collision until one of them lands and the other needs rework.

**Suggested order: this one first.** #1918 needs nothing further from you, while #1998 still owes the one-line cppcheck fix I asked for (`const char *node_vars[CYP_MAX_VARS] = {0};` and the same for `edge_vars`). Once this merges, #1998 will need a rebase of its test file — a small, mechanical one, since the conflict is additive test cases rather than competing logic.

If you would rather land #1998 first, that is fine too; just say so and I will sequence it that way. I only want the rebase to be expected rather than a surprise.

I am not merging this yet because the two attributable reds are still on the board and I would rather re-run than merge past them. Our Actions queue is finally draining, so a rebase would produce a clean run reasonably quickly if you want to force one.

DeusData:
Merged — thank you! 🙏

`RETURN *` reading the live scope (rather than the parse-time pattern) fixes two real observable bugs at once — the duplicate columns, and the empty values after a `WITH` — and the second commit refusing a `WITH` wider than the binding capacity closes the hole before it can become a memory bug. The test coverage in `test_cypher.c` for both shapes is very welcome.

Landed on `main` as merge `8f64c36d8575`; it will be in the next release.

PR #1921, opened 2026-08-29
fix(tests): copy the runner, not /bin/cat, as the POSIX blocked image

OhOkThisIsFine:
## Problem

`daemon_runtime_process_fingerprint_never_hashes_replacement_path` (POSIX variant) copies `/bin/cat` to a temp file named `image` and executes the copy. On Ubuntu 25.10+/26.04 — and any system with uutils/Rust coreutils — `/bin/cat` is a multi-call binary. A copy invoked under the name `image` prints `coreutils: unknown program 'image'` and exits 1, so the test fails at `ASSERT(setup)`.

Reproduced on WSL Ubuntu 26.04 (uutils coreutils 0.8.0):

```
$ cp /bin/cat "$d/image" && echo hi | "$d/image"
coreutils: unknown program 'image'
exit=1
```

## Fix

Mirror the Windows variant of the same test, which already copies the test runner and parks it in a `__cbm_runtime_image_holder` argv mode:

- `tests/test_main.c` — the `__cbm_runtime_image_holder` mode gains a POSIX branch: block reading stdin until EOF, exactly the behaviour the `cat` donor provided.
- `tests/test_daemon_runtime.c` — `runtime_test_spawn_blocked_executable` passes the holder argument; its target must now be a copy of the runner (it has exactly one caller). The POSIX fixture copies the runner via `runtime_test_copy_self_image` instead of `/bin/cat`.
- The `/bin/echo` replacement donor is gone too. It was only hashed, never executed — and on a uutils system `cat` and `echo` can be the same multi-call bytes, which would break the fixture's `original != replacement` precondition. The replacement is now a small data file, mirroring the Windows variant.

No macOS signing change is needed: the image copy stays byte-identical to the runner, and the identical-copy fixture already proves such a copy executes.

## Verification

`scripts/test.sh --suites daemon_runtime BUILD_DIR=build/wsl` on WSL Ubuntu 26.04 (uutils coreutils 0.8.0): **44 passed / 0 failed**, including `daemon_runtime_process_fingerprint_never_hashes_replacement_path`.

### DIFF
--- a/tests/test_daemon_runtime.c
+++ b/tests/test_daemon_runtime.c
@@ -382,6 +382,9 @@ static bool runtime_test_copy_executable(const char *source, const char *destina
     return ok;
 }
 
+/* PATH must be a copy of this runner: the child runs the
+ * __cbm_runtime_image_holder mode, which blocks reading the release pipe
+ * wired to its stdin. */
 static pid_t runtime_test_spawn_blocked_executable(const char *path, int *release_fd_out) {
     int ready[2] = {-1, -1};
     int input[2] = {-1, -1};
@@ -427,7 +430,7 @@ static pid_t runtime_test_spawn_blocked_executable(const char *path, int *releas
     if (input[0] != STDIN_FILENO) {
         (void)posix_spawn_file_actions_addclose(&actions, input[0]);
     }
-    char *const child_argv[] = {(char *)path, NULL};
+    char *const child_argv[] = {(char *)path, "__cbm_runtime_image_holder", NULL};
     pid_t child = -1;
     if (posix_spawn(&child, path, &actions, NULL, child_argv, environ) != 0) {
         child = -1;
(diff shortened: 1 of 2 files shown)


DeusData:
Merged as `0a8a1f29`.

This is a fix for a failure most of the project has not seen yet, which is the useful kind. `/bin/cat` being a multi-call binary under uutils means the copy-and-rename trick stops working the moment a distro switches coreutils implementations — so this test was going to start failing across the board on its own schedule, with a message (`coreutils: unknown program 'image'`) that points nowhere near the actual cause. Catching it on Ubuntu 26.04 before it arrives everywhere else is worth more than the diff size suggests.

Three things done right:

**You mirrored the Windows variant instead of inventing a mechanism.** The `__cbm_runtime_image_holder` argv mode already existed and already solved exactly this problem on the other platform. Extending it with a POSIX branch that blocks on stdin — the one behaviour the `cat` donor was actually providing — leaves one pattern in the tree rather than two.

**You found the second instance of the same root cause.** Removing the `/bin/echo` replacement donor because on a uutils system `cat` and `echo` can be *the same multi-call bytes*, which would silently break the fixture's `original != replacement` precondition, is the part that would have been missed. That donor was only ever hashed, never executed, so nothing pointed at it — you got there from the mechanism rather than from a failure.

**You addressed the platform question before it was asked.** Noting that the image copy stays byte-identical to the runner, and that the identical-copy fixture already proves such a copy executes, closed the macOS-signing concern in one sentence.

Reproduced with the exact failing command and output, fixed, and verified 44/0 on the affected platform. Thank you — a clean first contribution.

PR #1922, opened 2026-08-29
fix(cypher): refuse a name that is not in scope instead of projecting a blank column

### LINKED ISSUE #1919: query_graph: a variable the WITH clause dropped is still accepted, and its column comes back empty with no error
### Version

`codebase-memory-mcp dev` — built from source at `main` `997d087b`.

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

A `RETURN` may name a variable that the `WITH` before it dropped. The query
runs, exits clean, prints a column for that variable, and fills the column with
empty strings. Nothing reports an error.

Real Cypher refuses this query and names the offending variable. Here the
caller cannot tell a wrong query from a graph that holds no such data.

The empty column is the harmful part. A reader — a person or an agent — sees a
column of nothing and reads it as "the graph has no data here". The true answer
is "your query named something that is out of scope".

### Reproduction

The repository's own cypher test fixture reproduces it, so no outside code is
needed. `setup_cypher_store()` in `tests/test_cypher.c` builds a store with
three `CALLS` edges.

1. Code being indexed: the fixture built by `setup_cypher_store()` in
   `tests/test_cypher.c`.

2. The query:

   ```
   MATCH (f:Function)-[:CALLS]->(g) WITH f.name AS caller RETURN caller, g.name
   ```

   `g` does not survive the `WITH`. A plain `MATCH` does this as readily as an
   `OPTIONAL MATCH`, so the optional half is not the cause.

3. Result: three rows. Column `g.name` is present and every value is `""`.
   Return code 0. No error text anywhere.

   Expected: a parse or execution error naming `g` as out of scope.

To see it without writing a test, run the same query through `query_graph`
against any indexed project, using a variable the `WITH` did not carry.

### Logs

```
PROBE rc=0 col_count=2 row_count=3
PROBE column[0]="caller"
PROBE column[1]="g.name"
PROBE row[0] = ["HandleOrder"] [""]
PROBE row[1] = ["HandleOrder"] [""]
PROBE row[2] = ["ValidateOrder"] [""]
```

Captured by a throwaway test against the fixture above, on `main` `997d087b`
plus the one commit in #1918. That commit only touches `RETURN *`, so it does
not affect this path.

### Where it comes from

Nothing checks a projected name against the live scope. An unbound variable
resolves to NULL and renders as `""`, and that is deliberate:

- `src/cypher/cypher.c:2568` — `return NULL; /* unbound variable */`
- `src/cypher/cypher.c:3345` and `:3350` — an unmatched `OPTIONAL MATCH` target
  stays unbound on purpose, and the projection renders it `""`.

That convention is right for the `OPTIONAL MATCH` case. Its cost is that a name
which was never carried through looks exactly like an optional row that did not
match, so no error can be raised at projection time.

A search for `not in scope`, `unknown variable`, `undefined variable` and
`scope check` across `src/cypher/cypher.c` at `997d087b` returns nothing.

### Suggested fix

Add one check after parsing, before execution: every variable a `RETURN` or
`WITH` item names must be either a pattern variable still in scope, or an alias
the previous `WITH` created. Report the name that fails.

The check belongs at parse time, which keeps the `""` convention for an
unmatched `OPTIONAL MATCH` target exactly as it is. The check is about names
that were never carried through, not about rows that did not match.

### Related

- #1192 — `UNWIND` is parsed but silently ignored and should error. Same shape:
  a query that is wrong runs quietly instead of failing.
- #1918 — fixes a neighbouring fault where `RETURN *` after a `WITH` answered
  every value empty. That one is the star projection reading the query pattern
  instead of the live bindings. This issue is the named-column path and is a
  separate fix, kept out of that PR on the one-issue-per-PR rule.

### Confirmations

- [x] My reproduction uses shareable code — this repository's own test fixture.

CaptainMittens:
Fixes #1919.

## What was wrong

A `RETURN` could name a variable the `WITH` before it dropped. The query ran,
exited clean, printed a column for that name, and filled it with empty strings.
Nothing reported an error.

```
MATCH (f:Function)-[:CALLS]->(g) WITH f.name AS caller RETURN caller, g.name
```

`g` does not survive the `WITH`. Measured through `query_graph` on a real
indexed project before the fix:

```
rows: 5  (cols: caller g.name)
  appWith          -
  send             -
  appWithNoPartner -
```

The silence is the harm. A reader — a person or an agent — sees a column of
nothing and reads it as "the graph holds no such data". The true answer is
"your query named something that is out of scope". For a code graph, a false
"nothing calls X" is the worst answer it can give.

This is the same failure shape as #373, where an unknown function used to
project a blank column and now fails loudly. Same answer here.

## The fix

One check on the parsed query, between parse and execute. Every name a `RETURN`
or `WITH` item uses must be a variable some pattern declares, or an alias the
previous `WITH` made.

**The check reads the parsed query, not the run-time bindings, and that split is
the whole point.** It asks whether the query DECLARED the name — a different
question from whether a row happened to bind it. So the existing convention at
`cypher.c:3345` and `:3350` is untouched: an `OPTIONAL MATCH` target that
matched nothing is still declared, still legal, and still projects `""`. One of
the three tests below holds that line.

Two placeholders are skipped, because neither names anything the query
declared: `count(*)` stores `"*"` (`cypher.c:1614`) and a `CASE` expression
stores `"CASE"` (`cypher.c:1643`). Both were found by the existing suite —
skipping them turned 7 red tests green.

A query with more names than the guard can model skips the check rather than
guessing. A wrong refusal costs the caller a working query, which is worse than
the silence this removes.

The error names the offending variable and the clause, because a message that
does not say WHICH name is wrong sends the reader back to guessing:

```
variable 'g' is not in scope for RETURN — the WITH clause did not carry it through
```

## Tests

Three added, red before the change:

| Test | What it holds |
|---|---|
| `cypher_rejects_projection_of_dropped_with_var_issue1919` | the fault — must error and name `g` |
| `cypher_optional_match_target_still_allowed_issue1919` | an unmatched `OPTIONAL MATCH` target must still project `""` |
| `cypher_with_alias_stays_in_scope_issue1919` | a `WITH` alias, and a variable carried through whole, still work |

Red-green, both runs quoted:

```
before:  cypher_rejects_projection_of_dropped_with_var_issue1919  FAIL  ASSERT(rc != 0)
         185 passed, 1 failed
after:   cypher_rejects_projection_of_dropped_with_var_issue1919  PASS
         186 passed
```

The two guard tests passed before the change as well as after, which is what
makes them useful — they were there to catch a fix that over-rejects, and they
did: a first attempt broke 7 existing tests before the placeholder skip.

## Gates run locally

Both through the entry points `CONTRIBUTING.md` names, on macOS arm64:

| Gate | Result |
|---|---|
| `scripts/lint.sh --ci` | exit 0 — "All linters passed" |
| `scripts/test.sh --suites cypher` | exit 0 — 186 passed, 0 failed |

The full suite was not run locally. CI covers it.

## Scope

One issue, one fix. This does not touch #1918, which fixes a neighbouring fault
in `RETURN *` and can land in either order — the two change different
functions.

Names inside a `CASE` expression and inside a multi-argument function's
arguments are still unchecked. Widening the check there is a separate change
with its own over-rejection risk.

### DIFF
--- a/src/cypher/cypher.c
+++ b/src/cypher/cypher.c
@@ -4894,6 +4894,164 @@ static int execute_single(cbm_store_t *store, cbm_query_t *q, const char *projec
 
 /* ── Main entry point ─────────────────────────────────────────── */
 
+/* ── Is every name a RETURN or WITH item uses actually in scope? ──
+ *
+ * An unbound variable resolves to NULL and renders as "" on purpose: an
+ * OPTIONAL MATCH target that found no row has to project a blank rather than
+ * drop the row. The cost of that convention is that a name the query NEVER
+ * carried through looks exactly the same on screen. A typo, or a variable a
+ * WITH dropped, then reads as "the graph holds no such data" instead of "your
+ * query named something that is not there" (#1919). Same failure shape as
+ * #373, and the same answer: say so out loud.
+ *
+ * The check runs on the parsed query, not on the run-time bindings, and that
+ * is the whole trick. It asks whether the query DECLARED the name — a
+ * different question from whether a row happened to bind it. So the OPTIONAL
+ * MATCH convention above is untouched: a target that did not match is still
+ * declared, still legal, and still projects "".
+ */
+
+/* Cap for one query's declared names. CYP_MAX_VARS is the binding cap for
+ * nodes alone; a pattern also names edges, so this is the roomier of the two.
+ * A query that overruns it skips the check rather than guessing — a wrong
+ * refusal costs the caller a working query, which is worse than the silence
+ * this guard removes. */
+enum { CYP_SCOPE_MAX_NAMES = 32 };
+
+static bool scope_holds(const char *const *names, int count, const char *want) {
+    for (int i = 0; i < count; i++) {
+        if (names[i] && strcmp(names[i], want) == 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
+/* Every name the query's patterns declare, plus an UNWIND alias.
+ * Answers -1 when there are more names than the cap holds. */
+static int collect_declared_names(const cbm_query_t *q, const char **out, int cap) {
+    int n = 0;
+    for (int pi = 0; pi < q->pattern_count; pi++) {
+        const cbm_pattern_t *pat = &q->patterns[pi];
+        for (int ni = 0; ni < pat->node_count; ni++) {
+            const char *var = pat->nodes[ni].variable;
+            if (var && !scope_holds(out, n, var)) {
+                if (n >= cap) {
+                    return -1;
+                }
+                out[n++] = var;
+            }
+        }
+        for (int ri = 0; ri < pat->rel_count; ri++) {
+            const char *var = pat->rels[ri].variable;
+            if (var && !scope_holds(out, n, var)) {
+                if (n >= cap) {
+                    return -1;
+                }
+                out[n++] = var;
+            }
+        }
+    }
+    if (q->unwind_alias && !scope_holds(out, n, q->unwind_alias)) {
+        if (n >= cap) {
+            return -1;
+        }
+        out[n++] = q->unwind_alias;
+    }
+    return n;
+}
+
+/* What a WITH leaves behind: its alias where it made one, and the plain
+ * variable where it carried one through whole. `WITH f.name AS caller` leaves
+ * `caller` and nothing else — `f` is gone, which is the case #1919 is about. */
+static int collect_with_names(const cbm_return_clause_t *wc, const char **out, int cap) {
+    int n = 0;
+    for (int i = 0; i < wc->count; i++) {
+        const cbm_return_item_t *item = &wc->items[i];
+        const char *name = NULL;
+        if (item->alias) {
+            name = item->alias;
+        } else if (item->variable && !item->property && !item->func) {
+            name = item->variable;
+        }
+        if (name && !scope_holds(out, n, name)) {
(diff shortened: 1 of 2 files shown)


DeusData:
Merged as `a6468ed4`.

The framing in your description is the reason this went in quickly, and it is worth repeating back because it is the right standard for this project: *"The silence is the harm."* A blank column reads as "the graph holds no such data" when the truth is "your query named something that is out of scope" — and for a code graph a false **nothing calls X** is the worst answer it can give. That is a graph-quality argument, not a validation nicety, and it is why this is a fix rather than a feature.

Two things made it reviewable without a round trip:

**You drew the line the fix could have blurred, and then tested it.** Checking the *parsed query* rather than the run-time bindings is the whole distinction — "did the query declare this name" is a different question from "did a row happen to bind it". That is what keeps an `OPTIONAL MATCH` target that matched nothing legal and still projecting `""`, and you have `cypher_optional_match_target_still_allowed_issue1919` holding that line explicitly rather than leaving it to be discovered later.

**Two negative controls, not just the reproduction.** The `OPTIONAL MATCH` case and the `WITH`-alias case mean the fix cannot pass by over-rejecting. A guard like this fails in the over-eager direction far more often than the under-eager one, and both controls are registered in a suite that actually runs.

Citing #373 as the same failure shape was also the right move — it made the precedent do the arguing instead of the description.

Thank you. 35 checks green, no round trips needed.

PR #1925, opened 2026-08-30
Add memory-aware concurrent indexing for large repositories

zhiyuzhang001-a11y:
## Summary

- scope daemon index workers to the canonical request repository instead of inheriting the daemon starter workspace boundary
- add bounded two-pass indexing so large repositories do not retain the full extraction set in memory
- add a global memory-aware scheduler for concurrent indexing across projects, while preserving exact output parity

## Motivation

This enables multiple project-scoped MCP clients to share the Provider safely. Small daily repositories can run concurrently; large jobs are admitted according to measured source bytes and memory budget instead of starting without a global bound.

## Validation

- rebased onto current upstream `main`
- `git diff --check` passes
- ASan/UBSan focused suites: 788 passed, 1 platform skip
- suites: subprocess, daemon_application, discover, pipeline, extraction
- the same three commits previously passed the full local Provider suite: 7477 passed, 4 platform skips
- downstream Codebase Atlas product suite: 237/237
- independent root acceptance suite: 47/47

No release or binary distribution is included in this PR.

### DIFF
--- a/internal/cbm/arena.c
+++ b/internal/cbm/arena.c
@@ -9,6 +9,7 @@ void cbm_arena_init(CBMArena *a) {
     a->block_size = CBM_ARENA_DEFAULT_BLOCK_SIZE;
     a->blocks[0] = (char *)malloc(a->block_size);
     if (a->blocks[0]) {
+        a->block_sizes[0] = a->block_size;
         a->nblocks = SKIP_ONE;
     }
 }
@@ -26,6 +27,7 @@ static int arena_grow(CBMArena *a, size_t min_size) {
         return 0;
     }
     a->blocks[a->nblocks] = block;
+    a->block_sizes[a->nblocks] = new_size;
     a->nblocks++;
     a->block_size = new_size;
     a->used = 0;
@@ -39,6 +41,16 @@ void *cbm_arena_alloc(CBMArena *a, size_t n) {
     // 8-byte alignment
     n = (n + 7) & ~(size_t)7;
 
+    if (n <= 64) {
+        a->alloc_le_64 += n;
+    } else if (n <= 256) {
+        a->alloc_le_256 += n;
+    } else if (n <= 4096) {
+        a->alloc_le_4096 += n;
+    } else {
+        a->alloc_gt_4096 += n;
+    }
+
     if (a->nblocks == 0) {
         return NULL;
     }
@@ -51,6 +63,7 @@ void *cbm_arena_alloc(CBMArena *a, size_t n) {
 
     char *ptr = a->blocks[a->nblocks - SKIP_ONE] + a->used;
     a->used += n;
+    a->total_alloc += n;
     return ptr;
 }
 
@@ -60,6 +73,7 @@ char *cbm_arena_strdup(CBMArena *a, const char *s) {
     size_t len = strlen(s);
     char *dst = (char *)cbm_arena_alloc(a, len + SKIP_ONE);
     if (dst) {
+        a->strdup_alloc += len + SKIP_ONE;
         memcpy(dst, s, len + SKIP_ONE);
     }
     return dst;
@@ -70,6 +84,7 @@ char *cbm_arena_strndup(CBMArena *a, const char *s, size_t len) {
         return NULL;
     char *dst = (char *)cbm_arena_alloc(a, len + SKIP_ONE);
     if (dst) {
+        a->strdup_alloc += len + SKIP_ONE;
         memcpy(dst, s, len);
         dst[len] = '\0';
     }
@@ -91,6 +106,7 @@ char *cbm_arena_sprintf(CBMArena *a, const char *fmt, ...) {
     if (!dst) {
         return NULL;
     }
+    a->sprintf_alloc += (size_t)needed + SKIP_ONE;
 
     va_start(args, fmt);
     vsnprintf(dst, (size_t)needed + SKIP_ONE, fmt, args);
--- a/internal/cbm/arena.h
+++ b/internal/cbm/arena.h
@@ -1,41 +1,13 @@
-#ifndef CBM_ARENA_H
-#define CBM_ARENA_H
+/*
+ * Extraction compatibility include.
+ *
+ * CBMArena used to be duplicated here and in src/foundation/arena.h. The
+ * production binary links the foundation implementation, so any field drift
+ * between those definitions is an ABI violation. Keep one canonical layout.
+ */
+#ifndef CBM_EXTRACTION_ARENA_COMPAT_H
+#define CBM_EXTRACTION_ARENA_COMPAT_H
 
-#include <stddef.h>
+#include "../../src/foundation/arena.h"
 
-// CBMArena is a simple bump allocator that allocates from fixed-size blocks.
-// All memory is freed at once via cbm_arena_destroy(). Individual frees are not
-// supported — this is by design for per-file extraction where all data has the
-// same lifetime.
-#define CBM_ARENA_MAX_BLOCKS 256
-#define CBM_ARENA_DEFAULT_BLOCK_SIZE (64 * 1024) // 64KB initial
-
-typedef struct {
-    char *blocks[CBM_ARENA_MAX_BLOCKS];
-    size_t block_sizes[CBM_ARENA_MAX_BLOCKS]; // per-block sizes (for stats)
-    int nblocks;
-    size_t block_size;
-    size_t used;        // bytes used in current block
-    size_t total_alloc; // cumulative bytes allocated (for stats)
-} CBMArena;
-
-// Initialize an arena with the default block size.
-void cbm_arena_init(CBMArena *a);
-
-// Allocate n bytes from the arena. Returns NULL on OOM or block exhaustion.
-// All returned pointers are 8-byte aligned.
-void *cbm_arena_alloc(CBMArena *a, size_t n);
-
-// Duplicate a string into arena memory. Returns arena-owned copy.
-char *cbm_arena_strdup(CBMArena *a, const char *s);
-
-// Duplicate a string of known length into arena memory. NUL-terminates.
-char *cbm_arena_strndup(CBMArena *a, const char *s, size_t len);
-
-// sprintf into arena memory. Returns arena-owned string.
-char *cbm_arena_sprintf(CBMArena *a, const char *fmt, ...) __attribute__((format(printf, 2, 3)));
-
-// Free all blocks. Arena is invalid after this call.
-void cbm_arena_destroy(CBMArena *a);
-
-#endif // CBM_ARENA_H
+#endif /* CBM_EXTRACTION_ARENA_COMPAT_H */
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -105,14 +105,12 @@ void cbm_reset_profile(void) {
 
 #define GROW_ARRAY(arr, arena)                                                                   \
     do {                                                                                         \
+        (void)(arena);                                                                           \
         if ((arr)->count >= (arr)->cap) {                                                        \
             int new_cap = (arr)->cap == 0 ? CBM_SZ_32 : (arr)->cap * PAIR_LEN;                   \
-            void *new_items = cbm_arena_alloc((arena), (size_t)new_cap * sizeof(*(arr)->items)); \
+            void *new_items = realloc((arr)->items, (size_t)new_cap * sizeof(*(arr)->items));    \
             if (!new_items)                                                                      \
                 return;                                                                          \
-            if ((arr)->items && (arr)->count > 0) {                                              \
-                memcpy(new_items, (arr)->items, (size_t)(arr)->count * sizeof(*(arr)->items));   \
-            }                                                                                    \
             (arr)->items = new_items;                                                            \
             (arr)->cap = new_cap;                                                                \
         }                                                                                        \
@@ -1679,6 +1677,28 @@ void cbm_free_result(CBMFileResult *result) {
     free(result->owned_results);
     result->owned_results = NULL;
     result->owned_result_count = 0;
+#define FREE_RESULT_ARRAY(field)                                                                 \
+    do {                                                                                         \
+        free(result->field.items);                                                               \
+        result->field.items = NULL;                                                              \
+        result->field.count = 0;                                                                 \
+        result->field.cap = 0;                                                                   \
+    } while (0)
+    FREE_RESULT_ARRAY(defs);
+    FREE_RESULT_ARRAY(calls);
+    FREE_RESULT_ARRAY(imports);
+    FREE_RESULT_ARRAY(usages);
+    FREE_RESULT_ARRAY(throws);
+    FREE_RESULT_ARRAY(rw);
+    FREE_RESULT_ARRAY(type_refs);
+    FREE_RESULT_ARRAY(env_accesses);
+    FREE_RESULT_ARRAY(type_assigns);
+    FREE_RESULT_ARRAY(impl_traits);
+    FREE_RESULT_ARRAY(resolved_calls);
+    FREE_RESULT_ARRAY(string_refs);
+    FREE_RESULT_ARRAY(infra_bindings);
+    FREE_RESULT_ARRAY(channels);
+#undef FREE_RESULT_ARRAY
     cbm_arena_destroy(&result->arena);
     free(result);
 }
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -103,17 +103,18 @@ void cbm_reset_profile(void) {
 
 // --- Growable array push functions ---
 
-#define GROW_ARRAY(arr, arena)                                                                \
-    do {                                                                                      \
-        (void)(arena);                                                                        \
-        if ((arr)->count >= (arr)->cap) {                                                     \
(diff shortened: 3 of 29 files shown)


DeusData:
Read in full and routed for a maintainer fit decision. Since the only thing on this thread so far is the automated acknowledgement, here is honestly where it stands rather than silence.

CI is green and the validation you supplied is unusually thorough for a first contribution — ASan/UBSan focused suites, the full local suite, and named downstream results. That is not what is holding it.

What needs a maintainer ruling is that this is **three separable changes in one PR**, and one of them is a one-way door:

1. **Scoping daemon index workers to the canonical request repository.** This moves a workspace boundary, which is a security surface here — it was hardened deliberately in a recent round, on both entry points.
2. **Bounded two-pass indexing** so a large repository does not retain the full extraction set in memory. This one is self-contained and the easiest to reason about on its own.
3. **A global memory-aware scheduler admitting concurrent indexing across projects by measured source bytes against a memory budget.** This is the one-way door: a scheduler with a budget is a permanent maintenance surface, and its knobs outlive whoever tuned them.

Each of those is reviewable on its own evidence; together, a reservation about any one blocks all three. That is why the split matters and it is not a formality — but I am not going to ask you to do the splitting work before the direction call on (3) is made, because the answer changes what the split should look like.

One thing that would genuinely help the decision, if you want to write it while this is queued: what does the scheduler do when the budget is wrong — too low, or too high on a machine it mis-measures? Failing open, failing closed, and degrading to sequential are three different products, and which one it is matters more than the admission arithmetic.

I will come back with the ruling rather than leaving this to age.

DeusData:
The maintainer ruling is in. Short version: **please split this into three PRs**, and the third one needs to work on all three platforms before it can be reviewed.

## Split into three

1. **Daemon workers scoped to the canonical request repository.** Reviewable on its own. It moves a workspace boundary, which is a security surface here — it was hardened deliberately on both entry points recently — so it deserves to be looked at without a scheduler in the same diff.
2. **Bounded two-pass indexing.** Take this one first if you want the quickest path to a merge. It is self-contained, platform-independent, and it addresses a problem this project demonstrably has on its own benchmarks: a single Linux kernel index has peaked at 16.45 GB, and a 2× run OOM'd on a 36 GB machine. Not retaining the full extraction set is squarely on the roadmap regardless of anything else here.
3. **The global memory-aware scheduler.** See below.

## What (3) needs first: cross-platform measurement

The admission logic rests on `application_worker_observed_rss`, and that path is:

```c
DIR *directory = opendir("/proc");
if (!directory) {
    return false;
}
```

…then a walk of `/proc/<pid>/stat` and `/proc/<pid>/statm`.

`/proc` does not exist on macOS or Windows, both of which are first-class platforms here. On those, `opendir` fails, the probe returns false, `job->observed_rss_bytes` is never advanced from zero, `application_active_observed_rss_locked` therefore sums to zero, and `daily_memory_busy` can never become true. Admission falls back to slot counting with no memory awareness at all.

Failing open is the **right** choice over failing closed — nobody wants indexing blocked because a probe is unavailable. The problem is that the result is a scheduler which ships its code, its five environment knobs and its maintenance cost to all three platforms while protecting only one, and does so invisibly: there is no signal at runtime that the memory half is inert.

So (3) is reviewable once the RSS probe either works natively on macOS (`task_info` / `proc_pidinfo`) and Windows (`GetProcessMemoryInfo`), or explicitly degrades to a stated conservative bound and says so in a log line, rather than silently to none.

To be honest with you about scope: that is a substantial piece of platform work, and it is entirely reasonable to decide it is more than you want to take on right now. Splitting (1) and (2) out means neither is held hostage to that decision.

## One design question for whenever (3) comes back

`CBM_INDEX_RESOURCE_MODE`, `CBM_LARGE_REPOSITORY_FILE_THRESHOLD`, `CBM_LARGE_REPOSITORY_BATCH_FILES`, `CBM_LARGE_REPOSITORY_SOURCE_BYTES` and `CBM_STREAMING_BATCH_FILES` are five new environment variables, and an environment variable is a one-way door — once it is documented, it is supported. Worth asking which of them are genuinely operator-facing policy and which are internals that could be derived or fixed constants.

## Credit where it is due

The validation you supplied is well beyond what a first contribution usually carries: ASan/UBSan focused suites, the full local suite, and named downstream results rather than "tests pass". The `/proc` walk correctly filters by process group rather than assuming a flat child list, and the probe returns a status instead of a bogus zero — which is exactly why the platform gap was findable by reading rather than by an incident.

None of the above is a rejection of the idea. It is a request to let the two uncontroversial pieces land on their own evidence while the third gets the platform work it needs.

shafty023:
@DeusData @zhiyuzhang001-a11y  I built part 2 on its own and measured it against our repo. It does what it says: peak goes from
10.2 GB to 1.6 GB and the live set stays at one batch. Only part 2 below, nothing on the scheduler
or the workspace scoping.

13,999 files, mostly TypeScript, Linux aarch64, gcc 13.3, 12 GB container, three runs per row. Each
run is `CBM_CACHE_DIR=<private> CBM_PROFILE=1 CBM_WORKERS=4 [CBM_STREAMING_BATCH_FILES=512]
codebase-memory-mcp cli --json index_repository --repo-path <tree> --mode full`, and peak is the
highest `mem.phase peak_mb`.

| build | peak MB | post-extract RSS MB | nodes | edges | wall s |
|---|---|---|---|---|---|
| main at 3de05cd6 | 10242 / 10171 / 10272 | 9342 / 9932 / 9954 | 164693 | 835843 | 118 |
| this PR, streaming off (the ungated changes alone) | 6770 / 6818 / 6825 | 5332 / 5386 / 5393 | 164693 | 835845 to 835857 | 68 to 83 |
| this PR, `CBM_STREAMING_BATCH_FILES=512` | 1600 / 1620 / 1613 | n/a | 164693 | 835843 to 835847 | 119 to 145 |

Two of the three stock runs got SIGKILLed at `lsp_cross_prepare`. The streaming runs never went
over 1336 MB across all 56 batches. Parity between the last two rows is exact on nodes,
`parse_partial_count` (280) and registry (787960 / 805005 / 100027), and edges differ by at most 14,
which is inside the spread we get from an unmodified build. Every file is parsed twice, so our wall
clock mean went from 77 s to 130 s. Worth it at that ratio for us, but it is not free.

Four things before this lands on its own.

`dd3f4e84` does not compile at its parent. The streaming call to `cbm_pxc_collect_all_defs` is
missing its `ctx` argument and only `1e15c635` adds it, so split out first this breaks the build and
any bisect through it.

The streaming path builds the go, python, c, cs and ts cross registries but not java, which the
full-cache path does build. Java cross-file calls quietly fall back to per-file resolution in
large-repo mode. Is that intentional?

`CBM_DISABLE_LSP_CROSS` is only read in the full-cache path, so the documented opt-out for the large
TypeScript SIGSEGV does nothing in the mode that exists for large repositories, which is where I
would reach for it.

Only the batching is behind the knob. The inline `TSNodeStack`, the result-array ownership change
and the determinism rewrite of `pass_complexity.c` change every index on every repo, and on our
numbers they are 3.4 GB of the 8.6 GB drop against the batching's 5.2 GB. Is the intent for those
three to ship ungated inside the part 2 split, or should they be their own commits so the batching
lands on its own evidence? The `TSNodeStack` one also overlaps the retention fix in #2013, so we
should reconcile those two either way.

Nit: `compare_node_qn` in `pass_complexity.c` has no id tiebreak, unlike `compare_tld_target` right
above it, and qsort is not stable, so nodes sharing a qualified name still come out in arbitrary
order.

DeusData:
@shafty023 -- thank you for building part 2 in isolation and measuring it properly. Three runs per row, peak and post-extract RSS separated, parity checked on nodes / `parse_partial_count` / registry sizes: that is exactly the evidence the split needs, and 10.2 GB -> 1.6 GB with a bounded live set is the result the ruling above was betting on. The 77 s -> 130 s wall cost is stated honestly too, which matters; nobody should discover that after merging.

On your four points, from the maintainer side so @zhiyuzhang001-a11y is not left guessing:

1. **Every commit in the part-2 PR must build at its parent.** `dd3f4e84` calling `cbm_pxc_collect_all_defs` without the `ctx` that `1e15c635` introduces is a bisect trap; fold or reorder so the streaming call lands together with its signature.
2. **The Java cross registry missing from the streaming path is a defect, not a choice we made.** Large-repo mode silently falling back to per-file Java resolution would be a graph-quality regression precisely on the repos this mode exists for. Same class of problem for `CBM_DISABLE_LSP_CROSS` being read only on the full-cache path: the documented opt-out has to work in the mode where people reach for it.
3. **Yes -- the three ungated changes should land on their own evidence, separately from the batching.** Your numbers make the case: 3.4 GB of the drop comes from changes every index on every repo would get regardless of the knob. Each of those deserves its own commit (or PR) with its own before/after, so the batching PR is judged on the batching alone. The `TSNodeStack` piece needs reconciling against your #2013, which is already on main.
4. The `compare_node_qn` tiebreak nit is a real determinism hole given qsort is unstable -- worth fixing in the same pass as the ordering rewrite it belongs to.

Nothing here changes the shape of the ruling: part 2 first, self-contained, cross-platform-neutral, one buildable commit per change.

DeusData:
Heads-up, with thanks: while this PR is under review as a feature, two defects it repairs in passing are too valuable to wait, so they are being split out as their own small PRs with you credited as co-author — the index worker inheriting the daemon's `CBM_ALLOWED_ROOT` is #2076 (confirmed live: a daemon started with one allowed root refused every other session's repository inside the worker), and the worker-order dependence of the complexity pass follows next (confirmed run to run on adjacent equal-size cycle members). When you next rebase this branch those two hunks will drop out on their own; the feature itself stays under discussion here as before.

zhiyuzhang001-a11y:
Thanks for the review and the separate fixes. I have rechecked that #2076 and #2079 are open and #2013 is on main. I will leave worker scoping and complexity ordering to those PRs and prepare a standalone bounded two-pass indexing PR against current main; the global scheduler is deferred.

The batching split will include the missing Java cross-file registry and honor CBM_DISABLE_LSP_CROSS. I will keep the unconditional TSNodeStack/result-array ownership changes out of that diff, retain the #2013 scratch-lifetime fix, and leave the qualified-name tie-break with the ordering work. Each new commit will build independently and carry DCO sign-off.

This is the agreed scope, not a claim that the revised implementation or validation is already complete. I will link the new PR here with reproducible graph-parity, peak-memory and timing results, and distinguish local tests from platform CI. #1925 will remain the discussion entry point.

zhiyuzhang001-a11y:
The standalone batching PR is now #2082 (tracking issue #2081), with three signed commits separating the buildable primitives, the batching pipeline/tests, and reproducible measurements. It is now ready for review.

Implemented in the split: Java shared cross registry, CBM_DISABLE_LSP_CROSS parity, def collection after the complete registry/IMPORTS graph, whole-project macros/pkgmap, replay diagnostic deduplication, and final graph-ID handoff. The scheduler and unconditional stack/array/ordering changes are excluded; #2013 is retained and #2076/#2079 remain yours.

Local ASan/UBSan pipeline is 275/275; the primitives commit and final head each build independently. Three-run native macOS measurements now include whole-worker wait4 RSS, not only phase logs: generated 2,049-file TS corpus about 649 → 301 MiB median, 2.64 → 3.23 s worker wall; Repomix about 272 → 194 MiB, 2.06 → 2.42 s. All 15 worker runs match normalized persisted graph properties/relationships and LSP surfaces exactly against their baseline. The complete canonical scripts/test.sh entry point has now passed: 7,952 passed, 0 failed, 7 Windows-only skips across 141 suites, plus all contract/process guards. GitHub DCO passed; the other jobs remain queued, so native platform CI is still pending. This PR remains open as the overall discussion entry point.

asd59478126-commits:
# Lineage-Driven Event Sourcing Schema

## Reference Template

The following is a concrete example of how the schema is constructed and how the data is routed.

### 1. Lineage Graph Number

```text
Lineage Graph Number: 3
```

This number identifies the bounded lineage segment in which the data is located.

It is a routing and segmentation number.

It does not represent AI execution order.

---

### 2. Lineage Graph Record Time

```text
Record Date: 2026-09-07
Record Time: 05:00
```

The date and time record when this Lineage Graph was created, organized, or recorded.

No separate version number is required.

The temporal record itself provides the time-based version of the lineage structure.

---

### 3. Event Node Sequence Number Range

```text
Event Node Sequence Range: 12–27
```

This means that Lineage Graph `3` contains Event Nodes `12` through `27`.

Each Event Node corresponds to an actual position in the source data.

---

### 4. Extracted Text Unit

A later extraction process retrieves a specific text unit from the source:

```text
Extracted Text Unit:
Source lines 1840–1864
```

The extracted text unit is not required to be semantically interpreted at this point.

Its source position and boundary are sufficient for routing.

---

### 5. Routing to the Event Node

The routing procedure uses the extracted text unit's source position and the established segmentation boundaries to determine the corresponding Event Node.

```text
Extracted Text Unit
        ↓
Source Position
        ↓
Segmentation Boundary
        ↓
Routing Procedure
        ↓
Event Node 19
```

Result:

```text
Lineage Graph: 3
Event Node: 19
Source Position: lines 1840–1864
```

The routing procedure answers:

**Which Event Node does this extracted text unit belong to?**

It does not need to determine the semantic meaning of the event.

---

### 6. Event Unit

The located Event Node is the event itself.

```text
Event Node: 19
Source Position: lines 1840–1864
Event: [the actual source text belonging to this event]
```

The event is not converted into another artificial storage object simply for routing.

The event remains the event.

---

### 7. Semantic Processing

Only after the event has been located may semantic processing be performed.

```text
Event Node 19
        ↓
Semantic Processing
        ↓
Semantic Relation(s)
```

The semantic relationship is therefore attached after localization.

It does not determine the original physical position of the event.

---

# Complete Example

The complete record can therefore be represented as:

```text
Lineage Graph Number: 3
Record Date: 2026-09-07
Record Time: 05:00
Event Node Sequence Range: 12–27

Extracted Text Unit:
Source lines 1840–1864

Routing:
Source position + established boundary
        ↓
Event Node 19

Event Node:
19
Source lines 1840–1864

Semantic Processing:
Performed after localization
```

The complete process is:

```text
Lineage Graph Number
        ↓
Record Date / Time
        ↓
Event Node Sequence Range
        ↓
Extracted Text Unit
        ↓
Routing Procedure
        ↓
Event Node
        ↓
Semantic Processing
```

## Database Representation

A database record can therefore preserve the actual structure directly:

```text
Lineage Graph Number | Record Date | Record Time | Event Node | Source Position | Event
3                    | 2026-09-07 | 05:00       | 19         | 1840–1864       | [event text]
```

Semantic information may be derived afterward from the Event itself.

The important point is that the database does not need to create a separate artificial identity system in order to route the event.

The Lineage Graph Number provides the routing location.

The Event Node provides the event unit.

The source position provides localization.

The date and time provide the temporal record.

---

## Core Rule

**First establish where the data is.
Then establish which event unit it belongs to.
Then establish its semantic relationships.**

Or:

**Locate → Route → Identify Event → Interpret**

The Lineage-Driven Event Sourcing Schema is therefore primarily a method for making large quantities of data physically addressable and traceable before semantic interpretation is applied.


# Lineage-Driven Event Sourcing Schema

## Formal Schema Example

A formal instance of the schema can be represented as follows:

```text
Lineage Graph Number
    3

Lineage Graph Record Time
    2026-09-07 05:00

Event Node Sequence Number Range
    12–27

Extracted Text Unit
    Source position: lines 1840–1864

Routing Procedure
    Extracted Text Unit
        ↓
    Source Position
        ↓
    Lineage Boundary
        ↓
    Routing Algorithm
        ↓
    Event Node 19

Resolved Event Node
    Graph: 3
    Event Node: 19
    Source Position: lines 1840–1864
    Event: [actual event text]

Semantic Processing
    Applied after Event Node localization
```

The corresponding formal structure is:

```text
L = {
    lineage_graph_number,
    lineage_graph_record_time,
    event_node_sequence_range,
    extracted_text_unit,
    routing_procedure,
    event_node
}
```

For the example above:

```text
L = {
    lineage_graph_number: 3,
    lineage_graph_record_time: 2026-09-07 05:00,
    event_node_sequence_range: 12–27,
    extracted_text_unit: lines 1840–1864,
    routing_procedure: route(extracted_text_unit) → event_node_19,
    event_node: 19
}
```

The routing function is:

```text
route(extracted_text_unit)
    → source position
    → established boundary
    → corresponding Event Node
```

The semantic operation occurs only after the routing result has been established:

```text
route(extracted_text_unit)
    → Event Node
    → Semantic Processing
```

Therefore, the formal construction order is:

```text
Lineage Graph Number
→ Lineage Graph Record Time
→ Event Node Sequence Number Range
→ Extracted Text Unit
→ Routing Procedure
→ Event Node
→ Semantic Processing
```

The schema does not require a separate artificial identifier, a separate version number, or a separate storage layer.

The Lineage Graph Number is the routing location.

The Event Node is the event unit.

The source position provides localization.

The date and time provide the temporal record.

Semantic interpretation is performed afterward.

## Formal Request

**Please provide a formal schema example showing how a source text unit is mapped through the Lineage Graph Number, record time, Event Node sequence range, and routing procedure to the corresponding Event Node.**

DeusData:
Status note: the complexity-pass determinism piece from this PR (the pass depended on worker id order, so parallel runs were not byte-identical) landed on main as 29dbf0bfa150 with co-author credit (#2079), and the worker-scoping fix is queued separately as #2076. Both go into v0.10.9. The scheduler and bounded two-pass parts stay with this PR and #2082 for the design discussion. Thank you again for the work that surfaced them.

PR #1931, opened 2026-08-30
fix(extract): drop cgo's `import "C"` pseudo-package

### LINKED ISSUE #1926: fix(extract): Go — `import "C"` (cgo pseudo-package) resolves to an arbitrary project symbol named C
# fix(extract): Go — `import "C"` (cgo pseudo-package) resolves to an arbitrary project symbol named `C`

**Status: fix open — PR #1931** (drop the spec in `parse_go_import_spec()` when the path is exactly `"C"`). Measured: 27 false `IMPORTS` edges → 0. The general non-cgo case of the same mechanism is #1934/#1938.

## What happens

`import "C"` is not an import — `"C"` is reserved by `go/build`; no such package can exist. The clause only marks the preceding comment block as C. cbm's Go import parser keeps it as an ordinary spec; it fails Strategies 1–2 in `cbm_pipeline_resolve_import_node()` and lands in Strategy 3's symbol-name fallback (`pass_pkgmap.c`), which returns the lexicographically smallest project symbol named `C`.

## Damage measured (~1150-file Go+C repo)

27 files carry `import "C"` → 27 bogus `IMPORTS` edges (100%), all converging on one unrelated `C` member of a test helper. The winner is a lexicographic tie-break, so which symbol gets hijacked changes as the repo grows.

## Minimal reproduction

`fx/cgo_use.go`:
```go
package fx

/*
static int helper(int a) { return a + 1; }
*/
import "C"

func Run(a int) int { return int(C.helper(C.int(a))) }
```

`fx/sub/chan_holder.go`:
```go
package sub

type C struct{ N int } // ordinary project symbol that happens to be named C

func NewC() *C { return &C{} }
```

Before: `IMPORTS` edge `cgo_use.go → sub.C`. After: no rows.

## Detection recipe

```sql
SELECT count(*) FROM edges
WHERE type = 'IMPORTS' AND json_extract(properties, '$.local_name') = 'C';
```

Nonzero on a cgo repo = this bug.

## Fix + test

Skip the `"C"` spec before it is pushed — Go-specific knowledge belongs in the Go import parser, not the language-agnostic resolver. Reproduce-first test `go_cgo_pseudo_import_dropped` (RED without the change; also pins that the file's real `fmt` import survives).

Related: #1934 (general Go import→symbol case), #1929 (cgo modelling), #725; tracked in #1932.

ilyabrykau-orca:
Fixes #1926

## Problem

`import "C"` is not an import. `"C"` is reserved by `go/build`; no package with that
import path can exist. The clause only tells the Go toolchain to compile the
immediately preceding comment block as C.

The Go import parser treats it as an ordinary import spec, so every cgo file's
import list — and therefore the per-file import map that call resolution
consults — carries a package that cannot exist. Dropping it at extraction is the
fix: Go-specific knowledge belongs in the Go import parser rather than in the
language-agnostic resolver.

### History: the false-edge symptom this PR originally led with

When #1926 was filed (2026-08-30), the pseudo-import also produced false `IMPORTS`
edges: it fell through Strategies 1 and 2 and landed in the Strategy 3 symbol-name
fallback (`src/pipeline/pass_pkgmap.c`), which bound all **27** cgo files of the
measured Go+C repository to the same unrelated `C` member of a test helper.

That measurement **predates `7d76b88f`** ("fix(pipeline): never bind a Go import to
a symbol", merged 2026-08-31 via #1938), which gates Strategy 3 off for Go entirely.
Re-measured on current `main` (`8972ea69`) against the same repository — which now
carries **54** `import "C"` files: `IMPORTS` edges with target `C` (by node name or
by `local_name`/`module` property): **0**. The Go `IMPORTS` target histogram is
100% `Folder` (2,303 edges). The false edges are gone exactly as the Strategy-3
gating predicts, so this PR no longer removes any edge; its effect is confined to
the extraction layer.

## Change

`internal/cbm/extract_imports.c`, `parse_go_import_spec()` — skip the spec when the
import path is exactly `"C"`, before the import is pushed. The guard cannot lose a
real import: `go/build` reserves the path, so no resolvable package can ever be
spelled `"C"`.

## Effect on current main

- The file's import list, and the per-file import map derived from it, no longer
  name a package that cannot exist — so import-closure reasoning (e.g. the
  penalty logic in `unique_name` resolution) never consults a phantom package.
- Graph shape is unchanged: the repository-level `IMPORTS` count with target `C`
  is 0 both before and after this change on current main; the pre-#1938 history
  above is what the original description measured.

## Test

`tests/test_extraction.c` — `go_cgo_pseudo_import_dropped`.

Reproduce-first (extraction-level, so it stays RED-capable regardless of the
resolver gating): RED without the extractor change —

```
go_cgo_pseudo_import_dropped   FAIL tests/test_extraction.c:2881:
  ASSERT(strcmp(r->imports.items[i].module_path, "C") != 0)
  318 passed, 1 failed
```

GREEN with it — `319 passed`.

The test also asserts the same file's real `fmt` import survives, so the fix cannot
regress into dropping the whole import block.

## Checks

- `scripts/test.sh --suites extraction` — 319 passed, 0 failed.
- `clang-format` clean on both edited hunks.
- `scripts/lint.sh` could not complete locally: `clang-tidy` and `cppcheck` are not
  installed on this machine. Relying on CI for those two legs.

### DIFF
--- a/internal/cbm/extract_imports.c
+++ b/internal/cbm/extract_imports.c
@@ -133,6 +133,19 @@ static void parse_go_import_spec(CBMExtractCtx *ctx, TSNode spec) {
         return;
     }
 
+    /* cgo: `import "C"` names the cgo PSEUDO-package, not a real one — "C" is
+     * reserved by go/build and no package with that import path can exist.
+     * Emitting it as an ordinary import makes the import resolver fall through
+     * to its symbol-name fallback (cbm_pipeline_resolve_import_node Strategy 3,
+     * pass_pkgmap.c), which matches the literal name "C" against every project
+     * node called C and picks the lexicographically smallest: on a real Go repo
+     * all 27 cgo files pointed their IMPORTS edge at the same unrelated `C`
+     * member of a test helper. Drop it — the C side is not a graph package, and
+     * the file's remaining imports are unaffected. */
+    if (strcmp(path, "C") == 0) {
+        return;
+    }
+
     TSNode name_node = ts_node_child_by_field_name(spec, TS_FIELD("name"));
     const char *local_name =
         !ts_node_is_null(name_node) ? cbm_node_text(a, name_node, ctx->source) : path_last(a, path);
--- a/tests/test_extraction.c
+++ b/tests/test_extraction.c
@@ -3015,6 +3015,25 @@ TEST(go_imports) {
     PASS();
 }
 
+/* cgo's `import "C"` is a pseudo-package, not a real import: keeping it lets the
+ * import resolver name-match "C" onto an arbitrary project symbol called C. The
+ * real imports of the same file must survive. */
+TEST(go_cgo_pseudo_import_dropped) {
+    CBMFileResult *r = extract("package m\n\n/*\nstatic int helper(void) { return 1; }\n*/\n"
+                               "import \"C\"\n\nimport \"fmt\"\n\n"
+                               "func Run() { fmt.Println(C.helper()) }\n",
+                               CBM_LANG_GO, "t", "cgo.go");
+    ASSERT_NOT_NULL(r);
+    ASSERT_FALSE(r->has_error);
+    ASSERT(has_import(r, "fmt"));
+    for (int i = 0; i < r->imports.count; i++) {
+        ASSERT_NOT_NULL(r->imports.items[i].module_path);
+        ASSERT_TRUE(strcmp(r->imports.items[i].module_path, "C") != 0);
+    }
+    cbm_free_result(r);
+    PASS();
+}
+
 /* #1935: Go struct fields were never extracted — find_class_body() returns the
  * struct_type node, whose only named child is a field_declaration_list, so the
  * member loop matched nothing and every field was silently skipped (0 Field
@@ -6752,6 +6771,7 @@ SUITE(extraction) {
     RUN_TEST(python_imports);
     RUN_TEST(js_imports);
     RUN_TEST(go_imports);
+    RUN_TEST(go_cgo_pseudo_import_dropped);
     RUN_TEST(extract_go_struct_fields_have_nodes);
     RUN_TEST(java_imports);
     RUN_TEST(rust_imports);


DeusData:
Heads-up: this went red-to-merge because **#1937 (`fcd8bfc1`) and #1940 (`61de19bb`) landed**, not because of anything you changed.

Checked with `git merge-tree` against current `main`: the conflict is **exactly one file, `tests/test_extraction.c`**, and it is insertion-anchor drift — #1940 added its Go struct-field extraction test there. Every source file in your diff auto-merges cleanly.

Re-anchor your test and its `RUN_TEST` line and the rebase is done. Ping when it is up.

ilyabrykau-orca:
Rebased onto current main — as your merge-tree note predicted, the only conflict was test insertion-anchor drift (tests/test_extraction.c), re-anchored. Full scripts/test.sh on the rebased head: 7782 passed, 1 failed — the failure is test_daemon_runtime's marker_published assert, a load-timing flake from running three suites concurrently on this machine; it passes solo (47/47) and is unreachable from this diff (extract_imports only).

DeusData:
This is green and clean, and I was about to merge it — then checked the mechanism and stopped. I think the headline symptom may already be fixed, and I would rather ask than land it on a premise that has moved.

## The question

Your description routes the bug through Strategy 3:

> fell through Strategy 1 … and Strategy 2 …, and landed in the Strategy 3 symbol-name fallback (`pass_pkgmap.c:1904-1998`) … all **27** files carrying `import "C"` pointed their `IMPORTS` edge at the same unrelated `C` member of a test helper

But `pass_pkgmap.c` now has:

```c
bool cbm_import_symbol_fallback_allowed(CBMLanguage lang) {
    return lang != CBM_LANG_GO;
}
```

introduced by `7d76b88f` ("fix(pipeline): never bind a Go import to a symbol", 30 August) — whose comment describes exactly your failure mode, down to the example: *"the fallback instead bound the last path segment to an arbitrary same-named project symbol (`import "os/exec"` → a test harness's `exec()` method)"*.

**And `7d76b88f` is already an ancestor of this PR's head.** So on your own branch, Strategy 3 should never run for Go: `import "C"` misses Strategy 1, misses Strategy 2, and is then skipped rather than falling through — producing no edge at all.

If that reading is right, the 27 false `IMPORTS` edges are already gone, and this change no longer removes them.

**Was the 27-file measurement taken before `7d76b88f` landed?** That is the one fact I cannot determine from here, and it decides how this PR should be described.

## Why I still think it has value

Even with the fallback disabled, `parse_go_import_spec()` still pushes `"C"` into the file's import list — so the pseudo-package remains in **the import map that call resolution consults**. Your description already names that as a separate benefit:

> dropping it at extraction also keeps the pseudo-package out of the file's import map used by call resolution

That effect is untouched by `7d76b88f`, because that change gates *edge resolution*, not *extraction*. So the fix looks correct and worth having — but for the import-map reason, not the false-edge reason.

**The guard itself is right regardless.** `"C"` is reserved by `go/build` and no package with that import path can exist, so dropping it at the Go parser cannot lose a real import — this is a Go language fact, not a heuristic. Putting Go-specific knowledge in the Go import parser rather than the language-agnostic resolver is also the right layer.

## What I need

Please confirm whether the 27-edge measurement predates `7d76b88f`, and if so update the description to lead with the import-map effect. If you re-measure on current `main` and the false edges are still there, tell me — that would mean I have misread the gating, and it is the more interesting outcome.

Not blocking on anything else: 34 checks green, `CLEAN`, and I have cleared the current head.

ilyabrykau-orca:
Confirmed: **the 27-edge measurement predates `7d76b88f`.** It was taken for #1926, filed 2026-08-30, one day before #1938 merged the Strategy-3 gate. Your reading of the gating is right, not misread.

Re-measured today on current `main` (`8972ea69`), same repository (which has since grown to **54** `import "C"` files):

- `IMPORTS` edges from `.go` files with target node named `C`: **0**
- `IMPORTS` edges with `local_name`/`module` property `'C'`: **0**
- Go `IMPORTS` target-label histogram: `Folder` 2,303 — nothing else

So the false edges are already gone on main, exactly as the gate predicts, and this PR removes none. I have rewritten the description to lead with the import-map effect: the extractor still pushes `"C"` into the file's import list, so the per-file import map that call resolution consults names a package that cannot exist; this PR stops that at the parser, where the `go/build` reservation makes the skip a language fact rather than a heuristic. The original 27-edge story is kept as history with the predates-`7d76b88f` note.

The RED/GREEN evidence is unaffected — the reproduce-first test asserts at extraction level (`r->imports.items[i].module_path`), which stays RED-capable regardless of resolver gating.

DeusData:
Thank you for remeasuring and distinguishing cgo extraction/import hygiene from the resolver false-edge work that already changed. Current import extraction still handles the `C` pseudo-import at the identified location, so that narrower claim is the useful review boundary. We should review this contribution against that corrected scope rather than repeat the earlier broader claim.

DeusData:
Merged as `9b7d63f5`. Thank you — and thank you in particular for re-measuring rather than defending the original number. Finding that #1938's gate had already removed all 27 edges, and saying so, is the harder and more useful thing to do.

I reviewed it against the corrected scope, as I said I would on 19 September, and the smaller claim holds on today's `main`: the symbol-name fallback is gated, but a `"C"` entry in the file's import list still reaches the namespace-map strategy, and a reserved pseudo-package should never reach any resolver at all. Your test pins exactly that — `"C"` absent from the import list, `fmt` still present — and with the production line reverted it fails on the `"C"` assertion. `extraction`, `extraction_imports` and `edge_imports` pass on the merge result and the memory-core linter is unchanged.

One thing I will tidy in a follow-up commit of my own rather than cost you another CI round: the code comment still narrates the 27-edge case as if it were current. I will reword it to say what it now is — hygiene for a pseudo-package that no resolver should see — and keep the history as history.

PR #1933, opened 2026-08-30
fix(opencode): pass input args to augment hook and allow tool search in subagents

### LINKED ISSUE #1737: Generated OpenCode plugin passes output?.args to hook-augment, so augmentation never fires
## Summary

The OpenCode plugin generated by `cbm install` (`~/.config/opencode/plugins/cbm-augment.ts`, produced by `src/cli/client_adapter.c`) never augments `Grep`/`Glob` results because it reads the tool arguments from the wrong place.

## Version / Reproducible

- Installed: **v0.10.5**
- Confirmed still present in **v0.10.8**: `client_adapter.c` is identical (`git diff` between the v0.10.5 and v0.10.8 tags is empty for this file), and the released `v0.10.8` source still has the same line.

## Root cause

The generated `tool.execute.after` hook calls:

```js
const extra = await augment(tool, output?.args);
```

But the OpenCode plugin `tool.execute.after` hook signature is:

```ts
"tool.execute.after"?: (
  input: { tool: string; sessionID: string; callID: string; args: any },
  output: { title: string; output: string; ... },
) => Promise<void>
```

The tool arguments live in **`input.args`**, not `output` — `output` is the tool result and has no `args` field. So `output?.args` is always `undefined`, the payload sent to `hook-augment` is `{"tool_input":{}}`, the graph lookup matches nothing, and no `[codebase-memory]` context is ever appended. Real `Grep`/`Glob` results are never augmented, even though the hook and daemon are otherwise working.

## Fix

Use `input?.args` instead:

```js
const extra = await augment(tool, input?.args);
```

This is the location in the generator:

- `src/cli/client_adapter.c` — the `"tool.execute.after"` template (`const extra = await augment(tool, output?.args);`).

I verified the fix locally: after switching to `input?.args`, `Grep` results are augmented with the expected `[codebase-memory] N graph symbol(s) match ...` block, in both a fresh index and existing indexes.

Chewji9875:
## Summary

This PR resolves two issues affecting OpenCode integration:

1. **Fix `output?.args` typo in OpenCode plugin adapter (`src/cli/client_adapter.c`)**:
   In OpenCode's plugin API for `tool.execute.after(input, output)`, the arguments passed to the executed tool reside in `input?.args`, while `output` contains the result (`output?.output`). Passing `output?.args` passed `undefined`, causing `hook-augment` to receive an empty payload `{"tool_input":{}}` and silently failing to augment search results with graph context (Closes #1737).

2. **Add `tool_search` and `tool_search_regex` permissions to OpenCode subagent profiles (`src/cli/agent_profiles.c`)**:
   When OpenCode subagents are generated with `permission: "*": deny`, they were granted `read`, `grep`, and `glob` alongside direct MCP tool permissions. In environments with deferred MCP tool loading (e.g. `openstellar-tool-search`), OpenCode needs `tool_search` or `tool_search_regex` to discover and unlock deferred tools. Adding these permissions enables subagents to unblock graph tools as intended.

## Changes
- `src/cli/client_adapter.c`: Changed `output?.args` to `input?.args` in `cbm_client_adapter_opencode`.
- `src/cli/agent_profiles.c`: Updated `CBM_GRAPH_DIALECT_OPENCODE` subagent template to allow `tool_search` and `tool_search_regex`.
- `tests/test_agent_clients.c`: Added assertions verifying `input?.args` presence and `output?.args` absence.
- `tests/test_agent_profiles.c`: Added `agent_profiles_opencode_allows_tool_search_and_subagent_permissions` test.

## Verification
- Ran `./scripts/test.sh --suites "agent_clients agent_profiles" ` — 49/49 tests passing.
- Verified C binary compilation via `./scripts/build.sh`.

### DIFF
--- a/src/cli/agent_profiles.c
+++ b/src/cli/agent_profiles.c
@@ -547,6 +547,17 @@ static bool render_profile_text(profile_buffer_t *buffer, cbm_graph_profile_dial
         }
         return true;
     case CBM_GRAPH_DIALECT_OPENCODE:
+        if (!profile_buffer_append(buffer, "---\ndescription: ") ||
+            !profile_buffer_append(buffer, description) ||
+            !profile_buffer_append(
+                buffer, "\nmode: subagent\npermission:\n  \"*\": deny\n  read: allow\n  grep: "
+                        "allow\n  glob: allow\n  tool_search: allow\n  tool_search_regex: "
+                        "allow\n") ||
+            (direct && !append_permission_mcp_tools(buffer, dialect, tier)) ||
+            !profile_buffer_append(buffer, "---\n") || !profile_buffer_append(buffer, prompt)) {
+            return false;
+        }
+        return true;
     case CBM_GRAPH_DIALECT_KILO:
         if (!profile_buffer_append(buffer, "---\ndescription: ") ||
             !profile_buffer_append(buffer, description) ||
--- a/src/cli/client_adapter.c
+++ b/src/cli/client_adapter.c
@@ -315,7 +315,7 @@ char *cbm_client_adapter_opencode(const char *binary_path) {
               "        seen.add(sid);\n"
               "        pieces.push(await lifecycle());\n"
               "      }\n"
-              "      const args = output?.args ?? {};\n"
+              "      const args = input?.args ?? {};\n"
               "      const search =\n"
               "        input?.tool === 'grep' ? 'Grep' : input?.tool === 'glob' ? 'Glob' : null;\n"
               "      if (search) {\n"
--- a/tests/test_agent_clients.c
+++ b/tests/test_agent_clients.c
@@ -1283,6 +1283,9 @@ TEST(client_adapter_opencode_sends_the_required_hook_event) {
     ASSERT_NOT_NULL(js);
     ASSERT_NOT_NULL(strstr(js, "hook_event_name: 'PreToolUse'"));
     ASSERT_NOT_NULL(strstr(js, "tool.execute.after"));
+    ASSERT_NOT_NULL(strstr(js, "const args = input?.args ?? {};"));
+    ASSERT_NOT_NULL(strstr(js, "tool_input: args"));
+    ASSERT_NULL(strstr(js, "output?.args"));
     /* OpenCode reaches the tools over MCP already; this adapter must not
      * register any, or we reintroduce the second tool surface. */
     ASSERT_NULL(strstr(js, "registerTool"));
--- a/tests/test_agent_profiles.c
+++ b/tests/test_agent_profiles.c
@@ -360,6 +360,24 @@ TEST(agent_profiles_omp_direct_has_prefixed_tools_and_handoff_excludes_mcp) {
     PASS();
 }
 
+TEST(agent_profiles_opencode_allows_tool_search_and_subagent_permissions) {
+    char *direct = cbm_render_graph_profile(CBM_GRAPH_DIALECT_OPENCODE, CBM_GRAPH_TIER_VERIFY,
+                                            CBM_GRAPH_ACCESS_DIRECT, NULL);
+    ASSERT_NOT_NULL(direct);
+    ASSERT_NOT_NULL(strstr(direct, "mode: subagent\n"));
+    ASSERT_NOT_NULL(strstr(direct, "  \"*\": deny\n"));
+    ASSERT_NOT_NULL(strstr(direct, "  read: allow\n"));
+    ASSERT_NOT_NULL(strstr(direct, "  grep: allow\n"));
+    ASSERT_NOT_NULL(strstr(direct, "  glob: allow\n"));
+    ASSERT_NOT_NULL(strstr(direct, "  tool_search: allow\n"));
+    ASSERT_NOT_NULL(strstr(direct, "  tool_search_regex: allow\n"));
+    ASSERT_NOT_NULL(strstr(direct, "  \"codebase-memory-mcp_search_graph\": allow\n"));
+    ASSERT_NOT_NULL(strstr(direct, "  \"codebase-memory-mcp_check_index_coverage\": allow\n"));
+    ASSERT_NULL(strstr(direct, "delete_project"));
+    free(direct);
+    PASS();
+}
+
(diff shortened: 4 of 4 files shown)


DeusData:
Thanks for this, and apologies that the only thing on the thread so far was the automated acknowledgement.

Three things stand between this and review. One of them is mine, not yours.

**1. Mine — a conflict I caused.** #1487 merged about an hour ago (`1e3f2fd2`), extending the OpenCode plugin across its context surfaces, and it touches the same file you do. Checked with `git merge-tree` against current `main` (`20e86c02`): the conflict is **exactly one file, `src/cli/client_adapter.c`**. `agent_profiles.c` and both test files auto-merge cleanly. Worth reading what #1487 changed there before you re-anchor — the two changes look complementary rather than overlapping, but you will want to confirm your one-line edit still lands where you intend.

**2. Lint — two clang-format violations, both on one line:**

```
src/cli/agent_profiles.c:553:24
src/cli/agent_profiles.c:553:96
```

`make -f Makefile.cbm lint-format` reproduces it. One caveat that catches people here: use the **Homebrew LLVM** clang-format — a standalone `clang-format-20` reports spurious whole-file drift.

**3. DCO — the commit needs a sign-off.** It is a single commit (`9bd1cdd6`), so amending it with `-s` and force-pushing with lease covers it.

None of that is a comment on the change itself, which at +32/−1 across four files is exactly the shape that is pleasant to review — one claim, tests alongside it. It has not been read on the merits yet; the queue is deep and that part is on us. Clear those three and it goes into review properly.

JCVIDev:
#1487 is the branch that landed on top of you here — it merged into the same file about an hour after you opened this and restructured the block your one-line edit sits in. Sorry for the re-anchor; here is what changed, so you do not have to read my diff to find out.

The good news is that the edit gets simpler, not harder. On current `main` the rewritten `tool.execute.after` block has exactly one `args` binding:

```
/* src/cli/client_adapter.c:318 */
"      const args = output?.args ?? {};\n"
```

and both consumers read from it — `tool_input: args` at 325 for the grep/glob lookup, and `args.filePath ?? args.file_path ?? args.path` at 329 for the post-`read` coverage note that #1487 added. So your change is still a single line in a single place:

```
"      const args = input?.args ?? {};\n"
```

It now fixes two surfaces instead of one. #1487 extended the defect you found to the new post-`read` path without noticing it, so your PR is worth more against post-merge `main` than it was when you filed it.

Corroborating #1737 from a different angle, in case it is useful for the review: the arguments are on `input` at all five `tool.execute.after` trigger sites inside the OpenCode binary (v1.18.25) — native tools, the three MCP resource tools, generic MCP tools, and the subtask path — not just in the published typings.

```
yield* i.trigger("tool.execute.after",
  { tool: u.id, sessionID: H.sessionID, callID: H.callID, args: b },
  V);
```

`output` there is the tool result, `{ title, output, metadata, attachments? }`, with no `args` field on any path.

Happy to help with the rebase if it would save you time.

Chewji9875:
Good mate! @JCVIDev Thanks a lot for the heads up and the detailed context on #1487 — that made the re-anchor super smooth.

Rebased on latest `main` and cleared all three items:
1. **Rebase & Conflict:** Resolved the conflict in `src/cli/client_adapter.c` with `const args = input?.args ?? {};`, which now fixes both `tool_input: args` (grep/glob) and the post-`read` path from #1487.
2. **Lint:** Wrapped `src/cli/agent_profiles.c:553` to stay strictly under the 100-column clang-format limit.
3. **DCO:** Commit amended with `-s` sign-off.

Also aligned the regression assertion in `tests/test_agent_clients.c` with the new template layout and verified all 50 tests across `agent_clients` and `agent_profiles` pass. Ready for review!

DeusData:
Thank you for the quick rebase — all three blockers are cleared on `8d46d258` (clean merge, DCO signed, lint fixed), so I did the merit review.

**The `client_adapter.c` fix is real and I want to make sure it gets the credit it deserves.** I checked it against the plugin's own type definitions rather than taking either of our words for it — `packages/plugin/src/index.ts` on `sst/opencode` `dev`:

```ts
"tool.execute.after"?: (
  input: { tool: string; sessionID: string; callID: string; args: any },
  output: { title: string; output: string; metadata: any },
) => Promise<void>
```

`args` lives on **`input`** for the `after` hook (it is `output.args` only in `tool.execute.before`, which is where the original line was presumably copied from). So `output?.args ?? {}` was always `{}`: every grep/glob augmentation went out with an empty `tool_input`, and the `read` branch never found a `filePath` — the whole tool-event augment path was silently inert for OpenCode. One token, and the three assertions you added pin exactly the fault. 

**The `agent_profiles.c` half I need you to drop, or show me the source for it.** Two things:

1. On `main` the `case CBM_GRAPH_DIALECT_OPENCODE:` label falls through into the KILO body on purpose — they render the same frontmatter, and `tests/test_agent_profiles.c:24` already pins OpenCode getting `"*": deny` / `read` / `grep` / `glob`. The new body is a byte-for-byte copy of KILO's plus two lines, so it turns one shared body into two that must be kept in sync.
2. The two added lines are `tool_search: allow` and `tool_search_regex: allow`. I could not find a tool by either name in OpenCode: zero hits in `sst/opencode` for `tool_search`, `tool_search_regex`, `toolsearch` or `ToolSearch`, and the documented permission-key table in `docs/tools.mdx` does not list them. Permission keys are matched as wildcard patterns against the underlying tool name, so an unknown key matches nothing — these two are inert, and the new test would be pinning tool names that do not exist as a contract we then have to honour.

If they come from a specific OpenCode build or plugin, point me at the source that defines them and I will re-read with that in hand. Otherwise: please drop the `agent_profiles.c` hunk and `agent_profiles_opencode_allows_tool_search_and_subagent_permissions`, leave the `client_adapter.c` change and its three assertions in `test_agent_clients.c`, retitle to the fix (`fix(opencode): pass tool args to the augment hook`), and I merge on green. Amend + force-push your own branch is fine here.

DeusData:
Merged — thank you, @Chewji9875. Passing the input args through to the augment hook, plus the subagent tool-search permissions, is a clean and well-scoped fix.

Issue #1935, opened 2026-08-30
fix(extract): Go struct fields are never extracted — 0 Field nodes for ~5000 field declarations (structs miss one level of descent)


ilyabrykau-orca:
# fix(extract): Go struct fields are never extracted — `find_class_body` stops one level above `field_declaration_list`

**Status: fix open — PR #1940** (descend `struct_type` → `field_declaration_list` before `extract_class_fields()` iterates + skip blank-identifier `_` padding; stacked on #1937). Measured: 0 → **4533** Go `Field` nodes, each with its declared type in `return_type`. **Ordering:** must land with #1937 (cross-language refs) *and* #1944 (bare-ref-vs-Field, found while validating this) under it — see below.

## What happens

`go_field_types[] = {"field_declaration"}` is registered in `lang_specs.c`; the extractor just never reaches those nodes. `find_class_body()` returns `type_spec`'s `type` child:

- **interface**: `interface_type` holds its method specs directly → always worked;
- **struct**: `struct_type`'s only named child is a `field_declaration_list`, one level above the `field_declaration`s the member loop matches → every field silently skipped. `parse_partial` flags nothing (the file parses clean).

## Damage measured (~1150-file Go repo)

| | source | graph |
|---|---|---|
| Go struct field declarations | ~5055 (4846 named, 208 embedded) | **0** |
| `Field` nodes in the whole graph | — | 850, all C/C++ |
| Go interface method declarations | 342 | present |

No `Config.Timeout`, no `Event.PID`; `trace_path` data-flow stops at the struct; nothing exists for a declared-type resolver (#1927's guard rationale) to consult; func-typed-field call dispatch has no node to reach — probed with `type Router struct{ fn func(int) error }`: `r.fn(4)` binds `Router.fn` once fields exist, and without them the call is dropped or free to be hijacked by an unrelated same-named symbol.

## Minimal reproduction

```go
package fxf

type Config struct {
	Name    string
	Timeout int
	Nested  *Config
}

type Reader interface {
	Read(p []byte) (int, error)
	Close() error
}
```

Actual: `Struct Config`, `Interface Reader`, both interface `Method`s — no node for `Name`/`Timeout`/`Nested`. After the fix: three `Field` nodes with `return_type` `string`/`int`/`*Config`; interface members unchanged.

## Detection recipe

```sql
SELECT CASE WHEN file_path LIKE '%.go' THEN 'go' ELSE 'other' END, count(*)
[...]

Issue #1932, opened 2026-08-30
Meta: Go / cgo / Go+C extraction — tracking issue


ilyabrykau-orca:
Tracking issue for the Go / cgo / Go+C extraction work. Method: count a construct in **source**, count the same thing in the **graph**, treat any mismatch as a defect. Every item was minimised to a dummy fixture before filing, measured on one real Go + C/C++ repository (~1150 Go files, eBPF C probes, a Windows C++ driver — counts only), and carries a SQL detection recipe that works on any indexed repo. Each fix PR ships a reproduce-first test (RED proven before the change) and a fresh field census on the same tree.

## Ledger — bugs

| Issue | Defect | PR | State | Measured effect |
|---|---|---|---|---|
| #1906 | Weak short-name strategies fabricate CALLS for Go selector calls | #1907 | approved on shape; direction accepted 2026-09-03 on maintainers' k8s census; rebased onto current main (memory-core), suite green | `suffix_match` 3817 → 113, `same_module` 3267 → 224; weak share of CALLS 36.0% → 12.4%, `lsp_*` 8481 → 10672 (re-resolved, not deleted) |
| #1909 | Method QNs ignore the receiver — same-name methods collapse onto one node | #1913 | direction accepted 2026-09-03 (maintainers ran their own k8s census, 150-edge hand-classification); go_lsp caller-QN join gap fixed in-PR; rebasing behind #1907; 377-loss instrument answered on-thread | `Close` methods 15 → 86; stack census: LSP strategies end above baseline (`lsp_direct` 8896 → 9130), same_module 931 → 117 |
| #1910 | Multiple `init()` per package collapse onto one QN | #1915 | ordinal suffix implemented on review feedback; rebase held per maintainer until #1907/#1913 land | `init` nodes 7 → **29**; stability pinned by test |
| #1926 | cgo's `import "C"` resolves to an arbitrary project symbol named `C` | #1931 | open, CI green; maintainer verified the 27-edge symptom is already gone on main via #1938's Strategy-3 gate — description updated to lead with the import-map effect | 27 false `IMPORTS` (pre-#1938) → 0 on current main; remaining effect: pseudo-package removed from the per-file import map |
[...]

Issue #1929, opened 2026-08-30
feat(extract): cgo is invisible — the C preamble inside a .go file is never parsed, and C.f() calls produce no edges


ilyabrykau-orca:
# feat(extract): cgo is invisible — the C preamble inside a `.go` file is never parsed, and `C.f()` calls produce no edges

**Status: draft implementation-proposal open — PR #1947** for the three bounded pieces (//export linkage as `export_linkage` contracts, `C.` pseudo-namespace veto, preamble flagged in parse coverage); marked ready on maintainer design ack. The design-sized fourth piece — parsing the preamble with the C grammar — remains open here. **Measured:** exactly 22 `export_linkage` contracts (one per directive), `C.*` callee CALLS 1 → 0, C→Go CALLS total unchanged (upgrades in place). (The adjacent `import "C"` hijack is #1926, fixed by #1931.)

## What is missing

A cgo file is one file in two languages: the comment block above `import "C"` is C, the rest is Go. cbm parses it with the Go grammar only, so:

1. **Preamble-defined C functions/types get no node** — and no `parse_partial` entry either: the Go grammar sees a comment, not an error, so coverage reports the file as fully indexed.
2. **`C.f(...)` calls produce no CALLS edge** — nothing to bind to, and no notion of a `C` pseudo-namespace.
3. **`//export Name` linkage is unread** — inbound C→Go edges appear only by short-name luck.

No occurrence of the string `cgo` exists under `src/`.

## Damage measured (~1150-file Go+C repo)

| | source | graph |
|---|---|---|
| files with `import "C"` | 27 | — |
| `C.<ident>` references / call-shaped `C.f(` | 141 / 39 | **0 CALLS edges** |
| preamble-defined C functions (one file alone) | 12 | **0 nodes** |
| `//export` directives | 22 | 22 inbound edges, all `unique_name` @ 0.38–0.75 |
| cgo files flagged in `parse_partial` | — | **0** (silent) |

The `//export` row matters most: those 22 C→Go edges are right *by accident* — name guesses standing in for a **declared ABI contract**. Any name collision moves them silently. (Pre-#1906 there was exactly 1 `C.*` CALLS edge, itself a lucky `unique_name` guess; the Go guard correctly removed it.)

## Minimal reproduction

```go
// fx/cgo_use.go
package fx

/*
static int helper(int a) { return a + 1; }
static const int kAnswer = 42;
*/
import "C"

[...]

PR #1937, opened 2026-08-30
fix(pipeline): guard USAGE/WRITES/READS against cross-language binds

### LINKED ISSUE #1928: fix(pipeline): USAGE and WRITES edges have no cross-language guard — 31% of WRITES on a Go+C repo cross the Go/C boundary
# fix(pipeline): USAGE and WRITES edges have no cross-language guard

**Status: fix open — PR #1937** (`cbm_suppress_cross_language_ref()` at all four resolver sites — sequential *and* parallel twins; the sequential-only first cut left 344 Go→C WRITES alive). Measured: Go→C/C++ WRITES 835 → **0**, USAGE 1545 → **0**, C→Go 15/90 → 0; 6432 reference edges dropped, all cross-language, none same-language; CALLS/IMPORTS byte-identical. Unblocks #1935 (with #1942 as the same-language other half).

## What happens

#725 gave **CALLS** a cross-language guard (`cbm_suppress_cross_language_suffix_match`). `USAGE`, `WRITES` and `READS` resolve through the same short-name registry — `pass_usages.c` (`resolve_usage_edges` fallback, `resolve_rw_edges`) and their parallel twins in `pass_parallel.c` (`resolve_file_usages`, `resolve_file_rw`) — and never consult it. On a Go+C tree (any eBPF/kernel component), every identifier existing in both languages produces an edge across the boundary.

## Damage measured (~1150-file Go+C repo, 15.8k nodes)

| edge type | total | Go → C/C++ | C/C++ → Go | cross share |
|---|---|---|---|---|
| `WRITES` | 2681 | 835 | 15 | **31.7%** |
| `USAGE` | 29198 | 1545 | 90 | 5.6% |
| `CALLS` | 19920 | 76 | 31 | 0.5% (guard active) |

Representative: dozens of Go test functions each declaring a local `event` produced WRITES edges to a **local variable inside an eBPF C probe**. The full dropped-edge breakdown is wider than Go↔C: Go→`.json` 3034 (config keys), Go→`.hpp`/`.h` 1612, Go→`.sh`/`.yaml` 443.

## Minimal reproduction

```go
// probe_test.go
package app

import "testing"

func TestEvent(t *testing.T) {
    event := 1
    _ = event
}
```

```c
/* probe.c */
static int handle(void) {
    int event = 0;
    return event;
}
```

Index both → a WRITES edge from the Go test into the C `event`. Expected: none — Go cannot touch a C translation unit's identifiers.

## Detection recipe

```sql
SELECT e.type, count(*) FROM edges e
JOIN nodes s ON s.id = e.source_id JOIN nodes t ON t.id = e.target_id
WHERE e.type IN ('WRITES','USAGE','CALLS') AND s.file_path LIKE '%.go'
  AND (t.file_path LIKE '%.c' OR t.file_path LIKE '%.h'
    OR t.file_path LIKE '%.cpp' OR t.file_path LIKE '%.hpp')
GROUP BY 1;  -- repeat with source/target swapped
```

## Fix shape (as landed)

One pure predicate next to the #725 guard. Two deliberate differences from CALLS: **no strategy parameter** (a reference edge carries no import-closure evidence, so every registry strategy is a bare-name guess across a boundary — the unpenalized-`unique_name` carve-out of #1572 does not apply), and **C/C++ are one family** (`.h` maps to `CBM_LANG_CPP`; a `.c` file referencing its own header is not a boundary). LSP-backed semantic references resolve before the fallback and never reach it — that path is where a genuine cgo binding (#1929) would live. JS/TS family exemption as in #725.

Related: #725, #1935/#1940 (blocked on this), #1942 (same-language other half), #787; tracked in #1932.

ilyabrykau-orca:
Fixes #1928.

## What

`USAGE`, `WRITES` and `READS` edges resolve through the same short-name registry as CALLS but never consulted the #725 cross-language guard. On a Go tree with eBPF C probes, every Go identifier spelled like a C one produced a reference edge into the C file: **31.5% of all WRITES** on the measured repo (866 of 2793) crossed the Go→C boundary — dozens of Go test locals named `event` "writing" a C probe's automatic variable. Go code cannot touch a C translation unit's identifiers; every edge in that class is false.

## How

New pure predicate `cbm_suppress_cross_language_ref(caller_lang, target_file_path)` in `registry.c`, next to (and reusing the language machinery of) `cbm_suppress_cross_language_suffix_match`, consulted on:

- the **registry-fallback** branch of `resolve_usage_edges` (USAGE), and
- `resolve_rw_edges` (WRITES/READS), which now receives the file's language.

Two deliberate differences from the CALLS guard:

- **No strategy parameter.** A reference edge carries no import-closure evidence, so every registry strategy is a bare-name guess across a boundary — the unpenalized-`unique_name` carve-out that CALLS keeps (#1572) does not apply here. LSP-backed semantic references resolve before the fallback and never reach the predicate (that path is where a genuine cgo binding would live, #1929).
- **C and C++ are one family** — `.h` maps to `CBM_LANG_CPP` in the extension table, so a `.c` file referencing its own header's declarations must not read as a boundary. (JS/TS/TSX/ArkTS stay one family, as in #725.)

## Measured expectation

From the issue: Go→C/C++ WRITES 866 → ~0, USAGE 1596 → ~0 (plus the C→Go reverse direction, 15 + 90). This also unblocks #1935 — landing ~4600 Go `Field` nodes without this guard would mint ~5700 WRITES / ~22000 USAGE name-collision edges onto `err`/`ctx`/`_` fields.

## Tests

- Reproduce-first `pipeline_go_rw_usage_never_cross_into_c` (Go + C fixture). Both paths proven RED independently on main: first `FAIL … ASSERT(!(cross_file_edge_exists(s, project, "UsesHandle", "handle", "USAGE")))`, then — with only the USAGE branch guarded — `FAIL … ASSERT(!(cross_file_edge_exists(s, project, "WriteTotal", "total_events", "WRITES")))`. GREEN with both wired. Same-language control (`Bump` → `Counter` WRITES) survives.
- Unit test `cross_language_ref_drops_go_vs_c` next to the #725 unit test: Go↔C/C++/header drops both directions, same-language keeps, C↔.h keeps, JS↔TS keeps, unknown-language/NULL/empty keeps.
- Generalized the CALLS-only test helper into `cross_file_edge_exists(…, edge_type)`; `cross_file_call_exists` is now a one-line wrapper.
- Full `scripts/test.sh` venue leg (ASan+UBSan, all suites + contract steps): green. `git clang-format --diff`: clean.

Base: upstream `main` directly (independent of the #1907/#1913/#1915 stack). #1932 tracks the family.

### DIFF
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -2654,6 +2654,12 @@ static void resolve_file_usages(resolve_ctx_t *rc, resolve_worker_state_t *ws,
                 continue;
             }
             tgt = cbm_gbuf_find_by_qn(rc->main_gbuf, res.qualified_name);
+            /* #1928: the registry fallback is a bare-name guess — never let it
+             * bind a reference across a language boundary. Mirrors the
+             * sequential twin (pass_usages.c). */
+            if (tgt && cbm_suppress_cross_language_ref(lang, tgt->file_path)) {
+                continue;
+            }
             if (usage->semantic_reference_blocked && (usage->semantic_reference_local_shadow ||
                                                       cbm_pipeline_node_is_callable_target(tgt))) {
                 continue;
@@ -2713,7 +2719,7 @@ static void resolve_file_throws(resolve_ctx_t *rc, resolve_worker_state_t *ws,
 /* Resolve reads/writes for one file. */
 static void resolve_file_rw(resolve_ctx_t *rc, resolve_worker_state_t *ws, CBMFileResult *result,
                             const char *rel, const char *module_qn, const char **imp_keys,
-                            const char **imp_vals, int imp_count) {
+                            const char **imp_vals, int imp_count, CBMLanguage lang) {
     for (int r = 0; r < result->rw.count; r++) {
         CBMReadWrite *rw = &result->rw.items[r];
         if (!rw->var_name) {
@@ -2731,6 +2737,12 @@ static void resolve_file_rw(resolve_ctx_t *rc, resolve_worker_state_t *ws, CBMFi
         }
         const cbm_gbuf_node_t *tgt = cbm_gbuf_find_by_qn(rc->main_gbuf, res.qualified_name);
         if (!tgt || src->id == tgt->id) {
+            continue;
+        }
+        /* #1928: every resolution here is a bare-name registry guess — never
+         * let it bind a read/write across a language boundary. Mirrors the
+         * sequential twin (pass_usages.c). */
+        if (cbm_suppress_cross_language_ref(lang, tgt->file_path)) {
             continue;
         }
         const char *etype = rw->is_write ? "WRITES" : "READS";
@@ -3179,7 +3191,7 @@ static void resolve_worker(int worker_id, void *ctx_ptr) {
 
         /* ── READS / WRITES ────────────────────────────────────── */
         _ph_t0 = extract_now_ns();
-        resolve_file_rw(rc, ws, result, rel, module_qn, imp_keys, imp_vals, imp_count);
+        resolve_file_rw(rc, ws, result, rel, module_qn, imp_keys, imp_vals, imp_count, lang);
         atomic_fetch_add_explicit(&rc->time_ns_rw, extract_now_ns() - _ph_t0, memory_order_relaxed);
 
         /* ── INHERITS + DECORATES + IMPLEMENTS ──────────────────── */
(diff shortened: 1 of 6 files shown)


ilyabrykau-orca:
Amended (force-push): the field census caught the sequential-only version leaving 344 Go→C WRITES / ~1200 USAGE alive on the ~1150-file repo — USAGE and WRITES/READS each have a second, independent resolver in `pass_parallel.c` (`resolve_file_usages` / `resolve_file_rw`), and large repos resolve through that path. Both parallel twins now consult the same guard, mirroring their sequential counterparts, and the test gained a parallel twin (52-filler fixture) that was RED against the sequential-only fix.

Fresh census (this branch vs main, same tree): Go→C/C++ WRITES **835 → 0**, USAGE **1545 → 0**, C→Go **15/90 → 0**; 6432 reference edges dropped in total, every one cross-language (Go→`.json` 3034, Go→`.hpp`/`.h` 1612, Go→`.c` 746, Go→`.sh`/`.yaml` 443, …) and none same-language. CALLS and IMPORTS totals byte-identical to main. Full venue leg green.

PR #1938, opened 2026-08-30
fix(pipeline): never bind a Go import to a symbol

### LINKED ISSUE #1934: fix(pipeline): a Go import can resolve to a function or method — 89 false IMPORTS edges, including os/exec → a test helper method
# fix(pipeline): a Go `import` can resolve to a function or method — the name-guess fallbacks are language-blind

**Status: fix open — PR #1938** (both name-guess fallbacks — Strategy 3 symbol-name *and* Strategy 1b sibling-file — gated on the importing file's language via `cbm_import_symbol_fallback_allowed`, false for Go). Measured: Go IMPORTS targets `Method` 85 → **0**, `Function` 4 → **0**, `Folder` 2206 unchanged. Generalises #1926/#1931 (`import "C"` was the special case).

## What happens

A Go import path names a **package** — never a function, method, field, or a Makefile target. Strategy 1 in `cbm_pipeline_resolve_import_node()` resolves every correct Go import (module path → package `Folder`). When it misses — the normal case for stdlib/third-party — the correct result is **no edge**, but two fallbacks fire instead:

- **Strategy 3** (`pass_pkgmap.c`): takes the last path segment, `cbm_gbuf_find_by_name`s it project-wide, filters by the permissive `import_targetable_label()` (admits `Function`/`Method`/…, right for Python/Java where imports can name members), returns the lexicographically smallest survivor. `import "os/exec"` → any project symbol named `exec`.
- **Strategy 1b** (sibling-file resolution) admits symbol labels through the same filter, re-creating the bug one directory closer (found by field census after gating Strategy 3 alone: one surviving `os/exec` → a *same-package* `exec()` method).

## Damage measured (~1150-file Go repo)

| Go IMPORTS resolved to | count |
|---|---|
| `Folder` (correct) | 2206 |
| `Method`/`Function` (**false**) | **89** (3.9%) |

Breakdown: `os/exec` → a test harness's `exec` method (52); `"C"` → a test helper's `C` (27, = #1926); `.../wait` → a rate limiter's method (6); two imports → a `Function` extracted **from a Makefile** (4). Blast radius is confined to the IMPORTS relation (calls in those files resolve via LSP first; only 1 knock-on call edge) — but IMPORTS is what `get_architecture` and dependency views are built from, where a Go file "depends on" a test harness or a Makefile.

## Minimal reproduction

```go
// app/run.go
package app

import "os/exec"

func Run() error { return exec.Command("true").Run() }
```

```go
// helper/harness.go
package helper

type harness struct{ n int }

func (h *harness) exec(cmd string) error { return nil }
```

+ `go.mod`. Expected: `os/exec` is external → no IMPORTS edge. Actual: edge to `helper.harness.exec`.

## Detection recipe

```sql
SELECT t.label, count(*) FROM edges e
JOIN nodes s ON s.id = e.source_id JOIN nodes t ON t.id = e.target_id
WHERE e.type = 'IMPORTS' AND s.file_path LIKE '%.go'
GROUP BY 1 ORDER BY 2 DESC;
```

For Go, any row that is not `Folder`/`Module`/`File` is false.

## Fix (as landed)

The issue's "cheaper first cut": skip both name-guess fallbacks when the importing file is Go — the measurement found no correct Go import that needed either. Member-importing languages (Python `from m import f`, Java, Kotlin, Rust `use crate::ops::helper`) keep Strategy 3; build/markup grammars (SCSS partials, Meson subdir, Pony `use`) keep Strategy 1b. Reproduce-first exact-count probe (`ei_edge_count_is` — a floor can't catch a fabricated extra edge) with cross-package and same-package decoys.

Related: #1926/#1931, #725, #787; tracked in #1932.

ilyabrykau-orca:
Fixes #1934. Generalises #1926/#1931 — `import "C"` was the special case of the same mechanism.

## What

A Go import path names a **package** — never a function, method, field, or another language's build target. Strategy 1 resolves every correct Go import (module path → the package's `Folder` node). When it misses, the import is external (stdlib / a module outside the tree) and the correct result is **no edge**. Two name-guess fallbacks violated that:

- **Strategy 3** (symbol-name fallback, `pass_pkgmap.c`) matched the path's last segment against any project definition of the same simple name and returned the lexicographically smallest survivor: `os/exec` → a test harness's `exec` Method (52 edges), `.../wait` → a rate limiter's method (6), two imports → a Function extracted **from a Makefile** (4).
- **Strategy 1b** (sibling-file resolution) admits symbol labels through `import_targetable_label`, re-creating the same bug one directory closer: with only Strategy 3 gated, the field census still bound one `os/exec` import to a **same-package** `exec()` method.

89 of 2295 Go IMPORTS edges (3.9%) on the measured repo were false; 27 of those are the `import "C"` case (#1926/#1931), the remaining 62 are this.

## How

New pure predicate `cbm_import_symbol_fallback_allowed(CBMLanguage)` (declared with the testable helpers in `pipeline_internal.h`): false for Go, true for everything else. Both fallbacks are gated on it; Strategies 1, 2 and 4 are untouched. Member-importing languages (Python `from m import f`, Java, Kotlin, Rust `use crate::ops::helper` — the case Strategy 3's own comment cites) keep the symbol fallback, and build/markup grammars (SCSS partials, Meson subdir, Pony `use`) keep the sibling one. This is deliberately the issue's "cheaper first cut": the measurement found no correct Go import that needed either fallback.

## Field-validated (fresh index of the ~1150-file Go+C repo, this branch vs main)

| Go IMPORTS target label | main | this PR |
|---|---|---|
| `Folder` (correct) | 2206 | **2206** |
| `Method` | 85 | **0** |
| `Function` | 4 | **0** |

## Tests

- Reproduce-first `ei_go_import_never_binds_symbol`: exact-count probe — the internal `util` package import must be the *entire* IMPORTS relation, with both a cross-package and a same-package `exec` decoy present. RED `[IMPORTS] FAIL count=2 expected==1` before each gate (Strategy 3, then Strategy 1b), GREEN with both.
- Adds `ei_edge_count_is`, the exact-count sibling of the floor-only `ei_edge_present` — a floor cannot catch a fabricated extra edge (#1932's negative-assertion gap).
- Full `scripts/test.sh` venue leg (ASan+UBSan, all suites + contract steps): green. `git clang-format --diff`: clean.

Base: upstream `main` directly. #1932 tracks the family.

### DIFF
--- a/src/pipeline/pass_pkgmap.c
+++ b/src/pipeline/pass_pkgmap.c
@@ -1539,6 +1539,21 @@ static bool import_targetable_label(const char *label) {
     return false;
 }
 
+/* #1934: whether the name-guess import fallbacks — Strategy 1b (sibling file,
+ * whose label filter admits symbols) and Strategy 3 (symbol name) — may run
+ * for imports from this language. A Go import path names a package — never a
+ * function, method
+ * or field — and every correct Go import resolves in Strategy 1 (module path
+ * → the package's Folder node); when that misses the import is external and
+ * the correct result is NO edge. The fallback instead bound the last path
+ * segment to an arbitrary same-named project symbol (`import "os/exec"` → a
+ * test harness's exec() method, two imports → a Makefile target). Languages
+ * whose import genuinely can name a member (Python `from m import f`, Java
+ * `import com.example.Foo`, Rust `use crate::ops::helper`) keep it. */
+bool cbm_import_symbol_fallback_allowed(CBMLanguage lang) {
+    return lang != CBM_LANG_GO;
+}
+
 static const char *path_leaf(const char *path) {
     const char *leaf = path;
     for (const char *p = path; p && *p; p++) {
@@ -1817,10 +1832,24 @@ const cbm_gbuf_node_t *cbm_pipeline_resolve_import_node(const cbm_pipeline_ctx_t
         return target;
     }
 
+    /* Name-guess fallbacks below (Strategy 1b sibling-file, Strategy 3
+     * symbol-name) are gated per importing-file language — see
+     * cbm_import_symbol_fallback_allowed (#1934). */
+    const char *src_base = source_rel ? source_rel : "";
+    for (const char *pb = src_base; *pb; pb++) {
+        if (*pb == '/' || *pb == '\\') {
+            src_base = pb + SKIP_ONE;
+        }
+    }
+    const bool symbol_fallback_allowed =
+        cbm_import_symbol_fallback_allowed(cbm_language_for_filename(src_base));
+
     /* Strategy 1b: sibling-file resolution for build/markup grammars whose
      * import string is a sibling filename or directory (SCSS partials, Just/
-     * BitBake/func includes, Meson subdir, Pony use). */
-    {
+     * BitBake/func includes, Meson subdir, Pony use). Its label filter admits
+     * symbols too, so for Go it re-creates the Strategy-3 bug one directory
+     * closer (`os/exec` → a same-package exec() method) — gated the same. */
+    if (symbol_fallback_allowed) {
         const cbm_gbuf_node_t *sib =
             resolve_sibling_file(ctx, source_rel, source_file_qn, imp->module_path);
         if (sib) {
@@ -1898,7 +1927,8 @@ const cbm_gbuf_node_t *cbm_pipeline_resolve_import_node(const cbm_pipeline_ctx_t
     /* Strategy 3: symbol-name fallback.  Derive a representative imported
      * symbol (handling alias / glob / grouped forms) and match it against an
      * in-graph definition of the same simple name in another file
-     * (Rust `helper`, Java `Util`, Kotlin grouped, ...). */
+     * (Rust `helper`, Java `Util`, Kotlin grouped, ...).
+     * Gated per importing-file language, like Strategy 1b above (#1934). */
     char symbuf[256];
     /* Prefer the clean candidate from the module path; the local_name may be an
      * alias (Rust `as h`, Kotlin `as U`) that names no real symbol. */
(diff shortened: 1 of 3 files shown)


ilyabrykau-orca:
Amended (force-push): the field census caught a second path with the same disease — Strategy 1b's sibling-file resolution admits symbol labels and bound one `os/exec` import to a *same-package* `exec()` method after Strategy 3 was gated. Both fallbacks now sit behind `cbm_import_symbol_fallback_allowed`; the test gained a same-package decoy that reproduces it (RED `count=2 expected==1`). Fresh census with the amended branch: Go IMPORTS by target label `Folder 2206 / Method 0 / Function 0` (main: `2206 / 85 / 4`). Full venue leg green.

DeusData:
Merged as `8d7d3892`.

The principle is stated correctly and the evidence matches it: a Go import path names a **package**, so when Strategy 1 misses, the import is external and the right answer is no edge — not the lexicographically smallest same-named symbol. `os/exec` binding to a test harness's `exec` method, and an import binding to a Function extracted **from a Makefile**, are the kind of edges that make a graph confidently wrong rather than merely incomplete.

Three things carried it without a round trip:

**The `Folder 2206 → 2206` row.** That is the whole argument in one line. `Method 85 → 0` and `Function 4 → 0` only mean something because the correct population is provably untouched — otherwise it is a recall trade wearing a precision costume. Putting the unchanged row in the table is what made this reviewable in minutes.

**It is per-language and you said which languages and why.** `cbm_import_symbol_fallback_allowed` is false for Go and true for everything else, with the member-importing cases (Python `from m import f`, Java, Kotlin, Rust `use crate::ops::helper`) and the build/markup grammars named explicitly rather than left to be discovered. Suppressors staying per-language is a standing rule here, arrived at after a global one measured 32.5 edges destroyed per suspicious edge removed. You landed on the right side of it and justified it.

**You found the second instance before we did.** Gating Strategy 3 alone would have looked fixed — the same-package `exec` decoy is what exposed Strategy 1b re-creating the bug one directory closer. Both decoys in one probe is the reason that hole did not ship.

`ei_edge_count_is` is the most reusable thing here. "A floor cannot catch a fabricated extra edge" is exactly the negative-assertion gap #1932 tracks, and an exact-count assertion is the right shape for every future precision fix in this family.

One thing worth telling you, since it cost this PR time and none of it was yours: our diff scanner parked it on a false positive. The rule that fires on constructed-code execution matched the literal word `exec` in your comments — the ones discussing `os/exec`. A change about imports binding to `exec` was held up for mentioning `exec`. Noted on our side.

PR #1939, opened 2026-08-30
fix(go): resolve cross-package field-chain calls via type dispatch

xkchok:
## Summary

Go calls written through an aliased cross-package struct field (`h.svc.Method(ctx, req)`) previously fell through to fuzzy suffix matching instead of type dispatch. Three defects caused it:

1. **Dispatch walk bypassed** — the per-file Go dispatch walk in `cbm_pxc_dispatch_file` was bypassed, so field selectors were never descended.
2. **Struct fields dropped** — fields extracted as flat `Field` defs were discarded by the LSP label map, so every Go struct registered with zero fields in the cross-file registry.
3. **Import-alias receiver QNs unresolvable** — type texts written through aliases produced qualified names that exist nowhere in the registry.

## Changes

- `internal/cbm/lsp/go_lsp.c` — restore the Go field-chain walk: descend `receiver.field` selectors through the shared prebuilt registry and emit `lsp_type_dispatch` / `lsp_embed_dispatch` edges; add `go_requalify_via_imports` to re-qualify alias segments through the calling file's import map when the direct receiver-type lookup misses.
- `internal/cbm/extract_defs.c` + `src/pipeline/pass_lsp_cross.c` — fold per-field `Field` defs into their owning struct inside `cbm_pxc_collect_all_defs` (one site covers both the prebuilt-registry path and the per-file fallback).
- `tests/test_parallel.c` — regression: a cross-package field chain resolves end to end with an `lsp_type_dispatch` edge from the parallel resolver.
- `tests/test_go_lsp.c` — regression: an alias-typed struct field re-qualifies through the import map and lands on the real receiver type.

Separate chore commit (drop it if unwanted): env-gated dispatch tracing (`CBM_DISPATCH_TRACE` / `CBM_TRACE_FILE`, file sink) in `go_lsp.c`, since resolve-worker stderr is swallowed by the daemon supervisor. No behavior change when unset.

## Test plan

- [x] `make test-runner` clean build
- [x] `./build/test-runner parallel go_lsp` — 125 passed, regression fixtures dispatch as expected
- [x] `./scripts/build.sh` clean production build

## Verified scope

- Synthetic two-package Go fixture: 49/49 handler→service field-chain calls resolve at 0.95 confidence via type dispatch (previously fuzzy match 0.33).
- Known gap, separate issue: the same calls regress to fuzzy match when the module is indexed nested inside an umbrella monorepo root (receiver-QN reconstruction from nested module dirs), and calls through generated client factories (service-guard structs) still yield no cross-service edges.

### DIFF
--- a/internal/cbm/lsp/go_lsp.c
+++ b/internal/cbm/lsp/go_lsp.c
@@ -866,11 +866,53 @@ const CBMType* go_eval_builtin_call(GoLSPContext* ctx, const char* name, TSNode 
 
 // --- go_lookup_field: struct field lookup with embedding recursion ---
 
+// --- Import-alias re-qualification ------------------------------------
+//
+// parse_field_defs_into_type qualifies struct field type texts as
+// "<def_module>.<text>". When the author wrote the text through an import
+// alias ("Svc:svc.Svc" in module test.main), that yields a QN
+// ("test.main.svc.Svc") that exists nowhere in the project-wide registry —
+// the real QN is "<import_qn>.Svc" and only the calling file's import map
+// can say so. On an exact-QN miss this rewrites the alias segment through
+// the file's imports; returns NULL when no alias segment is involved.
+
+static const char *go_requalify_via_imports(GoLSPContext *ctx, const char *type_qn) {
+    if (!ctx || !type_qn || !type_qn[0] || ctx->import_count <= 0) return NULL;
+    for (int j = 0; j < ctx->import_count; j++) {
+        const char *alias = ctx->import_local_names[j];
+        const char *alias_qn = ctx->import_package_qns[j];
+        if (!alias || !alias[0] || !alias_qn || strchr(alias, '.')) continue;
+        size_t alias_len = strlen(alias);
+        /* Last occurrence of a "<dot>alias<dot>" segment in type_qn. */
+        const char *hit = NULL;
+        for (const char *p = type_qn;;) {
+            p = strstr(p, ".");
+            if (!p) break;
+            p++;
+            if (strncmp(p, alias, alias_len) == 0 && p[alias_len] == '.') {
+                hit = p;
+                p += alias_len;
+            }
+        }
+        if (hit) {
+            const char *rest = hit + alias_len + 1; /* past "<alias>." */
+            return cbm_arena_sprintf(ctx->arena, "%s.%s", alias_qn, rest);
+        }
+    }
+    return NULL;
+}
+
 static const CBMType* go_lookup_field(GoLSPContext* ctx,
     const char* type_qn, const char* field_name, int depth) {
     if (!type_qn || !field_name || depth > 5) return NULL;
 
     const CBMRegisteredType* rt = cbm_registry_lookup_type(ctx->registry, type_qn);
+    if (!rt && depth == 0) {
+        /* Import-alias re-qualification: field texts from cross-package defs
+         * may embed an alias segment only this file's import map resolves. */
+        const char* alt_qn = go_requalify_via_imports(ctx, type_qn);
+        if (alt_qn) rt = cbm_registry_lookup_type(ctx->registry, alt_qn);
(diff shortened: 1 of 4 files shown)


DeusData:
Reviewed against current `main`. The core of this is good and **not** superseded — but one hunk now is, and `main` moved under you about an hour ago, so here is exactly where things stand.

## You and #1940 independently found the same extraction bug

Your defect 2 — "every Go struct registered with zero fields" — was also diagnosed and fixed in #1940, which merged as `61de19bb` while this was open. Two people arriving at the same root cause from different directions is corroboration, not duplication, and it is worth recording that you got there independently.

Concretely, `extract_class_fields` on `main` now reads:

```c
TSNode body = find_class_member_body(class_node, ctx->language);
if (ctx->language == CBM_LANG_GO) {
    body = go_normalize_struct_body(body);   /* struct_type -> field_declaration_list */
}
```

**So please drop your `internal/cbm/extract_defs.c` hunk.** It is not a conflict — git will merge it cleanly, which is the problem. By the time your block runs, `body` is already the `field_declaration_list`, so your `strcmp(ts_node_type(body), "struct_type") == 0` can never be true and the hunk lands as dead code that reads like a live second fix.

## The rest of it is still needed — I checked rather than assumed

`pxc_map_label` in `src/pipeline/pass_lsp_cross.c` on `main` today admits type-like labels plus `Protocol`, `Function`, `Method` and `Variable` — and still **not** `Field`. So your premise holds exactly as written: the flat `Field` defs #1940 now produces are dropped on the way into the cross-file registry, and `pxc_fold_go_struct_fields` is what makes them reachable. #1940 created the nodes; this makes them usable. The two are complementary, and this half does not become redundant.

The `go_lsp.c` field-chain walk and `go_requalify_via_imports` are likewise untouched by anything that landed.

## Two smaller asks

1. **Take yourself up on the offer and drop the tracing commit** (`chore(go): add env-gated dispatch tracing`). You flagged it as separable, which was the right call — a debugging facility is a change with its own lifetime, and bundling it means the review of the fix has to carry it. Happy to look at it as its own PR.
2. **Rebase onto current `main`** after dropping the hunk, so the diff shows only what is genuinely yours.

## Worth calling out

The measurement is the part that makes this reviewable: 49/49 handler→service field-chain calls at 0.95 via type dispatch against 0.33 fuzzy, on a named fixture. And you recorded the two known gaps — nested-module receiver-QN reconstruction, and generated client factories — instead of leaving them for someone to discover later. That is the right way to hand over a partial fix.

Ping when it is rebased and I will re-review promptly.

xkchok:
Rebased onto current `main` — dropped the `internal/cbm/extract_defs.c` hunk (covered by #1940) and the env-gated tracing commit. Diff is now the four files shown, all tests green (`ci-ok`). Ready for re-review.

DeusData:
Merged as `a824d82a`. You did all three, cleanly.

The `extract_defs.c` hunk is gone, the tracing commit is out, and the rebase is clean — so what landed is only what is actually yours: the `go_lsp.c` field-chain walk, `go_requalify_via_imports`, and `pxc_fold_go_struct_fields`. That last one is the piece that makes #1940's recovered `Field` nodes reachable at all, since `pxc_map_label` still does not admit `Field`; the two changes are complementary rather than overlapping, which is why dropping the duplicate cost you nothing.

Worth repeating what made this quick to re-review: the measurement. 49/49 handler→service field-chain calls resolving at 0.95 via type dispatch against 0.33 fuzzy, on a named fixture, is a claim that can be checked. And recording the two known gaps — nested-module receiver-QN reconstruction, and generated client factories — rather than leaving them to be discovered means the next person working this area starts where you stopped instead of re-deriving it.

You also arrived independently at the same root cause as #1940 for the struct-field extraction. Two people reaching it from different directions is corroboration; it is a good sign for that whole family, not wasted effort.

Thank you.

PR #1940, opened 2026-08-30
fix(extract): descend into Go struct field_declaration_list

### LINKED ISSUE #1935: fix(extract): Go struct fields are never extracted — 0 Field nodes for ~5000 field declarations (structs miss one level of descent)

ilyabrykau-orca:
Fixes #1935. **Stacked on #1937** (the USAGE/WRITES cross-language guard) — the issue's own measurement shows why this must not land alone: ~4600 recovered `Field` nodes named `err`/`ctx`/`path`/`_` would otherwise mint ~5700 `WRITES` / ~22000 `USAGE` name-collision edges through the unguarded reference resolver. The two tip commits + the test are this PR's own change; the base commit is #1937's.

## What

Go struct fields were never extracted — 0 `Field` nodes for ~5055 field declarations on the measured repo, while the 850 `Field` nodes in the graph were all C/C++. `find_class_body()` returns `type_spec`'s `type` child:

- **interface**: `interface_type` holds its method specs directly → always worked;
- **struct**: `struct_type`'s only named child is a `field_declaration_list`, one level above the `field_declaration` nodes the member loop looks for → every field silently skipped. `go_field_types[] = {"field_declaration"}` was already registered in `lang_specs.c`; the extractor just never reached them.

## How

`go_normalize_struct_body()` descends `struct_type` → `field_declaration_list` before `extract_class_fields()` iterates, the same normalization step the Java `enum_body_declarations` path already does. A second commit skips the Go blank identifier `_` (struct padding in generated code — not a referenceable field; 55 nodes that collected 241 collision edges).

## Measured effect (from the issue)

0 → **4588** Go `Field` nodes, each carrying its declared type in `return_type`. Embedded fields (~208) stay unrepresented by construction (no `name` child) — separate follow-up; `lsp_embed_dispatch` is unaffected. With #1937 under it, `CALLS` onto Go fields stay guarded (2466 → 417 by #1907's gate in the issue's stacked measurement) and the `WRITES`/`USAGE` collision classes are dropped by the reference guard. Also unblocks func-typed-field call dispatch (`r.fn(4)` → `Router.fn`), which needs the field node to exist.

## Tests

- Reproduce-first `extract_go_struct_fields_have_nodes`: RED with the descend fix reverted — `count_defs_with_label(r, "Field") == 0, expected 3` — GREEN on the branch. Asserts the three named fields with declared types (`string`, `int`, `*Config`), the blank identifier's absence, and that interface members keep extracting exactly as before.
- Full `scripts/test.sh` venue leg (ASan+UBSan, all suites + contract steps): green on the stack. `git clang-format --diff`: clean.

#1932 tracks the family; ordering constraint (#1928 first) is from the issue's own two-index measurement.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -6962,6 +6962,13 @@ static void extract_class_fields(CBMExtractCtx *ctx, TSNode class_node, const ch
             continue;
         }
 
+        /* Go: `_` is the blank identifier, used for explicit struct padding in
+         * generated code. It is not a referenceable field, and emitting it gives
+         * every `_` in the repository a same-named node to collide with. */
+        if (ctx->language == CBM_LANG_GO && strcmp(name, "_") == 0) {
+            continue;
+        }
+
         const char *field_qn = cbm_arena_sprintf(a, "%s.%s", class_qn, name);
 
         CBMDefinition def;
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -4780,6 +4780,20 @@ static TSNode find_class_member_body(TSNode class_node, CBMLanguage lang) {
     return ts_node_is_null(declarations) ? body : declarations;
 }
 
+/* Go structs keep their field_declaration nodes one level below the body that
+ * find_class_body() returns: type_spec's `type` child is a struct_type whose
+ * only named child is a field_declaration_list. Interfaces need no such step --
+ * interface_type holds its method specs directly, which is why interface members
+ * extracted correctly while every struct field was silently skipped. Normalize
+ * here, the same way the Java enum_body_declarations step above does. */
+static TSNode go_normalize_struct_body(TSNode body) {
+    if (ts_node_is_null(body) || strcmp(ts_node_type(body), "struct_type") != 0) {
+        return body;
+    }
+    TSNode list = cbm_find_child_by_kind(body, "field_declaration_list");
+    return ts_node_is_null(list) ? body : list;
+}
+
 // Dart: resolve method name from method_signature/function_signature.
 static TSNode resolve_dart_method_name(TSNode child, const char *ck) {
     if (strcmp(ck, "method_signature") == 0) {
@@ -6656,6 +6670,9 @@ static void extract_class_fields(CBMExtractCtx *ctx, TSNode class_node, const ch
     }
 
     TSNode body = find_class_member_body(class_node, ctx->language);
+    if (ctx->language == CBM_LANG_GO) {
+        body = go_normalize_struct_body(body);
+    }
     if (ts_node_is_null(body)) {
         return;
     }
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -2654,6 +2654,12 @@ static void resolve_file_usages(resolve_ctx_t *rc, resolve_worker_state_t *ws,
                 continue;
             }
             tgt = cbm_gbuf_find_by_qn(rc->main_gbuf, res.qualified_name);
+            /* #1928: the registry fallback is a bare-name guess — never let it
+             * bind a reference across a language boundary. Mirrors the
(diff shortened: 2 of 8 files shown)


ilyabrykau-orca:
Rebased onto the amended #1937 (its guard now covers the parallel resolvers). No semantic changes to this PR's commits; extraction+pipeline suites green, full venue leg green on the stack.

Heads-up from field validation: with #1937 under it, this fix still surfaces ~26k same-language bare-name reference edges onto the new Field nodes (a local `err := …` binding whichever field is named `err`). That is its own bounded defect — filed as #1942 with counts, fix PR incoming stacked on this branch. Suggested merge order: #1937 → this → #1942's PR (or the last two together).

Issue #963, opened 2026-07-08
Best-effort indexing-coverage signal: mark files/regions that were not fully indexed


DeusData:
## Best-effort indexing-coverage signal

**Framing (read first):** this is a **best-effort** attempt to mark code that was **not fully indexed** into the graph. A flag means *"we know this file/region is incomplete — prefer grep here."* The **absence** of a flag is **NOT** a guarantee of completeness — it means no gap was *detected*, not that none exists. We never present this as a completeness proof.

### Why

Today the graph can silently drop constructs and nothing surfaces it. Concrete case: `cbm_path_within_root` (an `#ifdef`-split-brace function) was entirely absent from the graph while its callers stayed linked (#960, #961). An agent searching the graph has no way to know it might be reading an incomplete picture and should fall back to grep.

### What we can detect (best-effort, in decreasing confidence)

| Class | Example | Signal |
|---|---|---|
| Purposely ignored | gitignore / excluded dirs / `.cbmignore` / vendored | already in the index response `excluded` — surface as `not_indexed_by_design` (a *design* note, kept separate from gaps) |
| File-level skip | unreadable / oversized / parse-timeout / worker crash | already in `skipped[]` (#785) |
| **Parse-partial** | `#ifdef`-split brace, truncated file, unparseable syntax | **NEW — `ts_node_has_error(root)` + ERROR/MISSING line ranges.** Verified: tree-sitter reports `has_error=TRUE` for exactly this case. |
| **Extractor gap** | tree-sitter parsed a `function_definition`, extractor didn't emit a node | **NEW — per-file reconciliation** of tree-sitter recognized-construct count vs. emitted node count (analog of the existing `dump_verify`). Per-language, only where calibrated to zero false positives. |
| Clean-but-wrong misparse | valid tree, mis-categorized def, count matches | **Undetectable** — the honest residual; part of why this is best-effort. |

### Deliverables

- **Signal A (first increment):** capture `has_error` + ERROR line-ranges per file → mark the `File` node (a *present-when-incomplete* marker, so absence never implies "complete") → surface in the `index_repository` response as `parse_partial[]`. Reproduce-first test with the `#ifdef` fixture (RED before, GREEN after). This alone would have caught `cbm_path_within_root`.
- **Signal B (follow-up):** parse-vs-extract count reconciliation, per calibrated language.
[...]

PR #1941, opened 2026-08-30
fix(coverage): narrow parse-error ranges, and stop the report hiding what it dropped (#963)

### LINKED ISSUE #1984: index_status reports a whole-file parse-error range (src/cli/cli.c as 1-13047), which is not actionable
### Version

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

`index_status` reports a parse-error range covering an entire file when only a small part
of it failed to parse. On this repository it reported `src/cli/cli.c` as one error range
of `1-13047` — the whole file. The file indexed fine; the report was wrong.

A range that names the whole file is not actionable. "Grep 13,000 lines for the problem"
is the same as no signal at all, and it makes the coverage report look far worse than the
indexing actually is.

I expected the reported range to name the region that actually failed, so it points at
something a person can open and read.

### Reproduction

1. Index this repository itself:
   `codebase-memory-mcp cli index_repository '{"repo_path":"<path to this repo>"}'`
2. Call `check_index_coverage` (or `index_status`) and read the reported error ranges.
3. Result: `src/cli/cli.c` is reported as a single error range `1-13047`, the full file.
   Expected: a range naming the region that failed to parse.

This is the parent defect of the coverage-report work tracked in #963, which is closed.
The two narrower faults — a duplicate range, and a range ending one line past the end of
the file — are already tracked separately as #1965 and #1966. This issue is the remaining
one: the range itself is far too wide.

### Logs

```text
index_status: src/cli/cli.c  error range 1-13047
```

```text
Not applicable — a reporting-accuracy issue, not memory or performance.
```

Reproduces on this repository (roughly 649 hand-written C/H files).

### Confirmations

CaptainMittens:
## What this fixes

`index_status` reported `src/cli/cli.c` as one error range of `1-13047` — the
whole file. The file was indexed fine; the report was wrong, and "grep 13,000
lines" is not useful advice.

The cause is conditional compilation splitting a brace. Tree-sitter parses raw
text with no preprocessor, so this shape gives it two `if` headers and one
closing brace:

```c
#ifndef _WIN32
if (...) {
#else
if (...) {
#endif
```

At file scope it cannot recover, the root node becomes an error node, and the
old code blamed the whole file. `cli.c` has three such blocks. This is not
specific to us — a platform guard around `lstat`/`stat` is a common shape in C.

The fix reads the SECOND parse. The pipeline already preprocesses any file
containing `#ifndef`, and that expanded parse is clean. The report simply never
consulted it.

## Measured result

| file | before | after | share of file |
|:--|:--|:--|--:|
| `src/cli/cli.c` | `1-13047`, one range | 85 ranges, 1819 lines | 100% → **13.9%** |
| `tests/test_cli.c` | 6821 lines to EOF | 70 ranges, 441 lines | 48.6% → **3.1%** |
| `src/cli/activation_transaction.c` | 907 lines to EOF | 11 ranges, 175 lines | 38% → **7.5%** |
| whole repo | 78 flagged files | 58 flagged files | — |

What survives is honest. The three biggest ranges left in `cli.c` are real
discarded platform blocks that are genuinely absent from the graph on this
platform.

## Two silent caps, both fixed

Narrower ranges mean more of them, which pushed real files against limits that
used to be unreachable. There were two, in series, and both dropped ranges with
no signal at all:

| cap | where | was | now |
|:--|:--|--:|--:|
| `CBM_MAX_ERROR_REGIONS` | `internal/cbm/cbm.c` | 64 | 256 |
| `COVERAGE_RANGE_MAX` | `src/mcp/mcp.c` | 128 | 256 |

Both were measured binding at 64 — `cli.c` and `tests/test_cli.c` sat on
exactly 64 ranges each. That means every coverage number this project has ever
reported for those files was a floor, not a measurement.

A raised cap is still a cap, so a range string can now end with `,+<N>` naming
how many ranges were dropped, and `check_index_coverage` reports
`"ranges_truncated": true` — both when it sees that marker and when its own
limit stops the list.

## A third coverage kind: `parse_unusable`

A file whose single range still covers 80 percent or more of it now reports
`parse_unusable` instead of listing the range. The file WAS indexed; pointing a
reader at nearly every line just tells them nothing, so the report says "read
the source directly".

Its main customer is not C. The narrowing step only runs for C, C++ and CUDA,
so a Python, Java or Ruby file whose root node fails still reports one
whole-file range. Verified against real broken files in those languages.

It is deliberately not called `parse_failed`. "Failed" reads as a skip phase,
and a reader who thinks a file was skipped believes it is absent from the graph
entirely. Two places in the report already fell through to `"skipped"` for want
of an explicit branch; both now name the kind.

## This is the first of three PRs

The `scripts/ci/self-index-coverage-gate.sh` commit that used to sit here is now
#1968, at review request. It adds a required check to `pr-smoke`, which is a
maintainer decision rather than part of a parse-coverage fix.

A third PR sits between them, because two of the findings below are fixed rather
than only filed, and #1968's allowlist text quotes a figure that is only correct
once they are:

| PR | Carries | Merge |
|:--|:--|:--|
| **this one** | the parse-coverage product work | first |
| #1971 | the duplicate range and the past-EOF line (#1965, #1966) | second |
| #1968 | the CI gate | third |

Cross-fork PRs cannot base on a fork branch, so #1971 and #1968 each show this
PR's commits in their diff until this one merges. Both descriptions name which
commits are their own.

Nothing else moved. The five commits here touch no CI or `scripts/ci/` file at
all, and neither does the maintainer's merge of `main` on top of them.

## Rebased onto `main`, and one perf fix

Rebased onto `fe85a6b2`, and the maintainer has since merged current `main`
into the branch on top of that (`ddb6533c`), so the head is a merge commit
rather than the fifth commit. Three files conflicted in the rebase, and each is
resolved the way review asked:

- `src/mcp/mcp.c` — the three tool descriptions take `main`'s lean text
  verbatim, so this branch no longer touches that surface at all.
  `coverage_add_ranges` keeps `main`'s `static bool` shape and
  `return emitted > 0`, keeps this branch's `+<N>` marker branch and its
  reader-cap check, and emits ONE flag, `ranges_truncated`, whenever either
  fired. The flag sits outside the `emitted > 0` block, so a row that lost
  every range still says so. `coverage_add_row_json` keeps `main`'s
  `ranges_added` / `diagnostics_full` detail elision, with the kind test
  widened to `parse_partial || parse_unusable`.
- `tests/test_mcp.c` — both tests kept. The truncation-marker test now passes
  `"format":"json"` on its three calls, because `main`'s default is the lean
  tree format, and it asserts `ranges_truncated`.
- `tests/test_parse_coverage.c` — purely additive: this branch's Phase 2 tests
  and `main`'s three #1746 tests, both `RUN_TEST` groups.

The fifth commit is the perf fix. `cbm_refine_regions_with_pp_lines` asks
`cbm_line_is_toplevel_macro_call` about every preprocessor-parsed line, and
that check walked the source from byte 0 to find the line — bytes times lines,
on exactly the shape this refinement exists to narrow. It now builds one
line-offset table per file and hands the byte span to the check.

Measured on one C file whose file-scope brace is split by an `#ifndef`, padded
to 5k / 10k / 20k / 40k lines. Release build, one machine, seconds:

| lines | `main` | branch | with the fix |
|--:|--:|--:|--:|
| 5,000 | 0.98 | 1.14 | 0.98 |
| 10,000 | 3.30 | 4.02 | 3.41 |
| 20,000 | 11.64 | 14.77 | 12.56 |
| 40,000 | 44.04 | 58.58 | 46.90 |

The branch ran 33% over `main` at 40k lines and now runs 6.5% over, with the
same answer: `8-8`, where `main` reports the whole file as `5-40000`.

This fixture is heavier than the one used in review — it holds 40k
definitions, so `main`'s own per-definition cost dominates and the ratio here
looks smaller than the 2.8x measured there. A profile of the 40k run no longer
reaches the macro check. What is left is `cbm_line_contains` and
`cbm_span_contains_callable_def`, which `main` carries too and this branch does
not touch.

## One field renamed after review

Each `parse_unusable` entry carries one number, and it was called `lines`. That
reads as the length of the file, and the two are not the same number — a
grammar can end an error node past the last line, which this repo has already
met (`scripts/setup-windows.ps1` has 326 lines and its range ends at 327).
`lines` also already means a definition's line span everywhere else in the same
response.

The field is now `range_end`, at both places that emit it, with the caveat
written in a comment beside each. One test, proved RED first against the old
name. Deriving the real file length instead is not available here: neither
`cbm_file_error_t` nor `cbm_coverage_row_t` carries it.

## Tests

Every test was checked RED before it was kept, either by breaking the code
under it or by measuring first.

- The range narrows to the dropped branch; explained lines are excluded; a
  range never starts or ends on a directive; real garbage beside a split brace
  is still flagged; a clean file stays unflagged.
- The cap reports what it dropped, and an under-cap file carries no marker.
- The three coverage classes are told apart, including a non-C whole-file case.
- The Studio Export range join puts ONE marker at the end with the summed
  count — a marker left mid-string makes every reader stop there and silently
  lose the ranges after it.
- `check_index_coverage` emits every range in front of a marker, never turns
  the marker's digits into a range, and reports `ranges_truncated` from both
  caps.
- `tests/test_index_resilience.c` gained a ceiling beside its floor: exactly
  one of two fixture files is flagged and the clean neighbour is absent.
- The three `_Thread_local` grammar forms are pinned as measured, so a
  tree-sitter bump that changes them turns a test red instead of leaving a
  stale note behind.

Full suite, re-measured on the current head (`ddb6533c`) on 2026-09-06:
**7936 passed, 1 failed, 7 skipped** across 141 suites. The one failure is
`cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index`
at `tests/test_cli.c:2075`, unchanged in name and line since the earlier run. A clean worktree at `main` fails the same test at the
same line with the identical 305 passed / 1 failed, and this branch touches
neither `tests/test_cli.c` nor `src/cli/` at all. The 28 agent-client failures
this section used to name are gone — `main` fixed them.

## Found on the way, filed not fixed

All now have issues, per review:

| Finding | Issue |
|:--|:--|
| `scripts/setup-windows.ps1` names line 113 twice — two ERROR nodes on one line each push their own range | #1965, fixed in #1971 |
| The same file's range ends at line 327 of a 326-line file — a node ending at column 0 still gets +1 | #1966, fixed in #1971 |
| `cbm_count_lines` and the inline `orig_lines` counter disagree by one on a trailing newline | #1967 |
| `scripts/smoke-invariants.sh` asserts 14 tools when the server registers 16, and never checks the other direction | #1969 |

Related to #963.

Fixes #1984


DeusData:
`dco` is the only red check here — everything else is green — and since you will be amending for it anyway, there is one small extra thing to fold into the same pass.

**1. Sign-off.** None of the four commits carry a `Signed-off-by:` line, so all four need one. `git rebase --signoff main` (or an interactive reword) will do it in one go, then force-push with lease.

**2. While you are in there: please drop the `Claude-Session:` lines.** They point at a private session URL that nobody reading the history can open, so it is noise in the permanent record. `Co-Authored-By:` is fine and can stay — that is honest attribution and we are not asking you to remove it. This is a convention that was not written down anywhere, which is on us, not on you.

On the change itself: the commit messages here are better than most of what lands in this repo, and specifically for reasons worth naming.

You did not just state that the range was wrong — you found the mechanism (three `#ifndef _WIN32` blocks splitting a brace, so the raw parse cannot resync at file scope and the root becomes `ERROR`), and then noticed the pipeline already had the answer sitting unused in the preprocessed parse.

The three rules are all documented as *found by running it*, which is the part that usually goes missing:

- the blank-line rule, where treating a dropped branch's blank line as proof suppressed every C range in the suite;
- the directive-line rule, with its **cost stated explicitly** — a `#define` the raw parse really dropped no longer surfaces on its own;
- the top-level macro rule, tied to the recovery walker's deliberate refusal in #949.

Recording a known cost rather than quietly absorbing it is the right instinct, and it is what makes the rule reviewable instead of something a future reader has to re-derive.

Also noted: the ordering rationale for the three coverage steps is argued from *where each one's evidence lives*, not from what happened to work — and you fixed 13 pre-existing clang-format violations from earlier commits on the branch while confirming the reflowed tool descriptions concatenate byte-identically, so no output moved.

Get the sign-off on and this goes into review properly. Thanks for a strong first contribution.

DeusData:
One more ask, which I missed in my last comment and should have flagged first time — apologies for the two-part review.

**Please split `fccf8215` (the CI gate) into its own PR.**

That commit adds a new **blocking** step to `pr-smoke` in `.github/workflows/pr.yml`, a new gate script (`scripts/ci/self-index-coverage-gate.sh`), an exemption allowlist, and a hardcoded baseline ceiling. Those are maintainer-owned policy here: they add a required check every future PR must clear, and that decision belongs to the project owner rather than to a code review. It is not a judgement on the gate — building it with four checks each verified to fail independently is the right way to build one — but it should not ride in on a parse-coverage fix.

The good news is your commit boundary is already exactly on that line: `fccf8215` touches **only** `pr.yml` and the three `scripts/ci/` files, and the other three commits touch **zero** gate surfaces. So it should lift out with a clean `rebase -i` drop and no rework.

That leaves this PR as the ~1200 lines of product work, which is what I actually want to review.

So the full list, all mechanical:

1. `Signed-off-by` on all four commits (blocking — the only red check).
2. Drop the `Claude-Session:` lines. `Co-Authored-By:` is fine and stays.
3. Split `fccf8215` out into its own PR.

Two non-blocking notes while you are in there, take or leave:

- `add_parse_unusable_summary` labels `atoi(dash+1)` as `lines`, i.e. the file length — but that is the range *end*, and your own found-on-the-way note shows a range end can exceed the file length (245–327 on a 326-line file). Worth renaming or deriving properly in a change whose whole thesis is honest reporting.
- `orig_lines` in `cbm_extract_file_ex` counts every `\n` plus one, while `cbm_count_lines` skips a trailing newline. They feed different consumers so nothing breaks today, but one helper would be easier to keep honest.

And please do file your two found-not-fixed items as issues — the duplicated `113-113` range especially.

CaptainMittens:
All three done, plus both notes. No apology needed for the two-part review — the split was the right call and it lifted out cleanly.

**1. Signed-off-by.** All four commits carry it. `dco` is green.

**2. `Claude-Session:` lines.** Gone from every commit message, and from the PR description too. `Co-Authored-By:` stays.

**3. The gate is now #1968.** You were right that the boundary was already on the line — `git cherry-pick` of the other three onto `main` applied with no conflict, and the only tree difference against the old head is the four gate files. #1968 stacks on this PR and should merge after it, because checks 1 and 2 read the `parse_unusable` kind and the `+<N>` marker that this PR adds. Its description says so.

One thing you should know that was not visible before the split: a follow-up already exists that fixes **three ways the gate could report PASS without having looked at everything** — reading the ceiling could abort the whole gate under `pipefail`, the report's own `truncated` flag was never read, and check 1 named files it had already accepted. It is not in #1968 because its allowlist text depends on a range fix in a product PR I have not opened yet. It follows as its own PR. Given the gate is a required check, you may prefer to hold #1968 until that one is beside it — your call.

---

**Note 1, taken.** You are right, and it was worse than a mislabel: `lines` already means a definition's line span everywhere else in the same response, so one word carried two meanings in one document. The field is now `range_end` at both emit sites, with the caveat written in a comment beside each. Deriving the real file length was the other option and is not open here — neither `cbm_file_error_t` nor `cbm_coverage_row_t` carries it, only the path and the range string.

One test, proved RED first (`range_end is NULL` against the old name). It reads the expected end line from the persisted coverage row rather than a constant, so it states the property and not a measurement, and it asserts the old name is gone rather than kept beside the new one. `index_resilience`, `parse_coverage` and `mcp`: 283 passed, 4 skipped.

**Note 2, filed rather than changed:** #1967. You have it exactly — `cbm_count_lines` ignores a trailing newline, the inline `orig_lines` counter does not, and they disagree by one on nearly every file. Nothing breaks today because one sizes an array (generous is safe) and the other decides the 80% threshold (generous is harmless). I left it out of this PR because picking one convention is a real change with two defensible answers, not a rename, and this PR is already large. The issue says which answer needs writing down.

---

**Issues filed**, and I found two more while writing them up:

| Finding | Issue |
|:--|:--|
| The duplicated `113-113` range | #1965 |
| The range ending at 327 on a 326-line file | #1966 |
| The two line counters that disagree (your note 2) | #1967 |
| `smoke-invariants.sh` asserts 14 tools when the server registers 16 — and only checks one direction, so it prints `all 14 tools present` while covering 14 of 16 | #1969 |

That last one is the same shape as everything else in this PR: a check reporting success without having looked at everything.

This PR is now the ~1200 lines of product work plus the `range_end` rename. Ready for review.

CaptainMittens:
Thanks — all four points are addressed. Taking them in order.

## Sign-off (blocking)

Done. All four commits now carry `Signed-off-by`, and `dco` is green.

The cause was simple: this repo ships `scripts/install-git-hooks.sh`, and I had never run it. I have added a `prepare-commit-msg` hook locally instead, so the trailer goes on automatically from here. A commit template would not have worked — `git commit -m` and `git commit -F` ignore the template, and that is how most of these commits get written.

## The `Claude-Session:` lines

Gone from every commit. `Co-Authored-By:` stayed, as you said it could.

## Splitting the CI gate

Done — the gate is now #1968. You were right that it does not belong here: it adds a required check to `pr-smoke`, and that is your call to make, not a detail of a parse-coverage fix.

One thing changed while splitting it. A third PR now sits between the two, so the order is:

| PR | Carries | Merge |
|:--|:--|:--|
| **#1941** (this one) | the parse-coverage product work | first |
| #1971 | the duplicate range and the past-EOF line | second |
| #1968 | the CI gate | third |

#1971 exists because #1968's allowlist comment quotes `25.2%` for `scripts/setup-windows.ps1`, and that figure is only correct once the range fix lands. Without a middle PR the gate would either carry a wrong number or carry a product fix — which is the thing you asked me not to do.

Cross-fork PRs cannot base on a fork branch, so #1971 and #1968 each show this PR's commits until this one merges. Both descriptions name which commits are their own.

## Note 1 — `lines` is really the range end

Taken. You are right, and it is worse than a bad name: `lines` already means a definition's line span in six other places in the same response. One word, two meanings, one document.

The field is now `range_end` at both places that emit it, with the caveat written beside each:

```c
/* The end of the range, not the length of the file. A grammar can end
 * an error node past the last line, so this number can be larger than
 * the file. See range_end_is_not_file_length. */
```

I could not give you the real file length instead, which would have been the better fix. Neither `cbm_file_error_t` nor `cbm_coverage_row_t` carries it, so there is nothing to read. Renaming was the honest option available.

One test covers it, proved red first against the old name. It reads the expected end line from the persisted coverage row rather than hard-coding a number, and asserts the old `lines` key is absent.

## Note 2 — `orig_lines` versus `cbm_count_lines`

Confirmed and filed as #1967, not fixed here. Two counters over the same buffer disagree by one on a trailing newline. It is a real inconsistency but it is not what this PR is about, and the numbers it would move are pinned by tests in the commits above.

## The two found-not-fixed items

Both filed, and both fixed:

| Finding | Issue | Fix |
|:--|:--|:--|
| line 113 named twice — two ERROR nodes on one line each pushed a range | #1965 | #1971 |
| the range ends at line 327 of a 326-line file | #1966 | #1971 |

The duplicate you singled out was the more interesting of the two. Both copies count against `CBM_MAX_ERROR_REGIONS`, so a file with many multi-error lines could be clipped while holding fewer distinct lines than the cap allows. The merge in #1971 runs before the cap check, so a merged region is never miscounted as a dropped one.

The real file now reads `113-113,245-326`.

Two more turned up on the way and are filed rather than fixed: #1969 (`smoke-invariants.sh` asserts 14 tools when the server registers 16, and never checks the other direction) and #1972 (the gate's ceiling drifts against a moving `main`, so it reddens PRs that did not cause it — worth your opinion before #1968 merges).

## CI

Green on this PR: `dco`, `lint`, and all three `pr-smoke` legs. #1971 and #1968 are still running.

CaptainMittens:
Rebased onto `a824d82a`. The conflict came from `17b5a432` ("fix(perl): update grammar for format declarations"), which added two fixtures and two tests to `tests/test_parse_coverage.c` in the same place this branch adds its own. Both sides are purely additive, so both sets of tests are kept — nothing was dropped.

Rebasing surfaced a real regression in the branch stacked above this one (#1971), which I have fixed there rather than here. The short version: that commit merged overlapping error ranges into one, and the merge handed a wider range's covering definition to an error the definition does not explain. `perl_malformed_source_remains_partial_issue1838` — one of the tests `17b5a432` added — caught it. #1971 now drops only an exact repeat of the range already open, which still fixes the `113-113,113-113` duplicate it was written for.

Full suite at the top of the stack: `parse_coverage`, `extraction`, `pipeline`, `mcp` and `index_resilience` all pass.

CaptainMittens:
`test / test-lsan-macos` is red on this branch. I believe it is an unrelated flaky test, and I cannot re-run the job myself — I have no write access here. A re-run from a maintainer would confirm it.

The evidence, in case it saves you the dig:

**What failed** — one test out of 7789:

```
FAIL tests/test_watcher.c:1479: cbm_watcher_poll_once(w) == 0, expected 1
watcher   rc=1  pass=71  fail=1
7789 passed, 1 failed, 7 skipped
```

`watcher_identical_watch_preserves_dirty_baseline` appends a file, touches the watcher, re-registers the same path, then expects `cbm_watcher_poll_once` to report one change. It reported none.

**The leak is downstream of that assertion, not a separate problem.** The test opens a store at line 1459 and fails at line 1479, so the close never runs:

```
SUMMARY: AddressSanitizer: 232937 byte(s) leaked in 375 allocation(s)
  #12 store_open_internal                                   store.c:823
  #13 test_watcher_identical_watch_preserves_dirty_baseline  test_watcher.c:1459
```

Those 375 allocations are one abandoned sqlite connection.

**Why I do not think this branch causes it:**

1. This branch touches neither `tests/test_watcher.c` nor `src/watcher/watcher.c`.
2. Its only change to shared code is a comment. `src/store/store.c` shows 24 changed lines; every one is inside a doc block describing `parse_partial`, `parse_unusable` and the skip phases. No executable change.
3. `test-lsan-macos` passes on #1971, which contains every commit on this branch plus one more.

**Why the test can lose a race.** The assertion depends on the file write and the poll landing in different filesystem timestamp ticks. On a loaded macOS runner they can land in the same tick, and the watcher then correctly reports no change. If that reading is right, the fix belongs in the test rather than in any product code, and I am happy to open a separate issue for it — I have not, because one red run is not yet evidence of a pattern.

DeusData:
Thank you for the careful failure attribution. #1993 has now repaired `main`, so the inherited compile failure is gone. We will handle the requested rerun in the PR review queue rather than asking you to change this branch or file a watcher issue from a single run. Nothing further is needed from you at this stage, and thank you for tracing the leak to the failed assertion rather than treating it as a separate defect.

CaptainMittens:
The `test-lsan-macos` red here is not this PR, and I do not think a rebase clears it either. Details in case they save you a look.

The whole run is one failing assert out of 7789:

```
watcher rc=1 pass=71 fail=1 skip=0 secs=40
FAIL tests/test_watcher.c:1479: cbm_watcher_poll_once(w) == 0, expected 1 == 1
```

That is `watcher_identical_watch_preserves_dirty_baseline`, and line 1479 is the second poll — the one that must see the worktree go dirty after `th_append_file`. This PR touches no watcher code at all: the diff is coverage, pipeline, store and mcp.

The part you may not have visible: your own fix for this family is `32e910aa` ("fix(watcher): FILETIME-precision dirty signatures; flaky-family rebuilds"). It exists only on `feat/atlas-r1`. It is not an ancestor of `main`, and `main` has had no change under `src/watcher/` or `tests/test_watcher.c` since 2026-08-25. So rebasing this branch does not pick it up.

One caveat on my own reading, so you can discount it if you disagree. The headline half of `32e910aa` is second-granularity `st_mtime`, and that half should not apply on macOS — `main` already takes the nanosecond path at `src/watcher/watcher.c:586`:

```c
return ((int64_t)st->st_mtimespec.tv_sec * NS_PER_SEC) + (int64_t)st->st_mtimespec.tv_nsec;
```

The half that looks relevant is the other one in the same commit: the bounded wait-for-state helper replacing single fixed polls on the reach-assertions. A lone `poll_once` that expects detection to have already landed is racy on an 18-job runner, which matches a single-assert failure in an otherwise clean suite.

So from here this needs a re-run rather than anything from me. Happy to rebase if you would rather force a fresh run anyway — just say the word.

CaptainMittens:
Following up on my note above about `32e910aa` being stranded on `feat/atlas-r1` — I checked the rest of that branch, and it is not alone. Two more fixes there apply to `main` and have not reached it.

I went through all 15 `fix(` commits on `feat/atlas-r1` (51 commits ahead of `main`, tip `bda46086`, last touched 2026-08-28) and checked each against what `main` actually contains rather than against file paths. Three apply; twelve are correctly branch-local.

**`77eb7eb3` — fix(mcp): tell the truth about pre-publication index aborts.** Neither `aborted_previous_preserved` nor `persist_failed` appears anywhere under `main`'s `src/`, and both are in `atlas-r1`'s `src/mcp/mcp.c`. So on `main` today an aborted incremental run still answers with the generic `error — check repo_path exists`, sending people to debug a path that was fine, when the previous index is intact and a retry is the right move. That looks like a live user-facing defect on `main` rather than branch work. Happy to open an issue for it if you would like it tracked separately.

**`136685f1` — fix(test-infra): make container legs work from git worktrees.** `main`'s `test-infrastructure/run.sh` has no `CBM_GIT_COMMON_DIR`; `atlas-r1`'s has four references. Without it, `/src/.git` in a worktree checkout stays a pointer file into a directory outside the `..:/src` mount, so git inside the container fails and the git-based contracts die with "matched no files". This one I can confirm from the outside: every one of my checkouts is a worktree.

For completeness, the twelve I ruled out, so you do not have to re-check them:

- Nine are `fix(ui)`, serving the branch's own UI work.
- `409dce4d` sizes mkdtemp templates in the `cbm_hist` and `cbm_why` helpers. Those exist only on `atlas-r1` — `main`'s `tests/test_ui.c` has zero matches for either, and already uses the safe `snprintf(tmpdir, sizeof(tmpdir), ...)` form throughout.
- `0b736b2f` and `7f2c5632` adapt tests to a real `ingest_traces` contract. `main` still ships the discard-stub: `traces_received` is present on `main` and absent on `atlas-r1`, `pairs_unmatched` the reverse. Porting either would break `main`.

All three of the stranded fixes touch files `main` has barely moved since the 2026-08-21 split, so they should still port cleanly. That is less likely to hold as `main` keeps moving — it is 239 commits ahead of the branch already.

Nothing here needs anything from me. Flagging it only because `32e910aa` turned out to be a symptom rather than the whole thing.

DeusData:
Both of these are exactly the kind of note that saves a maintainer an hour — thank you.

**On the `test-lsan-macos` red:** agreed on every point. `watcher_identical_watch_preserves_dirty_baseline` at `test_watcher.c:1479` is a known flaky family here, this PR touches no watcher code, and your reading that the relevant half of `32e910aa` is the bounded wait-for-state helper (not the mtime precision, which macOS already has via `st_mtimespec`) matches mine.

One thing has changed since your note that affects the timing of a rebase. The same leg picked up a **second**, unrelated failure mode today: Homebrew's `llvm` formula moved to 23.1.0, and under Clang 23 the test-runner build dies at `src/cli/cli.c:1768` (`-Wunused-but-set-global`) before any test runs — whether a runner gets the 22 or 23 bottle is a draw. The fix is #2027, in flight now. So: yes to the rebase, but **please hold it until #2027 is on `main`** so one rebase picks up both the merged base and the compiler fix, rather than spending a run on a red we already understand.

**On the `feat/atlas-r1` audit:** this is careful work, and the method — checking each fix against what `main` actually contains rather than against paths — is the right one. Porting stranded fixes from that branch is a direction decision I want to make deliberately rather than in a PR thread; I have it queued with your list as the input. Yes, please open an issue for `77eb7eb3` — an aborted incremental run answering with the generic `check repo_path exists` while the previous index is intact is a live user-facing defect on `main` as you describe, and it deserves its own tracking regardless of how the branch question lands.

DeusData:
Thank you for the `feat/atlas-r1` audit — that took real effort, and it is appreciated.

One thing to know so it does not look ignored: `atlas-r1` is a draft branch we are actively working on right now, so those commits are moving targets rather than stranded work. They will land with the branch when it is ready, so no need to open separate issues for them; if anything from that audit still applies once atlas lands, we will pick it up then.

The plan for this PR stays as agreed: hold the rebase until #2027 is in, then one rebase clears both the LSan red and the conflict.

DeusData:
Review done on a tree trial-merged with current main, and the substance is solid: the coverage marker, `parse_unusable` and directive-line work all bind (six tests go red under main's `cbm.c`, two under main's `mcp.c`), the LSan red is gone on a rebased tree (699 pass incl. `watcher`, no leak), lint is clean, DCO is on all five commits, and the gate split-out means nothing here touches the enforcement surface any more. Two things before it merges:

1. **Rebase onto main** — three files conflict; this is the resolution I verified end to end:
   - `src/mcp/mcp.c`: (a) the three tool-description hunks (`index_repository` / `query_graph` / `index_status`): take main's lean text verbatim. (b) `coverage_add_ranges`: keep main's `static bool` shape and `return emitted > 0`, keep your `+<N>` marker branch and the reader-cap check, and emit one flag, `ranges_truncated` (main's name from #1597, still unreleased), whenever either fired — outside the `emitted > 0` block. Please use main's name rather than `truncated`: a row-level `truncated` collides with the lean contract's response-level key. (c) `coverage_add_row_json`: keep main's `ranges_added` / `diagnostics_full` detail elision and widen the kind test to `parse_partial || parse_unusable`.
   - `tests/test_mcp.c`: keep both tests; in `tool_check_index_coverage_reports_truncation_marker_issue963` add `"format":"json"` to its three calls (main's default is now the lean tree format — red without it, green with it) and assert `"ranges_truncated":true`.
   - `tests/test_parse_coverage.c`: purely additive — keep your Phase-2 tests and main's three #1746 tests, both `RUN_TEST` groups.

2. **One perf fix, measured and attributed**: `cbm_refine_regions_with_pp_lines` (`cbm.c:1388`) calls `cbm_line_is_toplevel_macro_call` once per PP-parsed line, and `cbm_span_is_macro_invocation` (`:1253`) rescans from byte 0 to locate that line — O(bytes × lines) on exactly the whole-file-ERROR C shape this PR targets. One C file with a file-scope `#ifndef`-split brace at 5k / 10k / 20k / 40k lines: main 0.14 / 0.35 / 1.04 / 2.99 s, this branch 0.27 / 0.78 / 2.37 / 8.23 s (2.8× at 40k and widening). With that one call neutralised the branch returns to 3.00 s with identical output (`8-8`, versus main's `4-39999`). Build a line-offset table once per file and hand the span to the macro check. Main's own ~3× per doubling on this shape is pre-existing and recorded separately — not yours to fix here.

Then #1971 follows straight after (its `cbm.c` hunk applies at an offset and its test hunk re-places after the #1746 tests; nothing else). Same sign-offs, and this merges on green. Thank you for splitting the gate out and for reproduce-first tests that fail for the right reasons — this is the right shape for coverage work.

CaptainMittens:
Both items are done, and #1971 is rebased on top.

## Rebase

Onto `fe85a6b2`, with your resolution exactly.

`src/mcp/mcp.c` now has no diff hunk anywhere in the `TOOLS` array — the three
descriptions are main's text byte for byte. `coverage_add_ranges` keeps main's
`static bool` and `return emitted > 0`, keeps the `+<N>` marker branch and the
reader-cap check, and emits one `ranges_truncated` outside the `emitted > 0`
block, so a row that lost every range still reports it. `coverage_add_row_json`
keeps main's `ranges_added` / `diagnostics_full` elision, with the kind test
widened. `tests/test_mcp.c` keeps both tests, with `"format":"json"` on the
three calls and `ranges_truncated` asserted. `tests/test_parse_coverage.c`
keeps both `RUN_TEST` groups.

Your note about the name collision earned its keep, so this is on the record:
my first pass renamed the flag with a plain search and replace, and it also
caught three of main's own `"truncated": true` assertions — the response-level
key. Reverted. Only the new test uses the row-level name.

The rebase also left one statement in `coverage_add_row_json` past the column
limit. It carries clang-format's wrap now, folded into the commit that already
touches that file.

## Perf fix

Fifth commit. `cbm_refine_regions_with_pp_lines` builds one line-offset table
per file and hands the byte span to the macro check. The line-based form stays
for the two callers that ask about a whole region, where a single walk beats a
table. If the table allocation ever fails, the old walk still runs, so the
answer cannot change.

Same fixture shape as yours — one C file whose file-scope brace is split by an
`#ifndef`, padded to 5k / 10k / 20k / 40k lines. Release build, one machine,
seconds:

| lines | `main` | branch | with the fix |
|--:|--:|--:|--:|
| 5,000 | 0.98 | 1.14 | 0.98 |
| 10,000 | 3.30 | 4.02 | 3.41 |
| 20,000 | 11.64 | 14.77 | 12.56 |
| 40,000 | 44.04 | 58.58 | 46.90 |

33% over main at 40k lines before, 6.5% after, and the same answer: `8-8`,
against main's `5-40000`.

My fixture is heavier than the one you measured — it pads with 40k function
definitions, so main's own per-definition cost dominates and the ratio reads
1.3x here where yours read 2.8x. The removed call is the same one; the baseline
underneath it is simply larger. A profile of the 40k run no longer reaches the
macro check at all.

## One finding from the sweep, not fixed here

`cbm_line_contains` (`internal/cbm/cbm.c:1064`) has the same shape: it walks
from byte 0 to reach a 1-based line, and `cbm_extract_file_ex` asks it once per
recovered definition. It is the top frame in that same profile. It is on main,
this branch does not touch it, and you have already said main's growth on this
shape is recorded separately — so I left it. Happy to open an issue for it if
you want it tracked.

## Tests

`parse_coverage`, `extraction`, `pipeline`, `mcp` and `index_resilience`: 964
passed, 4 skipped.

Full suite: 7932 passed, 1 failed, 7 skipped across 141 suites. The one failure
is `cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index` at
`tests/test_cli.c:2075`. A clean worktree at main fails the same test at the
same line with the identical 305 passed / 1 failed, and this branch touches
neither `tests/test_cli.c` nor `src/cli/` at all. The 28 agent-client failures
I reported earlier are gone — main fixed them.

## #1971

Rebased with no conflicts. The `cbm.c` hunk applied at an offset, the test hunk
re-placed after the #1746 tests, nothing else — exactly as you said. The change
is byte-identical to before the rebase.

Every commit on both branches carries a sign-off and no session line. The
description here is updated to match the final diff.

CaptainMittens:
CI is green on this branch: 34 of 34 checks pass and the PR reads `MERGEABLE / CLEAN` at `b04f5450`.

That covers the three platform smoke builds, ASan+UBSan (`test-diag`), LeakSanitizer (`test-lsan-macos`), MemorySanitizer (`test-msan`), ThreadSanitizer on macOS, x86 Linux and ARM Linux, the eight gcc `test-unix` shards, both Windows CLANG64 shards, and the `ci-ok` gate.

Ready when you are.

One note on #1971 before it follows. `test / test-windows (windows-latest, CLANG64, x86_64, 2/2)` failed there. The same shard passes here with `b04f5450` in the tree, so the perf commit is not the cause. That run is still going and its log is not published yet, so I have not read the assert. I will post the diagnosis on the #1971 thread once the log opens, and rebase onto main after this one lands.

DeusData:
Merged as aa44c28e, thank you for the careful rounds on this one. The rebase resolution matched the ask to the letter (checked with a range-diff against the pre-rebase chain), the perf fix brings the 40k-line case from 7.9 s back to main's 3 s with identical output, and the coverage report now says what it dropped instead of hiding it. #1971 is next: main now contains these commits, so please rebase that branch onto main (a plain merge of main into it conflicts); it should reduce to your one fix and merges on green.

Issue #1927, opened 2026-08-30
fix(pipeline): field_type_hint binds by receiver variable-name substring, not declared type — 1484 edges at conf 0.85 on one Go repo


ilyabrykau-orca:
# fix(pipeline): `field_type_hint` binds by receiver variable-name substring, not declared type

**Status: fix open — PR #1936** (owning-segment equality, `fth_owner_segment_is`; stacked on #1907→#1913→#1915). Measured: `field_type_hint` 1484 → **356**, single-character receivers 376 → 0, Go→C/C++ 34 → 0, nothing relocated to other strategies. Other strategies before/after: `suffix_match` 113 → 114, `unique_name` 1573 → 1573, `same_module` 224 → 224, all `lsp_*` 10672 → 10672; CALLS total 14316 → 13189 — the 1128 removed edges were fabricated on top of already-failed resolutions, so nothing absorbed them.

## What the code does

`try_field_type_hint()` (`src/pipeline/pass_parallel.c`, parallel resolver only): split `callee_name` at the first `.` → receiver identifier text; strip `_`/`m_`, uppercase the first letter → `type_name`; also build `iface_name = "I"+type_name`; then accept the first same-named candidate where `strstr(candidate_qn, type_name)` or `strstr(candidate_qn, iface_name)` hits, overwriting the resolution at confidence 0.85 as `field_type_hint`.

That is a raw substring test on the QN — no declared type, field type, or LSP type is ever read. For a single-letter receiver (`t`, `f`, `w`, `c`, `m`) `type_name` is one capital letter and matches almost any QN, the method name included (`"F"` ⊂ `"…FileStore.Close"`).

The Go weak-match guard (#1906/#1907) deliberately keeps `field_type_hint` on the stated premise that it is receiver-aware for Go. The implementation does not meet that premise; the pinning test (`lrp_go_s8_field_type_hint`, `repo`/`Repo`) passes for the wrong reason (its GREEN path is `lsp_cross`).

## Damage measured (~1150-file Go+C repo)

| | v0.10.8 | with #1906/#1909/#1913/#1915 fixes |
|---|---|---|
| CALLS total | 19653 | 14316 |
| `field_type_hint` | 493 (2.5%) | **1484 (10.4%)** |
| single-character receiver | — | **376** |

[...]

PR #1944, opened 2026-08-30
fix(pipeline): a bare Go reference never binds a struct Field

### LINKED ISSUE #1942: fix(pipeline): a bare Go reference can bind a struct Field — READS/WRITES/USAGE attach every local err to whichever Field is named err
# fix(pipeline): a bare Go reference can bind a struct Field — READS/WRITES/USAGE attach every local `err` to whichever Field is named `err`

**Status: fix open — PR #1944** (`cbm_go_suppress_bare_field_ref()` at the same four resolver sites as #1937's guard; stacked on #1937 → #1940). Measured: USAGE onto Go fields 21308 → **0**, WRITES 5191 → **0**; the combined stack ends below main's totals (USAGE 29198 → 21035, WRITES 2681 → 774) while adding 4533 `Field` nodes. Found while field-validating #1940 — the third leg of its ordering constraint.

## What happens

The READS/WRITES resolvers (`pass_usages.c:resolve_rw_edges`, `pass_parallel.c:resolve_file_rw`) and the USAGE registry fallback (`resolve_usage_edges` / `resolve_file_usages`) hand bare reference text to `cbm_registry_resolve()`. The registry contains `Field` nodes, so a dot-less local —

```go
err := doWork()
```

— resolves by simple name to **any** `Field` named `err`, project-wide.

In Go that binding is impossible by construction: a field is only reachable through a selector expression (`x.f`), and selector references resolve on the LSP path, never through this fallback. Unlike C#/Java/C++/Python, where a method body legitimately references its own members bare (`_count++` — `cp_reads_writes_cs_static_field` pins that as required), so the veto must be Go-gated, not global.

## Damage measured (#1940 stack, ~1150-file Go+C repo)

| onto Go `Field` nodes | count |
|---|---|
| `USAGE` | **21308** (top: a test struct's field `T` 3013, `events` 977, `err` 759, `Path` 242, `config` 221) |
| `WRITES` | **5191** |

**100%** of Field-targeted USAGE edges carried dot-less reference text — no legitimate selector-shaped population is mixed in. (The 612/468 refs onto C/C++ fields surviving #1937 are the legitimate own-member-in-method shape plus C-side noise; untouched by this fix.) The remaining field-targeted class is 2466 `CALLS` — #1906/#1907's selector-guard territory.

## Minimal reproduction (requires #1935's extraction)

```go
// state/state.go
package state

type Tracker struct {
    err error
}
```

```go
// app/app.go
package app

import "errors"

func Run() error {
    err := errors.New("x")
    return err
}
```

Expected: no edge from `Run` into `Tracker.err`. Actual (with #1940 alone): WRITES `Run` → `state.Tracker.err`.

## Detection recipe

```sql
SELECT e.type, count(*) FROM edges e
JOIN nodes t ON t.id = e.target_id
WHERE t.label = 'Field' AND t.file_path LIKE '%.go'
  AND e.type IN ('READS','WRITES','USAGE')
GROUP BY 1;
```

Any nonzero row is this bug.

## Fix (as landed)

Pure predicate next to #1937's: drop the bind when the file is Go, the target label is `Field`, and the reference has no `.`. Wired at both sequential resolvers and both parallel twins; reproduce-first probes on both paths (RED on the #1940 stack), plus unit tests (selector-shaped refs may bind; non-Field targets and non-Go languages untouched).

Related: #1927 (same "nothing checks what a name can denote" class on CALLS), #1928/#1937, #1935/#1940; tracked in #1932.

ilyabrykau-orca:
Fixes #1942. **Stacked on #1937 → #1940** — the reproduce-first probe needs Go `Field` nodes to exist (#1940) before a bare local can falsely bind one, and #1940 needs this before its node recovery is net-positive for reference edges. The tip commit is this PR's change; suggested merge order #1937 → #1940 → this (or the last two together).

## What

The READS/WRITES resolvers and the USAGE registry fallback hand bare reference text to the short-name registry, which contains `Field` nodes — so once Go struct fields exist, every Go local `err := …` binds whichever struct field is named `err`, project-wide:

| onto Go `Field` nodes (stack without this fix) | count |
|---|---|
| `USAGE` | **21308** (top target: a test struct's field `T`, 3013 edges) |
| `WRITES` | **5191** |

In Go the binding is impossible by construction: a field is only reachable through a selector expression (`x.f`), and selector references resolve on the LSP path — **100%** of the Field-targeted USAGE edges in the census carried dot-less reference text, so there is no legitimate population mixed in.

## How

`cbm_go_suppress_bare_field_ref(is_go, ref_name, target_label)` — pure predicate next to #1937's, consulted at the same four sites (both READS/WRITES resolvers, both USAGE registry fallbacks): drop when the file is Go, the target label is `Field`, and the reference has no `.`. **Go-gated**: C#/Java/C++/Python method bodies legitimately reference their own members bare — `cp_reads_writes_cs_static_field` pins that shape as required, and it stays untouched.

## Field-validated (fresh index, ~1150-file Go+C repo, full stack)

USAGE onto Go fields **21308 → 0**, WRITES **5191 → 0**. The only remaining field-targeted reference class is 2466 `CALLS`, which is #1906/#1907's selector-guard territory. Combined stack ends *below* main's reference totals (USAGE 29198 → 21035, WRITES 2681 → 774) while adding the 4533 field nodes.

## Tests

- Reproduce-first `pipeline_go_bare_ref_never_binds_field` + `_parallel` twin (both resolvers, per #1937's lesson): RED on the #1940 stack (`ASSERT(!(cross_file_edge_exists(s, project, "Run", "err", "WRITES")))` failed on both paths), GREEN with the predicate. Fixture asserts the field node exists, so the probe can't pass vacuously.
- Unit test `go_bare_ref_never_binds_field`: selector-shaped refs may bind, non-Field targets untouched, non-Go languages untouched, degenerate inputs safe.
- Full `scripts/test.sh` venue leg (ASan+UBSan, all suites + contract steps): green on the stack. `git clang-format --diff`: clean.

#1932 tracks the family. Same root class as #1927: a name-shape heuristic binding without checking what the name can denote.

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -6962,6 +6962,13 @@ static void extract_class_fields(CBMExtractCtx *ctx, TSNode class_node, const ch
             continue;
         }
 
+        /* Go: `_` is the blank identifier, used for explicit struct padding in
+         * generated code. It is not a referenceable field, and emitting it gives
+         * every `_` in the repository a same-named node to collide with. */
+        if (ctx->language == CBM_LANG_GO && strcmp(name, "_") == 0) {
+            continue;
+        }
+
         const char *field_qn = cbm_arena_sprintf(a, "%s.%s", class_qn, name);
 
         CBMDefinition def;
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -4780,6 +4780,20 @@ static TSNode find_class_member_body(TSNode class_node, CBMLanguage lang) {
     return ts_node_is_null(declarations) ? body : declarations;
 }
 
+/* Go structs keep their field_declaration nodes one level below the body that
+ * find_class_body() returns: type_spec's `type` child is a struct_type whose
+ * only named child is a field_declaration_list. Interfaces need no such step --
+ * interface_type holds its method specs directly, which is why interface members
+ * extracted correctly while every struct field was silently skipped. Normalize
+ * here, the same way the Java enum_body_declarations step above does. */
+static TSNode go_normalize_struct_body(TSNode body) {
+    if (ts_node_is_null(body) || strcmp(ts_node_type(body), "struct_type") != 0) {
+        return body;
+    }
+    TSNode list = cbm_find_child_by_kind(body, "field_declaration_list");
+    return ts_node_is_null(list) ? body : list;
+}
+
 // Dart: resolve method name from method_signature/function_signature.
 static TSNode resolve_dart_method_name(TSNode child, const char *ck) {
     if (strcmp(ck, "method_signature") == 0) {
@@ -6656,6 +6670,9 @@ static void extract_class_fields(CBMExtractCtx *ctx, TSNode class_node, const ch
     }
 
     TSNode body = find_class_member_body(class_node, ctx->language);
+    if (ctx->language == CBM_LANG_GO) {
+        body = go_normalize_struct_body(body);
+    }
     if (ts_node_is_null(body)) {
         return;
     }
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -2660,6 +2660,11 @@ static void resolve_file_usages(resolve_ctx_t *rc, resolve_worker_state_t *ws,
             if (tgt && cbm_suppress_cross_language_ref(lang, tgt->file_path)) {
                 continue;
             }
+            /* #1942: a bare Go reference can never denote a struct field. */
+            if (tgt &&
+                cbm_go_suppress_bare_field_ref(lang == CBM_LANG_GO, usage->ref_name, tgt->label)) {
+                continue;
+            }
             if (usage->semantic_reference_blocked && (usage->semantic_reference_local_shadow ||
                                                       cbm_pipeline_node_is_callable_target(tgt))) {
                 continue;
@@ -2743,6 +2748,10 @@ static void resolve_file_rw(resolve_ctx_t *rc, resolve_worker_state_t *ws, CBMFi
          * let it bind a read/write across a language boundary. Mirrors the
          * sequential twin (pass_usages.c). */
         if (cbm_suppress_cross_language_ref(lang, tgt->file_path)) {
+            continue;
+        }
+        /* #1942: a bare Go reference can never denote a struct field. */
+        if (cbm_go_suppress_bare_field_ref(lang == CBM_LANG_GO, rw->var_name, tgt->label)) {
             continue;
         }
         const char *etype = rw->is_write ? "WRITES" : "READS";
--- a/src/pipeline/pass_parallel.c
+++ b/src/pipeline/pass_parallel.c
@@ -2654,6 +2654,12 @@ static void resolve_file_usages(resolve_ctx_t *rc, resolve_worker_state_t *ws,
                 continue;
             }
             tgt = cbm_gbuf_find_by_qn(rc->main_gbuf, res.qualified_name);
+            /* #1928: the registry fallback is a bare-name guess — never let it
+             * bind a reference across a language boundary. Mirrors the
+             * sequential twin (pass_usages.c). */
+            if (tgt && cbm_suppress_cross_language_ref(lang, tgt->file_path)) {
+                continue;
+            }
             if (usage->semantic_reference_blocked && (usage->semantic_reference_local_shadow ||
                                                       cbm_pipeline_node_is_callable_target(tgt))) {
                 continue;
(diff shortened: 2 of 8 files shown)


ilyabrykau-orca:
CI retrigger (no-op amend, identical tree — no rerun rights): `test-windows-guards` died in its own harness setup — `SETUP FAIL: ASCII baseline did not index … 'nodes': None` on the guard's 2-file **TypeScript** fixture, before any guard ran. Same environmental signature #1907's first run hit and its retriggered run passed with identical code; this PR's diff is Go-gated end to end. No semantic changes.

ilyabrykau-orca:
CI retrigger (no-op amend): `test-windows-guards` again, identical `SETUP FAIL: ASCII baseline did not index … nodes: None` harness-setup signature — second consecutive hit on this PR, tracked in #1952. No semantic changes.

ilyabrykau-orca:
Fresh-day CI retrigger (no-op amend): third consecutive `test-windows-guards` harness-setup failure with the identical `SETUP FAIL … nodes: None` signature (#1952 has the matrix — this PR is now its most-hit entry). Yesterday's runner pool appears to have been the common factor; if this run hits it again I'll leave the job red and defer to maintainers on #1952 rather than keep force-pushing.

DeusData:
**#1937 and #1940 are merged** — `fcd8bfc1` and `61de19bb`. This one is queued behind its last check.

First, the thing you should not have had to do: **stop force-pushing to chase `test-windows-guards`.** You were right on all three counts. The failure is `SETUP FAIL: ASCII baseline did not index … 'nodes': None` in the guard's own harness, on a two-file **TypeScript** fixture, before any guard runs — and your diff is Go-gated end to end, so it cannot be reaching that code. #1952 is ours to fix, not something to out-wait. Three no-op amends is three more than you owed us, and you were explicit each time about what you were doing and why, which is the right way to handle it.

On the review itself — this is unusually careful work, and the specific things that made it reviewable are worth naming, because they are the things that usually go missing:

- **Both resolvers.** The sequential and parallel twins, not one. `parallel` has bitten a merged change in this repo before precisely because it was not in the "obviously related" set.
- **The probe cannot pass vacuously.** The fixture asserts the `Field` node *exists* before asserting nothing binds it. Without that, the test would have gone green if the change had simply deleted the nodes.
- **Revert-checked.** RED with the fix reverted, on both paths, quoted with the failing assertion — not "tests pass".
- **Go-gated, and you said why.** `cp_reads_writes_cs_static_field` pins the C# shape that must keep working. Call-resolution suppressors staying per-language is a standing rule here, arrived at after a global one measured **32.5 edges destroyed per suspicious edge removed**. You landed on the right side of it without being told.
- **Honest about what is left.** Embedded fields staying unrepresented, and the remaining 2466 `CALLS` being #1906/#1907 territory rather than folded in silently.

I verified the load-bearing claim in #1937 by reading the predicate rather than trusting the description: it fails **open** on unknown caller language, missing or empty path, unknown target language, same language, the JS/TS family, and the C/C++ family. It only drops when both languages are known and differ across families. That is the property that makes it a correctness fix rather than a recall trade, and it is why it could merge without a measurement argument.

The stack order you specified was also correct and load-bearing, so it was followed exactly: #1937 first, then #1940. Landing the ~4600 `Field` nodes without the reference guard underneath would have minted the collision edges your own measurement predicted.

Thank you — this is a strong first contribution to the project.

DeusData:
Post-merge correction, because a review pass after the merge found something I should have caught before it: **the `strchr(ref_name, '.') == NULL` test in `cbm_go_suppress_bare_field_ref` is inert for Go.** Filed as #1962.

Both call paths strip the receiver before the predicate ever sees the name:

- `resolve_lhs_write_name` documents its own contract — `obj.Field = ...` writes the **trailing** name, so `t.err = x` yields `err`, never `t.err`.
- `is_reference_node` never treats `selector_expression` as a reference kind; the reference is the inner `field_identifier`, whose span is bare.

So the guard is a blanket veto on Field-targeted references rather than the precise one it documents. Two consequences worth stating plainly, since they bear on evidence I accepted at merge time:

1. The unit test's `ASSERT_FALSE(..., "t.err", "Field")` pins an input Go's extractor cannot produce — a dead branch, not a production guarantee.
2. **The "100% of Field-targeted USAGE edges carried dot-less reference text" census is a tautology.** The extractor guarantees it. I quoted that figure back to you approvingly when I merged; it is not evidence of what it appeared to show, and that is on me for not checking the extractor before believing the number.

**None of this is a regression** — `main` had no Go `Field` nodes at all before #1940, so nothing that previously worked is lost. The honest description of the landed stack is "~4588 recovered nodes that reference edges cannot currently reach", not "edges removed". The work stands; the reasoning attached to one third of it does not.

The signal the guard actually wants — `field_identifier` versus a bare `identifier` — is computed at extraction time and discarded before the resolver runs, so a precise version is reachable rather than hypothetical. #1962 has the detail and a cheap falsification (index a Go repo, count Field-targeted USAGE/READS/WRITES; the prediction is zero).

No action needed from you unless you want to take #1962 — and no criticism intended in any of it. You disclosed the collision class this stack creates before anyone asked, which is why the follow-up was findable at all.

ilyabrykau-orca:
Thank you for the merge review — and especially for the post-merge correction. You are right on both counts: the extractor strips the receiver on every path that reaches the resolver, so the dot test cannot fire, and the census I presented was measuring the extractor's contract back at itself. I should have traced the ref_name provenance before claiming the 100% figure as evidence.

Taking #1962: the plan is exactly the signal you point at — record is-member-access at extraction (the field_identifier half of a selector) on CBMUsage/CBMReadWrite, and let the guard consume that instead of string shape, so genuine `t.err = x` / `cfg.Timeout` references bind Field nodes while bare locals stay suppressed. Reproduce-first test + a field census on a real Go repo (predicted: Field-targeted USAGE/WRITES go from the current 0 to the genuine selector population) will be in the PR.

Issue #1911, opened 2026-08-29
Go: build-tag twin files collide on one QN — the stub can shadow the real implementation (Rust cfg got #495)


ilyabrykau-orca:
**Status: draft implementation-proposal open — PR #1946** (marked ready on maintainer design ack). Minimal scope as proposed here: per-file τ folded into func/method QNs, #495 `#`-suffix style; τ-aware resolution is the follow-up. Sole-constrained-variant callers keep their edges; twin-pair callers from unconstrained files resolve as ambiguous (honest) instead of binding the lexicographically-smallest stub. **Measured:** 2177 τ-suffixed func/method nodes across 372 constrained files; swallowed-twin files 319 → 281; CALLS total **+62** (recovered targets outweigh fail-closed ambiguity). Receiver qualification (#1913) does **not** separate twins — generated-twin evidence below.

### Version

codebase-memory-mcp 0.10.8 (also reproduced on a source build @ d706c33ac2f9)

### Platform

macOS (Apple Silicon)

### Install channel

Built from source

### Binary variant

standard

### What happened, and what did you expect?

Go build-constrained twin files (`//go:build unix` / `//go:build !unix`, GOOS/GOARCH filename suffixes) legally define the **same symbols** in one package. The indexer ignores the constraints; the upsert keeps one node per QN by the deterministic smallest-file-path rule. Measured:

- `hostmirror_other.go` (2-line stub) beat `hostmirror_unix.go` (63-line real implementation) because `o` < `u`: `get_code_snippet` returns the stub and **all 25 inbound CALLS landed on the stub**;
- same for a `nofollow_unix.go`/`nofollow_other.go` constant pair;
- one generated GOARCH twin pair loses 39 of 40 nodes — 100% of the losing file;
- the dropped twin leaves **no trace**: both files parse clean, so `index_status`/`check_index_coverage` report nothing.

Expected: both build variants exist as nodes — exactly what #495 already does for Rust `#[cfg(...)]` twins.

### Reproduction

```go
// mirror_unix.go
//go:build unix

package mirror

func MirrorConfig(path string) (string, error) {
	dst := path + ".bak"
	// ... real work ...
	return dst, nil
}
```

```go
// mirror_other.go
//go:build !unix

package mirror

func MirrorConfig(path string) (string, error) { return "", nil }
```

Actual: one node, `file_path = mirror_other.go` (stub wins); the real implementation has no node. Expected: two nodes, one per constraint.

[...]

PR #1946, opened 2026-08-30
fix(extract): give Go build-constrained twin files distinct QNs

ilyabrykau-orca:
Part of #1911 — **draft implementation-proposal**: this changes what gets extracted (Go func/method QNs in build-constrained files), so per CONTRIBUTING it's opened as a draft for design review; the design is the one proposed in the issue body (the #495 Rust-cfg `#`-suffix pattern). Will mark ready on a maintainer design ack.

## What

Build-constrained twin files (`//go:build` lines, GOOS/GOARCH filename suffixes) legally define the same symbols; the upsert kept ONE node per QN by the smallest-file-path rule — a 2-line stub beat the 63-line implementation and took all 25 inbound CALLS; a bpf2go GOARCH twin pair lost 39/40 nodes; coverage stayed silent.

## How

- τ = compacted `//go:build` expression, else the official GOOS/GOARCH filename suffix (`cbm_go_build_tau`, computed once per file). Folded into **func/method** QNs (`MirrorConfig#unix`), mirrored on the call-scope side (`go_tau_scope_qn`) so body-call attribution stays exact. Types/vars stay plain — parent_class/`DEFINES_METHOD` joins keep working.
- Two `#`-aware seams make suffixed defs resolvable (both also close a latent gap for Rust cfg twins, whose QNs already carry `#`):
  - `cbm_registry_add` indexes the simple name as the part before `#` (identifiers cannot contain one);
  - the LSP registry's exact `pkg.name` lookup falls back to a **sole** `#`-suffixed variant (`lookup_func_sole_tau_variant`) — the dominant real-world shape is a constrained file with **no** in-tree twin, and its callers keep their `lsp_direct`/`lsp_strategy_cross_file` edges. Two+ variants are genuinely ambiguous without the caller's build configuration and fail closed. The pipeline's LSP-join leaf gate learns the same rule.

## Scope (deliberately the issue's "minimal PR")

QN disambiguation only. τ-aware resolution (preferring the caller's own constraint, and `module.name#τ` same-module keys) is the declared follow-up — unconstrained callers of a genuine twin *pair* currently resolve as ambiguous rather than binding an arbitrary twin, which is strictly better than today's bind-everyone-to-the-stub.

## Tests (reproduce-first)

- `extract_go_buildtag_tau_in_func_qns`: `//go:build linux && amd64` → `#linux&&amd64`, `_windows_amd64.go` → `#windows_amd64`, `_linux_test.go` → `#linux`, types stay plain, unconstrained files carry no `#`. RED on main.
- `pipeline_go_buildtag_twins_both_survive`: RED on main (`count_nodes_named == 1, expected 2`); GREEN with both twins as nodes AND a sole-variant caller edge staying alive (`UseFlush → FlushDisk#linux`, `lsp_strategy_cross_file` @ 0.92).
- 854 green across extraction/pipeline/registry/lsp probes/go_lsp/parallel; full `scripts/test.sh` venue leg green; `git clang-format --diff` clean.

## Field census

Landed (same-day baseline, repo @ f555e5ce): **2177** τ-suffixed func/method nodes across **372** constrained files; swallowed-twin files 319 → **281**; `Close` methods 15 → 27; CALLS total **+62** — recovered twin targets outweigh the fail-closed twin-pair ambiguity. Full table in the census comment below.

#1932 tracks the family.

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -1311,6 +1311,10 @@ CBMFileResult *cbm_extract_file_ex(const char *source, int source_len, CBMLangua
         .root = root,
         .macro_table = macro_table,
         .return_type_table = return_type_table,
+        /* #1911: per-file Go build constraint, folded into func/method QNs by
+         * the def and scope builders so build-tag twin files stop colliding. */
+        .go_build_tau =
+            language == CBM_LANG_GO ? cbm_go_build_tau(a, source, source_len, rel_path) : NULL,
     };
 
     // Run extractors: defs + imports use separate walks (unique recursion patterns),
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -599,7 +599,20 @@ typedef struct {
      * class-body variable def records which class declares it (parent_class)
      * without changing its module-level qualified name. NULL elsewhere. */
     const char *var_parent_class;
+    /* #1911: the Go file's build constraint τ (compacted //go:build expression,
+     * else the GOOS/GOARCH filename suffix), or NULL for an unconstrained
+     * file. Folded into func/method QNs by extract_defs.c and mirrored by the
+     * scope builder in extract_unified.c so build-tag twin files stop
+     * colliding in the graph upsert. Set once in cbm_extract_file. */
+    const char *go_build_tau;
 } CBMExtractCtx;
+
+/* #1911: resolve a Go file's build constraint τ — the compacted //go:build
+ * expression when present (constraint lines precede the package clause), else
+ * the official GOOS/GOARCH filename suffix (name_GOOS.go, name_GOARCH.go,
+ * name_GOOS_GOARCH.go, each optionally followed by _test) — or NULL for an
+ * unconstrained file. Defined in helpers.c. */
+const char *cbm_go_build_tau(CBMArena *a, const char *source, int source_len, const char *rel_path);
 
 // --- Public API ---
 
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -3778,6 +3778,16 @@ static void extract_func_def(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec 
         def.is_test = rust_def_is_test(def.decorators);
     }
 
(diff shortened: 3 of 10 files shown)


ilyabrykau-orca:
CI retrigger (no-op amend): `test-lsan-macos` failed in `test_cli` installer/skill probes and `test_daemon_ipc` (no LeakSanitizer report anywhere in the log) — none of which this Go-extraction diff touches, and the same job is green on every sibling PR in today's waves (#1907 #1913 #1937 #1938 #1944 #1947 #1949). Same runner-environment day as #1952. No semantic changes.

ilyabrykau-orca:
Field census landed (isolated `CBM_RUNTIME_DIR` unblocked the one-shots; same-day baseline vs this branch on the ~1150-file Go+C repo, repo @ f555e5ce):

| | main | #1946 |
|---|---|---|
| τ-suffixed func/method nodes | 0 | **2177** across **372** build-constrained files |
| files whose only node is their own `__file__` (swallowed twins) | 319 | **281** (−38) |
| nodes total | 15858 | 16071 (+213) |
| `Close`-named method nodes | 15 | 27 |
| CALLS total | 19823 | **19885 (+62)** |

The call graph *grew* while twins were being disambiguated — recovered twin targets more than offset the twin-pair-ambiguity fail-closed cases, so the minimal-scope worry (losing edges from unconstrained callers) did not materialize at repo scale.

DeusData:
Thank you for the measured build-constraint twin analysis and for keeping constraint-aware resolution outside this initial draft. Qualified-name changes affect extraction, joins, and compatibility across the graph, so we need more time to review the design and failure modes carefully. Please keep it in draft; we will come back with a grounded decision as soon as capacity allows.

ilyabrykau-orca:
Rebased onto current main. Conflicts were test insertion-anchor drift in tests/test_extraction.c and tests/test_pipeline.c (re-anchored, plus the count_nodes_named forward declaration my new test now needs above its insertion point); all source files auto-merged, including registry.c around the merged #1944 changes. Full scripts/test.sh green on this head (isolated CBM_RUNTIME_DIR — the earlier local run collided with the live 0.10.8 session daemon, same class as #1952's venue findings).

PR #1971, opened 2026-08-31
fix(coverage): stop reporting a duplicate range and a line past EOF (#963)

### LINKED ISSUE #1965: fix(coverage): a line with two ERROR nodes reports the same range twice
`scripts/setup-windows.ps1` has 326 lines. Its parse-coverage report reads:

```
113-113,113-113,245-327
```

Line 113 is named twice. Found while building the CI coverage gate in #1941; not fixed there.

## What happens

Line 113 carries two separate ERROR nodes, at columns 25-29 and 31-32. `cbm_error_regions_push` pushes a range for each one. A line range is advice — "read these lines" — and saying it a second time adds nothing.

## Why it matters beyond the noise

Both copies count against `CBM_MAX_ERROR_REGIONS`. A file with many multi-error lines can therefore be clipped, and report `,+<N>`, while carrying fewer distinct lines than the cap allows.

## Suggested fix

Merge a region that overlaps the range already open instead of appending a repeat. The walk visits children in source order, so a region that starts at or before the open range's end really does overlap it. The merge has to run **before** the cap check, or a merged region is miscounted as a dropped one.

Related: #963, #1941. Sibling finding: the past-EOF end line in the same range string.

### LINKED ISSUE #1966: fix(coverage): a parse-error range can end one line past the end of the file
`scripts/setup-windows.ps1` has 326 lines. Its parse-coverage report ends the last range at line **327**, which does not exist:

```
113-113,113-113,245-327
```

Found while building the CI coverage gate in #1941; not fixed there.

## What happens

The tree-sitter node for that region is `start=(244,2) end=(326,0)`. An end column of 0 means the node stopped right after the previous line's newline, so it holds no text on the row it points at. `cbm_error_regions_push` adds 1 to that row to convert to 1-based, which names one line past the end of the file whenever a region runs to EOF.

## Why it matters

A reader told to open line 327 of a 326-line file has been given advice that cannot be followed, in a report whose whole job is to say honestly which lines are missing. It also inflates the range's share of the file, which the coverage gate added in #1941 measures against a 25% threshold.

## Suggested fix

Clamp the end to the row above when the end column is 0 and the node spans more than one row.

Related: #963, #1941. Sibling finding: the duplicated `113-113` range in the same string.

CaptainMittens:
Two faults in one range string, both in `cbm_error_regions_push`. `scripts/setup-windows.ps1` has 326 lines and its parse-coverage report read:

```
113-113,113-113,245-327
```

The same line named twice, and an end line that does not exist. Found while building the coverage work in #1941 and filed rather than fixed there, at review request.

## The past-EOF end line

The tree-sitter node is `start=(244,2) end=(326,0)`. An end column of 0 means the node stopped right after the previous line's newline, so it holds no text on the row it points at. Adding 1 to that row to make it 1-based named a line past the end of the file whenever a region ran to EOF.

Fixed by clamping the end to the row above when the end column is 0 and the node spans more than one row.

## The duplicate range

Line 113 carries two separate ERROR nodes, at columns 25-29 and 31-32, and each pushed its own range. A line range is advice — "read these lines" — and it says nothing new the second time. Both copies also count against `CBM_MAX_ERROR_REGIONS`, so a file with many multi-error lines could be clipped while holding fewer distinct lines than the cap allows.

Fixed by dropping a range that **exactly repeats** the one already open — same start line and same end line. Ranges that merely overlap are left alone, and that distinction is the whole point.

Each range is judged later by `cbm_region_is_recovered`, which asks whether definitions starting inside the range cover it. Two ranges with identical numbers always get the same verdict, so dropping one changes nothing. Two different ranges do not: merging `3-3` into `2-3` hands the wider range's covering definition to an error that definition does not explain, and a real parse failure then disappears from a report whose only job is to be honest about failures. `perl_malformed_source_remains_partial_issue1838` pins that case.

The drop runs **before** the cap check, so a repeat that was never a distinct range is never counted as one the cap threw away.

The real file now reports `113-113,245-326`.

## Tests

Two, both proved RED first with the exact expected text:

| Test | Red output |
|:--|:--|
| `coverage_repeated_error_line_reports_one_range_issue963` | `"2-2,2-2"` != `"2-2"` |
| `coverage_range_never_ends_past_the_last_line_issue963` | `"1-5"` != `"1-4"` |

`parse_coverage`, `index_resilience` and `mcp`: 285 passed, 4 skipped. `make -f Makefile.cbm lint-format` clean.

## Stack order

This is the middle of three. #1941 merged as `aa44c28`, and this branch is now rebased onto it, so the diff here is exactly one commit — `e5c17de6`, `fix(coverage): stop reporting a duplicate range and a line past EOF`, touching only `internal/cbm/cbm.c` and `tests/test_parse_coverage.c`. #1968 stacks on this one and should merge after it.

| PR | Carries |
|:--|:--|
| ~~#1941~~ | the parse-coverage product work — merged as `aa44c28` |
| **this one** | the two range faults above |
| #1968 | the CI gate |

#1968 depends on this PR, not only on #1941: its allowlist entry quotes the post-fix figure of 25.2% for `setup-windows.ps1`, which is only true once the range above is corrected.

Closes #1965. Closes #1966. Part of #963.

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -794,12 +794,47 @@ typedef struct {
 } cbm_error_regions_t;
 
 static void cbm_error_regions_push(cbm_error_regions_t *acc, TSNode n) {
+    TSPoint start = ts_node_start_point(n);
+    TSPoint end = ts_node_end_point(n);
+    uint32_t start_line = start.row + 1;
+    uint32_t end_line = end.row + 1;
+
+    /* A node that ends at column 0 stopped right after the previous line's
+     * newline, so it holds no text on the row it points at. Counting that row
+     * named a line past the end of the file whenever the region ran to EOF:
+     * scripts/setup-windows.ps1 has 326 lines and reported "245-327". */
+    if (end.column == 0 && end.row > start.row) {
+        end_line = end.row;
+    }
+
+    /* One line can carry several error nodes, and repeating the same line range
+     * says nothing new. Line 113 of scripts/setup-windows.ps1 has two error
+     * nodes, at columns 25-29 and 31-32, and the report read "113-113,113-113".
+     * Drop the repeat.
+     *
+     * Only an EXACT repeat of the range already open is dropped. Do not merge
+     * ranges that merely overlap. Each range is judged separately later by
+     * cbm_region_is_recovered, which asks whether definitions starting inside
+     * that range cover it. Two ranges with the same numbers always get the same
+     * verdict, so collapsing them changes nothing. Two DIFFERENT ranges do not:
+     * merging 3-3 into 2-3 hands the wider range's covering definition to an
+     * error the definition does not explain, and a real parse failure then
+     * disappears from the report. tests/test_parse_coverage.c pins that case in
+     * perl_malformed_source_remains_partial_issue1838.
+     *
+     * This runs BEFORE the cap check, so a dropped repeat never counts as a
+     * range the cap threw away. */
+    if (acc->count > 0 && start_line == acc->starts[acc->count - 1] &&
+        end_line == acc->ends[acc->count - 1]) {
+        return;
+    }
+
     if (acc->count >= CBM_MAX_ERROR_REGIONS) {
         acc->dropped++;
         return;
     }
-    acc->starts[acc->count] = ts_node_start_point(n).row + 1;
-    acc->ends[acc->count] = ts_node_end_point(n).row + 1;
+    acc->starts[acc->count] = start_line;
+    acc->ends[acc->count] = end_line;
     acc->count++;
 }
 
--- a/tests/test_parse_coverage.c
+++ b/tests/test_parse_coverage.c
@@ -834,6 +834,51 @@ TEST(c_thread_local_grammar_limit_is_pinned_issue963) {
     PASS();
 }
 
+/* Two error nodes can sit on ONE line. Line 113 of scripts/setup-windows.ps1
+ * does exactly that, and the report used to read "113-113,113-113" — the same
+ * line named twice. A line range says nothing new the second time, so repeated
+ * or overlapping regions must collapse into one. */
+static const char *PS_TWO_ERRORS_ONE_LINE = "Write-Host \"start\"\n"             /* 1 */
+                                            "wsl.exe -- bash -c $Command 2>&1\n" /* 2 */
+                                            "Write-Host \"end\"\n";              /* 3 */
+
+/* An error region that runs to the end of the file stops just after the last
+ * newline. Tree-sitter calls that position row N, column 0 — a row that holds
+ * no text. Reading it as a line number named a line past the end of the file:
+ * scripts/setup-windows.ps1 has 326 lines and the report said "245-327". */
+static const char *PS_ERROR_TO_EOF = "} else {\n"             /* 1 */
+                                     "    if ($a) {\n"        /* 2 */
+                                     "        Write-Host x\n" /* 3 */
+                                     "}\n";                   /* 4 */
+
+TEST(coverage_repeated_error_line_reports_one_range_issue963) {
+    CBMFileResult *r =
+        cbm_extract_file(PS_TWO_ERRORS_ONE_LINE, (int)strlen(PS_TWO_ERRORS_ONE_LINE),
+                         CBM_LANG_POWERSHELL, "covproj", "two_errors.ps1", 0, NULL, NULL);
+    ASSERT_NOT_NULL(r);
+    ASSERT_TRUE(r->parse_incomplete);
+    ASSERT_NOT_NULL(r->error_ranges);
+    /* Line 2 carries two separate error nodes. It must be named once. */
+    ASSERT_STR_EQ(r->error_ranges, "2-2");
+    ASSERT_EQ(r->error_region_count, 1);
+    cbm_free_result(r);
+    PASS();
+}
+
+TEST(coverage_range_never_ends_past_the_last_line_issue963) {
+    int len = (int)strlen(PS_ERROR_TO_EOF);
+    CBMFileResult *r = cbm_extract_file(PS_ERROR_TO_EOF, len, CBM_LANG_POWERSHELL, "covproj",
+                                        "error_to_eof.ps1", 0, NULL, NULL);
+    ASSERT_NOT_NULL(r);
+    ASSERT_TRUE(r->parse_incomplete);
+    ASSERT_NOT_NULL(r->error_ranges);
+    /* The file has four lines and ends with a newline. Line 5 does not exist. */
+    ASSERT_STR_EQ(r->error_ranges, "1-4");
+    ASSERT_NULL(strstr(r->error_ranges, "5"));
+    cbm_free_result(r);
+    PASS();
+}
+
 SUITE(parse_coverage) {
     RUN_TEST(c_ifdef_split_brace_sets_parse_incomplete);
     RUN_TEST(c_ifdef_split_brace_neighbors_still_extracted);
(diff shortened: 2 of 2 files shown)


DeusData:
Reviewed. Both fixes are right, and the reasoning in the code is better than I expected going in.

**The past-EOF clamp.** `end.column == 0 && end.row > start.row` is the correct guard — the second condition is what stops a single-line node collapsing to line 0, and it would be easy to omit. "An end column of 0 means the node stopped right after the previous line's newline, so it holds no text on the row it points at" is the sentence that makes this reviewable rather than a magic `-1`.

**The duplicate drop, and a correction you made that is worth calling out.** The PR description says this merges "a region that overlaps the one already open". The code does not do that, and the comment explains why it must not:

> merging 3-3 into 2-3 hands the wider range's covering definition to an error the definition does not explain, and a real parse failure then disappears from the report

That is a genuine hazard — an overlap merge would have silently *removed* real findings from a report whose entire purpose is honesty — and you caught it, narrowed to an exact-repeat drop, and pinned the distinction with `perl_malformed_source_remains_partial_issue1838`. Good.

**So the one thing I would ask: update the PR description to match.** It is the first thing the next reader sees, and right now it describes an approach you deliberately rejected for a good reason. That reason deserves to be in the description, not only in the source.

Ordering the drop *before* the cap check is also correct, and the comment says so explicitly — a repeat that was never a distinct range must not be counted as one the cap threw away.

## Sequencing

**#1941 has to land first**, and it is currently red on three legs — `test-lsan-macos`, `test-unix (ubuntu-24.04-arm, gcc, 3/3)` and `test-unix (macos-15-intel)`.

I want to be careful here rather than hand you a wrong excuse: those failures completed at **14:35**, and `main` broke at **~15:08**, so they are *not* the broken-main incident. They are real and they are on that branch. (By contrast #1739's failures started at 15:22 and *are* the incident — I have told them so there.)

Your green on this PR is from the same 13:21 push, so it predates the breakage too; it is stale rather than wrong, and will need a re-run once #1993 fixes `main`.

**#1968** carries the CI gate. That one is a maintainer decision rather than a review — a contribution that adds a required gate is something we look at separately on principle, never on the merit of the change. It is not being ignored; it just does not move on the same track as this.

One observation, meant kindly: you currently have five PRs open across two stacks, several sharing files. That is a lot of coupling to hold in your head, and #1976's description already carried one assumption that had gone stale. Landing #1941 and #1896 first would collapse most of it.

CaptainMittens:
The description is fixed — I edited it at 18:25Z, about an hour after your review, so you were reading the old one. It now says:

> Fixed by dropping a range that **exactly repeats** the one already open — same start line and same end line. Ranges that merely overlap are left alone, and that distinction is the whole point.

and the paragraph under it carries the `3-3` into `2-3` hazard and names `perl_malformed_source_remains_partial_issue1838` as the test that pins it. You were right that it belonged there and not only in the source.

## #1941's red legs — two of the three are now green

I checked before repeating your numbers back, and the current run disagrees with the 14:35 one. Run `33512870122` on #1941 is **32 pass, 2 fail**:

| Leg you named | Now |
|:--|:--|
| `test-unix (ubuntu-24.04-arm, gcc, 3/3)` | pass |
| `test-unix (macos-15-intel)` | pass |
| `test-lsan-macos` | still red |

(`ci-ok` is the second failure, which is the aggregate gate reporting the first.)

The remaining one fails here:

```
FAIL tests/test_watcher.c:1479: cbm_watcher_poll_once(w) == 0, expected 1 == 1
SUMMARY: AddressSanitizer: 232937 byte(s) leaked in 375 allocation(s)
```

That assertion writes a file and expects the next poll to report one change. #1941 changes twelve files — `Makefile.cbm`, `internal/cbm/cbm.{c,h}`, `src/mcp/mcp.c`, three pipeline files, `src/store/store.c` and four test files — and none of them is the watcher or its test. The leak summary follows the failed assertion rather than preceding it, so I read it as fallout from the abort, not a second fault.

I am not going to call it environmental from here, because I have been wrong about a "flaky" leg on this repo before. I will run it down on #1941 and report there.

## Sequencing

Agreed, and taken. I am not opening anything further on these two stacks until #1941 and #1896 are in.

One new PR does exist — #1998, for #1995 — but it is off `main`, touches only `src/cypher/cypher.c` and `tests/test_cypher.c`, and shares no file with any of the five. It adds no coupling to the pile you described.

On #1968: understood, and no argument. A contributor should not be the one deciding that a gate becomes required.

DeusData:
Reviewed on a tree with #1941 trial-merged: both fixes bind (red `"2-2,2-2"` / `"1-5"` without your hunk, green with it) and I have no findings on the code. Order is #1941 first — its rebase and one perf fix are on that thread — then a rebase here: your `cbm.c` hunk applies at an offset and the test hunk re-places after main's #1746 tests, nothing else changes. The description refresh from earlier still applies (the code drops exact repeats, it does not merge overlaps — and your comment explaining why overlap-merging would hide a real parse failure is exactly the note that keeps this correct later). Merges on green after that; thanks for the careful reasoning.

CaptainMittens:
CI here is red on three checks, and all three trace to one flake in the Windows harness self-test. Nothing in this PR's diff is involved.

**Root cause — `test / test-windows (windows-latest, CLANG64, x86_64, 2/2)`**

The shard died in `tests/test_parallel_harness_contract.sh`, not in a C suite:

```
Traceback (most recent call last):
  File "<stdin>", line 104, in <module>
ValueError: invalid literal for int() with base 10: ''
rm: cannot remove '.../logs/timeout_exit_race.log': Device or resource busy
```

The script reads a PID out of a file with `int(path.read_text())`. Windows still held a handle on the log (`Device or resource busy`), the read returned an empty string, and `int('')` raised. Every C test in that shard passed first — `pass_1` through `pass_32`, `hang_after_summary`, `no_summary`, `pass_after` and `stubborn_tree` all report `rc=0 pass=1 fail=0`.

**The other two are consequences**

The crash happened before the manifest upload step, so that step logged `No files were found with the provided path: build/c/test-logs/shard-manifest.txt`. `scripts/ci/verify-shard-union.sh` then saw only one of the two shards:

```
OK:   macos-14-cc           — 1 shard(s), union of slices == full suite list
OK:   macos-15-intel-cc     — 1 shard(s), union of slices == full suite list
OK:   ubuntu-24.04-arm-gcc  — 3 shard(s), union of slices == full suite list
OK:   ubuntu-latest-gcc     — 3 shard(s), union of slices == full suite list
FAIL: windows-latest-CLANG64 shard indices [1] != 1..2
```

That fails `test / shard-completeness`, and `ci-ok` gates on both.

**Attribution**

- Neither this branch nor #1941 changes `tests/test_parallel_harness_contract.sh` — zero lines.
- The same shard, same platform and compiler, passes on #1941 with `b04f5450` in the tree.
- The 31 other checks here are green, including ASan+UBSan, LeakSanitizer, MemorySanitizer, ThreadSanitizer on three platforms, all eight gcc `test-unix` shards, and Windows shard 1/2.

`timeout_exit_race` kills processes and reads their PID files, and Windows holds file handles longer than POSIX does, so that is where this class of flake surfaces.

No re-run needed on my account — I will rebase this branch onto main once #1941 lands, as agreed, and the force-push restarts every check. Flagging it so the red is not read as a fault in the diff.

DeusData:
#1941 landed on main as aa44c28e, so this branch now only needs your one fix on top of main. A merge of main into it conflicts (the stacked commits were re-merged on the way in), so a rebase onto main is the clean path: `git rebase origin/main` should leave exactly f89f01e9's change. Keep the sign-off and it merges on green; the review is already done.

CaptainMittens:
Rebased onto `aa44c28`. The branch is now exactly one commit — `e5c17de6`, the duplicate-range and past-EOF fix — with the sign-off intact. The five stacked commits dropped out as already-applied, as you expected.

Diff against main is `internal/cbm/cbm.c` (+39/-2) and `tests/test_parse_coverage.c` (+47).

Checked before pushing:
- `make -f Makefile.cbm lint-format CLANG_FORMAT=clang-format-20` — clean.
- `make -f Makefile.cbm test-focused TEST_SUITES="parse_coverage"` — 37 passed, including `coverage_repeated_error_line_reports_one_range_issue963` and `coverage_range_never_ends_past_the_last_line_issue963`.

One note on the earlier red, in case it comes back: `test-windows-guards` failed on `section_cold_storm`, which is the shared flake tracked in #2057, not this branch. It hit four unrelated branches the same day.

DeusData:
Merged as 98b94a24a3c8 — thank you! Pinning both shapes (the duplicate range from two ERROR nodes on one line, and the range that ended one line past EOF) in test_parse_coverage.c is exactly what keeps the coverage report honest, and it lands cleanly on top of #1941. This ships in v0.10.9.

PR #1976, opened 2026-08-31
fix(extraction): reach a Swift URL built by a constructor

CaptainMittens:
Swift has no URL literal, so real code almost never passes a bare string to a request. It writes `URL(string: "https://…")!`, and the literal then sits two levels below the argument list — past the trailing `!`, which the grammar models as a `postfix_expression`, and inside the constructor's own `value_arguments`.

`extract_url_or_topic_arg` saw only the outer node and gave up, so the URL never reached the service-pattern table and no Route node formed. That is the shape #1892 reported from a real project.

### This PR stacks on #1896

Rebased on 2026-09-02, after #1986 merged as `620613ed`. The branch is now two commits:

| Commit | Belongs to |
|:--|:--|
| reach Swift call arguments through `call_suffix` | **#1896** — approved, please review there |
| reach a Swift URL built by a constructor | **this PR** |

GitHub cannot base a cross-fork pull request on a branch in the fork, so #1896's commit still shows here. It disappears on its own when #1896 merges.

#1896 is the layer underneath: Swift models a call as a target plus a `call_suffix`, so `args` was null for every Swift call and no URL ever arrived. Reaching a bare string argument was only half the problem — this PR reaches the constructor that wraps it.

### The fix

`swift_unwrap_url_constructor()` steps past both wrappers. It unwraps only `URL`, `URLComponents` and `URLRequest`, so any other constructor keeps its own meaning and the outer call does not borrow the inner call's string. A non-literal argument such as `URL(string: base + path)` falls through to the ordinary handling unchanged.

The `value_argument` unwrap added for the bare-string case became `swift_argument_value()`, because the nested argument list needs the same step and the code was identical.

### The vendored grammar fix now comes from main

An earlier version of this branch carried #1986's three commits — the two Swift scanner shift corrections and the `MANIFEST.md` row — because this PR's test is the first thing in the tree to reach the Swift scanner's suppressor path, the rule that stops `try!` emitting its `!` as its own token, which a force-unwrapped `URL(...)!` hits.

#1986 has merged, so the branch no longer carries any of that. This PR touches no vendored file: the diff is `internal/cbm/extract_calls.c`, `tests/test_extraction.c` and `tests/test_pipeline.c` only.

### Tests

Three in `tests/test_extraction.c`, one of them a negative control:

| Test | Covers |
|:--|:--|
| `swift_nested_url_constructor_issue1892` | `URL(string: "…")!` — the force-unwrapped shape from the report |
| `swift_nested_url_no_bang_issue1892` | `URLRequest(url: "…")` — no `!`, so no `postfix_expression` wrapper |
| `swift_non_url_constructor_untouched_issue1892` | `Formatter(pattern: "%s-%d")` must yield **no** string, so the unwrap cannot over-reach |

Plus a pipeline test in `tests/test_pipeline.c` that the Route node now forms.

Local run after the rebase, sanitized build (ASan + UBSan): extraction 335 passed, pipeline 266 passed, no failures.

Refs #1892

### DIFF
--- a/internal/cbm/extract_calls.c
+++ b/internal/cbm/extract_calls.c
@@ -2257,6 +2257,61 @@ static const char *extract_binary_concat_suffix(CBMExtractCtx *ctx, TSNode node)
 }
 
 // Try to extract URL/topic from a positional argument (string or constant).
+/* Swift names its call arguments, so each one is a value_argument that may lead
+ * with a value_argument_label — the `from:` in `data(from: url)`. Return the
+ * value itself, and leave any other node exactly as it came in. */
+static TSNode swift_argument_value(TSNode arg) {
+    if (strcmp(ts_node_type(arg), "value_argument") != 0 || ts_node_named_child_count(arg) == 0) {
+        return arg;
+    }
+    TSNode val = ts_node_named_child(arg, 0);
+    if (strcmp(ts_node_type(val), "value_argument_label") == 0 &&
+        ts_node_named_child_count(arg) > 1) {
+        val = ts_node_named_child(arg, 1);
+    }
+    return val;
+}
+
+/* Swift has no URL literal, so almost no real code passes a bare string to a
+ * request. It writes `URL(string: "https://…")!` instead, and the literal then
+ * sits two levels down: past the trailing `!`, which the grammar models as a
+ * postfix_expression, and inside the constructor's own argument list.
+ *
+ * Unwrap both so that literal is as reachable as a bare one. Only the three
+ * Foundation types that take a URL string are unwrapped — any other call keeps
+ * its own meaning, and a non-literal argument such as `URL(string: base + path)`
+ * falls through to the ordinary handling unchanged. */
+static TSNode swift_unwrap_url_constructor(CBMExtractCtx *ctx, TSNode arg) {
+    /* Step past a trailing "!" or "?". */
+    if (strcmp(ts_node_type(arg), "postfix_expression") == 0) {
+        TSNode target = ts_node_child_by_field_name(arg, TS_FIELD("target"));
+        if (!ts_node_is_null(target)) {
+            arg = target;
+        }
+    }
+    if (strcmp(ts_node_type(arg), "call_expression") != 0) {
+        return arg;
+    }
+    TSNode callee = ts_node_named_child(arg, 0);
+    if (ts_node_is_null(callee) || strcmp(ts_node_type(callee), "simple_identifier") != 0) {
+        return arg;
+    }
+    const char *name = cbm_node_text(ctx->arena, callee, ctx->source);
+    if (!name || (strcmp(name, "URL") != 0 && strcmp(name, "URLComponents") != 0 &&
+                  strcmp(name, "URLRequest") != 0)) {
+        return arg;
+    }
+    TSNode suffix = cbm_find_child_by_kind(arg, "call_suffix");
+    if (ts_node_is_null(suffix)) {
+        return arg;
+    }
+    TSNode inner = cbm_find_child_by_kind(suffix, "value_arguments");
+    if (ts_node_is_null(inner) || ts_node_named_child_count(inner) == 0) {
+        return arg;
+    }
+    return swift_argument_value(ts_node_named_child(inner, 0));
+}
+
 static const char *extract_positional_url(CBMExtractCtx *ctx, TSNode arg, const char *ak) {
     /* JS/TS template literals: `/things/${id}` normalizes to "/things/{}" so the
      * client URL joins the server route's canonical placeholder (issue #1006). */
@@ -2300,15 +2355,12 @@ static const char *extract_url_or_topic_arg(CBMExtractCtx *ctx, TSNode args) {
         }
         /* Swift wraps each argument in a value_argument that may lead with its
          * label, so `data(from: url)` would otherwise yield the label `from`
-         * rather than the value. Step past a leading value_argument_label. */
-        if (strcmp(ts_node_type(arg), "value_argument") == 0 &&
-            ts_node_named_child_count(arg) > 0) {
-            TSNode val = ts_node_named_child(arg, 0);
-            if (strcmp(ts_node_type(val), "value_argument_label") == 0 &&
-                ts_node_named_child_count(arg) > 1) {
-                val = ts_node_named_child(arg, 1);
-            }
-            arg = val;
+         * rather than the value. */
+        arg = swift_argument_value(arg);
+        /* A Swift URL is usually built rather than written bare, and the
+         * literal then sits inside that constructor. */
+        if (ctx->language == CBM_LANG_SWIFT) {
+            arg = swift_unwrap_url_constructor(ctx, arg);
         }
         const char *ak = ts_node_type(arg);
 
--- a/tests/test_extraction.c
+++ b/tests/test_extraction.c
@@ -4186,6 +4186,52 @@ TEST(swift_labeled_call_string_arg_issue1892) {
     PASS();
 }
 
+/* Swift has no URL literal, so real code builds one and force-unwraps it. The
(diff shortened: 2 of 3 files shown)


DeusData:
Reviewed. The extraction change is good and I have nothing to ask of it.

`swift_unwrap_url_constructor` unwraps an explicit three-name allowlist rather than any constructor, every step is null-guarded and returns the node unchanged on the way out, and `URL(string: base + path)` falls through to ordinary handling. `swift_non_url_constructor_untouched_issue1892` is the test that makes that claim checkable rather than stated — a negative control on an unwrap is exactly the right instinct, because the failure mode of this kind of fix is silently borrowing an inner call's string.

Pulling the `value_argument` unwrap out as `swift_argument_value()` once the nested list needed the same step is the right call too.

Three things about the stack, so you can sequence it without guessing.

**1. #1896 should go first, and it can go on its own.** It is the base layer, it is self-contained, and — unlike this one — it touches no vendored file, so it has no manifest obligation. Once it merges, this PR narrows to its own commit and gets much easier to read.

**2. This PR needs a `MANIFEST.md` row, same as #1986.** It changes `internal/cbm/vendored/grammars/swift/scanner.c` and updates `scripts/vendored-checksums.txt`, but not `internal/cbm/vendored/grammars/MANIFEST.md`. Any change to vendored sources has to be recorded in that file's "Local source patches" table — the vendoring table claims our copy matches the pinned upstream commit, and a local patch makes that false unless it is written down, so a future re-vendor would silently drop the fix. The suggested row is on #1986.

**3. #1977 is closed — #1986 supersedes it, and that changes one of your assumptions here.** #1986 is by you and strictly larger: it fixes `eat_operators` as #1977 did *and* `OP_SYMBOL_SUPPRESSOR` (`1UL <<` → `1ULL <<`), which matters because `unsigned long` is 32-bit under LLP64, so that one is undefined on exactly the Windows leg you identified. Full reasoning is on #1977.

The consequence for this branch: this PR's description says the carried grammar commit is byte-identical to the standalone one and will therefore merge without conflict. That was true of #1977 but is **not** true of #1986 — the superset has a hunk yours does not. So if #1986 lands first, expect this branch to need a rebase rather than to shrink on its own. Worth knowing before it surprises you.

**One unrelated heads-up:** `main` does not currently compile — two changes landed the same type short-name index into `type_registry.h` and git merged them into duplicate members. Your green here predates that, so it is stale rather than wrong; checks will need a re-run once main is fixed. #1993 has the repair. Nothing for you to do.

Thanks for splitting the grammar fix out into its own PR rather than leaving it buried in this one, and for laying the stack out in the description — the commit-to-PR table made this quick to follow.

CaptainMittens:
Manifest row added, branch rebuilt on current `main`, and the description corrected. Thank you for the #1977 → #1986 catch — it changed what the right fix here was.

## The stale assumption, and what I did with it

You were right that "byte-identical, so it merges without conflict" was true of #1977 and false of #1986. Rather than delete the claim, I made it true again: **this branch now carries #1986's three commits in full** — `eat_operators`, `OP_SYMBOL_SUPPRESSOR`, and the `MANIFEST.md` row — instead of #1977's single commit.

I checked that rather than assuming it. Diffing `origin/main...HEAD` for `swift/scanner.c`, `MANIFEST.md` and `scripts/vendored-checksums.txt` across this branch and #1986 returns nothing — the vendored change is identical on both. So this PR shrinks on its own when #1986 lands, which is what the description now says.

That also answers your manifest point: the row arrives with #1986's commit rather than as a separate edit here, so the two branches cannot drift into two different rows for one patch.

## Current stack

| Commit | Belongs to |
|:--|:--|
| `2fd719d8` reach Swift call arguments through `call_suffix` | **#1896** |
| `68cc0658` reach a Swift URL built by a constructor | **this PR** |
| `93ebe29e` + `8f00939d` + `82e1de8c` the scanner shift fix and its manifest row | **#1986** |

Merge order as you set it: #1896, then #1986, then this one narrows to a single commit.

Verified on the rebuilt branch with `-fsanitize=address,undefined`: `test-runner extraction pipeline grammar_regression` → **598 passed**, no failures and no UBSan runtime reports. All four Swift tests pass, including `swift_force_unwrap_scanner_shift` which is the one that reaches the suppressor path. `bash scripts/security-vendored.sh` passes too.

## On the coupling

Taken, and it was fair. Five open PRs across two stacks with shared files is how a stale assumption survived into a description in the first place — I could not hold all of it, so I stopped re-reading the parts I thought I already knew.

I am opening nothing further on these stacks until #1896 and #1986 are in.

DeusData:
Status update: **the thing this was waiting on has landed, and that is also why it now conflicts.**

#1986 merged earlier today as `620613ed`, bringing the Swift scanner shift fix and the `MANIFEST.md` row onto `main`. This branch carries its own copies of those same commits — `93ebe29e`, `8f00939d`, `82e1de8c` — which is exactly why GitHub now reports `CONFLICTING`.

**A rebase onto `main` should resolve it and shrink the diff considerably**: the three Swift-scanner commits become no-ops because their content is already upstream, leaving just the change this PR is actually about — `extract_calls.c` reaching a Swift URL built by a constructor, plus its tests.

There is also the chain to #1896 ("reach Swift call arguments through `call_suffix`"). That one is approved on our side and is also `DIRTY` now, so it needs the same rebase. If #1896 goes first, this one rebases on top of it cleanly; if you would rather collapse them, that works too — they touch the same file and neither is large.

My earlier review stands: reviewed and cleared, blocked only on sequencing. Nothing about the change itself needs to move.

Sorry for the wait — the queue is finally moving properly today, and your Swift work is a good chunk of what has landed: #1986 merged, and #2008, #2009, #1875, #1877 besides.

CaptainMittens:
Rebased — #1896 first, then this branch on top of it. Both are green on my machine and CI is re-running.

This PR is now the two commits you described: #1896's `call_suffix` change, and this PR's own URL constructor change. It touches no vendored file.

One detail worth recording, because it changes what a plain rebase does here. Of the three #1986 commits this branch carried, only the try-bang one matched its merged counterpart by patch-id. The int-shift commit and the `MANIFEST.md` row did not — their content had drifted from the copies that landed in `620613ed`. So `git rebase origin/main` would have replayed those two rather than dropping them, and produced either a conflict or a second MANIFEST row. I dropped all three explicitly and replayed only the URL constructor commit.

I also rewrote the description. It still described the five-commit stack and explained why the branch carried the vendored diff, and neither is true now.

Local run after the rebase, sanitized build (ASan + UBSan): extraction 335 passed, pipeline 266 passed, no failures. That run includes `swift_force_unwrap_scanner_shift` from main, so the scanner fix and this change hold together.

DeusData:
#1896 is on `main` as `5802ccdd` — thank you for landing the rebase in exactly that order, it made the first merge a clean one-commit read.

This branch is now the single commit on top, as intended. Its CI run was queued before the merge, so I re-queued the full matrix against the new `main` — standard practice here after every merge, not a signal about the change. When it comes back green this merges; nothing needed from you.

DeusData:
Merged as `28b4093c` — and I owe you an apology. I said "when it comes back green this merges" on 3 September. It came back green and then sat for nineteen days with nothing needed from you. That was my backlog, not your PR.

Verified tonight on the merge result with today's `main`: the functions you changed have not been touched on `main` since your base, the Swift gating holds (`ctx->language == CBM_LANG_SWIFT` guards the constructor unwrap), the shared argument-label unwrap is byte-for-byte what every other language had before, `extraction` and `pipeline` pass, the route materialises in `pipeline_swift_nested_url_makes_route_issue1892`, and the tests go red with the production change reverted.

The detail I want on the record: you checked the three carried commits by patch-id, found only one matched, and dropped all three explicitly rather than letting a plain rebase replay two and mint a second `MANIFEST.md` row. That is exactly the failure I would have merged without noticing. Thank you.

Issue #558, opened 2026-06-22
Class nodes always have in_degree/out_degree = 0 despite real INHERITS edges existing


Ilyes512:
### Version
codebase-memory-mcp 0.8.1

### Platform
macOS (Apple Silicon)

### What happened, and what did you expect?

`search_graph` returns `Class` nodes with `in_degree: 0` and `out_degree: 0` even when real `INHERITS` edges exist for those nodes in the graph.

**Example — `AbstractAttachmentDto` has 8 subclasses, but `search_graph` reports `in_degree: 0`:**

```json
search_graph(name_pattern=".*AbstractAttachmentDto.*")

→ { "name": "AbstractAttachmentDto", "label": "Class", "in_degree": 0, "out_degree": 0 }
```

**Confirmed via Cypher that the edges actually exist:**

```cypher
MATCH (child)-[:INHERITS]->(parent)
WHERE parent.name = 'AbstractAttachmentDto'
RETURN child.name
```
→ Returns 8 rows correctly.

**Also confirmed that the degree properties are stored as `0` on the node itself, not computed live:**

```cypher
MATCH (n:Class {name: 'AbstractAttachmentDto'})
RETURN n.in_degree, n.out_degree
```
→ 0, 0

### Root cause hypothesis

The `in_degree`/`out_degree` properties appear to be written onto nodes during indexing (not computed dynamically from the graph). It looks like these counts are not being updated for `Class` nodes after `INHERITS` edges are resolved, leaving them at their initial value of `0`.

### Impact

- `search_graph` returns misleading connectivity data for all classes
- Any `min_degree` filter to find well-connected or abstract classes will silently exclude everything
- Agents fall back to grep because `search_graph` results look like isolated/unused nodes

### Workaround

Use `query_graph` with a Cypher traversal directly — the edges are correct, only the stored property is wrong:

```cypher
MATCH (child)-[:INHERITS]->(parent {name: 'MyAbstractClass'})
RETURN child.name, child.file_path
```

### Confirmations
- [x] I searched existing issues and this is not a duplicate.

DeusData:
Hey, thx for this finding. Will check ASAP

mvanhorn:
[...]

PR #1986, opened 2026-09-01
fix(grammar): stop the Swift try-bang entry shifting past its own type

### LINKED ISSUE #1978: Swift grammar scanner shifts past the width of its type in two places — one is undefined on Windows
### Version

### Platform

Windows (x64)

### Install channel

Built from source

### Binary variant

standard

The vendored Swift grammar scanner shifts a literal `1` past the width of its own type in
two places. Both are undefined behaviour.

| Line | Code | Left operand | Shift count | State |
|--:|:--|:--|--:|:--|
| 514 | `1 << suppressor` | `int`, 32 bits everywhere | reaches 32 | fixed in #1977 |
| 131 | `1UL << FAKE_TRY_BANG` | `unsigned long`, **32 bits on Windows** | exactly 32 | open |

Both live in `internal/cbm/vendored/grammars/swift/scanner.c`.

For line 514, `TOKEN_COUNT` is 33, so the loop counter reaches 32. The mask being tested
is a `uint64_t`, which makes the intent unambiguous.

For line 131, `FAKE_TRY_BANG` is 32 — I counted the `TokenType` enum in that file: 33
entries, and `FAKE_TRY_BANG` is the last. On Linux and macOS `unsigned long` is 64 bits
and the shift is fine. On Windows it is 32 bits, so the shift count equals the full width
of the type.

I expected the entry to produce `1 << 32` as a 64-bit value. On Windows it produces
whatever the compiler decides an undefined shift produces.

**Neither of these is only my reading of the standard.** Upstream
`alex-pinkus/tree-sitter-swift` fixed both independently:

- line 514 in [`fb63a7004f07`](https://github.com/alex-pinkus/tree-sitter-swift/commit/fb63a7004f07) (2026-04-06), "Fix UB in eat_operators"
- line 131 in [`6ab8d1d74ebd`](https://github.com/alex-pinkus/tree-sitter-swift/commit/6ab8d1d74ebd) (2026-08-10), "Fix undefined 32-bit shift in OP_SYMBOL_SUPPRESSOR on LLP64 targets"

Both changes are one token each: `1` to `1ULL`, and `1UL` to `1ULL`.

`MANIFEST.md` pins the Swift grammar at `8abb3e8b3325`, dated 2026-03-20. Both upstream
fixes landed after that pin. Our copy is a faithful vendor rather than a local edit — I
checked it byte-identical to upstream at that commit.

**Why nothing here catches it.** This is the part I think is worth your time, because it
explains why Windows CI has stayed green:

1. Line 131 is a compile-time constant expression, so no sanitizer sees it at run time.
2. `Makefile.cbm:719` compiles every vendored grammar with `-w`, which switches off all
   warnings — including `-Wshift-count-overflow`, the one diagnostic that names this exact
   defect. The comment on line 718 gives the reason: "upstream code has warnings."
3. `CLANG64` on `windows-latest` is in `CORE_WIN` (`.github/workflows/_test.yml:63`), so
   the affected platform is built on every pull request and reports nothing.

> **A suggestion, not part of this report.** Turning on just `-Wshift-count-overflow` for
> vendored grammars — one narrow warning, not `-Wall -Werror` — would catch this class
> across all 104 vendored scanners without reintroducing the upstream warning noise that
> `-w` exists to suppress. I have not tried it and I am not proposing it here. Happy to
> open a separate issue if you want it explored.

### Reproduction

**I have no Windows machine, so I have no runtime reproduction on the affected platform,
and I would rather say that than invent one.** Two honest substitutes.

**1. The mechanism, which you can run anywhere.** `unsigned int` is 32 bits on every
target, so it stands in for Windows' `unsigned long`:

```c
#include <stdint.h>
#include <stdio.h>
int main(void) {
    uint64_t m = 1u << 32;   /* same shape as scanner.c:131 on Windows */
    printf("%llu\n", (unsigned long long)m);
}
```

With warnings on, the compiler names the defect:

```
repro.c:4:21: warning: shift count >= width of type [-Wshift-count-overflow]
    4 |     uint64_t m = 1u << 32;
      |                     ^  ~~
```

Compiled with `-w`, as the grammar build does, that warning disappears and the program
runs. The result is not merely wrong — it is unstable. Three runs of one binary on Apple
clang 21.0.0, arm64:

```
4338156640
4364321888
4377183328
```

At `-O2` it settles on a different value again (`8447164672`). `m` is never actually
assigned; the compiler treats the shift as poison and leaves stack garbage behind. Writing
the same line correctly as `(uint64_t)1 << 32` prints `4294967296` every time.

That is the mechanism, on my platform, with `unsigned int` standing in. **It is not a
measurement of what Windows does with `1UL << 32`** — I have not run that, and I am not
claiming the two produce the same value.

**2. How to confirm the real effect, if you have a Windows box.** Index a Swift file
containing a force-unwrap or `try!` on Windows and on Linux, then compare the extracted
tokens. `OP_SYMBOL_SUPPRESSOR[BANG]` is what stops `try!` emitting its `!` as a token of
its own. If that entry is wrong on Windows, Swift extraction differs there and nothing
currently fails to say so.

**What I am claiming, and what I am not.** Established: the expression is undefined, its
location, and that upstream judged both lines worth fixing. Not established: that a
symptom is visible in extracted output today. I did not want to overstate a bug I found by
reading rather than by running.

### Logs

```text
Not applicable — this is a source-level defect found by inspection, not a runtime failure.
The compiler diagnostic that would normally appear is suppressed by -w on grammar builds.
```

```text
Not applicable — not a memory or performance issue.
```

Not relevant — the defect is in a vendored grammar, independent of project size.

### Confirmations

CaptainMittens:
One token in the same vendored file as #1977. This is the second of the two shifts in #1978, and the one that is undefined on Windows.

Stacked on #1977, because both changes rewrite the same recorded hash in `scripts/vendored-checksums.txt`. Two branches off `main` would collide there. Until #1977 merges this PR shows both commits; after it merges the diff narrows to the one token below.

### The bug

`internal/cbm/vendored/grammars/swift/scanner.c:131` builds the entry that suppresses the `!` of a `try!`:

```c
1UL << FAKE_TRY_BANG, // BANG,
```

`FAKE_TRY_BANG` is the last member of a 33-entry `TokenType` enum, so its value is 32.

`unsigned long` is 64 bits on Linux and macOS. On Windows it is 32 bits, so the shift count equals the width of the type, which is undefined. `1ULL` is 64 bits on every target this project builds for.

### Why nothing catches it

Three things line up, and the third is the one I would not have guessed:

1. The expression is a **compile-time constant**. The compiler folds it, so no sanitizer sees it at run time. UBSan cannot report what never executes.
2. `Makefile.cbm:719` compiles every vendored grammar with `-w`. That switches off `-Wshift-count-overflow`, the one diagnostic that names this exact defect. The comment on line 718 gives the reason: upstream code has warnings.
3. The affected platform is not exotic. `CLANG64` on `windows-latest` sits in `CORE_WIN` (`.github/workflows/_test.yml:63`), so it builds on every pull request and reports nothing.

### What I did and did not verify

I have no Windows machine, so I did not observe a symptom in the product. What I did do is compile the same shape locally, using `unsigned int` because it is 32 bits everywhere:

```c
uint64_t m = 1u << 32;
```

`cc -Wall` warns `shift count >= width of type`. Adding `-w`, as the grammar build does, silences it.

I expected the folded value to be `0` and said so in #1978 before testing. It is not. The variable is simply never assigned, and three runs of one binary printed `4338156640`, `4364321888` and `4377183328`; at `-O2` it printed `8447164672`. The correct form, `(uint64_t)1 << 32`, prints `4294967296` every time. So the practical effect on Windows is a suppressor entry holding whatever happens to be there, not a predictable zero.

### No test, and why

A test here would pass before and after the change on every machine CI can run it on. `unsigned long` is already 64 bits on Linux and macOS, so the expression is correct there; the defect only exists on a target the test suite does not execute. Adding a test that cannot fail would suggest a guarantee that does not exist. The existing `swift_force_unwrap_scanner_shift` from #1977 already keeps a force-unwrap parsing at all.

### Upstream

Not my own reading alone. `alex-pinkus/tree-sitter-swift` made the same one-token change in `6ab8d1d74ebd` (2026-08-10). Our copy is pinned at `8abb3e8b3325` (2026-03-20), so that fix landed after our pin. I checked our copy byte-for-byte against upstream at that pin: it is a faithful vendor, not a local edit that drifted.

Re-vendoring picks up both fixes and five months of other work, but `TOKEN_COUNT` moves from 33 to 34 upstream, so the grammar itself changed. `CONTRIBUTING.md:134` puts a vendored dependency change behind a design discussion, so I have not proposed it here.

### Files

| File | Change |
|:--|:--|
| `internal/cbm/vendored/grammars/swift/scanner.c` | `1UL` to `1ULL` on line 131 |
| `scripts/vendored-checksums.txt` | the one recorded hash for that file, as `scripts/security-vendored.sh --update` writes it |

Layer 8 of the security gate compares vendored content against that manifest, so the edit and its recorded hash land together. `scripts/security-vendored.sh` exits 0 on this branch.

### One thing worth your call

`-w` on the grammar build is what hid this. Turning on the single warning `-Wshift-count-overflow` for vendored grammars — one narrow diagnostic, not `-Wall -Werror` — would catch this class across all 104 vendored scanners without touching the rest of the noise. I have not made that change here. Say the word and I will open it as its own issue.

Fixes #1978

Refs #1892


DeusData:
Good catch, and the fix is right. One piece of paperwork is missing before it can land.

**Please add a row to the "Local source patches" table in `internal/cbm/vendored/grammars/MANIFEST.md`.** Any change to vendored sources has to be recorded there, because the vendoring table's claim is that our copy matches the pinned upstream commit — a local patch makes that claim false unless it is written down, and a future re-vendor would silently drop your fix.

You are in good company: three of the four existing rows are the same class of UBSan fix in a vendored scanner.

Suggested row, matching the existing shape:

```
| swift | `swift/scanner.c`, `OP_SYMBOL_SUPPRESSOR` + `eat_operators` | `1 <<` / `1UL <<` → `1ULL <<` | UBSan: `1 << suppressor` shifts an `int` by up to `TOKEN_COUNT` bits, undefined once the index reaches 31, while the mask it feeds is `uint64_t` |
```

Worth upstreaming too, if you have the appetite — `alex-pinkus/tree-sitter-swift` carries the same bug at the pinned commit, and an upstream fix means the next re-vendor drops the local patch instead of re-applying it.

On the change itself, two things I want to name because they are the reason this was easy to review:

**You found a bug that the normal test build actively hides.** Recovering UBSan prints the diagnostic and continues, so the shift has been executing and being ignored for as long as it has existed. Identifying that the Windows CLANGARM64 leg runs UBSan in trap mode — where the same shift is an illegal instruction rather than a log line — is what turns "a warning nobody reads" into "a crash on one platform".

**And you wrote the test's own limitation into the test.** "This test cannot go red here… parsing this file at all is the check" is exactly the right comment to leave. A future reader who sees it pass on macOS will not mistake that for evidence. That honesty is worth more than the four-character fix.

Add the manifest row and this is ready.

CaptainMittens:
Manifest row added, and the branch is rebased onto current `main`.

## The row

I used your shape and extended the reason to name both sites, since this PR fixes two:

```
| swift      | `swift/scanner.c`, `OP_SYMBOL_SUPPRESSOR` + `eat_operators` | `1UL <<` / `1 <<` → `1ULL <<` | UBSan: `1 << suppressor` shifts an `int` by up to `TOKEN_COUNT` bits, undefined once the index reaches 31, while the mask it feeds is `uint64_t`. `1UL << FAKE_TRY_BANG` is the same defect on Windows, where `unsigned long` is 32 bits and `FAKE_TRY_BANG` is 32; the CLANGARM64 leg runs UBSan in trap mode, so there it is an illegal instruction rather than a log line |
```

**The commit touches a second file, and that is deliberate rather than scope creep.** `MANIFEST.md` is itself listed in `scripts/vendored-checksums.txt`, so writing the row breaks its own recorded checksum unless that line moves with it.

## The rebase, and what the conflict was

The branch was 44 commits behind and conflicting. The conflict was one line, and it was that same self-reference: `MANIFEST.md`'s checksum, bumped on `main` by another patch row and bumped here by mine.

I resolved it by recomputing the hash from the merged file rather than taking either side. Both recorded values described a `MANIFEST.md` that no longer exists once both rows are present, so either choice would have left the file failing its own check while looking resolved.

`bash scripts/security-vendored.sh` passes on the rebased branch, and the sanitized build is clean:
`test-runner extraction pipeline grammar_regression` → **591 passed**, no failures and no UBSan runtime reports, built with `-fsanitize=address,undefined`.

## Upstreaming

Yes, and I have not done it yet — I will open it against `alex-pinkus/tree-sitter-swift` and link it here rather than leave you to guess whether it happened. If it lands, the next re-vendor drops this local patch instead of re-applying it, which is the better end state for a row like this.

## On the two things you named

The UBSan point is the one I want to be honest about: I did not identify the trap-mode difference from insight. I hit the recovering-UBSan diagnostic while running an unrelated test, went looking for which leg treats it as fatal, and found the CLANGARM64 one. The finding was real; the route to it was ordinary.

The test comment was deliberate. A test that cannot fail on the machine most people run it on is worth less than it looks, and writing that down felt cheaper than having someone later mistake a green macOS run for coverage.

CaptainMittens:
The MANIFEST row is in, and I have to correct two things I said.

### The upstream fix is already upstream — I cannot open it

I said I would open this against `alex-pinkus/tree-sitter-swift` and link it here. I went to do that and found both changes already there, in commits that landed after our pinned revision:

| Our change | Upstream commit | Landed | Their issue |
|:--|:--|:--|:--|
| `eat_operators`, `1 <<` → `1ULL <<` | `fb63a7004f07` | 2026-04-06 | #558 |
| `OP_SYMBOL_SUPPRESSOR`, `1UL <<` → `1ULL <<` | `6ab8d1d74ebd` | 2026-08-10 | — |

`fb63a7004f07` is byte-identical to `3f831936` here, at the same line. We pin `8abb3e8b3325` (2026-03-20), which predates both. So this is a stale pin, not an unfixed upstream bug, and there is nothing for me to send. That retracts the offer you recorded on #1978 — please do not wait on it.

Both fixes still matter here. `main` today still carries `1UL <<` at line 131 and `1 <<` at line 514.

### The row I first wrote would have misled a re-vendor

Your reason for wanting the row is that a re-vendor would otherwise silently drop the fix. That reason is exactly right for the other four rows, where upstream never took the change. It inverts for this one. The table says:

> Re-vendoring from upstream must re-apply these.

Anyone following that on the Swift row after a re-vendor past 2026-08-10 would hand-apply a change the newer file already has.

So the row differs from your suggested wording in two ways. The reason column now names both upstream commits, our pin, and the dates, and says to **delete** this row on a re-vendor rather than re-apply it. The table intro gains one clause allowing that case, because it is a property of the table rather than of one row:

> Re-vendoring from upstream must re-apply these, unless the reason column names an upstream commit that already carries the change — then drop the row instead.

The `location` and `patch` columns are unchanged from your suggestion.

### What is not here

Re-vendoring the Swift grammar is the real end state and it is not in this pull request. `TOKEN_COUNT` goes 33 to 34 between our pin and upstream `main`, so the generated `parser.c` moves with the scanner. That is a bigger change and yours to schedule.

`bash scripts/security-vendored.sh` passes — `MANIFEST.md` is itself listed in `scripts/vendored-checksums.txt`, so its own recorded checksum moved with the edit.

I amended the MANIFEST commit rather than stacking a correction on it, so the branch is still three commits. The two code commits are untouched.

DeusData:
Merged as `620613ed`. You did more than I asked for, and the extra part is the valuable bit.

I asked for a `MANIFEST.md` row and suggested upstreaming the fix if you had the appetite. You went and checked whether it was *already* upstream — and it was, twice: `fb63a7004f07` (2026-04-06, upstream #558) for `eat_operators`, and `6ab8d1d74ebd` (2026-08-10) for the `OP_SYMBOL_SUPPRESSOR` entry. Our pin `8abb3e8b3325` (2026-03-20) predates both, so this is a backport rather than a local invention. That is a materially different fact about the patch than the one I was working from.

**And you encoded the consequence in the manifest rather than only in the row:**

> Re-vendoring from upstream must re-apply these, unless the reason column names an upstream commit that already carries the change — then drop the row instead.

That turns a static instruction into a self-retiring one. Without it, a re-vendor past 2026-08-10 would dutifully re-apply a patch upstream no longer needs, and the local-patch table would accumulate entries nobody can safely remove because the reason for each has been lost. Naming the upstream commit *in the row* is what makes the deletion decidable later. That is a better answer than the one I proposed.

The original diagnosis stands too: recovering UBSan prints and continues, which is why `1 << suppressor` executed unnoticed for as long as it existed, and the Windows CLANGARM64 leg runs UBSan in trap mode where the same shift is an illegal instruction rather than a log line. Writing that lane-binding into the test's own comment — "this test cannot go red here… parsing this file at all is the check" — is the sort of honesty that stops a future reader mistaking a macOS pass for evidence.

Before merging I confirmed the 12-commit gap to `main` was inert: no commit had touched `swift/scanner.c`, `MANIFEST.md`, `vendored-checksums.txt` or `test_extraction.c` since your base, so the green described exactly what landed.

Thank you — that is your second merge today, after #2008.

PR #1992, opened 2026-09-01
docs: name the history-growth cost of the team artifact, and how to bound it

DeusData:
Distilled from #1697.

@angusgastle hit this on a real monorepo: the team-shared artifact regenerates on every index, they committed each regeneration, and history reached ~6 GB across ~350 commits of that one path. They measured it, rewrote it out with `git-filter-repo`, and reported a ~10x clone reduction afterwards. That is a real cost the README never warned about, and finding it is their credit.

This lands the guidance with two corrections.

**The instruction as written would not have worked.** #1697 said to replace the auto-created `merge=ours` line with `.codebase-memory/graph.db.zst filter=lfs …`. That line lives in `.codebase-memory/.gitattributes` with the bare pattern `graph.db.zst`, and a slash-bearing pattern inside that file anchors to `.codebase-memory/.codebase-memory/` — it matches nothing. Following it would silently leave LFS disengaged *and* drop `merge=ours`, which is worse than doing nothing, and `ensure_gitattributes` is `O_EXCL` create-only-if-absent (`src/pipeline/artifact.c:776`) so nothing repairs it later. The tracking line belongs in the repo-root `.gitattributes`, alongside the auto-created file rather than instead of it: gitattributes resolve per attribute from the nearest file outward, so `merge=ours` keeps coming from `.codebase-memory/` and only `filter=lfs` comes from the root.

**LFS is not free, and a recommendation we own should say so.** It relocates the bloat to a metered store — GitHub bills LFS storage and bandwidth, LFS objects cannot be pruned without contacting support, and a teammate without `git lfs install` checks out a pointer file, which the integrity-checked import refuses, dropping them back to the full reindex the artifact exists to avoid.

So the cheaper fix leads: **don't commit every regeneration.** LFS follows, for teams that genuinely need the artifact to move on every commit.

Also names the file the `merge=ours` line is written to. The old bullet said only "a `.gitattributes` line", which is what made the wrong placement the natural reading.

Docs only; no code paths touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,13 @@ Commit a single compressed file to your repo and your teammates skip the reindex
   - **Best** (`zstd -9` + index strip + `VACUUM INTO`) — written on explicit `index_repository`
   - **Fast** (`zstd -3`) — written by the watcher for low-latency incremental updates
 - **Bootstrap**: when no local DB exists but the artifact is present, `index_repository` imports the artifact first, then runs incremental indexing — avoiding the full reindex cost
-- **No merge pain**: a `.gitattributes` line with `merge=ours` is auto-created on first export, so concurrent edits don't produce conflicts on the binary artifact
+- **No merge pain**: a `.codebase-memory/.gitattributes` line with `merge=ours` is auto-created on first export, so concurrent edits don't produce conflicts on the binary artifact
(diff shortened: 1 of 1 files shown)


DeusData:
Closing this. #1697 is the right home for the change — its author rebased it hours ago and is plainly still on it, so reimplementing it here would be taking work off an active contributor for no reason.

The review that produced this diff is on #1697, and the wording has been handed over there for them to use directly. Nothing is lost by closing.

PR #1996, opened 2026-09-01
fix(store,ui): export paired free functions for FFI callers (#1762)

### LINKED ISSUE #1762: cbm_store_fetch_call_edges / cbm_leiden / cbm_layout_to_json document caller-owned memory but export no matching free function

I'm compiling `codebase-memory-mcp` as a DLL to call it from C# (P/Invoke), and three public API functions document caller-owned memory but export no matching free function — unlike the rest of the API:

- `cbm_store_fetch_call_edges` (store.h:471) — *"caller frees both"* — two `int64_t*` arrays
- `cbm_leiden` / `cbm_louvain` (store.h:843, 847) — *"the caller frees it"* — one `cbm_louvain_result_t*` array
- `cbm_layout_to_json` (layout3d.h:72) — *"Caller must free()."* — one `char*`

The rest of the API does follow a paired-free pattern: `cbm_store_find_nodes`→`cbm_store_free_nodes`, `cbm_layout_compute`→`cbm_layout_free`, `cbm_store_get_schema`→`cbm_store_schema_free`, etc. From C, a bare `free()` works fine because of static linking — but from C# there's no way to invoke the DLL's internal `free()` unless it's exported by name.

### Proposed solution

Export a paired free for each of the three, following the existing pattern. I can implement this. Two questions on approach before I send the PR:
1. 3 specific free functions (`cbm_store_free_call_edges`, `cbm_leiden_free`, `cbm_layout_free_json`), or a generic `cbm_free(void*)`?
2. One PR for all 3, or one per function?

### Alternatives considered

None that hold up across an FFI boundary — a foreign caller has no way to reach the DLL's internal `free()` without an exported symbol.

### Confirmations

Jumaga2015:
## Summary

Fixes #1762. Three public API functions document caller-owned memory but
export no matching free function, unlike the rest of the API:

- `cbm_store_fetch_call_edges` (store.h) — *"caller frees both"* — two `int64_t*` arrays
- `cbm_leiden` / `cbm_louvain` (store.h) — *"the caller frees it"* — one `cbm_louvain_result_t*` array
- `cbm_layout_to_json` (layout3d.h) — *"Caller must free()."* — one `char*`

The rest of the API already follows a paired-free pattern
(`cbm_store_find_nodes`→`cbm_store_free_nodes`,
`cbm_layout_compute`→`cbm_layout_free`,
`cbm_store_get_schema`→`cbm_store_schema_free`, …). From C a bare `free()`
works because of static linking, but a caller across an FFI boundary has no
way to reach the DLL's internal `free()` unless it is exported by name.

Concretely: with no exported free, a P/Invoke caller must either skip these
functions entirely (losing cycle/SCC analysis in the case of
`cbm_store_fetch_call_edges`) or leak on every call. Releasing from the CLR
side is not an option — `Marshal.FreeHGlobal` uses the Win32 heap and
`Marshal.FreeCoTaskMem` the COM heap, neither of which allocated this
memory, and using the wrong allocator does not throw: it corrupts the heap
and takes the process down later, somewhere unrelated.

This adds the three missing pairs, following the existing naming:
`cbm_store_free_call_edges`, `cbm_leiden_free`, `cbm_layout_free_json`.

I asked in the issue whether you preferred three specific frees or a generic
`cbm_free(void*)`, and one PR or three. With no reply I went with what the
codebase already does everywhere else — specific names, one PR — but I am
happy to reshape this either way if you prefer the other option.

## Notes for review

- **`free()` is the right call inside the DLL.** The project links with
  `-Wl,--wrap=malloc,free,...` on MinGW (#581), so these three functions
  compile to `jmp __wrap_free`, exactly like the allocating side. The
  allocate/free pair still matches; `__wrap_free` itself decides whether the
  block belongs to mimalloc or the CRT.
- All three types are flat (`int64_t`, a `{int64_t; int}` struct, a yyjson
  string) — no owned inner pointers, so a single `free()` is complete.
  All three are NULL-safe by `free()`'s contract.
- I also updated the one internal production caller (`mcp.c`, SCC analysis)
  to use `cbm_store_free_call_edges` rather than a bare `free()`. Happy to
  drop that hunk if you would rather keep the PR strictly additive.

## Test plan

- [x] Three new tests in `test_store_arch.c` — `fetch_call_edges_basic`,
      `fetch_call_edges_truncated`, `fetch_call_edges_empty_project`.
      `cbm_store_fetch_call_edges` had no direct test coverage before this;
      the truncation test uses a 5-edge fixture against a cap of 2, so it
      exercises the limit rather than passing vacuously.
- [x] Existing `cbm_leiden`/`cbm_louvain`/`cbm_layout_to_json` tests updated
      to release through the new paired frees.
- [x] `store_arch`, `ui`, `mcp` suites green under real ASan+UBSan
      (clang 22, `x86_64-w64-windows-gnu`): 326 passed, 10 skipped
      (all skips are platform guards, POSIX-only fork/alarm), no sanitizer
      reports.
- [x] Builds clean under `-Wall -Wextra -Werror`.
- [x] `clang-format` clean on every touched file.

Note: the full `scripts/test.sh` venue leg (steps 0o–0u) packages and
checksums darwin-arm64 / linux-amd64 / windows-amd64 release bundles, which
is not reachable from a single Windows dev box — those steps fail here with
or without this change. CI on Linux is the real gate for that leg.

Made with [Claude Code](https://claude.com/claude-code)


DeusData:
Approved on merit. I checked every claim against the tree rather than taking them, and they all hold:

- `store.h:571` *"caller frees both"*, `store.h:991` *"the caller frees it"*, `layout3d.h:71` *"Caller must free()."* — three documented caller-owned allocations with no exported release.
- The paired-free convention is real and consistent: `cbm_layout_free` (`layout3d.h:69`), `cbm_store_schema_free` (`store.h:772`), `cbm_store_free_nodes` (`store.h:1008`). Your three names follow it.
- The `--wrap` reasoning is correct: `Makefile.cbm:145` builds `MIMALLOC_WRAP_FLAGS` from `-Wl,--wrap=$(sym)`, so inside the binary a `free()` in your new functions lowers to `jmp __wrap_free` exactly as the allocating side does. The allocate/free pair stays matched and `__wrap_free` decides mimalloc-vs-CRT. (`Makefile.cbm:150` notes macOS ld has no `--wrap` and stays census-only — which does not disturb the argument, since both halves take the same path within one binary either way.)

**The heap-mismatch paragraph is what makes this a bug rather than a nicety**, and I want to quote it back because it is the part a reviewer might otherwise skim:

> `Marshal.FreeHGlobal` uses the Win32 heap and `Marshal.FreeCoTaskMem` the COM heap, neither of which allocated this memory, and using the wrong allocator does not throw: it corrupts the heap and takes the process down later, somewhere unrelated.

That is exactly right, and it is why "the caller can just free it" is true from C and false across an FFI boundary. The failure it produces is the worst kind — silent, delayed, and attributed to unrelated code. So today a P/Invoke caller's only safe options are to leak on every call or to skip `cbm_store_fetch_call_edges` entirely, which means giving up cycle and SCC analysis. Neither is acceptable for a documented public API.

**Noting that all three payloads are flat** — `int64_t`, a `{int64_t; int}` struct, and a yyjson string, no owned inner pointers — is the check that makes a single `free()` complete rather than merely plausible. That is the question I would have asked first.

Updating the internal `mcp.c` SCC caller to use `cbm_store_free_call_edges` instead of a bare `free()` is the right instinct too: it keeps the one in-tree caller honest against the convention, so the pairing does not rot.

## On the unanswered question

> I asked in the issue whether you preferred three specific frees or a generic `cbm_free(void*)`, and one PR or three. With no reply I went with what the codebase already does everywhere else.

You asked, we did not answer, and you chose by following existing convention rather than guessing. That is the best possible way to handle our silence, and the answer you arrived at is the one I would have given: **three specific frees.** A generic `cbm_free(void*)` would be a new, untyped public entry point that no existing pair uses, and it would make every future caller's ownership less legible, not more. One PR is right too — the three are one defect class.

Sorry you had to proceed without a reply. That is the third PR today where a contributor asked a design question on an issue and got silence, which is a queue problem on our side, not yours.

## Before merge

This adds three exported symbols, so it is a public-surface addition — additive, conventional, and reversible only in the sense that removing an export is a break, which is worth being deliberate about. I am satisfied it earns them.

CI is heavily backlogged on this repo right now (44 runs queued against 3 executing), so I am deliberately not forcing a re-run. Once it reports green I will merge.

Thank you — this is a careful, well-evidenced fix.

DeusData:
Your CI came back 31 green / 4 red. **None of the four are yours** — I traced them, and it is worth telling you exactly what happened so you do not go looking.

The root failure is on the Windows shard, and it is a harness error rather than a test assertion:

```
FAIL: parallel scheduler infrastructure error: parallel scheduler cleanup failed:
hang_after_summary: suite 'hang_after_summary' leader exited leaving live descendants
```

That string is raised at `scripts/run-test-wave.py:172`, inside the parallel scheduler's own process-tree cleanup. `hang_after_summary` is a deliberately-hanging fixture suite whose whole purpose is to exercise that cleanup path — it is our test harness testing itself, and it has nothing to reach into three new `free()` wrappers.

The other three fall out of it: `shard-completeness` fails at "Verify cross-shard union per leg" because the Windows shard 2/2 died before reporting its suites, and `ci-ok` aggregates.

**We already have the fix open.** PR #1823 changes precisely that line:

```diff
-            if windows_descendants(process.pid, kill_grace):
+            if windows_descendants(process.pid, windows_helper_timeout(kill_grace)):
```

Its reasoning is exactly your symptom: `--kill-grace` budgets how long a *doomed process* may take to die, while the Windows helper call needs process spawn plus CIM startup, which on a loaded runner routinely exceeds a second. Conflating the two makes a small grace value fail the whole wave. Our Actions pool is currently saturated — 47 runs queued against 3 executing — which is the loaded-runner condition that makes this fire.

So: nothing to change on your side. #1823 is blocked on a maintainer step, I am chasing it, and I will re-run this once it lands.

My review from earlier stands — approved on merit, merging on green.

DeusData:
Thank you, and I am sorry for the silence: I approved this on merit on 2 September and then left it for twenty days. The review stands — three documented caller-owned allocations with no exported release, named to match the existing convention, and your `--wrap`/mimalloc reasoning is correct.

One thing changed underneath you, and it is ours to resolve, not yours. On 13 September we added a memory-core linter with a per-file ratchet of raw allocator call sites (`scripts/lint-memory-core.py`, `scripts/memory-core-baseline.txt`), and it is now part of `lint-ci`. Your branch was linted before that gate existed. On the merge result it fails:

```
src/store/store.c:   358 -> 361
src/ui/layout3d.c:    66 ->  67
```

Your four `free()` calls are *right*: the allocating side (`cbm_store_fetch_call_edges` and friends) is still raw `malloc`, so routing the releases through the memory core would mismatch the heap — the exact bug this PR exists to prevent. So the honest choices are ours: migrate the allocating side to the core in the same change so the count can go down, or record a deliberate exception in the baseline. I will settle that with the maintainer and come back here; nothing is needed from you, and please do not rebase or "fix" the lint in the meantime, because either resolution is a policy call rather than a patch.

Issue #1334, opened 2026-07-29
query_graph: multi-key ORDER BY silently drops LIMIT (6326 rows / 117 KB instead of 5) on v0.9.1-rc.1


b1rdex:
**Version:** `v0.9.1-rc.1` (built from the tag, `560ad40`, macOS arm64)
**Tool:** `query_graph`

## Summary

A `LIMIT` is silently ignored when the query has an `ORDER BY` with **more than one sort key**. Single-key `ORDER BY … LIMIT` works; adding a second key makes the whole result set come back.

This is a token-safety problem rather than a cosmetic one: on a 24k-node graph my first natural hot-path query returned **6339 rows / ~587 KB (~147k tokens)** into an agent's context instead of the 8 rows I asked for. On a bigger graph that is a context-window wipeout from a query that looks completely ordinary.

## Reproduction

Same project, same binary, same DB — only the `ORDER BY` arity changes.

```bash
P=my-project

# A) single-key ORDER BY + LIMIT 5  -> correct
cbm cli query_graph --project $P \
  --query 'MATCH (f:Method) WHERE f.file_path STARTS WITH "src/" RETURN f.name ORDER BY f.complexity DESC LIMIT 5'
# rows: 5      (113 bytes)

# B) two-key ORDER BY + LIMIT 5     -> LIMIT DROPPED
cbm cli query_graph --project $P \
  --query 'MATCH (f:Method) WHERE f.file_path STARTS WITH "src/" RETURN f.name ORDER BY f.complexity DESC, f.name DESC LIMIT 5'
# rows: 6326   (117277 bytes)   <-- expected 5

# C) LIMIT 5, no ORDER BY          -> correct
cbm cli query_graph --project $P \
  --query 'MATCH (f:Method) WHERE f.file_path STARTS WITH "src/" RETURN f.name LIMIT 5'
# rows: 5      (113 bytes)

# D) two-key ORDER BY + LIMIT 5, plus the MCP-level guard -> correct
cbm cli query_graph --project $P --max-rows 5 \
  --query 'MATCH (f:Method) WHERE f.file_path STARTS WITH "src/" RETURN f.name ORDER BY f.complexity DESC, f.name DESC LIMIT 5'
# rows: 5      (113 bytes)
```

| case | `ORDER BY` keys | `LIMIT` | rows returned | bytes |
|------|-----------------|---------|---------------|-------|
| A | 1 | 5 | 5 | 113 |
[...]

PR #2002, opened 2026-09-02
fix(cypher): reject a non-numeric SKIP/LIMIT operand instead of dropping it

### LINKED ISSUE #1994: query_graph: a non-numeric LIMIT/SKIP operand (e.g. `LIMIT $limit`) is silently dropped — the query runs unbounded and reports success
### What happens

`parse_return_or_with()` accepts `SKIP`/`LIMIT` only when the next token is a number. When it is not, the clause is discarded and parsing continues as if it were never written. The query then executes with **no** limit and returns `isError: false`.

Reproduced on `main` (3627eff), built from source, against a 26-node fixture repo:

```
$ cbm cli --json query_graph --project demo --query 'MATCH (n) RETURN n.name LIMIT 1'
rows: 1

$ cbm cli --json query_graph --project demo --query 'MATCH (n) RETURN n.name LIMIT $limit'
rows: 26          # LIMIT silently ignored, isError=false

$ cbm cli --json query_graph --project demo --query 'MATCH (n) RETURN n.name SKIP $offset LIMIT 1'
rows: 26          # the orphaned operand also swallows the LIMIT that follows

$ cbm cli --json query_graph --project demo --query 'MATCH (n) RETURN n.name LIMIT abc'
rows: 26
```

Expected: a parse error naming the offending token. Actual: a successful response containing the entire result set.

`$param` is not an exotic input — it is what a Neo4j-shaped client or an LLM agent writes by default. On a real repository the same query returns every matching row up to the 100k ceiling, or dies with `result exceeded 100k rows — use narrower filters or add LIMIT`, which tells the user to add a LIMIT they did write.

### Root cause

`src/cypher/cypher.c:1821-1834`:

```c
    /* Optional SKIP */
    if (match(p, TOK_SKIP)) {
        const cbm_token_t *num = expect(p, TOK_NUMBER);
        if (num) {
            r->skip = (int)strtol(num->text, NULL, CBM_DECIMAL_BASE);
        }
    }

    /* Optional LIMIT */
    if (match(p, TOK_LIMIT)) {
        const cbm_token_t *num = expect(p, TOK_NUMBER);
        if (num) {
            r->limit = (int)strtol(num->text, NULL, CBM_DECIMAL_BASE);
        }
    }
```

`expect()` (cypher.c:492-500) fills `p->error` and returns NULL, but both call sites only test `if (num)` and fall through to `*out = r; return 0;`. `r->limit` keeps the `-1` initialiser set at cypher.c:1770, which cypher.c:4823 reads as "no LIMIT":

```c
    rb_apply_skip_limit(rb, ret->skip, ret->limit >= 0 ? ret->limit : max_rows);
```

Two further layers hide the failure:

1. `cbm_lex()` skips unknown characters (cypher.c:410-411, `/* Unknown character — skip */ i++;`), so `$` disappears and the operand lexes as an ordinary token.
2. `parse_post_where()` does not require `TOK_EOF` after the RETURN clause, so the orphaned token is dropped as well — which is why `SKIP $offset LIMIT 1` loses both clauses.

Because every sub-parse returned 0, `cbm_parse()` never consults `p.error` and the caller gets `error == NULL`.

### Why this is the failure mode #1334 already banned

`tests/test_cypher.c:410-411` says of the ORDER BY cap fix:

> more keys than the modeled maximum is a loud parse error - the old failure mode (ignore the remainder, drop the LIMIT) must never come back.

Same failure mode, different route: #1334 closed the ORDER BY key-list path; the SKIP/LIMIT operand path was left open.

### Fix

Propagate the `expect()` failure in both branches — `free_return_clause(r); return CBM_NOT_FOUND;` — matching the ORDER BY branch immediately above. `cbm_parse()` then surfaces the existing `expected token type ... got ... at pos N` message.

PR follows with four regression tests: three fail on `main` and pass with the fix, and a control asserting `SKIP 2 LIMIT 10` still parses and still carries 2/10. `scripts/test.sh` is green locally (clean ASan+UBSan build, all suites, contract steps).

Separately, the `(int)strtol(..., NULL, ...)` truncation at the same lines means a literal ≥ 2^31 also lands on a wrong value, but that needs an absurd literal and is not worth its own issue; the operand check above is the reachable defect.

### Environment
- `main` @ 3627eff, built from source with `scripts/build.sh`

metehanulusoy:
Fixes #1994

### The defect

`parse_return_or_with()` accepted `SKIP`/`LIMIT` only when the next token was a number. When `expect(p, TOK_NUMBER)` failed it filled `p->error` and returned NULL, but both call sites only tested `if (num)` and then fell through to `*out = r; return 0;`. `r->limit` kept the `-1` "unset" initialiser, which execution reads as "no LIMIT":

```c
rb_apply_skip_limit(rb, ret->skip, ret->limit >= 0 ? ret->limit : max_rows);
```

So a bounded query ran unbounded and reported success. `SKIP $offset LIMIT 10` lost both clauses at once, because the orphaned operand also blocked the following `match(p, TOK_LIMIT)`.

Measured on the built binary before the fix, against a 26-node fixture:

| query | rows |
|---|---|
| `MATCH (n) RETURN n.name LIMIT 1` | 1 |
| `MATCH (n) RETURN n.name LIMIT $limit` | **26** |
| `MATCH (n) RETURN n.name SKIP $offset LIMIT 1` | **26** |
| `MATCH (n) RETURN n.name LIMIT abc` | **26** |

### The change

Propagate the `expect()` failure in both branches — `free_return_clause(r); return CBM_NOT_FOUND;` — exactly as the `ORDER BY` branch directly above already does. `cbm_parse()` then surfaces the existing `expected token type ... got ... at pos N` message instead of silently continuing.

This is the failure mode #1334 banned ("the old failure mode — ignore the remainder, drop the LIMIT — must never come back"), reached by a different route.

### Tests

Four cases in `tests/test_cypher.c`, next to the #1334 regression:

- `cypher_parse_nonnumeric_limit_rejected_issue1994` — `LIMIT $limit`
- `cypher_parse_nonnumeric_skip_rejected_issue1994` — `SKIP $offset LIMIT 10`, covering the swallowed second clause
- `cypher_parse_word_limit_operand_rejected_issue1994` — `LIMIT abc`
- `cypher_parse_numeric_skip_limit_still_accepted` — control: `SKIP 2 LIMIT 10` still parses and still carries 2/10

The first three fail on `main` and pass with this change; the control passes both ways.

### Verification

- `scripts/test.sh` — green (clean ASan+UBSan build, all suites, contract steps, prod-binary guards)
- `make lint-format` with `clang-format-20` (the version `_lint.yml` pins) — no drift
- `make lint-tidy-diff` — clean
- `cppcheck` — clean
- Branch is on current `main` (17786374) — not stale

One note on how I ran that: `scripts/test.sh` is green when invoked directly, but the `scripts/hooks` pre-commit hook failed five times in a row on `git_context_linked_worktree`. That turned out to be unrelated to this change — git exports `GIT_DIR` into hook environments, `GIT_DIR` overrides `git -C`, and the git-shelling tests in `tests/test_pipeline.c` therefore run against the real repository instead of their temp fixture. Reproduced without any hook: `GIT_DIR=/path/to/repo/.git scripts/test.sh --suites pipeline` fails, plain `scripts/test.sh --suites pipeline` passes. `main` with no changes shows the same thing, so it is pre-existing. I filed it separately as #2003 rather than bundling it here; this commit therefore used `--no-verify`, with the gates run by hand (all listed above).

### Behavior change — worth calling out

This is stricter than `main`: a query whose `SKIP`/`LIMIT` operand is not a number now returns a parse error where it previously returned rows. Anyone who was unknowingly sending `LIMIT $limit` was already getting the wrong answer, so the error is the correct outcome — but it is a visible change, not a pure internal fix, and I would rather flag it than have it surprise you. If you would prefer a warning-and-continue instead of a hard error, say so and I will rework it.

### Scope

One file changed in `src/`, 13 insertions / 6 deletions. No MCP tool signature change, no new dependency, no new `system()`/`popen()`/network call.


DeusData:
Approved on merit. Both CI failures are unrelated to your change — details below — so this needs a rebase and a re-run, not a code change.

## The diagnosis holds

I checked it against `main` rather than taking it:

- `cypher.c:1826-1834` — the `SKIP` and `LIMIT` branches both do `if (num) { ... }` with no `else`, then fall through to `*out = r; return 0;`.
- `cypher.c:4823` — `ret->limit >= 0 ? ret->limit : max_rows`, so the untouched `-1` initialiser reads as "no LIMIT". A bounded query runs unbounded and reports success.
- Your `#1334` reference is exact: `cypher_parse_order_by_over_cap_rejected_issue1334` (`test_cypher.c:410`) already established that this parser's over-cap case must be *a loud parse error*, and its comment names the same banned failure mode.

**The fix is the right shape because it removes an inconsistency rather than adding a rule.** The `ORDER BY` branch immediately above already does `free_return_clause(r); return CBM_NOT_FOUND;`, and so does the `r->count > CBM_SZ_32` bound above that. `SKIP`/`LIMIT` were the two branches that didn't. After this, all four behave the same way, which is a much better state than three-plus-a-special-case.

The measured before-table is what makes this reviewable in one pass — `LIMIT $limit` → 26 rows against a 26-node fixture says "silently unbounded" in a way no prose does. And catching that `SKIP $offset LIMIT 10` loses *both* clauses, because the orphaned operand blocks the following `match(p, TOK_LIMIT)`, is the detail that justifies fixing both branches together.

## The two red checks

**`test / test-msan` — infrastructure.** It fails at the *"Build MSan image (cached layers)"* step and the *"MSan suite"* step is `skipped`, so no test of yours ever ran. This is a known image-build failure on our side.

**`test / test-windows-guards` — not attributable to this diff.** It goes red on `tests/windows/test_daemon_stability.py :: section_start_status_port` ("the daemon did not accept the UI configuration; browser was not opened"), with two `SETUP FAIL` lines and three precondition skips alongside it.

I did look for a mechanism by which stricter `SKIP`/`LIMIT` parsing could break a daemon guard, because that is the one way this could genuinely be yours: if any shipped query used a non-numeric operand, it would now be rejected. It does not. Every `LIMIT ?N` / `LIMIT %d` in the tree is **SQLite** (`store.c`, `mcp.c`), which is a different parser, and the Windows guard suite does not exercise Cypher at all. The same job passed on another PR earlier today, so it is not a standing red either.

**Please rebase** — you are 3 commits behind (`#1703`) — and the re-run should settle both. If windows-guards stays red on a fresh base, tell me and I will take it; it will not be yours to fix.

## Your `GIT_DIR` finding

> git exports `GIT_DIR` into hook environments, `GIT_DIR` overrides `git -C`, and the git-shelling tests in `tests/test_pipeline.c` therefore run against the real repository instead of their temp fixture

That is a real bug and a good one — it means those tests silently assert against whatever repo state the developer happens to have, and it only shows up when something else (your pre-commit hook) sets `GIT_DIR`. **Please open it as its own issue** with the `GIT_DIR=... scripts/test.sh --suites pipeline` reproducer. It deserves a fix on its own terms rather than a footnote here, and running it down while it was blocking you — then correctly concluding it was unrelated — is exactly the right instinct.

One heads-up: #1998 also touches `src/cypher/cypher.c`, in `check_projection_scope` around line 5006. Different region, so no textual conflict, but whichever lands second may want a rebase.

Rebase and I will merge on green.

metehanulusoy:
Rebased onto `5fbab7bb` — clean, no conflicts; upstream's three commits do not touch either file. Cypher suite green on the new base (190 passed, including the four new cases).

On the `GIT_DIR` finding: already filed as #2003, with the `GIT_DIR=... scripts/test.sh --suites pipeline` reproducer and a note that it reproduces on clean `main` too, so it is not something my branch introduced. I linked it from the PR description shortly after opening, so you may have read the earlier version.

Thanks for tracing the two red checks yourself rather than bouncing them back to me.

DeusData:
Rebased and re-running — thank you. You are now 10 commits behind rather than 3 behind a `main` that has moved four times today, which is as current as anything in the queue.

And thank you for filing **#2003** for the `GIT_DIR` finding. That was the right call and I want to be clear it is not a footnote: git exporting `GIT_DIR` into hook environments, where it overrides `git -C`, means the git-shelling tests in `tests/test_pipeline.c` have been asserting against whatever repository the developer happened to be standing in. A green there proves less than it appears to, and that is worse than an outright failure — it is the same "silently wrong rather than loudly broken" shape as the defect this PR fixes. It is now tracked on our side.

My review stands: approved on merit, merging on green.

For the two reds you saw earlier — both were ours and neither was yours. `test-msan` failed at the *image build* step with the suite skipped, and `test-windows-guards` went red on `test_daemon_stability.py`, which does not exercise Cypher at all. I checked the one mechanism by which stricter `SKIP`/`LIMIT` parsing could genuinely have caused a Windows failure — a shipped query with a non-numeric operand — and there is none: every `LIMIT ?N` / `LIMIT %d` in the tree is SQLite, a different parser entirely.

Fair warning that green may take a while. Our Actions pool is servicing roughly one job at a time against 45 queued runs, so the delay is ours, not yours.

DeusData:
Merged — and I owe you an apology. This was **approved on merit on 2026-09-02**, you rebased the same day, and then it sat green for eighteen days. You did everything asked within hours; the delay was entirely ours.

The fix deserves restating because it is more serious than its diff size suggests. A non-numeric `SKIP`/`LIMIT` operand did not fail — `expect()` returned NULL, the clause was quietly dropped, `r->limit` stayed at its `-1` "no LIMIT" sentinel, and the query reported **success**. So a caller asking for a bounded result received the entire result set and had no way to know. That is the same failure mode #1334 banned, arriving by a different route.

The detail that makes it matter in practice is the one you put in the comment: `$limit` is not an exotic input. It is what a Neo4j-shaped client writes by default, so the everyday path was the broken one.

Failing loudly with `CBM_NOT_FOUND` and freeing the clause is the right call — for a query engine, answering the wrong question confidently is worse than refusing to answer.

Thank you as well for filing #2003 for the `GIT_DIR` finding rather than folding it in here. Keeping an unrelated discovery in its own issue is what made this one reviewable at a glance, and it is a habit I wish were more common.

Issue #1153, opened 2026-07-17
method calls through object instances/pointers are dropped in CALLER EDGES


sreehari-malapati-myprod:
### Version

codebase-memory-mcp 0.9.0

### Platform

Linux (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

The dropped calls share a trait: they're method calls through object instances/pointers:

instEntry->processChange() ❌
instEntry->processAdd() ❌
instEntry->setApp() ❌
appEntry->samplerInstEntry() ❌
params->printParams() ❌
The captured calls are mostly free functions / static / uniquely-named:

gSampler.sampleInstEntry(...) ✅ (0.38, unique_name)
createPlatInstEntry(...) ✅ (free function)
insertSampleInstEntry(...) ✅

### Reproduction

codebase-memory-mcp cli index_repository   --repo-path /data/src/repo  --mode fast --name branch --persistence true

HOME=/data/codebase codebase-memory-mcp cli query_graph --project branch   --query "MATCH (caller:Method)-[c:CALLS]->(callee) WHERE caller.qualified_name = 'master.forwarding.aftman.sampler.src.SamplerMsgHandler.SamplerMsgHandler.processAdd' RETURN callee.name, callee.qualified_name, callee.file_path, c.confidence, c.candidates, c.strategy, c.line ORDER BY c.line LIMIT 50"

### Logs

```text

```

### Diagnostics trajectory (memory / performance / leak issues)

```text

```

### Confirmations

- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

DeusData:
Thanks for the focused receiver-call evidence. I removed the incorrect Windows label and placed this in the high-priority 0.9.1-rc query/call-resolution queue under #594. The RED guard should cover C/C++ pointer and instance receivers independently, prove the enclosing Method remains the caller, and distinguish a missing edge from an edge incorrectly attributed to a File or Module node.

sreehari-malapati-myprod:
@DeusData ..

Relates to / extends #1153 (method calls through object instances/pointers dropped in caller edges).

Summary:
[...]

PR #2005, opened 2026-09-02
fix(parsing): recover export-macro-hidden type definitions (#1989)

### LINKED ISSUE #1989: C/C++: export macros (<MODULE>_API / <lib>_EXPORT) between class-key and name extract the macro as the type name; enums and free functions are lost
### Version

codebase-memory-mcp 0.10.8

### Platform

Windows (x64)

### Install channel

GitHub release archive / install.ps1

### Variant

standard

Indexing C/C++ headers that place a build-system **export macro** (`<MODULE>_API`, `<lib>_EXPORT`, …) between the class-key and the type name extracts the **macro token as the type name** instead of the real class name — and in several forms (enums, free functions) the declaration is lost entirely (ERROR region → `parse_partial`).

Expected: `class MYMODULE_API Foo` yields a class definition named `Foo`.

This pattern is generated by every symbol-exporting build system, not a niche convention:

- **Unreal Engine** UBT defines `<MODULE>_API` empty on the compiler command line (`/D "MODULE_API="`) — every UE plugin header uses it
- **CMake** `generate_export_header()` emits `<lib>_EXPORT` used identically
- Verified public code: [`spdlog/include/spdlog/logger.h:50`](https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/logger.h#L50) `class SPDLOG_API logger`; [`fmt/include/fmt/os.h:222`](https://github.com/fmtlib/fmt/blob/master/include/fmt/os.h#L222) `class FMT_API file`; [`protobuf/src/google/protobuf/descriptor.h`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.h) `class PROTOBUF_EXPORT LazyDescriptor`

### Reproduction

Self-contained dummy snippet mirroring UE5 plugin conventions (reflection macros are harmless here — the breaking token is the `DUMMYPLUGIN_API` export macro):

```cpp
// ue_style.h — minimal UE-style header
#pragma once
#include "CoreMinimal.h"

UCLASS(BlueprintType)
class DUMMYPLUGIN_API UDummyPoolSubsystem : public UTickableWorldSubsystem
{
    GENERATED_BODY()
public:
    UFUNCTION(BlueprintCallable, Category = "Dummy|Pool")
    AActor* SpawnActorFromPool(UClass* ActorClass, const FTransform& SpawnTransform);
private:
    UPROPERTY()
    TMap<UClass*, TArray<TObjectPtr<AActor>>> Pools;
};

UINTERFACE(MinimalAPI, BlueprintType)
class UDummyPoolable : public UInterface
{
    GENERATED_BODY()
};

class IDummyPoolable
{
    GENERATED_BODY()
public:
    virtual void OnAcquiredFromPool(AActor* Actor) = 0;
};

UCLASS()
class ADummySpawner : public AActor
{
    GENERATED_BODY()
public:
    ADummySpawner();
    UPROPERTY(EditAnywhere, Category = "Dummy")
    int32 PoolSize = 16;
};

struct DUMMYPLUGIN_API FDummyCfg { int32 X; };
enum class EDummyMode : uint8 { A, B };
```

1. Code: the snippet above (save as `ue_style.h` in an empty dir; the `UCLASS`/`GENERATED_BODY` tokens can be left as-is — they do not affect the result)
2. Command: `codebase-memory-mcp cli index_repository '{"repo_path":"<dir>"}'` then `cli search_graph '{"project":"...","name_pattern":".*Dummy.*"}'`
3. Actual extraction of declarations (probed directly against tree-sitter-cpp, same grammar family as the vendored `grammar_cpp.c`):

| Declaration | Extracted |
|---|---|
| `class Foo { ... }` (no macro) | ✅ Class `Foo` |
| `UCLASS()` then `class Foo : ...` | ✅ Class `Foo` (the `UCLASS()` call itself is a harmless ERROR) |
| `class DUMMYPLUGIN_API UDummyPoolSubsystem : ...` | ❌ Class named **`DUMMYPLUGIN_API`**; real name swallowed into an ERROR region |
| `struct DUMMYPLUGIN_API FDummyCfg { ... }` | ⚠️ Struct named **`DUMMYPLUGIN_API`** — silent wrong name, no error flag |
| `enum class EDummyMode : uint8 { A, B };` (isolated) | ✅ Enum `EDummyMode` |
| `DUMMYPLUGIN_API void DoThing();` | ❌ Function lost (ERROR) |

AST evidence for the class case:

```
function_definition
  class_specifier
    class
    type_identifier    # 'DUMMYPLUGIN_API'   <- macro taken as the name
  ERROR                <- real name 'UDummyPoolSubsystem' lands here
    identifier
    :  public
```

`UINTERFACE`-style blocks cascade worse: one parse error makes a single ERROR region swallow **every following top-level declaration** in the header (verified: `UDummyPoolable`/`IDummyPoolable`/`ADummySpawner` all vanish into one `field_declaration` fragment).

In the real CBM graph (indexed a large UE5 plugin, v0.10.8) this surfaces as: `UObjectPoolSubsystem` appears only as a `Function`-label node named after the class, while the `Class` label has no entry for it — so `query_graph("MATCH (c:Class) WHERE c.name CONTAINS 'Subsystem'")` returns 0 rows and `trace_path` inbound misses real callers (adjacent open issues #694/#763).

### Scale

Real-world UE5 plugin repo (731 C++ files): **230/364 headers** carry `<MOD>_API` on a class/struct/enum declaration; **365 files** flagged `parse_partial`. The same shape exists in spdlog/fmt/protobuf and any CMake `generate_export_header` project, so this is a general C/C++ extraction gap.

### Root cause

tree-sitter parses without preprocessor state, so `MODULE_API` (empty on the real compile command line) is an ordinary identifier → `class_specifier.name = MODULE_API`. The existing simplecpp second pass (`cbm.c:~1397`) already re-parses C/C++ with `extra_defines`, but its defs rescue (#961) only adopts defs that intersect raw ERROR regions and whose name line exists in raw source — the raw pass already supplied the *wrong* name (`class_specifier` itself parses "successfully"), so nothing is corrected.

### Proposed fix (direction — will implement per your preference)

The machinery already exists; this is assembly, not new infrastructure:

1. **Synthesized empty defines (recommended)**: for C/C++, collect candidate export macros from the raw source (identifier shape `^[A-Z][A-Z0-9_]*_(API|EXPORT|IMPORT|DLLEXPORT|DEPRECATED)$`, capped), add them as empty `NAME=` entries to `extra_defines` for `cbm_preprocess_with_map`, then let the existing second pass + #961 rescue adopt corrected defs (line remap via `original_line_by_expanded_line` already exists). Locally verified with tree-sitter-cpp: expanding `-DDUMMYPLUGIN_API= -DMYMODULE_API=` flips extraction to `UDummyPoolSubsystem` / `FDummyCfg` correctly.
2. **Plumb user/project defines**: accept defines from config (e.g. `CBM_EXTRA_DEFINES`) or project metadata (UE `*.Build.cs` `PublicDefinitions`, CMake export headers) — complements (1), generally useful, zero heuristics.
3. **Extractor-level fallback (cheapest)**: when `class_specifier.name` matches the export-macro shape and an ERROR region holds the following `identifier`, adopt that identifier as the name. Doesn't fix enums/functions.

(1) fixes class/struct/enum/function loss in one place and reuses battle-tested code; (2) is the deterministic path for compile_commands-less projects; (3) is a minimal stopgap. No vendored-grammar changes.

### Questions

- Is the `^[A-Z0-9_]*_(API|EXPORT|IMPORT|DLLEXPORT|DEPRECATED)$` shape acceptable as the candidate-macro filter, or would you prefer an explicit config-driven list only (option 2)?
- Since this changes what gets extracted (per CONTRIBUTING), may I open a focused PR for (1)+(2)? Would keep it <500 lines with tests in `tests/test_pipeline.c` / `tests/test_extraction.c` and a spdlog-derived fixture as the shareable test bed.

### Confirmations

- I searched existing issues and this is not a duplicate. Closest: #592 (tracking), #694/#763 (C/C++ CALLS attribution — adjacent, distinct root cause), #946 (generic C++ parse bug).
- My reproduction uses shareable code (self-contained dummy snippet + public OSS repos spdlog / fmt / protobuf).

XIYBHK:
## What does this PR do?

Fixes #1989.

Recover C/C++ definitions hidden by conventional build-system export macros through the existing preprocessing second pass. Candidate collection is lexically bounded to 32 names per file; explicit caller defines retain priority. Recovered definitions pass line mapping, original-source visibility, and qualified-name reconciliation gates. The change covers class/struct/enum definitions, free functions, and inline methods, with negative controls for ordinary all-caps identifiers.

Rebased onto `main` at `92abefa3`. The implementation preserves the upstream error-region `dropped` counter initialization and uses `cbm_calloc`/`cbm_free` with `CBM_MEM_CLASS_EXTRACT` for its temporary superseded-definition bitmap. All 19 original regression tests remain registered. The PR contains one signed-off commit, `fcf451c6`, touching only the original four files. Project-level macro configuration remains in follow-up #2049.

## Validation

- `git diff --check`, diff-scoped formatting checks, and strict GCC/G++ syntax checks passed for the changed translation units.
- `scripts/security-audit.sh` passed.
- Local runtime tests have not passed: after supplying zlib, the unsanitized MinGW GCC runner exited with heap corruption in the first existing extraction test, before the export-macro regressions ran. No local sanitizer validation is claimed.
- The maintainer confirmed the current upstream lint/test blockers are being addressed in #2257 and will update this branch and run full CI plus local extraction tests after that fix lands. This PR does not include the unrelated request-handler cleanup or relax any lint baseline.

Known boundaries: one-character export prefixes such as `X_API` remain intentionally rejected; header-only prototypes retain the existing extraction behavior; UINTERFACE cascade handling remains outside this PR.

### DIFF
--- a/internal/cbm/cbm.c
+++ b/internal/cbm/cbm.c
@@ -1029,6 +1029,63 @@ static bool cbm_span_contains_callable_def(const char *src, int src_len, uint32_
     return false;
 }
 
+/* #1989: type-def counterpart of cbm_span_contains_callable_def. A rescued
+ * class/struct/enum definition has its name followed by a base clause (':') or
+ * a body ('{'), not a parameter list, so the callable validator would reject
+ * every type def pulled from the expanded tree. */
+static bool cbm_span_contains_type_def(const char *src, int src_len, uint32_t start_line,
+                                       uint32_t end_line, const char *name) {
+    if (!src || src_len <= 0 || !name || !name[0] || start_line == 0 || end_line < start_line) {
+        return false;
+    }
+    int span_start = 0;
+    uint32_t line = 1;
+    while (span_start < src_len && line < start_line) {
+        if (src[span_start++] == '\n') {
+            line++;
+        }
+    }
+    if (line != start_line) {
+        return false;
+    }
+    int span_end = span_start;
+    while (span_end < src_len && line <= end_line) {
+        if (src[span_end++] == '\n') {
+            line++;
+        }
+    }
+    size_t name_len = strlen(name);
+    for (int pos = span_start; pos + (int)name_len <= span_end; pos++) {
+        if (strncmp(src + pos, name, name_len) != 0 ||
+            (pos > 0 && cbm_identifier_char(src[pos - 1])) ||
+            (pos + (int)name_len < src_len && cbm_identifier_char(src[pos + name_len]))) {
+            continue;
+        }
+        int open = pos + (int)name_len;
+        while (open < span_end && isspace((unsigned char)src[open])) {
+            open++;
+        }
+        // Base clause (": public B"), body ("{"), or template args ("<...").
+        if (open < span_end && (src[open] == ':' || src[open] == '{' || src[open] == '<')) {
+            return true;
+        }
+    }
+    return false;
+}
+
+/* #1989: label buckets for the misparse-correction gates. A raw def extracted
+ * from a broken `class MOD_API Foo` shape lands under a callable label while
+ * the expanded tree yields a proper type label for the same name. */
+static bool cbm_def_label_is_type(const char *label) {
+    return label != NULL && (strcmp(label, "Class") == 0 || strcmp(label, "Enum") == 0 ||
+                             strcmp(label, "Interface") == 0 || strcmp(label, "Type") == 0 ||
+                             strcmp(label, "Struct") == 0 || strcmp(label, "Union") == 0);
+}
+
+static bool cbm_def_label_is_callable(const char *label) {
+    return label != NULL && (strcmp(label, "Function") == 0 || strcmp(label, "Method") == 0);
+}
+
 /* Remap an expanded-source definition back to the original input file. Every
  * line in the definition must be attributable to the main file; generated
  * macro bodies, included headers, and ambiguous spans fail closed. */
@@ -1423,6 +1480,14 @@ static CBMFileResult *extract_file_ex_body(const char *source, int source_len, C
     // Defs keep original-source line numbers; only CALLS are extracted from expanded source.
     if (language == CBM_LANG_C || language == CBM_LANG_CPP || language == CBM_LANG_CUDA) {
         uint64_t pp_start = now_ns();
+        /* #1989: collect build-system export-macro candidates from the raw
+         * source. They are predefined empty in the expanded buffer (see
+         * preprocessor.cpp) so `class MOD_API Foo` parses with its real name,
+         * and the rescue loop below can adopt corrected type defs. Bounded:
+         * at most CBM_EXPORT_MACRO_MAX names per file. */
+        char export_cands[CBM_EXPORT_MACRO_MAX][CBM_EXPORT_MACRO_NAME_MAX];
+        int export_cand_count =
+            cbm_export_macro_candidates(source, source_len, export_cands, CBM_EXPORT_MACRO_MAX);
         CBMPreprocessedSource *preprocessed = cbm_preprocess_with_map(
             source, source_len, rel_path, extra_defines, include_paths, language != CBM_LANG_C);
         if (preprocessed && preprocessed->source) {
@@ -1503,17 +1568,51 @@ static CBMFileResult *extract_file_ex_body(const char *source, int source_len, C
                      * recover defs from it — adopting ONLY those that
                      * intersect a raw ERROR region, whose name is visible on
                      * the raw source line, and whose QN the raw pass did not
-                     * already extract. */
-                    if (ts_node_has_error(root)) {
+                     * already extract.
+                     *
+                     * #1989: build-system export macros (<MOD>_API,
+                     * <lib>_EXPORT, ...) are empty on the real compile line but
+                     * opaque to tree-sitter, so `class MOD_API Foo` misparses
+                     * with the macro as the type name — often with NO raw ERROR
+                     * region at all (the class_specifier "succeeds" with the
+                     * wrong name; enums and free functions do error out). When
+                     * export-macro candidates were collected for this file, also
+                     * adopt expanded defs that replace a raw def literally named
+                     * as a candidate, or that correct a raw misparse label (the
+                     * raw pass extracted the same name as Function/Method from
+                     * the broken function_definition shape; the expanded tree
+                     * yields a proper type def). Superseded raw defs are
+                     * dropped so the corrected definition is not shadowed by
+                     * the qualified-name dedup. With no candidates the block
+                     * reduces to the original #961 behavior. */
+                    if (ts_node_has_error(root) || export_cand_count > 0) {
                         cbm_error_regions_t raw_regs = {{0}, {0}, 0};
                         cbm_collect_error_regions(root, &raw_regs, source, source_len);
-                        if (raw_regs.count > 0) {
+                        if (raw_regs.count > 0 || export_cand_count > 0) {
                             int defs_before = result->defs.count;
                             cbm_extract_definitions(&pp_ctx);
                             int w = defs_before;
+                            char *superseded = NULL;
+                            if (export_cand_count > 0 && defs_before > 0) {
+                                superseded = (char *)calloc((size_t)defs_before, 1);
+                            }
+                            /* #1989: spans of defs already adopted from the
+                             * expanded tree. A def nested inside one of these
+                             * (an inline method of a rescued class) was dropped
+                             * by the raw pass — the misparsed
+                             * `class MOD_API Foo {...}` parsed as a
+                             * function_definition, and nested defs in its body
+                             * are never walked. Bounded: one span per adopted
+                             * def, capped at 64. */
+                            struct {
+                                uint32_t start;
+                                uint32_t end;
+                            } rescued_spans[64];
+                            int rescued_count = 0;
                             for (int i = defs_before; i < result->defs.count; i++) {
                                 CBMDefinition *d = &result->defs.items[i];
                                 bool adopt = false;
+                                int supersede_j = -1;
                                 if (cbm_remap_preprocessed_def(d, preprocessed)) {
                                     for (int rj = 0; rj < raw_regs.count && !adopt; rj++) {
                                         if (d->start_line <= raw_regs.ends[rj] &&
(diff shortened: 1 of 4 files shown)


DeusData:
Holding note, because you are owed one before any review verdict.

On #1989 you asked two direct questions:

> Is the `^[A-Z0-9_]*_(API|EXPORT|IMPORT|DLLEXPORT|DEPRECATED)$` shape acceptable as the candidate-macro filter, or would you prefer an explicit config-driven list only (option 2)?
>
> Since this changes what gets extracted (per CONTRIBUTING), may I open a focused PR for (1)+(2)?

Nobody answered. You did the right thing by asking, waited, and then built it anyway rather than letting the report rot — but this PR's description calls it "the agreed option 1", and I want to be straight with you: **it was never agreed, because we never replied.** That is our failure, not a liberty you took.

I have put the direction question — heuristic shape versus config-driven defines versus both — in front of the maintainer now, and I will bring you a real answer rather than letting it sit again.

Two things I can say already, having read the diff rather than the summary:

**The report is exemplary.** Naming UBT's `/D "MODULE_API="` and CMake's `generate_export_header()` as the mechanism, then pinning it to three public headers — spdlog `logger.h:50`, fmt `os.h:222`, protobuf `descriptor.h` — turns "my code doesn't index" into a defect anyone can reproduce in a minute. Enumerating three fixes with costs, and asking which we wanted, is more than most maintainers get.

**And the implementation is careful in the places that are easy to get wrong.** `skip_raw_string` handles the fact that `"` is a legal d-char, so `R"""(...)"""` does not derail the scan; `skip_non_code` handles line-spliced `//` comments, where a backslash before the newline continues the comment; the `>= slen + 2` prefix rule keeps a single-letter `X_API` out; and the uncertain paths return -1 and stop collecting rather than guessing. Failing toward *not* injecting a define is the right direction for a heuristic that changes what gets extracted.

None of that decides the direction question, which is genuinely the maintainer's. But it does mean that if the answer is yes, this is in good shape.

Thank you for the patience, and sorry it took a PR to get a reply.

XIYBHK:
Thanks for the clarification. I may be interpreting the timeline differently, so I checked the issue history.

Before opening this PR, I asked whether to pursue the heuristic option 1 or the config-driven option 2, and whether to combine them. The reply on #1989 explicitly asked me to "open a focused PR for option 1 only" and specified the candidate cap, negative controls, no user-configured define plumbing, and the class/struct/enum/free-function test boundary: https://github.com/DeusData/codebase-memory-mcp/issues/1989#issuecomment-5500705435

I used that as the scope for this PR. I agree that the exact regex was not approved character-for-character; the implementation made the candidate filter deliberately conservative and bounded. If the exact filter or any other part of the scope should change, I am happy to adjust it.

I just wanted to clarify that the option-1 direction and scope were discussed in the issue before implementation.

DeusData:
You are right, and thank you for correcting the record.

The maintainer reply on #1989 did explicitly ask for a focused PR implementing option 1 with the bounded candidate scan and the stated negative controls. Our earlier comment here saying that the direction had never been agreed was wrong. I am sorry: you followed the scope we gave you.

The option-1 direction therefore remains the approved scope. The exact candidate filter and implementation still receive normal review on their merits, but you do not need to re-litigate why this PR exists. Thank you for responding with the thread reference and keeping the correction factual.

XIYBHK:
Rebased onto the latest main (2f9828d6) and resolved the conflict in tests/test_extraction.c by preserving both the upstream extraction tests and the #1989 regression coverage. The PR remains a single commit and still changes only the four intended files.

The rebased files pass single-file compilation and clang-format checks locally. The full sanitizer test runner cannot be linked in the local MinGW environment because its sanitizer runtime/spec file is unavailable; the refreshed CI matrix is running now.

DeusData:
Direction settled — thank you for building this, and apologies that the two questions on #1989 went unanswered before you did. That was on us, and the answer was yes.

The zero-config shape rule (option 1) is accepted as the default: the candidate collection is bounded and lexically careful, the recovered definitions go through the line-mapping / visibility / qualified-name gates before they replace anything, and explicitly configured defines keep priority. That is the right layer and the right amount of caution.

Two things so you know what happens next:

1. **Merge sequencing.** Your branch is green and cleared. Three other PRs that also append to `tests/test_extraction.c` (#1976, #1721, #1730) are mid-way through their CI right now; landing yours first would void all three matrices in a very loaded queue, so I will let them land and then bring this branch up to date once and merge on green. Nothing needed from you — if the branch update needs a hand I will say so here.

2. **Follow-up, not a blocker:** an explicit override for the rule (extend the candidate set for macros the shape rule cannot see, exclude a name that is a genuine symbol) — the reporter's option 2 layered on top of yours. Filed as #2049; you are the natural person to take it if you want to, no timing pressure.

DeusData:
Thank you for rebasing onto today's `main` — and please stop there for a moment, because you are about to spend effort on a problem that is ours, not yours.

**The `lint / lint` red you are trying to get past is `main`'s own.** The memory-core linter counts raw `free()` sites in `src/mcp/mcp.c` against a ratcheted baseline, and a merge on our side on Saturday pushed that file three over (`814 -> 817`). Every PR based on `main` fails that check right now, whatever it touches. Your second commit, `6eb89924` ("share index request cleanup on early exits"), is a reasonable way to claw the count back — but:

- **it is already being fixed**, in #2257, which gives `handle_index_repository` a single release helper for its three argument strings (817 -> 785) and is in its last CI legs as I write this;
- **your commit conflicts with it** — both reshape the same early-return blocks, so once #2257 lands your branch would go `CONFLICTING` on a file your PR has no reason to touch;
- and it folds the cross-repo-mode branch into the policy-error branch, which changes *when* `handle_cross_repo_mode` runs relative to the policy load. That may well be harmless, but it is a behaviour question in a request handler, and it should not ride along inside a C/C++ parsing fix where nobody will think to look for it.

So: **please drop `6eb89924`** and keep the PR at the single commit `fcf451c6` and the four files it was always meant to touch (`git reset --hard fcf451c6 && git push --force-with-lease` — force-pushing your own PR branch is completely fine here). Do not worry about the lint red in the meantime; it will not be held against this PR.

**What happens next, and it is all on our side:** #2257 lands (it also fixes a second `main` red, in the test stage, that was hidden behind the lint failure). Then I bring your branch up to date with `main`, which gives it a fresh, full CI run; I build and run the extraction suites on the merge result locally; and it merges. The direction was settled on 4 September and nothing about that has changed — the sequencing I described then simply took far longer than it should have, and I am sorry for that.

Thank you for your patience, and for keeping the PR current through a very busy stretch on `main`.

XIYBHK:
Thank you for the clear explanation and for coordinating the fix in #2257.

I've dropped `6eb89924` and restored the branch to `fcf451c6`. The PR is back to one signed-off commit and the original four files, and I've updated the description accordingly. Sorry for adding an unrelated cleanup while trying to get past the CI failure.

I'll leave the branch as-is and let you handle the update and validation after #2257 lands. Thank you again for reviewing this and keeping us informed.

DeusData:
Merged as c7488b03 — thank you, and thank you for your patience through a sequencing promise that took far longer than it should have.

Verified on the merge result before landing: extraction 369/0 with all the #1989 tests green; c_lsp, pipeline, lang_contract, parse_coverage, grammar_regression and complexity green; with production reverted the tests no longer compile, so they bind to `cbm_export_macro_candidates`; memory-core linter clean; no raw allocator or `fopen` in the added lines.

A nice coincidence worth telling you about: on the morning of the merge a reporter on #1153 ("method calls through object pointers are dropped", priority/high) independently traced that bug to exactly this defect — `class MYMODULE_EXPORT Worker {` being read as a function. I ran their example through `main` before and after your change: `Worker` goes from `Function` back to `Class` and the missing `CALLS` edge returns, identical to the no-macro control. So this fix closes more than the issue it was written for.

The explicit override layered on top of the shape rule (the reporter's option 2 on #1989) remains open as a follow-up, as discussed.

PR #2006, opened 2026-09-02
fix(pipeline): key the Go Field guard on the recorded selector shape, not reference text

### LINKED ISSUE #1962: fix(pipeline): cbm_go_suppress_bare_field_ref's dot-check is inert — the Go guard is a blanket veto, and its census is tautological
## Summary

`cbm_go_suppress_bare_field_ref` (merged in #1944, `20e86c02`) drops a reference edge when the target label is `Field`, the file is Go, and `strchr(ref_name, '.') == NULL`. **The dot test can never be false for Go**, so the predicate is a blanket veto on Field-targeted references rather than the precise guard it documents.

Found during post-merge review of the #1937 → #1940 → #1944 stack.

## Why the dot can never appear

Both call paths strip the receiver before the resolver ever sees the name.

**READS/WRITES** — `resolve_lhs_write_name` (`internal/cbm/extract_semantic.c`) documents its own contract:

```
//   - field/member/selector access (`self.total = ...`, `obj.Field = ...` →
//     write the trailing field name `total`/`Field`)
```

So `t.err = x` yields `ref_name == "err"`, never `"t.err"`.

**USAGE** — `is_reference_node` (`internal/cbm/extract_usages.c`) never treats `selector_expression` as a reference kind; the only occurrence of that node type in the file is a *parent* check. The reference is the inner `field_identifier`/`identifier`, whose own span is bare.

## Consequences

1. **The unit test pins a dead branch.** `ASSERT_FALSE(..., "t.err", "Field")` asserts "selector-shaped refs may bind", but Go's extractor cannot produce that input. It is not a production guarantee.
2. **The census that justified the change is a tautology.** "100% of Field-targeted USAGE edges carried dot-less reference text, so there is no legitimate population mixed in" is guaranteed by the extractor, not measured from the data. It cannot distinguish noise from signal, which is the job the predicate was given.
3. **Go `Field` nodes are reference-unreachable.** The ~4588 `Field` nodes recovered by #1940 receive no `USAGE`/`READS`/`WRITES` edges at all, including genuine `t.err = x` writes and `cfg.Timeout` reads. Only `CALLS` can reach them.

## Not a regression

`main` had zero Go `Field` nodes before #1940, so nothing that previously worked is lost. The accurate description of the merged stack is "adds ~4588 nodes that reference edges cannot reach", not "removes edges".

## Cheap falsification

Index a Go repository on current `main` and count `Field`-targeted `USAGE`/`READS`/`WRITES` edges. The prediction is **zero**. A non-zero count disproves this issue.

## The signal exists but is discarded

The distinction the predicate wants — `field_identifier` (the field half of a selector) versus `identifier` (a bare local) — is computed at extraction time and thrown away before the resolver runs. Threading it through would make the guard precise instead of total, and would let genuine `x.f` references bind while still refusing bare locals.

## Suggested resolution

Either thread the `field_identifier` signal through to the resolver, or keep the blanket veto and make the docstring and unit test say so honestly — the current pair asserts precision the code does not have.

ilyabrykau-orca:
## Summary

Fixes the defect you found in post-merge review of #1944: `cbm_go_suppress_bare_field_ref`'s `strchr(ref_name, '.')` test is inert because the extractor strips the receiver on every path that reaches the resolver — `resolve_lhs_write_name` records the trailing field name of a selector LHS, and `is_reference_node` records the inner `field_identifier`. The guard was therefore a blanket veto: every Go `Field` node was unreachable by `USAGE`/`READS`/`WRITES`, genuine selector references included.

As #1962 says, the signal the guard wants is computed at extraction time and discarded. This PR records it and threads it through:

- `CBMUsage` / `CBMReadWrite` gain `is_member_access`. The usage recorder sets it when the reference node is a `field_identifier` (the member half of a selector); `resolve_lhs_write_name` reports it through an out-param when it takes the field/member LHS branch. The receiver is stripped either way — the flag is the only surviving record of selector shape.
- `cbm_go_suppress_bare_field_ref(is_go, is_member_access, target_label)` now refuses a `Field` bind only for references that were never the member half of a selector. All four resolver sites (sequential + parallel, USAGE + READS/WRITES) pass the recorded flag. Still Go-gated — the C#/Java/C++/Python bare-member shapes stay untouched (`cp_reads_writes_cs_static_field` still pins that).
- The dead `ASSERT_FALSE(..., "t.err", "Field")` unit case — an input Go's extractor cannot produce — is replaced by flag-based cases, so the unit test now pins reachable behavior.

## RED / GREEN (reproduce-first)

Extended the #1942 fixtures with a sibling-file method doing genuine selector references (`t.err = nil`, `return t.n`). With the blanket veto restored, both resolver twins go RED at the new asserts:

```
FAIL tests/test_pipeline.c:4913: ASSERT(cross_file_edge_exists(s, project, "Reset", "err", "WRITES"))
FAIL tests/test_pipeline.c:4948: ASSERT(cross_file_edge_exists(s, project, "Reset", "err", "WRITES"))
```

With the fix, both twins are GREEN and the original bare-local negatives (`Run`'s `err := …` must not bind the field) still hold. Full `scripts/test.sh` (ASan+UBSan) green.

## Field census (same repo as the #1937/#1940/#1944 numbers)

Your cheap falsification, run on current `main`: **zero** Field-targeted `USAGE`/`READS`/`WRITES` edges (4497 Go `Field` nodes, reachable only by CALLS/DEFINES/TESTS) — the prediction was exact.

With this fix, same repo:

| | main | this PR |
|---|---|---|
| `USAGE` → Go `Field` | 0 | **8503** |
| `WRITES` → Go `Field` | 0 | **1368** |
| distinct `Field` nodes reached | 0 | **2220** of 4497 |

For scale: the pre-#1944 bare-local pollution was 21308 USAGE / 5191 WRITES. The recovered population is the genuine selector subset, not a return of the noise — the bare-local negatives in the fixtures and the flag's default-false both hold that line.

Fixes #1962.

### DIFF
--- a/internal/cbm/cbm.h
+++ b/internal/cbm/cbm.h
@@ -292,6 +292,10 @@ typedef struct {
     uint32_t site_start_byte;             // exact reference-token span; end > start when present
     uint32_t site_end_byte;               // exclusive byte offset in the source file
     CBMSourceOrigin source_origin;        // raw source or C-family preprocessed buffer
+    bool is_member_access;                // token is the member half of a selector/attribute
+                                          // (Go x.f — field_identifier). The extractor strips
+                                          // the receiver, so this is the only surviving record
+                                          // of selector shape (#1962). Default false.
 } CBMUsage;
 
 typedef struct {
@@ -303,6 +307,9 @@ typedef struct {
     const char *var_name;          // variable name
     const char *enclosing_func_qn; // QN of enclosing function
     bool is_write;                 // true = write, false = read
+    bool is_member_access;         // var_name is the field half of a selector/member LHS
+                                   // (`t.err = x` → "err"); the receiver is stripped here,
+                                   // so this is the only record of selector shape (#1962)
 } CBMReadWrite;
 
 typedef struct {
--- a/internal/cbm/extract_semantic.c
+++ b/internal/cbm/extract_semantic.c
@@ -146,7 +146,10 @@ static void walk_throws(CBMExtractCtx *ctx, TSNode root, const CBMLangSpec *spec
 //   - field/member/selector access (`self.total = ...`, `obj.Field = ...` →
 //     write the trailing field name `total`/`Field`)
 // Returns NULL if no simple write target can be determined.
-static char *resolve_lhs_write_name(CBMExtractCtx *ctx, TSNode left) {
+// Sets *is_member_out when the written name is the field half of a
+// member/selector LHS — the receiver is stripped from the returned name, so
+// this out-param is the only surviving record of the selector shape (#1962).
+static char *resolve_lhs_write_name(CBMExtractCtx *ctx, TSNode left, bool *is_member_out) {
     // Unwrap a single-element expression_list (Go).
     if (strcmp(ts_node_type(left), "expression_list") == 0) {
         if (ts_node_named_child_count(left) != 1) {
@@ -184,6 +187,9 @@ static char *resolve_lhs_write_name(CBMExtractCtx *ctx, TSNode left) {
             fld = ts_node_child_by_field_name(left, TS_FIELD("name"));
         }
         if (!ts_node_is_null(fld)) {
+            if (is_member_out) {
+                *is_member_out = true;
+            }
             return cbm_node_text(ctx->arena, fld, ctx->source);
         }
         return NULL;
@@ -251,12 +257,14 @@ static void try_emit_assignment_write(CBMExtractCtx *ctx, TSNode node, const cha
     if (ts_node_is_null(left)) {
         return;
     }
-    char *name = resolve_lhs_write_name(ctx, left);
+    bool is_member = false;
+    char *name = resolve_lhs_write_name(ctx, left, &is_member);
     if (name && name[0] && !cbm_is_keyword(name, ctx->language)) {
         CBMReadWrite rw;
         rw.var_name = name;
         rw.is_write = true;
         rw.enclosing_func_qn = func_qn;
+        rw.is_member_access = is_member;
         cbm_rw_push(&ctx->result->rw, ctx->arena, rw);
     }
 }
@@ -329,12 +337,14 @@ void handle_readwrites(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec,
         TSNode left = resolve_write_lhs_node(node);
 
         if (!ts_node_is_null(left)) {
-            char *name = resolve_lhs_write_name(ctx, left);
+            bool is_member = false;
+            char *name = resolve_lhs_write_name(ctx, left, &is_member);
             if (name && name[0] && !cbm_is_keyword(name, ctx->language)) {
                 CBMReadWrite rw;
                 rw.var_name = name;
                 rw.is_write = true;
                 rw.enclosing_func_qn = state->enclosing_func_qn;
+                rw.is_member_access = is_member;
                 cbm_rw_push(&ctx->result->rw, ctx->arena, rw);
             }
         }
--- a/internal/cbm/extract_usages.c
+++ b/internal/cbm/extract_usages.c
@@ -2673,6 +2673,10 @@ void handle_usages(CBMExtractCtx *ctx, TSNode node, const CBMLangSpec *spec, Wal
         usage.ref_name = name;
         usage.enclosing_func_qn = state->enclosing_func_qn;
         usage.lexical_scope_id = usage_lexical_scope_id_for_node(ctx, state, node);
+        /* The member half of a selector is its own reference node
+         * (field_identifier); record that shape — the name alone cannot carry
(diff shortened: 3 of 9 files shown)


DeusData:
Approved. I verified the inertness claim against `main` rather than taking it, and it holds exactly:

- `extract_semantic.c:181-187` — `resolve_lhs_write_name` matches `selector_expression`/`field_access`, takes the `field` (or `name`) child, and returns `cbm_node_text` of **that node alone**. `t.err = nil` records `err`.
- `extract_usages.c:113/117/121` — `is_reference_node` accepts `field_identifier`, the member half of the selector. Same result: `err`.

So `strchr(ref_name, '.')` is unconditionally NULL on both paths that reach the guard, `cbm_go_suppress_bare_field_ref` returns true for every `Field` target, and no Go struct field was reachable by `USAGE`/`READS`/`WRITES` at all. Your census is the confirming measurement: **zero** such edges across 4497 Go `Field` nodes on current `main`.

## The dead test is the real lesson here

`tests/test_registry.c:848`:

```c
/* A selector-shaped reference may bind a field. */
ASSERT_FALSE(cbm_go_suppress_bare_field_ref(true, "t.err", "Field"));
```

That assertion passes. It has always passed. And it is worthless, because `"t.err"` is an input the extractor **cannot produce** — the receiver is stripped before the resolver ever sees the name. The guard had a unit test covering precisely the case that mattered, the test was green, and the behaviour it described never occurred in production.

That is why this survived #1942 and #1944. A test written against a hypothetical input gives exactly the confidence of a test, and none of the coverage. Replacing it with flag-based cases that pin *reachable* behaviour is the most valuable change in this PR, more than the edges it recovers.

It is also the second time in two days this repo has produced this shape — the other was a suite whose `RUN_TEST`s sat in a decoy suite that never ran. Worth remembering as a class: **when a guard has a passing test and is provably inert, suspect the test's inputs before the guard's logic.**

## On the fix itself

**Recording the signal instead of reconstructing it is the right call.** #1962 identified that selector shape is known at extraction time and thrown away; the alternative — reconstructing "was this a selector?" downstream from a name with the receiver already gone — is not recoverable at all. There is no clever resolver-side fix here, which is what makes threading the flag the honest answer rather than the lazy one.

**A distinct `is_member_access` rather than overloading an existing flag** matches the precedent set when `callee_is_locally_bound` was kept separate from `is_method`: `is_method` has a documented contract and other readers, and widening it to mean two things corrupts them silently.

**Still Go-gated, with all four sites updated.** Sequential and parallel, `USAGE` and `READS`/`WRITES` — the twin-resolver split is the standing trap in this area and you covered both, with `cp_reads_writes_cs_static_field` still pinning that C#/Java/C++/Python bare-member references are untouched.

I also confirmed `CBMUsage`/`CBMReadWrite` are in-memory extraction structs — neither appears in `sqlite_writer.c` or `artifact.c` — so the two new fields need **no index-format bump**. Worth stating explicitly in the description, since a reader seeing struct growth will ask.

## Before merge

CI is still queued. Once it is green I will merge — this recovers ~10k edges that were being silently discarded, which is squarely graph-quality-first.

Thank you for taking the post-merge finding and coming back with the falsification run first. Reporting the census on `main` as **zero**, and calling it "your cheap falsification", is the part that made this reviewable in one pass: you proved the bug existed before proving your fix removed it.

DeusData:
Merged as `192f2960`. This recovers roughly **9,900 edges** that were being silently discarded — 8,503 `USAGE` and 1,368 `WRITES` — and it is the most consequential graph-quality fix I have merged today.

The finding I keep coming back to is not the edges, though, it is *why this survived two prior PRs*. The guard had a unit test for exactly this case:

```c
ASSERT_FALSE(cbm_go_suppress_bare_field_ref(true, "t.err", "Field"));
```

Green, always — and worthless, because `"t.err"` is an input the extractor cannot produce. `resolve_lhs_write_name` returns the text of the field node alone, and `is_reference_node` records the inner `field_identifier`; both strip the receiver before the resolver ever sees the name. So `strchr(ref_name, '.')` was unconditionally NULL and the guard was a blanket veto over every Go `Field` node — 4,497 of them, reachable by nothing.

Replacing that assertion with flag-based cases that pin *reachable* behaviour is worth more than the edges, because it is what stops the next version of this bug. I have written the pattern down on our side: when a guard is provably inert, suspect the test's inputs before the guard's logic, and give any guard whose effect should be selective a census — `0` or `all` is the tell. Your "cheap falsification" run was exactly that census, and you did it before writing the fix.

Recording the signal at extraction time rather than reconstructing it downstream was the right call too: once the receiver is stripped, "was this a selector?" is not recoverable, so there was no clever resolver-side alternative to find.

Before merging I confirmed the 29-commit gap to `main` was inert — no commit had touched any of your nine files since your base — so the green described exactly what landed.

PR #2016, opened 2026-09-02
feat(ui): show serving binary version (#1820)

### LINKED ISSUE #1820: UI: show the serving binary version in the header
## What problem does this solve?

The browser management UI does not show which codebase-memory-mcp binary version is serving it. That makes support reports and upgrade verification harder, especially when a long-lived daemon or multiple install channels may leave uncertainty about the active binary. The frontend package version is not sufficient because it can differ from the release version of the server embedding it.

## Proposed solution

- Add the serving binary's compiled `CBM_VERSION` to the existing `GET /api/ui-config` response.
- Show a compact read-only `vX.Y.Z` label beside the Codebase Memory name in the UI header.
- Hide the label when the config request fails or returns no valid version, preserving compatibility and avoiding a new error state.
- Cover the backend response with the native HTTP test and the visible label with a focused component test.

This reuses an existing same-origin endpoint and requires no new MCP method, external request, dependency, or configuration key.

## Alternatives considered

- Use the version from `graph-ui/package.json`: this identifies the frontend package, not necessarily the binary serving it.
- Ask users to run `codebase-memory-mcp --version`: useful in a terminal, but it does not identify the server behind the currently open browser tab.
- Add a dedicated status endpoint: unnecessary for one small read-only field already suited to `/api/ui-config`.

## Confirmations

- [x] I searched existing issues and pull requests and this is not a duplicate.

rarepops:
## What does this PR do?

Fixes #1820.

- adds the compiled serving binary version to the existing `/api/ui-config` response
- shows a compact `vX.Y.Z` label beside the Codebase Memory name
- hides the label when the endpoint fails or returns no string version
- adds focused frontend coverage for visible, invalid, and unavailable states
- adds native endpoint coverage for the compiled version value

This uses the existing same-origin config endpoint and adds no external request, dependency, MCP method, or configuration key.

## Validation

- `cd graph-ui && npm test` (46 passed)
- `cd graph-ui && npm run build`
- `./build/c/test-runner httpd` under GCC ASan/UBSan (64 passed, 1 platform skip)
- `make -f Makefile.cbm lint-format CLANG_FORMAT=clang-format-20`
- `git diff --check`

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) (full native suite not run; focused UI and HTTP suites passed)
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) (full lint not run; pinned formatting gate passed)

## Screenshot

Header showing the serving binary version badge

### DIFF
--- a/graph-ui/src/App.test.tsx
+++ b/graph-ui/src/App.test.tsx
@@ -1,0 +1,59 @@
+/* @vitest-environment jsdom */
+import "@testing-library/jest-dom/vitest";
+import { cleanup, render, screen, waitFor } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { App } from "./App";
+import { messages } from "./lib/i18n";
+
+vi.mock("./components/GraphTab", () => ({ GraphTab: () => null }));
+vi.mock("./components/StatsTab", () => ({ StatsTab: () => null }));
+vi.mock("./components/ControlTab", () => ({ ControlTab: () => null }));
+vi.mock("./lib/i18n", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("./lib/i18n")>();
+  return { ...actual, useUiMessages: () => messages.en };
+});
+
+describe("App", () => {
+  afterEach(() => {
+    cleanup();
+    vi.unstubAllGlobals();
+    window.history.replaceState(null, "", "/");
+  });
+
+  it("shows the serving binary version", async () => {
+    vi.stubGlobal("fetch", vi.fn(async () =>
+      new Response(JSON.stringify({ lang: "en", version: "0.10.8" }), {
+        status: 200,
+        headers: { "Content-Type": "application/json" },
+      }),
+    ));
+
+    render(<App />);
+
+    expect(await screen.findByText("v0.10.8")).toBeVisible();
+  });
+
+  it("hides the version when the config has no string version", async () => {
+    const fetchMock = vi.fn(async () =>
+      new Response(JSON.stringify({ lang: "en", version: 108 }), { status: 200 }),
+    );
+    vi.stubGlobal("fetch", fetchMock);
+
+    render(<App />);
+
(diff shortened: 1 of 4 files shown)


DeusData:
Approved on merit. Small, well-scoped, and it follows the existing conventions rather than inventing any.

**The `CBM_VERSION` fallback is exactly right.** I checked before flagging it, and

```c
#ifndef CBM_VERSION
#define CBM_VERSION "dev"
#endif
```

is verbatim the pattern already in `src/main.c:103`, `src/mcp/index_supervisor.c:36` and `src/cli/cli.c:99`, with CI injecting `-DCBM_VERSION` (`Makefile.cbm:74`). So a build without the flag degrades the same way everywhere else does, instead of introducing a fourth convention.

**Extending `/api/ui-config` rather than adding an endpoint** is the right call too. It is same-origin, already fetched during load, and your description names what is *not* changing — no external request, no dependency, no MCP method, no config key. Stating the negative space is more useful than describing the positive change, because it is what a reviewer would otherwise have to go and check.

**Handling the three display states separately** — visible, invalid, unavailable — is what makes this safe on the frontend. Hiding the label when the endpoint fails or returns no string means a version that is missing never renders as `vundefined`, and each state has its own test.

On exposure: the version is already public via `--version` and the UI is loopback-only, so surfacing it on that endpoint discloses nothing new. Worth having considered, not worth blocking.

## One note

Your validation shows `make lint-format CLANG_FORMAT=clang-format-20`. Our gate requires the **Homebrew LLVM** build, and the two disagree on some large files — usually reporting phantom drift, though I have just seen the reverse on another PR, where `clang-format-20` passed locally and CI's pinned formatter found real violations in newly-added lines. So treat CI's `lint / lint` as authoritative here rather than your local pass. Nothing to do unless it goes red.

CI is still early (3 of 11 reported). I will merge on green — our Actions queue has been heavily backlogged today, though it is finally draining.

Issue #2003, opened 2026-09-02
pre-commit hook fails the test suite: git exports GIT_DIR into hooks, which overrides `git -C` in tests/test_pipeline.c


metehanulusoy:
### What happens

`CONTRIBUTING.md` tells contributors to install the repo hooks (`git config core.hooksPath scripts/hooks`). The pre-commit hook runs the full test suite. Git exports `GIT_DIR` into hook environments, `GIT_DIR` overrides `git -C`, and the tests in `tests/test_pipeline.c` that shell out to git therefore operate on the **real repository** instead of their temp fixture.

The result: with the documented hook installed, `git commit` fails the suite every time.

### Reproduction — no hook needed, one variable

```
$ scripts/test.sh --suites pipeline
  264 passed

$ GIT_DIR="$PWD/.git" scripts/test.sh --suites pipeline
  git_context_linked_worktree   FAIL tests/test_pipeline.c:7444: run_cmd(cmd) == 32768, expected 0 == 0
  261 passed, 3 failed
```

Verified on `main` with no local changes (3627eff), macOS 15 arm64, git 2.50.1.

Line 7444 is:

```c
snprintf(cmd, sizeof(cmd), "git -C \"%s\" checkout -b main >%s 2>&1", repo, null_dev);
ASSERT_EQ(run_cmd(cmd), 0);
```

With `GIT_DIR` set, `-C` no longer selects the repository: the command runs against the real checkout, where `main` already exists, so git exits 128.

### It is not only a failing assert — the suite writes to the real repository

While tracing this I ran the test's own `worktree add` line with `GIT_DIR` set:

```
$ GIT_DIR=/path/to/codebase-memory-mcp/.git \
    git -C "$TMP/repo with space" worktree add -b feat2 "$TMP/wt with space"
Preparing worktree (new branch 'feat2')
Updating files: 100% (2118/2118), done.
HEAD is now at 17786374 Merge pull request #1993 from DeusData/fix/unify-type-short-name-index
```

It checked out 2118 files of the actual repository into the temp path and created a branch in the contributor's clone. So a contributor who follows the documented setup gets branches and worktree registrations created in their working repo by the test suite. `git worktree prune` plus `git branch -D` cleans it up, but nothing warns them it happened.

### Why it is easy to miss

[...]

PR #2019, opened 2026-09-03
test(cli): stop the registry plan check depending on the developer's machine

CaptainMittens:
## The problem

`cli_agent_client_registry_routes_plan_install_and_uninstall` builds a fixture with its own `HOME` and `PATH`, then asserts the whole install plan contains no path with `/plugins/` in it.

Agent detection does not stay inside `HOME` and `PATH`. `cbm_find_cli` also looks in `/usr/local/bin` and, on macOS, `/opt/homebrew/bin`. A developer with OpenCode installed there has it detected inside the fixture, and OpenCode ships a real plugin file — so the plan legitimately carries a `/plugins/` path and the assertion fails. On a machine without OpenCode the same assertion passes, which is why CI has never seen it.

## The change

Name the two directories the check actually meant — the fixture's Qoder and Pi directories, neither of which has a plugin directory, so a planned path under one would be invented:

```c
char qoder_plugins[700];
char pi_plugins[700];
snprintf(qoder_plugins, sizeof(qoder_plugins), "%s/plugins/", qoder_dir);
snprintf(pi_plugins, sizeof(pi_plugins), "%s/plugins/", pi_dir);
```

The check now tests the same thing regardless of what the developer has installed.

One file, `tests/test_cli.c`, 12 insertions and 1 deletion. **No production behaviour changes.** Searching well-known install directories is deliberate, so that install finds an agent whose command is not on the current `PATH` — that is not what this touches.

## Evidence

Run on a machine that has OpenCode installed, so the trigger condition is present.

Before, at `c38aa35c`:

```
FAIL tests/test_cli.c:6725: CLI install/plan/uninstall must route the agent-client
registry, preserve foreign entries, and keep Pi free of invented MCP configuration
  cli_registry_installs_kimi_rovo_amp_durable_context
  Detected agents: OpenCode Kimi Code CLI Rovo Dev CLI Amp
```

After, on this branch:

```
cli_agent_client_registry_routes_plan_install_and_uninstall
  Detected agents: OpenCode Qoder CLI Roo Code Amazon Q Developer IDE Pi Sourcegraph Cody
```

OpenCode is detected in the fixture in both runs, so the condition that provokes the failure is present either way. The assertion fires in the first and not in the second.

## One thing worth stating plainly

The `cli` suite has other failures on this machine that this change does not address and does not claim to. They vary between runs — the agent-client install and uninstall assertions cannot drain a cohort while a permanent daemon is running, so the set differs each time. `origin/main` itself reports 536 passed / 6 failed here, and three of those six print the same `Detected agents: OpenCode` line, at `tests/test_cli.c:9390` and `:9491` alongside the one fixed here.

So this fixes one instance of a wider pattern rather than the whole pattern. I did not extend it to the other two, because the right fix for each depends on what its own assertion means and I did not want to guess that on your behalf. Happy to follow up on those if the approach here is the one you want.

https://claude.ai/code/session_01ERsNp8UNLaixL4uUz7hRTX

### DIFF
--- a/tests/test_cli.c
+++ b/tests/test_cli.c
@@ -6561,8 +6561,19 @@ TEST(cli_agent_client_registry_routes_plan_install_and_uninstall) {
     char *plan = cbm_build_install_plan_json(tmpdir, binary_path);
     yyjson_doc *plan_doc = plan ? yyjson_read(plan, strlen(plan), 0) : NULL;
     yyjson_val *plan_root = plan_doc ? yyjson_doc_get_root(plan_doc) : NULL;
+    /* Neither of these agents has a plugin directory, so a planned path under
+     * one would be invented. Name the two directories rather than searching the
+     * whole plan for "/plugins/": OpenCode does ship a real plugin file, and
+     * agent detection finds a command in /usr/local/bin or /opt/homebrew/bin
+     * whatever HOME and PATH say, so a blanket search passes or fails according
+     * to what the developer happens to have installed. */
+    char qoder_plugins[700];
+    char pi_plugins[700];
+    snprintf(qoder_plugins, sizeof(qoder_plugins), "%s/plugins/", qoder_dir);
+    snprintf(pi_plugins, sizeof(pi_plugins), "%s/plugins/", pi_dir);
     bool plan_ok =
-        plan && !strstr(plan, "/plugins/") && !strstr(plan, "plugin_files") &&
+        plan && !strstr(plan, qoder_plugins) && !strstr(plan, pi_plugins) &&
+        !strstr(plan, "plugin_files") &&
         test_json_string_array_contains(plan_root, "config_files_planned", qoder_settings) &&
         test_json_string_array_contains(plan_root, "config_files_planned", amazon_config) &&
         test_json_string_array_contains(plan_root, "config_files_planned", roo_config) &&


DeusData:
Approved. I verified the mechanism rather than taking it: `src/cli/cli.c:977` and `:982` hardcode `/usr/local/bin/%s` and `/opt/homebrew/bin/%s`, so agent detection genuinely escapes the fixture's `HOME` and `PATH`, exactly as you describe.

**This is the more valuable half of the fix**, and worth stating plainly: the assertion's verdict depended on **what the developer happened to have installed**. A machine with OpenCode in `/opt/homebrew/bin` fails; one without passes. A test like that does not report on the code — it reports on the workstation, and a green from it proves nothing.

That is the second finding of this shape we have seen in as many days. #2003 (filed by another contributor) is the same class from the other direction: `git` exports `GIT_DIR` into hook environments, where it overrides `git -C`, so the git-shelling tests in `test_pipeline.c` assert against whatever repository the developer is standing in. Both are tests whose result is a function of ambient environment rather than of the change under test.

**And you fixed it the right way round.** The tempting repair is to harden the fixture — sandbox `PATH` harder, stub the search directories — which fights the production code and tends to drift. Instead you narrowed the assertion to what it actually meant: the fixture's own Qoder and Pi directories, neither of which has a plugin directory, so a planned path under either would be invented. The check now tests one thing, deterministically, and stays true as detection changes.

## Your one red is ours

`test / test-msan` fails at the **"Build MSan image (cached layers)"** step, with the "MSan suite" step `skipped` — so no test of yours ever ran. That is a known image-build failure on our side, unrelated to a twelve-line test change.

Merging once the remaining checks report. Thank you — this is the sort of fix that pays for itself the next time someone trusts a green.

DeusData:
Merged as `bac8edc5` — thank you!

This is the second machine-dependent assertion you have caught this week, and the fix is the right way round again: the test now names the two fixture directories it actually cares about instead of searching the whole plan for `/plugins/`, so a developer with OpenCode in `/opt/homebrew/bin` and one without get the same verdict. Exactly the kind of test that stops the suite lying to whoever runs it next.

Small note for the record, not a change request: the retained `!strstr(plan, "plugin_files")` clause guards a key the plan builder never emits, so it is inert — harmless, just not load-bearing.

PR #2040, opened 2026-09-03
perf(ui): make the graph UI usable on very large projects (435k nodes / 4.2M edges)

### LINKED ISSUE #2039: Graph UI exhausts browser memory and layout takes minutes on very large projects (435k nodes / 4.2M edges)
### Version / platform
`codebase-memory-mcp dev` built from `main` @ 5802ccdd (also reproduces on 0.9.0). macOS 25.6 arm64, Chrome (default 4 GB V8 heap limit), 24 GB RAM.

Displaying a very large project in the graph UI (435k nodes / 4.2M edges, a private Go monorepo; `c1.db` is 1.4 GB) exhausts browser memory and takes minutes per layout request. I expected memory to stay roughly proportional to what is rendered and to stay flat when adjusting display settings.

Measured against the full graph (`max_nodes=435000`):

| Stage | JS heap used | Renderer footprint | GPU process footprint |
|---|---|---|---|
| After render | 855 MB | ~1.0 GB | ~0.4 GB |
| After 14 edge-brightness slider changes | 3949 MB | 3.4 GB | 4.2 GB |
| 25 s idle afterwards | 3954 MB | no change | no change |
| Tab closed | | | 172 MB |

Each slider change added ~400 MB and nothing was reclaimed. One more change would have hit the V8 heap limit. At a 150k budget the same test plateaued (GC keeps up because the per-change garbage is smaller), which is why this only shows on very large projects.

Server side, `GET /api/layout?project=c1&max_nodes=N`:

| N | time | daemon peak RSS | JSON | edges returned |
|---|---|---|---|---|
| 5 000 (default) | 7.3 s | 461 MB | 1.8 MB | 538 |
| 50 000 | 10 s | 478 MB | 22 MB | 60k |
| 150 000 | 20 s | 948 MB | 84 MB | 518k |
| 435 000 | 490 s | 3.3 GB (+1 GB compressed) | 382 MB | 4.2M |

### Root causes (verified in source and by measurement)

**Client**
1. `graph-ui/src/components/EdgeLines.tsx` `useMemo` (deps `nodes, edges, highlightedIds, targetNodes, brightness`) allocates a new `THREE.BufferGeometry` with two `Float32Array(edges.length*6)` (≈203 MB at 4.2M edges) on every change and never calls `geometry.dispose()`. react-three-fiber does not dispose a replaced `geometry` prop. three.js keeps a strong reference to every attribute of an undisposed geometry in `WebGLBindingStates` and only `gl.deleteBuffer`s from the dispose handler, so both CPU arrays and GPU buffers accumulate. The only `dispose()` in `graph-ui/src` is for label textures.
2. There is a 75k point-mode threshold for nodes (`NodeCloud.tsx`) but no cap, sampling, or LOD for edges.
3. `useGraphData.ts` buffers the whole response, concatenates the chunks, decodes and `JSON.parse`s it (several transient copies of a 382 MB payload); 4.2M edge objects each carry their own `type` string instance. `GraphTab.tsx` re-filters all nodes/edges and rebuilds id `Set`s on every checkbox; `NodeDetailPanel.tsx` rebuilds a 435k-entry `Map` and scans all edges on every node click.

**Server**
4. `compute_call_depth` (`src/ui/layout3d.c:447-454`) rescans every edge for every dequeued BFS node: O(n·e). A `sample` of the daemon during the 435k request put 100 % of the HTTP thread's time in this loop. This is the 490 s.
5. `cbm_layout_compute` loads every edge of every type project-wide via `cbm_store_find_edges_by_type` (no `LIMIT`, and the unused `properties` column is fetched and strdup'd) and only then filters by the node sample. The node sample itself is `ORDER BY n.name, n.id LIMIT max_nodes` (`src/store/store.c:5308`), i.e. alphabetical, so the default 5 000-node view of this project has only 538 edges.
6. `handle_layout` (`src/ui/http_server.c:1579-1741`) serializes the layout, re-parses it with `yyjson_read`, deep-copies it with `yyjson_doc_mut_copy`, re-serializes, then `cbm_http_replyf("%s")` copies it once more (`httpd.c:770-786`). The payload exists ~4× concurrently. `layout3d.h` says positions are cached in SQLite; no such cache exists.

### Reproduction
Any project with >1M edges. Open `/?tab=graph&project=<name>`, set the node budget to the full node count, wait for render, then drag the edge-brightness slider a dozen times while watching `performance.memory.usedJSHeapSize` and the GPU process footprint.

### Related
#726 (layout drops connection for large graphs), #498 (browser freezes >10K nodes), #593 (large-repo performance task).

### Proposed fix
PR to follow: allocate edge/node geometry once and update attributes in place with disposal on cleanup, brightness as a material multiplier; an edge render budget; interning + fewer payload copies; adjacency index for the BFS; edge SQL scoped to the sampled ids without `properties`; degree-ordered node sampling; single JSON write straight to the reply buffer.

tthayer:
## What does this PR do?

Fixes #2039.

Makes the graph UI usable on very large projects (tested on a 435k-node / 4.2M-edge index). The browser side stopped leaking GPU and JS memory on every display-setting change, and the server side no longer takes minutes and gigabytes to lay out the graph. One commit per fix so they can be reviewed or split independently:

**Client (graph-ui)**
1. `fix(graph-ui)`: edge and node geometry is allocated once, refilled in place, and disposed on cleanup; edge brightness is applied through the material instead of rewriting 4.2M vertex colors. This is the fix for the leak in #2039.
2. `feat(graph-ui)`: deterministic edge render budget with a "Rendering X of Y edges" notice, mirroring the existing node budget.
3. `perf(graph-ui)`: fewer copies of the layout payload, interned edge/node strings, one shared adjacency index instead of per-click rebuilds, partial top-k for labels.

**Server**
4. `perf(ui)`: call-depth BFS uses a CSR adjacency index (O(n+e) instead of O(n·e)). This was 100 % of the 490 s.
5. `perf(ui)`: edge query is scoped to the sampled node ids and no longer fetches `properties`; node sample is ordered by degree so small budgets produce a connected subgraph.
6. `perf(ui)`: layout JSON is built once and handed straight to the reply buffer, removing the parse → deep-copy → rewrite → printf round trip.

### Measurements

Same machine, same `c1.db`, before = `main` @ 5802ccdd, after = this branch.

`GET /api/layout?project=c1&max_nodes=N`

| N | before | after |
|---|---|---|
| 5 000 (default) | 7.3 s, 461 MB peak, 538 edges (1.8 MB) | 3.7 s, 145 MB peak, 386k edges (22 MB) |
| 150 000 | 20 s, 948 MB peak, 518k edges (84 MB) | 13.5 s, 1.18 GB peak, 2.96M edges (208 MB) |
| 435 000 (all) | 490 s, 3.3 GB peak, 4.2M edges (382 MB) | 31 s, 2.2 GB peak, 4.2M edges (382 MB) |

The degree-ordered sample returns far more edges for the same node budget (that is the point: the old alphabetical sample was almost edge-less), so payloads for small budgets grow. Peak server memory at the full budget is still dominated by the 4.2M-edge JSON document itself; a streaming writer would be the next step and is out of scope here.

Browser, full 435k budget, 14 edge-brightness slider changes:

| | before | after |
|---|---|---|
| Time to first render (incl. server) | ~9 min | 66 s |
| JS heap after render | 855 MB | 639 MB |
| JS heap after 14 changes | 3949 MB (limit 4192), no recovery after 25 s idle | 642 MB (flat) |
| Renderer footprint after 14 changes | 3.4 GB | 0.9 GB |
| GPU process footprint after 14 changes | 4.2 GB | 1.35 GB (browser-wide GPU helper; baseline 0.5 GB) |

The HUD now reads "Rendering 1,000,000 of 4,226,606 edges (render budget)"; node click, detail-panel connections, and edge-type filter toggles were smoke-tested at the 150k budget with no console errors.

### Notes for reviewers
- Two intentional response changes in `/api/layout`: `nodes[]` is now ordered by degree (desc, then name, id) for every request, not alphabetically; and `linked_projects` is always present (possibly `[]`) instead of being omitted when there are none. Edge `source`/`target` still index into the same response's `nodes[]`.
- The layout header comment claiming positions are cached in SQLite was stale; there is no cache. Adding one is a reasonable follow-up but out of scope here.
- `cbm_store_search` ordering is unchanged (it is a paginated public API); degree ordering is a separate opt-in query used only by the layout.
- Edge geometry capacity only grows within a mounted graph; buffers sized for the largest edge count seen stay allocated until the graph unmounts.
- `scripts/test.sh` (ASan+UBSan, 141 suites) passed except one pre-existing flaky test (`index_second_inprocess_run_survives_issue773`) that also fails intermittently on the merge base. `clang-format`, `cppcheck`, and diff-scoped `clang-tidy` are clean on the changed files; note `src/ui/*.c` is not in the Makefile's lint file list.
- This bundles six fixes for one issue because they are all needed for the large-project case to work; happy to split into per-commit PRs if preferred.

## Checklist

### DIFF
--- a/graph-ui/src/components/EdgeLines.test.ts
+++ b/graph-ui/src/components/EdgeLines.test.ts
@@ -1,0 +1,130 @@
+/* @vitest-environment jsdom */
+import { describe, expect, it, vi } from "vitest";
+import { renderHook } from "@testing-library/react";
+import * as THREE from "three";
+import {
+  createEdgeBuffers,
+  fillEdgeBuffers,
+  useEdgeGeometry,
+} from "./EdgeLines";
+import type { GraphEdge, GraphNode } from "../lib/types";
+
+function node(id: number, x = 0): GraphNode {
+  return { id, x, y: 0, z: 0, label: "Function", name: `n${id}`, size: 1, color: "#fff" };
+}
+
+const NODES: GraphNode[] = [node(1, 0), node(2, 10), node(3, 20)];
+const EDGES: GraphEdge[] = [
+  { source: 1, target: 2, type: "CALLS" },
+  { source: 2, target: 3, type: "IMPORTS" },
+];
+
+describe("createEdgeBuffers / fillEdgeBuffers", () => {
+  it("allocates position/color attributes sized to capacity", () => {
+    const buf = createEdgeBuffers(2);
+    expect(buf.geometry.getAttribute("position").array.length).toBe(2 * 6);
+    expect(buf.geometry.getAttribute("color").array.length).toBe(2 * 6);
+    expect(buf.capacity).toBe(2);
+  });
+
+  it("fills positions/colors and sets the draw range to the valid edge count", () => {
+    const buf = createEdgeBuffers(EDGES.length);
+    const validCount = fillEdgeBuffers(buf, NODES, EDGES, null, undefined);
+    expect(validCount).toBe(2);
+    expect(buf.geometry.drawRange).toEqual({ start: 0, count: 4 });
+
+    const positions = buf.positionAttr.array as Float32Array;
+    /* First edge: node 1 (x=0) -> node 2 (x=10) */
+    expect(positions[0]).toBe(0);
+    expect(positions[3]).toBe(10);
+  });
+
+  it("skips edges with a dangling endpoint and still fills the valid ones", () => {
+    const buf = createEdgeBuffers(3);
+    const edgesWithDangling: GraphEdge[] = [
+      { source: 1, target: 999, type: "CALLS" },
+      ...EDGES,
+    ];
+    const validCount = fillEdgeBuffers(buf, NODES, edgesWithDangling, null, undefined);
+    expect(validCount).toBe(2);
+  });
+
+  it("refills the SAME TypedArray in place — no reallocation for equal capacity", () => {
+    const buf = createEdgeBuffers(EDGES.length);
+    const positionsArray = buf.positionAttr.array;
+    const colorsArray = buf.colorAttr.array;
+    fillEdgeBuffers(buf, NODES, EDGES, null, undefined);
+    fillEdgeBuffers(buf, NODES, EDGES, new Set([1, 2]), undefined);
+    expect(buf.positionAttr.array).toBe(positionsArray);
+    expect(buf.colorAttr.array).toBe(colorsArray);
+  });
+});
+
+describe("useEdgeGeometry", () => {
+  it("reuses the same geometry instance across re-renders when capacity doesn't grow (e.g. a brightness-only change)", () => {
+    const { result, rerender } = renderHook(
+      ({ highlightedIds }: { highlightedIds: Set<number> | null }) =>
+        useEdgeGeometry(NODES, EDGES, highlightedIds, undefined),
+      { initialProps: { highlightedIds: null as Set<number> | null } },
+    );
+
+    const first = result.current;
+    expect(first).toBeInstanceOf(THREE.BufferGeometry);
+
+    /* Simulate a brightness slider tick: brightness is applied to the
+     * material, not the geometry, so it is not even a hook input — but we
+     * also verify that re-rendering with the SAME edges/nodes/highlight
+     * (as happens on every brightness change) keeps the same geometry. */
+    rerender({ highlightedIds: null });
+    expect(result.current).toBe(first);
+
+    /* A highlight change refills in place too (same edge count → same
+     * capacity), so it must not allocate a new geometry either. */
+    rerender({ highlightedIds: new Set([1]) });
+    expect(result.current).toBe(first);
+  });
+
+  it("disposes the geometry when the capacity grows (new geometry replaces the old, disposed one)", () => {
+    const { result, rerender } = renderHook(
+      ({ edges }: { edges: GraphEdge[] }) =>
+        useEdgeGeometry(NODES, edges, null, undefined),
+      { initialProps: { edges: EDGES } },
+    );
+
+    const first = result.current;
+    const disposeSpy = vi.spyOn(first, "dispose");
+
+    const biggerEdges: GraphEdge[] = [
+      ...EDGES,
+      { source: 3, target: 1, type: "CALLS" },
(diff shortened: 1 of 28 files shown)


DeusData:
Thank you for this, @tthayer — genuinely one of the best-diagnosed first contributions this repository has received. The root-cause section of #2039 is more precise than most maintainer triage: naming the `WebGLBindingStates` retention behind the `EdgeLines` geometry leak, the exact `useMemo` dependency list, and profiling the 490 s down to `compute_call_depth`. We reviewed the branch properly — built it, ran the C and graph-ui suites, and reproduced the server-side gains on a synthetic 200k-node / 1M-edge store: the scoped edge fetch is properly indexed (0.013 s for a 5k-node sample instead of streaming the whole edge set), and that together with the CSR BFS is where the 490 s → 31 s comes from. The work is sound.

We are closing it anyway, and the reason is timing rather than quality: the graph UI is being rebuilt from the ground up. The current 3D view and its `/api/layout` feed are being superseded by a new version designed around exactly the problems you hit — level of detail decided on the server rather than a client-side budget, and a payload that no longer scales with the edge set. Merging a large change into the surface that is about to be replaced would give it a very short life and put you through a rebase-and-rework cycle for no lasting gain, and that felt like the wrong way to treat this level of effort.

What carries forward is your analysis. Connected (degree-ordered) sampling instead of an alphabetical slice, scoping the edge fetch to the sampled ids, the CSR adjacency for call depth, and dropping the serialize → parse → re-serialize round-trip in `handle_layout` are all going into the new design as requirements, with #2039 as the reference.

Two small things we found during review, in case you carry any of this code elsewhere: the defensive `else` branch in the new edge-mapping loop (`layout3d.c`, around line 644) frees `type`/`project`/`properties_json` and then `free_edge_array` frees them again — a `memset` of the struct after the frees, as the sibling branch already does, closes it; and `buildGraphIndex` allocates two `Connection` objects per edge on the unsampled set and is rebuilt on every filter toggle, which at 4.2M edges is the same allocation-churn shape the PR fixes elsewhere.

Thank you again — and please keep the reports coming. A reporter who profiles this carefully is worth a great deal to a project at this scale.

Issue #1954, opened 2026-08-31
uninstall: one symlinked agent config aborts the whole run — executable and indexes (386 MB) left behind


Jmeg8r:
### Version

0.10.8

### Platform

macOS (Apple Silicon)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

## Summary

`uninstall` aborts entirely when **any one** agent config path is a file
symlink. A single unwritable Cursor path (`~/.cursor/mcp.json`, symlinked into
a dotfiles repo) stopped the run before executable and index removal started,
leaving 386 MB on disk after a reported "uninstall".

The refusal to write through a symlink looks deliberate and defensible. The
problem is the blast radius: a Cursor-specific config check gates teardown
steps that have nothing to do with Cursor.

## Observed

```
Found 2 index(es):
  /Users/<me>/.cache/codebase-memory-mcp/hr-moderate.db
  /Users/<me>/.cache/codebase-memory-mcp/_config.db
Delete these indexes? (y/n): y (auto)
Stopping active CBM sessions and operations for uninstall...
error: agent_config agent=Cursor op=mcp_uninstall path=/Users/<me>/.cursor/mcp.json (target: symlink, 43 bytes)
error: one or more agent cleanup operations failed; executable and index removal were not started
error: activation stopped after one or more agent configuration or cleanup operations failed; the published/current executable was kept, and configuration changes that completed may remain. Please restart your coding-agent sessions after resolving the errors above.
```

`~/.cursor/mcp.json` is `lrwxr-xr-x ... -> ~/.dotfiles/cursor/mcp.json` — a
normal dotfiles arrangement, present long before CBM was installed.

## Left behind

| item | size |
|---|---|
| `~/.local/bin/codebase-memory-mcp` | 282 MB |
| `~/.cache/codebase-memory-mcp/` | 104 MB |
| `~/.local/bin/install.sh` (updater) | — |
| `~/.zshrc` PATH line | — |

All four had to be removed by hand. Note this install used `--skip-config`, so
there was no CBM entry in the Cursor file at all — the run aborted over a file
it had never written to and had nothing to remove from.

## Expected

Teardown of CBM-owned artifacts (executable, indexes, updater) should not be
gated on agent-config cleanup. Suggested, in rough priority order:

1. **Decouple.** Remove the executable and indexes even when some agent config
   cleanup fails, then report the skipped configs and exit non-zero. The user
[...]

PR #2041, opened 2026-09-03
test(cli): point PATH at the fixture so uninstall isolation holds on any machine

CaptainMittens:
## What this changes

One test moved `HOME` and `CBM_CACHE_DIR` into a temporary directory but left `PATH` alone.
This points `PATH` at the same fixture and restores it afterwards, using the
`save_test_env` / `restore_test_env` pair that the Copilot and VS Code tests in this file
already use. Ten added lines, one test, no production code.

## Why the test failed

Agent detection reads `PATH` before anything else, so a real agent binary on the
developer's machine is found even though `HOME` points at an empty fixture. `uninstall`
then tries to edit that agent's config file inside the fixture, fails because the fixture
never created one, and stops before removing the binary and the index — which is the thing
`cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index` measures.

On my machine the agent is Goose, at `/Users/<me>/go/bin/goose`:

```
error: agent_config agent=Goose op=mcp_uninstall path=/tmp/cli-daemon-uninstall-7gY0EY/.config/goose/config.yaml (target: does not exist or cannot be inspected)
error: one or more agent cleanup operations failed; executable and index removal were not started
```

The test then fails on `ASSERT_EQ(rc, 0)` with `rc == 1`.

## Evidence

Same tree, same built `test-runner`, only `PATH` differs:

| Run | Result |
|:--|:--|
| `goose` on `PATH`, before this change | `291 passed, 1 failed` |
| `goose` removed from `PATH`, before this change | `292 passed, 0 failed` |
| `goose` on `PATH`, after this change | `292 passed, 0 failed` |

The third row is the one that matters: the suite is green while `goose` is still installed
and still on `PATH`, so the fix is the redirect rather than a quieter environment.

## This does not fix the underlying defect

The abort it exposed is real, and I reported it on #1954, which already tracks the same
abort reached through a symlinked Cursor config. One agent config that cannot be edited
stops teardown for every other agent and for the executable and indexes. That wants a
product decision I did not want to make for you — treat "nothing to remove" as success at
the two call sites, or keep it an error that no longer gates the executable and index
removal — so this PR only stops the test suite from depending on what the developer has
installed.

## Scope I deliberately left alone

Fourteen tests call `cli_activation_save_env`. Twelve leave `PATH` open the same way; only
`cli_install_config_failure_keeps_published_binary` and
`cli_update_already_current_does_not_quiesce_sessions` redirect it today.

I changed only the one that demonstrably fails. The other eleven pass on this machine, and
redirecting `PATH` in a passing test changes what it exercises, so I did not want to do that
without a failure to point at. Happy to send that as a follow-up if you would rather have
the whole class covered at once.

https://claude.ai/code/session_01ERsNp8UNLaixL4uUz7hRTX

### DIFF
--- a/tests/test_cli.c
+++ b/tests/test_cli.c
@@ -2033,6 +2033,15 @@ TEST(cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index) {
     char *old_cache = NULL;
     cli_activation_save_env(&old_home, &old_cache);
     cbm_setenv("HOME", tmpdir, 1);
+    /* PATH has to move with HOME. Agent detection asks cbm_find_cli, which
+     * reads PATH before anything else, so a real agent binary on the developer's
+     * PATH is found even though HOME points at this fixture. Uninstall then
+     * tries to edit that agent's config file here, fails because the fixture
+     * never created one, and stops before removing the binary and the index —
+     * which is exactly what this test measures. Redirecting PATH makes the
+     * result the same on every machine. */
+    char *old_path = save_test_env("PATH");
+    cbm_setenv("PATH", tmpdir, 1);
 
     char cache_dir[512];
     char index_path[640];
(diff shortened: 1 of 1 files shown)


DeusData:
Merged — thanks for the clean isolation fix. Pointing PATH at the fixture makes the uninstall-quiesce test give the same verdict on every developer machine, exactly the kind of determinism this suite needs.

PR #2056, opened 2026-09-04
fix(extract): extract structure from Svelte/HTML/Astro inline scripts

### LINKED ISSUE #1807: Svelte <script> function/call symbols never extracted — component call-graph is empty (imports work per #223, calls don't; Svelte analog of #1410)
### Summary
For `.svelte` components, the `<script>`/`<script lang="ts">` block is parsed structurally but its **function definitions and call sites are never emitted as graph symbols**. Each component yields only `File` + `Module` nodes. Imports *are* extracted (thanks to #223/#224) and SvelteKit routes too (#369), but the call graph inside components is absent.

This is the Svelte analog of #1410 (Vue `<script setup>`), with the same consequence: incomplete call graph and **false-positive dead-code** for component logic.

### Environment
- v0.10.8, macOS arm64 (darwin), npm install
- tree-sitter Svelte grammar present (parses `script_element`)

### Repro
1. Index any repo containing a `.svelte` component whose `<script lang="ts">` defines named functions (e.g. `function toggleOpen() {…}`).
2. Query the graph:
   ```
   query_graph  MATCH (f:Function) WHERE f.file_path ENDS WITH '.svelte' RETURN count(*)
   query_graph  MATCH (n) WHERE n.file_path ENDS WITH '.svelte' RETURN n.name, labels(n)
   ```

### Observed
- `Function` count for `.svelte` files: **0**
- Component nodes: only `["File"]` and `["Module"]`
- `CALLS` edges originating in real `.svelte` files: **0**
- `IMPORTS` edges from `.svelte`: present ✅ (working, per #224)

Real-world index (a SvelteKit app, 294 components): 294/294 components have File+Module nodes and import edges, but **zero** `<script>`-internal function or call symbols.

### Expected
`<script>` bodies re-parsed as JS/TS (language injection on `script_element` → TS grammar), with normal symbol + `CALLS`/`DEFINES` extraction and line offsets mapped back to the component — matching what already happens for a `.ts` file.

### Evidence that injection is sufficient
Extracting a component's `<script>` verbatim into a sibling `.ts` file makes the same functions appear as full `Function` nodes with call edges:

| Source | `Function` nodes emitted |
|---|---|
| `TreeNode.svelte` (`toggleOpen`, `transformDataForRecursiveTree` in `<script>`) | 0 |
| identical body extracted to `TreeNode.script.ts` | 2 (both, with edges) |

So the grammar + extractor already handle the code — only the injection step from `script_element` into TS symbol/call extraction is missing (the imports path added in #223 is the precedent).

bmcnaboe:
Fixes #1807. Follow-up to #1852, carrying the generalization agreed on #1410.

## What it does

#1852 built the shared embedded-script seam (included-ranges parse over the original buffer, `lang=`/`src=` resolution) and gated structural extraction to Vue. This widens it: Svelte, HTML and Astro `<script>` blocks — and Astro's frontmatter fence — now yield `Function` definitions and `CALLS` edges in host-file coordinates, exactly as a standalone `.ts`/`.js` file does.

Concretely, in `internal/cbm/extract_imports.c` and `internal/cbm/lang_specs.c`:

- The `extract_structure` boolean is gone; every embedded block extracts definitions, imports and calls.
- `vue_embedded_language()` is now `script_embedded_language()` and runs for all four markup hosts. It starts from the spec row's language rather than hard-coding JavaScript, so an attribute-free block keeps its declared grammar. It also learns HTML's `type=`: absent/empty, `module`, `text/javascript` and `application/javascript` pass; `application/json`, `importmap`, `text/x-template`, `application/ld+json` and anything else bail, so data blocks cannot leak garbage definitions. `src=` and unsupported `lang=` bail as before.
- CFML's `<cfscript>` carries no attributes and keeps its spec-fixed language (the resolver is not consulted for it).
- Astro's frontmatter fence and `<script>` bodies default to **TypeScript**, deliberately: Astro type-checks both as TS with no `lang=` attribute to say so, and the fence cannot carry one. The TypeScript grammar parses untyped JavaScript unchanged.

## Tests

- The #1852 control pinning Svelte/HTML/Astro at zero functions and zero calls is flipped to assert the real symbols (`embedded_structure_sibling_hosts_issue1807`).
- Stay-at-zero controls: `<script type="application/json">` and `<script src>` in every host, plus `importmap` and `text/x-template` for HTML (`embedded_structure_inert_blocks_issue1807`). HTML's own tag walker still records the `src=` reference as an import edge; that names the external file and is asserted explicitly.
- Both Svelte blocks contribute — `<script context="module">` and Svelte 5's bare `module` — with per-block `lang=` and host-file line numbers (`svelte_embedded_structure_both_blocks_issue1807`).
- HTML's accepted `type=` forms with line numbers, and a data block on the same page ignored (`html_embedded_structure_issue1807`).
- Astro: TS frontmatter (interface + typed signatures) and a `<script>` body, both in host-file coordinates (`astro_embedded_structure_issue1807`).
- End-to-end `CALLS` edge inside a `.svelte` file, the exact shape #1807 reports missing (`es_calls_svelte_embedded_issue1807`).
- The per-grammar label goldens for Svelte and Astro gain the `Variable` their fixtures always declared (`test_grammar_labels.c`).

## Validation

Indexed a 213-component Vue corpus (~26k nodes, ~75k edges) with `main` and with this branch, isolated caches: node counts per label, per host extension (`.vue`, `.html`, `.ts`), `CALLS`/`IMPORTS` counts per extension, and `trace_path` output for Vue-defined functions are identical. Vue behaviour is unchanged.

`scripts/test.sh` and `scripts/lint.sh` are green locally. DCO signed.


DeusData:
Merged as 2b030a5b — thank you for this one. Inline scripts in Svelte, HTML and Astro files now contribute their structure to the graph instead of vanishing, the tests pin both the extraction and the label side, and you did the overlap check on the sibling PRs yourself, which made the review easy. Much appreciated.

PR #2059, opened 2026-09-04
fix(cli): YAML removers answer OK for an absent config, so uninstall no longer refuses

dgilperez:
## Problem

`codebase-memory-mcp uninstall` exits 1 on any macOS where Hermes is installed. Hermes is detected by its binary on PATH (`cbm_agent_cli_exists("hermes", ...)`), not by its home, so a HOME without `.hermes/` reaches the Hermes uninstall ops with `~/.hermes/config.yaml` absent. The JSON removers already treat a missing config as "nothing to remove" (`cbm_json_like_read_document` returns 1 → `CLI_OK`). The four YAML removers acquire the lock first, and the lock directory is created beside the target, so with the parent directory absent the lock fails and the adapter records:

```
error: agent_config agent=Hermes op=pre_llm_hook_uninstall path=.../.hermes/config.yaml (target: does not exist or cannot be inspected)
error: agent_config agent=Hermes op=mcp_uninstall path=.../.hermes/config.yaml (target: does not exist or cannot be inspected)
error: one or more agent cleanup operations failed; executable and index removal were not started
```

This is also why `cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index` fails on such a machine: on a pristine `main` (2026-09-04) the suite gives `7837 passed, 1 failed, 7 skipped` here, with Hermes at `~/.local/bin/hermes`.

## Fix

`cbm_yaml_remove_mapping_entry`, `cbm_yaml_remove_owned_mapping_entry`, `cbm_yaml_remove_mapping_sequence_item` and `cbm_yaml_remove_string_list_item` short-circuit with their OK value on a true ENOENT (`lstat`, or `ERROR_FILE_NOT_FOUND`/`ERROR_PATH_NOT_FOUND` on Windows) and create nothing. A dangling symlink still reaches the reader and keeps its byte-identical rejection (`config_yaml_edit_nested_sequence_rejects_symlink_byte_identically` unchanged).

## Validation

- New test `config_yaml_edit_remove_on_absent_config_is_a_no_op` covers the four removers against a path whose directory does not exist.
- `make -f Makefile.cbm test` on macOS arm64 with Hermes installed: all passed, 7 skipped (the previously failing `cli_uninstall_quiesces_active_cohort_before_removing_binary_and_index` included).
- DCO sign-off included.

### DIFF
--- a/src/cli/config_yaml_edit.c
+++ b/src/cli/config_yaml_edit.c
@@ -3601,8 +3601,49 @@ int cbm_yaml_upsert_mapping_entry(const char *file_path, const char *section_key
     return result == 0 && release_result == 0 ? 0 : YAML_ERROR;
 }
 
+/* Removal is idempotent: a config that does not exist has nothing to remove.
+ * The JSON removers already answer OK here (cbm_json_like_read_document's
+ * "missing" result); the YAML removers instead took the lock first, and the
+ * lock directory is created beside the target — so with the parent directory
+ * absent (`~/.hermes/` never created because Hermes was detected by its
+ * binary on PATH, not by its home) every uninstall reported
+ * "does not exist or cannot be inspected" and refused to remove the
+ * executable. Only a true ENOENT short-circuits: a dangling symlink still
+ * reaches the reader and its byte-identical rejection. */
+static bool yaml_remove_target_absent(const char *path) {
+    if (!path) {
+        return false;
+    }
+#ifdef _WIN32
+    /* FILE_FLAG_OPEN_REPARSE_POINT, as in toml_read_file: opens the link
+     * itself rather than following it, so a dangling symlink does NOT count as
+     * absent and still reaches the reader — matching the POSIX branch, where
+     * lstat does not follow either. */
+    wchar_t *wide = cbm_utf8_to_wide(path);
+    if (!wide) {
+        return false;
+    }
+    HANDLE handle = CreateFileW(
+        wide, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, NULL,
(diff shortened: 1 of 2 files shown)


DeusData:
Thanks — this is right, and I verified it the hard way: with a fake hermes on PATH and a fresh HOME without .hermes/, main's uninstall really does stop with the Hermes error and keeps the binary; your test is red against main's config_yaml_edit.c and green with your change; the merged tree passes config_yaml_edit, cli, activation_transaction, daemon_bootstrap and daemon_frontend (411 cases) plus the end-to-end run. The JSON and TOML removers already behave this way, so this aligns YAML with them. One thing keeps CI red: both Windows shards fail to COMPILE — tests/test_config_yaml_edit.c:1851 calls lstat, and the new TEST body sits outside the #ifndef _WIN32 block while its RUN_TEST is inside one. Wrap the test definition in the same #ifndef _WIN32 / #endif (or drop the lstat assertion), keep the sign-off, and I'll merge on green.

dgilperez:
Thanks for verifying it that thoroughly - fixed both shards.

The test now checks "nothing was created" with cbm_path_info_utf8 instead of lstat, and its RUN_TEST moves out of the #ifndef _WIN32 block, so it runs on Windows too. Wrapping the test would have left the Windows branch untested, and that branch had a bug: GetFileAttributesW follows symlinks, so a dangling symlink counted as absent and skipped the reader, unlike the POSIX branch. It now uses CreateFileW with FILE_FLAG_OPEN_REPARSE_POINT, mirroring toml_read_file.

DeusData:
Merged as a6aea0bc — thank you. Uninstall no longer trips over a YAML config that was never written, the test pins it, and the Windows probe correction (open the reparse point instead of following it) was the right call on top of the compile fix I asked for. Much appreciated.

dgilperez:
Happy to help! Thank you for maintaining this nice tool!

Issue #1722, opened 2026-08-19
install fails all Claude Code agent/hook writes when ~/.claude is a directory symlink — misleading "(target: does not exist or cannot be inspected)"


chen5541741:
### Version

0.10.8

### Platform

Windows (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

## Summary

On Windows, `install` (v0.10.8, via `install.ps1`) fails every Claude Code
agent-configuration write when `~/.claude` is a **directory symbolic link**
(here `C:\Users\mayn\.claude` → `I:\.claude`, relocated because the system
drive is nearly full). The paths exist and are writable — the installer
appears to refuse to write through a reparse point, but the error says the
opposite.

## Environment

- Windows 10 Pro 10.0.19045, amd64
- codebase-memory-mcp 0.10.8 (installed via `install.ps1`)
- `C:\Users\mayn\.claude` = SymbolicLink → `I:\.claude`
- Detected agents: Claude Code, VS Code

## Observed

```
Claude Code:
  skills: 0 installed
error: agent_config agent=Claude Code op=agent_install path=C:/Users/mayn/.claude/agents/codebase-memory-scout.md (target: does not exist or cannot be inspected)
error: agent_config agent=Claude Code op=agent_install path=C:/Users/mayn/.claude/agents/codebase-memory.md (target: does not exist or cannot be inspected)
error: agent_config agent=Claude Code op=agent_install path=C:/Users/mayn/.claude/agents/codebase-memory-auditor.md (target: does not exist or cannot be inspected)
  mcp: C:/Users/mayn/.claude.json
error: agent_config agent=Claude Code op=hook_script_install path=C:/Users/mayn/.claude/hooks/cbm-code-discovery-gate.cmd (target: does not exist or cannot be inspected)
error: agent_config agent=Claude Code op=hook_script_install path=C:/Users/mayn/.claude/hooks/cbm-session-reminder.cmd (target: does not exist or cannot be inspected)
error: agent_config agent=Claude Code op=hook_script_install path=C:/Users/mayn/.claude/hooks/cbm-subagent-reminder.cmd (target: does not exist or cannot be inspected)
...
error: installation failed (exit code 1)
```

What this suggests:

- Every write to a **real** path succeeded: `~/.claude.json` (MCP registration),
  `~/.copilot/**` (VS Code agents/skill/hooks), `AppData\...\Code\User\mcp.json`.
[...]

PR #2060, opened 2026-09-05
fix(cli): follow agent-root symlinks owned by the invoking user

TheStranjer:
## What does this PR do?

Lets `install` write through an agent config directory that is a symlink the
invoking user owns.

Refs #1722. Not `Fixes`: this resolves the POSIX directory-symlink case, and
the issue was opened for the Windows reparse-point case, which is a separate
code path (see Scope). Bug fix, so submitted under the CONTRIBUTING exception
for focused fixes; the contract change is called out below because it may
warrant the design discussion the maintainer reserved in #1722.

### The bug

`~/.config/opencode -> /run/media/<me>/<drive>/opencode/config`, a user-owned
directory symlink. Every OpenCode write failed:

```
error: agent_config agent=OpenCode op=mcp_install path=/home/<me>/.config/opencode/opencode.json (target: regular file, 899 bytes)
error: agent_config agent=OpenCode op=instructions_install path=/home/<me>/.config/opencode/AGENTS.md (target: does not exist or cannot be inspected)
error: agent_config agent=OpenCode op=agent_install path=/home/<me>/.config/opencode/agents/codebase-memory-scout.md (target: does not exist or cannot be inspected)
...
error: activation stopped after one or more agent configuration or cleanup operations failed
```

Same shape as #1722 (Windows, and the macOS/stow confirmation in its comments).
The `(target: ...)` detail is a red herring: the file is fine, the refusal is
the parent-chain walk in `cbm_mkdir_p`, which opens every component with
`O_NOFOLLOW` and only retries root-owned links. A user-owned link fails with
ELOOP, `prepare_config_parent` returns false, and every write under that root
is reported as an opaque `agent_config` error.

### The change

`cbm_open_directory_component` now also follows a symlink whose owner is the
invoking effective uid.

Why that is safe: the refusal exists so the installer never writes through a
redirection created by someone with different authority: a link planted by
another account in a shared ancestor, or a user-owned link followed by a
privileged install. A symlink's owner is its creator, so a link owned by the
invoking account was made by that account or by root, and following it cannot
reach anything the user could not already write directly. The rule rests only
on POSIX symlink ownership, so it holds on macOS and the BSDs as on Linux. It
is the rule the Linux kernel's `fs.protected_symlinks` applies, and the same
trust `activation_posix_intermediate_secure` already extends to ancestor
directories (`st_uid == 0 || st_uid == geteuid()`).

What stays refused:

- a link owned by any other account
- a user-owned link when the installer runs as root (euid 0)
- symlinked config *files*: the `O_NOFOLLOW` writers in `config_*_edit.c`
  are untouched (#1560, #1954 are a different case)

The `O_NOFOLLOW` walk itself is unchanged, so the validate-then-write race
stays closed; only the one trusted component is reopened without the flag.
The first `openat`'s errno is preserved when a link is refused so the caller's
ENOENT check never reads `fstatat`'s.

Of the three remedies floated in #1722 (resolved-target validation, explicit
opt-in, clearer hard refusal), this is the first, with link ownership as the
validation.

### Contract change, stated plainly

`cli_installer_rejects_symlinked_agent_roots` (from `a3903caa`, refined in
`34f9ad1b`) asserted that a user-owned link is refused. Its stated model was
"only root can create root-owned links, so they are outside the attacker
model" and "the contract under test is refusal of USER-planted links". That
reasoning is about who could have created the link, and it never distinguished
a link the user made from one another account planted. The test is reworked as
`cli_installer_follows_symlinked_agent_roots_only_for_trusted_owners`: own
links are followed everywhere; when the suite runs as root (containers) it
additionally demotes the link to uid 65534 and asserts refusal, which is the
only way a foreign-owned link can be created in a test.

### Scope

- POSIX only. The Windows walk (`cbm_windows_mkdir_component`) is a separate
  path and is unchanged, so the Windows reparse-point report in #1722 is not
  fixed here. A port there needs its own design: junctions need no privilege
  to create, and objects created from an elevated shell are owned by the
  Administrators group SID rather than the user's.
- Detection is unchanged: a client detected only by `dir_exists()` on its root
  (which uses `lstat`) is still skipped when that root is a symlink. Separate
  issue, no error is printed for it.
- Known limitation shared with every existing `st_uid == geteuid()` check in
  the tree: on mounts that synthesize ownership (ntfs3/CIFS with `uid=`,
  DrvFs), the owner check is only as good as the reported uid.

### Tests

- `platform_mkdir_p_follows_symlink_owned_by_caller`: creates through an
  own symlink; fails closed on a dangling one without creating anything.
- Reworked installer contract test, above.

Red/green: without the fix both fail, and the installer test prints the exact
error text from the report. With it, every C suite (119, via
`scripts/test.sh --suites ...`) passes except six failures that are local to
my machine and fail identically on `main`:

- 5 in `cli` (`test_cli.c` 1088, 1136, 1202, 1291, 1498): the tests stage
  the runner binary from the repo path, and this checkout lives on a mount
  whose root is 0777, which the ancestor policy refuses
  (`ancestor_directory_world_writable (mode 0777, uid 1000)`).
- 1 in `mcp` (`detect_changes_seeds_only_touched_symbol_issue1363`): my
  global git config sets `diff.mnemonicprefix=true`, so `git diff` emits
  `+++ w/` and `pass_gitdiff.c` only recognises `+++ b/`. Passes with
  `GIT_CONFIG_GLOBAL=/dev/null`. Pre-existing; I will file it separately.

Verified end to end: built from this branch, installed on the affected
machine, all nine detected agents configured with zero errors, OpenCode files
present in the symlink target, and the installed server answering MCP
`initialize`. Used in a real session afterward.

## Checklist

- [x] Tests pass locally (`make -f Makefile.cbm test`) — all 119 C suites
      run; the six local failures above are environment-specific and
      reproduce on `main` unchanged
- [x] Lint passes (`make -f Makefile.cbm lint-ci`) — ran the exact CI
      command, `scripts/lint.sh --ci` with clang-format 20.1.8 and cppcheck
      2.20.0: cppcheck, clang-format, no-skips policy and NOLINT check all
      pass. clang-tidy on the changed lines is clean apart from the
      pre-existing cognitive complexity of `cbm_mkdir_p`.

### DIFF
--- a/src/foundation/compat_fs.c
+++ b/src/foundation/compat_fs.c
@@ -837,14 +837,32 @@ FILE *cbm_fopen(const char *path, const char *mode) {
     return fopen(path, mode);
 }
 
+/* Symlink policy for the parent-chain walk. Every component is opened with
+ * O_NOFOLLOW; a symlink is followed only when its OWNER is trusted, never by
+ * default. Trusted owners are root (distro /home indirection, macOS /tmp:
+ * only root can create those, so they are outside the attacker model) and the
+ * invoking account itself: a link the caller owns is the caller's own
+ * arrangement, not something planted on them. That second case is how dotfile
+ * managers and configs kept on another volume look (~/.config/opencode ->
+ * /mnt/...), and refusing it made every agent-config write under such a root
+ * fail with an opaque agent_config error. It is the same rule the Linux
+ * kernel's fs.protected_symlinks applies, and the ancestor policy the
+ * activation transaction already uses. A link owned by any OTHER account
+ * (planted in a group- or world-writable ancestor) stays refused, and a
+ * privileged install (euid 0) still refuses user-owned links. */
 static int cbm_open_directory_component(int parent, const char *component, int flags) {
     int descriptor = openat(parent, component, flags);
 #if defined(O_NOFOLLOW) && defined(AT_SYMLINK_NOFOLLOW)
     if (descriptor < 0) {
+        /* The caller decides on errno from the FIRST open (ENOENT means
+         * "create it"); a refused link must not leak fstatat's errno instead. */
+        int open_errno = errno;
         struct stat state;
         if (fstatat(parent, component, &state, AT_SYMLINK_NOFOLLOW) == 0 &&
-            S_ISLNK(state.st_mode) && state.st_uid == 0U) {
+            S_ISLNK(state.st_mode) && (state.st_uid == 0U || state.st_uid == geteuid())) {
             descriptor = openat(parent, component, flags & ~O_NOFOLLOW);
+        } else {
+            errno = open_errno;
         }
     }
 #endif
--- a/src/foundation/compat_fs.c
+++ b/src/foundation/compat_fs.c
@@ -558,6 +558,13 @@ bool cbm_mkdir_p(const char *path, int mode) {
     return ok;
 }
 
+bool cbm_mkdir_p_ex(const char *path, int mode, unsigned int policy) {
+    /* The Windows walk has its own reparse-point policy (see
+     * cbm_windows_mkdir_component); the POSIX symlink policy does not apply. */
+    (void)policy;
+    return cbm_mkdir_p(path, mode);
+}
+
 int cbm_unlink(const char *path) {
     wchar_t *wpath = cbm_path_to_wide(path);
     if (!wpath) {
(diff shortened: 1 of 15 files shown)


DeusData:
Thanks for taking this on — the installer case you describe is real, and the errno save/restore plus the fstatat(AT_SYMLINK_NOFOLLOW) ownership check are the right ingredients for it.

One scoping request before I can take it: please keep the new ownership-based symlink following confined to the installer's own walk (`prepare_config_parent` and the agent-config roots it resolves under `$HOME` / XDG), and leave `cbm_mkdir_p` itself with its current refusal. That helper is generic with two dozen callers, and the trust argument in your description ("the link's owner is the invoking user, so they created it") is one the installer can make for the paths it derives from the home directory; the shared helper cannot make it for its other callers, so the policy has to live at the call site. Concretely: a small wrapper or flag used only by the installer walk, unchanged behaviour everywhere else, and your tests pinned to that path (the current tests pass on a root runner because root links were already followed, so a non-root control would make them bind).

Separately, this PR settles which remedy #1722 gets; that choice is still an open design decision on the issue rather than something to decide inside a PR, so I'd keep the PR to the mechanism and let #1722 carry the "which remedy" discussion.

With those two, this is a good fix and I'll review the new head promptly.

TheStranjer:
Do you prefer I make these changes now or wait for a resolution on which method from #1722 will ultimately be used, if any?

DeusData:
Thank you for this — the direction is the one we want (a directory component that is a symlink owned by root or the invoking user is followed; other owners, root-run user links, and symlinked files stay refused), the precedent in `activation_transaction.c` holds, and the platform test binds (RED without the `compat_fs.c` change, GREEN with it). Two small asks before merge, both hardening rather than correctness:

1. After the follow-open in `cbm_open_directory_component`, `fstat()` the returned descriptor and require the target directory's `st_uid` to be 0 or `geteuid()` as well. That bounds what a swapped link can reach in the `fstatat` → `openat` window, the same before/after-identity idiom `activation_transaction.c` uses around its own opens. If you think it is unnecessary, a one-line comment saying why is fine too.

2. The PR text says a user-owned link is refused when running as root, but the test only asserts a foreign-owned link is refused as root. Please add the leg for a user-owned (`lchown` to a non-root uid) link under euid 0 so the claim is pinned; it can stay skipped on non-root runs like the existing refusal leg.

One note for the record, no change needed: `cbm_mkdir_p` is the only caller, so this is a global policy change (cache and artifact directories follow user-owned links too, ~24 call sites), which is acceptable and worth one sentence in the PR description.

DeusData:
One more, and it is the important one — it came out of a second security read of the change, not of your code as such.

`cbm_mkdir_p` is not only used for agent roots and cache directories under HOME: `artifact.c` calls it for `<repo_path>/.codebase-memory`, inside the checkout being indexed, automatically after every persisted index. Git creates symlinks owned by whoever cloned, so with an "owner == invoking user" rule a repository that ships `.codebase-memory -> ~/.claude` (or `~/.ssh`, or an attacker-controlled world-writable directory with a pre-planted `<name>.tmp` link) would be followed, and the artifact writer's temp-file + rename would land there. "User-owned" and "user-intended" are not the same thing once the path root is a checkout.

So the follow needs to be opt-in per call site rather than the default for every `cbm_mkdir_p` caller: for example `cbm_mkdir_p_ex(path, mode, CBM_MKDIR_FOLLOW_OWNED)` (or a flags parameter) used only by HOME/XDG-rooted callers — agent roots, config, cache and log directories — while every `repo_path`-derived caller keeps today's strict `O_NOFOLLOW` walk. Please also keep the post-open `fstat()` check from the earlier ask (target is a directory, owned by root or the invoking user, and not world-writable unless sticky), and add a test that a repository-level `.codebase-memory -> <user-owned dir>` link is still refused while `~/.claude -> ~/dotfiles/claude` is followed.

With that in place this is a merge, and it unblocks the stow / moved-drive users in #1722. Thank you for bearing with the extra round.

src/foundation/compat_fs.c
```diff
@@ -837,14 +837,32 @@ FILE *cbm_fopen(const char *path, const char *mode) {
...
 #if defined(O_NOFOLLOW) && defined(AT_SYMLINK_NOFOLLOW)
     if (descriptor < 0) {
+        /* The caller decides on errno from the FIRST open (ENOENT means
+         * "create it"); a refused link must not leak fstatat's errno instead. */
+        int open_errno = errno;
         struct stat state;
         if (fstatat(parent, component, &state, AT_SYMLINK_NOFOLLOW) == 0 &&
-            S_ISLNK(state.st_mode) && state.st_uid == 0U) {
+            S_ISLNK(state.st_mode) && (state.st_uid == 0U || state.st_uid == geteuid())) {
             descriptor = openat(parent, component, flags & ~O_NOFOLLOW);
```
AmirF194: I extracted `cbm_open_directory_component` verbatim (head `3a8b8e8`) and ran it in a clean container against the ownership check.

`fstatat` (line 861) and the `openat` that follows the link (line 863) are separate syscalls on the same entry, with nothing pinning what gets opened to what got checked. I simulated the window with an LD_PRELOAD shim that swaps the symlink for a foreign-owned one right after the real `fstatat` returns, the same effect a writer to the parent directory would have:

```
baseline: FOLLOWED resolved_to=/tmp/race/legit_target
shim armed (link swapped between check and open):
FOLLOWED resolved_to=/tmp/race/evil_target
$ stat -c '%U %u' /tmp/race/parent/link
nobody 65534
```

The check passed against a root-owned link; the code then opened one now owned by an untrusted account, the case this function says must stay refused. I'm not claiming this races under real thread timing, only that the check and the open aren't atomic.

One fix: open the component with `O_PATH | O_NOFOLLOW` to get a handle bound to that inode, `fstat` that fd for the decision, then follow it via `/proc/self/fd/<n>` instead of reopening by name.

AmirF194: I re-ran the race harness against 8622d6a0 (real cbm_mkdir_p_ex, LD_PRELOAD swaps the symlink to a directory owned by an untrusted uid right after the ownership check, same as before). The post-open fstat on the followed descriptor rejects the swapped target and cbm_mkdir_p_ex returns false; no directory gets created under the attacker's path. This closes the race, thanks.

TheStranjer:
@DeusData @AmirF194 I think I got all of the things listed, tell me if there are more (I will check in the morning)

DeusData:
Merged — thank you for this, @TheStranjer. Binding the symlink ownership judgement to the resolved target (owned by the invoking user) while refusing other-owner links is exactly the right call, and it held up under the security review. Appreciate the careful iteration through the review rounds.

PR #2066, opened 2026-09-05
test(mcp): over-long grep records never invent matches (#2011)

### LINKED ISSUE #2011: search_code invents file paths and line numbers: fgets splits a >2047-byte grep record and the continuation is parsed as a fresh file:line:content
### What happens

`collect_grep_matches()` reads the scan output with `fgets` into a 2048-byte buffer and parses each line as `file:line:content`. A grep record longer than 2047 bytes is split across two `fgets` calls, and the continuation chunk is then parsed as if it were a fresh record: the text before its first `:` becomes the **file**, and the text after it becomes the **line number**.

Any file with a line longer than ~2 KB containing colons reproduces it — a minified bundle, generated code, a long data row.

### Reproduction

```sh
mkdir -p /tmp/fx && cd /tmp/fx && git init -q
python3 -c "
pairs = ','.join('k%d:\"v%d\"' % (i, i) for i in range(260))
open('big.js','w').write('var CFG=({' + pairs + ',NEEDLEmarker:1});\n')
open('normal.js','w').write('const NEEDLEmarker = 42;\n')
"
git add -A && git -c user.email=t@t -c user.name=t commit -qm init
codebase-memory-mcp cli index_repository --repo-path /tmp/fx
codebase-memory-mcp cli --json search_code --project <project> --pattern NEEDLEmarker
```

`big.js` is a single 2927-byte line. The response comes back `isError: false`, and contains:

```
raw: 1  (cols: file line content)
  "v181",k182     0     "v183",k184:"v184",k185:"v185", … ,NEEDLEmarker:1});

total_grep_matches: 3
total_results: 2
```

The `file` is a fragment of JavaScript — `"v181",k182` — and the line number is `0`. No such file exists. `total_grep_matches` is 3 for a repository containing 2 matches, because the split produced a third record.

An agent that reads this calls `get_code_snippet` on a path that was never in the repository, or reports to the user that a symbol lives at a file and line that do not exist.

### Mechanism

`src/mcp/mcp.c:10261` `collect_grep_matches()`:

```c
char line[CBM_SZ_2K];                       /* 2048 — src/foundation/constants.h:36 */

while (fgets(line, sizeof(line), fp) && gm_count < grep_limit) {
    ...
    char *sep1 = strchr(line, (unsigned char)sep);
    if (!sep1) { continue; }
    char *sep2 = strchr(sep1 + SKIP_ONE, (unsigned char)sep);
    if (!sep2) { continue; }
    *sep1 = '\0';
    *sep2 = '\0';
    const char *path = line;                                  /* mcp.c:10300 */
    ...
    snprintf(gm[gm_count].file, sizeof(gm[0].file), "%s", file);
    gm[gm_count].line = (int)strtol(sep1 + SKIP_ONE, NULL, CBM_DECIMAL_BASE);
```

Nothing distinguishes a continuation chunk from a record start. When the tail happens to contain two colons the chunk becomes a fabricated match; when it does not, the `continue` silently drops a genuine one instead — so the same split can either invent a match or lose one.

### Suggested fix

Detect the truncation `fgets` already tells you about: a chunk that does not end in `\n` is a partial record. Either accumulate until the newline arrives, or consume and discard the remainder of the over-long record and count it as skipped, rather than feeding the tail back into the `file:line:content` parser. Validating the parsed line number (reject a non-numeric or `<= 0` field) would catch the fabricated rows as a second line of defence, but on its own it would still drop the real match silently.

### Environment
- Built from `main` @ 5fbab7bb (plus the unrelated cypher patch in #2002; `collect_grep_matches` is untouched by it)

kavish-19:
Regression test for #2011. **The production fix already landed via #1597 (`3c7427e`)** — this PR is now the test only.

## What #2011 was

A grep record longer than the read buffer was split across two reads, and the continuation was parsed as a fresh `file:line:content` record: a fragment of matched content became the **file**, and the text after the next delimiter became the **line number**. An agent reading that result called `get_code_snippet` on a path that was never in the repository. The same split could also *lose* a real match, when the tail held fewer than two delimiters and the record was dropped by a `continue` instead.

## Why this is test-only now

`search_code` no longer reads into a fixed buffer. On current `main`, `scan_and_classify_grep_matches` reads each record with `cbm_getline` into a growable buffer:

```c
ssize_t line_length = cbm_getline(&line, &line_capacity, fp);
```

so a record cannot split, and `collect_grep_matches` is gone. I confirmed that before rebasing — the only remaining mention of the old function on `main` is a word in a comment.

That closed #2011 as a side effect of the lean-output work rather than as a targeted fix, so **nothing on `main` guards the shape today**. A future rewrite of the read loop that reintroduced a fixed buffer would restore the bug silently. Hence the test.

## The test

`tests/test_mcp.c` → `search_code_long_line_does_not_invent_matches`, beside the existing `search_code_path_filter_*` tests and registered in the suite. It writes a >2 KiB colon-laden line plus a normal one, calls `search_code` through `cbm_mcp_server_handle`, and asserts:

- `total_grep_matches == 2` — the repository's two real matches, not the three the split produced
- no row carries `line: 0`, the value the fabricated record used to yield

Against the original buffered implementation it failed exactly as #2011 describes (`grep_matches == 3, expected 2`); against `main`'s `cbm_getline` version it passes, as you confirmed on your side.

## Changes from the first version of this PR

- Rebased onto current `main`; the `src/mcp/mcp.c` drain hunk is dropped entirely — `main`'s file taken as-is, per your resolution.
- Retitled from `fix(mcp): …` to `test(mcp): …`, since there is no production change left.
- Commit re-authored under `kavish-19 <63698788+kavish-19@users.noreply.github.com>`; the earlier push used the wrong email of mine, which I have corrected on #2065 too. Sign-off matches the author on both.

Diff against `main` is `tests/test_mcp.c` only.

### DIFF
--- a/tests/test_mcp.c
+++ b/tests/test_mcp.c
@@ -10118,6 +10118,73 @@ TEST(search_code_path_filter_prefilter_keeps_matches) {
     PASS();
 }
 
+/* #2011: a grep record longer than the read buffer used to be split across two
+ * fgets calls, and the continuation was parsed as a fresh file:line:content
+ * record — inventing a file path out of matched content and a line number of 0.
+ * A single >2 KiB line containing colons reproduces it: the repository holds
+ * two matches, and the split used to report three. */
+TEST(search_code_long_line_does_not_invent_matches) {
+    char tmp[512];
+    snprintf(tmp, sizeof(tmp), "/tmp/cbm_srch_longline_XXXXXX");
+    ASSERT_TRUE(cbm_mkdtemp(tmp) != NULL);
+
+    char big_path[768], normal_path[768];
+    snprintf(big_path, sizeof(big_path), "%s/big.js", tmp);
+    snprintf(normal_path, sizeof(normal_path), "%s/normal.js", tmp);
+
+    /* One line well over the 2 KiB read buffer, full of colons, with the
+     * needle at the very end so the match lands past the split point. */
+    FILE *fp = fopen(big_path, "w");
+    ASSERT_NOT_NULL(fp);
+    fprintf(fp, "var CFG=({");
+    for (int i = 0; i < 260; i++) {
+        fprintf(fp, "k%d:\"v%d\",", i, i);
+    }
+    fprintf(fp, "NEEDLEmarker:1});\n");
+    fclose(fp);
+
+    fp = fopen(normal_path, "w");
+    ASSERT_NOT_NULL(fp);
+    fprintf(fp, "const NEEDLEmarker = 42;\n");
+    fclose(fp);
+
+    cbm_mcp_server_t *srv = cbm_mcp_server_new(NULL);
+    ASSERT_NOT_NULL(srv);
+    const char *proj = "longline-search";
+    cbm_mcp_server_set_project(srv, proj);
+    cbm_store_upsert_project(cbm_mcp_server_store(srv), proj, tmp);
+
+    char *resp = cbm_mcp_server_handle(
+        srv, "{\"jsonrpc\":\"2.0\",\"id\":96,\"method\":\"tools/call\","
+             "\"params\":{\"name\":\"search_code\","
+             "\"arguments\":{\"pattern\":\"NEEDLEmarker\",\"project\":\"longline-search\"}}}");
+    ASSERT_NOT_NULL(resp);
+    char *inner = extract_text_content(resp);
+    ASSERT_NOT_NULL(inner);
+
+    /* Exactly the two real matches — the split must not produce a third. */
+    int grep_matches = -1;
+    const char *g = strstr(inner, "\"total_grep_matches\":");
+    if (g) {
+        sscanf(g, "\"total_grep_matches\":%d", &grep_matches);
+    } else if ((g = strstr(inner, "total_grep_matches: ")) != NULL) {
+        sscanf(g, "total_grep_matches: %d", &grep_matches);
+    }
+    ASSERT_EQ(grep_matches, 2);
+
+    /* Every reported line number belongs to a real line: the fabricated row
+     * carried line 0, which no grep -n record can produce. */
+    ASSERT_TRUE(strstr(inner, "\"line\":0") == NULL);
+
+    free(inner);
+    free(resp);
+    cbm_mcp_server_free(srv);
+    unlink(big_path);
+    unlink(normal_path);
+    rmdir(tmp);
+    PASS();
+}
+
 /* PR #756 (distilled): path_filter matching ZERO indexed files. With the
  * prefilter the scoped filelist has 0 records, and handle_search_code now
  * skips the grep subprocess entirely (xargs on an empty filelist is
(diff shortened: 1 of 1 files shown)


DeusData:
Thanks — this is a correct fix for the tree it was written against, and I checked it the hard way: at your base (`ceba4272`), reverting only your `src/mcp/mcp.c` hunk makes your new test fail exactly as #2011 describes (`tests/test_mcp.c:5402: grep_matches == 3, expected 2`), and restoring it passes; `len == sizeof(line)-1 && line[len-1] != '\n'` is precisely `fgets`'s truncation signal and the drain sits before every `continue`. Nicely done.

What changed underneath you: #1597 merged on main this morning (`3c7427ef`) and replaced `collect_grep_matches` altogether — `search_code` now reads each grep record with `cbm_getline` (unbounded, doubling buffer) in `scan_and_classify_grep_matches`, so an over-long record can no longer split. I ran your test against main's production code: it passes, and the whole `mcp` suite is green with it (310 passed). So the production half of this PR is already on main, by a side effect nobody guarded — which is exactly why your test is still valuable: nothing on main protects the >2 KiB record shape today, and #2011 (which I've closed with the fixing SHA) deserves a regression test that names it.

Could you rebase and keep just the test? The one conflict is delete-vs-modify on `src/mcp/mcp.c`; the resolution is to take main's version as-is:

```
git fetch upstream main && git rebase upstream/main
# conflict in src/mcp/mcp.c: main already reads each record with cbm_getline,
# so your drain hunk has nothing left to fix — take main's file
git restore --source=upstream/main src/mcp/mcp.c && git add src/mcp/mcp.c
git rebase --continue            # tests/test_mcp.c merges on its own
```

Then retitle to something like `test(mcp): over-long grep records never invent matches (#2011)` and note in the body that the production fix landed via #1597. Keep the sign-off, and I'll merge it on green. Welcome aboard — and #2065 is being reviewed alongside.

kavish-19:
Thank you — and for checking it at my base rather than just against `main`; that made the situation unambiguous.

Verified your account of `main` before rebasing: `scan_and_classify_grep_matches` reads each record with `cbm_getline(&line, &line_capacity, fp)` into a growable buffer, so the split cannot happen, and `collect_grep_matches` survives only as a word in a comment at `mcp.c:12327`. My drain hunk has nothing left to fix.

Rebased exactly as you set out — `git restore --source=upstream/main src/mcp/mcp.c`, and `tests/test_mcp.c` merged on its own. The diff against `main` is now that one test file, +68/-0. Retitled, and the body says the production fix landed via #1597 (`3c7427e`) and explains why the shape is still worth guarding: it was closed as a side effect, so nothing on `main` protects it if the read loop is ever rewritten with a fixed buffer again.

One correction on my side, unprompted: the earlier pushes on this branch and on #2065 were authored under the wrong email of mine. Both are re-authored under `kavish-19 <63698788+kavish-19@users.noreply.github.com>`, with the sign-off matching the author. Content is byte-identical — I diffed the new commits against the old ones to be sure only the identity changed.

Same caveat as on #2065 about what my "verified" covers: `scripts/build.sh` fails on this machine at `internal/cbm/preprocessor.cpp` (`fatal error: 'cctype' file not found` — the C++ standard-library headers are missing from its Command Line Tools), and that blocks the test runner, so I have not executed the test myself. You have, which is the run that counts here. If CI disagrees I will fix it promptly.

PR #2070, opened 2026-09-06
fix(daemon): keep one separator spelling for the session root on Windows

DeusData:
## What

On Windows the daemon kept the client's `session_root` / `allowed_root` and the auto-index job's `repo_path` in the platform's native backslash spelling: `cbm_mcp_server_set_session_context` stored the roots verbatim, and `application.c` canonicalized with the bare `cbm_canonical_path` at three sites. Every tool handler, however, normalizes separators after canonicalizing. One directory therefore had two names inside the daemon.

## Symptom

`application_index_args_equal` compares roots exactly, so an explicit `index_repository` request for the session root never equalled the running auto-index job's args and was refused as **`OPTIONS_CONFLICT`** instead of joining that job.

## Fix

- `cbm_mcp_server_set_session_context` normalizes the separator spelling of both roots before storing them.
- `application.c` canonicalizes **and** normalizes at the three sites through one helper, `application_canonical_root()` (= `cbm_canonical_path` + `cbm_normalize_path_sep`).
- New platform-independent test `daemon_session_context_keeps_one_spelling_of_a_root` in `tests/test_daemon.c`. Separators fold on every platform, so it binds wherever the suite runs.

Distilled from #1726 with co-author credit to @liuchong -- the fix only, nothing from the feature.

## Verification

- **Revert-check** (test kept, production reverted, runner rebuilt): the new test is RED on main for the right reason -- `tests/test_daemon.c:483: "C:\repos\cbm" != "C:/repos/cbm"` (daemon suite 12 passed, 1 failed). Fix re-applied byte-identical: green.
- **Suites** (macOS, ASan/UBSan runner): `daemon daemon_application mcp daemon_ipc cli` -- 728 passed, 4 skipped.
- **Lint**: `make -f Makefile.cbm lint-ci` (cppcheck, clang-format, NOLINT whitelist) green; `scripts/check-no-test-skips.sh` OK.
- The local Windows VM leg was skipped (host memory pressure); CI's Windows shards are the Windows proof for this change.

Refs #1726

### DIFF
--- a/src/daemon/application.c
+++ b/src/daemon/application.c
@@ -379,6 +379,20 @@ static bool application_canonical_directory_exists(const char *path) {
     return cbm_path_info_utf8(path, &info) == 0 && info.is_directory;
 }
 
+/* The daemon's one spelling of a repository root. cbm_canonical_path answers in
+ * the platform's native form (backslashes on Windows) while every tool handler
+ * normalizes the separators of the paths it canonicalizes. A root kept native
+ * is a second name for one directory, and the exact comparison downstream -
+ * index-args equality when a request joins the job already running for that
+ * root - never matches it. */
+static bool application_canonical_root(const char *path, char *out, size_t out_size) {
+    if (!cbm_canonical_path(path, out, out_size)) {
+        return false;
+    }
+    cbm_normalize_path_sep(out);
+    return true;
+}
+
 static cbm_daemon_application_watch_t *application_find_watch_locked(
     cbm_daemon_application_t *application, const char *project) {
     for (cbm_daemon_application_watch_t *watch = application->watches; watch; watch = watch->next) {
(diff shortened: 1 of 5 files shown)


DeusData:
**Windows red, and what changed.**

`test / test-windows (windows-latest, CLANG64, x86_64, 1/2)` failed two tests: `tests/test_daemon_application.c:2256` and `:2380`, both `ASSERT(sensitive_blocked)`. Both tests set `HOME` to the `cbm_canonical_path` spelling of a directory (native backslashes on Windows) and expect a session rooted there to be refused for auto-index and for watch. The first cut respelled the session root with forward slashes on the way into the session policy, and the sensitive-root / allowed-root containment walk (`ws_is_ancestor_or_equal`, `src/foundation/workspace.c`) compares prefixes byte-exact - so `C:/Users/...` no longer matched a `HOME` of `C:\Users\...` and `$HOME` was admitted. Wrong layer.

The follow-up commit (9846c3f1) keeps the policy roots in their canonical native spelling (`mcp.c` and the three `application.c` sites are back to what `main` runs) and folds the separator spelling at the one comparison the original defect lives in: `application_index_args_equal`, where the running job's args meet an explicit `index_repository` request's. The fold acts on the parsed copies only, so nothing the daemon stores changes spelling and every other option stays exact.

Tests: `daemon_session_context_keeps_one_spelling_of_a_root` asserted the respelling and is replaced by `daemon_session_context_keeps_the_policy_spelling_of_a_root`; the comparison is exposed via `application_internal.h` and bound by `daemon_application_index_args_compare_repo_path_separator_equivalently` (binds on every platform - `cbm_normalize_path_sep` folds everywhere). Windows was not run locally; the Windows shards are the proof for the two previously red tests.

Issue #1955, opened 2026-08-31
Linux 0.10.5: new MCP/CLI clients intermittently wait 15–30s or time out while healthy daemon serves existing clients


chris-han:
### Version

codebase-memory-mcp 0.10.5

### Platform

Linux (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

Client admission and read-only queries must remain available while indexing is running or other clients are connected; only
mutation of the same project index should serialize.

### Reproduction

  This failure occurs during daemon client admission, before any graph query or
  index access, so it does not depend on the code being indexed.

  Environment:

  - OS: Linux / WSL2
  - codebase-memory-mcp: 0.10.5
  - Multiple MCP clients: Codex sessions connected to the same per-user daemon
  - Daemon mode: session-managed

  Steps:

  1. Start at least three Codex/MCP sessions configured to use
     `codebase-memory-mcp`.

  2. Confirm that the shared daemon is healthy:

     ```bash
     codebase-memory-mcp daemon status

  Output during the failure:

     daemon: active (session-managed)
       pid: 2458
       build: 0.10.5 (dfd7db016354...)
       committed clients: 3
         - pid 55306
         - pid 49179
         - pid 110190
       ui: configured at http://127.0.0.1:9749

  3. Existing MCP sessions can still execute read operations such as
     search_graph, search_code, and get_code_snippet.

  4. From another terminal, attempt to start a new client with a read-only tool
     call:

     time codebase-memory-mcp cli --json index_status

     No project argument is intentionally supplied. Once admitted, the expected
     result is an immediate validation error saying that project is required.
     This isolates daemon admission from database/index behavior.

  5. Intermittently, the new client waits for the complete 30-second startup
     deadline and exits with:

     codebase-memory-mcp: CBM daemon is active or starting but could not accept
     this client within 30000 ms
     error: CBM daemon is active or starting but could not accept this client
     within 30000 ms

  6. Repeat the identical command. In one observed retry it connected after
     approximately 15 seconds and returned the expected tool-level validation
     response:

     {
       "error": "missing required argument: project",
       "hint": "Pass the project as the \"project\" argument..."
     }

  Expected behavior:

  A healthy daemon with 3 committed clients should admit the new client promptly.
  The source configures HOST_MAX_CLIENTS = 64. Read-only client admission should
  not wait for or be blocked by index mutation coordination. Only concurrent
[...]

Issue #1830, opened 2026-08-25
Linux: ancestor walk fails unconditionally in Docker userns-remap containers (single-uid mapping, `/` owned by nobody/65534)


zhaoops:
# Linux: ancestor walk fails unconditionally in Docker userns-remap containers (single-uid mapping, `/` owned by nobody/65534)

## Version

Would affect any version from v0.10.0 onward (the daemon coordination layer). Not currently installed — cannot run due to this issue.

## Platform

Linux (x64): Docker Desktop 29.x on Windows 11 WSL2, container running Debian 13 (trixie), kernel 6.18.35.2-microsoft-standard-WSL2

## Install channel

Binary (attempted previously with v0.10.6; same issue expected on v0.10.8)

## What happened, and what did you expect?

Every cbm process exits immediately with `secure CLI coordination could not be created (endpoint)`. The ancestor walk rejects every possible directory path because **no ancestor directory from `/home/kirocrew` to `/` is owned by uid 0 or uid 1000** — they are all owned by uid 65534 (nobody).

**Expected:** cbm should be able to run in a standard Docker container that uses user namespace remapping (a single-uid mapping), which is a common production Docker security pattern.

## Root cause

Docker's `userns-remap` feature (or equivalently, Podman's rootless mode) creates a user namespace where only a single uid is mapped:

```
# /proc/self/uid_map
      1000       1000          1
```

This means:
- The process runs as uid 1000 (kirocrew)
- uid 0 does **not exist** inside the container namespace
- All filesystem objects owned by uids outside the mapped range appear as `nobody` (65534) per kernel overflow behavior

The directory ownership chain:

```
mode=755  owner=65534:65534  /          ← nobody (unmapped host root)
mode=755  owner=65534:65534  /home      ← nobody
mode=700  owner=1000:1000    /home/kirocrew  ← ✅ current user
mode=1777 owner=65534:65534  /tmp       ← nobody (not uid 0!)
```

The ancestor walk (`posix_directory_parent_secure`) requires each ancestor to be owned by either:
1. The current process euid (1000), OR
2. uid 0 (root)

Neither condition is satisfiable for `/`, `/home`, or `/tmp` because:
- uid 0 doesn't exist in the namespace (no mapping)
- These directories show owner uid 65534 (the kernel's overflow uid for unmapped uids)

The `/tmp` sticky-bit special case also fails: it requires root-owned (uid 0) + sticky, but `/tmp` is uid 65534 + sticky.

[...]

Issue #1533, opened 2026-08-11
Windows: ancestor walk refuses app container / capability SIDs, so 0.10.0 cannot start inside a containerized host


Kiborgik:
## Summary

On Windows, `win_private_directory_tree_secure` walks every ancestor of the runtime directory and refuses any that grants mutation rights to an identity it does not trust. App container and capability SIDs (`S-1-15-2-*`, `S-1-15-3-*`) are not on the trusted list, so when the process runs inside a containerized host application the walk refuses the container's own `AppData` and no endpoint can be created.

The result is that 0.10.0 cannot start at all from inside such a host, with no remedy available to the user: the offending ACE is on a directory the container owns, not on anything the user can edit.

## Evidence

Instrumented build from `main`, printing `ipc_validation_detail_buffer` at the failing gate in `cbm_daemon_ipc_endpoint_new`:

```
DIAG gate=final dir=1 wide=1 names=1 pipe=1 mutex=1 tree_secure=0 err=0
     runtime_dir=C:\Users\petro\AppData\Local/cbm-daemon-c888acc3ae367a1e
DIAG detail=C:\Users\petro\AppData: DACL entry 0 grants mutation rights 0x000d0152
     to untrusted identity (other S-1-15-3-3557520199-3666692283-3112367039-3524159787-2791857073-3163583606-3692855932)
```

Every other condition in that gate passes. The refusal is on an **ancestor** (`C:\Users\petro\AppData`), not on the runtime directory itself.

Without instrumentation the only surfaced message is:

```
codebase-memory-mcp: secure CLI coordination could not be created (endpoint)
```

`CBM_LOG_LEVEL=debug` adds nothing, because the failure happens before the log layer is initialized.

## Environment

- Windows 11 26200, non-elevated standard token
- codebase-memory-mcp 0.10.0 (also reproduced with a local build of `main`)
- Host application runs the MCP server inside an app container, so its `%LOCALAPPDATA%` resolves to a container-owned object whose DACL carries the capability SID above
- The same binary run from an ordinary terminal on the same machine works: that `AppData` has no such ACE

## Version boundary

0.9.0 works from inside the same host. The ancestor walk is part of the strict permission gate introduced in 0.9.1-rc (#1416), so the regression window is 0.9.0 → 0.9.1-rc.

## Why this differs from #1416

[...]

PR #2072, opened 2026-09-06
feat(mcp): add CBM_IN_PROCESS to serve stdio MCP without the daemon

pmaxvsbobo:
## What does this PR do?

Adds an opt-in `CBM_IN_PROCESS` environment variable that serves an MCP session
in-process over stdio, without starting or connecting to the coordination daemon.

**Why it's needed:** on hosts whose sandbox denies socket syscalls, the daemon
handshake can never complete. A macOS seatbelt profile that permits filesystem
access but applies `(deny network*)` makes both `bind()` and `connect()` fail
with `EPERM` — `AF_UNIX` included — so there is no reachable rendezvous and no
alternative transport. Today that surfaces as a 30s hang and a misleading
*"CBM daemon is active or starting"* error;

The change is small because the server already existed: `cbm_mcp_server_run()`
is a complete socket-free stdio loop (zero socket/IPC references in
`src/mcp/mcp.c`) whose only caller was `tests/test_mcp.c`. This routes MCP
clients to it behind a gate, so default behaviour is unchanged.

Measured under an active sandbox: **31.1s failure → 4.6s `initialize` result**.

**Trade-off** (documented in `README.md` and `docs/CONFIGURATION.md`): no daemon
means no cross-session coordination — no shared watchers, no shared indexing
jobs, no UI, no exact-build admission barrier.

## Checklist

- [X] New behavior is covered by a test (reproduce-first for bug fixes)


DeusData:
Thanks for this, and for measuring it (31 s to 4.6 s is a real pain point, and the analysis of why the failure is misreported is right: the connect result carries no errno, so an `EPERM` connect is classified as a reserved endpoint and retried blindly). Before we decide on the direction I want to understand the use case better, because a daemon opt-out is trickier than the diff suggests: the daemon is where the cross-session mutation lease, the index executor, the project-mutation guard and the exact-build admission live, and the in-process server as wired here gets none of them while background tasks stay on, so an in-process `index_repository` writes the shared cache without the lease that protects a concurrent daemon session on the same machine. Could you tell us more about:

1. **The environment.** Which host, sandbox or agent product applies `(deny network*)` to the MCP server process, and is that its default profile or one you wrote? A concrete reproduction (profile snippet or product name plus version) would help us judge how common this is.
2. **Why AF_UNIX cannot be allowed.** Most seatbelt profiles that deny network still permit unix-domain sockets to a specific path (`(allow network* (local unix-socket))` style rules, or a path-scoped allow). Is that ruled out where you are, and if so by whom?
3. **What you need inside that sandbox.** Is it read-only querying of an index built elsewhere, or does the session also have to index and re-index? If reads are enough, a read-only in-process mode sidesteps the lease problem entirely; if writes are needed, how would you propose to coordinate with a daemon session on the same cache?
4. **Whether the real defect alone would do for now.** If the connect failure surfaced immediately as "the sandbox denies unix sockets: …" instead of a 30 s "daemon is active or starting", would that unblock you while the design question is settled? That fix (errno propagation in the connect result) is small, clearly right, and we would take it as its own PR straight away.

No decision yet on the mode itself; the maintainer wants to think about it with your answers in hand. Thank you again for the careful write-up.

pmaxvsbobo:
Thanks — you're right about the lease, and it was worse than "no lease is acquired". Fixed in`c72cb8b`; the mode is now read-only.
`mcp_project_mutation_begin()` is `return !srv->mutation_begin || srv->mutation_begin(...)`, so an  **unset** guard fails *open* and the mutation proceeds. I'd wired only `set_tool_profile` and  `set_session_context`, left `background_tasks` at its default `true`, and installed no mutation guard, no try guard and no index executor. Every other construction site in the tree already gets this right — `daemon/application.c`, `ui/http_server.c`, and `main.c`'s own local-CLI path in the same function I edited. I should have mirrored them.
 I verified it rather than taking it on description: with the three calls removed,  `index_repository` on an in-process session doesn't merely skip the guard, it returns  `{"status":"indexed"}` with `isError:false` and writes the cache. So this was live, not theoretical. Thank you for catching it.
 
What `c72cb8b` does — mirroring `ui/http_server.c`:
- `background_tasks(false)`, so `maybe_auto_index()` can't index on the initialize path
- a mutation guard that refuses, with the **try guard deliberately left NULL**: with `mutation_begin` set and `mutation_try_begin` NULL, `mcp_project_mutation_try_begin()` also returns false, so opportunistic writes during a read are refused too
- an index executor that rejects with a message naming the reason, rather than letting the guard report "blocked by an active index" when nothing is blocking
The test now asserts both that the call is refused **and** that no `.db` appears under `CBM_CACHE_DIR` — an error message alone could be emitted after a partial write, and only the second check distinguishes a refusing guard from an absent one. I confirmed that assertion fails against a build with the three calls removed.

pmaxvsbobo:
Answers to your questions:
 **1. The environment.** An enterprise-managed build of Claude Code, deployed by my employer's
security org — **not** a profile I wrote, and not one I can change. It applies a seatbelt profile
to every subprocess it spawns, MCP servers included: `(allow default)` for the filesystem,
 `(deny network*)`, then four narrow re-allows. The profile file is regenerated from a compiled-in
constant at every launch and sha256-verified, so local edits are reverted; changing the policy
requires review by the team that owns it. I'd rather not paste my employer's profile wholesale
in a public issue, but I can share the exact deny/allow lines if that helps. The shape is:
filesystem unrestricted, all networking denied, a handful of vendor-owned control sockets
re-allowed by exact filename.

 I can't tell you how common this is beyond my own org. But the *pattern* — sandbox the agent's
subprocesses, deny networking, leave the filesystem alone — isn't exotic, and CBM is unusually
exposed to it because coordination is mandatory and the only transport is AF_UNIX. #1533/#2023
 (Windows app-container ACLs) and #1830 (Docker userns-remap) look like the same class arriving
through different mechanisms.
 
**2. Why AF_UNIX can't be allowed.** It's ruled out, by the team that owns the profile. Two specifics that matter:
 - `network*` covers `network-bind` as well as `network-outbound`, so `bind()` **and** `connect()`  are both `EPERM` for `AF_UNIX`. There's no connect-only loophole.
 - The only unix-socket re-allows are `(allow network-outbound (remote unix-socket (regex ...)))` matching an **exact filename pattern** for the vendor's own control sockets. They're not directory-scoped, so no value of `CBM_RUNTIME_DIR` can produce a matching path — CBM hardcodes both the `cbm-daemon-<uid>` subdirectory and the `cbm-<hash>.sock` filename.
Measured inside one sandboxed process: `connect()` to a live socket under `$HOME` → `EPERM`;
`connect()` to the CBM socket under `/private/tmp` → `EPERM`; `bind()` in `$HOME`, `/private/tmp`
and `$TMPDIR` → `EPERM`, 8/8. The loopback UI port is `EPERM` too.

There is an escape hatch — the vendor lets a user exempt a named binary from the sandbox entirely — but I don't think that's the better answer. It trades a scoped capability for removing containment altogether from a process that indexes source code, and it's a per-machine manual step the project can't rely on.

 **3. What I need inside the sandbox — reads only.** This is the useful answer, and it's why your suggestion is simply better than what I sent. Querying an index built elsewhere is enough.  Indexing already happens outside the sandbox, from an ordinary terminal where the daemon works normally and owns the watcher and the lease. The sandboxed session only needs `search_graph`, `search_code`, `get_code_snippet`, `trace_path` and friends against the cache the daemon built. So the read-only mode in `c72cb8b` is not a concession — it's what I actually needed, and it removes the coordination question rather than answering it. If you'd prefer this expressed through the existing `--tool-profile` machinery rather than a new env var, say so and I'll rework it; a read-only profile is arguably where this belongs.

 **4. Would the errno fix alone unblock me?** Not on its own — an accurate instant error still leaves the tools unusable, so I'd still be reaching for a terminal. But it's worth far more than its size, and I'd rather it landed first regardless. It cost me most of a day: the 30 s wait lands *just* past the 30 s MCP handshake budget my host imposes, so all I saw was "connection timed out after 30000ms" — the actual error never arrived. And when it does arrive it actively misleads: the daemon was healthy throughout, and a *running* daemon is what makes the wait last the full 30 s, since holding the cohort lock is what makes presence read COORDINATED and the `EPERM` classify as RESERVED/waitable. `daemon start` made it worse, which is the opposite of what the message implies.

I'll send that as its own PR: an `os_error` field on `cbm_daemon_runtime_connect_result_t`, and a fast-fail for non-retryable values (`EPERM`, `EACCES`) in `cbm_daemon_bootstrap_classify_failed_connect()` with a message naming the cause. Independent of whatever you decide here, and it should help #1955 too — that report hits the same message, and it currently can't distinguish "denied" from "contended" from "still starting".
 
On CI: DCO is fixed and both commits are signed off. The `clang-format` failures in `cypher.c`, `runtime.c`, `version_cohort.c`, `frontend.c` and `cli.c` are pre-existing and not from this  branch — this PR touches four files, none of them those. I only hit them locally because Homebrew ships clang-format 23 while you pin 20. The one violation that *was* mine (an over-wrapped `fprintf` in `src/main.c`) is fixed.

pmaxvsbobo:
@DeusData I think this is needed for making codebase-memory-mcp more popular on enterprise env, can u pls prioritize reviewing it?

DeusData:
Thank you for clarifying the read-only and sandbox use cases. Those answers address the use-case questions already raised. We need more time to decide the in-process/read-only execution contract before making an integration decision; there is no additional questionnaire or speculative rebase requested from you at this stage.

PR #2073, opened 2026-09-06
fix(tests): create nested pipeline fixtures reliably

### LINKED ISSUE #2034: tests: write_temp_file creates only one directory level — nested-package fixtures index nothing and negative assertions pass vacuously
## Summary

`write_temp_file` in `tests/test_pipeline.c:9764-9776` creates only **one** directory level (`/* mkdir -p (simple version, one level) */` → a single `cbm_mkdir(parent)`). A fixture that uses nested package directories — `registry/batch/cronjob/strategy.go`, the shape of most real Go/Java/C# layouts — silently writes nothing but the top-level files, the pipeline logs `pipeline.discover files=1`, and any **precision** assertion ("this edge must NOT exist") passes vacuously.

## Why it matters

This bit a fixture while reproducing a cross-package call-resolution defect: the first version of the test passed on the unfixed tree because nothing had been indexed. A negative assertion over an empty graph is always green. The class of tests most at risk is exactly the class we are adding most of (false-positive CALLS edges, wrong-package bindings).

## Proposed fix

1. Make `write_temp_file` a real `mkdir -p` (iterate over every `/` in the relative path).
2. Grep the suite for negative assertions (`ASSERT_FALSE(... edge_exists ...)`, `ASSERT_EQ(count, 0)`) whose fixtures use nested paths and add a positive existence check first — the committed reproduction for the cross-package shape does this with a `go_node_with_qn_tail_exists` helper that asserts both same-named methods are Method nodes before asserting the absence of the edge.

astandrik:
## What does this PR do?

The pipeline test writer only created the immediate parent directory, so fixtures with multiple missing directory levels could silently disappear. Use `cbm_mkdir_p` and require the expected source nodes before negative edge checks.

Add regressions for exact file contents, nested Go indexing, a regular file blocking directory creation, and exact node counting. Cover deep paths, Unicode, overwrites, and 1023/1024/1025-byte boundaries; make envscan and pkgmap rely on the writer for nested directories.

## Verification

The final commit `aec9bfe3` changes only fixture comments; executable text and line counts are unchanged. The completed CI results below are from `8e29e156`; checks on the latest commit are shown in GitHub.

- [Verified source `8e29e156` CI](https://github.com/DeusData/codebase-memory-mcp/actions/runs/34115081932): all 137 suites per platform passed — Linux x64/ARM64 **7,792 each**, macOS Intel/ARM64 **7,788 each**, Windows x64 **7,667**; **0 failed**. All ten shard manifests and totals were audited. `ci-ok`, lint and the existing sanitizer jobs also passed.
- [Native Windows ARM64](https://github.com/astandrik/codebase-memory-mcp/actions/runs/34117601772/job/101727946559): full `scripts/test.sh CC=clang CXX=clang++`, **141 suites, 7,833 passed, 0 failed, 69 platform skips**. Native ARM64 binary, trap-UBSan flags, unchanged source and every suite log verified.
- On the original fix `5abd77f5`: RED/GREEN for directory creation and exact contents; **15 controls passed; all 32 mutations failed at the intended assertions**. Full macOS/Linux tests and CI lint passed; known baseline diagnostics are listed below. The fix's added/deleted lines are identical after the main merge.
- Cases include depth 0/1/2/3/8, existing/partial parents, spaces/Unicode, 480-byte paths, empty files, line endings, overwrites, blocked parents, and 1023/1024/1025-byte boundaries with extra/truncated/changed contents.

Known unrelated issues: local Linux reports six UBSan diagnostic lines reproduced without the patch; full clang-tidy reports 6,754 diagnostics in unchanged files. The [fixed Windows comparison](https://github.com/astandrik/codebase-memory-mcp/actions/runs/34117601772/job/101727946770) completed 22 attempts: the same cold-storm endpoint failure occurred **2/11 on base and 2/11 on candidate**, with no timeouts or precondition skips. That auxiliary job remains red; its strict cleanliness check also flagged only generated `graph-ui/tsconfig.tsbuildinfo` in both builds. C/guard source hashes remained unchanged. The upstream Windows guard passed on `8e29e156`.

Fixes #2034

## Checklist

- [x] Canonical tests pass locally; sanitizer limitations are disclosed above.
- [x] CI lint (`scripts/lint.sh --ci`) and the memory analyzer pass locally.
- [x] New behavior has reproduce-first regressions and mutation checks.


DeusData:
Thank you for the focused nested-fixture fix in tests/test_pipeline.c. This is queued for review, and we need a little more time to check the fixture lifecycle and failure handling before giving a decision.

DeusData:
Here is the review this was queued for on 2026-09-09. Eleven days is too long for a test-only change, and I am sorry — you were told it was waiting on us, and then it kept waiting.

**The verdict: this is better than its title, and I want to be specific about why.**

The headline is reliable nested fixtures, and `cbm_mkdir(parent)` → `cbm_mkdir_p(parent, 0755)` does fix that — the old comment cheerfully admitted it was a "simple version, one level", which is a bug waiting for the first two-level fixture path. But the more valuable half of this diff is the assertion hardening, and it addresses a failure mode this repository keeps getting bitten by.

**`count_nodes_named(s, project, "err") >= 1` is not a test.** It passes if *any* node anywhere in the graph happens to be called `err`, in any file, with any label. It would keep passing if the node it was actually written to check disappeared entirely, as long as some unrelated `err` survived. Replacing it with

```c
fixture_node_count(s, project, "state/state.go", "err", "Field") >= 1
```

pins the file path and the label, so the assertion can only be satisfied by the thing it is about. That is the difference between a test and a green light.

**And you found an inert fixture.** This one is the part I would not have caught by reading:

```c
-  int event = 0;          /* block-local */
+  static int event = 0;   /* file scope */
```

Block-local C variables are not emitted as nodes, so the cross-language collision guard — the whole point of that probe — had nothing to collide with. The test was passing because it was checking nothing. A verified-broken fixture getting fixed rather than deleted or worked around is exactly the right call.

**Verified rather than assumed.** I built your head and ran the suite you touch: **280 pipeline tests, 0 failures**. Your strengthened assertions hold against what production actually emits, which is the thing that could have gone wrong when you tightened them.

**The `test-lsan-macos` red is not yours.** It fails at `tests/test_convergence_probe.c:636 ASSERT(grpc >= 1)` — a different test file from the only file you touch. I ran that suite three times locally and it passed each time, so it does not reproduce here; I have re-run the CI job against your identical commit to confirm it is nondeterministic rather than real. I am not treating a green rerun as the end of it either — a required check that passes on retry is a defect of ours to attribute, not something to shrug past — but it will not block you.

Once that check settles I will merge this. Thank you for the patience, and for tightening assertions that were quietly protecting nothing.

Issue #2057, opened 2026-09-04
test-windows-guards: test_daemon_stability turns setup failures and timeouts into REGRESSION verdicts on unrelated PRs


DeusData:
## Summary

`test / test-windows-guards` is a required check, and it goes red on pull requests whose diffs cannot touch it. The red always comes from `tests/windows/test_daemon_stability.py`, and it comes from three mechanisms in the guard's own harness rather than from the product. Until they are fixed the leg is a lottery: a green needs a rerun that changes nothing but the dice, which is exactly what the CI-determinism rule forbids.

## Observed (unrelated diffs, same leg, same file)

| PR | diff | run / job |
|---|---|---|
| #1875 | Cypher | job 99662589846 (line 820: `SETUP FAIL: ASCII baseline did not index: {nodes: None, edges: None, definition_nodes: 5}`) |
| #1876 | Windows Unicode paths | reddened 2026-08-28 |
| #1896 | Swift extraction | reddened 2026-09-01 |
| #2002 | Cypher SKIP/LIMIT | `section_start_status_port`: "the daemon did not accept the UI configuration; browser was not opened" |
| #2055 | `scripts/run-test-wave.py` + one contract test | run 33892540012, job 101150446542: `section_crash_recovery` ("the daemon did not accept the UI configuration; browser was not opened") and `section_cold_storm` ("cold-storm client 2 failed (racing daemon spawn): secure CLI coordination could not be created (endpoint)") |

## Mechanisms (verified in the tree on 2026-09-03; line numbers as of that audit)

1. **A setup failure is reported as a regression.** The file header of `test_daemon_stability.py` documents the exit contract `0 == all sections green, 1 == regression, 2 == setup error`, and `scripts/test-windows.ps1` (around line 190) treats exit 2 as a documented precondition skip. But every setup-failure branch in the stability guard (`return False` at roughly lines 206 / 218 / 265 / 319) exits **1**, while its sibling guards return **2** for the same class of failure. The same daemon-start failure is therefore an invisible skip in one guard and a REGRESSION in this one. That asymmetry is why the red reads like a daemon bug.
[...]

PR #2075, opened 2026-09-06
fix(watcher): back off re-forking a hard-failing index worker

### LINKED ISSUE #2015: Watcher re-forks a permanently-failing index worker forever — no backoff on rc < 0 (2233 workers in 4h43m); gap in #937
### Version

codebase-memory-mcp 0.10.8

### Platform

macOS on Apple Silicon

### Install channel

GitHub release archive

### Binary variant

standard

## Summary

When a watcher-triggered index worker fails with a **hard** error (`index_fn` returns `rc < 0`), `poll_project()` has no error backoff, no failure counter, and no circuit breaker. The baseline stays uncommitted — correct, per #937, so the change is not lost — but the next poll re-forks a worker that fails identically. If the failure is persistent rather than transient, this becomes an unbounded fork loop at the adaptive poll cadence.

On this host that produced **2233 consecutive failed index workers over 4h43m** — one every ~7.5 s, sustained — across two watched projects. Each iteration forks the full 295 MB binary. The daemon stayed up and answered every MCP query normally throughout, so from the client side CMM simply felt intermittently unresponsive with no error surfaced anywhere.

I expected a persistent index failure to back off and eventually stop re-forking, logging the reason once, rather than retrying forever at 5–7 s intervals.

## The loop

`src/watcher/watcher.c` (`poll_project`, ~L1428-1455):

```c
int rc = ctx->w->index_fn(s->project_name, s->root_path, ctx->w->user_data);
if (rc == 0) {
    /* commit baselines, refresh file_count + interval_ms */
} else if (rc > 0) {
    /* Busy-skip: baseline stays uncommitted, next poll retries. */
    cbm_log_info("watcher.index.retry", "project", s->project_name);
} else {
    cbm_log_warn("watcher.index.err", "project", s->project_name);   /* ...and that is all */
}

s->next_poll_ns = ctx->now + ((int64_t)s->interval_ms * US_PER_MS);
```

Two things compound:

1. The `rc < 0` branch has no state at all — nothing records that this project just failed, so the next poll is indistinguishable from the first.
2. `interval_ms` is only refreshed inside the **success** branch, so a project that starts failing keeps whatever interval it last had. With `POLL_BASE_MS 5000` (L113) that can pin the retry at ~5 s indefinitely.

Busy-skip (`rc > 0`) has real backpressure by design. Hard-error has none.

## Observed

Log directory: `~/.cache/codebase-memory-mcp/logs/`

- **2233** orphaned `.worker-log-*` files, 14 MB total.
- Every single one ends with the identical line:
  ```
  CBM index worker could not start: a pre-coordination or unverified CBM generation is active
  ```
- mtime histogram: one outlier at `2026-09-01 09:53:50`, then a continuous storm from **13:32 to 18:15:23** at a steady **8 workers/minute**.
- Two watched projects only:

  | repo_path | failed workers |
  |---|---|
  | `/Users/ahby/Sources/add-translate-buttons` | 2039 |
  | `/Users/ahby/Sources/liverpool-cleanup` | 194 |

  Within the storm window the daemon log contains **zero** `reap outcome=clean` events and **no** `watcher.changed` for any other project — these two were simply the only repos being actively edited. So the evidence does not distinguish "these two projects were poisoned" from "the endpoint was globally poisoned and only these two had changes to trigger it"; the latter is at least as consistent with the data.

Daemon log (`cbm-daemon.log`) over the same period:

| event | count |
|---|---|
| `watcher.changed` | 2923 |
| `index.supervisor.reap` (all `exit_nonzero exit_code=1`) | 2945 |
| `index.supervisor.worker_failed` | 2233 |
| `watcher.index.err` | 2226 |

repeating verbatim:

```
watcher.changed → index.supervisor.reap outcome=exit_nonzero exit_code=1
  → index.supervisor.worker_failed → watcher.index.err → watcher.changed → …
```

The storm ended only when the daemon was restarted. It has recurred across daemon generations.

## Relationship to #937

#937 fixed the loop where the watcher re-indexes a dirty worktree **successfully** over and over (write amplification), by staging observations in `pending_*` and committing baselines only after a successful reindex. The comment at `watcher.c` ~L1223 cites it.

This report is the gap that fix left: the loop where the reindex **never succeeds**. The at-least-once guarantee is right and should stay; what is missing is that an unsuccessful attempt is retried at full cadence forever instead of decaying.

## The trigger (secondary — no repro, filed for context only)

The underlying failure is that the forked index worker aborts at `src/main.c` ~L2745:

```c
int worker_seal_status = cbm_daemon_ipc_local_transition_seal_legacy(worker_transition);
```

No `version_cohort.uncoordinated_daemon` line appears in any of the 2233 worker logs, which means the return was **-1**, not 0 — so one of the two `posix_named_lock_is_retained` checks, `cbm_daemon_ipc_lifetime_reservation_probe`, or `posix_stale_generation_cleanup_locked` in `cbm_daemon_ipc_local_transition_seal_legacy` (`src/daemon/ipc.c` ~L3256) returned -1. All are hard conditions, so every fork hits it identically. Notably the **daemon itself was unaffected** — MCP queries were served normally the entire time; only forked workers were poisoned.

Within the storm span the daemon served **57 read-only MCP calls successfully** (`search_code`, `search_graph`, `get_code_snippet`, `query_graph`, `trace_path`, `index_status`) — so the daemon process itself was healthy throughout. But it also logged **5 × `tool=index_repository status=error`**: the user-facing `index_repository` MCP tool failed too, for the whole window.

That is the part worth emphasising. The seal failure is not confined to background watcher workers — it takes out *every* index attempt, foreground included, while leaving all read paths working. From the client's seat CMM answers queries normally but silently cannot reindex anything, for hours, until the daemon is restarted. There is no error a user would ever see unless they go read `.worker-log-*` by hand.

This looks like the same coordination-state family as #1757 and #1760, but neither covers this manifestation: both of those block startup/install *visibly*, whereas here everything appears healthy while all indexing — background and explicit — dies silently and permanently.

I could not reduce this to a repro and am deliberately **not** filing it as a separate issue. The backoff fix above is worth having on its own merits precisely because it makes any future instance of this — or any other persistent worker failure — self-limiting instead of machine-saturating.

## Suggested fix

In `project_state_t`, add a consecutive-hard-failure counter. In the `rc < 0` branch:

- increment it, and schedule `next_poll_ns` with exponential backoff (e.g. `interval_ms << min(n, k)`, capped at a few minutes) instead of the plain interval;
- reset it to 0 on `rc == 0`;
- after N consecutive failures, stop re-forking for that project until the next daemon generation or an explicit `index_repository`, logging the give-up once.

Roughly 15 lines in one file.

## Secondary: this was much harder to diagnose than it should have been

Two small things would have made the above self-diagnosing rather than a mtime-forensics exercise:

1. **`cbm-daemon.log` has no timestamps on any line.** The entire 4h43m timeline had to be reconstructed from `.worker-log-*` file mtimes, because the daemon log itself cannot say when anything happened or how fast.
2. **`watcher.index.err` logs only `project=`** — not the `rc`, and not the worker log path. The supervisor logs `log=<path>` on `worker_failed` one line earlier, but the watcher's own error line drops it, so the daemon log alone never explains *why* a worker died.

Also minor: failed workers appear to leak their `.worker-log-*` files (2233 accumulated, 14 MB). Successful ones seem to be cleaned up.

## Reproduction

I do not have a clean repro for the underlying seal failure. The retry-storm behaviour itself is structural and readable directly from `poll_project`: any `index_fn` returning a persistent `rc < 0` reproduces it. A synthetic test that stubs `index_fn` to always return -1 and asserts that poll intervals grow (and that forking eventually stops) would cover the fix.

### Relevant log output

```text
$ ls ~/.cache/codebase-memory-mcp/logs/.worker-log-* | wc -l
    2233

$ for f in .worker-log-*; do tail -1 "$f"; done | sort | uniq -c
2233 CBM index worker could not start: a pre-coordination or unverified CBM generation is active

$ stat -f '%Sm' -t '%Y-%m-%d %H:%M' .worker-log-* | sort | uniq -c
   1 2026-09-01 09:53
   1 2026-09-01 13:32
   4 2026-09-01 13:34
   8 2026-09-01 13:35
   8 2026-09-01 13:36
   ...  (8/min sustained) ...
   7 2026-09-01 18:13
   7 2026-09-01 18:14
   3 2026-09-01 18:15

$ grep -h -o '"repo_path":"[^"]*"' .worker-log-* | sort | uniq -c
2039 "repo_path":"/Users/ahby/Sources/add-translate-buttons"
 194 "repo_path":"/Users/ahby/Sources/liverpool-cleanup"

$ cat .worker-log-BmwDoV
{"level":"control","event":"index.worker.start","version":"0.10.8","build":"996bad5fe6fb89c0c50363d87f03dc78f6e536aeea746c06679a294d14eca435","pid":"23647","repo_path":"/Users/ahby/Sources/liverpool-cleanup","args":"{\"repo_path\":\"/Users/ahby/Sources/liverpool-cleanup\"}"}
level=info msg=version_cohort.claimed_unheld build=996bad5fe6fb89c0c50363d87f03dc78f6e536aeea746c06679a294d14eca435
CBM index worker could not start: a pre-coordination or unverified CBM generation is active

$ # ordering check within the storm span (daemon log lines 1996..10977)
$ sed -n '1996,10977p' cbm-daemon.log | grep -c 'reap outcome=clean'
0
```

halindrome:
Closes #2015

## What

`poll_project()` treats a hard index failure (`index_fn` returning `rc < 0`) as a transient event: it emits `watcher.index.err` and schedules the next poll at the plain interval. A *persistent* start failure — a poisoned coordination endpoint, an unreadable DB — therefore re-forks an identically-failing worker at the normal cadence for as long as the daemon lives (2,233 workers in 4h43m in the report).

This adds a per-project consecutive-failure counter with a doubling, capped backoff on that path only:

- `index_failures` on `project_state_t`, incremented on `rc < 0`, reset to `0` by any successful reindex.
- `cbm_watcher_index_backoff_ms(interval_ms, consecutive_failures)` — pure function, doubling per consecutive failure up to a shift cap and an absolute ceiling, so a permanently-failing project decays to occasional retries instead of every-poll retries.
- `watcher.index.err` now carries `rc` and `consecutive` so the streak is visible in the log rather than inferred from line count.
- `cbm_watcher_index_failure_count()` accessor so the counter is testable without reaching into the struct.

## The watcher guarantee is preserved

Per the maintainer note on #2015 — a failed observation must not be silently committed. It isn't: the `rc < 0` arm only logs and lengthens the next poll. It does not touch `last_dirty_sig`, `pending_dirty_sig`, or `pending_head`. The baseline stays uncommitted exactly as #937 intends, so the pending change is still picked up whenever the underlying failure clears; only the retry *cadence* decays. The busy-skip (`rc > 0`) path is untouched.

## Field evidence

The patched build has been running locally since 2026-09-02. On 2026-09-06 an unrelated daemon-coordination fault made every index worker fail to start — the same permanent-failure class as the original report — across three watched projects.

| build | `watcher.index.err` events for a permanent failure |
|---|---|
| before (unpatched) | 2,226 |
| after (this patch) | 14 total, across 3 projects; longest streak 6 |

The underlying fault was not fixed by this change and is not meant to be — the point is that the watcher stopped re-forking within seconds instead of accumulating thousands of doomed workers, and the streak counter in the log made the fault diagnosable at a glance.

## Tests

`tests/test_watcher.c` adds coverage for the backoff arithmetic (monotonic, capped, saturating), the counter's increment/reset contract, and the guarantee that a failing poll leaves the baseline uncommitted.

## Notes

Opened as a draft. Three QA rounds were run against this diff before submission (commits `4fc5999a`, `61ccd905`, `08642202`); reports follow as comments.

https://claude.ai/code/session_01Y9gerhYqHmbgnCxmqpVKFc

### DIFF
--- a/src/watcher/watcher.c
+++ b/src/watcher/watcher.c
@@ -75,6 +75,14 @@ typedef struct {
     uint64_t last_dirty_sig;       /* committed dirty-state signature */
     uint64_t pending_dirty_sig;    /* observed at check time */
     char pending_head[CBM_SZ_128]; /* HEAD observed at check time */
+    /* Consecutive hard index failures (index_fn < 0). A hard error is
+     * usually persistent — a poisoned coordination endpoint, an unreadable
+     * DB — so retrying it at the plain poll interval re-forks a worker that
+     * fails identically, for as long as the daemon lives. #937 deliberately
+     * leaves the baseline uncommitted so the change is never lost; this
+     * decays the retry cadence so "never lost" does not also mean "retried
+     * forever". Reset to 0 by any successful reindex. */
+    int index_failures;
     /* Hop from root_path up to the repository root ("" when they are the same),
      * from `rev-parse --show-cdup`. Porcelain paths are repository-relative, so
      * the signature needs this to stat them. Resolved once at baseline. */
@@ -114,6 +122,16 @@ struct cbm_watcher {
 #define POLL_FILE_STEP 500 /* add 1s per this many files */
 #define POLL_MAX_MS 60000
 
+/* Hard index-failure backoff. Doubling per consecutive failure, capped, so a
+ * persistently failing project costs ~12 attempts/hour instead of ~480 while
+ * still recovering on its own within the ceiling once the cause clears. The
+ * shift cap keeps the intermediate value well inside int64 for any interval. */
+#define INDEX_FAIL_SHIFT_MAX 6
+#define INDEX_FAIL_CEILING_MS 300000 /* 5 min */
+/* Log a distinct line once the failures are clearly not transient, so the
+ * daemon log names the stuck project instead of only repeating the warning. */
+#define INDEX_FAIL_SUSTAINED 10
+
 /* Stale-root pruning (#286): a watched project whose root directory stays
  * missing is pruned — its cached DB is deleted and the watch entry removed.
  * Deletion is destructive (the DB can hold user-authored data such as the
@@ -142,6 +160,19 @@ static int64_t now_ns(void) {
 }
 
 /* ── Adaptive interval ──────────────────────────────────────────── */
+
+int cbm_watcher_index_backoff_ms(int interval_ms, int consecutive_failures) {
+    if (interval_ms < 0) {
+        interval_ms = 0;
+    }
+    if (consecutive_failures <= 0) {
+        return interval_ms;
+    }
+    int shift =
+        consecutive_failures < INDEX_FAIL_SHIFT_MAX ? consecutive_failures : INDEX_FAIL_SHIFT_MAX;
+    int64_t delay = (int64_t)interval_ms << shift;
+    return delay > INDEX_FAIL_CEILING_MS ? INDEX_FAIL_CEILING_MS : (int)delay;
+}
 
 int cbm_watcher_poll_interval_ms(int file_count) {
     int ms = POLL_BASE_MS + ((file_count / POLL_FILE_STEP) * CBM_MSEC_PER_SEC);
@@ -1429,6 +1460,7 @@ static void poll_project(const char *key, void *val, void *ud) {
         int rc = ctx->w->index_fn(s->project_name, s->root_path, ctx->w->user_data);
         if (rc == 0) {
             ctx->reindexed++;
+            s->index_failures = 0;
             /* Commit the baselines OBSERVED AT CHECK TIME — the state whose
              * reindex just succeeded. A commit/edit landing during the
              * reindex is deliberately not absorbed: the next poll sees it
@@ -1447,11 +1479,28 @@ static void poll_project(const char *key, void *val, void *ud) {
             /* Busy-skip: baseline stays uncommitted, next poll retries. */
             cbm_log_info("watcher.index.retry", "project", s->project_name);
         } else {
-            cbm_log_warn("watcher.index.err", "project", s->project_name);
+            /* itoa_buf returns one shared per-thread buffer, so two of them
+             * in one call would print the same value twice. */
+            char rc_text[CBM_SZ_32];
+            char streak_text[CBM_SZ_32];
+            if (s->index_failures < INT_MAX) {
+                s->index_failures++;
+            }
+            snprintf(rc_text, sizeof(rc_text), "%d", rc);
+            snprintf(streak_text, sizeof(streak_text), "%d", s->index_failures);
+            cbm_log_warn("watcher.index.err", "project", s->project_name, "rc", rc_text,
+                         "consecutive", streak_text);
+            if (s->index_failures == INDEX_FAIL_SUSTAINED) {
+                cbm_log_warn("watcher.index.sustained_failure", "project", s->project_name,
+                             "consecutive", streak_text);
+            }
         }
     }
 
-    s->next_poll_ns = ctx->now + ((int64_t)s->interval_ms * US_PER_MS);
+    /* Failures back off; success and busy-skip keep the adaptive cadence. */
+    s->next_poll_ns =
+        ctx->now +
+        ((int64_t)cbm_watcher_index_backoff_ms(s->interval_ms, s->index_failures) * US_PER_MS);
 }
 
 /* Callback to snapshot project state pointers into an array. */
--- a/src/watcher/watcher.c
+++ b/src/watcher/watcher.c
@@ -171,7 +171,15 @@ int cbm_watcher_index_backoff_ms(int interval_ms, int consecutive_failures) {
     int shift =
         consecutive_failures < INDEX_FAIL_SHIFT_MAX ? consecutive_failures : INDEX_FAIL_SHIFT_MAX;
     int64_t delay = (int64_t)interval_ms << shift;
-    return delay > INDEX_FAIL_CEILING_MS ? INDEX_FAIL_CEILING_MS : (int)delay;
+    if (delay > INDEX_FAIL_CEILING_MS) {
+        delay = INDEX_FAIL_CEILING_MS;
+    }
+    /* Backing off must never schedule SOONER than the project's own cadence.
+     * Unreachable today (POLL_MAX_MS < the ceiling), but clamping here keeps
+     * the function monotonic for every input rather than only for the inputs
+     * the current constants can produce — the caller's contract is "a delay
+     * that never shrinks as failures accumulate". */
+    return (int)(delay < interval_ms ? interval_ms : delay);
 }
 
 int cbm_watcher_poll_interval_ms(int file_count) {
@@ -1159,6 +1167,17 @@ void cbm_watcher_touch(cbm_watcher_t *w, const char *project_name) {
         s->next_poll_ns = 0;
     }
     cbm_mutex_unlock(&w->projects_lock);
+}
+
+int cbm_watcher_index_failure_count(cbm_watcher_t *w, const char *project_name) {
+    if (!w || !project_name) {
+        return -1;
+    }
+    cbm_mutex_lock(&w->projects_lock);
+    project_state_t *s = cbm_ht_get(w->projects, project_name);
+    int failures = s ? s->index_failures : -1;
+    cbm_mutex_unlock(&w->projects_lock);
+    return failures;
 }
 
 int cbm_watcher_watch_count(cbm_watcher_t *w) {
--- a/src/watcher/watcher.h
+++ b/src/watcher/watcher.h
@@ -92,10 +92,19 @@ void cbm_watcher_stop(cbm_watcher_t *w);
 int cbm_watcher_watch_count(cbm_watcher_t *w);
 
 /* Return a watched project's consecutive hard-index-failure count, or -1 when
- * it is not watched. Exposed because the failure state machine (increment on
- * a hard error, reset on success) is otherwise unobservable from a test: the
- * only other thing it feeds is a poll deadline, and gating on that deadline
- * needs a controllable clock the watcher does not have. */
+ * it is not watched. Exposed so the failure state machine (increment on a
+ * hard error, reset on success) can be asserted directly rather than inferred
+ * from the poll deadline it feeds.
+ *
+ * Memory visibility: this reads under projects_lock, but poll_project WRITES
+ * the counter outside that lock — it runs against a state snapshot taken
+ * while the lock was held, which is the same discipline every other
+ * per-project field in this struct already follows. Reading concurrently with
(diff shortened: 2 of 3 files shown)


DeusData:
Thank you for this one — the per-project failure counter with a capped doubling backoff on the `rc < 0` path only, leaving the dirty-signature baseline alone so #937's at-least-once guarantee survives, is exactly the shape we wanted here, and the 2,233-worker log from #2015 made the case better than any description could.

We are assembling a patch release (v0.10.9) around the install and memory reports, and this belongs in it. Whenever you consider the draft ready, please mark it ready for review and we will take it through the normal review right away. If there is a piece you are still unsure about, say so in the PR and we can work through it together rather than wait.

halindrome:
Marked ready for review — thanks for the confirmation on the shape, and glad it lands in v0.10.9.

Three QA rounds have been run against the diff and are already pushed (`4fc5999a`, `61ccd905`, `08642202`); the branch is current with `main` and mergeable.

**On the one red check:** `test / test-windows-guards` is failing, and I believe it is pre-existing rather than something this PR introduced. The failure is in `tests/windows/test_daemon_stability.py`:

```
RED: cold-storm client 0 failed (racing daemon spawn)
```

This diff touches only `src/watcher/watcher.c`, `src/watcher/watcher.h`, and `tests/test_watcher.c` — no daemon startup, no client path, nothing Windows-specific. The symptom matches #2057 ("test-windows-guards: test_daemon_stability turns setup failures and timeouts into REGRESSION verdicts on unrelated PRs") exactly. Everything else in the matrix is green, including all three `pr-smoke` platforms, tsan/msan/lsan, and the full `test-diag` shard.

Happy to re-run that job if you would like a second data point, or to work through anything else you want changed during review.

halindrome:
Could I ask for a re-run of `test / test-windows-guards` when you get a moment? I don't have rerun rights on the repo (`Must have admin rights to Repository`), and it's the only thing keeping `ci-ok` red.

For context on the run itself: it is from 2026-09-06, but it **is** against the current head SHA `08642202` — all three QA-round commits were already pushed when the PR was opened, so there is nothing stale about what it tested, only about when. Everything else in the matrix is green.

The failure remains `tests/windows/test_daemon_stability.py` → `RED: cold-storm client 0 failed (racing daemon spawn)`, matching #2057. This diff touches only `src/watcher/watcher.c`, `src/watcher/watcher.h`, and `tests/test_watcher.c`.

## Some field data that arrived since

The backoff wedged-daemon scenario recurred on my machine on 2026-09-11 — every index worker refusing to start for 427 minutes — which gave an unplanned before/after for this patch in a single `cbm-daemon.log`. The `watcher.index.err` line only carries `rc` / `consecutive` on the patched build, so the two populations are separable:

| build | `watcher.index.err` lines | projects affected | max streak |
|---|---|---|---|
| unpatched | 2226 | 1 | unthrottled |
| patched (this PR) | ~560 | 8 | 81, sitting at the 5-min `INDEX_FAIL_CEILING_MS` |

`watcher.index.sustained_failure` fired at `consecutive=10` for six projects, which is exactly the intended signal — it made the fault obvious at a glance instead of requiring a line count.

Worth being straight about the limit, though: the backoff bounds the retry *rate*, not the total. Over a 427-minute wedge the 5-minute ceiling still permits ~85 retries per project, so dead workers accumulate linearly with daemon uptime until the underlying fault is cleared by hand. That is not a gap in this PR — it is the separate root cause, which I've now written up as #2178 with the mechanism traced through `version_cohort_active_daemon_presence()`. Happy to take that one on separately if the analysis looks right to you.

DeusData:
You asked for a rerun on 11 September and got silence for nine days. I am sorry — that is on us, and the field data you posted in the meantime deserved a faster answer than that.

**Review: I have no changes to ask for.** I read the production diff line by line:

- `cbm_watcher_index_backoff_ms` is a pure function with an `int64_t` intermediate, a shift cap and an absolute ceiling, and the final clamp makes it monotonic for *every* input rather than only for the ones today's constants can produce — the comment explaining why an unreachable branch is there is exactly the kind of comment worth having.
- The counter lives on `project_state_t`, is read under `projects_lock`, saturates instead of overflowing, and is reset by any successful reindex. Busy-skip and success keep the adaptive cadence; only `rc < 0` backs off.
- It leaves the dirty-signature baseline uncommitted, so #937's at-least-once guarantee survives: "never lost" no longer also means "retried forever".
- `watcher.index.sustained_failure` fires once at the threshold, and `rc` / `consecutive` on the warning line are what made your before/after separable in a single log. 2,226 unthrottled lines for one project against ~560 across eight, sitting at the five-minute ceiling, is about as good as field evidence gets.
- The tests: six on the pure function (doubling, ceiling, monotonicity, degenerate inputs, never-sooner-than-the-interval) and state tests that drive `poll_once` with a control proving the watcher is merely gated, not dead. No sleeps, no wall-clock assertions — which matters more than usual here, because the watcher suite is where our most expensive flaky-test family lives.

**About the red.** You read it correctly: `test_daemon_stability.py` → `section_cold_storm` is a known nondeterministic guard on our side (the #2057 class), it has reddened docs-only PRs, and nothing in `src/watcher/` can reach it. I am *not* going to rerun that one job, for a reason worth stating: a rerun re-tests the merge commit GitHub computed on 6 September, and `main` has moved 398 commits since. A green there would say nothing about today's tree. What I will do instead:

1. `main` currently has two reds of its own (a linter ratchet and a worker-policy script, both fixed by #2257, which is in its last CI legs right now). Until that lands, every PR's CI is red for reasons that are not the PR's.
2. Once it is in, I bring your branch up to date with `main` — your textual merge is clean — which gives this PR a fresh, full CI run against the current tree.
3. Before that I build and run the watcher suite on the merge result locally; this week taught us twice that a textually clean merge can still be a wrong one.
4. Then it merges. This is a contained bug fix with its evidence attached, and it does not need a further decision.

If `section_cold_storm` reddens the fresh run as well, that is ours to deal with and will not be held against this PR.

Thank you for the fix, for three QA rounds you ran on your own, and for coming back with production numbers rather than a reminder.

DeusData:
Merged as a7187ad2 — thank you.

Verified on the merge result before landing: watcher suite 83/83 three runs in a row with the three `issue2015` tests green each time; with the production change reverted the test binary no longer links, so the tests bind; memory-core linter clean. It was then verified a second time as part of the combined result of the five PRs that landed together tonight (ten suites, no failures). CI on your updated head came back fully green, including the Windows guards job that had been red for reasons unrelated to you.

It ships in the next patch release. The 2,226-versus-560 log comparison you posted is going into the release notes as the reason this matters.

PR #2084, opened 2026-09-07
fix(ui): rank Accept-Language tags by q-value instead of substring scan

### LINKED ISSUE #1829: `/api/ui-config` ignores `Accept-Language` ordering and q-values — returns `zh` if Chinese appears anywhere in the list
### Version

codebase-memory-mcp 0.10.8

### Platform

Windows (x64)

### Install channel

### Binary variant

ui

# Summary

The UI language selection in the web dashboard (`http://localhost:9749/`) does not respect the user's preferred language. The `/api/ui-config` endpoint returns `"lang":"zh"` whenever **any** Chinese tag appears anywhere in the `Accept-Language` header, regardless of position or quality values. As a result, any user with a bilingual browser language list (e.g. English (United States) first, Chinese third) gets the Chinese UI even though English is their top preference.

# Impact

For a site, per RFC 9110, the server should use the *highest-preference* language it supports. In practice:
- Bilingual users (very common, e.g. en-US + zh-CN in Edge/Chrome language lists) always get the Chinese UI, which most of them did not opt into.
- The only current workaround is removing the secondary language from the browser's preferred-language list entirely — which breaks other sites that legitimately want the user's full preference list.

# Notably, the frontend already implements the correct logic

The UI bundle's own fallback path (used when the `/api/ui-config` fetch fails) parses the Accept-Language-style string properly:

```js
function JT(i, e) {
  if (e === "zh" || e === "en") return e;
  if (!i) return "en";
  const t = i.split(",").map(n => {
    const [r, ...o] = n.trim().split(";");
    const u = o.map(d => /^\s*q\s*=\s*([\d.]+)\s*$/i.exec(d)).find(Boolean);
    return { tag: r.trim().toLowerCase(), q: u ? Number(u[1]) : 1 };
  })
  .filter(({tag: n, q: r}) => n && Number.isFinite(r) && r > 0)
  .sort((a, b) => b.q - a.q)          // sort by q-value
  .find(({tag: n}) => n.split("-")[0] === "zh" || n.split("-")[0] === "en");
  return t != null && t.tag.startsWith("zh") ? "zh" : "en";
}
```

The server-side parser could reuse the same approach: parse each tag with its q-value (default 1.0), sort descending, and return the first tag whose base language is `zh` or `en`.

# Suggested fix

1. **Server (primary):** parse the `Accept-Language` header per RFC 9110 — split into weights, sort by weight descending, return `"zh"` only if the highest-preference matched tag is Chinese. (Or, more simply: return the first `zh`/`en` tag in the list's *given order*, since browsers already sort by preference.)
2. **Optional (feature):** a way to pin the UI language so bilingual users don't have to manage their browser's language list per-site — e.g. a `CBM_UI_LANG` env var (default: `auto` = header-based) or a persisted `lang` override in the UI itself.

# Environment

- Windows 11/10, daemon installed via the official `install.ps1` (binary date: 2026-08-25; no embedded version metadata)
- Browser: Microsoft Edge (Chromium). Same behavior would apply to any browser whose `Accept-Language` contains a Chinese tag after English.

# Current workaround

Remove the secondary (Chinese) language from the browser's preferred-languages list, or use a browser profile with an English-only language list.

### Reproduction

With the daemon running (default port 9749):

```bash
# Top preference is English (q=0.95), Chinese is secondary (q=0.5):
curl -H "Accept-Language: en-US;q=0.95,zh-CN;q=0.5" http://localhost:9749/api/ui-config
# => {"lang":"zh", ...}   ❌ expected "en"

# Plain ordering, en first:
curl -H "Accept-Language: en-US,en;q=0.9,zh-CN;q=0.8" http://localhost:9749/api/ui-config
# => {"lang":"zh", ...}   ❌ expected "en"

# Chinese genuinely preferred:
curl -H "Accept-Language: zh-CN,zh;q=0.9,en-US;q=0.8" http://localhost:9749/api/ui-config
# => {"lang":"zh", ...}   ✅

# English only:
curl -H "Accept-Language: en-US,en;q=0.9" http://localhost:9749/api/ui-config
# => {"lang":"en", ...}   ✅
```

The effective behavior is `if (list contains a zh tag) return "zh"; else return "en";` — order and `q` values are not considered.

### Logs

```text

```

```text

```

_No response_

### Confirmations

AmirF194:
## What does this PR do?

`detect_ui_lang()` (`src/ui/http_server.c`) picks the `/api/ui-config` language with `strstr(accept_language, "zh")`, so any Chinese tag anywhere in the header serves the Chinese UI, regardless of position or q-value. A bilingual browser list whose top preference is English (`en-US;q=0.9,zh;q=0.5`) still gets Chinese.

Per RFC 9110 12.5.4, the fix parses each tag's q-value (default 1.0), treats `q=0` as explicitly unacceptable (12.5.1), and returns the highest-ranked tag whose base language is `zh` or `en`, keeping the first occurrence on a tie. This mirrors the UI bundle's own `JT()` fallback parser (quoted in the issue), which already does this correctly as a client-side fallback when the endpoint is unreachable.

## Checklist

- [x] Tests pass locally (`scripts/test.sh --suites httpd`)
- [x] Lint passes on the changed files

## Verification

- The two new tests assert `"lang":"en"` and get `"lang":"zh"` on plain `main`; both pass on this branch. Ran both ways in a clean `ubuntu:24.04` container. Full `httpd` suite: 66 passed, 1 skipped (pre-existing, Windows-only) on both sides.
- `detect_ui_lang` has a single call site (`handle_ui_config`); no other reader needed updating.
- `cppcheck` and `clang-format --dry-run` on the two changed files: same output before and after this change, so nothing new here.
- I have not run this on Windows or macOS; the code has no OS-specific branches, so I'd expect it to behave the same there.

Fixes #1829

### DIFF
--- a/src/ui/http_server.c
+++ b/src/ui/http_server.c
@@ -122,11 +122,64 @@ static void update_cors(const cbm_http_req_t *req, int port) {
     snprintf(g_cors_json, sizeof(g_cors_json), "%sContent-Type: application/json\r\n", g_cors);
 }
 
+/* RFC 9110 §12.5.4: rank Accept-Language tags by q-value (default 1.0,
+ * q=0 excluded per §12.5.1) and return the top-ranked "zh"/"en" tag; ties
+ * keep whichever came first. */
 static const char *detect_ui_lang(const char *accept_language) {
-    if (accept_language && (strstr(accept_language, "zh-CN") || strstr(accept_language, "zh"))) {
-        return "zh";
+    if (!accept_language) {
+        return "en";
     }
-    return "en";
+
+    char buf[CBM_SZ_256];
+    snprintf(buf, sizeof(buf), "%s", accept_language);
+
+    const char *best_tag = NULL;
+    double best_q = 0.0; /* only a strictly positive q is acceptable */
+    char *cursor = buf;
+
+    while (cursor) {
+        char *comma = strchr(cursor, ',');
+        if (comma) {
+            *comma = '\0';
+        }
+
+        char *entry = cursor;
+        while (*entry == ' ' || *entry == '\t') {
+            entry++;
+        }
+
+        double q = 1.0;
+        char *semi = strchr(entry, ';');
+        if (semi) {
+            *semi = '\0';
+            char *qpos = strstr(semi + 1, "q=");
+            if (qpos) {
+                q = strtod(qpos + 2, NULL);
+            }
+        }
+
+        size_t len = strlen(entry);
+        while (len > 0 && (entry[len - 1] == ' ' || entry[len - 1] == '\t')) {
+            entry[--len] = '\0';
+        }
+
+        if (len >= 2 && q > best_q) {
+            char base0 = (char)tolower((unsigned char)entry[0]);
+            char base1 = (char)tolower((unsigned char)entry[1]);
+            bool base_ends = len == 2 || entry[2] == '-';
+            if (base_ends && base0 == 'z' && base1 == 'h') {
+                best_q = q;
+                best_tag = "zh";
+            } else if (base_ends && base0 == 'e' && base1 == 'n') {
+                best_q = q;
+                best_tag = "en";
+            }
+        }
+
+        cursor = comma ? comma + 1 : NULL;
+    }
+
+    return best_tag ? best_tag : "en";
 }
 
 static void handle_ui_config(cbm_http_conn_t *c, const cbm_http_req_t *req) {
--- a/tests/test_httpd.c
+++ b/tests/test_httpd.c
@@ -1637,6 +1637,46 @@ TEST(ui_server_ui_config_prefers_config_lang) {
     PASS();
 }
 
+TEST(ui_server_ui_config_ranks_by_qvalue_not_substring_issue1829) {
+    th_server_t ts;
+    ASSERT_EQ(th_server_start(&ts), 0);
+
+    /* English is ranked first (q=0.9 vs zh's q=0.5); a substring scan over
+     * the raw header would still match "zh" and serve Chinese. */
+    char resp[4096];
+    int n = th_http(cbm_http_server_port(ts.srv),
+                    "GET /api/ui-config HTTP/1.1\r\n"
+                    "Accept-Language: en-US,en;q=0.9,zh;q=0.5\r\n"
+                    "\r\n",
+                    resp, sizeof(resp));
+    ASSERT_TRUE(n > 0);
+    ASSERT_EQ(th_status(resp), 200);
+    ASSERT_NOT_NULL(strstr(resp, "\"lang\":\"en\""));
+
+    th_server_stop(&ts);
+    PASS();
+}
+
+TEST(ui_server_ui_config_treats_qzero_as_unacceptable_issue1829) {
+    th_server_t ts;
+    ASSERT_EQ(th_server_start(&ts), 0);
+
+    /* q=0 means "not acceptable" (RFC 9110 12.5.1), so English must win even
+     * though zh appears first in the header. */
+    char resp[4096];
+    int n = th_http(cbm_http_server_port(ts.srv),
+                    "GET /api/ui-config HTTP/1.1\r\n"
+                    "Accept-Language: zh;q=0, en\r\n"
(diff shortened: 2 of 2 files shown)


DeusData:
Thank you for contributing the Accept-Language quality-value handling and HTTP-server tests. This is queued as a UI behavior fix. We need more time to review the parsing and fallback cases before giving a decision.

AmirF194:
The macOS test-tsan failure is a pre-existing flake in daemon_runtime, unrelated to this diff (which only touches http_server.c); the sibling ubuntu-arm TSAN leg is green. Anything else you'd want changed here?

DeusData:
Thank you for the quality-value implementation and for asking about next steps. Current `detect_ui_lang` still uses a substring check for `zh`, so the HTTP-side problem remains on main. This PR is relevant to #553 and #1829, including `q=0`, supported-language matching and deterministic ties. I have no additional contributor request from this triage; the next step is maintainer review, not another routine rebase.

DeusData:
You were right, and I want to confirm it rather than leave you waiting on an unexplained red.

Your 2026-09-14 diagnosis — that the macOS `test-tsan` failure is a pre-existing `daemon_runtime` flake unrelated to a diff touching only `http_server.c` — checks out. I re-ran the failed job against the **identical commit and base**, changing nothing but the runner, and it came back **35 success, 0 failures**. Same code, different result, which is the definition of a flake and not a verdict on your work.

The signature is now recorded on our side so the next person who hits it does not have to re-derive it:

```
tests/test_daemon_runtime.c:2726  ASSERT(capacity_rejected)
daemon_runtime_connection_cap_covers_slow_hello_and_stopping_is_terminal
on test-tsan (macos-14)
```

I am also not treating a green rerun as the end of it. A required check that passes on retry is a lottery, not a fixed test, so that flake is logged as a defect of ours to attribute properly — production race versus test-side timing — rather than something to shrug at. That work is on us, not on you, and it will not block this PR.

Separately, the review note left on 2026-09-19 about `detect_ui_lang` still using a substring check for `zh` is the open item here. That one is a real question about scope, and it is yours to answer when you have time — no rush from me, and apologies that you were left chasing a red that was never yours.

AmirF194:
Appreciate you tracking down the tsan flake signature so nobody has to re-derive it.

On detect_ui_lang: this PR's diff already replaces the substring check with the RFC 9110 q-value ranking (12.5.1, q=0 excluded, ties broken deterministically), covered by the two new httpd tests. Nothing else queued on that path, the substring check only survives on main until this merges.

DeusData:
@AmirF194 — you are right: your diff already replaces the substring scan, and my 19 September note read as if it did not. That was my misreading, not an open item on your side. Thank you for answering it patiently.

I am nevertheless going to close this, and you deserve the real reason rather than a vague one: **we are relaunching the UI soon, and this change would be redundant the moment that lands.** The current graph UI and the server code that serves it — `detect_ui_lang` lives in `src/ui/http_server.c` and exists only to pick the language that UI is served in — are being replaced as a whole rather than evolved. Anything merged into that surface today has a very short life, and I would rather say so now than merge it and have it quietly disappear. I should have told you this on the 19th instead of sending you around another loop; I am sorry for that.

What I want you to know about the work itself:

- **It is correct.** Ranking by q-value with a default of 1.0, excluding `q=0`, breaking ties by order, and matching `zh`/`en` only on a tag boundary (`len == 2 || entry[2] == '-'`) is what RFC 9110 asks for, and the old `strstr(…, "zh")` would indeed have picked Chinese for `Accept-Language: en, zh;q=0.1`. Your two `httpd` tests pin exactly those cases.
- **The requirement carries forward.** "Negotiate the UI language by q-value, never by substring" goes into the new UI's requirements as written.
- **The red check on 14 September was never yours.** You diagnosed the macOS TSan failure as a pre-existing `daemon_runtime` flake from a diff that only touches `http_server.c`, and you were right; it is recorded on our side with its signature.

You have several other PRs in this queue that are *not* on a disappearing surface — #2089, #2181, #2170 and #2085 among them — and those are being reviewed and landed on their merits. #2089 is next on my list tonight. Thank you for the steady stream of careful, well-tested fixes; it is noticed.

Issue #2020, opened 2026-09-03
index_repository reports an abort that preserved the index as an indistinguishable `error`


CaptainMittens:
Related to #2001, which quotes this same hint text. That report is about an unreadable
directory aborting a run that should have continued. This one is about the shape of the
response after an abort that is legitimate. Different cause, different fix, so I filed it
separately rather than adding to that thread.

## What happens

`cbm_pipeline_run` publishes by renaming a fully validated staging database over the
destination. Every abort before that rename leaves the existing database in place and
still serving. The pipeline says so in its return value: it has two distinct failure
codes, and it returns them from many places.

`handle_index_repository` throws that away. It tests `rc == 0` and puts everything else
into one branch:

```c
} else {
    yyjson_mut_obj_add_str(doc, root, "status", "error");
    yyjson_mut_obj_add_str(doc, root, "hint",
                           "Pipeline failed. Check repo_path exists and contains source files. "
                           "Try mode='fast' for a quicker diagnostic run.");
}
```

So a caller sees `status: "error"` and a hint naming `repo_path`, when `repo_path` was
fine and the previous index is intact.

## Why it costs the reader time

Two things go wrong at once.

The hint sends people to check a path that is not the problem. The usual real cause is
that files changed while the run was in flight, or a discovery or manifest phase failed
transiently. Neither has anything to do with `repo_path`.

The reader also cannot tell the two failures apart, and the right next step differs:

- Aborted before publication. The old index is intact and still answering queries. Retry.
- Publication failed. The staging database was validated but could not be put in place.
  The old index may have been rolled back. Check free disk space and permissions on the
  cache directory before retrying.

Today both arrive as the same string.

## The header already promises this

`src/pipeline/pipeline.h:59-68` tells callers they may distinguish the failures:

> Treating any non-zero as "the run failed" is always correct. Callers that need to know
> whether the PREVIOUS generation survived can distinguish the failures by value: the run
> publishes by renaming a fully validated staging database over the destination, so every
> abort before that rename leaves the existing database in place. Those codes
> (CBM_PIPELINE_ABORT_PRESERVE_DB and CBM_PIPELINE_PERSIST_FAILED) are defined in
> pipeline_internal.h alongside the stages that raise them.

The two constants live at `src/pipeline/pipeline_internal.h:35` and `:37`. A public caller
should not include that header, so the contract invites callers to read a value it does
not give them a name for.

## Verified on `c38aa35c`

| Fact | Where |
[...]

PR #2085, opened 2026-09-07
fix(pipeline): skip an unreadable subdirectory in the control-file walk instead of aborting the whole index

### LINKED ISSUE #2001: index_repository: one unreadable directory aborts the whole run with a generic `Pipeline failed`, and the hint points elsewhere (0.10.8)

AmirF194:
Picking up from the thread on #2001: the abort happens after discovery finishes clean, in `cbm_pipeline_build_semantic_manifest`'s own control-file walk.

`semantic_manifest_walk_controls` recursively walks the whole repo looking for `.gitignore`/`.cbmignore`/package-manifest files. When it hits the `locked` directory and `cbm_opendir` fails, it returns `CBM_NOT_FOUND`, which propagates all the way up and aborts the entire index. `discover.c`'s `walk_dir` already treats the identical failure as "nothing to discover here" and moves on; the control walk had no reason to be stricter; a permission-denied subdirectory just has no control files to contribute.

Fix skips a nested directory it can't open instead of failing the whole walk. The root call still fails closed (depth 0), so `pipeline_semantic_manifest_rejects_non_directory_root` keeps passing unchanged.

Added `pipeline_semantic_manifest_skips_unreadable_subdirectory`: a locked sibling next to a readable one, asserts the walk succeeds and still finds the sibling's `.gitignore`. Needs a non-root run to mean anything, since root ignores the chmod and the check would be a no-op (same reasoning as `test_cli.c`'s existing root-demotion comment); verified that way in a container both ways.

Not fixed here, and worth flagging so it doesn't get lost: `#2020` is the separate issue about the generic `error` hint collapsing distinct failure codes. It's real and independent of this one; I only touched the abort itself.

Fixes #2001

### DIFF
--- a/src/pipeline/pipeline_incremental.c
+++ b/src/pipeline/pipeline_incremental.c
@@ -338,7 +338,10 @@ static int semantic_manifest_walk_controls(semantic_manifest_builder_t *builder,
     }
     cbm_dir_t *dir = cbm_opendir(abs_dir);
     if (!dir) {
-        return CBM_NOT_FOUND;
+        /* A nested directory we can't open has no control files to
+         * contribute; skip it like discover.c's walk_dir already does.
+         * The root call (depth 0) still fails closed. */
+        return depth == 0 ? CBM_NOT_FOUND : 0;
     }
     int rc = 0;
     cbm_dirent_t *entry;
--- a/tests/test_pipeline.c
+++ b/tests/test_pipeline.c
@@ -4123,6 +4123,45 @@ TEST(pipeline_semantic_manifest_rejects_non_directory_root) {
     PASS();
 }
 
+#ifndef _WIN32
+/* A nested directory the process cannot open (permission-denied) must not
+ * abort the whole manifest walk: it has no control files to contribute, so
+ * it is skipped, and control files in sibling directories still get found.
+ * Root-not-a-directory (above) is a different, still fail-closed case. */
+TEST(pipeline_semantic_manifest_skips_unreadable_subdirectory) {
+    char tmp[256];
+    snprintf(tmp, sizeof(tmp), "/tmp/cbm_manifest_locked_dir_XXXXXX");
+    ASSERT_NOT_NULL(cbm_mkdtemp(tmp));
+    write_temp_file(tmp, "readable.py", "def Readable():\n    return 1\n");
+    write_temp_file(tmp, "locked/secret.txt", "should never be opened\n");
+    write_temp_file(tmp, "sibling/.gitignore", "*.log\n");
+
+    char locked_path[512];
+    snprintf(locked_path, sizeof(locked_path), "%s/locked", tmp);
+    ASSERT_EQ(chmod(locked_path, 0), 0);
+
+    cbm_file_hash_t *manifest = NULL;
+    int manifest_count = -1;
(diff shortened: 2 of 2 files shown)


DeusData:
Thank you for the focused control-file walk change and the pipeline test. We need more time to review how an unreadable directory should affect indexing and exclusion handling before giving a decision.

AmirF194:
The test-windows-guards failure is the same pre-existing section_cold_storm issue as its sibling PR #2084, unrelated to this diff. Anything else needed here?

DeusData:
Thank you for asking what is still needed. The current recursive control-file walk aborts when an `opendir` fails, so the supplied case addresses a concrete path. I do not have an additional contributor task to request from this triage: review of exclusion/error semantics and attribution of the Windows result are now maintainer-owned.

PR #2086, opened 2026-09-07
feat(lang): add Visual Basic 6 / VBA grammar (harumiWeb/tree-sitter-vba)

Tokarzewski:
## What does this PR do?

Adds **Visual Basic 6 / VBA** as a language (`CBM_LANG_VB6`) — part two of #721, following the merged short-term disambiguation in #2074. The VB6 `.frm`/`.cls` files that #2074 routed to "unsupported" now parse with a real grammar; `.bas`/`.ctl`/`.dsr`/`.pag` are mapped directly.

Refs #721.

**Grammar:** [harumiWeb/tree-sitter-vba](https://github.com/harumiWeb/tree-sitter-vba) @ `63b2f8d0d65c` (v0.13.0, MIT, ABI 15, lexer-only — no external scanner). It parses the module export headers natively (`VERSION`, `Begin … End`, `BeginProperty … EndProperty`, GUID blocks, `Attribute` lines), so the `.frm` designer block flagged in #721 as the main technical risk needs no preprocessing step — it becomes ordinary `frm_begin_block` nodes and the code below parses normally. VB6-specific statements are covered (`Declare`, `Implements`, `Event`/`RaiseEvent`, `WithEvents`, `Property Get|Let|Set`). Rejected alternatives: arrmee-wt/tree-sitter-vba (GPL-3.0 — outside `scripts/license-policy.json`), joannefan/tree-sitter-vb6 (no licence file, no generated parser, `.frm` metadata "poorly parsed" per its README), CodeAnt-AI/tree-sitter-vb-dotnet (VB.NET — that is #1073).

**Provenance caveat — please decide:** upstream gitignores `src/parser.c`, so the vendored 46 MB `parser.c` was generated from the pinned `grammar.js` with `tree-sitter-cli 0.26.9` (`tree-sitter generate --abi 15`, the project's own `pnpm generate`). Byte-identity with an upstream artefact therefore cannot be claimed; `MANIFEST.md` records the exact regeneration command (the vba row is marked `NOT-IN-REGISTRIES (see note)`). The size comes from per-keyword case-insensitive expansion; `lean`'s parser is 102 MB, so it is within precedent.

**Extraction (generic paths plus small language hooks):**
[...]

Issue #721, opened 2026-06-30
VB6 .frm/.cls misclassified as FORM/Apex; .bas/.ctl unmapped; no Visual Basic language


matigumma:
## Summary

Visual Basic 6 source files are indexed with the **wrong tree-sitter grammar** (or not at
all), because their extensions collide with two other supported languages and there is no
Visual Basic language in the table:

| Ext | Current mapping (`src/discover/language.c`) | Correct |
|-----|---------------------------------------------|---------|
| `.frm` | `CBM_LANG_FORM` (line 104) — the FORM symbolic-manipulation language (`vermaseren/form`) | Visual Basic |
| `.cls` | `CBM_LANG_APEX` (line 333) — Salesforce Apex | Visual Basic |
| `.bas`, `.ctl`, `.vbs` | *(absent from `EXT_TABLE`)* | Visual Basic |

There is **no `CBM_LANG_*` for Visual Basic / Basic** among the 449 extension entries.

## Impact

On a real VB6 codebase (~800 `.frm`/`.cls` files), indexing produced **zero function nodes
for every `.frm` file** and parsed `.cls` with the Apex grammar. Forms — which hold most of a
VB6 app's logic — are effectively invisible to `search_graph` / `query_graph` /
`get_code_snippet`. Example: a 9,960-line form with 100 procedures yielded 0 `Function` nodes
(only a handful of stray identifiers mislabeled as `Variable`).

This is an **extension collision**, not merely a missing language: `.frm` is shared with FORM
and `.cls` with Apex, so both currently win over (nonexistent) VB6 handling.

## Repro

1. Index any VB6 project containing `.frm`/`.cls` files (e.g. forms with `Private Sub … End Sub`).
2. `query_graph("MATCH (n:Function) WHERE n.file_path CONTAINS '.frm' RETURN count(n)")` → `0`.

## Signatures (for disambiguation)

VB6 files are reliably distinguishable by their first line (Windows-1252 encoded, CRLF):

- VB6 form `.frm`: `VERSION 5.00` followed by `Begin VB.Form …`
- VB6 class `.cls`: `VERSION 1.0 CLASS`
- FORM `.frm`: `Symbol x;`, `#procedure …` (no `VERSION`/`Begin VB.`)
- Apex `.cls`: `public class …`, `@isTest` (brace-based)

## Proposed fix

**(a) Short term — content disambiguation.** The codebase already has the exact pattern:
[...]

PR #2087, opened 2026-09-07
feat(extract): VB6 class modules as Class + Methods (stacked on grammar PR)

Tokarzewski:
## What does this PR do?

**Draft, stacked on https://github.com/DeusData/codebase-memory-mcp/pull/2086.** Only the top two commits are this change; until that PR merges the diff here also shows its grammar commit. Rebased onto `main`, this is an 8-file change (+379/−3). Opened as a draft so the design can be reviewed alongside the grammar — happy to fold it into the grammar PR, keep it separate, or drop it, whichever you prefer.

Refs #721.

**Why:** a VB6 `.cls`/`.frm`/`.ctl`/`.dsr`/`.pag` file *is* exactly one COM class, but the grammar has no class node. With the grammar PR alone every procedure in a class module is a free `Function`, and a VB6 project has no `Class` nodes, no `DEFINES_METHOD` edges and no `Implements`/form inheritance in the graph.

**What it does:**
- Synthesises one `Class` per class-module file, named from `Attribute VB_Name = "…"` (the IDE-written identity; falls back to the file stem), with base classes from the designer block type (`Begin VB.Form Form1` → `VB.Form`; GUID-typed blocks skipped) and every `Implements`.
- Walks the file inside that class scope, so procedures become `Method`s (`class.Name`, `parent_class` set → `DEFINES_METHOD`), and nested `Type`/`Enum` declarations nest under the class. `.bas` standard modules are untouched (still `Function`s).
- Call attribution: the scope-frame walker already saves `enclosing_class_qn` on every push and restores it on pop, so the seeding is one line in `cbm_extract_unified()` — the baseline becomes the file-class QN for VB6 class modules and stays `NULL` for everything else. `compute_func_qn` then produces exactly the Method QN the def side mints; both sides derive the class QN from a single helper (`cbm_vb6_file_class_qn`), so they cannot drift. No new state field.

**Tests:** regression — `Widget.cls` yields 1 `Class` (base `IShape`), 2 `Method`s, 0 `Function`s, every Method's `parent_class` is the class QN, and the call to `Helper` is attributed to `class.Go`; a `.bas` module yields no Class/Method. Contract — a `.cls` `CALL_CASES` row plus `contract_vb6_class_module_calls_source_from_method`, which indexes the fixture through the production pipeline and requires the only caller of `Helper` to be the Method `Go` (never the Module). Amends one assertion in the grammar PR's `vb6_implements_is_an_import` (`Go` is now a `Method`).

**Open design points for you:**
1. `Property Get/Let/Set` triplets share one QN (`class.Name`) and collapse to one `Method` node — consistent on the def and call side, and the same collapse the grammar PR has for `.bas`. A `.get/.let/.set` suffix would need matching changes on both sides.
2. `Declare … Lib` and `Event` declarations inside a `.cls` become Methods of that class (they are module members).
3. Module-level `Dim`/`Const` in a `.cls` remain module-level `Variable`s rather than class `Field`s — a possible follow-up.
4. Two commits (feature + tests) — squash if you prefer one.

## Checklist

- [x] Tests pass locally — stacked on the grammar branch, same Windows zig/clang toolchain **without** ASan/UBSan: `language extraction grammar_regression grammar_labels lang_contract` → 628 passed, 0 failed; `grammar_label_goldens` unchanged; `[CALLS-BREADTH] 56 langs: 0 FAILURES`; language-count contract unchanged (this PR adds no grammar). CI's sanitised runs are authoritative.
- [x] Lint: `clang-format-20 --dry-run --Werror` clean on `helpers.c/.h`, `extract_defs.c`, `extract_unified.c/.h`; cppcheck not run locally — relying on CI
- [x] New behaviour is covered by tests (above); all fixtures synthetic


Tokarzewski:
`test / test-windows (windows-latest, CLANG64, x86_64, 2/2)` failed on the last run: `tests/test_parallel_harness_contract.sh` — `FAIL: Windows timeout race refused without a surviving descendant to refuse over -- the fixture no longer exercises the race` (synthetic `stubborn_tree` subprocess scenario).

This looks pre-existing and unrelated to this PR: the fixture tests the CI harness's own process-timeout/kill logic under synthetic subprocess trees, not extraction/parsing, and the failure text describes the test's own race precondition not being hit on this run rather than an assertion mismatch. I don't have rights to re-run the job on this repo. Happy to investigate further if it recurs or if you'd like a second opinion.

PR #2089, opened 2026-09-07
fix(cli): emit the OpenCode plugin as a V2 default export

### LINKED ISSUE #2077: Generated OpenCode plugin uses V1 export shape; V2 loader rejects it (SchemaError: Missing key at ["default"])
## Summary

The OpenCode plugin generated by `codebase-memory-mcp install` (v0.10.8) at `.opencode/plugins/cbm-augment.ts` uses the **V1 plugin shape** — a named export returning hooks:

```ts
export const CodebaseMemory = async () => ({
  'tool.execute.after': async (input, output) => { ... },
});
```

The current OpenCode **V2** plugin loader only validates the module's **default export**, expecting `{ id, setup() | effect() }`. The plugin therefore fails to load on every server start:

```
message="failed to load plugin" target=.opencode/plugins/cbm-augment.ts ref=err_1077113d
cause="PluginModule.LoadError: Plugin must export a default definition with an id and an effect or setup
function. (cause: SchemaError(Missing key at [\"default\"]))"
```

## Environment

- codebase-memory-mcp 0.10.8
- OpenCode V2 (server plugin runtime)
- Linux

## Suggested fix

Emit the V2 shape. Note that pinning `@opencode-ai/plugin` is risky: versions like 1.18.18 don't export `Plugin.define`, so a plain default-export object is the most portable form:

```ts
export default {
  id: 'codebase-memory-augment',
  async setup(ctx) {
    await ctx.tool.hook('execute.after', async (event) => {
      const tool = event.tool === 'grep' ? 'Grep' : event.tool === 'glob' ? 'Glob' : null;
      if (!tool) return;
      // spawn `codebase-memory-mcp hook-augment` and append to event.result
    });
  },
};
```

## Additional context

- Hand-converting the file restores loading, but the file lives between `codebase-memory-mcp:start/end` markers and is regenerated on install/update, so the broken shape returns on every update.
- Possibly related: #1737 (same generated plugin also passes `output?.args` instead of tool input) and #1335 (hook-augment warm-daemon issue).

## Workaround used

Manual rewrite of the generated file to the V2 shape above; marker comments retained knowingly, expecting them to be overwritten on update.

AmirF194:
OpenCode's V2 plugin loader only reads the default export and requires an id plus a setup() or effect() function. `cbm_client_adapter_opencode` still generated the old named export (`export const CodebaseMemory = async (ctx) => {...}`), so every server start failed with `SchemaError: Missing key at ["default"]`, and the workaround gets overwritten on every install/update since the file is regenerated.

Wrapped the same hook logic in a default export with `id: 'codebase-memory-augment'` and `async setup(ctx)`; the returned hooks object (`tool.execute.after`, `experimental.session.compacting`) is unchanged.

Added a test asserting the generated module has the default export shape and no longer contains the old named export. Ran `scripts/test.sh --suites agent_clients` on Ubuntu 24.04: the new test fails without the change and the suite is green on the branch (38 tests).

Fixes #2077


DeusData:
Thanks for taking #2077 on so quickly, and for the RED/GREEN evidence — the diagnosis is right: OpenCode's newer config loader validates `default` against `{ id, effect | setup }`, so a default export is the way to go.

One thing turned up when running the generated module through OpenCode's own loaders, though. There are two of them, and the one that actually fires `tool.execute.after` is the server runtime (`packages/opencode/src/plugin/index.ts` → `readV1Plugin` in `shared.ts`). It reads `default.server`, not `default.setup`. With `{ id, setup }` it throws `Plugin … must default export an object with server()`, and the V2 loader that accepts `{ id, setup }` calls `setup()` but drops whatever it returns (`packages/core/src/plugin/promise.ts:91`) — its context has no `tool` domain yet. So on every 1.18.x checked (v1.18.18 / 25 / 29 and dev) this shape loads without an error line but registers zero hooks — the silent no-op #616 taught us to fear.

The shape that satisfies both loaders:

```js
export default {
  id: 'codebase-memory-augment',
  // server runtime: reads default.server and dispatches the hooks it returns
  server: async (ctx) => { /* the existing hook object */ },
  // V2 config loader: requires id + setup|effect; no tool domain yet, so nothing to register
  setup() {},
};
```

Could you switch `async setup(ctx) {` to `server: async (ctx) =>`, add the empty `setup() {}`, and update the new test to pin `server:` and `setup() {}` (keeping the "no `export const CodebaseMemory`" check)? Also, #1933 lands first and changes `output?.args` → `input?.args` in the same block, so after rebasing please keep `input?.args` inside your rewritten block — its three assertions will tell you if it slipped. Then this is a merge. The red `test-windows-guards` leg on your head is the known cold-storm daemon-spawn race, unrelated to this diff.

AmirF194:
This one's green across all 34 legs and has had no comments in a week. Anything you'd want changed before merging?

DeusData:
Thank you for the loader follow-up. The current head still returns hooks from asynchronous setup, while the earlier requested contract was to place them in the server entry and leave setup empty. Please reconcile both code and tests with that requested shape, preserving the `input?.args` guard. This is a check against our existing review request, not a new claim that the external loader contract has been independently revalidated.

AmirF194:
Pushed e171bd48: moved the hooks under server: async (ctx) => {...} and left setup() {} empty, matching the shape you asked for. Updated the emission test to check both server: and setup() {}, and to assert the old async setup(ctx) form is gone. Full local suite green, 7989 passed / 0 failed.

DeusData:
Read `e171bd48` against the 7 September request, and it is exactly the shape that was asked for: a default export with `id: 'codebase-memory-augment'`, an empty `setup() {}` for the V2 config loader, and the hook object returned from `server: async (ctx) => …` for the server runtime that actually dispatches `tool.execute.after`. The `input?.args ?? {}` guard is intact, the hook bodies are unchanged apart from indentation, and the emission test now pins both halves plus the absence of the old `async setup(ctx)` form. Both commits are signed off. Thank you for turning that around in a day, and I am sorry the round before it cost you two weeks.

**The one red check is ours.** `test-windows-guards` fails in `section_cold_storm` with a cold-start race in the Windows daemon's directory creation. It hit five unrelated PRs today, and a fix is written and being proven. Nothing in `client_adapter.c` can reach it.

What happens next is on my side: a local build of the merge result with the `agent_clients` and `cli` suites, then I bring the branch up to date with `main` (it is 166 commits behind) for a fresh CI verdict, and then it merges. I am holding the branch update for an hour or two only because the CI runner pool is saturated tonight. Nothing further is needed from you.

Issue #1754, opened 2026-08-20
Support Solidity by integrating solidity-parser-antlr


djmahe4:
### What problem does this solve?

Current indexing/parsing support does not cover Solidity files, so Solidity repositories can be skipped, partially indexed, or produce incomplete code graph/data. That makes the repository memory features much less useful for teams working with smart contracts.

### Proposed solution

Add Solidity support by integrating `solidity-parser-antlr` into the parsing pipeline. Use it to parse Solidity source files, extract declarations/imports/references, and feed the resulting AST into indexing and search so Solidity projects are handled consistently with other supported languages. Include representative public OSS Solidity repos as test beds to validate coverage and parsing quality.

### Alternatives considered

Alternative approaches could include extending a hand-rolled parser or relying on simpler regex-based extraction, but those would likely be less accurate and harder to maintain than using a dedicated Solidity parser.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

sanskar717:
Happy to pick this up — quick check before I start: solidity-parser-antlr looks unmaintained (last updates years old), while @solidity-parser/parser is the actively maintained fork with the same AST shape. OK if I implement against the fork instead? Wanted to confirm before starting so I don't build on a dead dependency.

djmahe4:
Actually I'm working on it in one of my forks and I can do one pr adding solidity as a single lang, but it would take a couple of weeks for me to do so.
Meanwhile you can proceed with `solidity-parser/parser`

asd59478126-commits:
This is an architecture reference for how I am currently thinking about agent task structure.

The idea is intentionally minimal:

**Purpose = Necessary Requirement**

A Purpose primarily provides a stable reference to two structures:

* Narrative Chain
* Fixed Logical Chain

For the first version, I am using five general task types as the initial vocabulary:

**Search / Generation / Writing / Analysis / Execution**

This is not intended to prescribe an implementation or constrain your Solidity integration. It is simply a reference model for thinking about how an agent can maintain task continuity while task states and operations change.

The implementation details are deliberately left open.

[Agent Purpose Construction — V1 Architecture Specification.md](https://github.com/user-attachments/files/31895277/Agent.Purpose.Construction.V1.Architecture.Specification.md)

sanskar717:
@DeusData @djmahe4

Opened a PR on the existing Tree-sitter Solidity path to address the missing modifier/decorator tracking.
[...]

PR #2090, opened 2026-09-07
fix(lang): track Solidity modifier invocations as decorators

sanskar717:
## Summary

Adds Solidity `modifier_invocation` support to the existing Solidity language specification so modifier usages such as `onlyOwner` are tracked as decorators in the code graph.

## Changes

* Added `solidity_decorator_types` with `modifier_invocation`
* Wired it into `CBM_LANG_SOLIDITY`
* Enables modifier definitions and invocations such as `onlyOwner` to be tracked correctly

## Verification

Verified locally with `search_code` using `onlyOwner`:

* Found the modifier definition
* Found the `onlyOwner` invocation on `withdraw`

 Relation to #1754

The issue proposes adding Solidity parser support. Since Solidity grammar is already vendored through Tree-sitter, this change extends the existing parsing path without introducing a Node.js runtime or JavaScript dependency.

This PR specifically fixes the missing modifier/decorator tracking in the current Solidity support.

### DIFF
--- a/internal/cbm/lang_specs.c
+++ b/internal/cbm/lang_specs.c
@@ -1038,6 +1038,7 @@ static const char *solidity_field_types[] = {"state_variable_declaration", "stru
 static const char *solidity_call_types[] = {"call_expression", "call", "new_expression", NULL};
 static const char *solidity_import_types[] = {"import_directive", "import", "using_directive",
                                               NULL};
+static const char *solidity_decorator_types[] = {"modifier_invocation", NULL};
 static const char *solidity_branch_types[] = {"if_statement",    "for_statement",
                                               "while_statement", "do_while_statement",
                                               "try_statement",   NULL};
@@ -2115,7 +2116,7 @@ static const CBMLangSpec lang_specs[CBM_LANG_COUNT] = {
                            solidity_field_types, solidity_module_types, solidity_call_types,
                            solidity_import_types, empty_types, solidity_branch_types,
                            solidity_var_types, solidity_assign_types, solidity_throw_types, NULL,
-                           empty_types, NULL, NULL, tree_sitter_solidity, NULL},
+                           solidity_decorator_types, NULL, NULL, tree_sitter_solidity, NULL},
 
     // CBM_LANG_TYPST
     [CBM_LANG_TYPST] = {CBM_LANG_TYPST, typst_func_types, empty_types, empty_types,


sanskar717:
The issue specifically requested `@solidity-parser/parser` (JS). Since this project is pure C and uses a single-binary setup, I completed the existing `tree-sitter-solidity` path instead, focusing on modifier/decorator tracking.

Verified using `search_code` with `onlyOwner`, covering both the modifier definition and its invocations.

DeusData:
Thank you for working within the existing C/tree-sitter Solidity path and clarifying the modifier-invocation scope. We need more time to review the extraction semantics and their regression coverage before giving a decision.

DeusData:
Here is the review this was waiting on since 2026-09-09. Eleven days for a two-line change is too long, and I am sorry.

**First, the judgement call you made was the right one and I want to endorse it explicitly.** The issue asked for `@solidity-parser/parser`, a JavaScript dependency. This project is a single C binary with vendored tree-sitter grammars, and adding a JS parser to it would have been a bigger architectural change than the feature justifies. Completing the existing `tree-sitter-solidity` path instead is exactly the call I would have made, and you explained the reasoning up front rather than quietly substituting one thing for another. That saved the review a round trip.

The change itself reads correctly: `solidity_decorator_types = {"modifier_invocation"}` puts Solidity on the same footing as `java_decorator_types`, `rust_decorator_types` and the rest, and feeds the existing `extract_decorators` path rather than inventing a parallel one.

**What I need before merging is a regression test, and it is the one thing the maintainer note from 2026-09-09 was circling.**

Right now nothing pins this behaviour. I checked: `tests/test_grammar_labels.c:107` asserts a Solidity census of `Class:1,Method:1,Module:1`, and your CI is fully green — which tells us the change adds no nodes to that fixture, and therefore that no existing assertion would notice if this line were deleted tomorrow. A two-line table entry with nothing holding it in place tends to get reverted by accident during the next refactor of that table.

The shape to follow already exists next door — `tests/test_grammar_probe_c.c:314` (`solidity_contract_node`) and `:334` (`solidity_interface_node`) are the pattern. Something like a contract with a modifier:

```solidity
pragma solidity ^0.8.0;
contract Vault {
    modifier onlyOwner() { _; }
    function withdraw() public onlyOwner { }
}
```

asserting that `withdraw` carries `onlyOwner` as a decorator. The test to aim for is one that **fails with your two lines reverted** — if it passes either way, it is not pinning anything, and this repository has shipped a few tests like that, so I check for it now rather than assume.

No rush and no pressure on timing. Push the test and I will review promptly — the table change itself is already settled as far as I am concerned.

sanskar717:
Pushed the regression test (7036c4d) — `solidity_modifier_decorator` in test_extraction.c, following the pattern from extract_java_method_annotations_issue382.

Verified: reverting the table entry (solidity_decorator_types → empty_types) makes it fail with `ASSERT(decorators_contain(m, "onlyOwner"))`; with the fix in place it passes.

DeusData:
Thank you — the regression test is exactly what was asked for, and saying how you proved it (table entry reverted, `ASSERT(decorators_contain(m, "onlyOwner"))` fails; restored, it passes) is what makes it a test rather than a decoration.

**The twelve red legs on this head are ours, not yours.** Every one of them fails on the same test, which has nothing to do with Solidity:

```
tool_check_index_coverage_freshness_uses_indexer_mtime_source_issue1714
tests/test_mcp.c:5806: strstr(response, ""freshness":"metadata_match"") is NULL
```

That test asserted a JSON field without asking the tool for JSON; it was fixed on `main` on 20 September (`96a2fc35`), and your CI run was built against a `main` from before that fix. Your branch is 226 commits behind, so the verdict you are looking at is stale rather than wrong.

Next steps are mine: a local build of the merge result with the extraction suites, then I update the branch with `main` for a fresh run. I am holding that update for an hour or two only because the CI runner pool is saturated tonight. After a green matrix this merges. Nothing further is needed from you.

sanskar717:
**Side note:** VS Code flags `CLOCK_MONOTONIC` as undefined at line 6049 (`extract_wide_flat_ms`) — `time.h` is included, so this may be Windows-specific (matches the `windows-CLANG64` comment just above it). Not something I touched, flagging in case it's useful.

PR #2099, opened 2026-09-08
fix: uninstall removes the executable and indexes even when an agent config cleanup fails

CaptainMittens:
Takes two of the three directions settled in #1954. The third — following a
config symlink owned by the invoking user — is not here; see "Not in this PR"
below.

## 1. Agent config cleanup no longer holds the executable and the indexes hostage

`cli_uninstall_activate` returned early as soon as any one agent configuration
failed to clean up, before index and executable removal had started. The
reporter's trigger was `~/.cursor/mcp.json` symlinked into a dotfiles
repository: the JSON editors open with `O_NOFOLLOW` on purpose and refuse to
write through the link, and that one refusal left a 282 MB executable and a
104 MB index cache on disk after a run that had already printed a per-client
report. All four leftovers had to be removed by hand.

Cleaning up an agent's configuration and removing this tool's own files are
separate jobs, so the gate is gone. Nothing about the failure gets quieter:

- every failure is still named on stderr as it happens, by
  `record_agent_config_error`;
- the closing line now states the count, because a reader who watches stdout
  alone never sees the stderr lines;
- the exit code is still non-zero — `return g_agent_uninstall_errors == 0 ? 0 :
  CLI_TRUE;` was already there and is unchanged.

`CBM_AGENT_EDIT_NOT_APPLICABLE` also stopped counting as a failure in the agent
client registry loop. That value means the client has no MCP editor for its
config shape, so there was never an entry of ours to take out. Treating it as an
error was a second route to the same refusal over a file this tool had never
written to.

## 2. The report says what the run actually did

Every per-client line was written in the past tense whatever happened. So:

- `--dry-run` said `removed` for files it never opened;
- a step that had just recorded an error said `removed` on the next line;
- a preserved foreign MCP entry printed `preserved` and `removed` one after the
  other.

Reading either output alone, an operator would conclude a dozen configuration
files had been rewritten — which is exactly what the report describes.

Two small helpers carry the fix:

- `uninstall_verb(dry_run)` returns `would remove` or `removed`, so a dry run
  never claims a past action;
- `uninstall_steps_succeeded(errors_before)` ties a line that covers several
  steps at once (`removed MCP config + hooks + instructions`) to whether every
  one of those steps worked.

A step that failed now prints no line of its own; `record_agent_config_error`
has already named the agent, the operation and the path on stderr.

The shape copies the Qoder and Devin blocks, which already reported
`planned` / `complete` / `failed` correctly — the correct version was already in
the tree.

## Tests

Both in `tests/test_cli.c`, against a temporary `HOME` with `PATH` redirected so
a real agent binary on the developer's machine cannot add an unrelated failure.
Neither runs a real uninstall against a real home.

- `cli_uninstall_removes_binary_and_index_despite_agent_config_failure`
  (POSIX only, needs `symlink`) builds the reporter's fixture — `.cursor/mcp.json`
  as a symlink into a dotfiles directory — and asserts the executable and the
  index are gone, the link's target is untouched, the exit code is still 1, the
  new failure count appears, and the old abort message does not.
- `cli_uninstall_dry_run_report_uses_the_future_tense` asserts a `--dry-run`
  prints `Cursor: would remove MCP config entry`, never
  `Cursor: removed MCP config entry`, and leaves the config file byte-identical.

Red before the change, green after.

## Two existing tests changed on purpose

Both encoded the contract this PR sets out to change, so both had to be
inverted rather than kept:

- `cli_agent_uninstall_reports_safe_editor_refusal` asserted that a malformed
  OpenClaw config left the executable in place, and that the activation guard
  produced an "executable was kept" diagnostic. It now asserts the executable is
  gone and no partial-activation diagnostic is produced. The two claims that
  mattered are unchanged: the exit code is still non-zero, and the malformed
  file is still byte-identical afterwards — it is never written through.
- `cli_codex_migrates_to_single_hook_representation` checked
  `stat(binary_path) == 0` after an uninstall whose Codex hook preflight failed,
  with the fixture content "installed binary must survive failed cleanup". Now
  `stat(binary_path) != 0`. Everything else it checks — the non-zero exit, the
  removed skill and agent files, the cleaned hooks, the emptied AGENTS.md — is
  untouched.

No test asserted any of the report wording, so the tense and outcome changes
broke nothing.

## Checked, not changed

The install path does not have the same shape. `install_generic_agent_config`
prints the config path (`  mcp: <path>`) rather than a past-tense claim, and its
dry run goes through the install plan (`g_install_plan`). Nothing to fix there.

## Not in this PR

- **Following a config symlink owned by the invoking user.** That is the
  `O_NOFOLLOW` decision in `src/cli/config_json_like.c`, and #2060 is already
  adding an ownership helper on the install side. Doing it here would collide.
- **A config file that exists but holds no entry of ours still reports
  `removed`.** The removers answer OK both when they deleted our entry and when
  there was nothing of ours to delete, so the caller cannot tell those apart
  without a fourth return value across the JSON and YAML removers. That is a
  wider change than this one and I would rather send it separately — say the
  word and I will.

Refs #1954

### DIFF
--- a/src/cli/cli.c
+++ b/src/cli/cli.c
@@ -10802,6 +10802,30 @@ int cbm_cmd_install(int argc, char **argv) {
 
 /* ── Subcommand: uninstall ────────────────────────────────────── */
 
+/* One wording for every line the uninstall report prints.
+ *
+ * The report used to be written in the past tense whatever happened: a
+ * --dry-run said "removed" for files it never opened, and a step that had just
+ * recorded an error said "removed" on the next line (#1954). A reader could
+ * only conclude that a dozen configuration files had been rewritten.
+ *
+ * Every report line now goes through this verb, and the caller prints the line
+ * only when the step really ran and worked. A failed step prints nothing here
+ * because record_agent_config_error has already named the agent, the operation
+ * and the path on stderr. The shape copies the Qoder and Devin blocks below,
+ * which already said "planned" for a dry run and "failed" for a failure. */
+static const char *uninstall_verb(bool dry_run) {
+    return dry_run ? "would remove" : "removed";
+}
+
+/* True when nothing has recorded a failure since `errors_before` was read from
+ * g_agent_uninstall_errors. Several report lines below cover a run of steps at
+ * once ("removed MCP config + hooks + instructions"), and this ties such a line
+ * to whether every one of those steps worked. */
+static bool uninstall_steps_succeeded(int errors_before) {
+    return g_agent_uninstall_errors == errors_before;
+}
+
 /* Remove Claude Code agent configs. */
 static void uninstall_claude_code(const char *home, const char *installed_binary, bool dry_run) {
     char config_dir[CLI_BUF_1K];
@@ -10812,7 +10836,7 @@ static void uninstall_claude_code(const char *home, const char *installed_binary
     char skills_dir[CLI_BUF_1K];
     snprintf(skills_dir, sizeof(skills_dir), "%s/skills", config_dir);
     int removed = cbm_remove_skills(skills_dir, dry_run);
-    printf("Claude Code: removed %d skill(s)\n", removed);
+    printf("Claude Code: %s %d skill(s)\n", uninstall_verb(dry_run), removed);
     char agent_path[CLI_BUF_1K];
     snprintf(agent_path, sizeof(agent_path), "%s/agents/codebase-memory.md", config_dir);
     uninstall_tiered_agent_profiles(
@@ -10827,10 +10851,14 @@ static void uninstall_claude_code(const char *home, const char *installed_binary
 
     char mcp_path[CLI_BUF_1K];
     snprintf(mcp_path, sizeof(mcp_path), "%s/.mcp.json", config_dir);
+    bool legacy_mcp_removed = true;
     if (!dry_run && cbm_remove_editor_mcp_owned(installed_binary, mcp_path) != CLI_OK) {
         record_agent_config_error(true, "Claude Code", "legacy_mcp_uninstall", mcp_path);
+        legacy_mcp_removed = false;
     }
-    printf("  removed MCP config entry\n");
+    if (legacy_mcp_removed) {
+        printf("  %s MCP config entry\n", uninstall_verb(dry_run));
+    }
 
     char mcp_path2[CLI_BUF_1K];
     snprintf(mcp_path2, sizeof(mcp_path2), "%s/.claude.json", user_root);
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for addressing the uninstall failure path and adding CLI coverage. The current cleanup-error guard can indeed stop uninstall before binary removal; this is also tracked in #1954 and #2106. We need more time to review this alongside #2110 so the overlapping fixes are handled coherently and your contribution is not lost.

DeusData (changes requested):
Thank you for this, and for your patience: I said on 9 September that I needed to handle it together with #2110 so nothing of yours got lost, and here is that review. The short version: **half of this PR is already on `main` through #2110 (merged 12 September, `b790be3d`), and the other half is exactly the follow-up #2110's description deferred — so this PR is that follow-up, not a duplicate, and I would like to land it.**

What #2110 already did, so it drops out of your diff on rebase: the abort gate in `cli_uninstall_activate` is gone (cleanups → indexes → binary), a failed cleanup ends in a structured stderr list of agent/op/path/reason plus "Uninstall finished with errors", the two test inversions (`cli_agent_uninstall_reports_safe_editor_refusal`, `cli_codex_migrates_to_single_hook_representation`) are identical to yours, and a RED-first hostage test with a dangling symlink covers the "binary and indexes survive a cleanup failure" contract. #2110 also started *following* user-owned symlinked configs under HOME, which changes what your hostage fixture means (a link to an existing `$HOME/dotfiles/mcp.json` is now cleaned, exit 0).

What is genuinely new here and not on `main`, verified by grep: the honest per-client report wording (`uninstall_verb` / `uninstall_steps_succeeded`, "would remove" on `--dry-run`, no "removed" line after a failed or preserved step), the `CBM_AGENT_EDIT_NOT_APPLICABLE` mapping in the client-registry loop so "this HOME never had that client" is not counted as a failure, and the test `cli_uninstall_dry_run_report_uses_the_future_tense`. I built your head and ran `cli` (311/0), then reverted only `cli.c` and got exactly the four expected reds, so the tests bind; on a local resolution of your branch onto today's `main` the suite is 320/0 with all three #1954 tests passing, clang-format and the memory-core linter clean.

Please rebase onto `main`; the conflicts are mechanical but one of them git will *not* flag, so here is the exact recipe:

1. `cli_uninstall_activate`: keep `main`'s slot as is (it already drops the gate and adds the Windows PATH removal); drop your comment there.
2. OpenHands block: keep `main`'s new settings.json / profile-refs lines and apply only your `%d skill(s) %s` wording to the first printf.
3. **Silent one:** drop your `else` branch at the end of `cbm_cmd_uninstall`. `main` prints its failure list and returns before that point, so your count line would merge cleanly into dead code.
4. Drop `cli_uninstall_removes_binary_and_index_despite_agent_config_failure`: `main`'s `…_when_agent_config_cleanup_fails` covers the contract, and since #2110 follows user-owned symlinks your fixture now exits 0.
5. Drop your edits to the two inverted tests — `main` has the same inversions.
6. Keep `uninstall_verb` / `uninstall_steps_succeeded`, the NOT_APPLICABLE branch, and the dry-run tense test with its `RUN_TEST` after the `#endif`.

One optional but worthwhile line while you are in there: `agent_continue_edit` returns NOT_APPLICABLE both when the document is missing (rc 1) and when the read fails or is refused (rc −1), so with your mapping a refused Continue config would print "no managed entry to remove" and not fail the run. Returning `CBM_AGENT_EDIT_ERROR` on −1 and NOT_APPLICABLE only on 1 keeps the new branch from ever hiding a refusal. Expected result after the rebase: `cli` 320/0. If you would rather I carry the rebase myself, say so and I will, with your commit kept as the author.

PR #2100, opened 2026-09-08
fix(mcp): prune cached projects with missing roots

### LINKED ISSUE #1431: Stale project indexes from inactive sessions are never pruned
## Summary

The stale-root pruning added by #817 only applies to projects that are **currently registered with the watcher**. When an MCP/agent session disconnects, its project is unwatched. If the corresponding worktree or directory is removed afterward, that cached project DB is never observed as missing and is retained indefinitely.

This leaves stale project entries and large cache files behind in common short-lived worktree workflows.

## Environment

- codebase-memory-mcp: `v0.9.1-rc.1`
- macOS arm64
- Client: Codex Desktop
- `auto_index=true`
- `auto_watch=true`
- Default cache directory

## Reproduction

1. Create a temporary Git worktree.
2. Start an agent/MCP session in that worktree and allow it to be indexed.
3. Close or disconnect that session.
4. Observe the daemon log registering and then unregistering the project:

   ```text
   watcher.watch project=<project> path=<worktree>
   watcher.unwatch project=<project>
   ```

5. Remove the worktree directory.
6. Keep other sessions running, or restart the daemon.
7. Wait longer than the default 10-minute prune grace period.
8. Call `list_projects`.

## Actual behavior

The project remains in `list_projects` and its `.db` remains in the cache indefinitely. There are no `watcher.root_missing` or `watcher.root_pruned` events because the historical project is no longer registered with the watcher.

A read-only audit of one real cache found:

- 99 `*.db` files
- 97 readable project indexes
- 28 project indexes pointing to 23 root paths that no longer exist
- approximately 3.39 GiB held by those stale project DBs

The stale indexes came primarily from short-lived worktrees and temporary agent tasks.

## Expected behavior

There should be a safe way to discover and prune cached projects whose recorded `root_path` no longer exists, even when those projects are not owned by a currently connected session.

## Existing behavior and known limitation

#286 requested automatic deletion after a project root disappears. #817 implemented conservative stale-root pruning, but its PR description explicitly notes:

> Only currently-watched projects are pruned. Stale cached DBs left behind by older sessions (projects never watched in this server's lifetime) are out of scope for this change.

This report covers that residual case.

#1429 has the same watcher-registration root cause and proposes watching all indexed projects, although it focuses on nested-repository freshness. #573 discusses longer-term worktree deduplication and garbage collection.

## Suggested solution

Add an explicit, coordination-safe cache pruning operation, for example:

```text
prune_projects(dry_run=true)
```

Possible behavior:

1. Enumerate project DBs and read each recorded `root_path`.
2. Treat only `ENOENT` / `ENOTDIR` as missing; preserve projects on `EACCES`, `EIO`, mount failures, or other uncertain errors.
3. Persist a first-missing timestamp so the grace period survives session disconnects and daemon restarts.
4. Default to dry-run and report project name, root path, age, and reclaimable bytes.
5. On confirmed prune, use the existing project mutation coordination and remove `.db`, `.db-wal`, and `.db-shm`.
6. Clear the missing state if the root reappears.

Alternatively, the daemon could register all cached projects for stale-root observation while any daemon session is active, but an explicit dry-run prune command would also support deterministic maintenance and avoid coupling cleanup to session lifetime.

## Acceptance criteria

- A project indexed by a previous/inactive session is discoverable as stale after its root remains absent for the configured grace period.
- Missing state survives daemon restarts, or an explicit prune command can safely evaluate it.
- A restored root is not deleted.
- Uncertain filesystem errors never trigger deletion.
- Dry-run performs no writes and reports the candidate count and reclaimable size.
- Confirmed pruning is coordinated with active indexing/query sessions.

lorenzozanee:
## What does this PR do?

Cached project indexes from disconnected sessions remain after their worktrees disappear because they are no longer registered with the watcher. This adds a dry-run-by-default `prune_projects` MCP tool that identifies confirmed missing roots and removes the database and sidecars only after coordination.

Closes #1431

## Checklist

### DIFF
--- a/src/mcp/mcp.c
+++ b/src/mcp/mcp.c
@@ -2878,8 +2878,8 @@ static int64_t project_cache_bytes(const char *dir_path, const mcp_project_recor
 }
 
 static const char *prune_project_cache(cbm_mcp_server_t *srv, const char *dir_path,
-                                        const mcp_project_record_t *record, bool dry_run,
-                                        int64_t *reclaimed_bytes) {
+                                       const mcp_project_record_t *record, bool dry_run,
+                                       int64_t *reclaimed_bytes) {
     int root_status = project_root_status(record->root_path);
     if (root_status == 0) {
         return "root_present";
@@ -3073,8 +3073,8 @@ static char *handle_prune_projects(cbm_mcp_server_t *srv, const char *args) {
         }
         candidate_count++;
         int64_t candidate_bytes = project_cache_bytes(dir_path, &records[i]);
-        const char *status = prune_project_cache(srv, dir_path, &records[i], dry_run,
-                                                 &reclaimed_bytes);
+        const char *status =
+            prune_project_cache(srv, dir_path, &records[i], dry_run, &reclaimed_bytes);
         if (strcmp(status, "deleted") == 0 || strcmp(status, "already_deleted") == 0) {
             deleted_count++;
         }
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for contributing the missing-root cleanup proposal and tests. Because this introduces a project-deletion tool and its dry-run contract, we need more time to review the API and deletion safeguards before giving a decision.

PR #2101, opened 2026-09-08
test(daemon): bound idle frontend maintenance probes

### LINKED ISSUE #2093: Idle MCP frontend still burns ~19% of a core on v0.10.8 (Windows) — looks like a regression of #1764
### Version

v0.10.8

### Platform

Windows (x64)

### Install channel

### Binary variant

standard

I'm on v0.10.8 (latest release) on Windows, and I'm seeing the same idle-CPU behavior that #1764 was supposed to fix. Reporting here in case the fix regressed or doesn't cover this path.

Isolated reproduction (no MCP client involved)

To rule out my editor/agent, I started a single frontend process by hand, kept stdin open, and sent zero JSON-RPC requests:

powershell
$psi = New-Object System.Diagnostics.ProcessStartInfo
$psi.FileName = "C:\...\codebase-memory-mcp.exe"
$psi.RedirectStandardInput = $true
$psi.RedirectStandardOutput = $true
$psi.RedirectStandardError = $true
$psi.UseShellExecute = $false
$p = [System.Diagnostics.Process]::Start($psi)
Start-Sleep 30
$p.Refresh()
"CPU after 30s idle: $([math]::Round($p.TotalProcessorTime.TotalSeconds,2))s  RSS: $([math]::Round($p.WorkingSet64/1MB,0))MB"
$p.Kill()
Result:

text
CPU after 30s idle: 5.78s   RSS: 16MB
That's ~5.78s CPU over 30s wall-clock ≈ 19% of a core, with no requests in flight and stdin idle. A correct stdio server should be doing a blocking read here (~0% CPU). This strongly suggests a busy-wait / poll loop on the idle stdin (or on the daemon connection) rather than a blocking wait.

Why it matters in practice

My editor spawns one frontend per session/subagent, and they are long-lived. I observed 23 frontends under a single client process, each burning ~40–110s CPU over a few minutes — together saturating multiple cores while completely idle. The daemon itself is fine (low CPU); the cost is entirely in the frontends. During this period the daemon log was quiet (no incoming prompts/get / tools/list), yet CPU kept climbing — confirming the spin is independent of request traffic.

Environment

codebase-memory-mcp v0.10.8 (--version confirmed), latest release
Windows (amd64)
Reproduces with a hand-started frontend, so it's independent of any specific MCP client
Happy to capture a CBM_DIAGNOSTICS=1 trajectory or a stack sample if that helps pinpoint the loop.

### Reproduction

<#
.SYNOPSIS
    Measure the idle CPU usage of a codebase-memory-mcp frontend process.

.DESCRIPTION
    Starts a single codebase-memory-mcp frontend, keeps stdin open, sends NO
    JSON-RPC request at all, and measures how much CPU the process burns while it
    is idle. A correct stdio MCP server should use ~0% CPU here (blocking read on
    stdin). If you measure substantial usage, that points to a busy-wait / poll loop.

    This tests the binary completely independently of opencode / OpenChamber.

.PARAMETER Exe
    Path to codebase-memory-mcp.exe. Defaults to the known install location.

.PARAMETER Seconds
    How long to measure while idle (default 30s).

.PARAMETER Samples
    Number of intermediate measurements (default 3), so you can see CPU growth over time.

.EXAMPLE
    powershell -ExecutionPolicy Bypass -File .\test-cbm-idle-cpu.ps1

.EXAMPLE
    powershell -ExecutionPolicy Bypass -File .\test-cbm-idle-cpu.ps1 -Seconds 60 -Samples 6
#>
[CmdletBinding()]
param(
    [string]$Exe = "<path_to_codebase>\codebase-memory-mcp\codebase-memory-mcp.exe",
    [int]$Seconds = 30,
    [int]$Samples = 3
)

$ErrorActionPreference = 'Stop'

if (-not (Test-Path $Exe)) {
    Write-Error "codebase-memory-mcp not found at: $Exe`nProvide the correct path with -Exe."
    exit 1
}

# Show version
$version = (& $Exe --version 2>&1) -join ' '
Write-Host "Binary  : $Exe"
Write-Host "Version : $version"
Write-Host "Measure : $Seconds s idle, $Samples samples (stdin open, NO requests)"
Write-Host ("-" * 60)

# Start the frontend with stdin/stdout/stderr redirected so the process thinks it
# was launched by an MCP client and keeps waiting on stdin.
$psi = New-Object System.Diagnostics.ProcessStartInfo
$psi.FileName               = $Exe
$psi.RedirectStandardInput  = $true
$psi.RedirectStandardOutput = $true
$psi.RedirectStandardError  = $true
$psi.UseShellExecute        = $false
$psi.CreateNoWindow         = $true

$p = [System.Diagnostics.Process]::Start($psi)
Write-Host "Started PID $($p.Id) - NOTHING is sent to stdin.`n"

$interval = [math]::Max(1, [int]($Seconds / $Samples))
$elapsed  = 0
$prevCpu  = 0.0

try {
    for ($i = 1; $i -le $Samples; $i++) {
        Start-Sleep -Seconds $interval
        $elapsed += $interval
        $p.Refresh()

        if ($p.HasExited) {
            Write-Warning "Process exited prematurely (exit=$($p.ExitCode)). stderr:"
            Write-Host ($p.StandardError.ReadToEnd())
            break
        }

        $cpu    = [math]::Round($p.TotalProcessorTime.TotalSeconds, 2)
        $rss    = [math]::Round($p.WorkingSet64 / 1MB, 0)
        $delta  = [math]::Round($cpu - $prevCpu, 2)
        # % of one core over this interval
        $pct    = [math]::Round(($delta / $interval) * 100, 1)
        $prevCpu = $cpu

        Write-Host ("t={0,3}s | CPU total={1,6}s | +{2,5}s this interval | ~{3,5}% of 1 core | RSS={4}MB" -f `
            $elapsed, $cpu, $delta, $pct, $rss)
    }
}
finally {
    if (-not $p.HasExited) {
        # Average over the whole run
        $p.Refresh()
        $totalCpu = [math]::Round($p.TotalProcessorTime.TotalSeconds, 2)
        $avgPct   = [math]::Round(($totalCpu / $elapsed) * 100, 1)
        Write-Host ("-" * 60)
        Write-Host ("FINAL RESULT: {0}s CPU over {1}s idle = ~{2}% of 1 core (average)" -f `
            $totalCpu, $elapsed, $avgPct)
        Write-Host ""
        if ($avgPct -lt 2) {
            Write-Host "OK: idle usage is negligible (blocking read, correct behavior)." -ForegroundColor Green
        }
        else {
            Write-Host "PROBLEM: idle frontend burns substantial CPU -> points to a busy-wait / poll loop." -ForegroundColor Yellow
        }
        Write-Host "`nCleaning up process (PID $($p.Id))..."
        $p.Kill()
        $p.WaitForExit()
    }
    if ($psi) { $p.Dispose() }
}

### Logs

```text

```

```text

```

_No response_

### Confirmations

lorenzozanee:
## Summary

Add focused regression coverage for the idle frontend maintenance probe cadence, preventing the old hot polling loop from returning during quiet sessions.

## Reviewers

@DeusData, @CaptainMittens, @LazyXuan

Fixes #2093

### DIFF
--- a/tests/test_daemon_frontend.c
+++ b/tests/test_daemon_frontend.c
@@ -1359,6 +1359,26 @@ TEST(daemon_frontend_idle_uses_one_maintenance_observer) {
     ASSERT_TRUE(fixture_closed);
     PASS();
 }
+
+/* An idle monitor must not repeatedly perform the expensive presence probe
+ * while no maintenance request exists. The short window is below the
+ * production idle cadence, so a second observation indicates a hot poll loop. */
+TEST(daemon_frontend_idle_maintenance_probe_is_bounded) {
+    frontend_maintenance_fixture_t fixture;
+    ASSERT_TRUE(frontend_maintenance_fixture_start(&fixture, "idle-cadence"));
+
+    cbm_daemon_frontend_test_observer_reset(false);
+    cbm_daemon_maintenance_monitor_t *monitor = cbm_daemon_maintenance_monitor_start(
+        fixture.manager, NULL, NULL, EXIT_SUCCESS, "idle cadence test");
+    ASSERT_TRUE(monitor != NULL);
+    cbm_usleep(100 * 1000U);
+    uint64_t observations = cbm_daemon_frontend_test_monitor_observations();
+    ASSERT_TRUE(cbm_daemon_maintenance_monitor_stop(&monitor));
+    frontend_maintenance_fixture_finish(&fixture);
+
+    ASSERT_EQ(observations, 1);
+    PASS();
+}
 #endif
 
 #ifndef _WIN32
@@ -1663,6 +1683,7 @@ SUITE(daemon_frontend) {
     RUN_TEST(daemon_frontend_rejects_non_notification_cancellation_shapes);
 #if defined(CBM_ENABLE_TEST_SEAMS) && CBM_ENABLE_TEST_SEAMS
     RUN_TEST(daemon_frontend_idle_uses_one_maintenance_observer);
+    RUN_TEST(daemon_frontend_idle_maintenance_probe_is_bounded);
 #endif
 #ifndef _WIN32
     RUN_TEST(daemon_frontend_maintenance_exits_while_stdio_reader_is_blocked);


DeusData:
Thank you for adding a focused regression test around idle frontend maintenance probes. Current main distinguishes the idle maintenance cadence from the active one in src/daemon/frontend.c; pinning that behavior is useful coverage for the reports in #2093. This is queued for review, and we need a little more time before giving a decision.

PR #2103, opened 2026-09-08
fix(daemon): trust the kernel overflow uid as an ancestor owner inside a restricted user namespace

### LINKED ISSUE #1830: Linux: ancestor walk fails unconditionally in Docker userns-remap containers (single-uid mapping, `/` owned by nobody/65534)

AmirF194:
## What does this PR do?

Inside a single-uid Docker userns-remap container (or some WSL2 devcontainer
setups), `cbm` fails to start with "secure CLI coordination could not be
created" on every path. The ancestor-owner trust check in
`posix_directory_owner_trusted` (`src/daemon/ipc.c`) only accepts an
ancestor owned by uid 0 or the caller's own euid. With a single-uid
mapping, uid 0 has no entry in the container's uid map, so the kernel
renders any unmapped host owner, including real root which owns `/`,
`/home` and `/tmp`, as the kernel overflow uid (65534 by default,
configurable) instead. Neither branch matches, so every ancestor is
refused.

The fix extends the trust check to also accept the overflow uid, read
once from `/proc/sys/kernel/overflowuid`, but only when this process's
own namespace genuinely has no mapping for uid 0 (read from
`/proc/self/uid_map`). Outside a restricted namespace the overflow uid is
an ordinary account (a service dropped to "nobody"), not a stand-in for
unmapped root, so it is not trusted unconditionally. The leaf private
directory's own ownership check is untouched: it still requires an exact
match against the caller's euid before it is chmod'd to 0700.

## Checklist

- [x] Tests pass locally (`scripts/test.sh --suites daemon_ipc`, 49/49)
- [x] Lint passes (`clang-format-20 --dry-run --Werror`, `cppcheck`)

## Verification

- New test `daemon_ipc_posix_private_directory_admits_overflow_uid_ancestor_inside_restricted_userns`
  goes red when just this fix's hunk in `posix_directory_owner_trusted` is
  reverted (`ASSERT(secured)`), and passes with it restored.
- New test `..._refuses_overflow_uid_ancestor_outside_restricted_userns` is
  a negative control: it stays green either way (the ordinary case is
  unchanged), and I confirmed it is not vacuous by temporarily making the
  trust check always return `true`, which turns it red.
- Full `scripts/test.sh --suites daemon_ipc` (49/49), `clang-format-20`
  and `cppcheck` all clean.
- I have not tried this against an actual Docker userns-remap container.
  Building a real restricted user namespace needs `CAP_SYS_ADMIN`, which
  is unavailable here even inside a fresh container (`unshare --user`
  fails `EPERM`), so both new tests fabricate the ownership condition
  directly and pin the namespace state through a test-only seam rather
  than a real namespace.

Fixes #1830

### DIFF
--- a/src/daemon/ipc.c
+++ b/src/daemon/ipc.c
@@ -1385,7 +1385,71 @@ static char *private_log_directory_path_copy(const char *directory_path) {
     return string_copy(directory_path);
 }
 
+#ifdef __linux__
+/* Inside a single-uid Docker userns-remap (or some WSL2 devcontainer) setups,
+ * uid 0 has no entry in the container's uid map, so the kernel renders any
+ * host-owned path whose real owner is unmapped (including real root, which
+ * owns "/", "/home", "/tmp") as the overflow uid instead. 65534 is the
+ * kernel default but is configurable, so read it rather than assume it. */
+uid_t cbm_daemon_ipc_posix_kernel_overflow_uid(void) {
+    FILE *overflow_file = fopen("/proc/sys/kernel/overflowuid", "r");
+    if (!overflow_file) {
+        return (uid_t)65534;
+    }
+    unsigned long parsed = 0;
+    int scanned = fscanf(overflow_file, "%lu", &parsed);
+    (void)fclose(overflow_file);
+    return scanned == 1 ? (uid_t)parsed : (uid_t)65534;
+}
+
+/* -1 = no override (read /proc/self/uid_map for real), 0 = force "mapped"
+ * (ordinary namespace), 1 = force "unmapped" (restricted userns). A real
+ * restricted userns needs CAP_SYS_ADMIN to construct, which CI runners do
+ * not grant, so tests pin this instead of shelling out to unshare(1). */
+static atomic_int g_posix_uid_zero_unmapped_override_for_test = ATOMIC_VAR_INIT(-1);
+
+void cbm_daemon_ipc_posix_uid_zero_unmapped_override_set_for_test(int override) {
+    atomic_store_explicit(&g_posix_uid_zero_unmapped_override_for_test, override,
+                          memory_order_release);
+}
+
+/* The overflow uid is only a stand-in for "unmapped root" inside a restricted
+ * user namespace. Outside one, uid 0 is mapped to itself and the overflow uid
+ * is an ordinary, sometimes-real account (services that drop privileges to
+ * "nobody"); trusting it unconditionally would let any ancestor a compromised
+ * nobody-owned service can write into pass this check. So only relax the
+ * check when our own namespace genuinely has no mapping for uid 0, read from
+ * /proc/self/uid_map (one line per contiguous range: "inner outer count"). */
+bool cbm_daemon_ipc_posix_uid_zero_unmapped(void) {
+    int override =
+        atomic_load_explicit(&g_posix_uid_zero_unmapped_override_for_test, memory_order_acquire);
+    if (override >= 0) {
+        return override != 0;
+    }
+    FILE *uid_map_file = fopen("/proc/self/uid_map", "r");
+    if (!uid_map_file) {
+        return false;
+    }
+    unsigned long inner = 0, outer = 0, count = 0;
+    bool zero_mapped = false;
+    while (fscanf(uid_map_file, "%lu %lu %lu", &inner, &outer, &count) == 3) {
+        if (inner == 0) {
+            zero_mapped = true;
+            break;
+        }
+    }
+    (void)fclose(uid_map_file);
+    return !zero_mapped;
+}
+#endif
+
 static bool posix_directory_owner_trusted(uid_t owner) {
+#ifdef __linux__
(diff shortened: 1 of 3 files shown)


DeusData:
Thank you for contributing the restricted-user-namespace case and the focused IPC tests. We need more time for a careful review of this ownership-policy change before giving a decision.

AmirF194:
Closing this: #2138 already landed the fix for #1830, and it covers more ground than this PR did. Mine trusted the overflow uid whenever uid 0 itself was unmapped; #2138 requires the whole uid_map to be a single-uid map, which also refuses a multi-uid subuid setup where uid 0 happens to be unmapped too. That's the right scope and this PR would only add a competing, narrower version of the same check.

PR #2140, opened 2026-09-10
feat(cli): point top-level cli --help at --format json

### LINKED ISSUE #2102: feat(cli): top-level --format / CBM_CLI_FORMAT default for tools (follow-up from #1867)
# Split out of #1867 per @DeusData's request

Split out from #1867 (secondary finding there: `search_graph` / `trace_path` / `detect_changes` emit the compact tree by default when stdout is not a TTY), per [the maintainer's invitation](https://github.com/DeusData/codebase-memory-mcp/issues/1867#issuecomment-5575970029) to give it its own discussion.

## Current state (verified on v0.10.8) — better than my original report assumed

- Tools that accept a `format` argument (notably `search_graph` / `trace_path` / `detect_changes`) already support `--format json`, and `<tool> --help` documents it (e.g. `search_graph --help`: "Response encoding. tree (default) … json: the SAME tree model as structured JSON")
- The README CLI section documents both layers: tool-level `--format json` (payload JSON) vs outer `--json` (full MCP envelope)
- The tree default is deliberate, and I understand why: the primary non-TTY consumers of this CLI are AI agents and bridges, and the compact tree is token-optimized for exactly that audience

So this is **not a bug report** — it is a small UX/discoverability suggestion.

## Remaining friction

1. **`--format json` is per-tool-call only.** A script or bridge invoking several tools must pass it on every single invocation; there is no way to set it once for a session.
2. **The top-level `cli --help` doesn't mention it.** You have to run `<tool> --help` for each tool to discover the flag. The README covers it, but `--help` is where habitual CLI users look first.

## Suggestion — pick whichever direction you prefer; happy to submit a PR

- **(a) Top-level flag:** `cli [--format tree|json] <tool> ...` acting as a *default* for tools that accept `format`; an explicit per-tool `--format` still wins.
- **(b) Env var:** `CBM_CLI_FORMAT=tree|json`, useful for bridges that spawn the binary from generated glue code without full control over argv on every call.
- **(c) Docs-only:** add one line to the top-level `cli` usage string pointing at tool-level `--format json`.

I'd lean towards **(a) + (b)**: both are fully backward compatible, and they deliberately do **not** flip the default on non-TTY stdout — auto-switching would silently change the output that model-facing bridges and agents see, which I assume is the audience the tree default was designed for.

I can implement this myself if you agree on a direction. From reading the source, the change looks localized:

- `run_cli()` in `src/main.c` — merge a top-level/env-provided `format` into `args_json` before tool dispatch (per-tool explicit value still takes precedence)
- `tests/test_cli.c` — precedence tests (top-level vs per-tool vs unset)
- `CLI_USAGE` string + README CLI section
- Per CONTRIBUTING (pre-commit hooks, `scripts/lint.sh`, DCO `Signed-off-by` trailer)

Lowpower:
## What does this PR do?

Help-only discoverability fix for #2102 (direction **c** only).

Top-level `cli --help` already described outer `--json` (the raw MCP result envelope) but did not mention that tools which accept `format` support `--format tree|json`. Habitual CLI users look at `cli --help` first; per-tool `--help` already documented `--format`.

This PR adds two lines to `CLI_USAGE`:

- Tools that accept format support `--format tree|json` (default: tree).
- `--format json` prints payload JSON; outer `--json` prints the full MCP envelope.

No session-wide `--format` flag, no `CBM_CLI_FORMAT` env var, and no change to output defaults.

`CBM_CLI_USAGE` lives in `src/cli/cli.h` so `tests/test_cli.c` can assert on the same string `run_cli()` prints.

## Checklist

- [x] Every commit is signed off (`git commit -s`) — required, CI rejects unsigned commits
- [x] CLI suite passes locally (`scripts/test.sh --suites cli` — including `cli_usage_points_to_tool_format_json_issue2102`)

Fixes #2102


DeusData:
Thank you for improving top-level help. The current diff also changes a daemon timeout test and retries the `search_code` JSON check, which are independent of help discoverability. Please separate those changes and keep this PR to the help text plus a deterministic help regression. The outer CLI already has a JSON-format path; the test should establish the advertised help contract without retrying another operation.

DeusData:
A quick request, because it is costing you time rather than saving it: **please do not rebase onto `main` again until this is merged.** Your previous head was one CI job from green — everything else had passed and the merge was queued on my side — and the rebase to `28b4093c` discarded that run and started a full matrix from zero, which takes about two hours on tonight's runner pool. The content is identical (same two commits), so nothing was gained.

`main` will keep moving tonight; that does not matter for this PR. The merge result is already verified locally against the current `main` (`cli` 320, linter unchanged, and the test does not compile without the production change). As soon as this run is green it merges. Nothing else is needed from you.

DeusData:
Merged — thank you @Lowpower, and thank you for splitting it exactly as asked: the PR is now the help text, one macro in `cli.h`, and a deterministic test that pins the wording, including that the synopsis still has no session-wide `--format`.

Verified on the merge result with the current `main` (`e783f73d`): `cli` 320 passed, the memory-core linter is unchanged, and with the production change reverted the test does not even compile (`CBM_CLI_USAGE` undeclared) — the strongest binding a test can have. The reds you saw on earlier runs were ours, not yours: a timing-sensitive MSan test that #2272 fixes, and a Windows cold-start race that #2275 fixes.

Small, and it removes a real point of confusion between `--json` and `--format json`.

PR #2158, opened 2026-09-10
fix(routes): accept relative JAX-RS @Path templates and compose Scala class-level @Path (#2147)

### LINKED ISSUE #2147: Java/Scala JAX-RS: relative @Path templates (no leading slash) produce no Route nodes; Scala class-level @Path not composed
### Version

JAX-RS allows `@Path` values without a leading slash, both on the class and on the method (`@Path("widgets")`, `@Path("{id}")`). The route extractor only accepts templates that start with `/`, so such resources produce **no `Route` nodes and no `HANDLES` edges at all**. In Java frameworks this spelling is common (JAX-RS spec, Jersey, RESTEasy, Quarkus samples).

A second, related gap: for Scala JAX-RS resources the class-level `@Path` prefix is not joined with the method path (it is for Java and Kotlin), so a Scala method route is emitted as the bare method template instead of `<class path>/<method path>`.

Expected: relative templates are accepted and normalized so the composed path is always rooted (`/widgets/{id}`), for Java, Kotlin and Scala alike.

Follow-up to #1005 (PR #1007 fixed class/method composition for rooted `@Path` values; templates without a leading slash are still rejected before composition runs).

### Reproduction

1. Dummy snippet (Java):

```java
@Path("widgets")
public class WidgetResource {
    @GET @Path("{id}")
    public Widget get(@PathParam("id") String id) { return null; }
}
```

Scala variant:

```scala
@Path("/widgets")
class WidgetResource {
  @GET @Path("{id}")
  def get(@PathParam("id") id: String): Widget = null
}
```

2. Command: `codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/repro"}'` then
   `codebase-memory-mcp cli search_graph '{"project":"repro","label":"Route"}'`

3. Result: no `Route` node for the Java resource; for the Scala resource a `Route` with path `{id}`.
   Expected: `Route` `/widgets/{id}` with a `HANDLES` edge from `WidgetResource.get` in both cases.

### Proposed fix

I have a two-commit fix with tests (relative-template acceptance + Scala class-level composition), rebased on current `main`:
https://github.com/htarnacki/codebase-memory-mcp/tree/fix/jaxrs-relative-path — I will open a PR referencing this issue.

### Confirmations

htarnacki:
## What does this PR do?

Fixes #2147.

JAX-RS allows `@Path` templates without a leading slash (`@Path("widgets")`, `@Path("{id}")`); the route extractor rejected them, so such resources produced no `Route` nodes and no `HANDLES` edges. Two commits:

1. `fix(routes): accept relative JAX-RS @Path templates` — accept relative templates on class and method level and normalize the composed path so it is always rooted (`/widgets/{id}`). Existing tests updated to use the relative spelling for the JAX-RS cases, a rooted case is kept.
2. `fix(routes): compose class-level @Path for Scala JAX-RS resources` — the class-level prefix was joined with the method path for Java and Kotlin but not for Scala; Scala now composes the same way. New pipeline test.

No behaviour change for Spring / other frameworks: rooted templates compose exactly as before (follow-up to #1005).

## Checklist


htarnacki:
CI status: everything is green except `test / test-windows-guards`, which went red in `tests/windows/test_daemon_stability.py::section_cold_storm` ("cold-storm client 0 failed (racing daemon spawn): secure CLI coordination could not be created (endpoint)").

I believe that one is pre-existing / environmental rather than caused by this PR: the change is confined to JAX-RS `@Path` extraction in `internal/cbm/extract_defs.c` plus tests, and the same guard failed in the same way on recent unrelated PR runs (`fix/decision-b-win-sid-trust`, `fix/1954-uninstall-symlinked-config`, `ci/windows-helper-spawn-timeout`). Happy to rebase/re-push to re-run if you prefer.

DeusData:
Thank you for the JAX-RS relative-path cases and Scala coverage. Current extraction rejects the relative forms at the identified path, so these regressions address a concrete gap. Please retain the rooted, non-JAX-RS negative control so the change cannot accidentally broaden unrelated route extraction. No empty commit or CI-only push is being requested.

htarnacki:
Updated per your note — the branch base is unchanged, only the test coverage moved:

- The original rooted fixtures (`@Path("/api/v1/widgets")` / `@Path("/count")`) in `extract_java_jaxrs_path_composition_issue1005` and `handles_jaxrs_java` are restored unmodified.
- Relative templates are now covered by separate tests: `extract_java_jaxrs_relative_path_templates` (class + method + `{id}/tags` template) and `handles_jaxrs_java_relative_path`.
- Added a non-JAX-RS negative control in both suites (`extract_java_spring_relative_string_not_route_path`, `handles_spring_java_relative_string_not_route`): `@RequestMapping("api")` + `@GetMapping("orders")` still yields the `/` fallback and no `/api/orders` Route node.

Verified locally by swapping in `main`'s `extract_defs.c`: rooted fixtures and both negative controls pass there, the relative variants fail there and pass with this change.

DeusData:
Thank you — this is exactly what was asked for, and the way you verified it is the part I appreciate most: swapping in `main`'s `extract_defs.c` and showing that the rooted fixtures and both negative controls pass there while the relative variants fail is a proper fails-without-the-fix proof, not just a green run.

Two things on our side:

**The `test / test-msan` red is ours.** It failed on `daemon_bootstrap_fails_fast_when_daemon_dies_at_publication` (`tests/test_daemon_bootstrap.c:1116`). We attributed that test yesterday: before it listens, the daemon hashes its own ~450 MB test image, which takes about 26 s under MSan against a 30 s client budget, so the verdict rides on a few seconds of runner speed. Nothing in a route-extraction change can influence it. Rebuilding that test so it no longer depends on timing is on our list.

**I have updated your branch with `main`** (it was 93 commits behind), which starts a fresh full CI run against today's tree. That is the verdict that counts; if the MSan leg trips on the same test again I will rerun it from here. Nothing further is needed from you for now — the next step is mine: reading the new tests against the restored originals and building the merge result locally.

DeusData:
The local half is done, on the merge result of your head `61cfb09b` with today's `main`:

| check | result |
|---|---|
| build (`-Werror`, ASan + UBSan) | clean |
| `extraction` | 371 passed |
| `edge_types_probe` | 62 passed |
| `pipeline` / `lang_contract` | 281 / 41 passed |
| memory-core linter | no growth |
| **production reverted, your tests kept** | `extract_java_jaxrs_relative_path_templates` fails with `"/" != "/api/v1/widgets"`, and the route probe reports `/api/v1/widgets` and `/api/v1/widgets/count` missing |

So the tests bind to the fix, which is the property that matters, and the change is scoped the way it should be: `allow_relative` is `true` at exactly one call site (the JAX-RS `@Path` branch) and `false` at every generic mapping, so nothing else can start inventing routes from bare strings. Adding Scala to the class-level composition is right too.

**One small edge, and it is a behaviour change, so I would like it closed before this merges.** `route_path_from_string_node` now ends with

```c
return path[0] ? cbm_arena_sprintf(a, "/%s", path) : path;
```

For `@Path("")` that returns the empty string rather than `NULL`. On `main` an empty value is rejected (`path[0] != '/'`), the method's path stays unset, and `extract_route_from_annotations` falls back to `"/"`; with this change `jax_path` is `""`, which is non-NULL, so `*out_path` becomes `""`. `@Path("")` on a method is legal JAX-RS and means "the class path itself". Returning `NULL` for the empty case keeps today's behaviour:

```c
return path[0] ? cbm_arena_sprintf(a, "/%s", path) : NULL;
```

plus one assertion for `@Path("")` next to your relative-template test. If you push it now it costs nothing: your CI run is still queued behind a busy runner pool, and a new push simply replaces it.

After that and a green matrix this merges. Thank you for the careful follow-through on the negative controls.

htarnacki:
Closed the `@Path("")` edge in `246e4190`, on top of your merge commit (no rebase):

```c
return path[0] ? cbm_arena_sprintf(a, "/%s", path) : NULL;
```

One note on the test shape. A single assertion next to the relative-template test would not have bound to the fix: with a class-level `@Path` present, `join_route_paths` already treats an empty method path as "use the prefix", so `""` and the `"/"` fallback compose to the same `/api/v1/widgets`. The case where the behaviour change is observable is a resource **without** a class-level `@Path` — there `join_route_paths(NULL, "")` returns the `NULL` prefix and the route is dropped altogether, whereas `main` yields `"/"`.

So the regression is a dedicated `extract_java_jaxrs_empty_path_means_class_path` covering both shapes:

| fixture | expected | with `: path` (previous revision) |
|---|---|---|
| `@Path("api/v1/widgets")` class, method `@GET @Path("")` | `/api/v1/widgets` | passes (masked by composition) |
| class without `@Path`, method `@GET @Path("")` | `/` | **fails**: `index->route_path is NULL` |

Verified locally: with the fix `extraction` 372 passed / `edge_types_probe` 62 passed; with the previous `: path` line the new test fails at the second row.

The 4 `clang-format` findings in `tests/test_extraction.c` (around the `parseJsonBody<T>` TypeScript fixture) come in with the merge from `main`, not from this branch; I left them alone.

PR #2160, opened 2026-09-10
fix(cli): accept JSON array literals for array-typed arguments (#2150)

### LINKED ISSUE #2150: cli: JSON array value for array-typed flags (--semantic-query '["a","b"]') is stored as a single string element
### Version

`codebase-memory-mcp cli <tool> --flag value` builds the tool's JSON arguments from typed flags. Array-typed parameters such as `semantic_query` are only accepted as **repeated flags** (`--semantic-query a --semantic-query b`). Passing the JSON spelling that the MCP schema documents (`--semantic-query '["a","b"]'`) is stored as a single string element, so the search runs against the literal text `["a","b"]` instead of two keywords. Agents that copy the argument shape from the tool schema hit this every time.

Expected: a value that is a JSON array of strings is expanded into individual elements; every other value keeps today's behaviour (repeated flags and plain strings unaffected).

### Reproduction

1. Any indexed project.
2. Command:
   `codebase-memory-mcp cli search_graph --project repro --semantic-query '["create","billing","account"]'`
3. Result: the request JSON contains `"semantic_query":["[\"create\",\"billing\",\"account\"]"]` (one element).
   Expected: `"semantic_query":["create","billing","account"]`.

Unit-level: `cbm_cli_build_args_json("search_graph", 2, {"--semantic-query", "[\"create\",\"billing\",\"account\"]"}, &err)`.

### Proposed fix

In `cli_add_typed`, when the value starts with `[` and parses as a non-empty JSON array of strings, add its elements; otherwise fall through to the existing single-element path. One-commit fix with a `test_cli.c` case:
https://github.com/htarnacki/codebase-memory-mcp/tree/fix/cli-json-array-args — I will open a PR referencing this issue.

### Confirmations

htarnacki:
## What does this PR do?

Fixes #2150.

`codebase-memory-mcp cli <tool> --flag value` stored a JSON array literal passed to an array-typed flag (`--semantic-query '["a","b"]'`) as a single string element, so the search ran against the literal text. `cli_add_typed` now expands a value that starts with `[` and parses as a non-empty JSON array of strings into individual elements; every other value (repeated flags, plain strings, non-array JSON) keeps the current behaviour. One commit, unit test in `test_cli.c`.

## Checklist


DeusData:
Thank you for the CLI target-array correction. Current argument handling can pass array-looking text as a literal value, matching the reported contract gap. Review should preserve repeatable-flag compatibility and explicitly cover empty, malformed, mixed and repeated inputs. Maintainer bundle #2227 overlaps this area, so we need to reconcile the contributions before integration rather than ask you to redo the same work.

htarnacki:
Added the explicit coverage you asked for (same base, single commit amended):

| input | result | test |
|---|---|---|
| `--semantic-query '["a","b"]'` | `["a","b"]` | `cli_build_args_json_array_literal` |
| `--semantic-query '[]'` | `[]` (no items; previously became the literal keyword `"[]"`) | `..._empty` |
| `--semantic-query '["a",'` (malformed) | unchanged behaviour: one literal string, no error | `..._malformed` |
| `--semantic-query '[1,2]'` (non-string items) | unchanged behaviour: one literal string | `..._non_string_items` |
| `a`, `'["b","c"]'`, `d` (mixed) | `["a","b","c","d"]` in argument order | `..._mixed_with_repeated` |
| `'["a"]'`, `'["b"]'` (repeated literals) | `["a","b"]` | `..._repeated` |

The only behavioural change relative to the previous revision is the empty-array case: a well-formed empty array is now expanded (to nothing) instead of falling through to the literal-string path. Everything that is not a well-formed array of strings keeps the pre-existing literal-string behaviour, and the existing repeated-flag test (`cli_build_args_json_repeated_array_issue680`) is untouched.

Issue #1406, opened 2026-08-01
fast mode silently excludes `scripts/` — first-class Python source dropped from graph (built-in skip list?)


Limebird01:
### Version

codebase-memory-mcp 0.9.0

### Platform

Linux (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard (portable static build, linux-amd64)

### What happened, and what did you expect?

`index_repository --mode fast` silently excluded the entire `scripts/` directory of a mid-size Python repo. Only 46 of 156 `.py` files were indexed, and `scripts` appeared in the `excluded.dirs` list of the index output. Re-indexing the same repo with `--mode full` captured everything (5,286 nodes / 17,328 edges, 6.0 s, ~232 MB max RSS).

No `.gitignore` or `.cbmignore` pattern matches `scripts/`. In this repo (as in many Python projects) `scripts/` holds first-class production entry-point modules — it is a common Python layout, not a bin of throwaway files.

Expected: fast mode should reduce **analysis depth** (cheaper resolution strategies, fewer passes), not silently drop first-class source directories from the graph — or at minimum it should emit a visible warning that a top-level source dir was skipped by a built-in heuristic.

This looks like the same silent-exclusion family as #1184 (built-in skip list drops `deploy`/`deployment(s)` dirs, mode full, not overridable via `.cbmignore`) and #411 (closed; moderate mode silently drops subtrees), and #1219 (default `assets` exclusion swallows API route files) — but none of those covers `scripts/` in fast mode specifically. Note the docs state the built-in skip list is not user-overridable, so there is currently no workaround other than switching to `--mode full`.

### Reproduction

1. Code: any generic Python repo with a top-level `scripts/` directory containing real importable modules, e.g.:

   ```
   repo/
     src/pkg/core.py
     scripts/deploy_tool/main.py   # first-class entry points, plain .py modules
     scripts/reporting/report.py
   ```

2. Commands:

   ```
   codebase-memory-mcp cli index_repository --repo-path /path/to/repo --mode fast --persistence false
   codebase-memory-mcp cli index_repository --repo-path /path/to/repo --mode full --persistence false
   ```

3. Result: fast-mode output lists `scripts` under `excluded.dirs`; indexed file count drops to 46/156 `.py`; every symbol under `scripts/` is absent from the graph. Full mode indexes all 156 files. Expected: identical file coverage in both modes (only analysis depth should differ), or an explicit warning naming the skipped directory.

[...]

Issue #1184, opened 2026-07-20
Built-in skip list silently excludes source dirs named `deploy`/`deployment(s)` — not overridable via `.cbmignore`


sakivks:
## Summary

The hardcoded directory skip list includes `deploy`, `deployment`, and `deployments`. On repos where deployment logic **is the product domain** (deploy control planes, CD tooling, IaC orchestrators), this silently drops core source code from the graph, and `docs/cbmignore.md` states the built-in list is "Not overridable from any ignore file today."

## Repro (v0.9.0, darwin-arm64)

Next.js/TypeScript repo with real source at `src/lib/deploy/` (23 .ts files) and `src/app/api/webhooks/deploy/`:

```
codebase-memory-mcp cli index_repository --repo-path <repo> --mode full
```

Response:

```json
"excluded":{"dirs":[".claude",".git",".vscode",".superpowers","src/lib/deploy","src/app/api/webhooks/deploy"],"count":6}
```

- No `.gitignore` / nested gitignore / global excludes / `.cbmignore` match these paths (`git check-ignore` exits 1).
- `strings` on the binary confirms `deploy`, `deployment`, `deployments` in the built-in name list.
- Symbols from those dirs (verified via `search_graph`) are absent from the graph.
- A `.cbmignore` negation can't rescue them — per docs, negation only overrides the git-global-excludes layer.

## Why it matters

Unlike `dist`/`target`/`vendor`, `deploy` is a common **source** directory name (deployment engines, `deploy/` K8s dirs that the README's own IaC-indexing feature targets). The exclusion is silent unless the user inspects `excluded` — and it's capped at 25 entries, so on bigger repos it may not even be visible there.

## Suggested fixes (any of)

1. Let `.cbmignore` negation (`!src/lib/deploy/`) override the built-in list — most flexible.
2. Remove `deploy`/`deployment(s)` from the always-on list (or demote them to fast/moderate modes only, like `docs`/`examples`).
3. Only skip these names when they contain no parseable source files (heuristic).

## Workaround used

Index the excluded dirs as companion projects:

```
codebase-memory-mcp cli index_repository --repo-path <repo>/src/lib/deploy --name myrepo-deploy-lib --mode full
```

— works, but loses cross-file edges into the main project graph.

DeusData:
[...]

PR #2161, opened 2026-09-10
fix(discover): retain external source directories in fast mode (#2151)

### LINKED ISSUE #2151: fast/moderate discovery silently skips every directory named external (FAST_SKIP_DIRS)
### Version

`external` is listed in `FAST_SKIP_DIRS` (`src/discover/discover.c`), so **fast and moderate discovery silently drop every directory named `external`**. In many codebases that name holds first-party code — external-facing API modules, adapters, integration layers — not vendored third-party sources. The vendored case is already covered by the `vendor`, `vendored`, `third_party`, `thirdparty` and `3rdparty` entries.

The effect is hard to notice: the indexing summary reports fewer files, but nothing indicates that a whole source tree was excluded, and `trace_path` / `search_graph` simply return nothing for symbols that live there.

Expected: `external` is indexed in every mode (only genuinely vendored directory names are skipped in fast mode).

Same silent-exclusion family as #1406 (`scripts/`) and #1184 (`deploy`/`deployment`); this report is scoped to the single `external` entry, for which a one-line fix is enough.

### Reproduction

1. Dummy layout:
   ```
   repro/
     core/Service.java
     external/PublicApi.java      # first-party, references core.Service
   ```
2. Command: `codebase-memory-mcp cli index_repository '{"repo_path":"/tmp/repro","mode":"fast"}'` then
   `codebase-memory-mcp cli search_graph '{"project":"repro","name_pattern":"PublicApi"}'`
3. Result: no node. Expected: `PublicApi` and its CALLS edge into `core.Service`.

Unit-level: `cbm_should_skip_dir("external", CBM_MODE_FAST)` returns `true`.

### Proposed fix

Remove `external` from `FAST_SKIP_DIRS`; add `no_skip_fast_external` to `test_discover.c`. Note: clang-format-20 switches the shortened initializer to one-item-per-line, so the list is regrouped with category comments in the same style as `ALWAYS_SKIP_DIRS` (order unchanged).
https://github.com/htarnacki/codebase-memory-mcp/tree/fix/discover-external-fast-mode — I will open a PR referencing this issue.

### Confirmations

htarnacki:
## What does this PR do?

Fixes #2151.

`external` was listed in `FAST_SKIP_DIRS`, so fast and moderate discovery silently dropped every directory with that name — in many codebases first-party code (external-facing APIs, adapters, integration layers). The vendored case remains covered by `vendor`, `vendored`, `third_party`, `thirdparty`, `3rdparty`. This removes the single `external` entry and adds `no_skip_fast_external` to `test_discover.c`.

Diff note: `clang-format-20` reflows the shortened initializer to one item per line, so the list is regrouped with category comments in the same style as `ALWAYS_SKIP_DIRS`; order and the other entries are unchanged.

Related: #1406, #1184 (same silent-exclusion family; not addressed here).

## Checklist


DeusData:
Thank you for the discovery fix and focused mapping regression. Current non-full discovery still applies the built-in `external` exclusion, so the test targets an identifiable discovery path. Review will need to retain the distinction between this directory heuristic and the remaining ignore rules; this is not approval to relax every exclusion.

PR #2164, opened 2026-09-10
fix(php): add fallback for self:: static calls when enclosing_class_q…

Carnival-z:
…n is NULL

Problem:
PHP LSP's resolve_static_call() returns early when enclosing_class_qn is NULL, causing self::/static::/parent:: calls to fall back to low-confidence heuristic strategies (unique_name 0.75 or suffix_match 0.55) instead of high-confidence php_self_static (0.95).

Root Cause:
When process_class_decl doesn't populate enclosing_class_qn (e.g., complex method bodies with closures), resolve_static_call short-circuits without generating any LSP resolution, leaving pass_calls.c to use registry fallbacks.

Fix:
Add fallback logic to infer class_qn from enclosing_func_qn when handling self::/static:: scopes. Extract class QN by stripping the last .method segment from "Project.path.Class.method" → "Project.path.Class", then continue with normal method lookup and emit php_self_static strategy at 0.95 confidence.

Impact:
- PHP only (no other languages affected)
- Improves self:: call resolution from 0.55-0.75 → 0.95 confidence
- Verified on iCMS v8.0.0: AdminAdmincp.do_index self::setContext/tableData/view

## Checklist

### DIFF
--- a/internal/cbm/lsp/php_lsp.c
+++ b/internal/cbm/lsp/php_lsp.c
@@ -1760,8 +1760,32 @@ static void resolve_static_call(PHPLSPContext *ctx, TSNode call, CBMResolvedKind
             }
         }
     }
-    if (!class_qn)
-        return;
+    if (!class_qn) {
+        /* Fallback: enclosing_class_qn is NULL (shouldn't happen in well-formed
+         * code, but may occur if process_class_decl didn't run). Emit a low-
+         * confidence self-reference so pass_calls can still attempt registry
+         * resolution instead of dropping the call entirely. */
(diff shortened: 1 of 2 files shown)


DeusData:
Thank you for the PHP investigation. The current diff also changes shared `resolve_name_lookup` behavior for callers without imports across languages, so this is broader than a PHP-only adjustment. Please separate that shared policy decision from the focused PHP correction and add the relevant PHP regression controls, including closure and `parent::` behavior. That will let the concrete fix be reviewed without implicitly approving a cross-language binding change.

PR #2170, opened 2026-09-10
fix(registry): recognize C# lower-case type keywords in receiver_chain_admits

### LINKED ISSUE #2121: BCL/external calls fabricate CALLS edges to unrelated internal classes (phantom callers; e.g. int.TryParse → internal TryParse)
### Version

codebase-memory-mcp 0.10.8

### Platform

Linux (x64)

### Install channel

AUR

### Binary variant

standard

Calls to **BCL / external static members** (`int.TryParse`, `Options.Create`, chained expressions) are being resolved to **unrelated internal classes** in the indexed project, producing fabricated CALLS edges:

- `int.TryParse(input, out _)` → CALLS edge to our internal `TsidNode.TryParse`
- `Options.Create(new object())` → CALLS edge to our internal `CqrsRegisterOptionsGet.Create`
- `Enum.GetValues<AssetRole>().Max` → CALLS edge to an internal class named `Field` (real repo, `unique_name` strategy)

The receiver has no resolvable relationship to these classes — the only thing they share is a **method name**. The edges carry `heuristic` strategy with non-trivial confidence (0.75–0.85 in our data), so they are not distinguishable as guesses by confidence alone.

Consequences:

- **Phantom callers**: internal methods appear to have callers they do not have. In our real repo, 4 phantom callers landed on a class with zero real callers — dead-code analysis would report it as live.
- **Blast-radius inflation**: `detect_changes` / impact maps include nodes that cannot be affected.
- On the real repository (~46K nodes / 121K edges, 11,251 CALLS edges) we measured a sample of 4,800 edges: **667 edges whose receiver text is a PascalCase identifier with no same-named internal type were still attached to an internal class**. Common method names (`Create`, `TryParse`, `Add`, `Get`, `Max`) make any internal class sharing them a phantom magnet.

Expected: calls whose receiver is not resolvable to an indexed type should produce either no CALLS edge or an explicit unresolved marker — never an edge to a class picked by method-name coincidence.

Related but distinct from #2120 (same-name class variant mis-attachment): that issue is about picking the **wrong real candidate**; this one fabricates edges where **no internal call exists at all**.

### Reproduction

Single project, two files.

`Contoso.App.csproj`:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup><TargetFramework>net8.0</TargetFramework></PropertyGroup>
  <ItemGroup><PackageReference Include="Microsoft.Extensions.Options" Version="8.0.0" /></ItemGroup>
</Project>
```

`src/Decoys.cs`:

```csharp
namespace Contoso.App;

public class TsidNode
{
    public static bool TryParse(string text, out TsidNode node) { node = null; return false; }
}

public class CqrsRegisterOptionsGet
{
    public object Create(object value) => value;
}

public class Field
{
    public int Max() => 0;
}
```

`src/Demo.cs`:

```csharp
using Microsoft.Extensions.Options;

namespace Contoso.App;

public class Demo
{
    public string M(string input)
    {
        _ = int.TryParse(input, out _);
        _ = Options.Create(new object());
        _ = Enum.GetValues<DayOfWeek>().Length;
        return input;
    }
}
```

Commands:

```bash
codebase-memory-mcp cli index_repository --repo-path /tmp/repro4 --mode full
codebase-memory-mcp cli query_graph --project tmp-repro4 \
  --query "MATCH (f)-[:CALLS]->(g) WHERE f.name = 'M' RETURN g.qualified_name"
```

Result:

```text
tmp-repro4.src.Decoys.TsidNode.TryParse
tmp-repro4.src.Decoys.CqrsRegisterOptionsGet.Create
```

`Demo.M` contains **zero** calls to internal classes — both edges are fabricated from BCL calls.

Evidence view (`trace_path --include-evidence true --direction outbound`):

```text
function: tmp-repro4.src.Demo.Demo.M
callees_total: 2
  CqrsRegisterOptionsGet.Create  hop 1  heuristic  0.75
  TsidNode.TryParse              hop 1  heuristic  0.75
```

(`Enum.GetValues<DayOfWeek>().Length` produced no edge — chained/property access on the array did not fabricate in this shape.)

### Logs

Included in the reproduction section above (query + trace outputs).

Real repo where observed: 46,186 nodes / 121,081 edges / ~2,500 files. Sample of 4,800 CALLS edges: 667 PascalCase-receiver edges with no same-named internal type attached to internal classes anyway. Repro above: 15 nodes.

### Confirmations

- [x] I searched existing issues and this is not a duplicate. (Companion report to #2120 — different failure mode: fabricated edges vs wrong-variant selection among real candidates.)

AmirF194:
## What does this PR do?

`receiver_chain_admits()` (`src/pipeline/registry.c`) fast-admits a lower-case-rooted receiver chain as a value rather than a type. That's correct for Go, Python and Swift (`vm.load`, `http.Get`, `os.path.join`), but C# spells its built-in types as lower-case keywords (`int`, `string`, `bool`, ...), so a static call like `int.TryParse(...)` skips the chain-consistency check entirely and can fabricate a `CALLS` edge to any unrelated method sharing the same name (#2121).

Fix: a closed set of C# type keywords is checked only when the receiver root is lower-case; anything not in that set still fast-admits exactly as before, so other languages' lower-case value roots are unaffected.

Fixes #2121

## What does this PR do NOT fix?

The issue's second example, `Options.Create(new object())`, no longer reproduces at HEAD: PR #1897 (merged 2026-09-04) already added the chain-consistency guard this PR extends, and it happens to cover that upper-case-rooted shape already.

## How was this verified?

- Added `registry_receiver_chain_refuses_lowercase_type_keyword_issue2121`. Without the fix, `ASSERT_NULL` in that test trips because the fabricated edge still resolves; with the fix it's clean. Ran both ways in the same Docker image (`ubuntu:24.04`, ASan+UBSan test build).
- Reproduced the issue's own two-file C# example end to end with the built CLI (`index_repository` + `trace_path`): `int.TryParse` fabricates an edge to an unrelated `TryParse` on main; zero edges on this branch.
- Added a second test, `registry_receiver_chain_keeps_type_keyword_when_chain_matches_issue2121`, confirming the guard still resolves a real project type whose chain matches, so it isn't just refusing everything lower-case.
- Full `pipeline` suite (279 tests) green on this branch; `cppcheck` and `clang-format --dry-run` clean on the changed file. Not checked against the exact CI toolchain: local cppcheck was 2.13.0 and clang-format 18.1.3, CI pins 2.20.0 and clang-format-20.

### DIFF
--- a/src/pipeline/registry.c
+++ b/src/pipeline/registry.c
@@ -925,6 +925,22 @@ static const char *qualified_suffix_match(const qn_array_t *arr, const char *cal
     return match;
 }
 
+/* C#'s built-in type aliases (int, string, bool, ...) are lower-case
+ * KEYWORDS naming a type, not an ordinary lower-case value root; closed list,
+ * no caller-language read, matching the guard below's own design. */
+static bool receiver_root_is_type_keyword(const char *root, size_t len) {
+    static const char *const keywords[] = {
+        "bool",   "byte",  "char",  "decimal", "double", "dynamic", "float",  "int",  "long",
+        "object", "sbyte", "short", "string",  "uint",   "ulong",   "ushort", "void",
+    };
+    for (size_t i = 0; i < sizeof(keywords) / sizeof(keywords[0]); i++) {
+        if (strlen(keywords[i]) == len && strncmp(root, keywords[i], len) == 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
 /* A dotted callee whose FIRST segment starts upper-case names a type — URLSession,
  * Calendar, JSONEncoder. That receiver chain is evidence the bare-name scorers
  * throw away, and throwing it away binds Foundation's URLSession.shared.data to
@@ -933,7 +949,8 @@ static const char *qualified_suffix_match(const qn_array_t *arr, const char *cal
  * appears somewhere in the chain. Calendar.utcGregorian.startOfDayUTC resolving
  * to AuthDTOs.Calendar.startOfDayUTC passes, because Calendar is in the chain.
  *
- * Only an upper-case first segment is guarded. A lower-case root names a value
+ * Only an upper-case first segment is guarded, plus the closed set of
+ * lower-case type keywords above. Any other lower-case root names a value
  * (vm.load, http.Get, os.path.join) whose declared type the chain does not
  * show, so the chain proves nothing there and the call passes through
  * unchanged. A callee with no separator passes through as well.
@@ -960,7 +977,14 @@ static bool receiver_chain_admits(const char *callee_name, const char *candidate
         return true; /* bare name — no receiver chain to judge */
     }
     if (dotted[0] < 'A' || dotted[0] > 'Z') {
-        return true; /* lower-case root names a value, not a type */
+        const char *first_dot = strchr(dotted, '.');
+        size_t root_len = (size_t)((first_dot ? first_dot : last_dot) - dotted);
+        if (!receiver_root_is_type_keyword(dotted, root_len)) {
+            return true; /* lower-case root names a value, not a type */
+        }
+        /* A reserved type keyword IS a type in receiver position
+         * (int.TryParse), so it falls through to the same chain-consistency
+         * check an upper-case root gets. */
     }
     /* A name written in capitals with underscores is a constant holding a
      * value, not a type: ISO_4217_URL.lower is a string's own method. JSON and
--- a/tests/test_pipeline.c
+++ b/tests/test_pipeline.c
@@ -9814,6 +9814,34 @@ TEST(registry_receiver_chain_ignores_lowercase_root_issue1893) {
     PASS();
 }
 
+/* Issue #2121: unlike vm.load's lower-case value root above, C#'s built-in
+ * type keywords (int, string, bool, ...) DO name a type and must get the
+ * same chain-consistency check an upper-case root gets. */
+TEST(registry_receiver_chain_refuses_lowercase_type_keyword_issue2121) {
+    cbm_registry_t *reg = cbm_registry_new();
+    cbm_registry_add(reg, "TryParse", "tmp.Decoys.TsidNode.TryParse", "Method");
+
+    cbm_resolution_t r = cbm_registry_resolve(reg, "int.TryParse", "tmp.Demo", NULL, NULL, 0);
+    ASSERT_NULL(r.qualified_name);
+
+    cbm_registry_free(reg);
+    PASS();
+}
+
+/* The true positive the new keyword check must not eat: the project really
+ * does define a type literally named "int" (or similar) and the chain shows
+ * it, so the match must still go through. */
+TEST(registry_receiver_chain_keeps_type_keyword_when_chain_matches_issue2121) {
+    cbm_registry_t *reg = cbm_registry_new();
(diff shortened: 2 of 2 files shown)


AmirF194:
Both red legs here (`test-windows CLANG64` timing race, and the downstream `shard-completeness` failure that depends on it) are outside this diff, neither touches `registry.c` or `test_pipeline.c`. Been open a week with no look yet, let me know if anything's blocking review.

DeusData:
Thank you for the C# keyword fix and negative controls. The current registry's lower-case receiver handling supports the reported gap. Review should preserve the C# keyword cases without admitting unrelated receiver names in other languages. We still need to attribute the Windows failure from its evidence; an unchanged Windows-specific file list alone would not prove it unrelated. The contribution is queued for that focused review.

AmirF194:
Read the job log rather than the file list this time. The windows-latest CLANG64 leg fails inside tests/test_parallel_harness_contract.sh's own scheduler self-test, step 0i: "FAIL: Windows timeout race refused without a surviving descendant to refuse over, the fixture no longer exercises the race" (job 103034332390). That's a timing fixture for the test harness's own process cleanup, nothing in registry.c or test_pipeline.c. shard-completeness and ci-ok both fail downstream of that missing shard manifest (job 103051948933 shows no manifest uploaded for windows-latest-CLANG64-0).

DeusData (changes requested):
Thank you for this. The C# bug is real and your fix for it is right: in `int.TryParse`, `string.Join` or `object.ReferenceEquals` the lower-case root is a type, and treating a reserved type keyword the way an upper-case root is treated fixes all of them cleanly. I verified it on the merge result with today's `main` (`e783f73d`): clean merge, even though both sides edited `tests/test_pipeline.c`; `pipeline` + `registry` 357 passed, including both `issue2121` tests; memory-core linter unchanged.

One change before it merges, and it is a rule of the codebase rather than a nit. Your comment says the keyword list is deliberately language-agnostic ("closed list, no caller-language read"). Our standing rule for call resolution is the opposite: suppressors stay per-language and are gated on the file's language, so a guard that fixes one language can never trade away recall in another. Every existing guard in `registry.c` follows it for exactly that reason. Here the words are reserved type keywords only in C#. In Python, JavaScript or Go, `object`, `string`, `char`, `float` and `dynamic` are ordinary identifiers, so a lower-case variable named `object` calling a method would now fall through to the chain-consistency check instead of passing as a value root, and that can drop a real CALLS edge in a language this PR never meant to touch.

What I would ask for:

1. Apply the type-keyword branch only when the calling file is C#. The registry already has the shape for this: the per-language gate is decided at the call sites and passed into the helper as `enabled`, never read inside it — see the comment above the Python / JS-TS guard. Those call sites are in `pass_calls.c` and `pass_parallel.c`, and they must enumerate the identical language set, or the sequential and parallel resolvers diverge. Please keep that lockstep.
2. Keep your two C# tests, and add the negative control that pins the other side: a Python (or Go) file in which a lower-case variable named `object` (or `string`) calls a method, asserting that the CALLS edge still resolves. That test is what will stop the next widening of the list from quietly costing another language.

If another language turns out to need the same treatment later, it can join the gate on its own, with its own test. Thank you for the careful true-positive test you already wrote; it is the reason this is a small ask rather than a redesign.

PR #2181, opened 2026-09-11
fix(discover): honor an enclosing repo's .gitignore for a git-less subfolder

### LINKED ISSUE #510: .gitignore (non repo root) gaps and overrides

AmirF194:
`discover_impl` only ever loaded `.gitignore` from `repo_path` itself (fixed for the first half of #510 by 3f754cf) and, when `repo_path` carries its own `.git`, from that repo's `info/exclude`. When `repo_path` is a subfolder with no `.git` of its own, the enclosing repository's root `.gitignore` was never consulted, so files the enclosing repo ignores (secrets, build output) get indexed and become searchable.

Added `resolve_enclosing_git_root()`: when `resolve_git_common_dir(repo_path, ...)` fails, walk up parent directories the same way `git` does from a subfolder, and load the first ancestor's `.gitignore` (plus its `info/exclude`, since `is_git_repo` becomes true once an ancestor is found). It's merged ahead of `repo_path`'s own `.gitignore` so the more specific file still wins on conflict via `cbm_gitignore_merge`, matching git's shallow-to-deep precedence.

Added three tests: enclosing root `.gitignore` honored, enclosing `info/exclude` honored, and the indexed subfolder's own `.gitignore` still overriding the enclosing one on conflict. All three go red without the fix. Ran `scripts/test.sh` (every suite, ASan+UBSan) on this branch and it's fully green. `clang-format` is clean on both files; `cppcheck`/`clang-tidy` flag a couple of pre-existing things in `compat_regex.c`/`cli.c`/`mcp.c` that this diff doesn't touch, same result on plain main.

Fixes #510

### DIFF
--- a/src/discover/discover.c
+++ b/src/discover/discover.c
@@ -1168,6 +1168,37 @@ static bool resolve_git_common_dir(const char *repo_path, char *common_dir, size
     return true;
 }
 
+/* When repo_path itself carries no .git (resolve_git_common_dir already
+ * returned false for it), walk upward looking for an enclosing repository,
+ * exactly as `git` itself would when run from a subfolder. Bounded by the
+ * filesystem/drive root: dir is truncated at each iteration, so the loop
+ * cannot run more times than repo_path is characters long. Returns true and
+ * fills ancestor_root (for the enclosing repo's own .gitignore) plus
+ * common_dir (for its info/exclude + config, via the same resolution
+ * resolve_git_common_dir already applies to an ordinary repo root). Fixes
+ * the remaining half of issue #510: only the indexed directory's own
+ * .gitignore was ever consulted, never an enclosing repo's. */
+static bool resolve_enclosing_git_root(const char *repo_path, char *ancestor_root, size_t ar_sz,
+                                       char *common_dir, size_t cd_sz) {
+    char dir[CBM_SZ_4K];
+    snprintf(dir, sizeof(dir), "%s", repo_path);
+    cbm_normalize_path_sep(dir);
+
+    for (;;) {
+        char *slash = strrchr(dir, '/');
+        /* No separator left, or only the root separator (POSIX "/") or a
+         * bare drive prefix (Windows "C:/"): nothing above this to check. */
+        if (!slash || slash == dir || (slash > dir && *(slash - 1) == ':')) {
+            return false;
+        }
+        *slash = '\0';
+        if (resolve_git_common_dir(dir, common_dir, cd_sz)) {
+            snprintf(ancestor_root, ar_sz, "%s", dir);
+            return true;
+        }
+    }
+}
+
 int cbm_discover(const char *repo_path, const cbm_discover_opts_t *opts, cbm_file_info_t **out,
                  int *count) {
     return cbm_discover_ex(repo_path, opts, out, count, NULL, NULL);
@@ -1215,11 +1246,13 @@ static cbm_discover_status_t discover_impl(const char *repo_path, const cbm_disc
 
     /* Load gitignore sources for ordinary repos AND linked worktrees.
      * Sources merged in order (later patterns win on conflict):
-     *   1. <repo>/.gitignore     — committed exclusions
-     *   2. <common>/info/exclude — per-clone exclusions, not committed
+     *   1. <enclosing>/.gitignore: enclosing repo's root exclusions, only when
+     *      repo_path itself has no .git of its own (see below)
+     *   2. <repo>/.gitignore: committed exclusions
+     *   3. <common>/info/exclude: per-clone exclusions, not committed
      * <common> is the git common dir, resolved via resolve_git_common_dir() so a
      * worktree (where .git is a gitlink file) reads the shared info/exclude/config
-     * just like a normal checkout. Both are folded into a single matcher so all
+     * just like a normal checkout. All are folded into a single matcher so all
      * downstream call paths remain unchanged. Fixes issue #489: OOM on repos whose
      * worktrees are excluded only via .git/info/exclude (e.g. Sandcastle). */
     cbm_gitignore_t *gitignore = NULL;
@@ -1231,11 +1264,34 @@ static cbm_discover_status_t discover_impl(const char *repo_path, const cbm_disc
     char git_common_dir[CBM_SZ_4K];
     bool is_git_repo = resolve_git_common_dir(repo_path, git_common_dir, sizeof(git_common_dir));
     bool has_git_config = false;
+    /* repo_path itself is not a git repo root: walk upward for an enclosing one,
+     * exactly as `git` does when run from a subfolder. Its root .gitignore is
+     * loaded FIRST below (least specific), so the indexed directory's own
+     * .gitignore and info/exclude (both more specific) still override it on
+     * conflict. Fixes the remaining half of issue #510: only the indexed
+     * directory's own .gitignore was ever consulted, never an enclosing repo's. */
+    char enclosing_root[CBM_SZ_4K];
+    if (!is_git_repo &&
+        resolve_enclosing_git_root(repo_path, enclosing_root, sizeof(enclosing_root),
+                                   git_common_dir, sizeof(git_common_dir))) {
+        is_git_repo = true;
+        char enclosing_gi_path[CBM_SZ_4K];
+        path_join(enclosing_gi_path, sizeof(enclosing_gi_path), enclosing_root, ".gitignore");
+        gitignore = cbm_gitignore_load(enclosing_gi_path);
+    }
     /* Always honour the .gitignore at the indexed-directory root, even when the
      * directory is not a git repo root (e.g. indexing a sub-package directly).
      * Fixes issue #510: a root .gitignore was silently ignored without .git/. */
     snprintf(gi_path, sizeof(gi_path), "%s/.gitignore", repo_path);
-    gitignore = cbm_gitignore_load(gi_path);
+    cbm_gitignore_t *local_gitignore = cbm_gitignore_load(gi_path);
+    if (local_gitignore) {
+        if (!gitignore) {
+            gitignore = local_gitignore;
+        } else {
+            (void)cbm_gitignore_merge(gitignore, local_gitignore);
+            cbm_gitignore_free(local_gitignore);
+        }
+    }
     if (is_git_repo) {
         path_join(gi_path, sizeof(gi_path), git_common_dir, "config");
         has_git_config = wide_stat(gi_path, &gi_stat) == 0 && S_ISREG(gi_stat.st_mode);
--- a/tests/test_discover.c
+++ b/tests/test_discover.c
@@ -1355,6 +1355,93 @@ TEST(discover_worktree_committed_gitignore) {
     PASS();
 }
 
+/* ── Enclosing-repo .gitignore tests (issue #510, second half) ──── */
+
+/* repo_path itself has no .git (indexing a git-less subfolder of a larger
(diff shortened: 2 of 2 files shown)


AmirF194:
The one red leg here (`test / test-windows-guards`, a concurrent-daemon-spawn stress test) is outside this diff, we only touch src/discover/discover.c and tests/test_discover.c. Been open a week with no look yet.

DeusData (changes requested):
Thank you for this, and apologies that it sat eleven days without a maintainer word — that is on us, not you. The diagnosis is right (#510's second half is exactly "the enclosing repo's rules are never consulted from a git-less subfolder"), the parent walk is sound and bounded (it stops at `/`, at a bare drive root and at a separator-less string; I traced UNC and `//?/` spellings and they only add a stat probe), the symmetry with `resolve_git_common_dir` is the right shape, and the three tests are RED for the right reason: with the production hunk reverted on today's `main` (`8b57fe2a`) exactly `discover_enclosing_repo_gitignore_issue510`, `discover_enclosing_repo_info_exclude` and `discover_enclosing_repo_gitignore_local_overrides` fail (`count == 2, expected 1`), and re-applied it is 124/124 on `discover`, 309/309 on `gitignore` + `pipeline`, memory-core linter unchanged, merge with `main` clean.

One thing blocks it, and it is a correctness problem rather than polish: **anchored patterns from the enclosing repo are re-anchored to the indexed subfolder.** The ancestor's patterns are merged into the same matcher whose `rel_path` is relative to `repo_path` (`src/discover/discover.c`, the `cbm_gitignore_merge` after `resolve_enclosing_git_root`), and a rooted pattern is matched against that path directly (`src/discover/gitignore.c`, `p->rooted ? glob_match_bounded(p->pattern, rel_path) : …`). Git evaluates every `.gitignore` relative to the directory that contains it. Two probes on the merge result, each checked against `git check-ignore` run from `pkg/`:

- root `.gitignore` = `/secret.py` → this branch **skips** `pkg/secret.py`; git does **not** ignore it. A root line like `/build`, `/dist` or `/src` (very common) now hides the same-named entry inside every indexed subfolder — silent index loss, which is the one thing discovery must never do.
- root `info/exclude` = `pkg/scratch/` → this branch **indexes** `pkg/scratch/tmp.py`; git ignores it. So the "secrets and build output get indexed" case from #510 persists for exactly the anchored form people write for a subdirectory.

Your three tests all use unanchored patterns (`secret.py`, `*.log`, `scratch/`), which match at any depth, so the suite cannot see either direction.

What I would ask for:

1. Keep the enclosing repo's matcher(s) separate instead of merging them into `repo_path`'s, and evaluate each against the path relative to the directory its file lives in — i.e. `<repo_path relative to that directory>/<rel_path>`. The `gitignore_link_t` chain in `discover.c` already encodes "deepest file with an opinion wins" for directories *below* the walk root; this is the mirror image (a base offset applied to `rel_path`) rather than a plain merge.
2. Load the `.gitignore` of every directory from the enclosing root down to `repo_path`, shallow to deep, not only the root's — git consults all of them.
3. Canonicalize `repo_path` (`cbm_canonical_path` in `src/foundation/compat_fs.h`) before the lexical walk: as written, `cbm index pkg` and `cbm index /abs/path/pkg` discover different things, because the walk is purely lexical on the caller's string.
4. Tests for both anchored directions, mirroring the two probes: root `/secret.py` must **not** hide `pkg/secret.py`, root `pkg/scratch/` **must** hide `pkg/scratch/tmp.py`, plus one intermediate-directory `.gitignore` case.
5. Small and optional: on a failed merge the code currently drops the more specific local patterns and keeps the ancestor's (the inverse of the documented policy on the `info/exclude` merge just below), and setting `is_git_repo = true` newly enables the user's global `core.excludesfile` for git-less subfolders — correct per git, but worth a line in the PR body since it is a visible behaviour change.

The Windows guard failure on your run is ours (a cold-start race in the daemon endpoint, fixed in #2275), not anything in this diff; I will update the branch once that lands so you get a clean run. If you would rather I take the base-relative matching from here, say so and I will, with your commits kept as the foundation — but the design is yours and I would be glad to see you finish it.

AmirF194:
You're right, and the two probes make it obvious once I see them: root `/secret.py` should never hide `pkg/secret.py`, and the merge collapses that distinction. Good catch on the intermediate directory case too, my tests never exercised a `.gitignore` sitting between the enclosing root and repo_path.

Take it from here. The matcher-offset redesign and the chain walk down from the enclosing root need the kind of familiarity with `gitignore_link_t` you already have, and guessing at that architecture is more likely to hand you a second bug than fix the first. Keep the commit as the foundation.

Thanks for tracing the Windows failure to #2275, that saves me chasing a dead end.

DeusData:
Thank you for handing this over so clearly, and for keeping the door open instead of guessing at the chain code. I took you at your word: your commit stays exactly as you wrote it, as the foundation, and I added one commit on top of it on this branch, with you credited as co-author.

What the added commit does:

- **Base-relative matching.** The enclosing repository's patterns are no longer merged into the indexed folder's matcher. The existing `gitignore_link_t` chain already carried a `prefix` for matchers *below* the walk root; an ancestor is the mirror image, so each link now also carries a `base` — the walk root's path relative to the matcher's own directory — which is prepended rather than stripped. A rooted `/secret.py` in the enclosing root now means the root's `secret.py`, as git means it.
- **Every intermediate directory.** The `.gitignore` of each directory from the enclosing root down to the indexed folder is loaded, shallow to deep, chained so the deepest opinion wins, negations included. The enclosing repository's `info/exclude` sits with the root ancestor's matcher, where git anchors it.
- **Canonical path first.** `repo_path` goes through `cbm_canonical_path` before the parent walk, so indexing `pkg` and `/abs/path/pkg` discover the same files.
- Nothing is merged into the indexed folder's own matcher any more, which also removes the old failed-merge path that kept the ancestor's patterns and dropped the local ones.

Tests: four new cases, each pinned first to real `git check-ignore` run inside the subfolder on an identical fixture — a rooted pattern that must not be re-anchored, an `info/exclude` sub-path that must apply, an intermediate `.gitignore`, and a deeper negation that must win. On your commit alone they fail (`discover` 120 passed, 4 failed); with the change they pass, and reverting just the production hunks brings back exactly the same four failures. Your three tests pass throughout. Memory-core linter unchanged. I also checked both commits merged with today's `main`, since your branch predates several merges: `discover` 128, `gitignore` 22, `pipeline` 287 and `git_context` 3 passed, with a clean merge.

The real-input check is the part I like best: this repository's own `scripts/` folder is a git-less subfolder under a root `.gitignore` with rooted `/memlab-*`. With your commit alone, indexing `scripts/` silently dropped `memlab-drive.py` and `memlab-report.py`, which git keeps; with the change both are back. So the defect was losing real files here, not only in the synthetic probes.

Honest limits: this was run on macOS only; the Windows legs of CI on this push are the first Windows run of the change, and the drive-root and UNC stop conditions are still your original loop, untouched. Symlinked `repo_path` values now resolve to the physical path, where git uses the logical one; no test covers that yet. If the Windows guard job goes red, that is the daemon endpoint cold-start race on our side (#2275), not this diff.

PR #2191, opened 2026-09-12
fix(csharp): index C# 14 extension declaration members

### LINKED ISSUE #2071: C# 14 extension declarations can cause file to be omitted from index
### Version

codebase-memory-mcp 0.10.8

### Platform

Windows (x64)

### Install channel

### Binary variant

standard

Minimal example:

using System;
using System.Linq;

{
    {
        public int SumPositive() =>
    }
}

**Expected result**

### Reproduction

2. Add a file containing:
using System.Linq;

{
    {
        public int SumPositive() =>
    }
}

**Actual Result:**

### Logs

```text

```

```text

```

_No response_

### Confirmations

lorenzozanee:
## Summary

C# 14 extension declarations are represented by older tree-sitter grammars as a constructor named `extension`. The indexer skipped the constructor body, so members such as `SumPositive` were absent from the graph. This change recovers method declarations from that body through the existing method-definition path.

Fixes #2071

## Testing

- `bash scripts/test.sh --suites grammar_regression,grammar_labels,cs_lsp`
- `git diff --check`

## Reviewers

- @DeusData
- @pcristin
- @bmcnaboe

### DIFF
--- a/internal/cbm/extract_defs.c
+++ b/internal/cbm/extract_defs.c
@@ -5014,6 +5014,28 @@ static void push_method_def(CBMExtractCtx *ctx, TSNode child, TSNode class_node,
     cbm_defs_push(&ctx->result->defs, a, def);
 }
 
+/* Older C# grammars parse an extension declaration as a constructor whose body
+ * contains local functions. Recover those members without descending into
+ * ordinary C# method bodies. */
+static void extract_csharp_extension_members(CBMExtractCtx *ctx, TSNode node, TSNode class_node,
+                                             const char *class_qn, const CBMLangSpec *spec) {
+    const char *kind = ts_node_type(node);
+    if (strcmp(kind, "local_function_statement") == 0 ||
+        strcmp(kind, "method_declaration") == 0) {
+        TSNode name_node = resolve_method_name(node, ctx->language);
+        if (!ts_node_is_null(name_node)) {
+            push_method_def(ctx, node, class_node, class_qn, spec, name_node);
+        }
+        return;
+    }
+
+    uint32_t count = ts_node_named_child_count(node);
+    for (uint32_t i = 0; i < count; i++) {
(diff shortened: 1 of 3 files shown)


pcristin:
I believe you tagged me by accident. Only @DeusData can review PRs.

internal/cbm/extract_defs.c
```diff
@@ -5014,6 +5014,28 @@ static void push_method_def(CBMExtractCtx *ctx, TSNode child, TSNode class_node,
...
+        }
+        return;
+    }
+
+    uint32_t count = ts_node_named_child_count(node);
+    for (uint32_t i = 0; i < count; i++) {
+        extract_csharp_extension_members(ctx, ts_node_named_child(node, i), class_node, class_qn,
+                                         spec);
+    }
+}
```
AmirF194, line 5037: `extract_csharp_extension_members` (extract_defs.c:5020-5037) only recognizes `local_function_statement` and `method_declaration`. A property member of an extension block, e.g. `public int P => x;`, cannot parse as a statement inside the constructor's block, so the grammar emits an ERROR node there instead and the property is silently dropped: no Method, no Field, nothing.

Checked against tree-sitter-c-sharp 0.23.5 in a clean container (closest I could get to the vendored grammar without building this repo; the method-only case parses exactly as this PR's own comment describes, which is why I'd expect this to generalize):

```
$ node -e 'const P=require("tree-sitter"),C=require("tree-sitter-c-sharp");
const p=new P(); p.setLanguage(C);
const t=p.parse("public static class E { extension(int x) { public int P => x; } }");
console.log("hasError:",t.rootNode.hasError);'
hasError: true
```

Same result whether the property is the extension's only member or sits beside a method. Worth at least detecting (`ts_node_has_error`) and skipping loudly rather than dropping the member with no signal.

Issue #1696, opened 2026-08-17
Audit smoke, soak, profiling, and benchmark harnesses for account daemon runtime isolation


astandrik:
Follow-up to #1691 and #1695.

PR #1695 isolates the focused security, watchdog, worker, and hook entry points that reproduced the account daemon shutdown. The longer-running smoke, soak, profiling, and benchmark harnesses have different lifecycle and measurement contracts, so they should be audited separately instead of expanding that fix.

## Scope

- `scripts/smoke-local.sh` and `scripts/smoke-test.sh`
- `scripts/soak-legs.sh` and `scripts/soak-test.sh`
- `scripts/memlab.sh`, `scripts/memlab-drive.py`, and `scripts/memlab-report.py`
- `scripts/benchmark-index.sh`, `scripts/benchmark-search-graph.sh`, and `scripts/clone-bench-repos.sh`

## Invariants to check

- An inherited `CBM_RUNTIME_DIR` or `CBM_CACHE_DIR` is never reused for a product process started by a harness.
- Concurrent harness runs receive different private roots.
- Unix roots are owner-only. Windows runs use the repository's protected temp-root helper rather than duplicating ACL logic.
- Install, update, uninstall, restart, and daemon cleanup stay inside the harness runtime.
- A sentinel client in account runtime A remains connected with the same daemon PID and generation, and runtime A records no `activation_shutdown`.
- A harness stops its own child processes before cleaning its runtime. If its private daemon remains active, cleanup leaves the root for diagnosis.
- Runtime isolation does not alter benchmark coverage or invalidate timing comparisons. Any performance claim includes comparable setup, cache, and total wall-clock measurements.

## Expected output

For each entry point, record whether it already satisfies these invariants. Add a focused fail-before reproduction for every confirmed gap, then submit scoped fixes separately from #1695.

DeusData:
Thank you for separating this audit from the focused fix in #1695. The boundary is correct: smoke, soak, profiling, and benchmark harnesses have lifecycle and measurement contracts that should not be folded into the security-script repair.

Please keep this issue as the audit ledger. Each confirmed isolation failure should have its own deterministic reproduction and focused fix, especially where changing setup could affect benchmark comparability. That gives every harness change an attributable reason and prevents a broad test-infrastructure sweep from becoming one unreviewable PR.

astandrik:
## Audit ledger — account daemon runtime isolation

[...]

PR #2196, opened 2026-09-13
fix(test): isolate soak harness daemon runtime

astandrik:
## What does this PR do?

Part of #1696 (audit ledger), follow-up to #1691/#1695.

`scripts/soak-test.sh` claimed to "isolate daemon coordination from interactive CBM
sessions" through a private `CBM_CACHE_DIR`, but only `CBM_RUNTIME_DIR` moves the
daemon rendezvous. The soak must be the only client of the daemon it measures (it
asserts that a session crash and the final shutdown each stop the daemon), so on a
shared rendezvous it either stopped the operator's daemon or was refused with a
cache-root conflict.

- `scripts/soak-test.sh` sources `scripts/test-runtime.sh` and calls
  `cbm_test_runtime_init`; off native Windows the helper's owner-only cache is the
  soak cache, so the daemon log path is unchanged in shape. `soak_cleanup` now runs
  `cbm_test_runtime_cleanup "$BINARY"` instead of `rm -rf` on the cache: the private
  daemon is stopped first and the root is kept when it will not stop. The
  native-Windows stamped root and its `icacls` handling are untouched
  (`tests/test_soak_daemon_recovery_contract.sh` pins them). Two FAIL messages no
  longer say "shared daemon".
- `tests/test_soak_runtime_isolation_contract.sh`: environment-probe fixture,
  same assertions as the smoke contract, `RESULTS_DIR` kept out of the checkout.
  Fails on `main` with `FAIL: soak-test exposed the caller CBM_RUNTIME_DIR to a
  product process`; passes with this change. Wired as Step 0d2 in `scripts/test.sh`.

`scripts/soak-legs.sh` starts no product process and needs no change.

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) — shell-only change; the new
      contract plus `test_soak_daemon_recovery_contract`, `test_venue_parity_contract`
      and `test_runtime_isolation_contract` were run; the full C suite was not run
      on this host
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) — no C changes


DeusData:
Merged — thank you. This is the fix for a problem that has been quietly taxing other people's pull requests for weeks, and you went after the actual cause rather than the symptom.

The diagnosis in your comment is exactly right, and it matches what we had already verified from the other end: `cbm_daemon_rendezvous_key()` hashes a compile-time constant, so the daemon cohort is machine-global, and **`CBM_RUNTIME_DIR` is the only hook that moves it**. A private `CBM_CACHE_DIR` looks like isolation and is not — which is why the soak either stopped the operator's daemon or was refused with a cache-root conflict.

Two things I want to call out specifically:

**Your contract test can fail.** I checked rather than trusting the green — reverting `scripts/soak-test.sh` to its previous state while leaving your test in place:

```
FAIL: soak-test exposed the caller CBM_RUNTIME_DIR to a product process
(exit 1)
```

Driving the harness with an environment-probe fixture and asserting on what the *product process* actually receives is the right design. It tests the contract, not the implementation, so it will keep working when the harness changes again.

**The cleanup is careful.** Routing through `cbm_test_runtime_cleanup` means you inherit its refusals — unexpected basename, symlinked root, replaced runtime path — and it is idempotent, so the early `trap ... EXIT` and the later `soak_cleanup` both firing is harmless. Leaving the root behind when the daemon will not stop is the right call for diagnosis, and ordering the `SOAK_WIN_ROOT` removal after the daemon check because the probing binary lives inside it is a detail that would have bitten later.

One note for your planning: the rest of this series (#2197, #2198, #2199) lands on `scripts/test.sh` too, so expect to rebase the remaining ones as each lands. I will keep them moving.

PR #2197, opened 2026-09-13
fix(test): isolate memlab harness daemon runtime

astandrik:
## What does this PR do?

Part of #1696 (audit ledger), follow-up to #1691/#1695.

`scripts/memlab.sh` gave the profiled run a private `CBM_CACHE_DIR` only, so the
process joined the operator's account daemon (refused with a cache-root conflict
when one was live; otherwise left as the account daemon while `cleanup()` removed
its cache underneath it).

- `scripts/memlab.sh` sources `scripts/test-runtime.sh`, calls
  `cbm_test_runtime_init` and runs `cbm_test_runtime_cleanup "$BINARY"` before
  removing its work directory (the native-Windows copy of the binary lives there).
  Off native Windows the helper's cache is the cache; the log fold-in reads from it.
- `tests/test_memlab_runtime_isolation_contract.sh`: memlab launches the product
  through its Python driver, which passes the environment through unchanged
  (`memlab-drive.py` uses no `env=`; the test pins that), so the contract shims
  `python3` and records the environment at the driver boundary. This keeps it
  deterministic on Windows CI, where the native Python cannot exec a shell fixture.
  Fails on `main` with `FAIL: memlab exposed the caller CBM_RUNTIME_DIR to the
  product`; passes with this change. Wired as Step 0g2 in `scripts/test.sh`.

Observation, not changed here: `scripts/memlab.sh` is tracked without an executable
bit although its usage line invokes it directly; the contract runs it through `bash`.

## Checklist

- [ ] Tests pass locally (`make -f Makefile.cbm test`) — shell-only change; the new
      contract plus `test_venue_parity_contract` and `test_runtime_isolation_contract`
      were run; the full C suite was not run on this host
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`) — no C changes


DeusData:
Merged — thank you. Same root cause as #2196 and the same careful treatment, with one addition I want to credit specifically.

**Recording the environment at the driver boundary is the right call, and the reasoning in your header is why.** `memlab-drive.py` passes no `env=` to `Popen`, so the driver hands the product exactly what it inherited — which makes the driver boundary an honest observation point rather than an approximation. Standing a `python3` shim in for the driver means the contract holds on Windows too, whose native Python cannot exec a shell fixture, and no product process needs to start at all. That is a test that will survive changes to the thing it tests.

**Your contract test can fail.** I verified it rather than trusting the green — reverting `scripts/memlab.sh` while leaving your test in place:

```
FAIL: memlab exposed the caller CBM_RUNTIME_DIR to the product
(exit 1)
```

The failure mode you describe in the header is worth restating because it is nastier than it first appears: with only a private `CBM_CACHE_DIR`, the profiled process joined the operator's account daemon — and was then either refused with a cache-root conflict, or left running as the account daemon *with its cache deleted underneath it*. The second is the kind of thing that surfaces as an unrelated failure hours later.

`scripts/test.sh` merged cleanly with #2196's addition, so no rebase was needed here. #2198 and #2199 touch the same file, and I will keep them moving.

PR #2201, opened 2026-09-13
fix(mcp): give the over-budget error a retry value, and say the peak is not the requirement

DeusData:
Found while benching v0.10.9 for release. Small, but it fixes a message that actively misleads.

## The problem

v0.10.9 introduces **Decision A**: an over-budget index fails whole with a named error instead of thrashing. The error said:

> Raise **CBM_MEM_BUDGET_MB**, lower CBM_WORKERS, or exclude large subtrees.

It names the knob but not a value, so the natural next step is to read `peak_rss_mb` and retry just above it.

**That retry fails every time, by construction.** The abort fires *when* RSS crosses the budget, so `peak_rss_mb` is pinned just above the budget — it is where indexing was **stopped**, not what the repo needs.

Measured on the Linux kernel (94,521 files / 43.1M LOC) during the release bench:

| | |
|---|---|
| default budget | 24,576 MB |
| aborted, reported peak | 25,622 MB |
| **actually needed to complete** | **31.75 GB** |

The reported peak understates the requirement by **24%**. A caller retrying at peak+10% burns another 30s to fail again with a near-identical message.

## The change

- adds `suggested_budget_mb` — **1.5× the current budget**, which would have cleared the kernel case (it needed 1.32×)
- the hint now states that `peak_rss_mb` is where indexing **STOPPED**, not a target

Computed as `(3*b+1)/2`, not `b + b/2`: integer division makes the latter degenerate to `b` at `b == 1`, so the "suggestion" would repeat the budget that just failed. **The extended test caught exactly that** on the first run.

## Deliberately not changed

The abort, the backpressure ladder, and worker cleanup are untouched.

Worth recording: worker cleanup was *suspected* broken (a 24.6 GB process appeared to outlive an abort). A clean test disproved it — process count 2 before the abort, 2 after, both idle at 0% CPU. The stray process belonged to an overlapping bench leg of my own. **No fix was needed, so none was made.**

## Verification

`tests/test_mcp.c` pins both halves: a concrete value strictly greater than the budget, and the hint stating the peak is where indexing stopped. Checked against the **real pre-fix output captured during the bench**, which contains neither string — so the test is genuinely red without this change.

- mcp suite: **317 passed, 0 failed, 4 skipped**
- `make -f Makefile.cbm lint-ci`: clean

## Release context

Benching v0.10.9 against the v0.10.0 baseline on the same host and corpus found **no performance regression**: cold index 289.6s → **226.7s (−21.7%)**, nodes stable (+0.02%). Edges fell 4.2%, traced by an A/B edge-set diff to removal of **false** cross-arch/cross-tooling calls (e.g. 23,745 edges pointing at a Python gdb helper; 10,323 bound to SPARC `writel` of which only 15 callers were SPARC) — a graph-quality improvement, not a loss.


DeusData:
Carried by #2202: the same change is on feat/memory-core as commit b5f10c22 (identical patch-id to 11679720), so it lands together with the memory-core increment. Closing this one to keep a single PR for the change.

PR #2203, opened 2026-09-13
fix(cli): handle trusted Linux system aliases during install activation

### LINKED ISSUE #2175: install subcommand always fails: 'failed to stage install candidate: activation transaction I/O failed' (v0.10.8, Linux btrfs)
## Repro
```
$ codebase-memory-mcp install -n
error: failed to stage install candidate: activation transaction I/O failed
codebase-memory-mcp install 0.10.8
```
Fails identically with `--dir=$HOME/.local/bin`, with a clean `TMPDIR`, and via `install.sh --skip-config` (which shells out to the same subcommand after checksum-verified download). Exit code is 0 despite the error.

## Environment
- Binary: `codebase-memory-mcp 0.10.8` (linux-amd64 portable tarball, checksum verified)
- OS: Fedora Atomic / bootc (kernel 7.1.5), x86_64
- Target fs: btrfs, 77G free, dir verified writable (`touch`+rm succeeds)
- TMPDIR unset (also tried clean TMPDIR — same failure)

## Notes
- The binary itself runs fine (indexing, `cli search_code`, MCP stdio serve all work). Only the `install` staging/activation path fails, before any file is written.
- Suspect a btrfs-incompatible staging op (reflink / renameat2 flag / dir-fsync assumption). No strace on this host so I could not confirm the exact syscall.
- Workaround used: manual `cp` of the portable binary to `~/.local/bin`.

DavidHLP:
## Summary

Fix Linux install staging when the target path is reached through a root-owned system alias.

`activation_posix_walk_path()` now resolves only the explicit Linux aliases `/tmp`, `/var`, and `/home` when they are root-owned symlinks. Arbitrary user-owned symlinks remain rejected, and macOS keeps its existing `/tmp` and `/var` handling.

## Why

On Linux layouts such as Atomic-style `/home -> /var/home`, and in the reproducer `/tmp -> /tmp-real`, the secure path walk rejected a system-managed alias before the install candidate could be staged. The command then reported:

```text
failed to stage install candidate: activation transaction I/O failure
```

## Validation

- `activation_transaction,cli`: 338 passed in the Ubuntu 24.04 Docker test image.
- Patched build and `git diff --check`: passed.
- Root-owned `/tmp` alias: both missing-target and existing-target cases succeeded.
- Root-owned `/home -> /var/home`: succeeded.
- Arbitrary user-owned symlink: rejected; target was not created.

## Review scope

- The exact Fedora bootc/SELinux image was not tested.
- The issue's reported exit code `0` was not reproduced; the controlled rejected-symlink case returned `1`.
- Exit-code handling is unchanged in this PR.

Fixes #2175


DavidHLP:
CI follow-up for the MSan result:

- The only failed job is `test / test-msan`.
- The failure is in the existing `daemon_bootstrap` suite, at `tests/test_daemon_bootstrap.c:1116` (`daemon_bootstrap_fails_fast_when_daemon_dies_at_publication`).
- The log shows an `ENOSPC` while creating `/tmp/cbm-bootstrap-failure-record-.../cbm-daemon-0/...sock.pending.tmp`, followed by the 30-second client-accept timeout. It does not report an MSan uninitialized-read finding.
- The PR diff still contains only `src/cli/activation_transaction.c`; the activation fix is not involved in this test failure.

The contributor token cannot rerun this job (`Must have admin rights to Repository`), so please rerun `test-msan` when convenient. Other completed checks are green; the remaining checks are still pending.

DavidHLP:
## Verification update

I rebuilt the patched `v0.10.8` binary in the existing Ubuntu 24.04 Docker test image and reran the focused validation.

Results:

- `activation_transaction,cli`: `338 passed`
- root-owned `/tmp` alias with a missing target parent: `rc=0`, binary installed
- existing target directory: `rc=0`
- Atomic-style root-owned `/home -> /var/home`: `rc=0`, binary installed
- arbitrary user-owned symlink: `rc=1`, target was not created

The reported direct-command exit-code discrepancy is not reproducible with this binary: the rejected user-owned-symlink case returns `1`, and the current `main` propagates the install handler result. I am not adding an unrelated exit-code change without a reproducer.

The PR workflow still has one underlying failing job, `test / test-msan`; the other PR checks are green. Its log reports a daemon-bootstrap test timeout without a MemorySanitizer report, and the contributor token cannot rerun that job because GitHub requires repository admin rights. Please rerun that job when convenient.

DavidHLP:
Follow-up for commit `581f0c7`:

The alias list is now Linux-only for `/home`; macOS retains its previous `/tmp` and `/var` policy.

I rebuilt this new head in Docker and reran the issue matrix:

- product build: `build_rc=0`
- root-owned `/tmp` alias + missing target parent: `rc=0`, target created
- existing target directory: `rc=0`, target created
- arbitrary user-owned symlink: `rc=1`, target not created
- root-owned Atomic-style `/home -> /var/home`: `rc=0`, target created
- final matrix status: `0`

The Linux fix remains minimal and the existing `O_NOFOLLOW` rejection boundary is preserved.

src/cli/activation_transaction.c
```diff
@@ -901,8 +909,13 @@ static char *activation_posix_walk_path(const char *directory) {
         }
         struct stat alias_status;
         char resolved[4096];
-        if (lstat(alias, &alias_status) != 0 || !S_ISLNK(alias_status.st_mode) ||
-            alias_status.st_uid != 0 || !realpath(alias, resolved)) {
+        if (lstat(alias, &alias_status) != 0) {
```
AmirF194, line 912: I extracted `activation_posix_walk_path` from this PR's head (581f0c7) verbatim into a standalone harness and ran it in a clean `python:3.12-slim` container as root.

At activation_transaction.c:912, when an alias path (`/tmp`, `/var`, `/home`) does not exist on the filesystem at all, `lstat(alias, &alias_status) != 0` returns NULL immediately. That aborts the whole install for ANY target path that merely starts with that prefix, even an ordinary non-symlink path that has nothing to do with the trusted-alias case this PR is adding. By contrast, when the alias exists but is not a symlink (line 915), the code correctly `continue`s past it and eventually falls through to `activation_string_copy(directory)`, passing the path through unchanged.

I confirmed this by removing `/home` from a container and calling the function with a target under `/home/...`:

```
$ docker run --rm -v /tmp/pr2203-repro:/repro:ro python:3.12-slim bash -c '
apt-get update -qq >/dev/null 2>&1 && apt-get install -qq -y gcc >/dev/null 2>&1
gcc -o /tmp/test_walk /repro/test_walk.c
rmdir /home
/tmp/test_walk "/home/newuser/.local/share/codebase-memory"
'
  [walk] lstat(/home) FAILED (errno=2 No such file or directory) -> returning NULL (hard fail)
RESULT: NULL (activation would ABORT / fail)
```

With `/home` present as an ordinary directory (the common case on most systems), the same call correctly returns the path unchanged.

This repo's own uninstall/agent support already targets minimal container and hardened-image environments (the PR's own repro is a stripped Ubuntu Noble image), so an image missing one of `/tmp`, `/var` or `/home` outright is a realistic shape here, not just a symlink-aliased one.

Fix: treat an `lstat` failure on the alias the same way as `!S_ISLNK` (line 915), using `continue` to move to the next alias instead of `return NULL`. That way a missing alias falls through to the unchanged-path return at the end of the function instead of aborting activation.

DavidHLP: Implemented in commit 3b7f2559.

When a trusted system alias is absent, activation_posix_walk_path now continues instead of returning NULL. A missing alias is therefore treated as an ordinary path prefix, while the existing O_NOFOLLOW traversal still validates the actual path. Root-owned alias resolution and rejection of arbitrary user-owned symlinks remain unchanged.

Validation in the Ubuntu 24.04 Docker environment (UID 1000, isolated HOME):
- product build: passed
- activation_transaction,cli regression suites: 338 passed
- final test return code: 0

Thanks for catching this edge case.

AmirF194: Ran commit 3b7f2559 against the same repro (removed `/home`, called `activation_posix_walk_path` on a path under it): it no longer aborts, and the path is returned unresolved as expected.

I also checked the two cases the one-line change touches directly, same container, same function body:
- `/home` present as a symlink owned by a non-root uid: still rejected (`NULL`), unchanged.
- `/home` present as a root-owned symlink: still resolved via `realpath`, unchanged.

Looks correct.

JohnGalt1717:
Please get this merged. Right now you can't install codebase-memory on Ubuntu 26.04 to name just one because of this.

DeusData:
Thank you for the Linux alias fix and the platform matrix. Current activation handling still makes this a real compatibility gap, but the trust-boundary details need more review. We need more time to review before giving an integration decision; this is not a request for speculative rebases or broader permission exemptions.

DavidHLP:
Thanks for the clarification. Understood. I’ve kept the change limited to the explicit Linux aliases and left the existing rejection of arbitrary user-owned symlinks unchanged. I won’t make speculative rebases or broaden permission exemptions. I also tightened the PR description to state the current scope, validation, and remaining limits clearly. I’ll wait for the trust-boundary review.

DeusData:
**Approved.** The trust-boundary question that was holding this has been settled, and I want to record the reasoning rather than just flipping a switch.

The gate here is narrow in three independent ways, and all three have to hold before anything is resolved:

1. **A fixed allowlist**, not a policy — `/tmp`, `/var`, `/home` on Linux. Nothing else is considered, so this cannot grow into general symlink following by accident.
2. **The entry must actually be a symlink** (`S_ISLNK`), so a real directory at those paths is untouched.
3. **The symlink must be root-owned** (`st_uid != 0` → `continue`). Arbitrary user symlinks still fail the `O_NOFOLLOW` walk below, which is the property that made the original macOS version safe.

That last point is what makes the extension sound: an attacker who can plant a root-owned symlink at `/home` already has root, so this grants no capability that did not already exist. Meanwhile the legitimate case it unblocks is real and common — immutable and Atomic Linux layouts expose `/home` as a root-owned alias for `/var/home`, which is precisely why installation fails there today.

I also appreciate that you **narrowed rather than broadened** when asked. Your 2026-09-20 note — keeping the change to explicit Linux aliases, leaving the rejection of arbitrary user-owned symlinks untouched, and declining to make speculative rebases — is the right instinct on a security-adjacent path. Splitting the `lstat` guard into separate early-`continue` branches is also a genuine readability gain at no semantic cost.

@JohnGalt1717 — thank you for the nudge; you were right that this is a real compatibility gap rather than an edge case.

**On timing:** `main` currently carries one failing test unrelated to this PR (a coverage test written against an older response format; fix in #2248). I am landing that first so this does not merge onto a red base, then merging here. Nothing further is needed from you.

DeusData:
Merged — `main` is green again as of 36f77084, so this lands on a clean base as promised. Thank you for the fix, the platform matrix, and for narrowing the scope when asked rather than broadening it. @JohnGalt1717, Ubuntu 26.04 installs should work from the next release.

PR #2205, opened 2026-09-14
fix(daemon): preserve disabled UI on manual start

### LINKED ISSUE #2114: `daemon start` silently overrides persisted `ui_enabled=false` and re-enables the graph UI listener
## Summary

On 0.10.8 (macOS arm64, official binary at `~/.local/bin`), an explicit `codebase-memory-mcp daemon start` re-enables the graph UI even when `ui_enabled` was just persisted to `false`, and writes `ui_enabled=true` back into the config. The auto-spawn path (daemon started on demand by the MCP front-end) respects the persisted setting, so the two start paths disagree.

## Repro

```
$ codebase-memory-mcp daemon stop
$ codebase-memory-mcp config set ui_enabled false
  ui_enabled = false
  (restart the daemon for this to take effect)
$ codebase-memory-mcp config list | grep ui_enabled
  ui_enabled                = false
$ codebase-memory-mcp daemon start
  daemon: started (permanent, pid ...)
  ui: warming asynchronously on configured port 9749
$ lsof -nP -i :9749
  codebase-…  LISTEN  # UI is back
$ codebase-memory-mcp config list | grep ui_enabled
  ui_enabled                = true      # persisted setting was overwritten
```

## Expected

`daemon start` honors the persisted `ui_enabled=false` — the `config set` output itself says "restart the daemon for this to take effect", implying a restart applies the setting rather than discarding it.

## Auto-spawn path behaves correctly

With `ui_enabled=false` persisted and no daemon running, connecting an MCP client (stdio front-end) auto-spawns the daemon and the UI stays off:

```
$ codebase-memory-mcp config list | grep ui_enabled
  ui_enabled                = false
$ < MCP initialize over stdio >
  # daemon auto-spawns, server initializes (~6s cold)
$ lsof -nP -i :9749
  # no listener
$ codebase-memory-mcp config list | grep ui_enabled
  ui_enabled                = false      # unchanged
```

## Impact

Users who deliberately disabled the UI (headless hosts, minimal footprint, port policy) silently regain a listening UI on 127.0.0.1:9749 after any manual `daemon start`. Since the warm daemon is the recommended way to avoid per-command startup cost, the only way to keep the UI off today is to never run `daemon start` and accept the cold auto-spawn.

## Environment

- codebase-memory-mcp 0.10.8 (build 996bad5fe6fb89c0c50363d87f03dc78f6e536aeea746c06679a294d14eca435)
- macOS 15, arm64
- MCP host: opencode v2 (beta-19296), `type: local` stdio server
- Rendezvous dir `/tmp/cbm-daemon-501` was freshly recreated (unrelated stale-state wedge per #1760 was cleared before testing)

Related: #1760 (rendezvous state), #1828 (misleading 30s timeout reporting).

vincentkoc:
## What does this PR do?

A cold `daemon start` now preserves a saved `ui_enabled=false` setting. Previously it unconditionally requested UI configuration, rewrote the preference to `true`, and started the HTTP listener.

An explicit cold `--port` or `--open` request keeps its existing UI-enabling behavior. Repeated bare starts report the disabled UI without incorrectly saying a browser request failed, and the configuration hint points to `config set ui_enabled true`. The configuration docs describe these semantics.

Fixes https://github.com/DeusData/codebase-memory-mcp/issues/2114

## Validation

- The new disabled-UI regression failed against the unchanged base and passed with this fix. All seven real-process readiness cases passed, covering cold and warm starts, saved enabled/disabled settings, explicit UI requests, delayed readiness, and foreign-listener rejection.
- `scripts/test.sh BUILD_DIR=build/full-tests` passed on macOS arm64: 8,022 tests passed, 10 platform skips, plus all prescribed process guards and security/ordering checks.
- `git diff --check`, formatting, cppcheck, and the NOLINT suppression check passed.
- **Full lint is not green:** `scripts/lint.sh` exited 2 with 6,972 repo-wide clang-tidy diagnostics. An exact-path VFS comparison of unchanged and candidate `src/main.c` produced 125 diagnostics each with the same diagnostic categories; none point to changed lines. The existing `main_run_daemon_ctl` cognitive-complexity diagnostic increases from 82 to 89, already above the threshold of 25. Tools: Apple clang-format 21, clang-tidy 22.1.8, cppcheck 2.21.0.

## Checklist

- [x] Tests pass locally.
- [ ] Full lint passes (the existing clang-tidy debt is described above).
- [x] New behavior is covered by a test, reproduced before the fix.


DeusData:
Thank you for the cold-start configuration fix. Current startup has separate loaded-configuration and explicit CLI override paths; review should prove that configured values survive a cold start while explicit flags retain precedence. Please keep the cold/warm and explicit-override cases as the acceptance boundary. This is queued for focused review.

Issue #2162, opened 2026-09-10
Intermittent cold-start race: version_cohort.claimed_unheld -> daemon.ipc.listen_failed stage=reservation_validation kills the stdio MCP server for the whole host session


Tomauskasz:
**Version:** `codebase-memory-mcp 0.10.8`
**Platform:** macOS (Apple Silicon)
**Install channel:** GitHub release archive / install.sh / install.ps1
**Binary variant:** ui

---

## What happened, and what did you expect?

Some **cold** daemon starts fail during listener handoff. The client then prints
`CBM daemon could not start within 30000 ms` and **exits**, closing stdio. MCP hosts do
not respawn a crashed stdio server, so that host session loses the graph permanently —
Codex reports every subsequent tool call as `Transport closed`.

Rate: with a deliberate cold-start probe (below) I hit it **2 times in ~30 cold starts**.
My `cbm-daemon.log` also carries **1 occurrence I did not cause**, against 83 successful
daemon starts in the same file. I cannot attribute any specific host-session outage to
this race — I am reporting a reproducible fault, not a measured production rate.

Two things make the blast radius much larger than the raw rate suggests:

1. The daemon exits the instant its last client disconnects
   (`daemon.runtime_stopping reason=last_committed_client_disconnected`) and tears down
   the whole rendezvous directory. So **every** new host session is a fresh cold start
   and a fresh roll of the dice. This is the same teardown behaviour reported in #2058.
2. `hook-augment` runs as a short-lived client on nearly every turn in both Claude Code
   and Codex, with a 2000 ms deadline. Cold start costs ~2 s, so those clients routinely
   abandon a start mid-flight — `hook-augment-timeouts.log` logged **1183
   `deadline_exceeded ms=2000` entries in a single day** on this machine. That is a lot
   of start/teardown churn feeding the race. (cf. #1335, #2058, #1767)

Expected: a cold start that loses a reservation race should retry or wait for the
incumbent, not hard-fail. And a hard failure should not be reported as a 30 s timeout —
the call returns in **0.8–2.1 s**.

## Reproduction

No project or indexed code required — this is purely daemon lifecycle. Run from any
directory; a public repo works fine as cwd.

```python
# repro.py — cold-start the stdio client repeatedly and count handshake failures.
# Each iteration attaches, handshakes, detaches, and waits long enough for the
# daemon to reach last_committed_client_disconnected and exit, so the next
# iteration is a genuine cold start.
[...]

PR #2207, opened 2026-09-14
fix(daemon): stop the lifetime-lock probe from dropping a held lock

AmirF194:
While chasing #2162 I found a real, separate bug in `posix_lifetime_lock_probe` (src/daemon/ipc.c): when the in-process registry says this process already holds the lifetime lock, it still opens a throwaway fd on the lock file just to close it again and return 1. `fcntl(2)` record locks are scoped to (process, inode), not (fd, inode), so that close silently drops the real lock too, even though the reservation's own fd stays open.

Two probes back to back are enough to lose it. I wrote a test that acquires the reservation, probes it twice, then forks a genuinely separate process that tries to take the same lock directly. On main it succeeds, the lock is gone. With the fix it fails as expected.

Fix: check the in-process claim before opening anything and return early. That branch only ever needed to trust the registry, not touch the file.

`scripts/test.sh --suites daemon_ipc`: the new test fails on main (`child_result == 1`), 50 passed / 1 skipped with the fix.

Not a fix for #2162 itself, I ran into this while reading that code, it's a separate bug.


DeusData:
Thank you for isolating the claim-lock descriptor path and providing a separate-process regression. This affects coordination between processes, so we need more time for the lifecycle review before making an integration decision. It should not be treated as proof that every cold-start race in #2162 is resolved.

PR #2097, opened 2026-09-08
test(windows): guard clean MCP stdio startup

lorenzozanee:
## What does this PR do?

Adds a native Windows MCP stdio regression guard for #654. The guard isolates daemon runtime cleanup and rejects the reported path-not-found diagnostics after a successful initialize/tools/list exchange.

## Checklist

- [ ] Every commit is signed off (`git commit -s`) — required, CI rejects
      unsigned commits ([DCO](../DCO), see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [ ] Tests pass locally (`make -f Makefile.cbm test`)
- [ ] Lint passes (`make -f Makefile.cbm lint-ci`)
- [ ] New behavior is covered by a test (reproduce-first for bug fixes)

Fixes #654

PR #2213, opened 2026-09-15
test(windows): guard clean MCP stdio startup

### LINKED ISSUE #654: [Windows] MCP stdio startup emits localized path-not-found message on stderr
## Summary

On the official Windows amd64 v0.8.1 binary, starting the MCP stdio server emits a localized OS error on `stderr` even when the cache is empty and the process completes successfully:

```text
level=info msg=mem.init budget_mb=... total_ram_mb=...
El sistema no puede encontrar la ruta especificada.
```

The second line means `The system cannot find the path specified`.

## Environment

- Windows 11, Spanish locale
- `codebase-memory-mcp 0.8.1`
- MCP `serverInfo.version`: `0.10.0`
- MCP protocol: `2025-03-26`
- Official `codebase-memory-mcp-windows-amd64.zip`
- ZIP checksum and GitHub attestation verified

## Minimal reproduction

Run the binary in stdio mode with a fresh empty `CBM_CACHE_DIR`, send only:

1. `initialize`
2. `notifications/initialized`
3. `tools/list`

Capture stdout and stderr separately. The behavior reproduces both from `C:\` and from a normal repository directory.

Equivalent Python harness:

```python
import json, os, subprocess, tempfile

cache = tempfile.mkdtemp()
env = os.environ.copy()
env["CBM_CACHE_DIR"] = cache
messages = [
    {"jsonrpc":"2.0","id":1,"method":"initialize","params":{
        "protocolVersion":"2025-03-26",
        "capabilities":{},
        "clientInfo":{"name":"minimal-repro","version":"1"}
    }},
    {"jsonrpc":"2.0","method":"notifications/initialized","params":{}},
    {"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}},
]
result = subprocess.run(
    [r"C:\path\to\codebase-memory-mcp.exe"],
    input="".join(json.dumps(m) + "\n" for m in messages),
    text=True,
    capture_output=True,
    env=env,
)
print(result.returncode)  # 0
print(result.stderr)
```

## Actual behavior

- Exit code is `0`.
- stdout contains valid JSON-RPC responses.
- stderr always includes the localized path-not-found line.
- The message also appears with no indexed projects, so it is not caused by a stale repository root.

## Expected behavior

No OS path error should be written during a successful minimal MCP startup, or the missing optional path should be handled silently / logged with identifying context.

## Impact

The server remains functional, but MCP clients may surface the line as an error. It also makes ad-hoc PowerShell users more likely to merge stderr with stdout, which corrupts captured JSON.

This appears separate from #423, which concerns CLI JSON/path argument handling.

lorenzozanee:
## What does this PR do?

Adds a native Windows MCP stdio regression guard for #654. The guard isolates daemon runtime cleanup and rejects the reported path-not-found diagnostics after a successful initialize/tools/list exchange.

## Checklist

Fixes #654

Supersedes #2097（原PR分支已删，按原提交重建）


DeusData:
Thank you for building a native Windows MCP stdio guard for #654 — that platform is where our hardest bugs hide and guards there are worth a lot.

**Before you spend any time on this one: I think it may be a double-submit, and I would rather ask than close it on you.**

This PR and **#2217** are the same guard for the same issue, opened thirteen seconds apart. #2217 looks like the fuller version:

| | #2213 (this one) | #2217 |
|---|---|---|
| `tests/windows/test_mcp_stdio.py` | +57 | +59 |
| `tests/windows/mcp_stdio.py` | — | +1 |
| `scripts/test-windows.ps1` | +1 | +1 |
| description | rejects path-not-found diagnostics | also captures stderr deterministically, handles the **Spanish** diagnostic text |

If that is right, #2217 is the one to keep and this can be closed — but it is your work and your call, so tell me which you want and I will do the rest. If they are actually different and I have misread, say so and I will review both.

Whichever survives needs the same two things:

**The `dco` check is red.** Your commits carry no `Signed-off-by` trailer — a headline with an empty body, so there is nothing to parse. (Not the blank-line trap, where a sign-off falls outside the trailer block and git's parser ignores it while `grep` still sees it; here there is no trailer.)

```
git push --force-with-lease
```

**`test / test-msan` is our problem, not yours.** That lane fails on its Docker image build whenever `apt.llvm.org` is flaky, and it reddens PRs that have nothing to do with it. Please ignore it — I track it separately and it will not block this.

#2241 and #2242 both merged today after review. Good work on those.

PR #2105, opened 2026-09-08
fix(mcp): reject stale AST coordinates after source changes

lorenzozanee:
Fixes #1750. Source edits can invalidate AST line ranges, so this update uses the canonical file metadata on Windows before exposing stored coordinates, rejects stale snippet reads, and preserves raw search matches without stale AST attribution.

PR #2216, opened 2026-09-15
fix(mcp): reject stale AST coordinates after source changes

### LINKED ISSUE #1750: Bug: Stale AST line coordinates in get_code_snippet / search_code after file modification

lorenzozanee:
Fixes #1750. Source edits can invalidate AST line ranges, so this update uses the canonical file metadata on Windows before exposing stored coordinates, rejects stale snippet reads, and preserves raw search matches without stale AST attribution.

Supersedes #2105（原PR分支已删，按原提交重建）


DeusData:
Thank you for carrying the stale-range work into this focused replacement. Current source still uses indexed source ranges, with freshness checks forming a separate contract. Review needs to retain the supplied boundary cases and check the Windows metadata and untracked-store behavior explicitly. The queue is busy, so it may take a little time to complete that review.

DeusData:
Thank you — rejecting stale AST coordinates after the source has changed is exactly the kind of correctness fix this project cares about. Returning a snippet keyed to bytes that have moved is worse than returning nothing.

Two blockers, both small.

**1. The memory-core linter (this is what `lint / lint` is failing on, not clang-format):**

```
  src/mcp/mcp.c: grew by 1 (814 -> 815); latest sites: 18147:free
```

One raw `free` at `src/mcp/mcp.c:18147`. Every allocation here routes through `src/foundation/mem_core.h` rather than the ~800 raw sites this codebase used to have, because the memory accounting, the budget and the spill machinery all depend on seeing each one — a raw `free` is invisible to them. The linter ratchets, so existing sites are tolerated and new ones are not.

Use `cbm_free` (and `cbm_alloc` / `cbm_calloc` / `cbm_realloc` / `cbm_mem_strdup` for the rest of the family). Keep the pairing consistent: memory that came from `cbm_alloc` must go to `cbm_free`, and memory from a raw `malloc` elsewhere must still go to raw `free`. If line 18147 releases something allocated by code you did not touch, tell me rather than guessing and we will sort it out.

`make -f Makefile.cbm lint-ci` reproduces it locally.

**2. The `dco` check is red.** Your commits have no `Signed-off-by` trailer — the message is a headline with an empty body, so there is nothing for the check to parse. (Not the blank-line trap, where the sign-off sits outside the trailer block and git's parser skips it while `grep` still sees it — here there is no trailer at all.)

```
git push --force-with-lease
```

#2241 and #2242 merged cleanly today with correct sign-offs, so the habit is already there — these branches just predate it.

PR #2217, opened 2026-09-15
test(windows): guard clean MCP stdio startup

lorenzozanee:
## What does this PR do?

Adds a native Windows MCP stdio regression guard for #654. The guard runs a successful initialize/tools/list exchange with an isolated cache and runtime directory, captures stderr deterministically, and rejects the reported English or Spanish path-not-found diagnostics. The guard is registered in `scripts/test-windows.ps1`, and the stdio helper now joins its stderr reader on close so the captured output is complete.

## Checklist

Refs #654

Supersedes #2182（原PR分支已删，按原提交重建）


DeusData:
Thank you for preserving the startup diagnostics. This overlaps #2213, but this version also carries sibling-runtime and stderr-join coverage. Could you identify one canonical PR retaining those cases and ensuring temporary runtimes are cleaned up on failures? We will keep both open until that handoff is clear; no duplicate closure or extra routine rebase is being requested.

DeusData:
Thank you for this — a native Windows MCP stdio guard for #654 is valuable, and the details here are the ones that matter: an isolated cache and runtime directory, deterministic stderr capture, and rejecting the diagnostic in both English and Spanish rather than assuming an English-locale runner.

**First, a housekeeping question.** This and **#2213** are the same guard, opened thirteen seconds apart. This one looks like the fuller version — +59 vs +57 in `tests/windows/test_mcp_stdio.py`, plus `tests/windows/mcp_stdio.py`, plus the Spanish diagnostic handling. I have asked on #2213 which you would like to keep; my guess is this one, but it is your work and your call, so I will not close either until you say.

**One real blocker: the `dco` check.** Your commits carry no `Signed-off-by` trailer — the message is a headline with an empty body, so there is nothing for the check to parse. (Not the blank-line trap, where a sign-off sits outside the trailer block and git's parser ignores it while `grep` still finds it; here there is genuinely no trailer.)

```
git push --force-with-lease
```

**Two of the red checks are ours, not yours:**

- `test / test-msan` fails on its Docker image build whenever `apt.llvm.org` is flaky.
- `test / test-windows-guards` is known to redden PRs it has nothing to do with — confirmed on three unrelated ones.

Please ignore both. I track them separately and neither will block this.

On the Spanish diagnostic specifically: matching localized runner output is the sort of thing that usually gets discovered the hard way, months later, on someone else's machine. Thank you for getting there first.

PR #2218, opened 2026-09-15
fix(csharp): index C# 14 extension declaration members

### LINKED ISSUE #2071: C# 14 extension declarations can cause file to be omitted from index

lorenzozanee:
## Summary

C# 14 extension declarations are represented by older tree-sitter grammars as a constructor named `extension`. The indexer skipped the constructor body, so members such as `SumPositive` were absent from the graph. This change recovers method declarations from that body through the existing method-definition path.

Fixes #2071

## Testing

- `bash scripts/test.sh --suites grammar_regression,grammar_labels,cs_lsp`
- `git diff --check`

## Reviewers

- @DeusData
- @pcristin
- @bmcnaboe

Supersedes #2191（原PR分支已删，按原提交重建）


DeusData:
Thank you for the C# 14 recovery work and for adding a grammar regression fixture. I have linked this with #2071 and the overlapping proposal #2231. Both changes touch extract_class_methods in internal/cbm/extract_defs.c, but their recovery logic and regression coverage differ, so we need to compare them rather than treat either contribution as automatically superseded.

The review queue is quite full, and this may take a little time. We appreciate your contribution and will come back with a review decision; there is no need to repeatedly refresh the branch solely to chase this acknowledgement. This is triage, not an approval or merge commitment.

DeusData:
Thank you — indexing C# 14 extension declaration members is genuinely useful, and covering it in both `test_grammar_labels.c` and `test_grammar_regression.c` is the right instinct.

Two blockers, both mechanical.

**1. clang-format**, one line:

```
internal/cbm/extract_defs.c:5023:57: error: code should be clang-formatted [-Wclang-format-violations]
```

CI lints with **clang-format 20**. One caution from our own history: a standalone `clang-format-20` binary and a Homebrew LLVM build disagree about some wrapped lines, and the mismatch produces false whole-file drift. If your local formatter reports changes far beyond line 5023, that is the disagreement rather than your code — reformat just that line and leave the rest alone. `make -f Makefile.cbm lint-ci` is the canonical check.

**2. The `dco` check is red.** Your commits carry no `Signed-off-by` trailer — the message is a headline with an empty body, so there is nothing for the check to parse. (Not the blank-line trap, where the sign-off falls outside the trailer block and git's parser ignores it while `grep` still finds it; here there is no trailer.)

```
git push --force-with-lease
```

One thing I want to flag for your own planning rather than as a review point: **#2231 by @yasmoradi also recovers C# 14 extension block members**, and the two may well overlap. I have not reviewed theirs yet and I am not asking you to change anything — but if the approaches differ I would rather we settle that deliberately than have one of you rebase onto the other's merge by surprise. I will look at both together.

#2241 and #2242 merged cleanly today with correct sign-offs, so the habit is already there — these branches just predate it.

DeusData:
Following up on the overlap I flagged earlier — I have now read both PRs properly, and it is closer than I thought.

**#2231 by @yasmoradi fixes the same bug and defines a function with the same name.** You have both added `extract_csharp_extension_members` to `internal/cbm/extract_defs.c`. They cannot both land as they stand. Our fault for not catching it sooner; you both found a real gap independently.

**Where your version is ahead:**

- You **recurse** rather than scanning only the block's direct children, and you also accept `method_declaration`, so a member nested more deeply still gets recovered.
- You ship a **grammar-label fixture** (`csharp_extension_declaration` with an expected census). That is the part I value most — it pins the behaviour so the next refactor of that table cannot silently undo it, and it follows the convention the surrounding tests already use.

**Where #2231 is ahead, and it is a correctness point rather than a style one:** both of you key off a `constructor_declaration` whose name is `extension`, but #2231 also checks the **enclosing type's** name. A constructor carries its own type's name, so a class genuinely called `extension` has a genuine constructor called `extension` — legal C#. Without that second check, such a constructor is silently reinterpreted as an extension block and its members are lifted onto the wrong type. Rare, but a graph that is quietly wrong is worse than one that misses something.

**What I would like, if you are both willing:** one PR carrying the type-name guard *and* your recursive traversal *and* your fixture. I will not pick a winner by filing order, and I am not closing either of you — you have both done good work on the same bug. If you and @yasmoradi would rather one of you carries it with the other credited as co-author, say which and I will back it.

Still outstanding here independently of any of that: the `dco` sign-off, and the clang-format violation at `internal/cbm/extract_defs.c:5023:57`.

Worth naming the real root cause: the vendored tree-sitter-c-sharp grammar has no `extension` production, which is why this parses as a constructor at all. A grammar refresh would make both PRs obsolete — until then, recovery like yours is the right approach.

PR #2219, opened 2026-09-15
test(fqn): guard against Unicode project-name collisions

### LINKED ISSUE #571: Project name strips non-ASCII (CJK) characters from path, resulting in truncated/unrecognizable names

lorenzozanee:
## What does this PR do?

Adds regression coverage for issue #571 by asserting that distinct non-ASCII project path segments produce distinct validator-safe project names. The existing project-name implementation encodes non-ASCII bytes as hexadecimal, so CJK path segments remain part of the project identity instead of collapsing into the same ASCII slug.

## Checklist

Fixes #571

Supersedes #2192（原PR分支已删，按原提交重建）


DeusData:
Thank you for adding the Unicode collision regression. Current FQN encoding already contains the underlying Unicode handling, so this is valuable protection of an existing fix rather than a newly delivered runtime correction. We should keep #571's requested reporter verification separate and not close it solely because this test lands.

DeusData:
Thank you — a guard against Unicode project-name collisions in FQNs is a sharp thing to think of, and 14 lines in one test file makes it easy to say yes to.

**One blocker: the `dco` check.**

Your commit carries no `Signed-off-by` trailer — the message is a headline with an empty body, so there is nothing for the check to parse. (Not the blank-line trap, where the sign-off sits outside the trailer block and git's parser skips it while `grep` still finds it; here there is no trailer at all.)

```
git commit --amend -s
git push --force-with-lease
```

Everything else is green (31 success, 1 skipped). Sign it off and I will merge it.

One question, genuinely curious rather than a review demand: does this guard fail today, or does it pin behaviour that is already correct? Both are worth having and I will take either — but they are different things, and if it is a pinning test I would like to say so in the merge so the next person reading it knows it was never a bug report.

#2241 and #2242 merged cleanly today with correct sign-offs — the habit is already there, these branches just predate it.

PR #2223, opened 2026-09-16
fix: describe graph reference gaps without declaring live code dead

### LINKED ISSUE #1642: Dead-code detection flags live code: constructors, dunders, and callback-by-reference all report "0 callers"
## Summary

The graph UI's dead-code classification (`status: "dead"` = "0 callers") flags a large number of **live** functions/methods as dead. The detector only counts direct `CALLS`/`USAGE`/`CALL_REFERENCE` edges and cannot see three common dispatch patterns, so it reports false positives for:

1. **Constructors** — `ClassName(...)` instantiation never creates a `CALLS` edge to `__init__`.
2. **Dunder / framework-invoked methods** — `__str__`, `__call__`, `_generate`/`_agenerate` (called by a framework's `invoke`/`ainvoke`), `_llm_type` (property access).
3. **Callback-by-reference** — functions passed by name as arguments (`asyncio.to_thread(fn, ...)`, `context.configure(include_object=fn)`, `DecisionPoint(apply=fn)`).

## Repro

On a Python repo, index it and open the graph UI. In my case (a Python monorepo), **50 of 50** "dead" nodes were false positives. Concrete example:

`ModeAwareModel.__init__` is flagged `dead`, but it is instantiated in three places:

```python
# ai/langgraph/services/model.py
def get_model() -> BaseChatModel:
    return ModeAwareModel(ChatOpenAI(...), role="supervisor")

def get_worker_model(session_id=None) -> BaseChatModel:
    return ModeAwareModel(ChatOpenAI(...), role="worker")

# inside bind_tools()
return ModeAwareModel(cast(BaseChatModel, bound), self._role)
```

The graph has no `CALLS` edge from `ModeAwareModel(...)` to `ModeAwareModel.__init__`, so the constructor is reported as having zero callers.

Other flagged-but-live examples:
- `_apply_updates` — passed as `apply=_apply_updates` to a `DecisionPoint`.
- `_include_object` / `_compare_type` — passed as `include_object=` / `compare_type=` to Alembic's `context.configure`.
- `_run_file_search` — passed to `asyncio.to_thread`.
- `__str__`, `__call__`, `_generate`, `_agenerate`, `_llm_type` — invoked by runtime/framework dispatch.

## Root cause

The classification in `src/ui/layout3d.c` (`cbm_layout_compute`):

```c
bool is_fn = label=="Function" || label=="Method";   // only these are candidates
int ic = in_calls[i];
int iu = in_usage[i] + in_call_reference[i];
...
else if (ic==0 && iu==0) status = "dead";
```

`dead` is defined purely as zero incoming `CALLS`/`USAGE`/`CALL_REFERENCE` edges. Static analysis cannot resolve constructor dispatch, dunder/framework invocation, or callback-by-reference, so all three produce false "dead" results.

## Suggested fix

In the dead-code classification, treat these as live (non-dead):

- **`__init__`** — live when its owning class has any instantiation (`Class(...)` call) or is itself referenced/exported.
- **Dunder methods** (`__str__`, `__call__`, `__repr__`, `__eq__`, etc.) — either treat as live, or exclude from dead-code classification entirely.
- **Callback-referenced functions** — a function referenced by name as an argument (e.g. `asyncio.to_thread(fn)`, `configure(include_object=fn)`) should count as having a caller.

At minimum, excluding `__init__` and dunders from the "dead" bucket would eliminate the majority of false positives.

## Environment

- codebase-memory-mcp v0.10.4 (single static binary, `--ui=true`)
- Python repo, ~13.4k graph nodes

mvanhorn:
## What does this PR do?

Update `FilterPanel.tsx` so the section is titled "Reference coverage", the candidate count reads "N with no known inbound references", and the existing filter reads "Show only nodes with no known inbound references". Add a short explanation in that section: "No known inbound references does not prove code is unused. Static analysis may miss constructors, implicit or framework calls, and callbacks." Change the `dead` entry in `STATUS_LEGEND` in `colors.ts` to "No known inbound references", and clarify the nearby status comment while retaining its existing key and color.

The graph UI presents functions and methods with no known incoming CALLS, USAGE, or CALL_REFERENCE edges as definitively dead, even though constructors, runtime/framework dispatch, and callbacks can execute without those edges being extracted. The reporter found 50 false positives in 50 sampled nodes, and another user reproduced the constructor case on v0.10.8.

DeusData asked that the first fix remove false certainty at the presentation boundary and leave extraction improvements to separate graph-quality work. This PR is that first step. It does not repair missing graph edges, so it references the issue rather than closing it.

Refs #1642

## Checklist

- [x] Every commit is signed off (`git commit -s`) - required, CI rejects
      Not run locally. CI currently fails on `test / test-unix (macos-15-intel)` and `ci-ok`; I am looking at that.
      Not run locally. CI `lint / lint` and `lint / lint-mem` both pass.


DeusData:
Thank you for improving this wording. Current UI labels say `Dead code` and `Dead (0 callers)`, although an absent recorded caller is not proof that live code is unused. Please use `Refs #1642` rather than an automatic closing reference: the wording improvement is useful, but it does not establish that the underlying edge-completeness concern is fixed.

mvanhorn:
Switched to Refs #1642 in the description, thanks. You are right that the wording change does not establish the edge-completeness concern is fixed, so it should not auto-close the report.

I also cleaned up the description while I was in there: a stray line from my own working notes had ended up in it, and the checklist was malformed. It now reflects the real state, including that CI is red on test-unix (macos-15-intel) and ci-ok, which I am looking at.

DeusData:
Thank you for this, @mvanhorn — and I owe you a straight explanation, because you did everything right here and I am still going to close it.

**We are relaunching the UI soon, and this change would be redundant the moment that lands.** The current graph view is being replaced as a whole rather than evolved, so anything merged into `graph-ui/` today has a very short life. I would rather tell you that now than merge it, have it disappear with the old surface, and leave you wondering what happened to your work. I should have said so on 19 September instead of asking you for another round; that one is on me.

What I want you to know about the change itself:

- **It is correct, and the point it makes is the right one.** "Dead code" and "Dead (0 callers)" claim something static analysis cannot know. "No known inbound references", plus the one-line note that constructors, framework dispatch and callbacks can be missed, is exactly the honest wording. Keeping the backend status key and colour untouched was the careful way to do it, and the test pins the wording.
- **The requirement carries forward.** Your phrasing — *no known inbound references; this does not prove the code is unused* — goes into the new UI's requirements as written, and #1642 stays open for the underlying edge-completeness work, as you set it up with `Refs`.
- **The red check was never yours.** `test-unix (macos-15-intel)` failed in a C test of the `search_code` tool (`tests/test_mcp.c:9744`), which a change to three files under `graph-ui/` cannot influence. It is logged on our side as ours to attribute. Please do not spend any more time on it.

Thank you for taking a reporter's "50 false positives out of 50" seriously enough to fix the claim at the boundary where users actually read it, and for the patience with the back-and-forth. If you would like to put that same eye on the edge-completeness side of #1642 — the constructor and callback cases that make those nodes look unreferenced in the first place — that work will outlive any UI, and I would be glad to review it.

mvanhorn:
Thanks for the straight answer, and no problem about the timing. Better to hear the UI is being replaced now than to watch the change vanish with the old surface later.

Glad the wording carries forward into the new UI's requirements. That was the part that mattered to me, not the diff.

I'll take a look at the edge-completeness side of #1642. The constructor and framework-dispatch cases are the more interesting half of the problem anyway, and you're right that the work outlives any UI.

Issue #2128, opened 2026-09-09
Dead-code status ignores framework-driven entry points


avigdorsky:
**Severity:** high — the feature is unusable as shipped.

359 nodes get `status: "dead"`; **247 (69%)** are reachable only dynamically:

- ASGI/WSGI middleware `__call__` for classes listed in a settings module
- `__init__` of classes instantiated via a registry or DI
- alembic `upgrade`/`downgrade`, Django management-command `handle`
- `pytest` fixtures, `@receiver` / `@shared_task` decorated functions

**Ask:** a configurable liveness list (dunder methods, decorator patterns,
declarative registrations) — and, more fundamentally, do not derive `dead` from a
call graph whose own measured precision is 52%. Reporting `unknown` when the graph
is unresolved would be honest and far more useful than a false negative.

---
Observed on v0.10.8 (build `996bad5fe6fb`), macOS arm64, Python monorepo (~36k nodes / ~179k edges). Happy to provide a minimal reproduction repo if useful.

bhaveshpatelapple:
Adding one more framework-dispatch pattern to the list, not yet mentioned: **Pydantic `@model_validator`**.

```python
from pydantic import BaseModel, model_validator

class Spec(BaseModel):
    start: str
    end: str

    @model_validator(mode="after")
    def _window_is_ordered(self) -> "Spec":
        if self.end <= self.start:
            raise ValueError("end must be after start")
        return self
```

`_window_is_ordered` is only ever invoked by Pydantic's own validation machinery when a `Spec` is constructed — there's no user-code call site to create a `CALLS` edge from. Flagged `status: "dead"` regardless.

Confirmed on v0.10.8, same project as my comment on #1642. This is the same "declarative registration" bucket you already called out for `@receiver`/`@shared_task` — Pydantic validators (and I'd guess `@field_validator`, `@validator`, dataclass `__post_init__`) belong in the same liveness-exclusion list, or under the `unknown` classification you proposed rather than a hard `dead`.

PR #2234, opened 2026-09-18
fix(ui): preserve framework route entry points from dead-code status

lorenzozanee:
## What does this PR do?

Treat Function/Method nodes with a HANDLES edge to a Route node as framework-driven entry points in graph layout dead-code classification. Add focused positive and negative regression coverage while preserving existing status precedence.

This addresses the verified route-handler representation from #2128; broader framework and configurable-liveness cases remain outside this focused change.

## Checklist

      The focused sanitizer-backed UI suite passes: `make -j4 -f Makefile.cbm test-focused TEST_SUITES=ui` (19/19).
      Not verified locally because `clang-tidy` and `cppcheck` are unavailable.

Refs #2128


DeusData:
Thank you for the targeted UI fix and the positive/negative HANDLES regression cases. I checked current main: the classification in src/ui/layout3d.c uses entry/route properties and incoming call/usage counts, but does not recognize a HANDLES edge to a Route as entry-point evidence. That grounds the false-positive problem this PR addresses.

I have queued this as a high-priority bug fix. The full review still needs to check failure handling and the cost of the additional lookups; this triage pass has not run or approved the change. Our review queue is quite full, so it may take a little time, but we appreciate the focused contribution and will come back with feedback.

DeusData:
Thank you for this — preserving framework route entry points from dead-code status is a real graph-quality improvement, and the test that comes with it is the right shape.

One blocker, and it is a five-line change rather than a rethink.

**The memory-core linter is what is failing, not clang-format.** The `lint / lint` job is red on this:

```
  src/ui/layout3d.c: grew by 3 (66 -> 69); latest sites: 818:free, 819:free, 820:free
```

This repository routes every allocation through one core (`src/foundation/mem_core.h`) rather than the ~800 raw `malloc`/`free` sites it used to have, because the memory accounting, the budget and the spill machinery all depend on seeing every allocation. A raw `free` is invisible to that accounting, so the linter ratchets: existing sites are tolerated, new ones are not.

The fix is to use the core's functions at `src/ui/layout3d.c:818-820`:

| instead of | use |
|---|---|
| `malloc` | `cbm_alloc` |
| `calloc` | `cbm_calloc` |
| `realloc` | `cbm_realloc` |
| `free` | `cbm_free` |
| `strdup` | `cbm_mem_strdup` |

Note the buffers you free there must also have been allocated through the core — mixing a raw `malloc` with `cbm_free` (or the reverse) is worse than either. If those three `free` calls release memory that something else allocated with raw `malloc`, tell me and we will look at it together rather than you guessing.

You can check locally before pushing with `make -f Makefile.cbm lint-ci`, and `--list` on the memory-core linter prints every site if you want the full picture.

Everything else here is green. Push the fix and I will re-review promptly.

PR #2242, opened 2026-09-19
fix(typescript): preserve awaited generic call edges

### LINKED ISSUE #2210: TypeScript: awaited generic calls omit CALLS edges
### Version

codebase-memory-mcp v0.10.8

### Platform

Linux x86_64 (reproduces with both `fast` and `full` indexing modes)

### What happened

TypeScript / TSX awaited calls with explicit generic type arguments do not produce `CALLS` edges. For example, `await parseJsonBody<T>()` is omitted, while the equivalent non-generic `await parseJsonBody()` call is indexed.

This makes inbound call tracing incomplete without any parse error or warning.

### Minimal reproduction

Create these two files in an otherwise empty directory.

`route-utils.ts`

```ts
export async function parseJsonBody<T>() { return {} as T; }
export async function plainLocal() { return await parseJsonBody(); }
export async function genericLocal() { return await parseJsonBody<{ message: string }>(); }
```

`gateway.ts`

```ts
import { parseJsonBody } from './route-utils';

type Payload = { message: string };

export async function plainImported() { return await parseJsonBody(); }
export async function genericImported() { return await parseJsonBody<{ message: string }>(); }
export async function genericPrimitive() { return await parseJsonBody<string>(); }
export async function genericNamed() { return await parseJsonBody<Payload>(); }
```

Index the directory, then call `trace_path` with:

```json
{
  "project": "awaited-generic-repro",
  "function_name": "parseJsonBody",
  "direction": "inbound",
  "depth": 1,
  "format": "json"
}
```

### Actual result

Only these two callers are returned:

- `plainLocal`
- `plainImported`

### Expected result

All six callers should be returned:

- `plainLocal`
- `genericLocal`
- `plainImported`
- `genericImported`
- `genericPrimitive`
- `genericNamed`

The behavior is the same in `.tsx` files. The generic calls are valid TypeScript, and the generic type argument should not change the callee from `parseJsonBody`.

### Notes

This appears to be an extraction issue around the AST shape of `await f<T>()`, rather than a resolver issue: the non-generic calls in the same files resolve correctly. Because the corrected extraction changes graph content while source files remain unchanged, an index semantic-version/cache invalidation may also be needed so existing indexes are rebuilt.

### Confirmations

- [x] I searched existing issues and did not find this exact awaited-generic-call case.
- [x] The reproduction contains only standalone, shareable code.

lorenzozanee:
## What does this PR do?

Normalize the TypeScript AST shape used for awaited generic calls so extraction and LSP resolution preserve the bare callee and emit the expected CALLS edges. Add focused regressions for plain and explicit-generic await calls.

Focused validation passed with `scripts/test.sh --suites extraction,ts_lsp` (653 tests). Full test and lint commands remain unverified locally because the sanitizer run exceeded the local time budget and clang-tidy/cppcheck are unavailable.

Fixes #2210

## Checklist


DeusData:
Merged — thank you, this is a well-built fix and it was a pleasure to review.

Three things you did that made it easy to say yes:

**You found the real shape of the bug.** `await parseJsonBody<string>()` parses with the `await_expression` sitting where the callee is expected, so the generic call was simply never reached. Unwrapping once, at the point the callee is selected, is the right place — not a special case scattered through the resolver.

**You fixed it in both paths.** The textual extractor (`extract_calls.c`) and the TypeScript LSP resolver (`ts_lsp.c`) each select a callee independently, and a fix in only one of them would have produced a graph that disagreed with itself. You patched both, and covered both with their own test.

**Your tests actually fail without your fix.** I verified this rather than trusting the green: reverting `extract_calls.c` and `ts_lsp.c` to their pre-PR state while leaving your tests untouched turns both red —

```
tests/test_extraction.c:4372  count_calls_named(r, "parseJsonBody") == 1, expected 2
tests/test_ts_lsp.c:738       find_resolved(r, "generic", "parseJsonBody") >= 0  -> not found
```

That check matters here: this repository has shipped tests that passed with the fix removed, so a green suite alone is not evidence. Yours is.

One observation, not a request: `unwrap_await_callee` is now defined identically in both translation units. That is ordinary C and I am not going to make you chase it — noting it only so the next person who touches either copy knows about the other.

Fixes #2210.

PR #2244, opened 2026-09-20
fix(cli): support non-ASCII Windows agent paths

### LINKED ISSUE #2145: Windows: `install` reports "Detected agents: (none)" when the profile path contains non-ASCII characters — agent detection still uses narrow stat() (v0.10.8)
### Version

codebase-memory-mcp 0.10.8

### Platform

Windows 11 Pro x64 (10.0.26200)

### Install channel

GitHub release archive / install.ps1

### Binary variant

standard

`install` completes successfully but configures **no agent at all** — it prints `Detected agents: (none)` — even though Claude Code is installed and both `%USERPROFILE%\.claude\` and `%USERPROFILE%\.claude.json` exist.

The cause is the Windows username containing a non-ASCII character. My profile is `C:\Users\İgal` (`İ` = U+0130, LATIN CAPITAL LETTER I WITH DOT ABOVE).

Expected: agents whose config directories exist should be detected and configured regardless of whether the profile path is representable in the process ANSI code page.

This is the **same root cause as #1715** (narrow `stat()` on Windows cannot see paths with characters outside the ANSI code page), but it is a **different code path**. #1715 was fixed for the daemon *session root* in 9d4ac2b9 / #1876; agent *detection* was not covered and is still narrow on `main` (checked at `055fbb7`):

```c
/* src/cli/cli.c — Agent detection */
static bool dir_exists(const char *path) {
    struct stat st;
#ifndef _WIN32
    return lstat(path, &st) == 0 && S_ISDIR(st.st_mode);
#else
    return stat(path, &st) == 0 && S_ISDIR(st.st_mode);   /* <-- narrow */
#endif
}
```

`cbm_detect_agents()` routes every agent through `dir_exists()`, so on such a profile *all* of them silently come back false — not just Claude Code.

### There is currently no escape hatch

`--clients=claude` does not help: `cli_clients_apply_selection()` only *clears* the flags that were not requested and never force-includes one, so it cannot override a failed detection.

```c
for (size_t i = 0; i < CLI_CLIENT_COUNT; i++) {
    if (!wanted[i]) {
        *(bool *)((char *)detected + CLI_CLIENTS[i].offset) = false;
    }
}
```

That matches #1798 ("`install --clients=<token>` accepted but inert"), which is closed but still reads as inert on `main`.

### Reproduction

1. On Windows, use an account whose profile path contains a character outside the active ANSI code page (e.g. `C:\Users\İgal`; CJK reproduces the same way).
2. Confirm `%USERPROFILE%\.claude\` and `%USERPROFILE%\.claude.json` exist.
3. Run `codebase-memory-mcp install -y --force`.

Result:

```
Detected agents: (none)

Added C:/Users/İgal/.local/bin to the current-user PATH
Install complete.
```

No MCP entry, no skill, no agents, no hooks are written.

### Workaround (verified)

Run `install` with `HOME` pointed at the profile's 8.3 short name, which is pure ASCII and resolves to the *same* directories (`cbm` checks `HOME` before `USERPROFILE`):

```powershell
$env:HOME = 'C:\Users\GAL~1'     # 8.3 short name of C:\Users\İgal
codebase-memory-mcp install -y --force --clients=claude
```

That produces the correct, complete result:

```
Detected agents: Claude-Code

Claude Code:
  skills: 1 installed
  agent: C:/Users/GAL~1/.claude/agents/codebase-memory-scout.md
  agent: C:/Users/GAL~1/.claude/agents/codebase-memory.md
  agent: C:/Users/GAL~1/.claude/agents/codebase-memory-auditor.md
  mcp: C:/Users/GAL~1/.claude.json
  hooks: PreToolUse Grep/Glob search augmentation + PostToolUse Read coverage (non-blocking)
  hooks: SessionStart / SubagentStart reminders
```

The 8.3 short name is discoverable programmatically, so it could also serve as an internal fallback:

```powershell
(New-Object -ComObject Scripting.FileSystemObject).GetFolder($env:USERPROFILE).ShortPath
```

### Suggested fix

Make `dir_exists()` (and any sibling existence checks used by detection) go through the wide API on Windows — `cbm_utf8_to_wide()` + `_wstat`/`GetFileAttributesW`. The conversion helper already exists in `src/foundation/compat_fs.c` and is what 9d4ac2b9 used for the session-root path, so this should be a small, consistent follow-up to that fix.

### Note

The failure is silent — `install` exits 0 and reports success. Since a user with a non-ASCII profile gets a completely unconfigured agent from a "successful" install, a warning when detection finds nothing at all might be worthwhile independent of the encoding fix.

lorenzozanee:
## What does PR do?

On Windows, agent detection could report no agents when the profile path contained non-ASCII characters because the lookup used narrow `stat()`. This switches detection to the existing UTF-8 path metadata wrapper, which uses the wide Windows filesystem API.

Adds a Windows regression test covering a non-ASCII profile path containing `.claude`. Production syntax and diff checks pass; related suites passed on the existing baseline binary. The Windows runtime regression, full rebuilt test suite, and lint require follow-up in CI or a suitable Windows/tooling environment.

## Checklist

- [x] Every commit signed off (`git commit -s`) — required, CI rejects unsigned commits ([DCO](../DCO), [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] New behavior covered test (reproduce-first bug fixes)

Fixes #2145


DeusData:
Thank you — this is the right fix, and I checked the part of it most likely to bite before saying so.

**The production change is correct on every platform.** `dir_exists` moving from `stat`/`lstat` to `cbm_path_info_utf8` is the same rule that governs `cbm_fopen`: a path that can come from a user has to go through the UTF-8 wrappers or it breaks on Windows outside the active code page. My one worry was POSIX: the old code deliberately used `lstat` there, so a symlink to a directory was *not* a directory, and `main` recently made symlink-following an explicit per-call-site opt-in. I read the wrapper — its POSIX branch uses `lstat` as well — so that behaviour is preserved exactly. No semantic change on macOS or Linux; on Windows both old and new report a directory.

**One real failure, and it is a one-character fix.** `test / test-windows (…, 2/2)` is red on your own new test:

```
cli_detect_agents_handles_non_ascii_windows_profile
    FAIL tests/test_cli.c:13085: could not create non-ASCII Windows profile path
```

It is not a Unicode problem. `test_mkdirp` (`tests/test_cli.c:718`) is errno-style — it returns **`0` on success** and `-1` on failure:

```c
return cbm_mkdir(tmp) == 0 || errno == EEXIST ? 0 : -1;
```

and the test reads it as a boolean:

```c
if (!test_mkdirp(claude_dir))
    FAIL("could not create non-ASCII Windows profile path");
```

`!0` is true, so the test fails **precisely when the directory was created successfully** — which means your production fix never even got exercised on the runner. `if (test_mkdirp(claude_dir) != 0)` fixes it. An easy one to write; the helper's name reads like a predicate and it is not one.

**The other red is ours, not yours:** `test / test-windows-guards` is a documented setup lottery on our side that reddens unrelated PRs. Please ignore it.

Push the one-line fix and I will re-review promptly — once that test actually runs, it becomes the proof that the production change works on a real non-ASCII Windows profile, which is the part I cannot verify locally today.

PR #2245, opened 2026-09-20
fix(install): reject unsupported Android portable installs

### LINKED ISSUE #1202: Termux/Portable Linux 64 Execution Issues + How It Sort Of Works Now
### Version

Dev build

### Platform

Linux (arm64)

### Install channel

### Binary variant

standard

Hi, I've been installing things on my Android phone via Termux with varying degrees of success.  While there were some issues getting this to work, I was eventually able to get it most of the way there, as in enough to make my Hermes Agent become extremely useful.  Thank you for that, by the way!  On to the issues here: it wasn't the cleanest install, and some dependencies to make it work on e.g. Android are missing from the install /instructions.  With the help of my very cute AI assistant, I've compiled a list of what went wrong, where and when, and what went right.  Feel free to make changes to the build as you see fit! 

Android / Termux codebase-memory-mcp installation and verification report
Date: 2026-07-21
Device: Android 13 aarch64
Termux Python: 3.13.13
Hermes Agent: v0.18.2 (2026.7.7.2)
Kernel: Linux localhost 4.19.113-27095354 #1 SMP PREEMPT Tue Mar 4 15:52:06 KST 2025 aarch64 Android

SUMMARY
- Source build from git succeeded.
- Binary installed to ~/.local/bin/codebase-memory-mcp
- Hermes MCP wiring succeeded; 15 tools discovered.
- Indexing is currently blocked on this device by a reproducible crash in the binary during the dump/artifact phase. MCP tool exposure is fine; graph creation is not usable yet.

REPRO STEPS THAT WORKED
1) Prebuilt installer failed
   - Command: curl -fsSL https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh | bash
   - Behavior: downloads codebase-memory-mcp-linux-arm64-portable.tar.gz, extracts, then crashes with "Bad system call" when running the bundled binary.
   - Root cause notes: The downloaded portable asset is an ELF static binary whose exec path uses a syscall/kernel path that Android 13 / kernel 4.19 / Bionic rejects at exec time. ldd does not reveal the issue; the failure is runtime kernel rejection.

2) Source build succeeded
   - git clone --depth=1 https://github.com/DeusData/codebase-memory-mcp.git
   - cd codebase-memory-mcp
   - bash scripts/build.sh
   - Output included: "Built: build/c/codebase-memory-mcp"
   - Resulting binary: ELF shared object, 64-bit LSB arm64, dynamic (/system/bin/linker64), for Android 30, built by NDK r29 (14206865), not stripped
   - Size: 294057640 bytes (~281 MiB)
   - This build runs, unlike the downloaded portable one.

3) Install
   - mkdir -p ~/.local/bin
   - cp build/c/codebase-memory-mcp ~/.local/bin/codebase-memory-mcp
   - chmod 755 ~/.local/bin/codebase-memory-mcp
   - ~/.local/bin/codebase-memory-mcp --version -> codebase-memory-mcp dev

TERMUX PACKAGES REQUIRED THAT ARE NOT IN THE UPSTREAM INSTALLER
- clang
- cmake
- make
- git
- zlib / zlib1g-dev

These were already present here, but the upstream install.sh assumes a normal Linux/macOS host with toolchains available. On a fresh Termux install you must install them first.

HERMES CONFIG CHANGES
~/.hermes/config.yaml
Added under mcp_servers:
  codebase-memory:
    command: /data/data/com.termux/files/home/.local/bin/codebase-memory-mcp
    args: []
    connect_timeout: 120.0
    timeout: 180

HERMES VERIFICATION RESULTS
- hermes mcp test codebase-memory
  - Connected: yes
  - Tools discovered: 15
- hermes mcp list
  - codebase-memory status: enabled

REPRODUCIBLE CRASH ON INDEXING
A) Real project: /data/data/com.termux/files/home/codebase-memory-mcp
   - Result: worker exited nonzero
   - Relevant logs:
     - level=info msg=parallel.extract.done nodes=22488 errors=0
     - level=info msg=parallel.resolve.done calls=145181 ...
     - ...
     - level=error msg=pipeline.err phase=dump
     - {"project":"...","status":"error","hint":"Pipeline failed. Check repo_path exists and contains source files. Try mode='fast' for a quicker diagnostic run."}
     - {"status":"error","outcome":"exit_nonzero","hint":"Indexing worker crashed on a file. The crash was contained (the server survived). Re-run to retry; a future release isolates the culprit file."}

B) Tiny synthetic test at ~/tmp-cbm-test/main.py with two classes
   - Result: same worker exit_nonzero at phase=dump
   - Relevant logs:
     - level=info msg=parallel.extract.done nodes=... errors=data-data-com.termux-files-home-tmp-cbm-test
     - level=error msg=pipeline.err phase=dump
     - {"status":"error","outcome":"exit_nonzero","hint":"Indexing worker crashed on a file. The crash was contained (the server survived). Re-run to retry; a future release isolates the culprit file."}

Because indexing never completes, list_projects cannot behave normally:
- {"error":"cannot read cache directory: /data/data/com.termux/files/home/.cache/codebase-memory-mcp","hint":"Check directory permissions or run index_repository first."}

CONCLUSION FOR USERS
- This MCP server can be wired into Hermes on Android/Termux and the handshake/tool discovery works.
- Do not rely on index_repository from Hermes sessions yet on this environment. The dump/artifact-write path crashes inside the worker in a contained way, but graph data is not persisted.
- If you only use tools that do not require a finished index, those may still work, but most valuable tools depend on indexing.

WHAT THIS IMPLIES FOR UPSTREAM
- The installer's portable linux-arm64 asset is not executable here because it hits a denied syscall at exec time, despite being labeled portable/static.
- The source build does launch, but another fault appears at artifact dump/write. This is the second distinct failure mode on Android/Bionic, separate from the exec failure.
- Crash isolation logs may not always be written to ~/.cache/codebase-memory-mcp/logs/.worker-*.log in run mode, which makes failure triage difficult for end users on Android/Termux.
- Android/Termux is not listed as an explicit supported target in the installer flow; presumably only macOS/Windows/Linux desktops were intended.

FILES CHANGED OR ADDED 
- ~/.local/bin/codebase-memory-mcp
- ~/.hermes/config.yaml  (added mcp_servers.codebase-memory)
- ~/codebase-memory-mcp/  (source checkout from git clone)
- ~/.cache/codebase-memory-mcp/  (cache directory created, but no usable graph database)

### Reproduction

1. https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh - > fails
2. - git clone --depth=1 https://github.com/DeusData/codebase-memory-mcp.git
   - cd codebase-memory-mcp
   - bash scripts/build.sh - > runs, with caveats.  See above.
3. Install
   - mkdir -p ~/.local/bin
   - cp build/c/codebase-memory-mcp ~/.local/bin/codebase-memory-mcp
   - chmod 755 ~/.local/bin/codebase-memory-mcp
   - ~/.local/bin/codebase-memory-mcp --version -> codebase-memory-mcp dev - > works

### Logs

```text

```

```text

```

_No response_

### Confirmations

lorenzozanee:
## What does PR do?

On Android/Termux, `install.sh` previously selected and downloaded an incompatible Linux portable binary. Detect Android/Termux, report the supported alternatives, and stop before download. Add a regression test that verifies the downloader is not invoked.

Fixes #1202

## Checklist
- [x] Every commit signed off (`git commit -s`) — required, CI rejects unsigned commits ([DCO](../DCO), see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] New behavior is covered by test (reproduce-first bug fixes)


DeusData (changes requested):
Thank you — this is the right fix and the right size. On Termux the installer currently downloads the `portable` asset, which then dies with "Bad system call" — it is statically linked glibc, and it makes system calls Android's seccomp filter rejects; stopping *before* the download, with a message that says why, is exactly what should happen. Detecting on three independent signals (`uname -o`, `TERMUX_VERSION`, the `$PREFIX` path) is thorough, guarding it behind `OS = linux` keeps every other platform untouched, and the test is nicely built — stubbing `curl` to exit 99 proves "rejected before download" rather than merely "rejected".

Three small things, and then I would like to merge it:

**1. The test will never run as it stands.** `scripts/test.sh` does not glob the `tests/` directory — it lists each shell contract test explicitly (`bash "$ROOT/tests/test_build_dir_safety.sh"` and so on, around lines 200–270), and every one of them is named `test_*.sh` with underscores. `tests/test-install-android-guard.sh` is neither listed nor named to the convention, so no lane would ever execute it, and a guard nobody runs protects nothing. Please rename it to `tests/test_install_android_guard.sh` and add one line for it next to the other contract tests in `scripts/test.sh`. (It would be the first installer test in that list — a good precedent.)

**2. `Fixes #1202` → `Refs #1202`, please.** #1202 is much wider than this: it also covers the gaps in building from source under Termux and an `ENOENT` on the temporary database that is still unexplained. Merging with `Fixes` would auto-close all of that. This PR resolves the first failure mode in that issue — cleanly — and should say so without closing the rest.

**3. One sentence in the error message promises slightly more than we can.** "Build from source with scripts/build.sh in Termux" reads as a supported path; the reporter on #1202 got it mostly working, but with open problems, and Termux support is an undecided question here. Something like

```
error: Android/Termux cannot run the Linux portable binary (static glibc; Android kills it with "Bad system call").
Building from source in Termux may work but is not a supported platform yet - see issue #1202.
```

tells the user the truth and gives them the thread where the workarounds are collected. Your test greps for `Android/Termux`, so it keeps passing.

Two things that are **ours, not yours**:

- The red `test-msan` you saw earlier was `daemon_bootstrap_fails_fast_when_daemon_dies_at_publication` — a daemon test that has nothing to do with `install.sh`. I re-ran it on the identical commit and it passed, which makes it a nondeterministic test on our side; it is recorded as such and will not be held against this PR.
- `lint / lint` is currently red on `main` itself for an unrelated reason, and a fix is in flight (#2257). If you rebase after that lands you will see a clean run.

A note on process, since you are new here: force-pushing your own PR branch is completely fine, and keep the `Signed-off-by` on any amended commit. Thanks again — small, careful fixes to the installer are some of the most valuable contributions this project gets, because that script is the first thing every new user runs.

PR #2261, opened 2026-09-20
Docs/contributor check prerequisites

tamerbak:
## Checklist


DeusData (changes requested):
Thank you — this is a gap every new contributor falls into: the README lists what you need to *build*, and then `scripts/test.sh` stops halfway through because something it never mentioned is missing. Writing that down where people look first is a real improvement, and keeping it to three lines is the right instinct.

Two small corrections and it can go in:

**1. The CONTRIBUTING.md line contains literal `\n\n`.** The new text was added to the existing line as

```
... Optional: Node.js 22+ (for graph UI).\n\n**Contributor check prerequisites**: `scripts/test.sh` also requires ...
```

Markdown does not interpret `\n`, so this renders as a backslash-n in the middle of the sentence instead of a paragraph break. It needs real blank lines between the two paragraphs.

**2. `curl` — can you point me at what needs it?** I checked every script `scripts/test.sh` invokes: `python3` is used by nineteen of the contract tests and `zip` by seven, so both of those claims hold. I could not find a `curl` call in any of them. If you hit a concrete failure without it, tell me which step (it may come from somewhere I did not look, e.g. inside the C test runner's fixtures) and it stays; if it was an assumption, please drop it — a prerequisites list is only useful if every entry is true, because people install what it tells them to. The `lint.sh` half (`cppcheck`, `clang-format`, `clang-tidy` for the full mode) checks out as written.

One optional suggestion: the README blockquote and the CONTRIBUTING paragraph now say the same thing in two places, which means two places to update. A single sentence in the README pointing to the CONTRIBUTING section would keep one source of truth — your call.

Housekeeping: the PR description is still the empty template. A two-line description ("what" and "why") helps whoever reads the history later; no need to tick the test/lint boxes for a docs-only change.

`lint / lint` is currently red on `main` itself for an unrelated reason (fix in flight, #2257) — not yours, and a rebase after it lands will show a clean run. Thanks for the careful, small PRs.

Issue #1168, opened 2026-07-18
Can't upgrade codebase-memory-mcp on windows platform.  Can't uninstall.


spam0115:
### Version

codebase-memory-mcp 0.8.1

### Platform

Windows (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

after running codebase-memory-mcp upgrade, the upgrade goes through the motions and even downloads something.  However, running "codebase-memory-mcp --version" afterwards shows that the version was stuck at 0.8.1.  

 Directory of C:\Users\...\AppData\Local\Programs\codebase-memory-mcp

06/30/2026  20:16    <DIR>          .
07/16/2026  21:44    <DIR>          ..
06/12/2026  06:13       269,067,776 codebase-memory-mcp.exe
06/12/2026  06:13       269,067,776 codebase-memory-mcp.exe.old

I had to delete the executable manually and then run "Invoke-WebRequest -Uri https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.ps1 -OutFile install.ps1" to finally upgrade it.

### Reproduction

1.  run "codebase-memory-mcp upgrade" on a windows machine.

### Logs

```text

```

### Diagnostics trajectory (memory / performance / leak issues)

```text

```

### Confirmations

- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

DeusData:
Thanks for the precise upgrade reproduction. I have placed this in `0.9.1-rc` at high priority. Review needs to cover Windows replacement of a running executable, the `.exe.old` fallback, and which binary `PATH` resolves after the upgrade reports success.

DeusData:
@spam0115 — the directory listing is what makes this unambiguous: `codebase-memory-mcp.exe` and `.exe.old` at identical sizes and timestamps after an `upgrade` that reported success. An upgrade that downloads, claims to have worked, and quietly leaves you on the old binary is worse than one that fails loudly.

Having to delete the executable by hand and re-run `install.ps1` is not something anyone should have to discover for themselves, and I am sorry that was the only way through.

[...]

Issue #1167, opened 2026-07-18
fails to install for opencode on windows (other os's unknown)


spam0115:
### Version

codebae-memory-mcp 0.9

### Platform

Windows (x64)

### Install channel

npm

### Binary variant

standard

### What happened, and what did you expect?

The install script finds claude, gemini, codex, etc but not opencode.  I have both opencode cli and app installed.  Tried this on two computers and the bug has been around for more than just this version.  Not sure if this is windows specific or if it affects other oses too.

### Reproduction

1.  Run codebase-memory install

### Logs

```text

```

### Diagnostics trajectory (memory / performance / leak issues)

```text

```

### Confirmations

- [x] I searched existing issues and this is not a duplicate.
- [x] My reproduction uses shareable code (a dummy snippet or a public OSS repository), not proprietary code.

DeusData:
Thanks for the Windows reproduction across two machines. I have placed this in `0.9.1-rc` as a high-priority installer/integration bug and linked it under #395. We will verify detection for both the OpenCode CLI and desktop app rather than inferring installation from unrelated configuration state.

DeusData:
Thanks @spam0115 — two machines, both the OpenCode CLI and the app installed, and your note that this predates 0.9 all matched what later reports showed. Your report was right and our `PATHEXT` fix was not the whole story.

Nothing here has been waiting on you. Honest status: the Windows detection path still has at least one live failure mode, and the fix needs to distinguish the actual install source and shim shape instead of assuming an extension probe finds it.

Tonight's v0.10.5 is being cut for the issues that make cbm unusable, and an installer that silently skips the agent you work in qualifies. Sorry it has taken this long for something you reported cleanly the first time.

Issue #1165, opened 2026-07-18
Windows: CBM_CACHE_DIR / CBM_ALLOWED_ROOT are lossily decoded — every non-ASCII char becomes U+FFFD, so the server silently uses a different path (v0.9.0)


sowa-content:
# Windows: `CBM_CACHE_DIR` / `CBM_ALLOWED_ROOT` are lossily decoded — every non-ASCII char becomes U+FFFD, so the server silently uses a different path (v0.9.0)

## Summary

On Windows v0.9.0, the values of `CBM_CACHE_DIR` and `CBM_ALLOWED_ROOT` are decoded
lossily: **each non-ASCII character is replaced with U+FFFD** (REPLACEMENT CHARACTER).
The server then operates on that mangled path instead of the one you set.

One defect, three very different-looking symptoms:

| Symptom | Underlying cause |
|---|---|
| `config set` reports success but nothing persists | DB written to the U+FFFD twin directory |
| `error: cannot open config database` | mangled path is not creatable (e.g. under `C:\Users\`) |
| every `index_repository` fails, even in-root | mangled allowed-root never matches the real repo path |

CLI arguments are **not** affected — `--repo-path` with the same characters works
correctly and encodes them properly into the project name (`ö` → `c3b6`). The bug is
specific to reading these environment variables.

This is distinct from #903 / #571, which concern the *repository path* (there indexing
succeeds and only retrieval/naming degrades).

## Environment

- codebase-memory-mcp **v0.9.0**, `codebase-memory-mcp-windows-amd64.zip` (standard variant)
- Windows 11 Pro 10.0.26200, AMD64
- Reproduced identically from both PowerShell 5.1 and Git Bash — not a shell-quoting artifact

## Proof: the tool creates a mojibake twin directory

Set the cache dir to a path with non-ASCII characters, then look at what appeared on disk:

```powershell
$env:CBM_CACHE_DIR = "C:\Tëst-Ünïcode\cache"
codebase-memory-mcp config set auto_watch false     # → "auto_watch = false"
codebase-memory-mcp config list                     # → auto_watch = false
```

Both commands report success. But `C:\Tëst-Ünïcode\cache` is never created. Instead:

```powershell
Get-ChildItem C:\ -Directory | Where-Object Name -like "T*st*code" |
  ForEach-Object { ($_.Name.ToCharArray() | % { "U+{0:X4}" -f [int]$_ }) -join " " }

# the directory I created:
[...]

Issue #1161, opened 2026-07-18
Installer targets legacy Gemini-CLI settings.json instead of Antigravity 2.0 config directory (resulting in duplicate 273MB binaries)


uongsuadaubung:
### Version

codebase-memory-mcp 0.9.0

### Platform

Windows (x64)

### Install channel

GitHub release archive / install.sh / install.ps1

### Binary variant

standard

### What happened, and what did you expect?

When running `codebase-memory-mcp install` (or using the Windows `install.ps1` script), the installer detects and configures the outdated/legacy agent **`Gemini-CLI`** rather than **`Antigravity`** (the current successor for Antigravity 2.0).

Consequently, it writes the configuration to the deprecated monolithic file **`C:\Users\<Username>\.gemini\settings.json`**.
In Antigravity 2.0, the configuration format is modular:
- MCP server configurations must reside in `~/.gemini/config/mcp_config.json`
- Hooks configurations must reside in `~/.gemini/config/hooks.json`

Because it targets the legacy file `~/.gemini/settings.json`, the newly installed MCP server and hooks are completely ignored by the Antigravity 2.0 agent.

Additionally, on Windows, this outdated installation process creates a duplicate binary:
1. `C:\Users\<Username>\AppData\Local\Programs\codebase-memory-mcp\codebase-memory-mcp.exe` (installed by `install.ps1`)
2. `C:\Users\<Username>\.local\bin\codebase-memory-mcp.exe` (copied by the binary's `install` command)
Since the binary is ~273 MB, this wastes disk space, and neither copy is recognized by Antigravity 2.0 out-of-the-box because of the incorrect path configuration.

---

## Root Cause Analysis
Inside the `codebase-memory-mcp` binary's `install` command logic:
1. It looks for and writes to the legacy Gemini CLI agent path: `~/.gemini/settings.json`.
2. It has not been updated to detect **`Antigravity`** and write sparse configurations to the modern configuration directory `~/.gemini/config/` (specifically `mcp_config.json` and `hooks.json`).

---

## Expected Behavior
1. The `install` command should detect the modern **`Antigravity`** agent (IDE & CLI).
2. It should write the MCP server configuration to `~/.gemini/config/mcp_config.json` instead of `~/.gemini/settings.json`.
[...]

PR #2262, opened 2026-09-20
Docs/windows troubleshooting

tamerbak:
## Checklist


DeusData (changes requested):
Thank you — a Windows troubleshooting page is something this project should have, and you built it the right way: every entry names the version it was reported on, points at its tracking issue, and the page ends by telling future editors to prune it rather than let it become a second tracker. We want this page. It needs one pass first, and the reason for that pass is our fault, not yours.

**All seven issues you cite are still open — but several of them are open only because we never closed them.** You could not have known that from the tracker. I checked each entry against the code at `v0.11.0`, and as written the page would tell a v0.11.0 user that bugs fixed two months ago are current. Here is what I found, with replacement wording you can take as is. Everything below is a code-level verdict (commits, file locations, guard tests); none of it was re-run on a Windows machine, and I say so where that matters.

**1. Non-ASCII `CBM_CACHE_DIR` / `CBM_ALLOWED_ROOT` (#1165) — root cause fixed in v0.10.0.** Environment variables are now read wide (`GetEnvironmentVariableW`, faf5160c) and `CBM_ALLOWED_ROOT` is canonicalised per session (0e00ef57); `tests/windows/test_non_ascii_cache_dump.py` guards a non-ASCII cache directory end to end. A few fallback paths still use the narrow read, so I am asking the reporter for a v0.11.0 retest rather than closing it blind. Suggested entry:
> **Affects v0.9.x and earlier.** Fixed in v0.10.0 (environment variables are read as wide strings). On an older build, use ASCII paths for these two variables or upgrade with `install.ps1`. Tracking: #1165.

**2. Upgrade completes but the version does not change (#1168) — fixed in v0.10.0.** In-process self-replacement on Windows was removed (a54ea957): `update` now only prints the `install.ps1` command, and `install.ps1` retires the locked executable and verifies `--version` afterwards. `tests/windows/test_windows_update_handoff.py` guards it. I am closing #1168 with that commit. Suggested entry:
> **Affects v0.9.x and earlier** (in-process self-update). From v0.10.0 `update` prints the `install.ps1` command instead of replacing itself. On an older build, run `install.ps1` once; it replaces the binary and verifies the version.

**3. Worker hangs, crashes, or leaves an empty `.worker-*.log` (#1130, #1132, #1145) — partly fixed.** The *empty log* is fixed in v0.10.5 (87717b0a: the Windows CRT fully buffers a redirected stderr; the log now always carries a flushed run header and one line per file as it starts), and the misleading "crashed on a file" report for what was really a returned error is fixed in v0.10.0 (8fc04c0b). From v0.11.0 the Windows worker is memory-capped. Whether the underlying hang (#1130) and the startup exit (#1145) still occur is **not something I can tell from code** — those stay open. Your `.cbmignore` workaround is correct and documented (`docs/cbmignore.md`). Suggested entry:
> **Symptom:** `index_repository` reports worker hangs or crashes. (An *empty* worker log is a separate, fixed bug: v0.10.4 and earlier.) From v0.10.5 the worker log always names the run and the files that were in flight; from v0.11.0 the Windows worker is memory-capped. For trees with large binary/media/model directories, exclude them with `.cbmignore`. If indexing still fails on v0.11.0, attach that log to #1130 or #1145 — it now says which file was being processed.

**4. Antigravity 2.0 configuration not picked up (#1161) — the framing needs correcting.** The installer does not target the legacy Gemini-CLI file when Antigravity is detected; even v0.9.0 wrote `~/.gemini/config/mcp_config.json` in that case. The reporter's machine was *detected as Gemini-CLI*, because detection is only "`~/.gemini/antigravity-cli` exists, or `antigravity` is on PATH". The duplicate-binary half is fixed for the `install.ps1` flow since v0.10.0. "Verify the configuration files Antigravity is actually loading" does not give a user anything to do. Suggested entry:
> Antigravity is configured only when `~/.gemini/antigravity-cli` exists or `antigravity` is on PATH; the target is `~/.gemini/config/mcp_config.json`, and no hooks are installed for it. If neither signal is present the machine is treated as Gemini-CLI. From v0.10.0 `install.ps1` keeps a single binary; running `codebase-memory-mcp install` by hand without `--dir` can still create a second copy in `~/.local/bin`. Tracking: #1161.

**5. OpenCode installed but not detected (#1167) — the workaround is factually wrong, and that one is worth fixing even if nothing else were.** The README's "Manual MCP Configuration" section covers Claude only; there is no manual OpenCode configuration in it to point at. What did change: from v0.10.0 OpenCode is also detected through its *config* (an existing `opencode.json(c)`, the `~/.config/opencode` directory, or `OPENCODE_CONFIG_DIR`), not only through PATH (a3903caa). The PATH probe still has real Windows weaknesses, so the issue stays open. Suggested entry:
> From v0.10.0 OpenCode is detected through its configuration as well as PATH. Make sure `%USERPROFILE%\.config\opencode\` exists (or set `OPENCODE_CONFIG` / `OPENCODE_CONFIG_DIR`), then re-run `codebase-memory-mcp install`. PATH-only detection can still fail on Windows (very long PATH, non-ASCII profile, npm's `%APPDATA%\npm` shim directory). Tracking: #1167.

Two small things on top: the intro sentence "confirmed … symptoms that are still tracked" should become something like "known Windows-specific symptoms, with the version range each affects" once the entries carry ranges; and the PR description is still the empty template — two lines of "what" and "why" are enough.

This is a bigger rewrite than a docs PR usually gets asked for, and I want to be clear that it reflects how much the ground moved under those issues between v0.9 and v0.11, not a problem with your work. If you would rather not redo it, say so and I will apply these edits on top of your commits with you as the author of the page. Thank you for taking the trouble to collect these in one place — it also just made us close a stale issue.

PR #2290, opened 2026-09-22
fix(windows): set UTF-8 environment values through wide CRT

knewstimek:
`cbm_setenv` already converts its UTF-8 input to UTF-16, but then passes the original bytes to `_putenv_s`. That call can reject byte sequences that are invalid in the active ANSI code page. Use `_wputenv_s` for the CRT update and keep `SetEnvironmentVariableW` for the process environment.

A narrow `getenv` of the variable now returns the CRT's active-code-page representation of the wide value, rather than the original UTF-8 bytes. Callers that need UTF-8 can use `cbm_safe_getenv`.

The Windows platform test now uses a Unicode value whose UTF-8 bytes include `0x81` and checks the wide environment, `cbm_safe_getenv`, and narrow `getenv`. In a local LLVM-MinGW/UCRT reproduction with code page 949, the old function returned 42 and left the wide variable unset; this change returned 0, preserved the wide value, and made `getenv` agree with `CP_ACP` conversion.

Split from #2268.


knewstimek:
`test-diag` failed in the unchanged Linux `pipeline_python_cross_module_call` test (`tests/test_pipeline.c:8060`, missing `CALLS` edge). This PR changes only the Windows `cbm_setenv` branch and its Windows test. I tried rerunning the failed jobs, but GitHub requires repository admin rights. Could you rerun them? I can investigate further if it repeats.

